PR #1727, opened 2020-06-16
OBPIH-2977 Fix expiration date display on packing list and receive pages

DominikaHatala:


grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -143,7 +143,7 @@ class ReceiptService {
             partialReceiptItem.binLocation = binLocation
         }
         partialReceiptItem.lotNumber = shipmentItem.inventoryItem?.lotNumber
-        partialReceiptItem.expirationDate = shipmentItem.inventoryItem?.expirationDate
+        partialReceiptItem.expirationDate = shipmentItem.requisitionItem?.expirationDate ?: shipmentItem.inventoryItem?.expirationDate
```
jmiranda, line 146: This is not how I would expect this to work. I think if we're storing expiration date on the requisition item it would be temporarily for that page. But the value should actually be set on the inventory item. I need to discuss with Kelsey though because we may not want to override the expiry date but instead create a new inventory item.

jmiranda:
Ticket was closed in October 2020 so closing PR as abandoned.
https://pihemr.atlassian.net/browse/OBPIH-2977

PR #1746, opened 2020-06-23
OBPIH 2956 Add breadcrumbs to react pages

julienreliefapps:


src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,70 @@
...
+    _.forEach(listItems, (name) => {
+      // Put a space before each uppercase
+      let nameFormated = name.replace(/([A-Z])/g, ' $1').trim();
+      // Put an uppercase at the beginning of each word
+      nameFormated = nameFormated.replace(/(^\w{1})|(\s{1}\w{1})/g, match => match.toUpperCase());
+      listItemFormated.push(nameFormated);
+    });
+    const listToReturn = listItemFormated.map((name, id) =>
+      (name === 'Openboxes' || name === '' ? null : (
+        <a key={`item-${id}`} href={window.location.pathname.split('/', id + 1).join('/')} className="item-breadcrumbs"> {name}
```
awalkowiak, line 35: Move `{name}` to the next line for better readability (here it goes out of the screen and was misleading for me)

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,70 @@
...
-);
+class Breadcrumbs extends Component {
+  constructor(props) {
+    super(props);
 
-export default Breadcrumbs;
+    this.openModal = this.openModal.bind(this);
+  }
+
+  componentDidMount() {
```
awalkowiak, line 15: Why there is this empty lifecycle method?

src/js/components/location/LocationChooser.jsx
```diff
@@ -113,13 +112,24 @@ class LocationChooser extends Component {
...
+  changeCurrentLocation,
+  showModal,
+  hideModal,
+})(LocationChooser);
 
 LocationChooser.propTypes = {
   /** Function called to change the currently selected location */
   changeCurrentLocation: PropTypes.func.isRequired,
+  // Boolean to show modal or not
+  modal: PropTypes.bool.isRequired,
```
awalkowiak, line 128: Should be a more fitting name for a boolean.

src/js/components/location/LocationChooser.jsx
```diff
@@ -113,13 +112,24 @@ class LocationChooser extends Component {
 const mapStateToProps = state => ({
   currentLocationName: state.session.currentLocation.name,
   defaultTranslationsFetched: state.session.fetchedTranslations.default,
+  modal: state.session.modal,
```
awalkowiak, line 115: Should be a more fitting name for a boolean.

src/js/reducers/sessionReducer.jsx
```diff
@@ -49,6 +51,7 @@ const initialState = {
   minimumExpirationDate: '',
   isPaginated: false,
   logoLabel: '',
+  modal: false,
```
awalkowiak, line 54: isOpen (or isLocationChooserOpen to indicate what modal is open) sounds better for boolean

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,70 @@
...
+
+  render() {
+    const listItems = window.location.pathname.split('/');
+    const listItemFormated = [];
+    _.forEach(listItems, (name) => {
+      // Put a space before each uppercase
+      let nameFormated = name.replace(/([A-Z])/g, ' $1').trim();
+      // Put an uppercase at the beginning of each word
+      nameFormated = nameFormated.replace(/(^\w{1})|(\s{1}\w{1})/g, match => match.toUpperCase());
+      listItemFormated.push(nameFormated);
```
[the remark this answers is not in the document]
julienreliefapps: The windows.location.pathname doesn't include the parameters, so for this url : http://localhost:8080/openboxes/stockMovement/createOutbound?direction=OUTBOUND

This is what I get : /openboxes/stockMovement/createOutbound

So we I think that we shouldn't have problems in the name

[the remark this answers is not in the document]
awalkowiak: You're right. It's a pathname, it should be fine.

src/js/reducers/sessionReducer.jsx
julienreliefapps: Done

src/js/components/location/LocationChooser.jsx
julienreliefapps: Done

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,70 @@
...
+  componentDidMount() {
+
+  }
+
+  openModal() {
+    this.props.showModal();
+  }
+
+  render() {
+    const listItems = window.location.pathname.split('/');
```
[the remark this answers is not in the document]
julienreliefapps: I have seen that in the grails part, it was done using the current controller and current action. As our URL are in this format controller/action, I think it will work well. 
But it's a proposition, this approach hasn't been approved yet @jmiranda

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,70 @@
...
+    const listItemFormated = [];
```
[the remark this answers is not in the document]
julienreliefapps: Done

jmiranda:
@awalkowiak let me know if this look ok now?

awalkowiak:
@jmiranda I approved it, but now I noticed that there is no translation option atm

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,67 @@
...
-export default Breadcrumbs;
+    this.openModal = this.openModal.bind(this);
+  }
+
+  openModal() {
+    this.props.showModal();
+  }
+
+  render() {
+    const listItems = window.location.pathname.split('/');
```
jmiranda, line 20: @julienreliefapps Yeah I was really hoping for the breadcrumbs to be pulled from the API in order to handle more complex requirements like translations on the backend. But if you can make this work with translations on the client-side then be my guest.

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,67 @@
...
+    const listItemFormatted = [];
+    _.forEach(listItems, (name) => {
+      // Put a space before each uppercase
+      let nameFormatted = name.replace(/([A-Z])/g, ' $1').trim();
+      // Put an uppercase at the beginning of each word
+      nameFormatted = nameFormatted.replace(/(^\w{1})|(\s{1}\w{1})/g, match => match.toUpperCase());
+      listItemFormatted.push(nameFormatted);
+    });
+    const listToReturn = listItemFormatted.map((name, id) =>
+      (name === 'Openboxes' || name === '' ? null : (
```
jmiranda, line 30: I don't love the hard-coded OpenBoxes here. Can you explain this?

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,67 @@
...
+        <a key={`item-${id}`} href={window.location.pathname.split('/', id + 1).join('/')} className="item-breadcrumbs">
+          {name}
+          <img className="item-breadcrumbs" alt="/" src="/openboxes/images/bc_separator.png" />
```
jmiranda, line 33: `/openboxes` needs to be configurable since the app will eventually need to be deployed under any context path.

src/js/components/Layout/Breadcrumbs.jsx
julienreliefapps: yes you are right I will pull it from the config

src/js/components/Layout/Breadcrumbs.jsx
julienreliefapps: Same as answered above

src/js/components/Layout/Breadcrumbs.jsx
julienreliefapps: Ah my bad I really didn't notice about the translation, and @jmiranda in this case, it's true that it may be easier pulling it from the backend

jmiranda: It actually might even be just as easy to render the breadcrumbs as a global javascript variable on the page since I'm realizing that getting it from the API might be tough to do. For example, in order for an endpoint to resolve the breadcrumbs you'll need to pass the full URL, which isn't terrible. Just seems odd. 

Alternatively, maybe the page itself (_react.gsp) can render a list of breadcrumbs as a global variable?

@awalkowiak Thoughts?

src/js/components/Layout/Breadcrumbs.jsx
```diff
@@ -1,11 +1,67 @@
...
+    _.forEach(listItems, (name) => {
+      // Put a space before each uppercase
+      let nameFormatted = name.replace(/([A-Z])/g, ' $1').trim();
+      // Put an uppercase at the beginning of each word
+      nameFormatted = nameFormatted.replace(/(^\w{1})|(\s{1}\w{1})/g, match => match.toUpperCase());
+      listItemFormatted.push(nameFormatted);
+    });
+    const listToReturn = listItemFormatted.map((name, id) =>
+      (name === 'Openboxes' || name === '' ? null : (
+        <a key={`item-${id}`} href={window.location.pathname.split('/', id + 1).join('/')} className="item-breadcrumbs">
```
jmiranda, line 31: What is ${id}? Is that a GSP variable?

julienreliefapps: This is the id of the name in the list in order to not have duplicate key for the <a>

jmiranda: Oh nevermind. I didn't realize you could do that kind of string interpolation in javascript.

src/js/components/Layout/Breadcrumbs.jsx
julienreliefapps: @jmiranda  should I use the translationService ?

PR #1759, opened 2020-07-01
OBPIH-2893 Add request type to outbound stock movement

DominikaHatala:


src/js/components/stock-movement-wizard/outbound/CreateStockMovement.jsx
```diff
@@ -174,6 +197,16 @@ class CreateStockMovement extends Component {
     }
   }
 
+  setRequestType(values, stocklist) {
+    console.log(values);
```
awalkowiak, line 201: Remove console.log

PR #1764, opened 2020-07-06
OBPIH-3023 Improve pagination to make fewer requests

DominikaHatala:


src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -117,15 +118,17 @@ class TableBodyVirtualized extends Component {
...
 
     return (
       <div>
         <InfiniteLoader
           loadMoreRows={loadMoreRows}
           isRowLoaded={isRowLoaded}
-          rowCount={totalCount || 50}
+          rowCount={totalCount}
+          minimumBatchSize={minimumBatchSize}
+          threshold={30}
```
awalkowiak, line 131: @DominikaHatala Perhaps this also could be configurable in config file?

DominikaHatala: It's a good idea to have minimum batch size configurable(now it's set to 50 but maybe we will think that 30 or 100 could be better) but I think there is no sense to make threshold configurable, this property should be constant.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -91,6 +91,7 @@ class ApiController {
         DateFormat dateFormat = new SimpleDateFormat("MM/DD/YYYY");
         String minimumExpirationDate = dateFormat.format(grailsApplication.config.openboxes.expirationDate.minValue)
         def logoLabel = grailsApplication.config.openboxes.logo.label
+        def minimumBatchSize = grailsApplication.config.openboxes.api.pagination.minimumBatchSize
```
jmiranda, line 94: Change the configuration property to pageSize.

grails-app/conf/Config.groovy
```diff
@@ -1135,6 +1135,7 @@ openboxes.receiving.receivingLocation.prefix = Constants.DEFAULT_RECEIVING_LOCAT
 
 // Pagination
 openboxes.api.pagination.enabled = true
+openboxes.api.pagination.minimumBatchSize = 50
```
jmiranda, line 1138: Set this to 10 by default. And get rid of the initial value of 30. I don't care if the user sees "Loading..." under the first 10 rows, I just want them to see those rows as quickly as possible.

src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -117,15 +118,22 @@ class TableBodyVirtualized extends Component {
 
   render() {
     // eslint-disable-next-line max-len
-    const { properties } = this.props;
-    const { totalCount, loadMoreRows, isRowLoaded } = properties;
+    const { properties, minimumBatchSize } = this.props;
+    const {
+      totalCount, loadMoreRows, isRowLoaded, isFirstPageLoaded,
+    } = properties;
+    const threshold = 30;
```
jmiranda, line 125: Again, use the default pageSize always.

DominikaHatala:
@jmiranda I have updated PR, applied all requested changes and also fixed pagination on all other workflows.

PR #1780, opened 2020-07-10
OBPIH-3047 Fix outbound stock movements download

DominikaHatala:


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -365,7 +365,7 @@ class StockMovementService {
 
         log.info "Stock movement: ${stockMovement?.shipmentStatusCode}"
 
-        def requisitions = Requisition.createCriteria().list(max: params.maxResults?:10, offset: params.offset?:0) {
+        def requisitions = Requisition.createCriteria().list(max: params.max, offset: params.offset) {
```
awalkowiak, line 368: Why you got rid of fallback options?

DominikaHatala: @awalkowiak If max or offset are null they should not be set to any other value because while downloading stock movements we need all results.

PR #1793, opened 2020-07-16
OBPIH-2822 Fix transaction report to not include transactions for end…

DominikaHatala:
… date + 1

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
@@ -678,7 +678,7 @@ class InventorySnapshotService {
                 'in'("transactionCode", transactionCodes)
             }
             transactionDateKey {
-                between("date", startDate, endDate+1)
+                between("date", startDate, endDate)
             }
             locationKey {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -699,7 +699,7 @@ class InventorySnapshotService {
 
     def getTransactionReportDetails(Location location, List<Category> categories, List<Tag> tags, List<ProductCatalog> catalogs, Date startDate, Date endDate) {
 
-        def transactionData = getTransactionReportData(location, startDate, endDate)
+        def transactionData = getTransactionReportData(location, startDate, endDate - 1)
```
awalkowiak, line 702: If you go to the getTransactionReportData, there is:
`transactionDateKey {
                between("date", startDate, endDate+1)
            }`
How about removing the `+1` there instead of doing the `- 1` here (and in the line below)?

DominikaHatala: @awalkowiak Thanks, that's right!

PR #1801, opened 2020-07-21
OBPIH-2958: Add mechanism to get handling icon to product

awalkowiak:


grails-app/taglib/org/pih/warehouse/ProductTagLib.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    def container = { attrs, body -> }
+
+    def renderHandlingIcons = { attrs, body ->
+        if (!attrs.product) {
+            out << ""
+        } else {
+            def iconElements = ""
+            def handlingIcons = attrs.product.getHandlingIcons()
+            handlingIcons.each { it ->
```
jmiranda, line 22: This is fine. The other I was thinking about doing this would have been to put all of the HTML into a GSP template file
```
out << g.render(template: '/taglib/handlingIcons', model: [attrs: attrs])
```
And then in /taglib/_handlingIcons.gsp you would just add 
```
<g:each var="handlingIcon" in="${product.handlingIcons}">
   <i class='fa ${handlingIcon.icon}' style='color: ${handlingIcon.color?:'inherit'}' ></i>&nbsp;
</g:each>
```

PR #1805, opened 2020-07-22
OBPIH-2958: Add new handling icons on product page

awalkowiak:


grails-app/views/product/edit.gsp
```diff
@@ -18,6 +18,10 @@
         #category_id_chosen {
             width: 100% !important;
         }
+        .handling-icon-2x > i > svg {
```
jmiranda, line 21: This is fine, but why can't we just add fa-2x to the icon attribute?

awalkowiak: You're right. I thought that it has to be with 'fa' class to be working and this is just font-size: 2em. I'll get rid of this class with my next changes.

awalkowiak: But instead of on icon, it should be on the outer element that is wrapping <g:renderHandlingIcons ... />

PR #1807, opened 2020-07-22
OBPIH-1879 MySQLTransactionRollbackException: Lock wait timeout excee…

jmiranda:
…ded; try restarting transaction


jmiranda:
Didn't work as expected, so closing.

PR #1812, opened 2020-07-27
OBPIH-2827 Disable supplier edition after adding items to PO

DominikaHatala:


DominikaHatala:
@jmiranda Error message that is added in here is not displayed and I have no idea why. Could you help me and take a look at this one?

awalkowiak:
I approved before I saw the comment that this solution does not work.

awalkowiak:
@DominikaHatala what happens if you add on("error").to("enterOrderDetails")?

DominikaHatala:
@awalkowiak it doesn't change anything

grails-app/controllers/org/pih/warehouse/order/PurchaseOrderWorkflowController.groovy
```diff
@@ -54,13 +54,18 @@ class PurchaseOrderWorkflowController {
         enterOrderDetails {
             on("next") {
                 log.info "Save order details " + params
-                flow.order.properties = params
-                try {
-                    if (!orderService.saveOrder(flow.order)) {
+                if (flow.order.orderItems && flow.order.origin.id != params.origin.id) {
```
jmiranda, line 57: The one issue I have with this is that it's not going to be applied whenever we save an order. It would only be applied when we run through this exact code. So I'm going to work on adding this as a constraint to the Order class.

PR #1814, opened 2020-07-27
OBPIH-3081: Add handling icons on outbound stock movement

awalkowiak:


awalkowiak:
@jmiranda I had to add some product object on some api items objects (eg. available items, substitution items, StockMovementItem, EditPageItem and couple more), but i have not removed the fields like 'productName', 'productCode' and so on to avoid any potential regression. These fields could be removed, but I think I'd need to refactor it in a separate task. What do you think about it? Or you'd like to have it refactored entirely in this PR?

jmiranda:
@awalkowiak Yeah that seems fine to me. Remove the attributes from the stock movement API, but leave them in product API just in case someone is depending on those.

PR #1824, opened 2020-07-30
OBPIH-3060 Improve add adjustment to PO tab

DominikaHatala:


grails-app/views/purchaseOrderWorkflow/purchaseOrder/showOrderItems.gsp
```diff
@@ -320,6 +321,8 @@
             success: function () {
               clearOrderItems();
               loadOrderItems();
+              $('#orderItems').html('<option></option>').trigger('change');
```
awalkowiak, line 324: What is this empty option for? I think that noSelection functions as an empty option.

DominikaHatala: It's used on item deletion to clear options in select. Otherwise, deleted item is still visible in dropdown because select doesn't remove it.

jmiranda: Actually I noticed that when I add items and then go to the adjustments tab the Order Items dropdown is empty. Which makes sense because we don't explicitly refresh it. You have to refresh the page in order for it to populate. Can you test whether this refresh would refresh under that use case as well? It seems it should but if not let's try to find a way to refresh it.

DominikaHatala: @jmiranda Refresh of the page helps in both cases - adding and deleting items. In this PR I have added a way to refresh this select after adding/deleting items. It work for both cases but in case of deleting I had to add this line above to clear options first.

grails-app/views/purchaseOrderWorkflow/purchaseOrder/showOrderItems.gsp
```diff
@@ -523,6 +527,29 @@
           });
         }
 
+        function updateOrderItems() {
```
awalkowiak, line 530: @DominikaHatala I approved it, but lets adjust this naming. Because now we have loadorderItems and updateorderitems and the naming is very vague. Update this to something more descriptive.

awalkowiak: @DominikaHatala i guess the best case would if the items were updated for select during loadOrderItems, but if its problematic, then let's distinghuish it with proper naming.

PR #1827, opened 2020-08-05
OBPIH-3066: Import lot number into shipment (from PO)

awalkowiak:


grails-app/views/order/shipOrder.gsp
```diff
@@ -141,7 +162,7 @@
 													   name="shipOrderItems[${status}].quantityToShip"
 													   min="${shipOrderItem.quantityMinimum}"
 													   max="${shipOrderItem?.quantityMaximum}"
-													   value=""/>
+													   value="${shipOrderItem?.quantityToShip ?: ''}"/>
```
DominikaHatala, line 165: It should be "value=""/>" so it defaults to blank.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -755,4 +789,199 @@ class OrderService {
...
+                        shipmentItem.removeFromOrderItems(orderItem)
+                    }
+                    shipment.removeFromShipmentItems(shipmentItem)
+                    shipmentItem.delete()
+                }
+            }
+        }
+
+        importedLines.each { line ->
+            if (line.quantityToShip > 0) {
```
DominikaHatala, line 938: It should be if ((line.quantityToShip as int) > 0)

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -755,4 +789,199 @@ class OrderService {
...
+                OrderItem orderItem = order.orderItems.find { it.id == line.id }
+                Product product = line.productCode ? Product.findByProductCode(line.productCode) : orderItem.product
+
+                InventoryItem inventoryItem = null
+                if (line.lotNumber && line.expiration) {
+                    inventoryItem = inventoryService.findOrCreateInventoryItem(product, line.lotNumber, line.expiration)
```
DominikaHatala, line 944: If there is no existing inventory item then a new one should be created (It's specified in ticket's description)

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -755,4 +789,199 @@ class OrderService {
...
+        orderItems.each { orderItem ->
+            def row = [
+                    'Order Item ID' : orderItem?.id,
+                    'Pack level 1'  : '',
+                    'Pack level 2'  : '',
+                    Code            : orderItem?.product?.productCode,
+                    Product         : orderItem?.product?.name,
+                    UOM             : orderItem?.unitOfMeasure,
+                    Lot             : '',
+                    Expiry          : '',
```
DominikaHatala, line 804: There should be format added in header (It's specified in ticket's description) so it should be "Expiry (dd/mm/yy)"

grails-app/views/order/shipOrder.gsp
awalkowiak: I shoud default to blank, but only on first enter, after the import that it should take the imported value. I see that this always defaults to shipOrderItem?.quantityToShip (even without any data). Ill fix that.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -935,13 +937,16 @@ class OrderService {
...
         importedLines.each { line ->
-            if (line.quantityToShip > 0) {
+            if ((line.quantityToShip as int) > 0) {
                 OrderItem orderItem = order.orderItems.find { it.id == line.id }
                 Product product = line.productCode ? Product.findByProductCode(line.productCode) : orderItem.product
 
                 InventoryItem inventoryItem = null
-                if (line.lotNumber && line.expiration) {
-                    inventoryItem = inventoryService.findOrCreateInventoryItem(product, line.lotNumber, line.expiration)
+                def expiry = null
```
DominikaHatala, line 945: Why expiry is set to null here and then used in inventoryService.findOrCreateInventoryItem(product, line.lotNumber, expiry)?

awalkowiak: Before doing findOrCreate the value of expiry is assigned. And I declared it here because I reference this later (if the inventory item is created, ofc). This is here just to get rid of an IDE warning.

PR #1835, opened 2020-08-06
OBPIH-3098: Calculating inventory snapshots after each transaction

jmiranda:


grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
```diff
@@ -0,0 +1,89 @@
...
+            def startDate = context.mergedJobDataMap.get('startDate')
+            String locationId = context.mergedJobDataMap.get('location')
+            String transactionId = context.mergedJobDataMap.get('transaction')?.id
+            Transaction transaction = Transaction.get(transactionId)
+            Location location = Location.get(locationId)
+
+            try {
+                log.info ("Refresh inventory snapshot " + retryCount + " out of " + maxRetryAttempts)
+
+                List productIds = transaction?.transactionEntries?.collect { it?.inventoryItem?.product?.id }?.unique()
```
jmiranda, line 60: I was trying to make sure you did not do this because it won't work in the delete case. If I delete a transaction I should not be able to "get" the transaction from the database. That's why it was important to get all product IDs in the event service and spawn separate jobs / threads to update each one separately.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
```diff
@@ -0,0 +1,89 @@
...
+
+                productIds.each { productId ->
+                    Product product = Product.get(productId)
+
+                    // I feel like with this check, the inventory snapshot not always are properly populated (at least during the inventory import).
```
jmiranda, line 65: I assume there's potential for a race condition. But the following line is supposed to prevent the job from being executed by two threads at the same time, so we should be safe.
```
def concurrent = false
```

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
```diff
@@ -0,0 +1,89 @@
...
+                    // No inventory snapshots saves, despite it should be saved (for example two imports with quantities changed right after each other).
+                    if (forceRefresh) {
+                        inventorySnapshotService.deleteInventorySnapshots(location, product)
+                    }
+
+                    // When I call this one (without any date passed) I am getting more transaction dates than I feel like I should.
```
jmiranda, line 71: Yes, this was what you were supposed to override so that it only ran on one date. What you're calling will update the inventory snapshots for all dates given the product and location. At some point we might need this process to be run on multiple dates. For example, when we update a transaction in the past we'd need to update all inventory snapshots from that date up until midnight tonight. But in the case we're trying to solve for we just want to update the inventory snapshot for tomorrow (i.e. new Date()+1).

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
awalkowiak: Oh right, I totally forgot about that one. I'll revisit that.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
awalkowiak: I think that this case occurred when I imported inventory right after the previous one.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
awalkowiak: Yes, I had at the beginning the new Date() + 1, but then I went back to the one without a date.

awalkowiak: I'll readd this.

jmiranda:
@awalkowiak What's the status of this PR?

awalkowiak:
@jmiranda I have to adjust it for the deleting case. Will do it after finishing the work on OBPIH-2936 and OBPIH-2937.

awalkowiak:
@jmiranda I fixed the deleting case. Now there are no LazyInit issues. If you don't like the solution I did, then there will be a need to trigger the event before the delete. Additionally, I added the check if it is deleting and in that case, there is no populating snapshot (only the inventory snapshot delete is called). If this is the wrong assumption, then let me know and I remove this flag.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
```diff
@@ -0,0 +1,87 @@
...
+                def date = new Date() + 1
+
+                productIds.each { String productId ->
+                    Product product = Product.get(productId)
+
+                    if (forceRefresh) {
+                        inventorySnapshotService.deleteInventorySnapshots(location, product)
+                    }
+
+                    if (!isDeleting) {
```
jmiranda, line 69: Can you explain this? I think we still want to refresh the inventory snapshots even when we're deleting a transaction.

grails-app/services/org/pih/warehouse/inventory/TransactionEventService.groovy
```diff
@@ -21,8 +20,20 @@ class TransactionEventService implements ApplicationListener<TransactionEvent> {
         Transaction transaction = event?.source
         def transactionId = transaction?.id
         def transactionDate = transaction?.transactionDate
-        def locationId = transaction?.inventory?.warehouse?.id
-        log.info "Refresh inventory snapshot date=$transactionDate, location=$locationId, transaction=$transactionId"
-        RefreshInventorySnapshotJob.triggerNow([location: locationId, forceRefresh: event.forceRefresh])
+        def locationId = event.associatedLocation
+        List productIds = event.associatedProducts
```
jmiranda, line 24: You don't need to pass this in the event. You can probably just pull this from the transaction. 
```
List productIds = transaction.associatedProducts
```

awalkowiak: I was afraid of getting LazyInitialization exceptions with pulling it from the transaction in the service. I'll check tomorrow morning if it works that way properly and I refactor it if it will be ok.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
awalkowiak: @jmiranda for example:
I had snapshot of the inventory of two products with some quantity. After deleting the transaction for these two entries they were readded to the inventory snapshot table during populateInventorySnapshots (with the same qoh values as it was before deleting). Is this desired behavior? Is there something that indicates if the inventory was deleted even if the quantity on hand has positive value?

awalkowiak: I'll reinvestigate it tomorrow together with the comment related to pulling associatedProducts from transaction.

awalkowiak:
@jmiranda 
1. I checked pulling associated products and location from transaction instead of event and I am getting LazyInitializationException in case of deleting. When passing these as params in event, then it works fine.
2. I rechecked the rightness of isDeleting bool, and the issue i described in my previous comment still occurs. Let me know what do you think about this case:
`I had snapshot of the inventory of two products with some quantity. After deleting the transaction for these two entries they were readded to the inventory snapshot table during populateInventorySnapshots (with the same qoh values as it was before deleting). Is this desired behavior? Is there something that indicates if the inventory was deleted even if the quantity on hand has positive value?`

jmiranda:
> @jmiranda I fixed the deleting case. Now there are no LazyInit issues. If you don't like the solution I did, then there will be a need to trigger the event before the delete. Additionally, I added the check if it is deleting and in that case, there is no populating snapshot (only the inventory snapshot delete is called). If this is the wrong assumption, then let me know and I remove this flag.

If we delete a transaction we definitely need to re-populate the inventory snapshot. Think of the delete as an edit rather than a delete.

grails-app/jobs/org/pih/warehouse/jobs/RefreshInventorySnapshotAfterTransactionJob.groovy
jmiranda: So let's say we have a Product A and that product has two items (A123 and A456). 

State A
Initial inventory is 0.

State B
We Record Stock and quantity for A123 is 10 and A456 is 10. Product A QoH = 20. Inventory Snapshot process runs and saves the two rows.

State C
Then we create a DEBIT transaction with quanitty for A123 = 10. So now we only A456 = 10. Product A = 10. Inventory Snapshot process runs and saves the one row.

State D
Then we delete the DEBIT transaction which reverts the state to the State B. The inventory snapshot process should run and save two records to the table. Product A QoH = 20.

jmiranda: Make sure that works properly and document any other issues (like the race condition) in the ticket.

DominikaHatala (approved):

PR #1841, opened 2020-08-11
OBPIH-3097 Fix inventory import template

DominikaHatala:


grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -934,6 +931,43 @@ class DocumentService {
         }
     }
 
+    void generateInventoryTemplate(OutputStream outputStream, List<Map> data) {
```
jmiranda, line 934: I would prefer if we did something more generic here and expanded the generateExcel() to handle formats and other features.

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -934,6 +931,43 @@ class DocumentService {
...
+            data.eachWithIndex { Map dataRow, rowIndex ->
+                Row excelRow = sheet.createRow(rowIndex + 1)
+                dataRow.keySet().eachWithIndex { columnName, cellIndex ->
+                    def cellValue = dataRow.get(columnName) ?: ""
+                    // POI can't handle objects so we need to convert all objects to strings unless they are numeric
+                    if (!(cellValue instanceof Number)) {
+                        cellValue = cellValue.toString()
+                    }
+                    excelRow.createCell(cellIndex)
+                    if (columnName == 'Expiration date' && cellValue != '') {
```
jmiranda, line 949: For example, instead of hardcoding this here, we could change the row data to include a dataType (date) and format ("MM/dd/yyyy"). For dataType = Number we toString() the cellValue.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -1402,9 +1402,9 @@ class InventoryController {
             def row = [
                     'Product code'    : it.product?.productCode,
                     'Product name'    : it.product?.name,
-                    'Lot number'      : StringEscapeUtils.escapeCsv(it.inventoryItem?.lotNumber ?: ""),
+                    'Lot number'      : it.inventoryItem?.lotNumber,
```
jmiranda, line 1405: So now the row would look something like this:
```
def row = [
    productCode: [label: g.message(code: "product.productCode.label"), value: it?.product?.productCode],
    productName: [label: g.message(code: "product.name.label"), value: it?.product?.productCode],
    expirationDate: [label: g.message(code: "inventoryItem.expirationDate.label"), value: it.inventoryItem?.expirationDate?.format("MM/dd/yyyy"), format: "MM/dd/yyyy"]
]
```
You could also pull the label and format out into a separate data structure so we're not duplicating that data on each row.
```
def columns = [
    productCode: [label: g.message(code:"product.productCode.label"), format: "MM/dd/yyyy"]
]
```

jmiranda:
@awalkowiak We're going to move the refactoring to another ticket. Could you review this PR as-is?

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
awalkowiak: @jmiranda that's the part I was not sure before (one the first sight). It looked very similar to the code that already is there. So I agree that this could be more generic.

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -934,6 +931,43 @@ class DocumentService {
...
+                    }
+                }
+            }
+            for (int i = 0; i <= data.get(0).size(); i++) {
+                sheet.autoSizeColumn(i)
+            }
+            workbook.write(outputStream)
+            outputStream.close()
+        } catch (IOException e) {
+            log.error("IO exception while generating import template")
```
awalkowiak, line 967: I guess this could log the exception too.

PR #1845, opened 2020-08-12
OBPIH-2936,2937: Update product packages after uom edits on placed PO

awalkowiak:


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -696,6 +696,9 @@ class OrderController {
             if (!order.save(flush:true)) {
                 throw new ValidationException("Order is invalid", order.errors)
             }
+            orderService.updateProductPackage(orderItem)
```
jmiranda, line 699: This is going to update product package and unit price every time an order item is saved, not just when the order has already been placed.

jmiranda: We should probably trigger an event to deal with this or just wrap this with an if statement to check if the order has been placed.

awalkowiak: @jmiranda Again I had something done in proper way first and then I decide to remove it. You're right this should be done only in the case after the order is placed. I added it as an if statement because I guess in the event it needs to be checked too.

awalkowiak:
@jmiranda as i did requested change, I am gonna merge it now.

PR #1849, opened 2020-08-13
OBPIH-3091 Remove internal locations API request on pick page

DominikaHatala:


src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -374,23 +373,6 @@ class PickPage extends Component {
       .catch(() => { this.props.hideSpinner(); });
   }
 
-  /**
-   * Fetches available bin locations from API.
-   * @public
-   */
-  fetchBins() {
```
jmiranda, line 381: This method is duplicated three times. Each one has different parameters, but it seems like we should create an OpenboxesApiClient (or several API clients) or even a utility library that contains a bunch of these commonly used API fetches and pushes so we don't have to rewrite the same boilerplate code in multiple places. What's the standard practice for methods like this in React?

awalkowiak: @jmiranda in the case when methods are simillar, then we should do the common util function. However in this case all the fetchBin methods that we have differ mostly in the url search params and state object structure. I am not sure if there is need for doing util function for bins fetching, especially if there are now two left and one of them is using additional object in state. 
On the other hand I do think that we could look through react methods and move to the utils as many stuff as we can.

PR #1854, opened 2020-08-17
OBPIH-3125 Add organizations csv download

DominikaHatala:


grails-app/controllers/org/pih/warehouse/core/OrganizationController.groovy
```diff
@@ -38,6 +39,28 @@ class OrganizationController {
                 }
             }
         }
+        if (params.format && organizationInstanceList) {
```
awalkowiak, line 42: Why this is in search action. I think this could have its own action with and getting organizationInstanceList could be moved to the separate method in the OrganizationService for clarity (which would be used in search and in download csv action.

grails-app/controllers/org/pih/warehouse/core/OrganizationController.groovy
```diff
@@ -12,33 +12,45 @@ package org.pih.warehouse.core
...
 
     def identifierService
+    def organizationService
     def scaffold = true
 
     def search = {
+        def organizationInstanceList = organizationService.getOrganizations(params)
+        render(view: "list", model: [organizationInstanceList:organizationInstanceList, organizationInstanceTotal:organizationInstanceList.totalCount])
+        if (params.format && organizationInstanceList) {
+            redirect(controller: "organization", action: "download", params: params)
```
awalkowiak, line 27: I wanted that to be called separately, not through the search action. Now you have to get organization list two times.

DominikaHatala: @awalkowiak good point, done!

PR #1865, opened 2020-08-26
OBPIH-2947 Create stockout fact table; implement refresh mechanism

jmiranda:


jmiranda:
@awalkowiak If this looks ok, please merge and move to QA so we can try to have it tested on Wednesday.

awalkowiak (approved):

PR #1866, opened 2020-08-26
OBPIH-3113 Send notification for recipients on shipped and received

DominikaHatala:


grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -248,6 +249,19 @@ class ReceiptService {
             }
         }
 
+        // Send notification email on completed receiving
```
jmiranda, line 252: I feel like the save shipment should occur before the email is sent. Can you explain the reason for not doing that?

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -248,6 +249,19 @@ class ReceiptService {
...
+        if (completed) {
+            def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+            def recipientItems = partialReceipt.partialReceiptItems.groupBy {it.recipient }
+            recipientItems.each {recipient, items ->
+                def subject = g.message(code: "email.yourItemReceived.message", args: [shipment.shipmentNumber])
+                def body = "${g.render(template: "/email/shipmentItemReceived", model: [shipmentInstance: shipment, receiptItems: items])}"
+                if (recipient) {
```
jmiranda, line 259: Move this up a few lines so we don't need to render the template unless there's a recipient.

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -1108,6 +1108,16 @@ class ShipmentService {
             // Add a Shipped event to the shipment
             createShipmentEvent(shipmentInstance, shipDate, EventCode.SHIPPED, locationInstance)
 
+            // Send notification email to recipients
+            def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+            def recipientItems = shipmentInstance.shipmentItems.groupBy {it.recipient }
+            recipientItems.each {recipient, items ->
+                def subject = g.message(code: "email.yourItemShipped.message", args: [shipmentInstance.shipmentNumber])
+                def body = "${g.render(template: "/email/shipmentItemShipped", model: [shipmentInstance: shipmentInstance, shipmentItems: items])}"
+                if (recipient) {
```
jmiranda, line 1117: Same here. Move up above subject line.

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -1108,6 +1108,16 @@ class ShipmentService {
...
+                    mailService.sendHtmlMail(subject, body.toString(), recipient.email)
+                }
+            }
```
jmiranda, line 1120: Also check out line 1139. I was hoping (but did not explicitly say) that we'd implement the recipient email sending feature using the same event bus used to send the status change emails. If possible, we should move the email sending code above to its own method in the NotificationService and then invoke it using the ShipmentStatusTransitionEventService. If you have time please make that change. 

Otherwise if @awalkowiak is ok with the PR feel free to merge it so we can get Katarzyna to test before the sprint is closed.

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -248,6 +249,19 @@ class ReceiptService {
             }
         }
 
+        // Send notification email on completed receiving
+        if (completed) {
```
jmiranda, line 253: I don't mind that we're only sending when the whole shipment has been received, but I interpreted Kelsey's requirements as "send whenever any quantity was received". It seems like this would be a quick fix once Kelsey has UAT'd the ticket so let's leave it for now.

DominikaHatala: completed means sending PARTIAL receiving. It can be either fully received or just partially. The same method is used just to save partial receipt.

awalkowiak (approved):

PR #1867, opened 2020-08-27
OBPIH-2602: Filter destination options in the demand table to be rele…

awalkowiak:
…vant to available entries


awalkowiak:
@jmiranda I could not base this list on the query that is in getDemandDetailsForDemandTab(), hence I added a dedicated one.

PR #1870, opened 2020-08-27
OBGM-141: Add context path handling

awalkowiak:


src/js/utils/apiClient.jsx
```diff
@@ -77,14 +77,17 @@ const handleError = (error) => {
   return Promise.reject(error);
 };
 
+// TODO: This is temporary cleaner. Once migration is complete it should be removed
+const cleanUrlFromContextPath = url => url.replace('/openboxes', '');
```
awalkowiak, line 81: I swapped lodash's trimStart with replace because previous version was cutting some unnecessary slashes.

awalkowiak:
@awalkowiak TODO add utility function for adding ${window.CONTEXT_PATH} to some urls.

awalkowiak:
I am closing this one because it is obsolete and it is done in another PR

PR #1873, opened 2020-08-28
OBPIH-2678 Add product color to stock list view and edit pages

DominikaHatala:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -866,7 +866,8 @@ class JsonController {
                         valueText     : localizedName,
                         desc          : product.description,
                         inventoryItems: inventoryItemList,
-                        handlingIcons : product?.getHandlingIcons()
+                        handlingIcons : product?.getHandlingIcons(),
+                        color         : product?.color
```
jmiranda, line 870: This is a super expensive convenience method. Could you see if there's a way to make this more performant?

grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
```diff
@@ -742,7 +742,8 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
                                             name: it.uom?.name
                                     ]
                             ]
-                        }
+                        },
+                        color       : product.color
```
jmiranda, line 746: Same here.

grails-app/views/requisitionTemplate/show.gsp
```diff
@@ -66,11 +66,12 @@
                         <g:set var="requisitionItems" value='${requisition?."$sortByCode.methodName"}'/>
                         <g:each var="requisitionItem" in="${requisitionItems}" status="i">
                             <tr class="prop ${i%2?'even':'odd'}" id="requisitionItem_${requisitionItem?.id }" requisitionItem="${requisitionItem?.id}">
-                                <td>
+                                <td style="color: ${requisitionItem?.product?.color}">
```
jmiranda, line 69: This is going to be slow as well for large requisition templates. Each time you display the color it invokes at least one extra query possibly more.

DominikaHatala: You're right. I'm trying to figure out a way to make this more performant while working on OBPIH-3064. I'm not sure what would be the best option but maybe we should add color field to Product class and update the value if catalog's color is updated. What do you think about it?

jmiranda: That would be really difficult to keep in sync. Grails has a derived property feature that we could potentially use here. So instead of generating lots of queries to get catalog and catalog items, we'd just add a subquery to any query for product. 

See the "More Complicated Example" in the following blog.
https://blog.eyallupu.com/2009/07/hibernate-derived-properties.html

Here's the official Grails doc on Derived Properties
https://grails.github.io/grails2-doc/1.3.9/guide/single.html#5.5.2.11%20Derived%20Properties

I don't know if this will be an improvement but we should investigate.

jmiranda: We can spin this off into a separate ticket once you've used up the time estimate on this. But let's try to get a proof of concept working in this ticket. My guess is that it'll look something like this

Product.groovy
```
static mapping = {
      color formula: "select color from product_catalog join product_catalog_item on product_catalog.id = product_catalog_item.product_catalog_id where product_catalog_item.product_id = id AND <complex logic> ORDER BY <order by clause> LIMIT 1"
}
```

And obviously, remove the getColor() method.

DominikaHatala: @jmiranda Thanks, that's right. I will update this PR once I finish working on edit page.

DominikaHatala: @jmiranda  I'm getting "SELECT command denied to user 'openboxes'@'localhost' for table 'product_catalog_item'" error when trying to add derived property. While executing this query(not exactly the same that is above but a bit changed) I get proper values. Do you know what may be the core?

jmiranda: @DominikaHatala Weird. 

That's another solution I was going to suggest. Just create a query in the getColor() method that returns a single color for the product. Something like the following should work but should be tested more thoroughly. There's no logic for dealing with multiple colors so we'll need to discuss with Kelsey.
```
    def getColor() {
        def results = ProductCatalogItem.executeQuery("select pci.productCatalog.color " +
                "from ProductCatalogItem pci where pci.product = :product " +
                "AND pci.productCatalog.color is not null", [product:this, max:1])
        return results ? results[0] : null
    }
```

DominikaHatala:
@jmiranda This PR is updated and ready to be merged.

jmiranda:
@DominikaHatala Do you know why the build is failing on travis?

DominikaHatala:
@jmiranda It seems that there are two tests failing. Only getColor() method has changed what shouldn't cause any issues but I have to look into that. How can I debug tests?

PR #1882, opened 2020-09-02
OBPIH-3162 Fix downloading packing list

DominikaHatala:


grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -851,6 +851,7 @@ class DocumentService {
                 CELL_INDEX = 0
                 log.debug "Adding item  to packing list " + itemInstance?.product?.name + " -> " + itemInstance?.container?.name
                 row = sheet.createRow((short) counter++)
+                def lastItem = new ArrayList(shipmentInstance.sortShipmentItemsBySortOrder()).last()
```
jmiranda, line 854: Is there ever a case where this code would be invoked with an empty set returned by sortShipmentItemsByOrder()?

jmiranda: A better way to do this (and I should have done this from the beginning) might have been to use ...
```
shipmentInstance.sortShipmentItemsBySortOrder().eachWithIndex { itemInstance, index ->
    ...
    boolean isLastItem = index == shipmentInstance.sortShipmentItemsBySortOrder().size()
    ...
```

jmiranda: One other note ... we're creating a new ArrayList for every loop. This seems like it should be pulled out of the loop. 

But instead of doing that, let's try the simpler approach using the `eachWithIndex`.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
DominikaHatala: @jmiranda Thanks. I've updated PR. Indexes start with 0 so in order to compare index and shipment items to get the last item I had to subtract 1 from shipment items size.

jmiranda (approved):

PR #1886, opened 2020-09-03
OBPIH-3167 Fix getting quantity revised for edit page items

DominikaHatala:


grails-app/migrations/views/stock-movement-item.sql
```diff
@@ -13,7 +13,8 @@ CREATE OR REPLACE VIEW stock_movement_item AS
         requisition_item.quantity_canceled,
         requisition_item.order_index,
         requisition_item.cancel_reason_code,
-        modification_item.quantity as quantity_revised
+        CASE WHEN modification_item.quantity THEN modification_item.quantity
```
jmiranda, line 16: Can you explain what you're trying to do here?

DominikaHatala: When quantity revised is 0 then there is no modification item created, only original item is canceled because it's quantity requested is equal to quantity canceled. Before this change we only got quantity revised if it was not equal to 0.

jmiranda: Why did you need the second CASE? Isn't what you wrote equivalent to the following or am I missing something?
```
CASE 
    WHEN modification_item.quantity THEN modification_item.quantity
    WHEN requisition_item.quantity_canceled = requisition_item.quantity THEN 0 
END AS quantity_revised
```

jmiranda: One thing I might also add is a default case ELSE at the end. I assume the default case is NULL or 0?

DominikaHatala: @jmiranda Yes, you're right, one CASE is enough. I've updated PR.

PR #1890, opened 2020-09-05
OBPIH-3085 Cannot import product sources csv download without error

jmiranda:


DominikaHatala:
@jmiranda Why are there two same PRs?

DominikaHatala (approved):

jmiranda:
@DominikaHatala first one was targeted to the wrong branch so i created a second one for the right branch.

PR #1895, opened 2020-09-07
OBPIH-3111 Fix of the filter not working well

julienreliefapps:


jmiranda:
@awalkowiak @DominikaHatala can you review this one?

awalkowiak (commented):
@jmiranda the react side looked ok to me, but the condition in the query (in IndicatorDataService.groovy) looked a bit weird and I think it has low readability. If you are ok with that part in IndicatorDataService.groovy, then I am ok with it too. Other than that it LGTM

jmiranda:
@awalkowiak That's what I was having trouble with as well. I couldn't figure out a way to do it without writing our own query builder. Can you think of a better way of making it more readable (besides duplicating the queries ... although even that would probably be an improvement)?

awalkowiak:
@jmiranda sure I'll look into it

jmiranda:
@awalkowiak Actually let's just merge it and refactor it later. Look into it a little and add some comments to the ticket.

PR #1898, opened 2020-09-08
OBPIH-3181 Fix loading edit page items

DominikaHatala:


grails-app/migrations/views/edit-page-item.sql
```diff
@@ -27,4 +27,5 @@ CREATE OR REPLACE VIEW edit_page_item AS
             AND product_summary.location_id = stock_movement_item.origin_id
 			LEFT OUTER JOIN
 		product_substitution_status ON product_substitution_status.product_id = stock_movement_item.product_id
-			AND product_substitution_status.location_id = stock_movement_item.origin_id;
+			AND product_substitution_status.location_id = stock_movement_item.origin_id
+			ORDER BY sort_order;
```
awalkowiak, line 31: @DominikaHatala I think that SQL will always want to return data in the most efficient way and will ignore this order during select * from without order by clause.

awalkowiak (changes requested):

grails-app/migrations/views/edit-page-item.sql
awalkowiak: @DominikaHatala Nvm, in our case it should work. But I don't think that this order by would be the reason for the edit page freeze.

awalkowiak:
This was not the reason. I am closing this one.

PR #1905, opened 2020-09-11
OBPIH-3190 Add cancel line functionality for order items

DominikaHatala:


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -968,4 +970,16 @@ class OrderController {
         }
         render (status: 200, text: "Successfully imported template")
     }
+
+    def changeOrderItemStatus = {
```
awalkowiak, line 974: I feel like i'd like this to be splitted. One action for cancellingOrderItem and one action to restoringOrderItem

awalkowiak (changes requested):

grails-app/views/purchaseOrderWorkflow/purchaseOrder/showOrderItems.gsp
```diff
@@ -337,6 +337,51 @@
           return false
         }
 
+        function cancelOrderItem(id) {
```
awalkowiak, line 340: This and restoreOrderItem could be one function, that gets params with id, action, and error message. Other than that lgtm

awalkowiak (changes requested):

awalkowiak (approved):

PR #1908, opened 2020-09-11
OBPIH-3139 Added database migrations for stockout views; triggered in…

jmiranda:
…ventory snapshot process on startup

### DIFF
--- a/grails-app/migrations/views/product-stockout.sql
+++ b/grails-app/migrations/views/product-stockout.sql
@@ -1,0 +1,35 @@
+CREATE OR REPLACE VIEW product_stockout AS
+(
+select stockout_fact_tmp.product_id,
+       stockout_fact_tmp.location_id,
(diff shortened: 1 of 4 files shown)


grails-app/migrations/views/product-stockout.sql
```diff
@@ -0,0 +1,35 @@
+CREATE OR REPLACE VIEW product_stockout AS
+(
```
awalkowiak, line 2: LGTM, but the alignment of the brackets in this file is misleading. I thought that this one was not closed.

awalkowiak (approved):

grails-app/migrations/views/product-stockout.sql
jmiranda: I thought the same thing but this is what IntelliJ does when I run Reformat Code.

PR #1911, opened 2020-09-16
Obpih 3139

julienreliefapps:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/tableroapi/IndicatorDataService.groovy
```diff
@@ -818,10 +818,10 @@ class IndicatorDataService {
         List<Integer> listData = []
 
         def stockOutLastMonth = dataService.executeQuery("""
-            select count(pss.product_id), pss.stockout_status 
-            from product_stockout_status as pss
```
awalkowiak, line 822: @julienreliefapps After this change do we need still this view to be present or this can be removed?

jmiranda: Actually, why are we changing it back? This should use product_stockout_status as defined in https://github.com/openboxes/openboxes/blob/develop/grails-app/migrations/views/product-stockout.sql

julienreliefapps: Ah ok, I didn't understand as in my local, no data were fetched with this new request. But I think it's because my product_summary table is empty. I change it back with this new view.

PR #1923, opened 2020-09-23
OBPIH-3184 Make budget code required for specific locations

DominikaHatala:


grails-app/views/order/_orderItemFormDialog.gsp
```diff
@@ -15,6 +15,9 @@
                     <format:product product="${orderItem.product}"/>
                     <g:hiddenField id="dlgProduct" name="product.id" value="${orderItem?.product?.id}"/>
                     <g:hiddenField id="dlgSupplierId" name="supplier.id" value="${orderItem?.order?.originParty?.id }"></g:hiddenField>
+                    <g:hiddenField id="hasBudgetCodeSupport" name="hasBudgetCodeSupport"
```
awalkowiak, line 18: @DominikaHatala I think, that this should have a different name. Now it is suggesting that the whole feature might be disabled/enabled, but in reality, it is indicating if the field is required. I suggest `isBudgetCodeRequired`.

awalkowiak (changes requested):

awalkowiak (approved):
@DominikaHatala @jmiranda I am merging as it is so Katarzyna can test it. If there will be a need for any further changes related to the `openboxes.budgetCode.enabled` flag, then it can be applied in the new PR.

PR #1925, opened 2020-09-23
Localize all labels / messages in the dashboard UI first translation …

julienreliefapps:
This is a first approach, I still didn't work on how to change the data coming from the database and the tooltip. This one could be in another ticket as said during the call.

In this one, I still have to finish all the labels that I can with the approach presented here.

There is also one bug with a cardTitle that I will fix tomorrow. The goal of this merge request is more for you to be able to have a look at it whenever you get a chance !

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -2740,62 +2740,62 @@ react.stockListManagement.alert.noManagerAssociated.label=There is no manager as
 react.stockListManagement.alert.noManagerSelected.label=Please add a manager as a recipient and resend.
 
 # Dashboard titles
-react.dashboard.numberData.inventoryByLotAndBin=Inventory Details by Lot and Bin
-react.dashboard.numberData.inProgressShipments=Your in Progress Shipments
-react.dashboard.numberData.inProgressPutaways=Your in Progress Putaways
-react.dashboard.numberData.receivingBin=Products in Receiving Bin
-react.dashboard.numberData.itemsInventoried=Items Inventoried this Month
-react.dashboard.numberData.defaultBin=Products in Default Bin
-react.dashboard.numberData.productWithNegativeInventory=Products with Negative Inventory
-react.dashboard.numberData.expiredProductsInStock=Expired products in Stock
+react.dashboard.numberData.inventoryByLotAndBin.label=Inventory Details by Lot and Bin
+react.dashboard.numberData.inProgressShipments.label=Your in Progress Shipments
+react.dashboard.numberData.inProgressPutaways.label=Your in Progress Putaways
+react.dashboard.numberData.receivingBin.label=Products in Receiving Bin
+react.dashboard.numberData.itemsInventoried.label=Items Inventoried this Month
+react.dashboard.numberData.defaultBin.label=Products in Default Bin
+react.dashboard.numberData.productWithNegativeInventory.label=Products with Negative Inventory
+react.dashboard.numberData.expiredProductsInStock.label=Expired products in Stock
 
-react.dashboard.indicatorData.expirationSummaryData=Expiration Summary
-react.dashboard.indicatorData.fillRate=Fill Rate Last 12 Months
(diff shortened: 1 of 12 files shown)


grails-app/i18n/messages.properties
```diff
@@ -2734,5 +2735,61 @@ react.stockListManagement.emailSubject.label=STOCK LIST UPDATE
...
+react.default.dashboard.indicatorData.incomingStock=Incoming Stock Movements By Status
+react.default.dashboard.indicatorData.discrepancy=Items received with a discrepancy
+react.default.dashboard.indicatorData.productsInventoried=Percent of Products Inventoried
+react.default.dashboard.indicatorData.delayedShipments=Delayed Incoming Stock Movements
+react.default.dashboard.indicatorData.lossCausedByExpiry=Value of Stock Lost to Expiry
+react.default.dashboard.indicatorData.percentageAdHoc=Stock vs ad-hoc requests last month
+react.default.dashboard.indicatorData.stockOutLastMonth=Stock out last month
+
+# Dashboard labels lists
+react.default.dashboard.labels.timeline.today=today
```
awalkowiak, line 2765: I think that `default` and `labels` in this case is redundant and the pattern is that `label` is the last part of a property.
So in your case it could be:
react.dashboard.numberData.XYZ.label
react.dashboard.indicatorData.XYZ.label
react.dashboard.timeline.XYZ.label, and so on.
We could also remove the `react` part from dashboard labels, but then you'd need to remember to include it in endpoint returning localization for react pages.

awalkowiak (changes requested):

jmiranda (commented):
I agree with @awalkowiak on the comment below. But other than that it looks good to me.

jmiranda:
@awalkowiak If you feel comfortable with this once @julienreliefapps make the changes, please merge and get it to QA so we can try to get this done before the end of the sprint.

grails-app/i18n/messages.properties
julienreliefapps: Hello ! Yes, this is the part that I didn't understand. I have put react.default because the translation didn't work without. Where should I include this endpoint ? Maybe the problem come from here. @awalkowiak

awalkowiak: @julienreliefapps Quick explaining steps:
1.  As you can see here: https://github.com/openboxes/openboxes/blob/develop/src/js/actions/index.js#L123
there is the prefix that is added after the `react` beginning. 
2. In the MainRouter there is `this.props.fetchTranslations('', 'default');`, that's why it worked for you when you added the `default` to the property key.
3. In order to pull the `react.dashboard.xyz.abc.label`, you'd need to add in the main dashboard component `this.props.fetchTranslations('', 'dashboard');`.

julienreliefapps: Ah ok I see ! I ll do that ! Thank you !

julienreliefapps: Done !

julienreliefapps: Done

grails-app/i18n/messages.properties
```diff
@@ -2737,5 +2738,64 @@ react.stockListManagement.emailSubject.label=STOCK LIST UPDATE
 react.stockListManagement.alert.emailSend.label=Email sent successfully
 react.stockListManagement.alert.noManagerAssociated.label=There is no manager associated with this stock list. Please add a manager and try again.
 react.stockListManagement.alert.noManagerSelected.label=Please add a manager as a recipient and resend.
+
+# Dashboard titles
+react.dashboard.numberData.inventoryByLotAndBin=Inventory Details by Lot and Bin
```
awalkowiak, line 2743: Please add the `.label` at the end of each property key to match the pattern.

awalkowiak (changes requested):

grails-app/i18n/messages.properties
julienreliefapps: ok !

awalkowiak (approved):

PR #1930, opened 2020-09-25
OBPIH-2939 : Localize all labels / messages in the dashboard UI.     Fix

julienreliefapps:
fix of labels on the abscissa, the name on the table, the all destination and implementation of a method to fix the color problem

Also a fix of the preview indicator

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -497,7 +497,13 @@ openboxes {
                             state6 = ["Expiration(s)"]
                         }
                         labels {
-                            state5 = ["today", "within 30 days", "within 90 days", "within 180 days", "within 360 days"]
+                            state5 = [
+                                [code : "react.dashboard.timeline.today.label", message : "today"],
+                                [code : "react.dashboard.timeline.within30Days.label", message : "within 30 days"],
+                                [code : "react.dashboard.timeline.within90Days.label", message : "within 90 days"],
+                                [code : "react.dashboard.timeline.within180Days.label", message : "within 180 days"],
+                                [code : "react.dashboard.timeline.within360Days.label", message : "within 360 days"]
+                            ]
                         }
                     }
                 }
--- a/src/js/consts/dataFormat/graphConfig.jsx
+++ b/src/js/consts/dataFormat/graphConfig.jsx
@@ -67,7 +67,20 @@ function loadGraphColors(payload) {
 
       payload.data.labels.forEach((label, index) => {
         const labelColor = Object.keys(payload.config.colors.labels)
-          .find(key => payload.config.colors.labels[key].includes(label));
+          .find((key) => {
+            let response = false;
+            payload.config.colors.labels[key].forEach((labelConfig) => {
+              if (labelConfig.code && labelConfig.message && label.code && label.message) {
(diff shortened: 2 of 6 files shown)


src/js/consts/dataFormat/graphConfig.jsx
```diff
@@ -67,7 +67,20 @@ function loadGraphColors(payload) {
 
       payload.data.labels.forEach((label, index) => {
         const labelColor = Object.keys(payload.config.colors.labels)
-          .find(key => payload.config.colors.labels[key].includes(label));
+          .find((key) => {
+            let response = false;
+            payload.config.colors.labels[key].forEach((labelConfig) => {
```
awalkowiak, line 72: What's the point of this part? And when the line 80 will need to be removed?

julienreliefapps: This is because we have to get the code of each labelConfig to compare it to the label.code. In fact, now, we don't send a string from the config for the color configuration but also a map

julienreliefapps: @awalkowiak  it will be removed when all the color configurations will sending a map of code and message.

awalkowiak (approved):

grails-app/conf/Config.groovy
```diff
@@ -497,7 +497,13 @@ openboxes {
                             state6 = ["Expiration(s)"]
                         }
                         labels {
-                            state5 = ["today", "within 30 days", "within 90 days", "within 180 days", "within 360 days"]
+                            state5 = [
+                                [code : "react.dashboard.timeline.today.label", message : "today"],
+                                [code : "react.dashboard.timeline.within30Days.label", message : "within 30 days"],
```
jmiranda, line 502: @julienreliefapps @awalkowiak @DominikaHatala At some point we should consider refactoring translation codes like this to a more generic version and pass arguments ...
```
[code : "react.default.withinXDays.label", args: ['30'], message : "within 30 days"],
[code : "react.default.withinXDays.label", args: ['90'], message : "within 90 days"],

```

jmiranda (approved):

grails-app/conf/Config.groovy
julienreliefapps: @jmiranda  yes good idea ! @awalkowiak , @DominikaHatala does the translation passing argument is already working in the react side ?

awalkowiak: @jmiranda Yeah this is the part of these translations I disliked the most, but I somehow forgot about the arguments in translations.

awalkowiak: @jmiranda @julienreliefapps there is an option to add arguments in translations on the react side, but there's a catch:
https://ryandrewjohnson.github.io/react-localize-redux-docs/#dynamic-translations
There might be one tiny issue with it. The placeholder for argument in translation in this react library is `"${argumentName1}${argumentName2}"`, into which arguments are passed as an object with argument names, and for grails it is `"{0}{1}{2}"` with arguments passed in as an array. If we want to do this for react side, then we won't be able to use `react.` properties on the grails side (if we don't need it there, then I think we're good).

julienreliefapps: Ok I see, I am gonna check that @jmiranda @awalkowiak

PR #1934, opened 2020-09-28
OBPIH-3242 Add source name to PO lines export and import

DominikaHatala:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -133,15 +133,15 @@ class ProductSupplierDataService {
...
+            return ProductSupplier.get(params.productSupplier)
         }
 
         Product product = Product.get(params.product.id)
         ProductSupplier productSupplier = new ProductSupplier()
-        productSupplier.code = params.productSupplier.id?:params.supplierCode
+        productSupplier.code = params.productSupplier?:params.supplierCode
         productSupplier.supplierCode = params.supplierCode
-        productSupplier.name = params.supplierCode
+        productSupplier.name = params.sourceName ?: params.supplierCode
```
jmiranda, line 144: I think we should ask kelsey about this, but I think this should probably be one of two options:

Option A
```
productSupplier.name = params.sourceName ?: product?.name
```
Option B
```
productSupplier.name = params.sourceName ?: null
```
I believe Kelsey mentioned that she would prefer Option B, but let's ask her.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
jmiranda: Kelsey says use product.name so let's go with Option A.

DominikaHatala:
@jmiranda I have updated this PR with requested changes

awalkowiak:
Looks like everything is fine now. I am merging this.

PR #1935, opened 2020-09-28
HOTFIX: OBPIH-3255 Add actual ready date to pending inbound tab for POs

DominikaHatala:


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -274,14 +274,16 @@ class InventoryItemController {
                 def quantityPicked =  it instanceof Requisition ? itemsMap[it].sum() { RequisitionItem requisitionItem -> requisitionItem.calculateQuantityPicked() } : 0
                 def quantityRemaining = it instanceof Shipment ? itemsMap[it].sum() { ShipmentItem shipmentItem -> shipmentItem.quantityRemaining } : 0
                 Integer quantityPurchased = it instanceof Order ? itemsMap[it].sum() { OrderItem orderItem -> orderItem.quantityRemaining * orderItem.quantityPerUom } : 0
+                def shipDate = it instanceof Order ? itemsMap[it].first().actualReadyDate : it.expectedShippingDate
```
jmiranda, line 277: This looks a little scary. What are we doing here and would itemsMap[it] ever return null?

DominikaHatala: @jmiranda I think that we should refactor the whole method to improve readability. itemsMap[it] in this case would never return null. There are items of two different classes - Order and Shipment and there are items in itemsMap grouped by order or shipment. Order doesn't have actualReadyDate so we have to take this value from item.

PR #1942, opened 2020-09-30
OBPIH-3256 Rewrite on order summary views

DominikaHatala:


jmiranda (approved):
LGTM. I will merge this as-is so we can have a baseline query. And any changes we make from now will be recorded in git history. Let me know if you get stuck while debugging.

PR #1955, opened 2020-10-07
OBPIH-3250 Adjust menu items visibility for users with different roles

DominikaHatala:


grails-app/conf/Config.groovy
```diff
@@ -917,7 +917,7 @@ openboxes {
         }
         analytics {
             enabled = true
-            requiredRoles = [RoleType.ROLE_ADMIN]
+            requiredRoles = [RoleType.ROLE_SUPERUSER, RoleType.ROLE_ADMIN]
```
jmiranda, line 920: @awalkowiak How does the requiredRoles work? Is it minimum required role? Or any of these roles? Based on your response, I'll let you decide whether this one should be merged or changed?

awalkowiak: @jmiranda there is used hasAnyRoles, so this usage is ok. But I am wondering if the minimum required role would not be better (cleaner)

awalkowiak: As it goes to hotfix then I am merging as it is

awalkowiak (approved):

PR #1957, opened 2020-10-07
OBPIH-3260 Stock movement design header and page fit changes

bamoure:
Removed the stock movement info box on all pages except the "Send" page


awalkowiak (approved):
@bamoure Should all those fields and lines be removed? That was requested? Nothing instead? If yes then I am approving this.
cc @jmiranda

PR #1966, opened 2020-10-12
OBPIH-3282 Refactor custom links in megamenu to use the same configur…

DominikaHatala:
…ation as the react menu

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1120,13 +1120,11 @@ openboxes {
             ]
         }
         customLinks {
-            enabled = false
+            enabled = true
             label = "customLinks.label"
             defaultLabel = "Custom Links"
             menuItems = [
-                [label: "requestItemCreation.label", defaultLabel: "Request Item Creation", href: "", target: "_blank"], // Fill in href
-                [label: "trainingVideos.label", defaultLabel: "Training Videos", href: "", target: "_blank"], // Fill in href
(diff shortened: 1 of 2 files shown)


grails-app/conf/Config.groovy
```diff
@@ -1124,9 +1124,8 @@ openboxes {
             label = "customLinks.label"
             defaultLabel = "Custom Links"
             menuItems = [
-                [label: "requestItemCreation.label", defaultLabel: "Request Item Creation", href: "", target: "_blank"], // Fill in href
-                [label: "trainingVideos.label", defaultLabel: "Training Videos", href: "", target: "_blank"], // Fill in href
-                [label: "tableroNuevo.label", defaultLabel: "Tablero Nuevo", href: "/${appName}/tablero"],
+                    [label: "requestItemCreation.label", defaultLabel: "Request Item Creation", href: "https://goo.gl/forms/x1red4Q7fL0ThH253", target: "_blank"],
```
awalkowiak, line 1127: @DominikaHatala I think this should not be here.

awalkowiak (changes requested):

grails-app/conf/Config.groovy
DominikaHatala: @awalkowiak why not?

awalkowiak: I meant the value of href, because we don't want to show such links in a public repo.

DominikaHatala: @awalkowiak Yeah sure, I've just updated PR.

jmiranda (commented):
Approved

PR #1971, opened 2020-10-13
OBPIH-3251 Investigate dashboard config issues

julienreliefapps:
Basically in that one I just control if a new graph has been added or if one has been removed checking if the size of the user config is still the same as the config in the config.groovy file.

If the size is the same we compare every keys in order to see if anything has changed.

We don't have to remove the user config because as soon as he will save a new config, his config will be updated taking as well the eventual graph added or removed.

In the end, with this fix, I don't think that we need a button to remove the user dashboardConfig. 

What do you think ? 

@awalkowiak @jmiranda

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -350,10 +350,14 @@ class UserService {
     def getDashboardConfig(User user) {
         def config = grailsApplication.config.openboxes.tablero
         def userConfig = user.deserializeDashboardConfig()
-
+        
         if (userConfig != null) {
-            updateConfig("graph", config, userConfig)
-            updateConfig("number", config, userConfig)
+            int userConfigSize = userConfig.graph.size() + userConfig.number.size()
+            int configSize = config.endpoints.number.size() + config.endpoints.graph.size()
+            if (userConfigSize == configSize) {
+                updateConfig("graph", config, userConfig)
+                updateConfig("number", config, userConfig)
+            }
         }
 
         return config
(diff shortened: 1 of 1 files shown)


julienreliefapps:
In the end I may have found a way to fix this problem as well

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -350,10 +350,33 @@ class UserService {
     def getDashboardConfig(User user) {
         def config = grailsApplication.config.openboxes.tablero
         def userConfig = user.deserializeDashboardConfig()
-
+        Boolean configChanged = false
```
awalkowiak, line 353: @julienreliefapps Since config has a JSON format, would not it be easier to include in it, for each "section" a flag that indicates if it has changed in comparison to the one from Config.groovy? For example in Config.groovy by default:
`config.endpoints.graph.default = true` (or `changed = false`)
`config.endpoints.number.default = true` (or `changed = false`)
and each time if the user swaps something you just set this to the opposite and if the user restores the default value it will be set back to the former version. Thanks to that you don't have to do size counts or checking for key changes, only the check: `if not userConfig.number.default then do update`
cc @jmiranda

julienreliefapps: Ok I think I understand, I ll try that @awalkowiak

julienreliefapps: If I set the default to false everytime that the user update a new configuration and I just check this before resetting or not the configuration of the user, he will never keep it, everytime that he will refresh the page the configuration will be reset.

The problem is, yes when the user has a dashboard configuration saved, but only when the config.groovy has changed, so I don't understand how we can check if they are different just checking the user configuration

But maybe I have misunderstood what you proposed

awalkowiak: @julienreliefapps You are right this might not be the proper solution to the existing issue. But still, we need to find a better way of doing this.

julienreliefapps: @awalkowiak  It's true, I ll try to find one

jmiranda (approved):

PR #1972, opened 2020-10-13
OBPIH-3176: Refactor Purchase Order workflow

awalkowiak:


jmiranda:
@awalkowiak Will review this tomorrow.

DominikaHatala (approved):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -364,7 +364,7 @@ class OrderController {
                 orderAdjustment.properties = params
                 if (!orderAdjustment.hasErrors() && orderAdjustment.save(flush: true)) {
                     flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'orderAdjustment.label', default: 'Order Adjustment'), orderAdjustment.id])}"
-                    redirect(controller:"purchaseOrderWorkflow", action: "purchaseOrder", id: orderInstance.id, params:['skipTo': 'adjustments'])
+                    redirect(controller:"purchaseOrder", action: "addItems", id: orderInstance.id, params:['skipTo': 'adjustments'])
```
jmiranda, line 367: So I assume we don't need `skipTo` any longer unless you implemented that to select the appropriate tab. In either case we can replace the skipTo parameter with the fragment attribute as that should work to select the right tab (at least I think). I think the selected tab is stored in the cookie anyway so it probably doesn't matter. 
```
fragment="add-adjustment"
```

jmiranda (approved):
LGTM. Take a look at the comment below. But I'll let you decide whether you want to make that change.

jmiranda:
Nevermind, I'm just going to merge it.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
awalkowiak: @jmiranda I just find this comment. 'skipTo' is also used to open the proper tab on the page here:
https://github.com/openboxes/openboxes/blob/develop/grails-app/views/purchaseOrder/_showOrderItems.gsp#L803

PR #1982, opened 2020-10-20
OBPIH-3206; OBPIH-3263 Add items page with split line functionality t…

DominikaHatala:
…o combined shipments worfklow plus add items modal


src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -0,0 +1,874 @@
...
+   * updated (comparing to previous state of line items).
+   * @param {object} lineItems
+   * @public
+   */
+  getLineItemsToBeSaved(lineItems) {
+    const items = AddItemsPage.updateSortOrder(lineItems);
+    const lineItemsToBeAdded = _.filter(items, item => !item.id);
+    const lineItemsToBeUpdated = _.filter(items, item => item.id);
+
+    return [].concat(
```
awalkowiak, line 279: Is there a need to do the two separate maps and concat them? Looks like the only difference here is the id field. I think this could be sent as `null` if there is no item.id (`id: item.id ? item.id : null`).

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -168,6 +169,14 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         return "Product"
     }
 
+    Integer getQuantityRemainingToShip(Shipment shipment) {
```
awalkowiak, line 172: I think this could be added to transients.

awalkowiak (changes requested):

awalkowiak:
@DominikaHatala The failing build is due to the missing modal in this pr. Once change requests are resolved I'll push my work into this branch.

awalkowiak (approved):

awalkowiak:
@jmiranda please review, otherwise, I'm gonna merge it tomorrow morning.

jmiranda:
@awalkowiak will do

grails-app/conf/UrlMappings.groovy
```diff
@@ -213,6 +213,23 @@ class UrlMappings {
             action = [DELETE: "removingItem"]
         }
 
+        // Combined shipments
+
+        "/api/orderNumberOptions"(parseRequest: true) {
```
jmiranda, line 218: I'm going to review / refactor the APIs a little later. We should discuss these in an upcoming tech huddle.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -0,0 +1,97 @@
...
+import org.pih.warehouse.product.Product
+import org.pih.warehouse.shipping.Shipment
+import org.pih.warehouse.shipping.ShipmentItem
+
+class CombinedShipmentItemApiController {
+
+    def getOrderOptions = {
+        def vendor = Location.get(params.vendor)
+        def destination = Location.get(params.destination)
+        List<Order> orders = Order.findAllByOriginAndDestination(vendor, destination)
```
jmiranda, line 26: This probably needs to be of PurchaseOrder type.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    def findOrderItems = {
+        List<Order> orders
+        if (params.orderIds) {
+            orders = Order.findAllByIdInList(params.orderIds)
+        } else {
+            def vendor = Location.get(params.vendor)
+            def destination = Location.get(params.destination)
+            orders = Order.findAllByOriginAndDestination(vendor, destination)
```
jmiranda, line 42: This probably needs to of PurchaseOrder type too. Just add a method to the service.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -0,0 +1,97 @@
...
+            itemsToAdd.each {
+                OrderItem orderItem = OrderItem.get(it.orderItemId)
+                ShipmentItem shipmentItem = new ShipmentItem()
+                shipmentItem.product = orderItem.product
+                shipmentItem.inventoryItem = orderItem.inventoryItem
+                shipmentItem.product = orderItem.product
+                shipmentItem.quantity = orderItem.quantity
+                shipmentItem.recipient = orderItem.recipient
+                shipmentItem.quantity = it.quantityToShip * orderItem.quantityPerUom
+                shipmentItem.save()
```
jmiranda, line 88: You don't need these two saves. The shipment.save() should cascade.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -500,35 +500,60 @@ class StockMovementService {
...
+        if (requisitionItem) {
+            removeRequisitionItem(requisitionItem)
+        } else {
+            removeShipmentItem(shipmentItem)
+        }
     }
 
     def getStockMovementItems(String id, String stepNumber, String max, String offset) {
         // FIXME should get stock movement instead of requisition
+        Shipment shipment = Shipment.get(id)
```
jmiranda, line 513: I don't know how I feel about this here. Let's move this into the else (line 542) since it's not used outside of that scope and it won't be tempting to use it outside of the scope.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -500,35 +500,60 @@ class StockMovementService {
...
+            requisitionItems.each { requisitionItem ->
+                StockMovementItem stockMovementItem = StockMovementItem.createFromRequisitionItem(requisitionItem)
+                stockMovementItems.add(stockMovementItem)
             }
         } else {
-            requisitionItems = RequisitionItem.createCriteria().list() {
-                eq("requisition", requisition)
-                isNull("parentRequisitionItem")
-                order("orderIndex", 'asc')
+            if (max != null && offset != null) {
```
jmiranda, line 542: Why isn't this just 
```
if (max && offset) { 
```

jmiranda: If there's a reason (offset can be 0 and therefore we need to be explicit) then I'm ok with it but I really don't like null comparisons after developing with Groovy for 10 years.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -500,35 +500,60 @@ class StockMovementService {
...
+        }
     }
 
     def getStockMovementItems(String id, String stepNumber, String max, String offset) {
         // FIXME should get stock movement instead of requisition
+        Shipment shipment = Shipment.get(id)
         Requisition requisition = Requisition.get(id)
         List<StockMovementItem> stockMovementItems = []
         List <RequisitionItem> requisitionItems = []
+        List <ShipmentItem> shipmentItems = []
```
jmiranda, line 517: Do we need this outside the else scope? It feels like we're converting these to stockMovementItems before we return from that branch. And if that's the case for the requisitionItems, let's move that declaration into the if as well.

jmiranda:
Feel free to merge if you feel comfortable.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
DominikaHatala: Just as you said, offset can be 0 and that's why we're checking if it's not null.

PR #1991, opened 2020-10-26
OBPIH-3205 Add links to create combine shipments

DominikaHatala:


grails-app/controllers/org/pih/warehouse/api/OrderApiController.groovy
```diff
@@ -0,0 +1,20 @@
...
+ * By using this software in any fashion, you are agreeing to be bound by
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.api
+
+import grails.converters.JSON
+import org.pih.warehouse.order.Order
+
+class OrderApiController {
```
awalkowiak, line 15: How about matching it with CombinedShipmentItemApiController name? (either rename this or the other one to match).

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/OrderApiController.groovy
DominikaHatala: @awalkowiak Done

awalkowiak (approved):

PR #1994, opened 2020-10-28
OBPIH-3199 Add items import to combined shipments workflow

DominikaHatala:


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -981,44 +981,6 @@ class OrderController {
         }
     }
 
-
-    def importTemplate = {
```
awalkowiak, line 985: This is still used on the Ship Order page. If there is no "transition" to the Ship Order page (or it is not used anymore), then its view should be removed too. I think we should try to get rid of the dead code.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
jmiranda: Not yet. We want the Ship Order page to be available while we work on this feature. We'll remove it once we're convince it's working well for the 1:1 use case.

jmiranda: With that said, it would be great to add a TODO for dead code that should be removed so we can do some clean up every so often.

DominikaHatala: @jmiranda @awalkowiak I have added TODOs for dead code.

awalkowiak (approved):

awalkowiak (approved):

PR #1999, opened 2020-10-29
OBPIH-3240 Enable edits of some fields in shipped PO lines

DominikaHatala:


grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -150,6 +152,12 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         }?:0
     }
 
+    Integer getQuantityInShipmentsInStandardUom() {
```
awalkowiak, line 155: This is already added in the combined shipments feature

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -164,6 +172,10 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         return quantityReceivedInStandardUom / quantityPerUom
     }
 
+    Integer getQuantityInShipments() {
```
awalkowiak, line 175: Same as above

awalkowiak (approved):
LGTM, but there is some code that is already in the combined shipments feature. When we have plans on merging the combined shipments?

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
DominikaHatala: But this feature branch is not yet merged to develop.

jmiranda (approved):

PR #2001, opened 2020-10-29
Obpih-3244, 3245:  Add GL Account Type and GL Account

awalkowiak:


grails-app/conf/Config.groovy
```diff
@@ -1197,6 +1197,8 @@ openboxes {
                         [label: "documents.label", defaultLabel: "Documents", href: "/${appName}/document/list"],
                         [label: "documentTypes.label", defaultLabel: "Document Types", href: "/${appName}/documentType/list"],
                         [label: "eventTypes.label", defaultLabel: "Event Types", href: "/${appName}/eventType/list"],
+                        [label: "glAccountType.label", defaultLabel: "GL Account Type", href: "/${appName}/glAccountType/list", requiredRoles: [RoleType.ROLE_ADMIN, RoleType.ROLE_SUPERUSER]],
```
jmiranda, line 1200: This isn't a big deal, but can you figure out a way to implement the requiredRoles using the lowest allowed role. I know this is implemented somewhere using getEffectiveRoles() or getHighestRole(). We basically just want to include the lowest common denominator here. It's possible that the taglib that is used to check if a user has a role in the megamenu does this same logic. And if that's the case just leave it as-is and we'll refactor later.

grails-app/controllers/org/pih/warehouse/core/GlAccountController.groovy
```diff
@@ -0,0 +1,81 @@
...
+* which can be found in the file epl-v10.html at the root of this distribution.
+* By using this software in any fashion, you are agreeing to be bound by
+* the terms of this license.
+* You must not remove this notice, or any other, from this software.
+**/
+package org.pih.warehouse.core
+
+class GlAccountController {
+
+    def scaffold = GlAccount
```
jmiranda, line 14: For what it's worth, the scaffold property allows you to exclude the actions below, unless you're overriding them.

jmiranda (approved):
Looks good to me. If @DominikaHatala approves, feel free to merge.

DominikaHatala (approved):

grails-app/controllers/org/pih/warehouse/core/GlAccountController.groovy
awalkowiak: I'll remove it in next pr

grails-app/conf/Config.groovy
awalkowiak: I'll create a ticket for this

PR #2009, opened 2020-11-02
OBPIH-3246, 3275: GL Account validation

awalkowiak:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -585,6 +585,9 @@ class OrderService {
                         if (!product) {
                             throw new ProductException("Unable to locate product with product code ${productCode}")
                         }
+                        if (order.destination.isAccountingRequired() && !product.glAccount) {
```
jmiranda, line 588: We can leave this for now but this feels like it could be added as a custom validation on order item.

jmiranda (approved):

PR #2022, opened 2020-11-05
OBPIH-3306 Scroll to bottom of items table when add new line

DominikaHatala:


src/js/components/form-elements/FieldArrayComponent.jsx
```diff
@@ -54,6 +54,8 @@ class FieldArrayComponent extends Component {
     const { maxTableHeight = 'calc(100vh - 400px)', virtualized } = fieldsConfig;
     const addRow = (row = {}, index = null) => {
       if (index === null) {
+        const table = document.querySelectorAll('[role="rowgroup"]')[0];
```
awalkowiak, line 57: Why does it work without this part on the inbound sm?

DominikaHatala: @awalkowiak It works on inbound, outbound and for requests because FieldArrayComponent is used on each of those workflows.

awalkowiak (approved):

PR #2028, opened 2020-11-09
OBPIH-3355; OBPIH-3258: Add cancel adjustment functionality and refre…

DominikaHatala:
…sh adjustments tab when delete item

### DIFF
--- a/grails-app/views/purchaseOrder/_showOrderItems.gsp
+++ b/grails-app/views/purchaseOrder/_showOrderItems.gsp
@@ -124,7 +124,7 @@
                     <g:form controller="order" action="saveAdjustment" onsubmit="return validateAdjustmentsForm();">
                         <g:hiddenField name="order.id" value="${order?.id}" />
                         <input type="hidden" name="canManageAdjustments" id="canManageAdjustments" value="${canManageAdjustments}">
-                        <table class="items-table">
+                        <table id="orderAdjustmentsTable" class="items-table">
                             <thead>
                             <tr class="odd">
                                 <th><warehouse:message code="default.type.label"/></th>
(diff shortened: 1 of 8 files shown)


DominikaHatala:
@awalkowiak @jmiranda This one requries database migration - new boolean field "isCanceled" was added to order adjustment.

grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -394,6 +392,17 @@
           });
         }
 
+        function getTotalAdjustments() {
```
jmiranda, line 395: This doesn't feel very DRY :) 

We could probably add a utility library for methods like this. I have the same or very similar code in a few places in _productDetails.gsp. Perhaps we could make one single function in a utils.js library that does the following

```
function refreshDataPoint(url, data, target, digits) { 
    $.ajax({
        url:url,
        data: data,
        success: function(data, textStatus){
              target.html(parseFloat(data).toFixed(digits));
        }
    });
}
```

And then on pages like this we'd just use the library method.
```
function getTotalAdjustments() { 
   var target = $("#totalAdjustments");
   var data = { orderId: $("#orderId").val() }
   var url = "${g.createLink( controller:'order', action:'getTotalAdjustments')}"
   refreshDataPoint(url, data, target, 2)
}
```

In this case, I understand this doesn't really save us all that much, but it'll be much easier to troubleshoot and extend if it's all in one place.

jmiranda (changes requested):
LGTM. just take a look at the comment about refactoring.

grails-app/views/purchaseOrder/_showOrderItems.gsp
DominikaHatala: @jmiranda I think it's a good idea. I'll create a new ticket for that. Does that sounds good?

DominikaHatala:
@jmiranda @awalkowiak  I have created new ticket (https://pihemr.atlassian.net/browse/OBPIH-3386) for adding utils library so I think we can merge this one.

awalkowiak (approved):

awalkowiak:
LGTM I am merging it because I need the code from this pr

PR #2031, opened 2020-11-10
OBPIH-3274: Add GL validation for order adjustment on PO

awalkowiak:


grails-app/views/order/editAdjustment.gsp
```diff
@@ -122,6 +122,30 @@
...
+
+	function saveOrderAdjustment() {
+		var data = $("#orderAdjustmentForm").serialize();
+		if (validateForm()) {
+			$.ajax({
+				url:'${g.createLink(controller:'order', action:'saveAdjustment')}',
+				data: data,
+				success: function() {
+					$.notify("Successfully saved new adjustment", "success");
+					window.location = '${g.createLink(controller: 'purchaseOrder', action: 'addItems', params: [id: orderInstance?.id, skipTo: 'adjustments'])}';
```
awalkowiak, line 134: I don't like it, but as I wanted to keep the GL Account notifications consistent I had to handle it via ajax request, and ajax does not properly handle redirects that occur in action it calls.

jmiranda: Yeah probably because AJAX requests are separate requests. I'm fine with this. I'm sure we've done the same in other places. Perhaps  a better approach would be to set the redirect URL as an attribute of whatever element is causing the saveOrderAdjustment() function to be called. Or as a hidden input value. That way we at least don't need to mix JS and GSP.

jmiranda (approved):

DominikaHatala (approved):

PR #2038, opened 2020-11-13
OBPIH-3032; OBPIH-3030 Add html and csv request detail report

DominikaHatala:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1771,9 +1774,48 @@ class JsonController {
     }
 
     def getRequestDetailReport = {
-        // TODO in another ticket
-        log.info params
-        render([aaData: []] as JSON)
+        DateFormat dateFormat = new SimpleDateFormat("MM/dd/yyyy")
+        params.startDate = dateFormat.parse(params.startDate)
+        params.endDate = dateFormat.parse(params.endDate)
```
jmiranda, line 1779: What about when these are empty?

grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1771,9 +1774,48 @@ class JsonController {
...
+                "Request Number" { it.requestNumber }
+                "Date Requested" { it.dateRequested }
+                "Date Issued" { it.dateIssued }
+                "Origin" { it.origin }
+                "Destination" { it.destination }
+                "Product Code" { it.productCode }
+                "Product Name" { it.productName }
+                "Quantity Requested" { it.qtyRequested }
+                "Quantity Issued" { it.qtyIssued }
+                "Reason Code" { it.reasonCode }
```
jmiranda, line 1794: I wanted all the quantity columns together. Otherwise it's hard to read.

src/groovy/org/pih/warehouse/core/ReasonCode.groovy
```diff
@@ -64,6 +64,8 @@ enum ReasonCode {
     REJECTED(32),
     REQUEST_ERROR(33),
     NOT_NEEDED_BASED_ON_CONSUMPTION(34),
+    CLINICAL_JUDGMENT(35),
+    INSUFFICIENT_QUANTITY_AVAILABLE(36),
```
jmiranda, line 68: These aren't actually reason codes. They are classifications for all of the reason codes. This will be fine for now, but we need to convert these to be in a separate enum or perhaps related as parents to each of the ReasonCode enum values.

jmiranda (approved):
I'll merge since we're on a tight deadline. But please take a look at these comments as we'll need to address them.

grails-app/views/report/showRequestDetailReport.gsp
```diff
@@ -207,10 +188,10 @@
...
             { "mData": "productCode" },
             { "mData": "productName" },
-            { "mData": "qtyRequested"},
-            { "mData": "qtyIssued"},
-            { "mData": "reasonCodes"},
-            { "mData": "qtyDemand"},
+            { "mData": "quantityRequested"},
+            { "mData": "quantityIssued"},
+            { "mData": "reasonCode"},
+            { "mData": "quantityDemand"},
```
awalkowiak, line 194: Here's another example of quantities not being together

awalkowiak (approved):
Other than what @jmiranda noticed it looks good to me (I found only one more case where quantity demand could be swapped with reason code).

PR #2042, opened 2020-11-18
OBPIH-3032 Minor changes to request detail report

DominikaHatala:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1777,48 +1777,50 @@ class JsonController {
...
-                        destination: StringEscapeUtils.escapeCsv(it.destination),
-                        productCode  : it.productCode,
-                        productName  : StringEscapeUtils.escapeCsv(it.productName),
-                        qtyRequested     : it.quantityRequested,
-                        qtyIssued     : it.quantityIssued,
-                        reasonCode     : it.reasonCode,
-                        qtyDemand     : it.quantityDemand,
-                ]
-            }
+        if (params.startDate && params.endDate) {
```
awalkowiak, line 1780: What will happen if there will be no startDate and endDate? You should throw some error in that case (I know that there is validation on the front end, but someone can use this method in another place without proper validation).

awalkowiak (changes requested):

awalkowiak (approved):

PR #2054, opened 2020-11-25
Resize dashboard widgets

julienreliefapps:


src/js/components/tablero/Tablero.jsx
```diff
@@ -130,7 +130,8 @@ class Tablero extends Component {
     this.state = {
       isDragging: false,
       showPopout: false,
-      showNav: false,
+      // Not showing if screen too small
+      showNav: window.innerWidth > 1115,
```
awalkowiak, line 134: I am not sure if I understand this part. Is it only checked on default page load? What with browser window resizing? Also where does `1115` come from? It is not any of the standard screen resolution widths (https://en.wikipedia.org/wiki/Display_resolution#/media/File:Vector_Video_Standards8.svg, https://en.wikipedia.org/wiki/Comparison_of_high-definition_smartphone_displays)

awalkowiak (changes requested):
Please include the ticket number in the commit names as it is done in the project (Jira ticket number or github issue hash)

src/js/components/tablero/Tablero.jsx
julienreliefapps: @awalkowiak  Yes, @jmiranda wanted to pin the left nav but I realized that for very small screen, we weren't gonna see well the dashboard, so I decided to pin it only for big screen (1115 is the size in which we start having the overlay) and only at the beginning, so people can close it if they want even more space. What do you think ?

awalkowiak: Ok, but please keep the commit name pattern

julienreliefapps: @awalkowiak yes my bad I forgot this time, do you want me to rename this commit ? Or you were saying that for next time ?

awalkowiak: I'll add it to the merge commit, but please remember for the future :)

julienreliefapps: thank you very much !

PR #2060, opened 2020-11-30
OBPIH-3371 Fixes to add items page on request workflow

DominikaHatala:


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -370,7 +370,7 @@ class AddItemsPage extends Component {
       ) {
         lineItemsToBeUpdated.push(item);
       } else if (newQty !== oldQty || !item.quantityRequested ||
-        oldItem.comments !== item.comments) {
+        (oldItem.comments !== item.comments && _.isNil(item.comment))) {
```
awalkowiak, line 373: Instead of `_.isNil(item.comment)` I think it should be `_.isNil(item.comments)`. Are you sure about the `&&` here?

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
DominikaHatala: @awalkowiak Yes, thanks. It should !_.isNil(item.comments), I've changed it.

awalkowiak (approved):

PR #2064, opened 2020-12-01
OBPIH-3215 Receiving design update

bamoure:
@awalkowiak I have an issue on this ticket regarding what you mentioned on Slack yesterday: when the disabled condition is set in `getDynamicAttr` the field does not get disabled even if the condition is true. Do you know why this is the case?

For example on the PartialReceivingPage, the checkboxes get disabled as well as the top Bin Location select (where the disabled is set directly), but not the other fields.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -682,6 +682,14 @@ breadcrumbsConfig {
             actionUrl = "/${appName}/stockMovement/createRequest/"
             listUrl = "/${appName}/stockMovement/list?direction=INBOUND"
         }
+        verifyRequest {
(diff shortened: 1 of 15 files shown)


awalkowiak:
@bamoure this is because my changes fixing it were not yet merged.

bamoure:
@awalkowiak So will it work automatically once your fix is merge? Or is there anything I need to change here?

awalkowiak:
@bamoure I merged mine, you can rebase this pr and check if everything works properly.

awalkowiak:
@bamoure in this case it is better to do the rebase locally and force push to the branch, now I have to go through my code during the review (note for future).

awalkowiak (approved):

PR #2065, opened 2020-12-01
OBPIH-3393 Putaway redesign

bamoure:


src/js/components/put-away/PutAwayCheckPage.jsx
```diff
@@ -50,20 +50,18 @@ class PutAwayCheckPage extends Component {
 
   constructor(props) {
     super(props);
-    const {
-      putAway, pivotBy, expanded, location,
```
awalkowiak, line 54: Why this was removed? You could do: 
`const { initialValues: { putAway, pivotBy, expanded, location } } = this.props;`

src/js/components/put-away/PutAwayCheckPage.jsx
```diff
@@ -216,13 +215,18 @@ class PutAwayCheckPage extends Component {
     }
   }
 
+  firstPage() {
```
awalkowiak, line 218: If you are using Wizard component, then you should use `goToPage` for handling this.

src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -29,20 +29,19 @@ const SelectTreeTable = (customTreeTableHOC(ReactTable));
 class PutAwaySecondPage extends Component {
   constructor(props) {
     super(props);
-    const {
-      putAway, pivotBy, expanded, location,
```
awalkowiak, line 33: `const { initialValues: { putAway, pivotBy, expanded, location } } = this.props;`

awalkowiak (changes requested):

awalkowiak (approved):

PR #2066, opened 2020-12-01
OBS-461 Set transactionNumber anytime new transaction is created

DominikaHatala:


awalkowiak (commented):
@DominikaHatala how about a few transactions that are created in `IventoryController` (in `saveInventoryChanges` and `saveTransaction`) and in `InventoryItemController` (in `saveInventoryItem` and `saveTransactionEntry`)? Do we need it there too?

DominikaHatala:
@awalkowiak **saveInventoryChanges**, **saveInventoryItem** and **saveTransactionEntry** are not used anywhere(at least I couldn't find any usage either in code or while testing all places with inventory transactions i napplications) and **saveTransaction** is using existing transaction(with transactionNumber already set) so I don't think

DominikaHatala:
@jmiranda Can I help you somehow with any work in this one?

jmiranda:
@DominikaHatala This one'll just go to develop i.e. the next release. If we end up needing it sooner we can create a hotfix.

awalkowiak:
@jmiranda seems like this one is ready to be merged. I'll merge it on Monday when Katarzyna will be back in the office.

PR #2080, opened 2020-12-07
OBPIH-3392 Fix tabs display on sm details page

DominikaHatala:


awalkowiak (approved):
I don't like the length of the line, but ok.

PR #2088, opened 2020-12-09
OBPIH-2806 Add incoming stock requests card to dashboard

DominikaHatala:


grails-app/services/org/pih/warehouse/tableroapi/NumberDataService.groovy
```diff
@@ -249,4 +251,34 @@ class NumberDataService {
...
+                ])
+
+        def title = [
+                code : "react.dashboard.numberData.openStockRequests.label",
+                message : messageService.getMessage("react.dashboard.numberData.openStockRequests.label")
+        ]
+
+        def subTitle = [
+                code : "react.dashboard.subtitle.requests.label",
+                message : messageService.getMessage("react.dashboard.subtitle.requests.label")
```
jmiranda, line 275: See discussion in this ticket (https://pihemr.atlassian.net/browse/OBPIH-3416) for deciding on a consistent localization key for these messages. Please comment in the ticket and let's discuss during tech huddle on Friday. For this one, let's have a quick discussion here to try to pin down generally what we want. I feel like subtitle before requests doesn't make sense. Also, requests by itself is too general, so let's give it a more specific id like openRequests or openStockRequests.
```
react.dashboard.requests.subtitle.label
```

jmiranda (commented):
LGTM. Let's just have a quick discussion about localization keys below.

grails-app/services/org/pih/warehouse/tableroapi/NumberDataService.groovy
DominikaHatala: @jmiranda I agree that removing subtitle is a good idea but in this case requests is not too general because this label literally says "Requests". We also have label for openStockRequests.

DominikaHatala: @jmiranda So I will change this label to
`react.dashboard.requests.subtitle.label`
and `react.dashboard.numberData.openStockRequests.label`
to` react.dashboard.openStockRequests.title.label`.
Is that fine?

awalkowiak: @DominikaHatala Yes. Let's follow the ￼`react.dashboard.<indicatorId>.<otherElementType>.labe￼l` pattern.

DominikaHatala: @jmiranda @awalkowiak Done.

awalkowiak (approved):

PR #2094, opened 2020-12-10
OBPIH-3421 Fix months difference calculation on request report

DominikaHatala:


grails-app/views/report/showRequestDetailReport.gsp
```diff
@@ -166,9 +166,9 @@
           }
           nRow.getElementsByTagName('th')[2].innerHTML = totalDemand;
           var secondRow = $(nRow).next()[0];
-          var startDate = moment($('#startDate').val());
-          var endDate = moment($('#endDate').val());
-          var monthsDifference = endDate.diff(startDate, 'months');
+          var startDate = moment($('#startDate').val()).format("MM");
+          var endDate = moment($('#endDate').val()).format("MM");
+          var monthsDifference = moment(endDate).diff(moment(startDate), 'months', true) + 1;
```
awalkowiak, line 171: Why there is again moment in `moment(endDate)` and `moment(startDate)`? You already did this in the two previous lines?

awalkowiak (changes requested):

awalkowiak (approved):

grails-app/views/report/showRequestDetailReport.gsp
DominikaHatala: @awalkowiak I've changed it to create a date with moment using start of the month so then we can calculate number of chosen months. The previous issue(connected to how moment's diff() works) was that even when there were two different months chosen but difference between them was less then 30 days, it returned 0. So we we have to take beginnings of two chosen months, calculate difference between them and add 1 to get number of chosen months.

PR #2120, opened 2020-12-21
OBPIH-3472 Fix changing quantity available for edit page item after s…

DominikaHatala:
…ubstitution

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -413,6 +413,10 @@ class ProductAvailabilityService {
         return getAvailableBinLocations(location, [product])
     }
 
+    List<AvailableItem> getAllAvailableBinLocations(Location location, Product product) {
+        return getAllAvailableBinLocations(location, [product])
+    }
+
     List<AvailableItem> getAvailableBinLocations(Location location, List products) {
         def availableBinLocations = getQuantityOnHandByBinLocation(location, products)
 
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -428,6 +428,21 @@ class ProductAvailabilityService {
         return availableItems
     }
 
+    // Include also bin locations with negative qty (needed for edit page items)
+    List<AvailableItem> getAllAvailableBinLocations(Location location, List products) {
```
awalkowiak, line 432: Either add the same for single product (and call it inside this on) or make it for a single product instead of list.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1168,7 +1168,7 @@ class StockMovementService {
         Location location = requisitionItem?.requisition?.origin
 
         // Qty Available
-        List<AvailableItem> availableItems = productAvailabilityService.getAvailableBinLocations(location, requisitionItem.product)
+        List<AvailableItem> availableItems = productAvailabilityService.getAllAvailableBinLocations(location, [requisitionItem.product])
```
awalkowiak, line 1171: Use the single product method instead of method for list of products

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
DominikaHatala: @awalkowiak Done.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
DominikaHatala: @awalkowiak Done.

awalkowiak (approved):

PR #2129, opened 2021-01-08
OBPIH-3401: Change default outbound stock movement list sort

awalkowiak:


grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -122,6 +124,7 @@ class Requisition implements Comparable<Requisition>, Serializable {
         id generator: 'uuid'
         requisitionItems cascade: "all-delete-orphan", sort: "orderIndex", order: 'asc', batchSize: 100
 
+        statusSortOrder formula: RequisitionStatus.getStatusSortOrderFormula("status")
```
jmiranda, line 127: Why do we need to pass status? Is there ever going to be another value? If not I'm ok moving this in to the method to make it more readable.

awalkowiak: @jmiranda I wanted it to be flexible if the RequisitionStatus would be used in other domains with different naming. I guess I can remove it for now.

awalkowiak: @jmiranda done

PR #2133, opened 2021-01-11
OBS-540 Fix stock vs ad hov indicator

DominikaHatala:


grails-app/services/org/pih/warehouse/tableroapi/IndicatorDataService.groovy
```diff
@@ -897,11 +897,9 @@ class IndicatorDataService {
...
 
     GraphData getPercentageAdHoc(Location location) {
-        Date today = new Date()
-        Integer previousMonth = today.month
-
-        // if previousMonth = december, we are in the previous year
-        Integer year = previousMonth == 11 ? today.year + 1899 : today.year + 1900
+        Calendar calendar = Calendar.instance
+        calendar.set(calendar.get(Calendar.YEAR), calendar.get(Calendar.MONTH), 1, 0, 0, 0)
+        def firstDayOfMonth = calendar.getTime()
```
jmiranda, line 902: Thanks for refactoring. We should create a util that would make this date fetching more reusable. For now, could you add a comment to explain what we're doing so we don't have to keep re-remembering. I think this is the first day of the month of the current month?

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/tableroapi/IndicatorDataService.groovy
DominikaHatala: @jmiranda Right. I've just added a comment.

awalkowiak (approved):

PR #2139, opened 2021-01-13
OBPIH-3333 Set default put away bin to preferred bin in putaway workflow

DominikaHatala:


src/groovy/org/pih/warehouse/api/PutawayItem.groovy
```diff
@@ -114,12 +114,12 @@ class PutawayItem {
...
                 "recipient.name"              : recipient?.name,
                 currentBins                   : currentBins,
-                preferredBin                  : preferredBin,
+                preferredBin                  : preferredBin?.name,
                 currentBinsAbbreviated        : currentBinsAbbreviated,
                 "putawayFacility.id"          : putawayFacility?.id,
                 "putawayFacility.name"        : putawayFacility?.name,
-                "putawayLocation.id"          : putawayLocation?.id,
-                "putawayLocation.name"        : putawayLocation?.name,
+                "putawayLocation.id"          : putawayLocation?.id ?: preferredBin?.id,
```
awalkowiak, line 121: I don't like this. This indicates that despite `pytawatLocation` is null, the returned map will state that it is existing. I think the `preferredBin.id` should be a separate field in this map.

src/groovy/org/pih/warehouse/api/PutawayItem.groovy
```diff
@@ -114,12 +114,12 @@ class PutawayItem {
...
+                "putawayLocation.name"        : putawayLocation?.name ?: preferredBin?.name,
```
awalkowiak, line 122: Same case as above.

awalkowiak (changes requested):

awalkowiak (approved):

PR #2157, opened 2021-01-24
OBPIH-3420: Add filters to request detail report

pmuchowski:


grails-app/services/org/pih/warehouse/forecasting/ForecastingService.groovy
```diff
@@ -271,16 +270,39 @@ class ForecastingService {
                 DATE_FORMAT(date_issued, '%d/%b/%Y') as date_issued,
                 origin_name,
                 destination_name,
-                product_code,
+                product_demand_details.product_code,
```
awalkowiak, line 273: Why need for this only here?

pmuchowski: because there is field with the same name in the products table

awalkowiak (approved):

PR #2158, opened 2021-01-25
OBPIH-3514:  Get quantity from product availability table on Putaway

awalkowiak:


awalkowiak:
@jmiranda I pushed my work to another branch (instead of https://github.com/openboxes/openboxes/tree/OBPIH-3514) because I additionally refactored few other places where the inventory service was used and I added a new method for getting getQuantityOnHandByBinLocation with only rows within stock instead of passing boolean as you did (which should be a bit faster than `.findAll{}`). Please take a look and let me know which version you like more. Right now there is only one spot left where inventoryService is used (and quantity is not pulled from product availability), it's here: https://github.com/openboxes/openboxes/blob/OBPIH-3514-v2/grails-app/services/org/pih/warehouse/putaway/PutawayService.groovy#L150

PR #2159, opened 2021-01-25
OBPIH-3536: Redirect to stock movement view page on global search of …

awalkowiak:
…shipment


grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -78,10 +78,10 @@ class DashboardController {
 
         def shipment = Shipment.findByShipmentNumber(params.searchTerms)
         if (shipment) {
-            if (shipment?.isStockMovement()) {
+            if (shipment?.requisition) {
```
DominikaHatala, line 81: I think we should remove this condition and have one redirect to stock movement's details page.

DominikaHatala (changes requested):

DominikaHatala (approved):

PR #2165, opened 2021-01-27
OBPIH-3534 Add word break for shipment name in pending inbound and ou…

DominikaHatala:
…tbound tabs

### DIFF
--- a/grails-app/views/inventoryItem/_showPendingInboundStock.gsp
+++ b/grails-app/views/inventoryItem/_showPendingInboundStock.gsp
@@ -63,7 +63,7 @@
                                 </g:link>
                             </g:else>
                         </td>
-                        <td class="center">
+                        <td class="center" style="word-break: break-word;">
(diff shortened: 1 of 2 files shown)


grails-app/views/inventoryItem/_showPendingInboundStock.gsp
```diff
@@ -63,7 +63,7 @@
                                 </g:link>
                             </g:else>
                         </td>
-                        <td class="center">
+                        <td class="center" style="word-break: break-all;">
```
awalkowiak, line 66: Is it ok to break names in the middle of a word? Perhaps the `break-word` instead of `break-all` would be better?

DominikaHatala: You're right. I've tried break-word but in a bad place and assumed it's not working. I will change it.

awalkowiak (approved):

PR #2172, opened 2021-02-01
OBPIH-3515: Add Preferred bin location to import Inventory Level temp…

pmuchowski:
…late


grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -167,7 +169,19 @@ class DataService {
 
             // Create inventory level for current location, include bin location
             if (location.inventory) {
-                addInventoryLevelToProduct(product, location.inventory, row.binLocation, row.minQuantity, row.reorderQuantity, row.maxQuantity, row.preferredForReorder)
+                Location preferredBinLocation = null
+
+                if (row.preferredBinLocation) {
+                    preferredBinLocation = location.getBinLocations().find {
```
jmiranda, line 175: We should probably do a search here or else this is going to be pretty expensive for large. Even that is going to be expensive (N+1 query) so let's think about the best way to do this without requiring a ton of new queries.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/DataService.groovy
jmiranda: @pmuchowski Looks like you were correct. It doesn't impact performance, even with thousands of bin locations.

jmiranda (approved):

PR #2184, opened 2021-02-09
OBPIH-3596 Set purchasing organization to destination's party on PO c…

DominikaHatala:
…reation


awalkowiak (approved):

jmiranda (approved):
I guess this works. It's hard to see what's going on without running the code, but I'll assume this is doing what we want.

PR #2187, opened 2021-02-10
OBPIH-3541 Added a start delay to the refresh product availability jo…

jmiranda:
…b when triggered after a transaction is created in order to allow for data to be flushed to the database (need to find a better solution)


jmiranda:
@awalkowiak Review and merge if you think it looks ok. Then investigate whether there's a better solution.

PR #2188, opened 2021-02-11
OBPIH-3603 Change email templates for shipping and receipt notifications

DominikaHatala:


grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -133,7 +133,7 @@ class NotificationService {
         def recipientItems = shipmentInstance.shipmentItems.groupBy {it.recipient }
         recipientItems.each { Person recipient, items ->
             if (emailValidator.isValid(recipient?.email)) {
-                def subject = g.message(code: "email.yourItemShipped.message", args: [shipmentInstance.shipmentNumber])
+                def subject = g.message(code: "email.yourItemShipped.message", args: [shipmentInstance.origin, shipmentInstance.destination, shipmentInstance.shipmentNumber])
```
jmiranda, line 136: We should ask Kelsey if this should be the location.locationNumber or location.name

jmiranda (changes requested):
LGTM ... just make the change below.

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
DominikaHatala: In example in the ticket there is "Please be informed that an item you requested has arrived at Miami Warehouse: Haiti Stock." so I assume that is should be name.

jmiranda: Ok, but we still need to reference the name propery explicitly in case the toString() ever changes.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
DominikaHatala: @jmiranda Done.

grails-app/views/email/_shipmentItemReceived.gsp
```diff
@@ -8,7 +13,7 @@
             ${warehouse.message(code: 'email.user.label', args: [recipient?.firstName]) },
         </div>
         <div>
-            ${warehouse.message(code: 'email.shipmentItemReceived.label') }:
+            ${warehouse.message(code: 'email.shipmentItemReceived.label', args: [shipmentInstance?.destination]) }:
```
jmiranda, line 16: Same thing here. We should use the name property rather than relying on the toString() method.

jmiranda (changes requested):
@awalkowiak Once the issue below is fixed please merge this. We need to get the hotfix tested and released this week.

awalkowiak (approved):

PR #2191, opened 2021-02-12
OBPIH-3488 Add product type to product creation

pmuchowski:


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -198,10 +198,12 @@ class ProductController {
         // when the session is closed.
         if (!productInstance?.id || productInstance.validate()) {
             if (!productInstance.productCode) {
-                productInstance.productCode = productService.generateProductIdentifier()
+                productInstance.productCode = productService.generateProductIdentifier(productInstance.productType)
             }
         }
 
+        productInstance.validateRequiredFields()
```
awalkowiak, line 205: Is this line required? What happens when it fails or succeeds?

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -294,10 +296,11 @@ class ProductController {
                 // when the session is closed.
                 if (productInstance.validate()) {
                     if (!productInstance.productCode) {
-                        productInstance.productCode = productService.generateProductIdentifier()
+                        productInstance.productCode = productService.generateProductIdentifier(productInstance.productType)
                     }
                 }
 
+                productInstance.validateRequiredFields()
```
awalkowiak, line 303: Same as line 205 also seems like productInstance is validated few lines above.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: > Is this line required? 

From my understanding, this seems critical and the only way to do it. I guess if you could somehow move this into the validate method i.e. as a global constraint on the object rather than a property constraint (although not sure if Grails supports that) then we might not need to call it explicitly.
 
> What happens when it fails or succeeds?

The method is similar to generic "validate()" in that it adds validation errors to the Errors object, which is interrogated through the productInstance.hasErrors() method below. So this seems like the right approach. 

As I mentioned above, the only better option would be to have this called implicitly within the validate() method, but I don't think Grails supports that without doing something ugly like rewiring Product class's static validate methods similar to how we do it with unit tests.

```
Product.metaClass.static.validate = { return validate() and validateRequiredFields() }
```

You could try to add that in Bootstrap.groovy or resources.groovy and see what happens.

jmiranda (approved):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
awalkowiak: Ok, I understand it better now. Thanks for the explanation.

awalkowiak (approved):

PR #2193, opened 2021-02-15
OBPIH-3592 Add product source preference export

DominikaHatala:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -803,9 +803,21 @@ class DataService {
 
     def transformObject(Object object, Map includeFields) {
         Map properties = [:]
+        def value = null
```
jmiranda, line 806: I feel like this needs to be inside the `each` so we clear the value on each iteration and don't do something that accidentally causes `value` to contain data from a previous iteration.

grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -803,9 +803,21 @@ class DataService {
...
         includeFields.each { fieldName, property ->
-            def value = property.tokenize('.').inject(object) { v, k -> v?."$k" }
-            properties[fieldName] = value ?: ""
+            if (property instanceof LinkedHashMap) {
+                value = property.value.tokenize('.').inject(object) { v, k -> v?."$k" }
+                if (property.format && value) {
+                    value = value.format(property.format)
+                }
+                if (property.defaultValue && !value) {
```
jmiranda, line 813: We should support a property that has a default value and requires a format ... 
```[propertyName: "someDomain.someDateProperty", default: new Date(), format: "MM/dd/yyyy"]```

File: grails-app/services/org/pih/warehouse/data/DataService.groovy [hunk 3]
```diff
@@ -803,9 +803,21 @@ class DataService {
...
     def transformObject(Object object, Map includeFields) {
         Map properties = [:]
+        def value = null
         includeFields.each { fieldName, property ->
-            def value = property.tokenize('.').inject(object) { v, k -> v?."$k" }
-            properties[fieldName] = value ?: ""
+            if (property instanceof LinkedHashMap) {
+                value = property.value.tokenize('.').inject(object) { v, k -> v?."$k" }
+                if (property.format && value) {
+                    value = value.format(property.format)
```
jmiranda, line 811: I assume this assumes value is a Date object. Should we formalize that? What would happen if value was a String?

jmiranda (changes requested):
I have a few requests and questions, but this was a really solid start.

grails-app/domain/org/pih/warehouse/product/ProductSupplierPreference.groovy
```diff
@@ -46,4 +46,14 @@ class ProductSupplierPreference {
         updatedBy(nullable: true)
         createdBy(nullable: true)
     }
+
+    static PROPERTIES = [
+            "Product Source Code"                  : "productSupplier.code",
+            "Organization Code"                    : "destinationParty.code",
+            "Organization Name"                    : ["value": "destinationParty.name", "defaultValue": "DEFAULT"],
+            "Preference Type"                      : "preferenceType.name",
+            "Preference Type Validity Start Date"  : ["value": "validityStartDate", "format": "MM/dd/yyyy"],
```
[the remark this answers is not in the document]
awalkowiak: `propertyName` sounds good, but it sometimes could be a `path` so perhaps the sole `property` will be less confusing comparing it to two other options?

grails-app/services/org/pih/warehouse/data/DataService.groovy
DominikaHatala: @jmiranda I couldn't find a place where date is a string so I don't think that we should formalize that. We should use format only for date formatting.

grails-app/services/org/pih/warehouse/data/DataService.groovy
DominikaHatala: That's right, I will add it

grails-app/domain/org/pih/warehouse/product/ProductSupplierPreference.groovy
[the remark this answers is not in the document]
DominikaHatala: I agree it should be "property"

DominikaHatala:
@jmiranda I have updated this PR.

grails-app/services/org/pih/warehouse/data/DataService.groovy
jmiranda: Sorry, by "formalize" I meant that we should handle each case separately. This code works because you only use the format attribute for dates, but someone could set value to a string or any other object, set the format attribute, and then this would fail. 

So use instanceof or some other means to formally (explicitly) handle these cases. Since string objects don't have a format method (note that String.class does have a static format method) we should probably throw an exception.

DominikaHatala: @jmiranda Okay, sure I will.

jmiranda: Actually, here's one thing we can do for now just so we don't get bogged down. Change "format" to "dateFormat", then we can assume the object is a date.

DominikaHatala: @jmiranda Done.

jmiranda (approved):

PR #2194, opened 2021-02-15
OBPIH-3544 Add document availability fields to Product Supplier

awalkowiak:


grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
```diff
@@ -61,6 +61,12 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
     // Additional comments
     String comments
 
+    // Documents Availability
+    Boolean coaAvailable = Boolean.FALSE
+    Boolean isoAvailable = Boolean.FALSE
```
jmiranda, line 66: When we talked about this either (during planning poker?) I said that I wanted this to be done using the Attribute domain in manner similar to ProductAttribute. It doesn't look like that was recorded in the ticket so assign this to me and I'll do it myself.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
awalkowiak: I forgot about this. I think could fix that, but I need a broader understanding of how you see it.

PR #2195, opened 2021-02-16
OBPIH-3488: Change product fields when product type changed

pmuchowski:


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1822,6 +1816,23 @@ class JsonController {
             throw new IllegalArgumentException("Start and end date are required")
         }
     }
+
+    def checkIfProductFieldRemoved = {
```
jmiranda, line 1820: The guts of this should probably go in a method on the ProductType class but we can leave that for now.

jmiranda (approved):

PR #2196, opened 2021-02-17
OBPIH-3405 Change validation for existing sources on PO import

DominikaHatala:


grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,9 +141,27 @@ class ProductSupplierDataService {
...
-        def productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        def productSupplier
+        if (params.productSupplier) {
+            productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        } else {
+            productSupplier = ProductSupplier.createCriteria().list {
+                eq("product", params.product)
+                if (params.supplier && params.supplierCode && params.product) {
+                    eq("supplier", params.supplier)
+                    eq("supplierCode", params.supplierCode.replaceAll('[,-]',''))
```
jmiranda, line 152: Looks like you missed whitespace.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,9 +141,27 @@ class ProductSupplierDataService {
...
+                    // TODO we need to find a way to replace dashes in attribues, solutions below are not working
+                    // sqlRestriction("REPLACE('{alias}.supplierCode','-','')");
+                    // def code = params.supplierCode.replaceAll('[,-]','')
+                    // sqlRestriction("replace(supplierCode, '-', '') = '%$code%'")
+                } else if (!params.supplierCode) {
+                    eq("manufacturer", params.manufacturer)
+                    eq("manufacturerCode", params.manufacturerCode.replaceAll('[,-]',''))
```
jmiranda, line 159: Same here.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,9 +141,27 @@ class ProductSupplierDataService {
...
+                }
+            }
+        }
         if (productSupplier) {
-            return productSupplier
+            return productSupplier instanceof ArrayList ? productSupplier.first() : productSupplier
```
jmiranda, line 164: We should be able to do this in the query using maxResults(1)

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,9 +141,27 @@ class ProductSupplierDataService {
...
+        def productSupplier
+        if (params.productSupplier) {
+            productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        } else {
+            productSupplier = ProductSupplier.createCriteria().list {
+                eq("product", params.product)
+                if (params.supplier && params.supplierCode && params.product) {
+                    eq("supplier", params.supplier)
+                    eq("supplierCode", params.supplierCode.replaceAll('[,-]',''))
+                    // TODO we need to find a way to replace dashes in attribues, solutions below are not working
```
jmiranda, line 153: I think this could be done in beforeInsert/beforeUpdate. We could strip these characters out before saving to the database. That might not be what Kelsey wants so I'll see if there's another way of doing this in SQL/HQL. You also need to make the filter case-insensitive so a toLowerCase/toUpperCase on each side.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -615,18 +615,17 @@ class OrderService {
...
-                                                  supplierCode: supplierCode,
-                                                  manufacturerCode: manufacturerCode,
-                                                  supplier: supplier,
-                                                  sourceName: sourceName]
-                            ProductSupplier productSupplier = productSupplierDataService.getOrCreateNew(supplierParams)
-                            orderItem.productSupplier = productSupplier
-                        }
+                    } else {
+                        Organization supplier = Organization.get(supplierId)
+                        Organization manufacturer = Organization.findByName(manufacturerName)
```
jmiranda, line 620: I need to look into this a bit more. This doesn't feel right.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
awalkowiak: In the ticket there was a request for `dash`, `space`, and `dot`? Here is `comma` and `dash`

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
awalkowiak: `dash`, `space` and `dot`

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
awalkowiak: I think, that after adding `maxResult` it still will be a list. Would  it be also helpful if you additionally change the `ProductSupplier.createCriteria().list` into `.get`? Can it work together with `maxResult`?

grails-app/services/org/pih/warehouse/order/OrderService.groovy
awalkowiak: Where is the check for `manufacturer and manufacturer code match`? Additionally, the name is not unique. Perhaps it should be searched by code and then compared to name?

awalkowiak (changes requested):

DominikaHatala:
@jmiranda  @awalkowiak I have updated this PR and created sql query instead using createCriteria. The only thing I'm not yet sure about is finding manufacturer by name.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
DominikaHatala: @awalkowiak manufacturer code given by user is not the same that "manufactrer.code" is. I'm still not sure how this should work. In exported data there is manufacturer's name given so I thought that users would always give name in this column.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,7 +142,39 @@ class ProductSupplierDataService {
...
 
     def getOrCreateNew(Map params) {
-        def productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        def productSupplier
+        def data
+        if (params.productSupplier) {
+            productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        } else {
+            String query = """
+                select
```
jmiranda, line 151: On the standup this morning I realized we could convert this query into a view that hides the complexity of what is happening. Essentially we'd return all of the same columns as the table, but we'd clean certain columns. We could honestly do this for a bunch of tables and maybe even use HQL/SQL to query the view and hydrate objects. I think you could do something like this https://blog.jdriven.com/2014/03/grails-goodness-using-hibernate-native-sql-queries/

```
def sessionFactory

def getOrCreateNew(Map params) { 
    sessionFactory.currentSession.createSQLQuery("select <columns> from product_supplier_view where <filters>") 
```

NOTE: I've started using a pattern to avoid the nasty branching logic. See the example below to see where you can basically ignore a filter by using `IFNULL(<user-param>, column_name)`

```
String forceRefreshStatement = String.format(
                            "delete from product_availability " +
                                    "where location_id = '${location.id}' " +
                                    "and product_id = IFNULL(%s, product_id);", productId)
                    stmt.addBatch(forceRefreshStatement)
```
This would be better as a prepared statement like you have below so see if you can figure that out.

jmiranda: Here's some pseducode for a full solution
```
CREATE OR REPLACE VIEW product_supplier_clean_thingy AS (
    SELECT 
        id,
        code,
        name,
        supplier_id,
        REGEXP_REPLACE(supplier_code, '[ .,-]+', ''),
        ...
    FROM product_supplier
)

def getOrCreateNew(Map params) { 

    ProductSupplier productSupplier = getProductSupplier(params)
    if (!productSupplier) {
        productSupplier = createProductSupplierWihtoutPackage(params)
    }
    return productSupplier
}

def getProductSupplier(Map params) {

    final session = sessionFactory.currentSession

    // FIXME Eventually we'd like the regular expression to be configurable, 
    // but I'm not sure how that would work in the view above so we can 
    // ignore that for now
    String supplierCode = params.supplierCode ? params.supplierCode.replaceAll('[ .,-]+','') : null
    ...

    // Query string with :startId as parameter placeholder.
    final String query = """
        select id, ... 
        from product_supplier_clean_thingy  
        where product_id = :productId
        AND supplier_id = IFNULL(:supplierId, supplierId)
        AND supplier_code = IFNULL(:supplierCode, supplierCode)
        ...        
    """

    // Create native SQL query.
    final sqlQuery = session.createSQLQuery(query)

    // Use Groovy with() method to invoke multiple methods
    // on the sqlQuery object. 
    // FIXME Not sure if this works with our version but I've seen similar solutions with Grails 1.3.9
    final results = sqlQuery.with {
        // Set domain class as entity.
        // Properties in domain class id, name, level will
        // be automatically filled.
        addEntity(ProductSupplier)

        // Set value for parameter startId.
        setString('productId', productId)
        setString('supplierCode', supplierCode)

        // Get all results.
        // FIXME Not sure if you can use get() or max/limit here. If not we'll just check 
        // the results to see if there's more than one and return the first result
        list()
    }

    results
}
```

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -141,7 +142,39 @@ class ProductSupplierDataService {
...
+            productSupplier = params.productSupplier ? ProductSupplier.get(params.productSupplier) : null
+        } else {
+            String query = """
+                select 
+                    id
+                FROM product_supplier
+                WHERE product_id = :productId
+                """
+            if (params.supplierId && params.supplierCode) {
+                query += " AND supplier_id = :supplierId AND REPLACE(REPLACE(REPLACE(supplier_code, ' ', ''), '-', ''), '.', '') = :supplierCode "
```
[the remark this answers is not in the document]
DominikaHatala: We're using MySQL 5.7 version and it's not possible to use regex.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
DominikaHatala: I've tried creating native sql query but it didn't work. I had weird errors that "version" column does not exist and also some other strange. I've updated PR with other changes that you have suggested.

DominikaHatala:
@jmiranda I have updated PR

jmiranda (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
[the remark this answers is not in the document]
jmiranda: Makes sense.

PR #2200, opened 2021-02-18
OBPIH-3442 Add preference type logic in PO

DominikaHatala:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1738,25 +1740,33 @@ class JsonController {
...
-        productSuppliers = productSuppliers.collect {[
-            id: it.id,
-            code: it.code,
-            supplierCode: it.supplierCode,
-            text: it.code + ' ' + it.name,
-            manufacturerCode: it.manufacturerCode,
-            manufacturer: it.manufacturer?.id,
-        ]}
+        productSuppliers = productSuppliers.collect {
+            ProductSupplierPreference preference = it.productSupplierPreferences.find { it.destinationParty == destinationParty }
```
jmiranda, line 1749: Is there any reason not to move this into the query above. If not, ProductSupplier probably needs its own service at this point. So let's create ProductSupplierService and add a method to retrieve ProductSuppliers using the given criteria.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -623,6 +626,10 @@ class OrderController {
         Order order = Order.get(params.order.id)
         OrderItem orderItem = OrderItem.get(params.orderItem.id)
         ProductSupplier productSupplier = null
+        if (params.validationCode == 'BLOCK') {
```
jmiranda, line 629: Let's use the enum ValidationCode.BLOCK.

DominikaHatala: I'ts not working with ValidationCode.BLOCK

grails-app/controllers/org/pih/warehouse/JsonController.groovy
DominikaHatala: Which query do you mean? What we do above is checking if preference for this destination party exists and if it exists then we check if validation code is different than hide (and then we include this product source in select) or if preference for this destination party doesn't exist at all (and then we also include it). I'm not really sure if I know how would you like it to look like.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
jmiranda: Sorry you'd need to convert params.validationCode to ValidationCode enum. Something like this should work but you'd need to do some error handling for when params.validationCode is null.
```
ValidationCode validationCode = params.validationCode ? params.validationCode as ValidationCode : null
if (validationCode == ValidationCode.BLOCK) { 
```

Other options that might also work ... but I prefer the above approach
```
ValidationCode validationCode = params.validationCode as ValidationCode
if (params.validationCode == ValidationCode.BLOCK.toString()) { 
```
or 
```
ValidationCode validationCode = params.validationCode as ValidationCode
if (params.validationCode == ValidationCode.BLOCK.name()) { 
```

grails-app/controllers/org/pih/warehouse/JsonController.groovy
jmiranda: Yeah I guess I was wondering if we could make it work so you can match on preferences in the query.
```
def productSuppliers = ProductSupplier.createCriteria().list { 
  eq("product", product)
  eq("supplier", suppler)
  productSupplierPreferences { 
    and { 
       eq("destinationParty", destinationParty)
       not { 
          eq("validationCode", ValidationCode.HIDE)
       }
    }
}
```
}

jmiranda: You might need to include something like this 
```
(productSupplierPreferences is empty or ...)
```

DominikaHatala:
@jmiranda PR is updated.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -21,12 +21,15 @@ import org.pih.warehouse.core.Document
...
 import org.pih.warehouse.core.UomService
 import org.pih.warehouse.core.User
+import org.pih.warehouse.core.ValidationCode
 import org.pih.warehouse.product.Product
 import org.pih.warehouse.product.ProductSupplier
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentItem
 import org.springframework.web.multipart.MultipartFile
 
+import javax.xml.bind.ValidationException
```
jmiranda, line 31: Not the right ValidationException. It should be a grails class.

jmiranda: Doesn't look like this is being used so you can probably get rid of the import.

jmiranda (changes requested):
LGTM other than the issue I brought up below. Once that is fixed feel free to merge.

awalkowiak (approved):

PR #2207, opened 2021-02-20
OBPIH-3544 Add document availability to product supplier

jmiranda:


grails-app/views/attribute/edit.gsp
```diff
@@ -94,6 +94,14 @@
                                     </a>
 								</td>
 							</tr>
+                            <tr class="prop">
+                                <td valign="top" class="name">
+                                    <label for="entityTypeCodes"><g:message code="entityTypeCodes.label" default="Entity Type Codes" /></label>
+                                </td>
+                                <td valign="top" class="value ${hasErrors(bean: attributeInstance, field: 'entityTypeCodes', 'errors')}">
+                                    <g:selectEntityTypeCode name="entityTypeCodes" value="${attributeInstance?.entityTypeCodes}" class="chzn-select-deselect"/>
```
awalkowiak, line 102: On edit, there is a multi-select, but when creating it is as a single-select. Should this be unified?

awalkowiak (approved):
Looks good to me. I like this approach. It was a bit fuzzy for me at the beginning, but now I see the whole picture and it is a way better approach. I have one minor question you can find below regarding the multi-select.

grails-app/views/attribute/edit.gsp
jmiranda: I initially had it as a multipe but realized there were edge cases I didn't want to handle.

PR #2210, opened 2021-02-22
Improve getLocationsSupportingActivity performance

alsanchez:
This method is a tad inefficient when you have many locations, since it needs to retrieve all the records from the database and hydrate them before filtering.

With the proposed changes the filtering is done on the database, which improves performance for this use case.

Anecdotal benchmark: On our instance with 150K locations it cut down the load time of the New Purchase Order page from 60+ seconds to 0.2 seconds.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/LocationService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/LocationService.groovy
@@ -292,8 +292,17 @@ class LocationService {
     }
 
     List getLocationsSupportingActivity(ActivityCode activity) {
-        return getAllLocations()?.findAll { it.supports(activity) }
+        def locations = [:]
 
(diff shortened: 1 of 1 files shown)


jmiranda:
Wow, thank you @alsanchez. I just created a bunch of performance tickets and this was one of the top issues. Thanks so much!!

awalkowiak (approved):
LGTM.
I tried this solution locally and compared it to this single query:
```
dataService.executeQuery("""
    select l.id, l.name FROM location as l
        left outer join location_type_supported_activities as ltsa on l.location_type_id = ltsa.location_type_id 
        left outer join location_supported_activities as lsa on l.id = lsa.location_id 
    WHERE ltsa.supported_activities_string = :activity OR lsa.supported_activities_string = :activity
    GROUP BY l.id, l.name
    """, [
        activity: activity.id
])
```
The version from this PR has a better performance.

jmiranda:
Merged. Thanks again @alsanchez!

PR #2218, opened 2021-02-24
OBPIH-3622 Fix existing ProductPrice data (set date_created and last_…

awalkowiak:
…updated)

### DIFF
--- a/grails-app/migrations/0.8.x/changelog-2021-02-24-1700-fix-product-price-data.xml
+++ b/grails-app/migrations/0.8.x/changelog-2021-02-24-1700-fix-product-price-data.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <changeSet author="awalkowiak" id="2402202117000-0">
+    <preConditions onFail="MARK_RAN">
+      <and>
+        <tableExists tableName="product_price"/>
+        <not>
+          <sqlCheck expectedResult="0">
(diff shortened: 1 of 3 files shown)


grails-app/migrations/0.8.x/changelog-2021-02-24-1700-fix-product-price-data.xml
```diff
@@ -0,0 +1,23 @@
...
+        <tableExists tableName="product_price"/>
+        <not>
+          <sqlCheck expectedResult="0">
+            SELECT COUNT(*) FROM product_price WHERE date_created IS NULL OR last_updated IS NULL
+          </sqlCheck>
+        </not>
+      </and>
+    </preConditions>
+    <sql>
+      UPDATE product_price
```
jmiranda, line 18: These should be two separate changesets (one for date_created and one for last_updated).

jmiranda: I don't really understand why this is an issue. How did these product_price records get created?

jmiranda (changes requested):

grails-app/migrations/0.8.x/changelog-2021-02-24-1700-fix-product-price-data.xml
awalkowiak: It was a migration of existing prices (when float was refactored into ProductPrice) and I forgot that date_created and last_updated can't be null.

jmiranda: And the reason I'm asking is because if it's a PIH specific issue, then we should just run these updates manually on the effected servers and not add these database migrations if they will never be triggered in any other system.

awalkowiak: Ok. I think these kinds of migrations could be discussed in the tech huddle meetings.

jmiranda: Ok, if the offending product prices were created in another database migration then we can merge this database migration as well. I'll review and merge it in a little while. But just to reiterate, the issue should be fixed on OBDEV so this should not block OBPIH-3589.

jmiranda (approved):

PR #2224, opened 2021-02-26
OBPIH-3662 Prevent user from creating empty combined shipment

DominikaHatala:


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -978,6 +978,11 @@ class OrderController {
 
     def createCombinedShipment = {
         def orderInstance = Order.get(params.orderId)
+        if (!orderInstance.orderItems.find {it.quantityRemainingToShip != 0 }) {
+            flash.error = "${warehouse.message(code:'purchaseOrder.noItemsToShip.label')}"
```
jmiranda, line 982: Just use flash.message. It's more warning / informational and not really an error anyway.

jmiranda (commented):
Otherwise looks good.

awalkowiak (approved):

PR #2226, opened 2021-02-26
OBPIH-3589 Fix Product Source import validation

pmuchowski:


grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -69,8 +63,8 @@ class ProductSupplierDataService {
                 command.errors.reject("Row ${index + 1}: Manufacturer with name '${manufacturerName}' does not exist")
             }
 
-            if (ratingTypeCode && !RatingTypeCode.values().any { it.name == ratingTypeCode }) {
-                command.errors.reject("Row ${index + 1}: Rating Type with value '${ratingTypeCode}' does not exist")
+            if (ratingType && !RatingTypeCode.values().any { it.name == ratingType.toString().toUpperCase() }) {
```
jmiranda, line 66: Could you explain this change?

jmiranda (commented):
Looks good, I just want to hear what's going on with the rating type field.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
jmiranda: Ok, so I read the comments in the ticket and now I understand the issue. There are two ways I would prefer to have this done. First we should be initializing the value 

So either of the following two are fine. 
```
def ratingTypeCode = params?.ratingTypeCode?.toUpperCase() as RatingTypeCode
```
OR
```
def ratingTypeCode = RatingTypeCode.valueOf(params?.ratingTypeCode?.toUpperCase()))
```
The above declaration will likely cause a hard failure on invalid values so the validation below is probably not necessary (I'm ok with that). But for completeness, let's include it. I feel uncomfortable converting to a string, so I think I'd prefer the following. 
```
if (ratingTypeCode && !(ratingTypeCode in RatingTypeCode.values()))
```
Or add a validation method that can be more easily unit tested to the enum itself.
```
RatingTypeCode.inList(RatingTypeCode)
```

jmiranda (changes requested):

jmiranda (approved):

awalkowiak (approved):

PR #2229, opened 2021-03-01
OBPIH-3594 Enable removing preference types for product sources

DominikaHatala:


grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -155,12 +158,15 @@ class ProductSupplierController {
                     defaultPreference.productSupplier = productSupplierInstance
                     productSupplierInstance.addToProductSupplierPreferences(defaultPreference)
                 }
+            } else {
+                if (defaultPreference) {
```
awalkowiak, line 162: This could be in one `else if` block

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -170,6 +176,11 @@ class ProductSupplierController {
                     productSupplierPreference.productSupplier = productSupplierInstance
                     productSupplierInstance.addToProductSupplierPreferences(productSupplierPreference)
                 }
+            } else {
+                if (productSupplierPreference) {
```
awalkowiak, line 180: This could be in one `else if` block

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -155,12 +158,15 @@ class ProductSupplierController {
                     defaultPreference.productSupplier = productSupplierInstance
                     productSupplierInstance.addToProductSupplierPreferences(defaultPreference)
                 }
+            } else {
+                if (defaultPreference) {
+                    productSupplierInstance.removeFromProductSupplierPreferences(defaultPreference)
+                    defaultPreference.delete()
```
awalkowiak, line 164: Should we allow removing global preference types?

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
DominikaHatala: Yes. It's not required anywhere.

awalkowiak (approved):

PR #2230, opened 2021-03-01
OBPIH-3663 Pass list page params when removing stock movement

DominikaHatala:


awalkowiak (approved):

jmiranda (commented):
@DominikaHatala I'll be honest, I'm nervous about the removal of the `.id` here. Was there a reason for doing that?

DominikaHatala:
@jmiranda When passing those parameters with .id there was error displayed "No such property: id for class: java.lang.String" because params were converted to strings while passing from remove method to list method. Why are you nervous about the removal of it?

jmiranda:
Because it's been working with IDs for years and there are potentially other places where we redirect to the list page and we're expecting parameters with names like `object.id`.

DominikaHatala:
@jmiranda In order to make deleting items work properly we need to pass all parameters so they're not changed when stock movement is removed. Initially I have fixed it to only not change inbound/outbound but then while testing I've noticed that it's confusing when you add some filters like origin, status and you're on other page then first one and then when you delete shipment there are no more filters applied and you're on the first page again. I've tried fixing this "No such property: id for class: java.lang.String" issue but it was always passed as String to list action and error always occured.

jmiranda:
Closing this PR in favor of https://github.com/openboxes/openboxes/pull/2238

PR #2231, opened 2021-03-01
OBPIH-3668 Set contract price and date when adding new product source

DominikaHatala:


awalkowiak (approved):

DominikaHatala:
@jmiranda we could use just one action for both - save and update. I can try to refactor it.

awalkowiak:
@jmiranda yeah you're right, seems like we could use save for create and update here.

However last time I was including this price parsing I was thinking if we could refactor it to some sort of parsing util function. I don't have anything particular in my mind yet. I wanted to investigate it further when I would have to add it again.

jmiranda:
@awalkowiak @DominikaHatala Yeah let's put our heads together during planning poker and try to refactor it. I'd be happy with moving the code to a controller method or better yet, a service method that throws an IllegalArgumentException or ValidationException that can be handled by the controller. I actually code that I just worked on last week that does something similar and I'd like to refactor that as well. See the updateAttributes() controller method in ProductController and ProductSupplierController. They are slightly different but essentially do the same thing.

PR #2232, opened 2021-03-01
OBPIH-3665 Fix display of demand tab

DominikaHatala:


grails-app/services/org/pih/warehouse/forecasting/ForecastingService.groovy
```diff
@@ -331,6 +332,7 @@ class ForecastingService {
         }
 
         data = data.collect {
+            def quantityIssued = RequisitionItem.get(it?.request_item_id)?.getQuantityIssued()
```
awalkowiak, line 335: Why this can't be included as a join in the query?

DominikaHatala: @jmiranda I remember that once we've discussed why quantity issued is not included in product demand table but unfortunatelly I don't remember the reason. Do you have any idea why it's not included there? (quantity issued is also displayed in demand tab and it's done the same way)

jmiranda: Because we cannot easily calculate it and quantity picked should be the same as quantity issued. If it's not then there are likely data errors.

jmiranda: Definitely don't do this. In fact, any time you find yourself doing a query inside a loop, please stop and bring it to my attention. This is going to lead to very bad performance.

jmiranda (changes requested):

awalkowiak (approved):

jmiranda (approved):

PR #2235, opened 2021-03-02
OBPIH-3590 Add flush for proper adding new attribute to product

awalkowiak:


jmiranda (commented):
I don't think this will resolve the issue. Whenever I see that "was not processed by flush()" it's usually because there's a validation error that we're not handling. However, I haven't been able to reproduce this locally.

jmiranda:
I'm going to merge the PR just to keep things moving, but please look at my comments in the ticket.

PR #2236, opened 2021-03-02
OBPIH-3664 Fix reason code classification in the product demand view

awalkowiak:


awalkowiak:
@jmiranda I guess this can hold since the additional column for the cancel or pick reason code needs to be included

PR #2237, opened 2021-03-02
OBPIH-3671 Fix french confirmation message on PO edit lines page

DominikaHatala:
Each time we use message as a string in java script we need to remamber to escape quotes.

### DIFF
--- a/grails-app/views/purchaseOrder/_showOrderItems.gsp
+++ b/grails-app/views/purchaseOrder/_showOrderItems.gsp
@@ -471,7 +471,7 @@
             var data = $("#orderItemForm").serialize();
             data += '&orderIndex=' + $("#orderItemsTable tbody tr").length;
             if (validateItemsForm()) {
-              if ($("#validationCode").val() == 'WARN' && !confirm('${warehouse.message(code: 'orderItem.warningSupplier.label')}')) {
+              if ($("#validationCode").val() == 'WARN' && !confirm('${warehouse.message(code: 'orderItem.warningSupplier.label').replaceAll( /(')/, '\\\\$1' )}')) {
                 return false
               } else {
                 $.ajax({


grails-app/i18n/messages_fr.properties
```diff
@@ -826,7 +826,7 @@ orderCommand.shippedOn.invalid.required=Propriété [{0}] de la classe [{1}] ave
 # Order Item messages
 orderItem.list.label=Liste d'articles en cours
 orderItem.missingGlAccount.label=Impossible d'ajouter un article de commande sans code de grand livre valide
-orderItem.warningSupplier.label=Cette source a le type de préférence "low quality" pour votre destination. Voulez-vous toujours l'ajouter au bon de commande?
+orderItem.warningSupplier.label=Cette source a le type de préférence \\"low quality\\" pour votre destination. Voulez-vous toujours l\\'ajouter au bon de commande?
```
jmiranda, line 829: Look at the line above this. It seems like we're going to run into this a lot. So is it possible to do the escaping on the javascript side?

jmiranda: @awalkowiak Thoughts?

jmiranda (changes requested):

grails-app/i18n/messages_fr.properties
jmiranda: @awalkowiak If you guys figure this out don't worry about me reviewing, just merge it if it works.

DominikaHatala: @awalkowiak @jmiranda I've figured it out - I replace all single quotes with \' to make sure it's displayed properly.

grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -471,7 +471,7 @@
             var data = $("#orderItemForm").serialize();
             data += '&orderIndex=' + $("#orderItemsTable tbody tr").length;
             if (validateItemsForm()) {
-              if ($("#validationCode").val() == 'WARN' && !confirm('${warehouse.message(code: 'orderItem.warningSupplier.label')}')) {
+              if ($("#validationCode").val() == 'WARN' && !confirm('${warehouse.message(code: 'orderItem.warningSupplier.label').replaceAll( /(['])/, '\\\\$1' )}')) {
```
awalkowiak, line 474: I think, that `.replaceAll( /(')/, '\\\\$1' )` will suffice, no need for `[ ]`

awalkowiak (changes requested):

grails-app/views/purchaseOrder/_showOrderItems.gsp
DominikaHatala: @awalkowiak you're right, done.

awalkowiak (approved):

PR #2249, opened 2021-03-04
OBPIH-3680 Add pending shipment items export

awalkowiak:


jmiranda (approved):
@awalkowiak you'll probably need to go back and wrap some column data with StringEscapeUtils.escapeCsv(), but I'll merge this for now and let QA determine which columns. I'm guessing description, product name, and destination.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -495,5 +495,41 @@ class StockMovementController {
...
+            "Product Code" { it.productCode }
+            "Product" { it.productName }
+            "Qty Requested" { it.quantity }
+            "Qty Picked" { it.quantityPicked }
+            "Reason code" { it.reasonCode }
+        })
+        pendingRequisitionItems.each { requisitionItem ->
+            csv << [
+                    shipmentNumber  : requisitionItem?.requisition?.requestNumber,
+                    description     : requisitionItem?.requisition?.description ?: '',
```
jmiranda, line 518: You'll probably need to wrap a few of these columns with StringEscapeUtils.escapeCsv().

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -495,5 +495,41 @@ class StockMovementController {
...
+                    destination     : requisitionItem?.requisition?.destination,
```
jmiranda, line 519: This works because the Location.toString() returns name, but it should be explicit in case we ever change the toString() implementation.
```
requisitionItem?.requisition?.destination?.name
```

PR #2253, opened 2021-03-06
OBGM-141 Added context path handling

jmiranda:
@awalkowiak I had to recreate your PR #1870 since it was not merged before we rebased. Can you make sure it looks alright?


awalkowiak (commented):
@jmiranda everything looks good however, the interceptor is used when there is an API request through `apiClient`, so the urls inserted into the `window.location` won't be parsed by the interceptor. In that case, the `contextPath` should be used in such links too (here's example: https://github.com/openboxes/openboxes/pull/1870/commits/3aca0e0ed9d1b19342f889d36e2ad5038b60ed7b?file-filters%5B%5D=.jsx#diff-d8323bb26d8eaa563ed93c8a7a5dd18d79764f1d0e8528048dc50cf017fedc08R1006). 
There was a TODO for the utility function for adding the `contexPath` to such links (it should look the same as `urlInterceptor`, but should take an URL string as a parameter, instead of an apiClient config).

jmiranda:
@awalkowiak I'm not sure I understand exactly what needs to happen. When you have time can you make any additional changes?

awalkowiak:
@jmiranda sure, I'll do it tomorrow

awalkowiak:
@jmiranda I pushed the missing part

PR #2257, opened 2021-03-09
OBPIH-3636 Allow user to change destination on PO for purchasing loca…

DominikaHatala:
…tions

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
@@ -10,8 +10,10 @@
 package org.pih.warehouse.order
 
 import grails.validation.ValidationException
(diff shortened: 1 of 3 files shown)


grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
```diff
@@ -25,18 +26,23 @@ class PurchaseOrderController {
 
     def create = {
         Order order = new Order()
```
jmiranda, line 28: Let's move the logic to a method on the order service so it's easier to unit test.
```
OrderService.createNewPurchaseOrder(Location currentLocation)
```

jmiranda (changes requested):
LGTM otherwise.

DominikaHatala:
@jmiranda I have updated this PR with required changes.

jmiranda (approved):

PR #2259, opened 2021-03-09
OBPIH-3700 Fix GL account validation

DominikaHatala:


awalkowiak (approved):

jmiranda (commented):
@DominikaHatala I'm confused. How does this fix the issue? And shouldn't the validation happen on the controller / services when we try to save the product?

DominikaHatala:
@jmiranda It was broken after the refactor of edit page. This is frontend validation and showing "Required" error next to the field. This validation is implemented in edit.gsp and while refactoring you moved hidden field "isAccountingRequired" to the new file and it was not working anymore.

jmiranda:
@DominikaHatala Oh because the validation code is in edit.gsp. Weird. So if we didn't do this, then the other solution would have been to use live query to get the attributes from the form. But since the form doesn't actually need these values the proposed solution will work fine.

jmiranda (approved):

jmiranda (changes requested):
But you're rendering the hidden field in the head of the page and it probably needs to be in the body, right?

DominikaHatala:
@jmiranda You're right, it's better when it's in the body although it also works when it's in head.

PR #2261, opened 2021-03-09
OBPIH-3680 Show only picked items on pending shipment items export

awalkowiak:


awalkowiak:
@jmiranda that how I understood the first option from the last comment in the Jira ticket

jmiranda:
@awalkowiak I didn't mean to suggest you did it wrong just that we may have interpreted Kelsey differently. The way you did it was fine and probably more flexible / better.

PR #2264, opened 2021-03-10
OBPIH-3707 Prevent user from creating shipment when only canceled ord…

DominikaHatala:
…er items left

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
@@ -978,7 +978,7 @@ class OrderController {
 
     def createCombinedShipment = {
         def orderInstance = Order.get(params.orderId)
-        if (!orderInstance.orderItems.find {it.quantityRemainingToShip != 0 }) {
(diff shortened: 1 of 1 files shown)


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -978,7 +978,7 @@ class OrderController {
 
     def createCombinedShipment = {
         def orderInstance = Order.get(params.orderId)
-        if (!orderInstance.orderItems.find {it.quantityRemainingToShip != 0 }) {
+        if (!orderInstance.orderItems.find {it.quantityRemainingToShip != 0 && it.orderItemStatusCode != OrderItemStatusCode.CANCELED }) {
```
jmiranda, line 981: @awalkowiak @DominikaHatala Here's another case where it would really good to have the logic in a service or domain method so we can easily add unit tests when the time comes. Please start putting yourself in that frame of mind when you're writing code. 

FWIW this logic could also happen in the createFromOrder() method. If you find that you're about to create an order with no items, throw an exception.

jmiranda (approved):

PR #2265, opened 2021-03-10
OBPIH-3704 Fix updating package on product supplier when place PO or …

DominikaHatala:
…edit items

### DIFF
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -536,6 +536,7 @@ class OrderService {
                     productPrice.price = packagePrice
                     productPackage.productPrice = productPrice
                 }
+                productPackage.lastUpdated = new Date()
             }
             // Associate product package with order item
             orderItem.productPackage = productPackage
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -536,6 +536,7 @@ class OrderService {
                     productPrice.price = packagePrice
                     productPackage.productPrice = productPrice
                 }
+                productPackage.lastUpdated = new Date()
```
awalkowiak, line 539: I think this should be updated automatically by GORM

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
DominikaHatala: @awalkowiak It's not working without this line.

jmiranda: Weird I would have expected it to be updated as well. I guess it's fine but we should investigate.

jmiranda (commented):
Are we addressing the exception that occurs when the order is placed?

DominikaHatala:
@jmiranda I'm not able to reproduce this issue.

PR #2275, opened 2021-03-11
OBPIH-3247 Add expected receipt date to stock movement

awalkowiak:


jmiranda (changes requested):
Receipt and Delivery are not necessarily the same thing. Let's discuss this with Kelsey on the standup. In either case we should be using the same property name throughout. So we'll either need to change all references of `dateDelivered` to `expectedReceiptDate`. Or we'll need to change all references of `expectedReceiptDate` and `dateDelivered` to `expectedDeliveryDate` in order to keep things consistent and be exact.

jmiranda (approved):

PR #2283, opened 2021-03-17
OBPIH-3616 Add invoice list page

DominikaHatala:


grails-app/migrations/views/invoice-list.sql
```diff
@@ -0,0 +1,20 @@
...
+         invoice.id,
+		 invoice.party_from_id,
+         invoice.invoice_number,
+         reference_number.identifier as vendor_invoice_number,
+         unit_of_measure.name as currency,
+         invoice.date_invoiced,
+         person.id as created_by_id,
+         CONCAT(person.first_name, ' ', person.last_name) as created_by_name,
+         count(invoice_item.id) as item_count,
+         count(invoice_item.amount * invoice_item.quantity) as total_value
```
pmuchowski, line 12: I'm not sure what this column is, but I think it should be `sum` instead of `count`

DominikaHatala: yes, it should be sum

grails-app/migrations/views/invoice-list.sql
```diff
@@ -0,0 +1,20 @@
...
+    FROM invoice
+    LEFT JOIN invoice_item ON invoice_item.invoice_id = invoice.id
+    JOIN unit_of_measure ON unit_of_measure.id = invoice.currency_uom_id
```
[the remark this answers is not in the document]
DominikaHatala: all except currency are always set but they are not required so it's better to change them to left joins

jmiranda (approved):

PR #2303, opened 2021-03-25
OBPIH-3637 Refactor validation for stock movement actions and allow S…

DominikaHatala:
…M rollback and edit in central purchasing location

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -2394,5 +2394,29 @@ class StockMovementService {
         }
         return lineItems
     }
+
+    def getDisabledMessage(StockMovement stockMovement, Location currentLocation, Boolean isEditing = false) {
+        def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+
+        boolean isSameOrigin = stockMovement?.origin?.id == currentLocation?.id
+        boolean isSameDestination = stockMovement?.destination?.id == currentLocation?.id
+        boolean isDepot = stockMovement?.origin?.isDepot()
+
+        if ((stockMovement?.hasBeenReceived() || stockMovement?.hasBeenPartiallyReceived()) && isEditing) {
+            return g.message(code: "stockMovement.cantEditReceived.message")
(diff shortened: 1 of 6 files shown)


DominikaHatala:
@jmiranda Just as we discussed yesterday, this is my proposal of solution for moving validation from stockMovement/show.gsp to controllers.

grails-app/controllers/org/pih/warehouse/partialReceiving/PartialReceivingController.groovy
```diff
@@ -1,17 +1,28 @@
...
     def receiptService
+    def stockMovementService
 
     def index = {
         redirect(action: "create")
     }
 
     def create = {
-        Shipment shipment = Shipment.get(params.id)
+        Location currentLocation = Location.get(session.warehouse.id)
```
jmiranda, line 17: Remind me to discuss this during our next tech huddle. This was a bad idea (and I kind of knew it would be) especially for the API since we should try to be stateless whenever possible.

jmiranda (approved):
This is so much better. The getDisabledMessage() doesn't feel quite right but I can't think of a better way off the top of my head.

grails-app/controllers/org/pih/warehouse/partialReceiving/PartialReceivingController.groovy
DominikaHatala: @jmiranda Sure, I will.

awalkowiak (approved):

awalkowiak:
@DominikaHatala approved, but some conflict popped out

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2394,5 +2394,29 @@ class StockMovementService {
         }
         return lineItems
     }
+
+    def getDisabledMessage(StockMovement stockMovement, Location currentLocation, Boolean isEditing = false) {
```
awalkowiak, line 2398: @jmiranda Perhaps `getMessageForDisabled` or `getMessageWhenDisabled` would be better?

DominikaHatala: I think that Justin didn't mean the name of the method but just the way it's done.

PR #2306, opened 2021-03-29
OBPIH-3730 Minor changes to Invoice

DominikaHatala:


grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -44,8 +43,8 @@ class Invoice implements Serializable {
 
     InvoiceType invoiceType
 
-    Party partyFrom // Party generating the invoice
-    Party party // Party responsible for paying
+    Organization partyFrom // Party generating the invoice
```
jmiranda, line 46: Why are we changing this to organization? Eventually I would like to support any party, not just an organization.

jmiranda: Ok, I think I see why. We either need to make name a common property of Party or we need an abstract getter method for name. I would go with the latter if we can.

DominikaHatala: @jmiranda Sure I will change it but can you tell me what do you mean by adding this abstract class? I have already tried but it always returns null value.

jmiranda: On second thought changing to Organization doesn't have any impact on whether we want to allow this to work for Party in the future. We can just change it back to Party when we implement a new subclass and then we can add an abstract method to Party and implement it in all subclasses. But yes, as you mentioned I'm not sure how Grails is going to handle having an abstract method on a persistable class since I should still be able to do 
```
new Party().save()
```

What I was thinking was something like this
```
class Party { 
    public String getFullName();
}
```
```
class Organization extends Party {
    public String getFullName() { 
        return name;
    }
}
```
```
class Organization extends Party {
    public String getFullName() { 
        return "${firstName} ${lastName};
    }
}
```

}

jmiranda (approved):

PR #2312, opened 2021-03-31
OBPIH-3531 Add documents tab on invoice view page

DominikaHatala:


grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -121,6 +122,19 @@ class Invoice implements Serializable {
         return totalValue * (currentExchangeRate ?: 1.0)
     }
 
+    def getDocuments() {
+        def documents = []
+        invoiceItems.each {invoiceItem ->
+            Order order = invoiceItem.orderNumber ? Order.findByOrderNumber(invoiceItem.orderNumber) : null
```
awalkowiak, line 128: When you get invoiceItem.orderNumber you look for orders in shipments and then you return the orderNumber from it. So why not add transient with order in InvoiceItem and pull it once here instead of doing additional findByOrderNumber?

awalkowiak (changes requested):
One thing seems wrong to me. Other than that it looks ok.

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
DominikaHatala: Good point. I will change it.

awalkowiak (approved):

PR #2314, opened 2021-03-31
OBPIH-3042 Fix date format on cycle count report

DominikaHatala:


grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -606,7 +606,7 @@ class ReportController {
                             "Product code"        : StringEscapeUtils.escapeCsv(row?.product?.productCode),
                             "Product name"        : row?.product.name ?: "",
                             "Lot number"          : StringEscapeUtils.escapeCsv(row?.inventoryItem.lotNumber ?: ""),
-                            "Expiration date"     : row?.inventoryItem.expirationDate ? row?.inventoryItem.expirationDate.format("dd-MMM-yyyy") : "",
+                            "Expiration date"     : row?.inventoryItem.expirationDate ? row?.inventoryItem.expirationDate.format("MM/dd/yyyy") : "",
```
jmiranda, line 609: @DominikaHatala Was this requested? 

If so, we should use the constant we've defined for expiration date.
```
Constants.EXPIRATION_DATE_FORMAT
```

A better solution might be to add a configuration property. 
```
// Minimum date for expiration date
openboxes.expirationDate.minValue = new Date("01/01/2000")
openboxes.expirationDate.format = Constants.EXPIRATION_DATE_FORMAT
```
```
def showCycleCountReport = {
    Location location = Location.load(session.warehouse.id)
    String dateFormat = grailsApplication.config.openboxes.expirationDate.format
    ...
        "Expiration date" : row?.inventoryItem.expirationDate ? row?.inventoryItem.expirationDate.format(dateFormat) : "",
```

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
DominikaHatala: @jmiranda Done

awalkowiak (approved):

PR #2315, opened 2021-03-31
OBPIH-3614 Add invoice items modal

awalkowiak:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -61,32 +62,32 @@ class InvoiceService {
...
-                    }
-
-                    if (StringUtils.isNotBlank(orderNumber)) {
-                        eq("orderNumber", orderNumber)
-                    }
-
-                    if (StringUtils.isNotBlank(shipmentNumber)) {
-                        eq("shipmentNumber", shipmentNumber)
-                    }
+            .list() {
```
awalkowiak, line 73: I temporarily removed pagination here because I found few issues with API which dragged me away from making it paginated on the front-end (it can be added later, but the performance seems to be very good)

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -152,6 +153,18 @@ class InvoiceService {
         invoiceItem.delete()
     }
 
+    def addItems(Invoice invoice, List items) {
```
jmiranda, line 156: Let's find a way to do add/update in one service method so we don't need to add an RPC endpoint on the API. I want this to be a clean REST API that we use as the standard to refactor everything else.

Something like this should work.
```
def updateInvoiceItems(Invoice invoice, List items) { 
  items.each { item ->
    InvoiceItem existingInvoiceItem = invoice.invoiceItems.find { it.id == item.id }
    if (existingInvoiceItem) { 
        existingInvoiceItem.quantity = item.quantity
    }
    else { 
      InvoiceCandidate candidateItem = InvoiceCandidate.get(item.id)
      if (candidateItem) {
          InvoiceItem invoiceItem = invoiceItemService.createFromCandidate(candidateItem)
          invoiceItem.quantity = item.quantityToInvoice
          invoice.addToInvoiceItems(invoiceItem)
      }
    }
  }
}
```

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -152,6 +153,18 @@ class InvoiceService {
...
+        items.each { item ->
+            InvoiceCandidate candidateItem = InvoiceCandidate.get(item.id)
```
jmiranda, line 158: It feels like an additional query here seems unnecessary. Why are we doing this when we could be using the POST data?

jmiranda (changes requested):

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -150,12 +153,13 @@ class AddItemsPage extends Component {
    */
   setInvoiceItems(response, startIndex) {
     this.props.showSpinner();
-    const { data } = response.data;
+    const { data, totalCount } = response.data;
```
DominikaHatala, line 156: I'm not sure if this is a good idea to set totalCount each time new lines are loaded. Perhaps we could update totalCount when items in modal are saved?

src/js/components/invoice/InvoiceItemsModal.jsx
```diff
@@ -1,44 +1,323 @@
...
   onSave() {
+    this.props.showSpinner();
+    const { invoiceId } = this.props;
+    const { selectedInvoiceItems } = this.state;
+
+    const payload = {
+      invoiceItems: _.map(selectedInvoiceItems, (item, key) => ({
+        id: key,
+        quantityToInvoice: _.toInteger(item.quantityToInvoice),
+        sortOrder: _.toInteger(item.sortOrder),
```
DominikaHatala, line 158: Is this sortOrder needed and saved anywhere? There is no such field in InvoiceItem.

DominikaHatala (changes requested):

src/js/components/invoice/AddItemsPage.jsx
awalkowiak: This is evaluated in `onResponse` when items in modal are saved.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
awalkowiak: In post we send only candidate id and quantityToInvoice. I could send all data back, but then I will need to find BudgetCode by code, GlAccount by code and uom basing on the formatted string value (`"${quantityUom?.code}/${quantityPerUom as Integer}"` made in transient). I feel the current version is smoother and cleaner

DominikaHatala (approved):

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
jmiranda: Ok sounds good. One issue we might need to deal with is when the candidate invoice item does not exist any longer (i.e. it was added to another invoice in the time between it was queried and saved in our invoice). Right now it seems like we're just silently ignoring that case but that could become confusing. It shouldn't happen very often but it'll be annoying to the user when it does.

jmiranda: So let's throw an IllegalArgumentException or ValidationException for that case.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
jmiranda: Sounds good.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
awalkowiak: Ok, I'll do it as a first thing tomorrow.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -61,32 +62,32 @@ class InvoiceService {
         return invoiceItems
     }
 
-    def getInvoiceCandidates(String id, String orderNumber, String shipmentNumber, String max, String offset) {
+    def getInvoiceCandidates(String id, String orderNumber, String shipmentNumber) {
```
jmiranda, line 65: @awalkowiak @pmuchowski Not sure where this comment should go, but next time we make changes to the invoice feature please change InvoiceCandidate to InvoiceItemCandidate (view, domain, etc). It's been bothering me since the initial PR, but I didn't want to delay the API any longer. But now that the API is working I'd like to refactor this.

Also, do we need an InvoiceService and InvoiceItemService? I'd prefer if we have one service per feature / component.

awalkowiak: Ok. I'll do it.

PR #2316, opened 2021-04-01
OBPIH-3637 Fix issues with receiving and editing

DominikaHatala:


src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -234,9 +234,8 @@ class StockMovement {
 
     Boolean isReceivingAuthorized(Location currentLocation) {
         boolean isSameDestination = destination?.id == currentLocation?.id
-        boolean hasBeenPlaced = hasBeenShipped() || hasBeenPartiallyReceived()
 
-        return !hasBeenReceived() && hasBeenPlaced && (hasBeenIssued() || hasBeenShipped()) && isSameDestination
+        return !hasBeenReceived() && (hasBeenIssued() || hasBeenShipped() || hasBeenPartiallyReceived()) && isSameDestination
```
awalkowiak, line 238: I feel like this is getting out of hand a bit. I think there should be at least few unit tests for checking if receiving is authorized basing on Katarzyna's test cases because I feel like this is going back and forth too much.

awalkowiak (approved):

src/groovy/org/pih/warehouse/api/StockMovement.groovy
DominikaHatala: Yeah, you're probably right. There are too many cases for both receiving and editing.

jmiranda: Agreed. @DominikaHatala Can you add a tech debt ticket for adding unit tests?

PR #2318, opened 2021-04-01
OBPIH-3649 Improve performance of stocklist item list

DominikaHatala:


DominikaHatala:
@jmiranda @awalkowiak That would be great to merge it today before Kasia leaves.

jmiranda (approved):
This looks great. Sorry if I did not get it reviewed / merged in time. One thing we could change in the future would be to add a new read-only domain class to allow us to fetch these through GORM.

PR #2330, opened 2021-04-09
OBPIH-3105 Enable autoincrement format on product code

awalkowiak:
@jmiranda this requires the dateService cleanup done here: https://github.com/openboxes/openboxes/pull/2329 to be merged before this one

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
@@ -31,6 +31,7 @@ class IdentifierService {
     boolean transactional = true
     def grailsApplication
     def dataService
+    def productTypeService
 
     /**
      * A: alphabetic
@@ -193,7 +194,7 @@ class IdentifierService {
 
     def generateSequentialProductIdentifier(ProductType productType) {
         def sequenceFormat = extractSequenceFormat(productType.productIdentifierFormat, Constants.DEFAULT_SEQUENCE_NUMBER_FORMAT_CHAR, 1)
-        def sequenceNumber = productType.getNextSequenceNumber()
+        def sequenceNumber = productTypeService.getAndSetNextSequenceNumber(productType)
 
         return generateSequentialIdentifier(productType.productIdentifierFormat, sequenceFormat, sequenceNumber.toString())
     }
@@ -203,7 +204,7 @@ class IdentifierService {
             throw new IllegalArgumentException("Missing product type or code")
         }
 
-        Integer sequenceNumber = productType.getNextSequenceNumber()
+        Integer sequenceNumber = productTypeService.getAndSetNextSequenceNumber(productType)
         String sequenceNumberStr = generateSequenceNumber(sequenceNumber.toString())
 
         String template = ConfigurationHolder.config.openboxes.identifier.productCode.format
--- a/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
@@ -153,27 +153,32 @@ class IdentifierService {
         return identifier.toUpperCase()
     }
 
-    def generateSequentialProductIdentifier(ProductType productType, String sequentialPatternChar, Integer allowedSequences) {
-        // If custom identifier contains more than one sequential part, then throw exception
+    def extractSequenceFormat(String productIdentifierFormat, String sequentialPatternChar, Integer allowedSequences) {
         Pattern pattern = Pattern.compile("${sequentialPatternChar}+")
-        Matcher matcher = pattern.matcher(productType.productIdentifierFormat)
+        Matcher matcher = pattern.matcher(productIdentifierFormat)
         int count = 0
         def sequenceFormat = ""
+
         while (matcher.find()) {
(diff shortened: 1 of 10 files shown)


grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -150,13 +153,83 @@ class IdentifierService {
...
+        if (!productType.productIdentifierFormat) {
+            return generateProductIdentifier()
+        }
+
+        // if does not contain sequential part, then generate identifier basing on custom format
+        if (!productType.productIdentifierFormat.contains("0")) {
+            return generateIdentifier(productType.productIdentifierFormat)
+        }
+
+        return generateSequentialProductIdentifier(productType, Constants.DEFAULT_SEQUENCE_NUMBER_PATTERN, 1)
```
jmiranda, line 219: Replace the Constants.DEFAULT_SEQUENCE_NUMBER_PATTERN with a configuration property. Each sequential number should have their own pattern added to Config.groovy and the default value should be set to Constants.DEFAULT_SEQUENCE_NUMBER_PATTERN. Actually if the code is getting to this point isn't productType.productIdentifierFormat probably not null? If so I think we should use the pattern in productType.productIdentifierFormat, right?

jmiranda: Sorry DEFAULT_SEQUENCE_NUMBER_PATTERN is different from DEFAULT_SEQUENCE_NUMBER_FORMAT.

jmiranda: So I guess the case I'm thinking about should be handled by the if above this line so I think we're good.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -150,13 +153,83 @@ class IdentifierService {
...
+
+    String generateProductIdentifier(ProductType productType) {
+        // If the product type is null, then generate product code from DEFAULT_PRODUCT_NUMBER_FORMAT
+        if (!productType) {
+            return generateProductIdentifier()
+        }
+
+        // If the product type does not have a custom identifier but has code, then generate sequential product code
+        if (!productType.productIdentifierFormat && productType.code) {
+            Integer sequenceNumber = productType.getNextSequenceNumber()
```
jmiranda, line 197: Does it buy us anything to move this into a separate method? Easier to unit test?

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -150,13 +153,83 @@ class IdentifierService {
         return identifier.toUpperCase()
     }
 
+    def generateSequentialProductIdentifier(ProductType productType, String sequentialPatternChar, Integer allowedSequences) {
```
jmiranda, line 156: Could this be broken down into separate methods or does this need to be done in one unit of work?

validateSequenceNumberFormat() for the top part and generateSequenceNumber() in the bottom. This could probably be used by purchase order number and other sequence numbers.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -150,13 +153,83 @@ class IdentifierService {
...
+        def sequenceFormat = ""
+        while (matcher.find()) {
+            sequenceFormat = matcher.group()
+            count++
+        }
+        if (count > allowedSequences) {
+            throw new IllegalArgumentException("Cannot have more sequence numbers than ${allowedSequences} in the same identifier")
+        }
+
+        def sequenceNumber = productType.getNextSequenceNumber()
```
jmiranda, line 170: At what point do we save this to the database? And how do we ensure that two threads don't try to save the same identifier. Again, having this fail on a write operation (i.e. unique constraint on product.productCode) rather than on the read operation is fine with me for now but it might be good to have another solution in mind for maybe locking on read at some point. I don't know how other systems handle this so maybe we can create a spike to look into it.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
awalkowiak: I'll rename the const to DEFAULT_SEQUENCE_NUMBER_FORMAT_CHAR. The productType is passed, because the next sequence number is pulled from it

grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
```diff
@@ -0,0 +1,20 @@
...
+ * which can be found in the file epl-v10.html at the root of this distribution.
+ * By using this software in any fashion, you are agreeing to be bound by
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.product
+
+class ProductTypeService {
+
+    synchronized getAndSetNextSequenceNumber(ProductType productType) {
```
awalkowiak, line 14: @jmiranda if this won't work, perhaps doing ProductType.lock(productTypeId) would be enough?

jmiranda (approved):

PR #2339, opened 2021-04-15
OBPIH-3617 Add columns to payment terms

DominikaHatala:


grails-app/controllers/org/pih/warehouse/core/PaymentTermController.groovy
```diff
@@ -11,5 +11,58 @@ package org.pih.warehouse.core
...
+    def edit = {
+        def paymentTerm = PaymentTerm.get(params.id)
+        return [paymentTerm: paymentTerm]
+    }
+
+    def save = {
+        def paymentTerm = PaymentTerm.get(params.id)
+        if (paymentTerm) {
+            paymentTerm.properties = params
+            if (!paymentTerm.hasErrors() && paymentTerm.save(flush: true)) {
```
awalkowiak, line 34: hasErrors check is not required

grails-app/migrations/0.8.x/changelog-2021-04-15-1340-insert-payment-term.xml
```diff
@@ -0,0 +1,95 @@
...
+        </preConditions>
+        <insert tableName="payment_term">
+            <column name="id" value="1"/>
+            <column name="version" valueNumeric="0"/>
+            <column name="code" value="PP100"/>
+            <column name="name" value="Prepayment 100"/>
+            <column name="description"/>
+            <column name="prepayment_percent" value="100"/>
+            <column name="days_to_payment"/>
+            <column name="date_created" valueDate="2021-04-15T00:00:00.0"/>
```
awalkowiak, line 19: I think you could declare:
`<property name="now" value="now()" dbms="mysql, postgresql"/>`
and then use `valueDate="${now}"` in each timestamp field's value, unless you really want to have this particular date here

awalkowiak (changes requested):

jmiranda (changes requested):

grails-app/migrations/0.8.x/changelog-2021-04-15-1340-insert-payment-term.xml
jmiranda: Ohh I like that suggestion.

grails-app/controllers/org/pih/warehouse/core/PaymentTermController.groovy
jmiranda: The hasErrors check is fine (Grails generates the controllers this way). The point I was making with your PR was that the validate() shouldn't have been needed because it happens during the binding above.

grails-app/migrations/0.8.x/changelog-2021-04-15-1340-insert-payment-term.xml
jmiranda: Where is that "now" property defined? Let's just support mysql for ... now. I don't want people looking at this code and thinking we support postgres until we actually fully support postgres.

awalkowiak: Properties are defined before changesets.

awalkowiak: @jmiranda seems like 1.9 version of liquibase does not support specifying dbms on properties yet. We'd need to upgrade liquibase to get this working in the future.

awalkowiak (approved):

DominikaHatala:
@jmiranda I have updated PR with all required changes.

jmiranda (approved):

PR #2346, opened 2021-04-19
OBPIH-3761 Add debounced organizations fetch to invoice create page

DominikaHatala:


grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -82,6 +81,9 @@ class OrganizationService {
                     ilike("description", "${params.q}%")
                 }
             }
+            if (params.name) {
+                ilike("name", "%${params.name}%")
```
awalkowiak, line 85: Can't you do it using the above part for it (pass the name as a `q`)?

awalkowiak (commented):
One question. Other than that it looks good to me.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
DominikaHatala: Sure I could but I think that in case of getting organizations for dropdown I think that we should only get those with matching name and not code or description.

awalkowiak: @jmiranda How do you see it?

jmiranda: @DominikaHatala Where is params.name coming from? I only see params.q in the form?

jmiranda: Nevermind, just realized this was probably being sent as a parameter from a React component. I think we probably want it to work the same as the search form on the Org list page.

jmiranda: So yeah let's pass in params.q and then if we get specific requirements from kelsey about how name or description should match then we can implement those separately.

jmiranda (changes requested):

awalkowiak (approved):

PR #2347, opened 2021-04-19
OBPIH-3756 Usability changes to add items modal

awalkowiak:


DominikaHatala (approved):

jmiranda (commented):
I'm slightly confused by this but don't know the requirements well enough to provide a better solution. For example, while I understand why you created the getOrderAndShipmentNumbers API, I think it would make just as much sense to have separate APIs for the two /api/invoice/$id/orders and /api/invoice/$id/shipments. In fact, I think it would be better since those two APIs might be applicable in other use cases. However, since this was probably done with performance in mind I'll bite my tongue. But if we ever refactor this please note that I'd be more pleased with separate reusable APIs.

PR #2349, opened 2021-04-20
OBPIH-3695 Update order item and adjustments after invoice posted

DominikaHatala:


grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -56,4 +57,8 @@ class OrderAdjustment implements Serializable {
     def getTotalAdjustments() {
         return amount ?: percentage ? orderItem ? orderItem?.subtotal * (percentage/100) : order.subtotal * (percentage/100) : 0
     }
+
+    def getIsInvoiced() {
+        return InvoiceItem.list().find { it.orderAdjustment == this && it.invoice.dateSubmitted } ? true : false
```
jmiranda, line 62: Write criteria query (InvoiceItem.createCriteria()) or HQL (InvoiceItem.executeQuery()) here as this will become a bottleneck with enough invoice items in the system. In fact, this should just be a query that returns true or false.

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -286,6 +290,20 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         return sortOrder
     }
 
+    def getSubmittedInvoiceItems() {
+        return InvoiceItem.list().findAll { it.orderItem == this && it.invoice.dateSubmitted }
```
jmiranda, line 294: Same as above.

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -286,6 +290,20 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
...
+    }
+
+    Integer getQuantityInvoicedInStandardUom() {
+        return submittedInvoiceItems.sum { InvoiceItem invoiceItem ->
+            invoiceItem?.quantity
```
jmiranda, line 299: Same as above. This would be the one place where I wouldn't mind leaving this as-is, but this can definitely be done in a single query.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
awalkowiak, line 294: Is there a better way to achieve this?

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak, line 62: Is there a better way to achieve this?

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
awalkowiak: Nvm, I just noticed @jmiranda comments.

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak: Nvm, I just noticed @jmiranda comments.

DominikaHatala: I wanted to do it but orderAdjustment and orderItem are derived fields on Invoice Item. Can createCriteria or executeQuery handle this?

jmiranda: At the very least you should do 
```
Boolean getIsInvoiced() { 
  return InvoiceItem.createCriteria().list { 
      invoice {
          notNull("dateSubmitted")
      }
  }.find { it.orderAdjustment == this }
}
```

You might not be able to reference the derived property in the criteria query but you should be able to achieve this using a join in HQL. Start by writing SQL to answer this question for a single order adjustment and then we can help you convert that into a HQL or criteria query. 

In general, we should always write SQL if something won't work with our current associations.

jmiranda:
```
select * 
from invoice_item 
join invoice on invoice_item.invoice_id = invoice.id 
join order_adjustment_invoice on order_adjustment_invoice.invoice_item_id = invoice_item.id 
where invoice.date_submitted is not null;
```

jmiranda: You should probably add a join parameter for the order adjustment and then pass the order adjustment ID as a parameter.

jmiranda: If possible turn this into an HQL query. If necessary we could add an association.

DominikaHatala:
@jmiranda @awalkowiak I have added commit with required changes.

awalkowiak (approved):

jmiranda (approved):

PR #2354, opened 2021-04-23
OBPIH-3777 Refactor order search field to include order name

DominikaHatala:


jmiranda (approved):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -44,15 +44,13 @@ class OrderService {
     def getOrders(Order orderTemplate, Date dateOrderedFrom, Date dateOrderedTo, Map params) {
         def orders = Order.createCriteria().list(params) {
             and {
-                if (orderTemplate.name || orderTemplate.description) {
```
awalkowiak, line 47: @DominikaHatala What if `q` will be empty, but `orderTemplate` will have either `name` or `description`. Seems like it is still possible to pass name and description with the `orderTemplate` here: https://github.com/openboxes/openboxes/blob/develop/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy#L68.
cc @jmiranda

DominikaHatala: This is the only place where gerOrders() is used and after the change I've applied in this ticket it is no possible to pass name and description with the orderTemplate.

awalkowiak (approved):

jmiranda (approved):

PR #2355, opened 2021-04-23
OBPIH-3805 Adjust destination filter for PO and fix styling on putawa…

DominikaHatala:
…ys list

### DIFF
--- a/grails-app/views/order/_filters.gsp
+++ b/grails-app/views/order/_filters.gsp
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.core.ActivityCode" %>
+<%@ page import="org.pih.warehouse.core.ActivityCode; org.pih.warehouse.order.OrderTypeCode" %>
 <div class="box">
     <h2><warehouse:message code="default.filters.label"/></h2>
     <g:form id="listForm" action="list" method="GET">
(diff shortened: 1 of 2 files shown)


grails-app/views/order/_filters.gsp
```diff
@@ -41,26 +41,40 @@
...
+            </g:if>
+            <g:elseif test="${params.orderTypeCode == OrderTypeCode.TRANSFER_ORDER}">
+                <div class="filter-list-item">
+                    <label><warehouse:message code="order.destination.label"/></label>
+                    <g:selectLocationViaAjax id="destination"
+                                             name="destination"
+                                             class="ajaxSelect2"
+                                             noSelection="['':'']"
+                                             value="${params.destination}"
+                                             data-ajax--url="${request.contextPath }/json/findLocations?activityCode=${org.pih.warehouse.core.ActivityCode.PLACE_ORDER}"
```
jmiranda, line 74: Why PLACE_ORDER for transfer order?

grails-app/views/order/list.gsp
```diff
@@ -153,7 +153,8 @@
 							</tbody>
 							<tfoot>
 								<tr class="odd">
-									<th colspan="12"></th>
+									<g:set var="colspan" value="${params.orderTypeCode == OrderTypeCode.PURCHASE_ORDER ? 12 : 8}"/>
```
jmiranda, line 156: It's things like this that make me want to rethink how we do the UI for order. Either separate into separate directories

transferOrder/list
purchaseOrder/list

Or make it a wholesale change by adding PurchaseOrder extends Order, etc.

jmiranda: Not for this ticket but should discuss and ticket our approach in a tech huddle.

awalkowiak: @jmiranda I think it would be better to have a more generic Order domain and a separate PurchaseOrder extending it.

awalkowiak (commented):
Other than what @jmiranda brought up, it LGTM

grails-app/views/order/_filters.gsp
DominikaHatala: It used to work like that before for both purchase order and transfer order(putaway)

grails-app/views/order/list.gsp
DominikaHatala: Yes, I agree. That is something that we should discuss during tech huddle.

grails-app/views/order/_filters.gsp
DominikaHatala: @jmiranda Do you think that it should be changed and not filtered by any activity code?

jmiranda: > It used to work like that before for both purchase order and transfer order(putaway)

Oh right. Sorry about that. 

> Do you think that it should be changed and not filtered by any activity code?

Good question. I'm not sure. We can leave it as-is for now since no one has complained.

grails-app/views/order/list.gsp
jmiranda: Cool, I've added that to the tech huddle notes so we can bring it up on Monday if we have time or add another tech huddle to discuss.

jmiranda (approved):

PR #2356, opened 2021-04-23
OBPIH-3713 Add order state view

awalkowiak:


grails-app/migrations/views/order-state-view.sql
```diff
@@ -0,0 +1,160 @@
...
+            order_item.id               AS order_item_id,
+            product.product_code        AS product_code,
+            SUM(order_item.quantity)    AS quantity_ordered,
+            SUM(shipment_item.quantity) AS quantity_shipped
+        FROM `order`
+            LEFT OUTER JOIN order_item ON order.id = order_item.order_id
+            LEFT OUTER JOIN product ON order_item.product_id = product.id
+            LEFT OUTER JOIN order_shipment ON order_item.id = order_shipment.order_item_id
+            LEFT OUTER JOIN shipment_item ON shipment_item.id = order_shipment.shipment_item_id
+        WHERE `order`.order_type_code = 'PURCHASE_ORDER' AND `order`.status NOT IN ('COMPLETED', 'CANCELLED') AND order_item.order_item_status_code != 'CANCELLED'
```
jmiranda, line 23: Can you separate these three conditions onto separate lines so they are more readable? And I think we'd like to keep orders which are in those two states, no?

jmiranda: In other words, I would like one row for every PO in the database. Therefore, Canceled and Completed are valid states.

jmiranda (commented):
I ran the query locally on OBNAV data. This is very promising. I'll continue to review tomorrow.

grails-app/migrations/views/order-state-view.sql
awalkowiak: Done

jmiranda (approved):

PR #2362, opened 2021-04-26
OBGM-141 Small improvement for empty and '/' context path

awalkowiak:


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -58,19 +58,23 @@ class StockMovementController {
     }
 
     def createOutbound() {
-        render(template: "/common/react", model: [contextPath: grailsApplication.config.server.contextPath], params: params)
+        def contextPath = grailsApplication.config.server.contextPath
+        render(template: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''], params: params)
```
jmiranda, line 62: I'd like to move this to a utility like ConfigHelper if possible. I can do that myself.
```
render(template: "/common/react", model: [contextPath: ConfigHelper.contextPath], params: params)
```
and then in ConfigHelper add the following method
```
static String getContextPath() { 
   String contextPath = getContextPathGrails3
   return contextPath: contextPath != '/' ? contextPath : ''
}
```
Also do you think I should be able to reference the ConfigHelper.contextPath in the grails-template.html instead of passing it along in the model. Or is that not really possible since we're compiling the template using webpack?

jmiranda: As an example for what I meant by `getContextPathGrails3`, this is how I would probably do it in Grails 1.3.9. I just need to do some more investigation of the right approach for Grails 3.
```
class ConfigHelper {
    static getContextPath() {
        def contextPath = ConfigurationHolder.config.server.contextPath
        return contextPath != '/' ? contextPath : ''
    }
    ...
}
```

jmiranda (approved):

PR #2364, opened 2021-04-26
OBPIH-3745 Changes to receipt items email

DominikaHatala:


grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -172,7 +172,11 @@ class NotificationService {
         recipientItems.each { Person recipient, items ->
             if (emailValidator.isValid(recipient?.email)) {
                 def subject = g.message(code: "email.yourItemReceived.message", args: [shipment.destination.name, shipment.shipmentNumber])
-                def body = "${g.render(template: "/email/shipmentItemReceived", model: [shipmentInstance: shipment, receiptItems: items, recipient: recipient])}"
+                items.each {
+                    def quacan = it.quantityToCancel
+                    def co = quacan
```
jmiranda, line 177: Can you explan these two lines?

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -69,11 +69,11 @@ class ReceiptService {
      * @param shipment
      * @return
      */
-    PartialReceipt getPartialReceiptFromShipment(Shipment shipment) {
+    PartialReceipt getPartialReceiptFromShipment(Shipment shipment, User user) {
 
         PartialReceipt partialReceipt = new PartialReceipt()
         partialReceipt.shipment = shipment
-        partialReceipt.recipient = shipment.recipient
+        partialReceipt.recipient = user
```
jmiranda, line 76: Can you explain why we're doing this?

jmiranda: Also, there's a better way to get the current user without passing it down through a bunch of methods. 
```
def currentUser = AuthService.currentUser.get()
```

But I'd like to know why we're passing the user into these methods before I recommend this approach.

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
DominikaHatala: That's a great question! I had some issued with calculating quantity canceled by user in receipt before I realized that the only option to cancel is to cancel quantity remaining. So this part was only for me to debug.

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
DominikaHatala: Recipient on receipt should be the person who created partial receipt so the current user. I don't know why we used to set shipment.recipient because this field is null and in database all receipts do not have recipient.

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -108,6 +108,7 @@ class ReceiptService {
         partialReceipt.recipient = receipt.recipient
         partialReceipt.dateShipped = receipt?.shipment?.actualShippingDate
         partialReceipt.dateDelivered = receipt.actualDeliveryDate
+        partialReceipt.recipient = receipt.recipient
```
awalkowiak, line 111: It's already there (line 108)

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
jmiranda: Ok, sounds good.

jmiranda (approved):
@awalkowiak once that small bug is fixed go ahead and merge

awalkowiak (approved):

PR #2374, opened 2021-04-28
OBPIH-3796 Generate prepayment invoice

awalkowiak:


grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -144,6 +144,10 @@ class InvoiceItem implements Serializable {
 
     // Total shipment item value
     def getTotalItemPrice() {
+        if (invoice.invoiceType?.code == InvoiceTypeCode.PREPAYMENT_INVOICE) {
```
jmiranda, line 147: Let's add a convenience method here as I suspect this will be a commonly repeated piece of code.

Invoice 
```
boolean isPrepaymentInvoice() { 
    return invoice.invoiceType?.code == InvoiceTypeCode.PREPAYMENT_INVOICE
}
```

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -58,6 +58,17 @@ class OrderAdjustment implements Serializable {
         return amount ?: percentage ? orderItem ? orderItem?.subtotal * (percentage/100) : order.subtotal * (percentage/100) : 0
     }
 
+    def getInvoiceItem() {
```
jmiranda, line 61: Couldn't there be multiple invoice items for a single order adjustment?

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak: I think that technically it can have only one invoice item. Because it has quantity = 1. And there won't be invoice item candidate if order adjustment already have invoice item associated.
https://github.com/openboxes/openboxes/blob/develop/grails-app/migrations/views/invoice-candidate-view.sql#L26

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
awalkowiak: done

jmiranda (approved):

PR #2376, opened 2021-04-29
OBPIH-3785 Allow user to import items to placed order

DominikaHatala:


DominikaHatala:
@jmiranda Please take a look at this proposed solution if you have time. It's not yet finished but I just wanted to show you how I started it.

jmiranda (commented):
LGTM. I think you're on the right track. Keep going down this path just try to propose a better name for the field mapping config property.

DominikaHatala:
@jmiranda I've applied your suggestions.

pmuchowski (approved):

PR #2389, opened 2021-05-07
OBPIH-3813 Filter invoice candidates by buyer organization

DominikaHatala:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -115,10 +125,48 @@ class InvoiceService {
...
+                order {
+                    eq("destinationParty", currentLocation.organization)
+                }
             }
 
-        return invoiceItemCandidates
+        return shipmentNumbers
+    }
+
+    def getOrderNumbers(String id) {
```
awalkowiak, line 140: Is this required? The Shipment Number and Order Number are already on the InvoiceItemCandidate. Why you can't use the previous more generic way of getting order or shipment number by passing the appropriate field to it? And instead nesting orderNumber or shipmentNumber in order or shipment you could have simple `ne(distinctField, "")` with `groupProperty(distinctField)`. You could add to the previous version this part with the destination party.

I think this was meant to split only the endpoints this service method was fine I guess (unless I missed something).

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
DominikaHatala: @awalkowiak I've changed it.

awalkowiak (approved):

PR #2391, opened 2021-05-10
OBPIH-3800 Hide prepayment button

DominikaHatala:


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -90,7 +90,7 @@ class Order implements Serializable {
             "received",
             "canceled",
             "completed",
-
+            "hasInvoice"
```
awalkowiak, line 93: Move under `"totalNormalized"`

awalkowiak (changes requested):

PR #2392, opened 2021-05-10
OBPIH-3819 Add invoice tab to order page

awalkowiak:


src/groovy/org/pih/warehouse/invoice/InvoiceStatus.groovy
```diff
@@ -0,0 +1,32 @@
...
+ * By using this software in any fashion, you are agreeing to be bound by
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.invoice
+
+enum InvoiceStatus {
+
+    PENDING(10),
+    SUBMITTED(20),
```
DominikaHatala, line 15: I think it should be called POSTED instead of SUBMITTED.

DominikaHatala (changes requested):

src/groovy/org/pih/warehouse/invoice/InvoiceStatus.groovy
awalkowiak: done

DominikaHatala (approved):

pmuchowski (approved):

PR #2395, opened 2021-05-11
OBPIH-3837, OBPIH-3843 Added fields and filters on invoice list page

awalkowiak:


grails-app/domain/org/pih/warehouse/invoice/InvoiceList.groovy
```diff
@@ -0,0 +1,38 @@
...
+class InvoiceList {
+    String id
+    Invoice invoice
+
+    String invoiceNumber
+    String vendorInvoiceNumber
+
+    String partyFromId
+    String partyId
+    String vendor // = party.code + " " + part.name
```
jmiranda, line 21: Why not just have vendorCode and vendorName?

jmiranda: ... and let the UI handle the concatentation.

grails-app/migrations/views/invoice-list.sql
```diff
@@ -1,20 +1,27 @@
 CREATE OR REPLACE VIEW invoice_list AS (
    	SELECT
-         invoice.id,
-		 invoice.party_from_id,
+         invoice.id as id,
```
jmiranda, line 3: This will break the PK if there's more than one VENDOR_INVOICE_NUMBER reference number, so let's remove this column and use a composite key.

jmiranda (changes requested):

awalkowiak:
@jmiranda done

jmiranda (approved):

PR #2403, opened 2021-05-14
OBPIH-3797 Get prepaid invoice items and update total value of invoice

DominikaHatala:


awalkowiak (commented):
I think that the prepayment item's total value should be returned as negative. In the final invoice, it should be displayed as negative (according to the example from the ticket's description).

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -145,6 +146,10 @@ class Invoice implements Serializable {
         return InvoiceStatus.PENDING
     }
 
+    Float getTotalPrepaidValue() {
```
jmiranda, line 149: Could you explain this? I was expecting something like the following. In my mind if there's a prepayment invoice, that includes all items and adjustments from the order at the prepayment rate. You can't have a prepayment invoice item that is not part of a prepayment invoice.
```
BigDecimal getTotalPrepaidValue() {
   return (invoiceType?.code == InvoiceTypeCode.PREPAYMENT_INVOICE) ? invoiceItems.sum { it.totalAmount } : 0
}
```

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -69,7 +69,7 @@ class Invoice implements Serializable {
         invoiceItems cascade: "all-delete-orphan"
     }
 
-    static transients = ['vendorInvoiceNumber', 'totalValue', 'totalValueNormalized', 'documents', 'status']
+    static transients = ['vendorInvoiceNumber', 'totalValue', 'totalValueNormalized', 'documents', 'status', 'totalPrepaidValue']
```
jmiranda, line 72: just to keep the language consistent, please call this totalPrepaymentValue

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -171,6 +179,32 @@ class InvoiceItem implements Serializable {
         return invoice.invoiceType?.code == InvoiceTypeCode.PREPAYMENT_INVOICE
     }
 
+    def getPrepaidItem() {
```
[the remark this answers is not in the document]
DominikaHatala: We need to get prepaid item for each final invoice item to display it in the last page of invoice items.

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
DominikaHatala: That's right. But in this case we want to get the value of prepayment for the final invoice.

DominikaHatala:
@awalkowiak @jmiranda I have updated this PR to include adding lines to the last page of final invoice. Now when it's finished I just thought of a better solution to this one. We could just add a flag to final invoice if prepayment invoice already exists for the order and then all prepayment items are just all order items multiplied by prepayment term percentage. I think that this is something that we could discuss and potentially refactor.

src/js/components/invoice/ConfirmInvoicePage.jsx
```diff
@@ -187,6 +187,9 @@ class ConfirmInvoicePage extends Component {
       if (!_.isNull(startIndex) &&
           this.state.values.invoiceItems.length !== this.state.values.totalCount) {
         this.loadMoreRows({ startIndex: startIndex + this.props.pageSize });
+      } else if (this.state.values.invoiceItems.length === this.state.values.totalCount
```
awalkowiak, line 190: Why are you checking `this.state.values.invoiceItems.length === this.state.values.totalCount`? What does this indicate?

DominikaHatala: When all regular invoice items are loaded then we fetch prepaid items and add to the end of the list of invoice items.

src/js/components/invoice/ConfirmInvoicePage.jsx
```diff
@@ -226,6 +229,34 @@ class ConfirmInvoicePage extends Component {
...
+  fetchPrepaymentItems() {
+    const url = `/openboxes/api/invoices/${this.state.values.id}/prepaymentItems`;
+    apiClient.get(url)
+      .then((response) => {
+        const { data, totalCount } = response.data;
+        const lineItemsData = _.map(
+          data,
+          val => ({
+            ...val,
+            totalAmount: val.totalAmount * (-1),
```
awalkowiak, line 241: @DominikaHatala I think it would be better to have these prapayment items with value `* -1` available from the invoice domain level because this probably will be included in the invoice view page too.
cc @jmiranda

src/js/components/invoice/ConfirmInvoicePage.jsx
awalkowiak: Makes sense

src/js/components/invoice/ConfirmInvoicePage.jsx
awalkowiak: Another thing I just realized. What if the line is a `DISCOUNT` order adjustment with an already negative value?

jmiranda: Agreed on the first comment. I think the second comment requires Kelsey to respond.

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -145,6 +146,14 @@ class Invoice implements Serializable {
         return InvoiceStatus.PENDING
     }
 
+    boolean getIsPrepaymentInvoice() {
+        return invoiceType?.code == InvoiceTypeCode.PREPAYMENT_INVOICE
+    }
+
+    Float getTotalPrepaymentValue() {
+        return !isPrepaymentInvoice && invoiceItems.any { it.prepaymentItem } ? invoiceItems?.collect { it?.prepaymentItem?.totalAmount }?.sum() : 0
```
jmiranda, line 154: I feel like I asked this one already, but don't see it anywhere. Why are we doing it this way? Why not just calculate this 
```
return (isPrepaymentInvoice) ? invoiceItems.sum { it.totalAmount } : 0
```
That way we remove the need for the N+1 query invoked by getPrepaymentItem().

jmiranda (changes requested):

DominikaHatala:
@jmiranda @awalkowiak I have applied all requested changes and also rebased and tested it after Artur's changes. Everything works as it was supposed to.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -159,6 +165,10 @@ class InvoiceItem implements Serializable {
         return totalAdjustments ?: totalItemPrice
     }
 
+    def getTotalPrepaymentAmount() {
+        return totalAmount * (-1)
```
awalkowiak, line 169: `return isPrepaymentInvoice ? totalAmount * (-1) : 0.0`

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
DominikaHatala: Done

awalkowiak (approved):

PR #2404, opened 2021-05-14
OBPIH-3840 Added data model changes to handle many-to-many associatio…

jmiranda:
…n between invoice item and order item

### DIFF
--- a/grails-app/migrations/0.8.x/changelog-2021-05-14-1654-create-table-order-invoice.xml
+++ b/grails-app/migrations/0.8.x/changelog-2021-05-14-1654-create-table-order-invoice.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="jmiranda (generated)" id="1621028114887-1">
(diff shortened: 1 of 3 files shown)


jmiranda:
@awalkowiak @pmuchowski @DominikaHatala Hey guys after wrestling this for days I came to the conclusion that I was probably overthinking this one a bit. So I added the most basic association since it doesn't seem we need anything more complicated than that. Please let me know if you can think of use cases that I'm neglecting.

DominikaHatala (approved):

awalkowiak (approved):

grails-app/migrations/0.8.x/changelog-2021-05-14-1654-create-table-order-invoice.xml
```diff
@@ -0,0 +1,39 @@
...
+      </not>
+    </preConditions>
+    <addForeignKeyConstraint baseColumnNames="invoice_item_id" baseTableName="order_invoice"
+      constraintName="FK6A1A433CB95ED8E0" deferrable="false" initiallyDeferred="false"
+      referencedColumnNames="id" referencedTableName="invoice_item"/>
+  </changeSet>
+  <changeSet author="jmiranda (generated)" id="1621028114887-3">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <foreignKeyConstraintExists foreignKeyName="FK6A1A433CB95ED8E0"/>
```
awalkowiak, line 31: @jmiranda I think, that this should be FK6A1A433C3BE9D843, looks like FK6A1A433CB95ED8E0 is from the previous changeset

awalkowiak (changes requested):

grails-app/migrations/0.8.x/changelog-2021-05-14-1654-create-table-order-invoice.xml
jmiranda: Thanks

PR #2409, opened 2021-05-17
OBPIH-3824 Improvements to combine shipment worfklow

DominikaHatala:


grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -97,6 +97,11 @@ class ShipmentItem implements Comparable, Serializable {
         return orderItemIds ? orderItemIds.first() : null
     }
 
+    String getOrderId() {
+        def orderNumbers = orderItems?.collect { OrderItem orderItem -> orderItem.order.id }?.unique()
```
awalkowiak, line 101: I know we did that before, but perhaps it would be better to pull an associated order? Could be refactored in near future?

awalkowiak (commented):
LGTM, beside one potential refactor (for future?)

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
awalkowiak: cc @jmiranda

jmiranda: This could return multiple order IDs, so you'd have to retrieve first/last or allow returning a list.

jmiranda: Ah nevermind didn't see the next line. Yeah returning the order would probably be better but I'd have to see the context in which this method is used.

PR #2411, opened 2021-05-18
OBPIH-3824 Fix setting date invoiced on prepayment invoices

DominikaHatala:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -254,7 +255,7 @@ class InvoiceService {
         invoice.description = order.description
         invoice.partyFrom = order.destinationParty
         invoice.party = order.origin.organization
-        invoice.dateInvoiced = new Date()
+        invoice.dateInvoiced = LocalDate.now().toDate()
```
awalkowiak, line 258: What is wrong with the previous version?

DominikaHatala: It set the date with time, filter is searching for the date with no time set so it didn't match.

jmiranda: I think we're ok with keeping the time.

jmiranda (commented):
LGTM. Unless the requirements specify that invoiced date should only include date, then we should allow the time component.

PR #2412, opened 2021-05-18
OBPIH-3799, OBPIH-3840 Generate Invoice for PO after prepayment

awalkowiak:


grails-app/conf/UrlMappings.groovy
```diff
@@ -256,9 +256,9 @@ class UrlMappings {
         }
 
         // Invoice API
-        "/api/invoices/$id/invoiceItems"(parseRequest: true) {
+        "/api/invoices/$id/items"(parseRequest: true) {
```
jmiranda, line 259: Interesting chose. Can/should we apply this to other APIs in the near future?

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
```diff
@@ -95,4 +95,15 @@ class InvoiceController {
         Invoice invoice = invoiceService.generatePrepaymentInvoice(order)
         redirect(action: "create", params: [id: invoice.id])
     }
+
+    def generateInvoiceAfterPrepayment = {
```
jmiranda, line 99: I think this can just be generateInvoice().

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -69,7 +69,7 @@ class Invoice implements Serializable {
         invoiceItems cascade: "all-delete-orphan"
     }
 
-    static transients = ['vendorInvoiceNumber', 'totalValue', 'totalValueNormalized', 'documents', 'status']
+    static transients = ['vendorInvoiceNumber', 'totalValue', 'totalValueNormalized', 'documents', 'status', 'hasPrepaymentAssociated']
```
jmiranda, line 72: Change to `hasPrepaymentInvoice` to keep it consistent.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -97,35 +97,34 @@ class InvoiceItem implements Serializable {
         createdBy(nullable: true)
     }
 
-    ShipmentItem getShipmentItem() {
-        return shipmentItems ? shipmentItems?.iterator()?.next() : null
+    OrderItem getOrderItem() {
+        return orderItems ? orderItems?.find { it } : null
```
jmiranda, line 101: Why the switch? I can see why it might help (choose the first non-null item) but do we think it'll ever be null?

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -97,35 +97,34 @@ class InvoiceItem implements Serializable {
...
     }
 
-    String getDescription() {
-        return orderAdjustment ? orderAdjustment.description : product?.name
-    }
-
-    OrderItem getOrderItem() {
-        if (orderAdjustments) {
-            return orderAdjustment?.orderItem
+    Order getOrder() {
```
jmiranda, line 116: This method scares me the most. I think I'd prefer to leave it up to the developer unless it causes us to write this branching logic whenever we need to access Order. 

If we do need it, perhaps since a shipment might include multiple orders we always return a list of Orders here to make sure our Invoice model exposes the right method signature. It would be bad if the system supported adding multiple orders per shipment / invoice, but the model itself did not.
```
List<Order> getOrders() { 
    return orderItem ? [orderItem.order] : orderAdjustment ? [orderAdjustment.order] : shipmentItem.orderItems*.order
}
```

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -175,8 +174,8 @@ class InvoiceItem implements Serializable {
     Map toJson() {
         return [
                 id: id,
-                orderNumber: orderNumber,
-                shipmentNumber: shipmentNumber,
+                orderNumber: order?.orderNumber ?: '',
```
jmiranda, line 177: Why empty string here and not null?

grails-app/migrations/views/invoice-candidate-view.sql
```diff
@@ -52,5 +52,6 @@ union
           join order_shipment on order_shipment.shipment_item_id = shipment_item.id
           join order_item on order_shipment.order_item_id = order_item.id
           join `order` on order_item.order_id = `order`.id
+          left join order_invoice on order_invoice.order_item_id = order_item.id
           left join shipment_invoice on shipment_invoice.shipment_item_id = shipment_item.id
- where shipment_invoice.invoice_item_id is null and shipment.current_status = 'SHIPPED');
+ where order_invoice.invoice_item_id is null and shipment_invoice.invoice_item_id is null and shipment.current_status = 'SHIPPED');
```
jmiranda, line 57: Each condition of the where clause should be on a separate line for readability.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -215,6 +215,73 @@ class InvoiceService {
...
+    Invoice generatePrepaymentInvoice(Order order) {
+        if (order.orderItems.any { it.hasInvoices } || order.orderAdjustments.any { it.hasInvoice }) {
+            throw new Exception("Some order items or order adjustments for this order already have been invoiced")
+        }
+
+        Invoice invoice = createFromOrder(order)
+        invoice.invoiceType = InvoiceType.findByCode(InvoiceTypeCode.PREPAYMENT_INVOICE)
+        createOrUpdateVendorInvoiceNumber(invoice, order.orderNumber + Constants.PREPAYMENT_INVOICE_SUFFIX)
+
+        if (order.orderItems.size() + order.orderAdjustments.size() == 0) {
```
jmiranda, line 232: Careful here. I think Hibernate/Grails will initialize the orderItems and orderAdjustments with an empty list, but for unit tests it might not so you'll need to add null safety. 

Plus I would suggest using something more readable like this.
```
if (!order?.orderItems?.empty && !order?.orderAdjustments?.empty)
```

grails-app/views/order/_summary.gsp
```diff
@@ -171,14 +171,22 @@
                                 ${warehouse.message(code: 'default.button.rollback.label')}
                             </g:link>
                             <g:if test="${!orderInstance?.hasInvoice && orderInstance?.paymentTerm?.prepaymentPercent &&
```
jmiranda, line 173: We should encapsulate this logic in a domain method. In fact, I think we use each component in other places so let's add methods for each of them.
```
Boolean isPrepaymentInvoiceAllowed() { 
    return !this.hasInvoice && this.isPrepaymentRequired && !this.empty()
}

Boolean isPrepaymentRequired() { 
    return this.paymentTerm?.prepaymentPercent > 0
}

Boolean isEmpty() { 
    return this.orderItems?.empty && orderInstance?.orderAdjustments?.empty
}
```

I can't remember the convention for is/has methods in Grails, so you might have to play around with isSomeCondition() vs getIsSomeCondition().

And isEmpty is not descriptive enough so please suggest a more descriptive name if you can think of one. The following is a little long and awkward, but descriptive.
```
isOrderItemsAndOrderAdjustmentsEmpty
```

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
awalkowiak: But technically invoice does not have prepayment invoice (order does)

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
awalkowiak: This change was more for IDE to get rid of underscore

grails-app/conf/UrlMappings.groovy
awalkowiak: I think we should if we can specify the same endpoint for different request methods. In this case we had two separate endpoints for no reason.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
awalkowiak: Thanks

grails-app/views/order/_summary.gsp
awalkowiak: I'll tackle it with refactors for Order.getInvoiceItems

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
awalkowiak: I'll investigate it with other refactors, so this can be tested

pmuchowski (approved):

PR #2421, opened 2021-05-21
OBPIH-3778 Add checkbox to include all items in invoice modal

DominikaHatala:


src/js/components/invoice/InvoiceItemsModal.jsx
```diff
@@ -372,6 +407,14 @@ class InvoiceItemsModal extends Component {
             cache={false}
           />
         </div>
+        <div>
+          <Checkbox
```
awalkowiak, line 411: Perhaps a label would be handy

DominikaHatala: Now the label is right above other checkboxes, it would look odd to add label on the left side of it.

PR #2426, opened 2021-05-26
OBAM-116 Add replenishment type to stock list

awalkowiak:


awalkowiak:
@jmiranda I know, that there is a bit of code duplication for common columns, but I think, that it has better readability and it will be easier to make changes.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -113,6 +113,7 @@ class Requisition implements Comparable<Requisition>, Serializable {
     String monthRequested
 
     Integer replenishmentPeriod = 0
+    ReplenishmentType replenishmentType = ReplenishmentType.PUSH
```
jmiranda, line 116: Change this to ReplenishmentTypeCode to be consistent with other enums. We eventually might add a static data class (ReplenishmentType) that should allow the user to use their own naming.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
awalkowiak: @jmiranda Done

jmiranda (approved):

PR #2428, opened 2021-05-27
OBPIH-3856 Fix setting buyer organization for invoices

DominikaHatala:


jmiranda (commented):
What problem does this solve? If the order.destination.organization is not set then the order should not pass validation. And if the order.destination.organization is not set, the order.destinationParty won't be set either.

DominikaHatala:
@jmiranda destinationParty is not always equal destination.organization. This one was already fixed once and was working fine but then ir was broken after @awalkowiak implemented generating final invoice.

awalkowiak (approved):

jmiranda (approved):

PR #2430, opened 2021-05-28
OBAM-223 Add zone to Location

pmuchowski:


grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -57,8 +58,9 @@ class Location implements Comparable<Location>, java.io.Serializable {
     Date dateCreated
     Date lastUpdated
 
-    static belongsTo = [parentLocation: Location, organization: Organization]
+    static belongsTo = [parentLocation: Location, organization: Organization, zone: Location]
```
jmiranda, line 61: I don't think we need this. We just want the zone_id to be added to the location table. It's the same as location group and will only be used to group bin locations by zone.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -231,6 +234,14 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return supportsAny(requiredActivities) && !binLocations?.empty
     }
 
+    Boolean isInternalLocation() {
+        return locationType?.locationTypeCode in LocationTypeCode.listInternalTypeCodes()
```
jmiranda, line 238: So if you add isInternalLocation to LocationType, then this method would probably look something like 
```
return locationType?.isInternalLocation()
```
and LocationType.isInternalLocation() would look like this
```
return locationTypeCode in LocationTypeCode.listInternalTypeCodes()
```

jmiranda: Correct?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -565,7 +564,7 @@ class ProductAvailabilityService {
 
 
     void updateProductAvailability(Location location) {
-        def isBinLocation = location?.locationType?.locationTypeCode in [LocationTypeCode.BIN_LOCATION, LocationTypeCode.INTERNAL]
+        def isBinLocation = location?.isInternalLocation()
```
jmiranda, line 567: It's ok if you already made these changes. We should just add some notes in the ticket to ensure we test them.

jmiranda (changes requested):

jmiranda (approved):

PR #2433, opened 2021-05-31
OBPIH-3784 Show documents on the last page of invoice

DominikaHatala:


jmiranda (commented):
LGTM

The one thing I would have recommended (but I'm not 100% sure if it's the right approach) is that we add the link generation code as a domain method.
```
String getLink() {
    def g = ApplicationHolder.application.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
    return document.fileUri ?: g.createLink(controller: 'document', action: "download", id: document.id, absolute: true)
}
```
The one negative is that we're retrieving the taglib once for every document as opposed to just once in this code. Not sure how expensive that is (probably not much).

@DominikaHatala @awalkowiak Any thoughts about that? If you don't feel strongly either way just merge.

DominikaHatala:
@jmiranda I've tried exactly same way that you've proposed above but unfortunately it didn't work while getting documents on invoice react page.

src/js/components/invoice/ConfirmInvoicePage.jsx
```diff
@@ -268,6 +279,38 @@ class ConfirmInvoicePage extends Component {
...
+                      </button>
+                      <div className={`dropdown-content print-buttons-container col-md-3 flex-grow-1
+                        ${this.state.isDropdownVisible ? 'visible' : ''}`}
+                      >
+                        {this.state.values.documents.length > 0 &&
+                        _.map(this.state.values.documents, (document, idx) => {
+                          if (document.hidden) {
+                            return null;
+                          }
+                          console.log(document);
```
awalkowiak, line 300: console.log(document);

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/invoice/Invoice.groovy
```diff
@@ -133,11 +134,22 @@ class Invoice implements Serializable {
...
         def documents = []
+        def g = ApplicationHolder.application.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
         invoiceItems.each {invoiceItem ->
             if (invoiceItem?.order?.documents) {
                 invoiceItem?.order?.documents?.each {document ->
-                    if (!documents.contains(document)) {
-                        documents.add(document)
+                    if (!documents.find { it.id == document.id }) {
+                        def link = document.fileUri ?: g.createLink(controller: 'document', action: "download", id: document.id, absolute: true)
+                        documents << [
```
awalkowiak, line 143: Move this to the toJson domain method and align colons

awalkowiak (approved):

PR #2436, opened 2021-05-31
OBPIH-3858 Fix shipping status in the order summary view

awalkowiak:


jmiranda (commented):
Phew. Glad this was easy. I'll review tomorrow and try to come up with some tests.

jmiranda (approved):
LGTM - a little worried about the inline comment but let's see if it gets executed as I'd like to add more comments to our views.

PR #2437, opened 2021-05-31
OBPIH-3877 Generate source code from product code and vendor id

awalkowiak:


grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -303,7 +304,7 @@ class ProductSupplierDataService {
     def createProductSupplierWithoutPackage(Map params) {
         Product product = Product.get(params.product.id)
         ProductSupplier productSupplier = new ProductSupplier()
-        productSupplier.code = params.sourceCode?:identifierService.generateProductSupplierIdentifier(product?.productCode)
+        productSupplier.code = params.sourceCode ?: "${product?.productCode ? product?.productCode + Constants.DEFAULT_NAME_SEPARATOR : ''}${params.supplier.id}"
```
awalkowiak, line 307: @jmiranda this is what was requested in the ticket. However, I am not sure if this product supplier code should not be unique (technically it is not restricted, but only mentioned in the comment in the domain class)

jmiranda: @awalkowiak It should probably be unique but . Why are we using ${params.supplier.id}? That should probably be ${organization.code}. I might actually keep the i
```
identifierService.generateProductSupplierIdentifier(product?.productCode, params?.supplier?.id)
```
and hide the details of how we generate it in that method.

jmiranda (changes requested):
See comment in ProductSupplierDataService

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
awalkowiak: @jmiranda I made the requested refactor and I also tackled the `uniqueness` part of it.

awalkowiak:
@jmiranda let me know if there is something more to do

jmiranda (approved):

PR #2438, opened 2021-06-01
OBAM-224 Enable admin creation of zone locations

pmuchowski:


grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -426,6 +422,23 @@ class SelectTagLib {
         out << g.select(attrs)
     }
 
+    def selectZoneLocationByLocation = { attrs, body ->
+        def location = Location.get(attrs.id)
+
+        if (location) {
+            attrs.from = Location.findAllByParentLocationAndActive(location, true).sort {
```
jmiranda, line 429: The same logic should be used for line 322 in LocationController, so let's move this to the service layer and implement it using a criteria query or HQL. You can add the sort / filter in there as well.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -242,6 +242,10 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return locationType?.isFacilityLocation()
     }
 
+    Boolean isZoneLocation() {
```
jmiranda, line 245: Let's just call this isZone() throughout.

jmiranda: I know it's confusing because you might think it should be `isBin()` instead of `isBinLocation()` but bin location is the actual name for the thing and zone is the actual name for the thing.

jmiranda (changes requested):

jmiranda:
@awalkowiak Once those changes are made feel free to merge.

awalkowiak (approved):

PR #2442, opened 2021-06-01
OBPIH-3852 Fix Split line feature validation in combine shipments

awalkowiak:
(after step transition)


jmiranda (approved):
LGTM but I have no idea what's going on here. :rofl:

PR #2447, opened 2021-06-02
OBAM-225 Add zone to bin import/export

pmuchowski:


grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -353,22 +353,35 @@ class LocationController {
...
+        }
 
         if (!location) {
             throw new IllegalArgumentException("Must specify location")
         }
 
-        if (location.locations) {
+        def binLocations = location.internalLocations
+
+        if (zone) {
```
jmiranda, line 369: Feels like this would be better off being passed to a method on Location i.e. Location.getInternalLocationsByZone(Location zone) so we only have to iterate through that internal locations once.

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -416,6 +418,17 @@ class LocationService {
                         binLocation.locationNumber = it.name
                         binLocation.parentLocation = location
                         binLocation.locationType = defaultLocationType
+
+                        if (it.zoneName) {
+                            Location zone = Location.findByNameAndParentLocation(it.zoneName, location)
+
+                            if (!zone) {
+                                throw new ValidationException("Zone with name: ${it.zoneName} does not exits", binLocation.errors)
```
jmiranda, line 426: Typo: exits -> exist

jmiranda (changes requested):

awalkowiak (approved):

PR #2450, opened 2021-06-07
OBPIH-3862 Fix issues with prepayment

DominikaHatala:


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -394,7 +394,7 @@ class Order implements Serializable {
     }
 
     Boolean getHasItemsOrAdjustments() {
```
awalkowiak, line 396: If you return only active this should be indicated in the naming, because this one suggests, that it will check for any kind of order items or adjustments.

awalkowiak (changes requested):

awalkowiak (approved):

awalkowiak (approved):

PR #2453, opened 2021-06-08
OBPIH-3838 Add invoice type to regular invoice

DominikaHatala:


grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -87,6 +89,8 @@ class InvoiceApiController {
             invoice.invoiceNumber = identifierService.generateInvoiceIdentifier()
         }
 
+        invoice.invoiceType = InvoiceType.findByCode(InvoiceTypeCode.INVOICE)
```
awalkowiak, line 92: This is evaluated both by `create` and `update` actions. If in the future we will have a save on the last page of invoice wizard (or whenever you make an update on the invoice), then you will overwrite the type even for the prepayment invoices. Either move it to create action or add an `if not type` (I think, the latter should be fine)

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
DominikaHatala: @awalkowiak You're right, thanks! Done.

awalkowiak (approved):

PR #2460, opened 2021-06-09
OBAM-106 Add price history tab to supplier page

DominikaHatala:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1869,6 +1869,50 @@ class JsonController {
 
         render([fieldRemoved: fieldRemoved] as JSON)
     }
+
+    def getPriceHistory = {
```
jmiranda, line 1873: I eventually want to get rid of the JsonController as it was a terrible idea.

So let's add this to the OrderController.

jmiranda: Or SupplierController if we've added that.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
DominikaHatala: @jmiranda Thats a good suggestion. Thanks, done!

awalkowiak (approved):

PR #2462, opened 2021-06-09
OBAM-25 Add receipt data to delivery note

awalkowiak:


grails-app/views/deliveryNote/_printPage.gsp
```diff
@@ -165,6 +168,14 @@
                                 </g:if>
                             </td>
                         </g:if>
+                        <td class="middle">
+                            <g:set var="received" value="${shipmentItems?.receiptItems*.findAll { it.inventoryItem == inventoryItem }?.quantityReceived?.flatten()?.sum()}"/>
```
jmiranda, line 172: It is possible to move these two to a domain? I assume they'd go into RequisitionItem since that is the grain we're dealing with.

Above this, we have shipmentItems set to this ... so perhaps that can go into RequisitionItem as well.
```
// Not sure if this needs to be flattened
List<ShipmentItem> getShipmentItems() { 
    requisition?.shipment?.shipmentItems?.findAll { it.requisitionItem == requisitionItem }}.flatten()
}

List<ReceiptItem> getReceiptItems() { 
    shipmentItems.receiptItems*.findAll { it.inventoryItem == inventoryItem}.flatten()
}
```
Then you could do something like this
```
requisitionItem.receiptItems.sum { it?.quantityReceived }
requisitionItem.receiptItems.collect { it?.comment }.join(",")
```

jmiranda (changes requested):

grails-app/views/deliveryNote/_printPage.gsp
awalkowiak: I'll check that out

jmiranda (approved):

PR #2464, opened 2021-06-10
OBAM-36 Add zones to electronic stock card

DominikaHatala:


grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -719,6 +737,22 @@ class SelectTagLib {
                 writer << '</optgroup>'
                 writer.println()
             }
+            noGroup.each {
```
jmiranda, line 740: Wow, what is going on here. Besides "handling null group case".

Could you also add some comments to explain this?

jmiranda (changes requested):
LGTM, just add some comments on the noGroup case.

DominikaHatala:
@jmiranda @awalkowiak I've added comments so I think this one is ready to be merged.

awalkowiak (approved):

PR #2465, opened 2021-06-11
OBAM-35 Add zone to outbound workflow

DominikaHatala:


src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -309,8 +309,7 @@ class AvailableItem {
                 "productCode"     : inventoryItem?.product?.productCode,
                 lotNumber         : inventoryItem?.lotNumber,
                 expirationDate    : inventoryItem?.expirationDate?.format("MM/dd/yyyy"),
-                "binLocation.id"  : binLocation?.id,
-                "binLocation.name": binLocation?.name,
+                binLocation       : binLocation,
```
jmiranda, line 312: Be careful making changes like this. We probably only use these APIs for the React pages, but we don't for sure. So removing properties like this could break an API client that is expecting these fields to exist. We need to deprecated before we remove.

src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -593,7 +592,7 @@ class PackPageItem {
                 product        : shipmentItem?.product,
                 lotNumber      : shipmentItem?.lotNumber,
                 expirationDate : shipmentItem?.expirationDate?.format("MM/dd/yyyy"),
-                binLocationName: shipmentItem?.binLocation?.name,
+                binLocation    : shipmentItem?.binLocation,
```
jmiranda, line 595: Same here.

src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
jmiranda: @awalkowiak Thoughts?

jmiranda: FWIW I don't want to unnecessarily keep fields around that are deprecated but we should try to formalize a process around removing fields before we do this.

awalkowiak: @jmiranda it's a good point. I added a cart to the retro board to discuss how would be best to handle such cases.

src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -593,13 +595,16 @@ class PackPageItem {
...
                 uom            : shipmentItem?.product?.unitOfMeasure,
                 quantityShipped: shipmentItem?.quantity,
                 recipient      : shipmentItem?.recipient,
                 palletName     : palletName,
                 boxName        : boxName,
                 sortOrder      : sortOrder,
+                // deprecated
+                "binLocation.id"  : shipmentItem?.binLocation?.id,
```
awalkowiak, line 606: I think it should be `binLocationName` instead of `"binLocation.id"` and `"binLocation.name"`

awalkowiak (changes requested):

src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
DominikaHatala: @awalkowiak You're right. I've updated it.

awalkowiak (approved):

PR #2467, opened 2021-06-11
OBAM-25 Display receive date from receipt instead of shipment

awalkowiak:


grails-app/views/deliveryNote/print.gsp
```diff
@@ -224,7 +224,7 @@
                                             <label><warehouse:message code="deliveryNote.receivedDate.label" default="Received date"/>:</label>
                                         </td>
                                         <td>
-                                            <g:formatDate date="${requisition?.shipment?.expectedDeliveryDate}" format="d MMMMM yyyy  hh:mma"/>
+                                            <g:formatDate date="${requisition?.shipment?.receipt?.actualDeliveryDate}" format="d MMMMM yyyy  hh:mma"/>
```
jmiranda, line 227: There's a one to many association between shipment and receipts. This is using a convenience method to get the first receipt. Is that the expected behavior? We might need to ask Kelsey.

awalkowiak: You're right. I think it should be the earliest received date in case there are two or more receipts. Let's see what Kelsey wants here.

jmiranda: Since it's not specified, I guess we can merge noq and ask questions later.

PR #2468, opened 2021-06-11
OBPIH-3887 Fixed bug that occurs on Inventory Browser when there are …

jmiranda:
…no searchable product types

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -647,7 +647,7 @@ class ProductAvailabilityService {
             }
         }
 
-        def searchableTypes = ProductType.listAllBySupportedActivity([ProductActivityCode.SEARCHABLE])
+        def searchableTypes = ProductType.listAllBySupportedActivity([ProductActivityCode.SEARCHABLE])?:[null]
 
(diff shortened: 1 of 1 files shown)


awalkowiak (approved):

DominikaHatala (approved):

awalkowiak:
@jmiranda looks weird, but I guess its necessary

jmiranda:
> @jmiranda looks weird, but I guess its necessary

@awalkowiak Just to give a bit more background ...

Yeah this was the only way without having to do something like the following 
```
String query = "first part of query"
if (condition) {
    query+= "conditional party of query"
}
```
I was about to rewrite using criteria queries when I came across a clue in a Hibernate ticket (https://hibernate.atlassian.net/browse/HHH-2045?focusedCommentId=50211).
```
I may well be reading it incorrectly, but as I understand the ISO SQL 2003 grammar as specified
 at http://savage.net.au/SQL/sql-2003-2.bnf then `in ()` is not valid - the first element of the list is non-optional. 
```
To me that says we need at least one item in the list. So we either needed to pass a list with a valid product type (i.e. we need add a default product type in the database), a bogus product type (new ProductType(productTypeCode: SEARCHABLE), or a null. Since the latter worked, I decided that was enough for now.

awalkowiak:
@jmiranda that's what I figured out, but thanks for the explanation.

PR #2469, opened 2021-06-14
OBPIH-3838 Fix issues with gl account for order adjustments in invoice

DominikaHatala:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -337,7 +337,7 @@ class InvoiceService {
         InvoiceItem invoiceItem = new InvoiceItem(
             budgetCode: orderAdjustment.budgetCode,
             product: orderAdjustment.orderItem?.product,
-            glAccount: orderAdjustment.glAccount ?: orderAdjustment.orderItem?.glAccount,
+            glAccount: orderAdjustment.orderItem?.glAccount ?: orderAdjustment.orderAdjustmentType?.glAccount,
```
jmiranda, line 340: This doesn't feel right. It seems the order of precedence should be 

order adjustment (override) -> order item (override) -> order adjustment type (default)

DominikaHatala: @jmiranda You're right, I've missed that. I updated PR.

jmiranda: @awalkowiak @DominikaHatala If the ticket says to do it the way you did it, then ignore my comments and merge it. If it's not clear then go with the precedence order that I specified.

```
orderAdjustment?.glAccount ?: orderAdjustment?.orderItem?.glAccount ?: orderAdjustment?.orderAdjustmentType?.glAccount

```

DominikaHatala: It was not clearly specified in the ticket so I applied your suggestions.

jmiranda (approved):

PR #2489, opened 2021-06-18
OBAM-229 Fix deleting zone locations

awalkowiak:


grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -140,12 +140,23 @@ class LocationController {
         def locationInstance = Location.get(params.id)
         if (locationInstance) {
             try {
+                if (locationInstance.locationType.locationTypeCode == LocationTypeCode.ZONE && Location.findAllByZone(locationInstance)) {
```
pmuchowski, line 143: you should use the helper method here `isZoneLocation()`

pmuchowski (changes requested):

pmuchowski (approved):

PR #2491, opened 2021-06-21
OBPIH-3895 Remove organizations without supplier role from vendor dro…

DominikaHatala:
…pdown

### DIFF
--- a/src/js/utils/option-utils.jsx
+++ b/src/js/utils/option-utils.jsx
@@ -183,7 +183,7 @@ export const debounceProductsInOrders = (waitTime, minSearchLength, vendor, dest
 export const debounceOrganizationsFetch = (waitTime, minSearchLength) =>
   _.debounce((searchTerm, callback) => {
     if (searchTerm && searchTerm.length >= minSearchLength) {
-      apiClient.get(`/openboxes/api/organizations?q=${searchTerm}`)
+      apiClient.get(`/openboxes/api/organizations?q=${searchTerm}&roleType=ROLE_SUPPLIER`)
         .then(result => callback(
           null,
           {


grails-app/controllers/org/pih/warehouse/api/OrganizationApiController.groovy
```diff
@@ -24,4 +25,10 @@ class OrganizationApiController extends BaseDomainApiController {
         render ([data:organizations] as JSON)
      }
 
+    def listSuppliers = {
```
jmiranda, line 28: Now that I've had a few more minutes to look at this, why don't we just move roleType to the URL in option-utils.jsx instead of creating a new endpoint.

src/js/utils/option-utils.jsx
```diff
@@ -183,7 +183,7 @@ export const debounceProductsInOrders = (waitTime, minSearchLength, vendor, dest
 export const debounceOrganizationsFetch = (waitTime, minSearchLength) =>
   _.debounce((searchTerm, callback) => {
     if (searchTerm && searchTerm.length >= minSearchLength) {
-      apiClient.get(`/openboxes/api/organizations?q=${searchTerm}`)
+      apiClient.get(`/openboxes/api/organizations/listSuppliers?q=${searchTerm}`)
```
jmiranda, line 186: Therefore, this should be 
```
apiClient.get(`/openboxes/api/organizations?q=${searchTerm}&roleType=ROLE_SUPPLIER`)
```
Since there's only one value here, you might need to add some processing in the list controller action to convert the parameter to a list.
```
params.roleType = params.list("roleType")
```
In addition you might need to also convert it from a string to the enum.
```
params.roleType = params.list("roleType").collect { return it as RoleType }
```

jmiranda (changes requested):

src/js/utils/option-utils.jsx
DominikaHatala: That's right. I didn't even have to add any processing since it was already added. I've tried to do it that way at the beggining but instead of "roleType" parameter I added "roleTypes" so I had a typo.

awalkowiak (approved):

PR #2499, opened 2021-06-25
OBAM-28 Add ability to group picklist by zone when printing

pmuchowski:


awalkowiak (commented):
Could this all sorting be done on the domain level? Seems like the readability heavily dropped.

pmuchowski:
Sorting cannot be done in domain class, because requisition items are grouped by zone, so not all picklist items are displayed in given zone.

awalkowiak (approved):

jmiranda (approved):
I agree with artur and think we should refactor this and move more code to the domain. But we can do that later since I was the one that created the original mess so don't think it would be fair to Pawel to make him unspaghetti this code.

PR #2502, opened 2021-06-28
OBPIH-3902 Fix display of substitution items

DominikaHatala:


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -675,22 +675,18 @@ class StockMovementService {
                 'max': max ? max.toInteger() : null,
         ]);
 
-        def editItemsIds = data.collect { it.id }
```
pmuchowski, line 678: I think you can fix this by changing it to:
`def editItemsIds = data.collect { "'$it.id'" }.join(',')`
so the editItemsIds is a string and looks like this: "'id1','id2','id3','id4'"

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -675,22 +675,18 @@ class StockMovementService {
...
+        def productsMap = Product.findAllByIdInList(data.collect { it.product_id })
+                .inject([:]) {map, item -> map << [(item.id): item]}
 
-        def substitutionItemsMap = dataService.executeQuery("""
+        def editPageItems = data.collect {
+            def substitutionItems = dataService.executeQuery("""
```
pmuchowski, line 682: you will send n request to the DB instead of one, it will affect the performance

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: Yes, this would be a very bad performance degradation.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
DominikaHatala: Unfortunately it's still not working after adding it.

DominikaHatala:
@pmuchowski @jmiranda I have updated PR and fixed those issues.

pmuchowski (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -675,16 +675,14 @@ class StockMovementService {
...
 
         def substitutionItemsMap = dataService.executeQuery("""
                     select
                        *
                     FROM substitution_item
-                    where parent_requisition_item_id in (:ids)
-                    """, [
-                'ids': editItemsIds,
-        ]).inject([:]) {map, item -> map << [(item.parent_requisition_item_id): item]}
+                    where parent_requisition_item_id in (${editItemsIds})
```
jmiranda, line 684: I still don't really understand why we can't use the more common syntax that was already there. I don't mind you doing this
```
def editItemsIds = data.collect { "'$it.id'" }
```
because I understand you're trying to wrap each ID with quotes, but why can't you then pass that into the query as a parameter like we were already doing?

```
def substitutionItemsMap = dataService.executeQuery("""
select *
FROM substitution_item
where parent_requisition_item_id in (:ids)""", ["ids":editItemIds])
```

jmiranda: Can you also explain the reason we need to replace inject by groupBy?

jmiranda (commented):
@awalkowiak @pmuchowski  I don't care that much about the changes request below. I just want to feel comfortable that we're doing the right thing. Can you review and merge if you feel ok.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
DominikaHatala: @jmiranda It didn't work while passing it into the query, that is really strange but it returned 0 rows. We need to replase inject by groupBy because there can be more than one substitution item and with inject it always returned only 1.

awalkowiak (approved):

PR #2508, opened 2021-06-29
OBPIH-3894 Allow managers with invoice permission to delete pending i…

pmuchowski:
…nvoices

### DIFF
--- a/grails-app/conf/RoleFilters.groovy
+++ b/grails-app/conf/RoleFilters.groovy
@@ -38,6 +38,10 @@ class RoleFilters {
             'user'            : ['impersonate']
     ]
 
+    def static invoiceActions = [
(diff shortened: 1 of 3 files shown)


jmiranda (approved):
LGTM. We still need to add a "permission" for ROLE_MANAGER and above. Let me know if you need help with that.

pmuchowski:
@jmiranda should this be added in the RoleFilters?
Should I add a managerActions, similar to adminActions and superuserActions?

pmuchowski:
@jmiranda I think this can be done in two ways:
- check the permissions in the InvoiceController before the item is deleted
- add new actions in the RoleFilters for the invoice user, like it's done for adminActions and superuserActions
Which one should I use?

PR #2509, opened 2021-06-30
OBPIH-3825 Add delete all button to combined shipment workflow

DominikaHatala:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -149,7 +150,12 @@ class StockMovementApiController {
      */
     def removeAllItems = {
         Requisition requisition = Requisition.get(params.id)
-        stockMovementService.removeRequisitionItems(requisition)
+        Shipment shipment = Shipment.get(params.id)
```
jmiranda, line 153: Depending on the step in the workflow, would you ever want to delete from requisition and shipment? Or does the removeShipmentItems method handle that already?

jmiranda: I can see it below and no it does not.

DominikaHatala: @jmiranda When there is requisition then there are deleted both - requisition items and shipment items. When there is no requisition then we only need to remove shipment items.

awalkowiak (approved):

Issue #2459, opened 2021-06-08
Add Organization has a warning "Could not find matching constructor for: org.pih.warehouse.core.PartyType(java.lang.String)


khorjames:
Even though when I go to Configuration | Organizations and click on the button - Add Organization, it displays a warning as shown in the Title above, I was still able to successfully create an organisation.
<img width="1036" alt="Screenshot 2021-06-08 at 11 34 41 PM" src="https://user-images.githubusercontent.com/31814643/121214907-1f4ff600-c8b2-11eb-9650-23bddbb91a9d.png">

jmiranda:
Thanks @khorjames. Adding to the next release.

jmiranda:
@khorjames I just remembered that you're running the Grails 3 version of OpenBoxes so I moved this to the 0.9.0 release which will be the first release using Grails 3.

khorjames:
Thanks Justin.

How do I know the release will be made available ?

Best Regards,
James

jmiranda:
Are you asking how will you know "when" the release will be made available? 

If so, I really don't know. Hopefully in the next 6 months. But I've been saying that for about two years. The migration project is very high on my priority list, but very low on our users priority lists at the moment. I don't think they understand the implications well enough. But essentially that means there aren't as much resources (developers, funding, time, etc) so it's on me as the project maintainer to move the migration project forward on my own. However, I'm hoping to have a developer concentrating on this in the next month or two, so while it'll be slowe we'll probably start to make progress again soon. 

In any case, thank you for adding bug reports for this. I'll send you a note when we have a new version to test.

jmiranda:
@npokharel Ah yeah, I guess that works. :)

jmiranda:
@npokharel @khorjames While testing this issue on our test server I discovered that the error only occurs if a party type record with id = ORG does not exist. So this is actually a data error. 

@khorjames Can you test something for me? Take the URL for the Create Organization page and replace partyType=ORG with the partyType=1. For example, if you're running on localhost, replace this URL 
```
http://localhost:8080/openboxes/organization/create?partyType=ORG
```
with this one
```
http://localhost:8080/openboxes/organization/create?partyType=1
```
and let me know if you encounter the error.

NOTE: This will likely be fixed when we rebase the Grails 3 branch onto develop.

npokharel:
[...]

PR #2516, opened 2021-07-06
#2459 removed explicit partyType string param

npokharel:
The partyType field is being set by the dropdown list anyway. I don't see any other uses of that parameter. It has been added to rectify OBGM-285 .


jmiranda (commented):
See comment in ticket #2459

jmiranda:
@npokharel Thanks for investigating. We can close the PR for now, but leave the ticket open until after the rebase happens so it can be tested again.

PR #2518, opened 2021-07-07
OBAM-157 Add quantity available to promise to global search and inventory report

DominikaHatala:
…n location report

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
@@ -515,15 +515,28 @@ class InventoryController {
         def location = Location.get(session.warehouse.id)
         def quantityMap = dashboardService.getInStock(location)
         def statusMap = dashboardService.getInventoryStatus(location)
+        def availableItems = []
+
         if (params.format == "csv") {
             def filename = "In stock - " + location.name + ".csv"
             response.setHeader("Content-disposition", "attachment; filename=\"${filename}\"")
             render(contentType: "text/csv", text: getCsvForProductMap(quantityMap, statusMap))
             return
+        } else {
+            quantityMap.each { Product product, quantity ->
+                def inventoryLevel = product.getInventoryLevel(session.warehouse.id)
+                def quantityAvailableToPromise = product.getQuantityAvailableToPromise(session.warehouse.id)
+                availableItems << [
+                        status: statusMap[product],
+                        product: product,
+                        quantity: quantity,
+                        quantityAvailableToPromise: quantityAvailableToPromise,
+                        inventoryLevel: inventoryLevel
+                ]
+            }
         }
 
-        render(view: "list", model: [quantityMap: quantityMap, statusMap: statusMap])
-
+        render(view: "list", model: [availableItems: availableItems])
     }
 
     def listLowStock = {
@@ -736,6 +749,7 @@ class InventoryController {
         csv += '"' + "${warehouse.message(code: 'inventoryLevel.maxQuantity.label')}" + '"' + ","
         csv += '"' + "${warehouse.message(code: 'inventoryLevel.forecastQuantity.label')}" + '"' + ","
         csv += '"' + "${warehouse.message(code: 'inventoryLevel.currentQuantity.label', default: 'Current quantity')}" + '"' + ","
+        csv += '"' + "${warehouse.message(code: 'default.quantityAvailableToPromise.label', default: 'Quantity ATP')}" + '"' + ","
         csv += '"' + "${warehouse.message(code: 'product.pricePerUnit.label')}" + '"' + ","
         csv += '"' + "${warehouse.message(code: 'product.totalValue.label')}" + '"'
         csv += "\n"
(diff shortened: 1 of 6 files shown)


grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -457,12 +458,23 @@ class Product implements Comparable, Serializable {
...
+            projections {
+                sum("quantityAvailableToPromise", "quantityAvailableToPromise")
+            }
+            location {
+                eq("id", locationId)
+            }
+            eq("product", this)
+        }
+
+        return productAvailability?.get(0)?.quantityAvailableToPromise ?: 0
```
jmiranda, line 477: The .get(0) might throw an exception if there are no records in the table, so you should consider handling that edge case.

grails-app/views/inventory/list.gsp
```diff
@@ -72,6 +73,7 @@
                                 <g:each var="entry" in="${quantityMap.sort()}" status="i">
                                     <g:set var="product" value="${entry.key}"/>
                                     <g:set var="quantity" value="${entry.value}"/>
+                                    <g:set var="quantityAvailableToPromise" value="${entry?.key?.getQuantityAvailableToPromise(session.warehouse.id)}"/>
```
jmiranda, line 76: a few problems here: 
1) This will cause a huge amount of queries (1 for every row) 
2) Database access should never occur within the view

For MVC applications, we always want the controller to gather the data, prepare the model, and pass the model to the view. The view just reads from the model and should not make calls to the service or data access layers. The only time where it's appropriate to do so would be when using a taglib that needs to access the database through the service. But this is not one of those times. 

Instead you want to create a model that has all the data we need. I think this could be done with a small refactoring where we drop the quantityMap and use a list of AvailableItems.

jmiranda (changes requested):

DominikaHatala:
@jmiranda I've added all required changes to this PR.

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -457,12 +458,23 @@ class Product implements Comparable, Serializable {
...
 
     /**
-     * Currently not implement since it would require coupling InventoryService to Product.
+     * Get quantity available to promise for this product given the location
      *
      * @param locationId
      */
-    def getQuantityAvailableToPromise(Integer locationId) {
-        throw new NotImplementedException()
+    def getQuantityAvailableToPromise(String locationId) {
```
awalkowiak, line 465: I think this should be in the ProductAvailabilityService with params `Location location` and `Product product`

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -457,12 +458,23 @@ class Product implements Comparable, Serializable {
...
+        def productAvailability = ProductAvailability.createCriteria().list {
```
awalkowiak, line 466: If you want to return only the quantityAvailableToPromise, then I think you could do `.get()` without a `resultTransformer` and with a simple `return productAvailability?:0`

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1225,7 +1225,7 @@ class InventoryService implements ApplicationContextAware {
             eq("product", product)
         }
 
-        return productAvailability?.get(0)?.quantityAvailableToPromise ?: 0
+        return productAvailability ? productAvailability?.get(0)?.quantityAvailableToPromise : 0
```
awalkowiak, line 1228: Refactor these criteria to `.get {}` without `resultTransformer`

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -515,15 +515,28 @@ class InventoryController {
...
+
         if (params.format == "csv") {
             def filename = "In stock - " + location.name + ".csv"
             response.setHeader("Content-disposition", "attachment; filename=\"${filename}\"")
             render(contentType: "text/csv", text: getCsvForProductMap(quantityMap, statusMap))
             return
+        } else {
+            quantityMap.each { Product product, quantity ->
+                def inventoryLevel = product.getInventoryLevel(session.warehouse.id)
+                def quantityAvailableToPromise = inventoryService.getQuantityAvailableToPromise(product, location)
```
jmiranda, line 528: This is still doing a query per product. I was saying that we no longer need to get the quantityMap (above). Also, we could probably replace the query for inventoryLevel by doing a join in the getQuantityAvailableToPromise() query.
```
select <fields>
from product_availability
left outer join inventory level on product and location
```
And use the resultTransformer to return AvailableItems, including the inventoryLevel.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
jmiranda: Just make sure you only get one inventory level per product and location (i.e. you might need to join a subquery in order to ensure there's only one row per product, location returned by the inventory_level. It might look something like this in SQL.
```
select <fields>
from product_availability
left outer join (
    select 
        location_id, 
        product_id, 
        ... all the rest of the fields needed
    from inventory_level 
    join location on location.inventory_id = inventory_level.id
    group by location_id, product_id
    limit 1) as tmp on tmp.location_id = product_availability.location_id and tmp.product_id = product_availability.product_id
```

jmiranda:
@DominikaHatala @awalkowiak Sorry if I missed something, but are there any updates on this one? I keep checking back but haven't seen any changes. It's ok if this has taken a backseat to bug fixes.

DominikaHatala:
@jmiranda I've worked on this query again but didn't manage to make it work as it is supposed to so decided not to spend too much time as we have higher priority right now(fixing bugs). As we discussed last time - I think that we should merge it as it is and create new tickets to refactor fetching data for this report(probably creating a view for it would be the best option). What do you think?

jmiranda (approved):

PR #2519, opened 2021-07-07
OBAM-234 Prevent autopick of recalled lot

pmuchowski:


src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -301,6 +301,10 @@ class AvailableItem {
         quantityAvailable(nullable: true)
     }
 
+    def getIsAutopickable() {
```
jmiranda, line 304: I think my only beef is with the camelcase so let's go with isAutoPickable to make it more readable.

grails-app/domain/org/pih/warehouse/inventory/InventoryItem.groovy
```diff
@@ -132,4 +132,7 @@ class InventoryItem implements Serializable {
         return [product?.id]
     }
 
+    def getIsAutopickable() {
```
jmiranda, line 135: Also I always forget the convention when it comes to boolean properties, but I think you can just add "autoPickable" to the transients and isAutoPickable() as the method name.

jmiranda: https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter

jmiranda (changes requested):

awalkowiak (approved):

PR #2520, opened 2021-07-07
OBDS-46: Implement Dockerfile for OpenBoxes Grails v3 image

kkaczmarczyk:


docker/openboxes.yml
```diff
@@ -0,0 +1,6 @@
+dataSource.dbCreate: none
+dataSource.url: jdbc:mysql://${DB_HOST:host.docker.internal}:${DB_PORT:3306}/${DB_NAME:openboxes}?useSSL=false
```
kkaczmarczyk, line 2: @jmiranda would you say we should keep it as is with multiple variables or just have a single DATASOURCE_URL env variable to set the whole url?

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM openjdk:8u292-jre-slim-buster
+
+LABEL maintainer="support@openboxes.com"
+EXPOSE 8080
+ENV JAVA_TOOL_OPTIONS="-Xms1024m -Xmx1024m -XX:+UseParallelGC -Djava.awt.headless=true"
```
kkaczmarczyk, line 5: @jmiranda would you say that these default JVM heap size settings look alright?

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM openjdk:8u292-jre-slim-buster
```
kkaczmarczyk, line 1: I used Debian OpenJDK image as it has better support than Alpine one. So if there are updates to openjdk 8 it will be easier to update.

jmiranda: Sounds good to me.

docker/Dockerfile
jmiranda: I think we also need to add PermSize = 128m or 256m. The default memory settings look ok. I don't know if the default should be the minimum (512m) or recommended (2GB), so I think 1GB is a good happy medium.

kkaczmarczyk: PermSize got removed in Java 8, which OB Grails 3 uses: https://stackoverflow.com/a/22509753/2572524

docker/openboxes.yml
kkaczmarczyk: @jmiranda .env file is only used for the runtime environment. The openboxes.yml is a static part of the build environment, it is contained in the image, and it just references the env variables.

Docker doesn't require the openboxes.yml - we can just use env variables sourced from .env. After thinking, l reckon I could get rid of openboxes.yml, just document instructions to set up .env (DATASOURCE_URL primarily) properly. And docs should mention that these 3 (DB) env variables have to be provided to the runtime environment.

I didn't consider AWS Elastic Beanstalk or ECS support at this point. I think it would be great to support it though since we are at the point of having a working OB Docker image. Supporting Elastic Beanstalk involves a few other changes (not a lot) to be done to the Docker-related files to make it work - I think this can be done in a separate ticket.

docker/Dockerfile
jmiranda: Ah, I always thought that error just meant that you could only set PermSize and not MaxPermSize. Didn't realize they replaced permanent generation with metaspace. I'll need to read up on that.

jmiranda (approved):

PR #2523, opened 2021-07-08
OBPIH-3791 Allow user to rollback and delete shipments from PO when i…

DominikaHatala:
…n central purchasing location


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -318,6 +320,10 @@ class StockMovementController {
         params.direction = (currentLocation == stockMovement.origin) ? StockMovementType.OUTBOUND :
                 (currentLocation == stockMovement.destination) ? StockMovementType.INBOUND : "ALL"
 
+        if (isCentralPurchasingEnabled) {
+            redirect(controller: 'order', action: "list", params: [orderTypeCode: OrderTypeCode.PURCHASE_ORDER])
```
awalkowiak, line 324: After release when this will go to the develop branch, then this will need to be changed into OrderType

Issue #2417, opened 2021-05-19
Configuration > Show Settings > NullPointerException on Locale.getDisplayString()


jmiranda:
## Details 

| Property | Value |
| ----------- | -------- |
| Location | NAIROBI (c1e344346faf2191016fc73e80bd0021) |
| targetUri | /admin/showSettings |
| exception.message | None |
| exception.date | Sat Apr 03 10:45:09 UTC 2021 |
| absoluteTargetUri | http://demo.openboxes.com/openboxes |
| summary | None | 
| comments | None | 

## Stacktrace
```
    org.codehaus.groovy.grails.web.errors.GrailsWrappedRuntimeException
	at org.codehaus.groovy.grails.web.errors.GrailsExceptionResolver.resolveException(GrailsExceptionResolver.java:79)
	at org.springframework.web.servlet.DispatcherServlet.processHandlerException(DispatcherServlet.java:987)
	at org.codehaus.groovy.grails.web.servlet.GrailsDispatcherServlet.doDispatch(GrailsDispatcherServlet.java:339)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:719)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:644)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:549)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:621)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:728)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:303)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:241)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:70)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:241)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
[...]

PR #2524, opened 2021-07-08
#2417 added default locale for user with no locale

npokharel:
Another approach could be add default locale in SecurityFilters during currentUser setup but I don't think it is a good idea.


jmiranda:
@npokharel I think you have changes from develop in this PR. So can you recreate the PR off release/0.8.16 with just the following commit?
https://github.com/openboxes/openboxes/pull/2524/commits/6dfe298d0d93bdcd2e8476467104a44bba7bf980

I think the workflow could look something like this, but I'm sure there are better ways to do it.
```
git checkout release/0.8.16
git checkout -b 2417-show-settings-locale-second-attempt
git cherry-pick 6dfe298d0d93bdcd2e8476467104a44bba7bf980
git commit -m "#2417 added default locale for user with no locale"
git push -u origin 2417-show-settings-locale-second-attempt
```

jmiranda:
And if you're not comfortable handling that, I can work on it tomorrow.

npokharel:
I have re-created a new PR https://github.com/openboxes/openboxes/pull/2525 . Think we can close this one.

PR #2527, opened 2021-07-12
OBAM-217 Change autopick to ignore picked stock

pmuchowski:


pmuchowski:
@jmiranda there were some issues with using quantityAvailableToPromise instead of quantityOnHand, so in the PR it's used only during autopick.
There is one issue when you undo the picklist changes, it will do an autopick, but it will not work correctly, because the previout picklist items are still affecting the quantityAvailableToPromise, it is not recalculated before the autopick.
I think we will have to recalculate the quantityAvailableToPromise before every autopick to fix this issue, what do you think?

jmiranda (approved):

PR #2529, opened 2021-07-13
OBPIH-3942 Remove purchase order specific fields from putaway

DominikaHatala:


grails-app/views/order/show.gsp
```diff
@@ -59,14 +59,16 @@
...
                                 </tr>
-                                <tr class="prop">
-                                    <td valign="top" class="name">
-                                        <label><warehouse:message code="order.originCode.label"/></label>
-                                    </td>
-                                    <td valign="top" class="value">
-                                        ${orderInstance?.origin?.organization?.code }
-                                    </td>
-                                </tr>
+                                <g:if test="${orderInstance.orderTypeCode == OrderTypeCode.PURCHASE_ORDER}">
```
awalkowiak, line 62: I think this is a second PR, that will need a revisit after merging it to develop. Perhaps we should create a Jira ticket with a list of all PRs that will need refactor into OrderType for easy tracking?

awalkowiak (approved):

PR #2534, opened 2021-07-14
OBPIH-3908 Refactor search products method to use sql view

awalkowiak:


grails-app/domain/org/pih/warehouse/product/ProductSearch.groovy
```diff
@@ -0,0 +1,43 @@
...
+        version false
+        cache usage: "read-only"
+    }
+
+    static constraints = {}
+
+    static transients = ['color']
+
+    String getColor() {
+        return product?.color ?: ''
```
jmiranda, line 41: Can we join this in the view?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -673,53 +674,40 @@ class ProductAvailabilityService {
...
-                // This is pretty inefficient if the previous query does not narrow the results
-                if (innerProductIds) {
-                    'in'("id", innerProductIds)
+            }
+            eq("location", command.location)
+            or {
+                isNull("type")
+                and {
+                    eq("isSearchableType", Boolean.TRUE)
+                    gt("quantityOnHand", 0)
```
jmiranda, line 708: I think we should leave this up to the caller. There are negative quantities that we probably want to return in some cases (i.e. data error detection)

grails-app/migrations/views/product-search.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE VIEW product_search AS (
+SELECT
+  concat(p.id, ' ', p_a.location_id)            as id,
+  p.id                                          as product_id,
+  p_a.location_id                               as location_id,
+  p.product_type_id                             as type_id,
+  CASE
+  	WHEN GROUP_CONCAT(p_t_s_a.product_activity_code, ' ') is null THEN FALSE -- concat with ' ' for easier check, because other codes might include SEARCHABLE key word in code
```
jmiranda, line 8: Ooh this might be pretty bad performance-wise since we won't be able to index this. We can give this a shot, but it might be possible to allow product search to be the explosion and the query can change the grain to product-location. In other words, we don't GROUP BY until we get to the query. Leave it as-is for now and we can investigate in another ticket.

jmiranda (changes requested):
Nice work. I like the approach. Just a few comments / suggestions.

grails-app/domain/org/pih/warehouse/product/ProductSearch.groovy
awalkowiak: Unfortunately not because there is no `color column` (it is a formula in the Product domain)

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: This is the initial request from the ticket related to the searchable product types (if the type is searchable, show only products that have QoH)

grails-app/migrations/views/product-search.sql
awalkowiak: Quick check: I am getting results in ~600ms for a single location search (no product specified) in the database workbench. When I removed this property that is using `GROUP_CONCAT`, I am getting the same results in ~730ms. So having the `GROUP_CONCAT` improves the performance in this case somehow.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
jmiranda: Ok.

jmiranda (approved):

PR #2536, opened 2021-07-14
OBPIH-3932 Fix creating new source on edit order item modal

DominikaHatala:


awalkowiak (commented):
How does this fix the issue?

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -656,16 +656,22 @@ class OrderController {
...
             productSupplier = ProductSupplier.findByCodeAndSupplier(params.sourceCode, supplier)
             if (params.sourceCode && productSupplier) {
                 render(status: 500, text: "Product source with given code for your supplier already exists")
                 return
             }
-        }
-        if (params.productSupplier || params.supplierCode) {
-            productSupplier = productSupplierDataService.getOrCreateNew(params)
+            try {
+                productSupplier = productSupplierDataService.createProductSupplierWithoutPackage(params)
```
awalkowiak, line 670: You are basically creating the productSupplier for the second time here.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
DominikaHatala: In this case you can create without having enter any data for new product supplier or without having product supplier chosen before. In the previous if you can only get already available product supplier or create new with given data.

DominikaHatala:
@awalkowiak Please take a look at my comment in requested changes.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
awalkowiak: 1. `productSupplierDataService.getOrCreateNew(params)` will call `productSupplierDataService.createProductSupplierWithoutPackage(params)` inside. So it's triggered twice.
2. If you will have params.productSupplier == "Create New" and params.sourceCode, then the line 666 will trigger. (first, you create it in line 660, and then line 664 will find it and will complain that it already exists).
3. Indentation of try-catch seems to be a little off.

DominikaHatala:
@awalkowiak I've applied some changes.

jmiranda:
@awalkowiak I'll let you merge this when you feel good about it. I don't love the magic string ("Create new") since the string should technically be translated. And even if it's a static string in the select list, it could lead to a regression if someone were to change that value in the UI without changing this string.

awalkowiak:
@jmiranda The "Create new" is an option that is injected on the frontend, but you right. I think we could refactor this after release and instead of injecting it on the frontend, we could inject from the backend as an option with translated label and id made out of some constant.

awalkowiak (approved):

PR #2538, opened 2021-07-15
OBPIH-3940 Fix NPE during Product Availability update

awalkowiak:


jmiranda (approved):
LGTM thanks for removing the magic strings

PR #2539, opened 2021-07-16
OBAM-129 Fix filtering items and completing the stock transfer

awalkowiak:


grails-app/services/org/pih/warehouse/stockTransfer/StockTransferService.groovy
```diff
@@ -167,4 +171,88 @@ class StockTransferService {
...
+            command.quantity = stockTransferItem.quantity
+            command.otherLocation = stockTransferItem.location
+            command.otherBinLocation = stockTransferItem.destinationBinLocation
+            command.order = order
+            command.transferOut = Boolean.TRUE
+            command.disableRefresh = Boolean.TRUE
+            inventoryService.transferStock(command)
+        }
+
+        grailsApplication.mainContext.publishEvent(new StockTransferCompletedEvent(stockTransfer))
```
jmiranda, line 198: Is this necessary? If I'm not mistaken, the transferStock method will create a transaction that will do the same thing.

jmiranda: To verify this, just see whether the transfer cause the refresh to be executed twice.

awalkowiak: @jmiranda you are right, that was totally redundant. I removed it.

PR #2542, opened 2021-07-20
OBDS-48 Implement a docker-compose configuration for OpenBoxes

kkaczmarczyk:


docker/docker-compose-containerdb.yml
```diff
@@ -0,0 +1,37 @@
...
+      container_name: openboxes-db
+      environment:
+        MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-root}
+        MYSQL_DATABASE: openboxes
+        MYSQL_USER: ${DATASOURCE_USERNAME:-openboxes}
+        MYSQL_PASSWORD: ${DATASOURCE_PASSWORD:-openboxes}
+      volumes:
+        - ./mysql/:/var/lib/mysql/
+      healthcheck:
+        test: mysqladmin ping -h localhost -u $$MYSQL_USER --password=$$MYSQL_PASSWORD
```
kkaczmarczyk, line 35: review: look into having more verbose statuses

docker/Dockerfile
```diff
@@ -3,13 +3,19 @@ FROM openjdk:8u292-jre-slim-buster
...
+
 # Default flags for the JVM. These can be replaced at the runtime.
 ENV JAVA_TOOL_OPTIONS="-Xms1024m -Xmx1024m -XX:+UseParallelGC -Djava.awt.headless=true"
 
 # Create openboxes user and group
 RUN useradd --user-group --shell /bin/false  --home-dir /app openboxes
 WORKDIR /app
 
+# Create and chown 'uploads' directory for the OB app
+RUN mkdir uploads && chown openboxes:openboxes uploads/
```
jmiranda, line 17: It would be great to avoid this, but I think it's ok workaround. I created a ticket in OBPIH to see if we can bypass the need for temporary file storage. https://pihemr.atlassian.net/browse/OBPIH-3963

docker/README.MD
```diff
@@ -1,22 +1,34 @@
...
 
 **Note**: Run the below commands in **main** project directory.
 
 
-1. Run grails build to produce executable openboxes.war file. This WAR file contains embedded Tomcat.
-`grails prod war`
-2. Prepare docker build with gradle. It copies war and configuration files to *build/docker*.
-`./gradlew prepareDocker`
-3. Build a docker image:
+1. Prepare docker build with gradle. It runs grails build to produce executable openboxes.war file. This WAR file contains embedded Tomcat. It also copies war and configuration files to *build/docker*.
```
jmiranda, line 27: Save yourself the trouble and use the markdown syntax for numbered list.

docker/docker-compose-containerdb.yml
jmiranda: This should just be added to the Nice-to-Have list on the wiki page.

docker/docker-compose-hostdb.yml
```diff
@@ -0,0 +1,22 @@
...
+    app:
+      extends: 
+        file: docker-compose-generic.yml
+        service: app
+      image: ghcr.io/openboxes/openboxes:feature-upgrade-to-grails-3.3.10
+      # Use below 'image' property (with proper tag) if you use locally-built image
+      # image: openboxes/openboxes:latest
+      environment:
+        # It's hardcoded - app connects to the host's DB using host.docker.internal
+        DATASOURCE_URL: jdbc:mysql://host.docker.internal:3306/openboxes?useSSL=false
```
jmiranda, line 12: This should be configurable.

jmiranda (changes requested):
As I mentioned on the call, the default behavior for `docker-compose up` (or whatever it is) should be to create the database in a container.

Also what does docker-compose-generic represent? If it's supposed to be the default then let's call it docker-compose-default.yml. If it's a base docker-compose file that others inherit / amend let's call it docker-compose-base.yml.

docker/docker-compose-hostdb.yml
kkaczmarczyk: OK, I'll make this one default, and keep the DATASOURCE_URL configurable

kkaczmarczyk:
> As I mentioned on the call, the default behavior for `docker-compose up` (or whatever it is) should be to create the database in a container.

Sure, I'll switch the config files.

> Also what does docker-compose-generic represent? If it's supposed to be the default then let's call it docker-compose-default.yml. If it's a base docker-compose file that others inherit / amend let's call it docker-compose-base.yml.

Heh, I named it initially docker-compose-base.yml but changed to docker-compose-generic.yml just before commit :)
Makes sense - 'base' is more intuitive.

docker/README.MD
kkaczmarczyk: I had some issues with the markdown numbered list not following the order when the text was indented - I'll try to see if this can be amended.

jmiranda: From my experience you need to intend the code within a bulleted list (sometimes 4 spaces works, sometimes it requires)

1. Here's the first point with some code
        
        here's teh code

1. Point two without code

jmiranda (approved):

PR #2553, opened 2021-07-28
OBPIH-3945 Add picklist qty to ATP for quantity available

pmuchowski:


grails-app/controllers/org/pih/warehouse/api/StockMovementItemApiController.groovy
```diff
@@ -55,7 +52,7 @@ class StockMovementItemApiController {
         log.debug "JSON " + jsonObject.toString(4)
         StockMovementItem stockMovementItem = stockMovementService.getStockMovementItem(params.id)
 
-        stockMovementService.removeShipmentItemsForModifiedRequisitionItem(stockMovementItem)
+        stockMovementService.removeShipmentAndPicklistItemsForModifiedRequisitionItem(stockMovementItem)
```
jmiranda, line 55: This is fine for now, but I'd love to come up with a less verbose naming convention for some of these methods. Is there a way to describe this in fewer words? 
```
refreshOnUpdate() 
resetOnUpdate() 
```
or maybe even 
```
updatePicklist()
```
since ultimately that's what we're doing.


I had the same problem when naming the methods in InventoryService and now I want to go back to a simple name to describe what we're doing. For example, getQuantity() or getQuantityDetails() rather than awful names like getProductQuantityByInventoryItemMap().

grails-app/jobs/org/pih/warehouse/jobs/RefreshProductAvailabilityJob.groovy
```diff
@@ -35,30 +31,7 @@ class RefreshProductAvailabilityJob {
...
-                productIds.each { productId ->
-                    Product product = Product.load(productId)
-                    productAvailabilityService.refreshProductAvailability(product, forceRefresh)
-                }
-            }
-            // Calculate product availability for all products within all locations
-            else {
-                productAvailabilityService.refreshProductAvailability(forceRefresh)
-            }
+            productAvailabilityService.refreshProductsAvailability(locationId, productIds, forceRefresh)
```
jmiranda, line 34: We need to be careful with this though. The method signature implies that we're updating all products for a given location. But there's a side effect in that if we pass in a null locationId then we'll be updating these products over all locations. Which might not be what we want in all cases.

grails-app/services/org/pih/warehouse/inventory/RefreshProductAvailabilityEventService.groovy
```diff
@@ -25,7 +25,12 @@ class RefreshProductAvailabilityEventService implements ApplicationListener<Refr
             return
         }
 
-        productAvailabilityService.triggerRefreshProductAvailability(event.locationId,
-                event.productIds, event.forceRefresh)
+        if (event?.synchronousRequired) {
+            productAvailabilityService.refreshProductsAvailability(event.locationId,
```
jmiranda, line 29: Again be careful here. This action might be doing a lot more than the caller intended which might lead to performance implications.

jmiranda (commented):
LGTM, but want to let @awalkowiak take a look before merging.

awalkowiak (approved):

PR #2554, opened 2021-07-29
OBAM-247 Fix importing lead time and replenishment period to inventor…

dbienkowska:
…y levels

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/DataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/DataService.groovy
@@ -175,7 +175,7 @@ class DataService {
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -175,7 +175,7 @@ class DataService {
                     }
                 }
 
-                addInventoryLevelToProduct(product, location.inventory, preferredBinLocation, row.minQuantity, row.reorderQuantity, row.maxQuantity, row.preferredForReorder)
+                addInventoryLevelToProduct(product, location.inventory, preferredBinLocation, row.minQuantity, row.reorderQuantity, row.maxQuantity, row.expectedLeadTimeDays, row.replenishmentPeriodDays, row.preferredForReorder)
```
awalkowiak, line 178: The amount of parameters is getting out of hand. Can you pass the entire row here instead of destructured parameters?

awalkowiak (approved):

PR #2557, opened 2021-07-29
OBAM-246 Fix creating Purchase Order

awalkowiak:


pmuchowski (approved):

grails-app/views/order/show.gsp
```diff
@@ -59,7 +60,7 @@
                                         <format:metadata obj="${orderInstance?.orderType?.name}"/>
                                     </td>
                                 </tr>
-                                <g:if test="${orderInstance.orderTypeCode == OrderTypeCode.PURCHASE_ORDER}">
+                                <g:if test="${orderInstance.orderType == OrderType.findByCode(OrderTypeCode.PURCHASE_ORDER.name())}">
```
jmiranda, line 63: @awalkowiak Why not just do orderInstance?.orderType?.orderTypeCode == OrderTypeCode.PURCHASE_ORDER?

awalkowiak: I think that it can't be `orderInstance.orderType.orderTypeCode` because the OrderTypeCode can be shared by more OrderTypes (example: TRANSFER_ORDER for both putaway and transfer). It could be `orderInstance.orderType.code`. I added this change in my current work and I'll include it with my next PR.

PR #2588, opened 2021-08-16
OBKN-37 Spike: Implement mechanism for parsing / processing XML messages

npokharel:
Marshal / UnMarshal XML as per the `new_status.xml` file, validated against `trip_execution.xsd`


jmiranda:
Merged in #2590

PR #2591, opened 2021-08-19
OBKN-69 Add AcceptanceStatus POJOs

jmiranda:


src/java/org/pih/warehouse/xml/executionstatus/TripExecution.java
```diff
@@ -1,4 +1,6 @@
-package org.pih.warehouse.xml;
+package org.pih.warehouse.xml.executionstatus;
+import org.pih.warehouse.xml.acceptancestatus.Header;
```
npokharel, line 2: Wrong Header class included.

npokharel (commented):
Need to fix the Header class inside executionstatus package

PR #2598, opened 2021-08-23
OBKN-130 Create new API endpoint for products details

jmiranda:


jmiranda:
@chetanmaharishi Here's my first stab at the API. I'm sure there are more details we'll need to add and I might need to fix some bugs. But you should be able to test this out again the app.

jmiranda:
Closed as it was contaminated during a rebase
https://github.com/openboxes/openboxes/pull/2602

PR #2616, opened 2021-08-30
OBPIH-4012 Fix revise edit item

pmuchowski:


jmiranda (approved):
I don't understand why this worked ... but I don't really care.

image

pmuchowski:
When the items are sent from frontend, the requisition item is null (in the bind method the requisition item is not set).

PR #2623, opened 2021-09-03
OBAM-178 Create association between order and picklist

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/ReplenishmentApiController.groovy
```diff
@@ -144,4 +151,38 @@ class ReplenishmentApiController {
...
     }
+
+    def createPicklist = {
+        OrderItem orderItem = OrderItem.get(params.id)
+        if (!orderItem) {
+            throw new IllegalArgumentException("Can't find order item with given id: ${params.id}")
+        }
+
+//        TODO: OBAM-169: Remove user modified picklist for given orderItem
+//        replenishmentService.removePicklist(orderItem)
```
jmiranda, line 162: Can you explain the commented out code?

grails-app/conf/UrlMappings.groovy
```diff
@@ -342,6 +342,16 @@ class UrlMappings {
             action = [DELETE: "removeItem"]
         }
 
+        "/api/replenishments/$id/createPicklist"(parseRequest: true) {
```
jmiranda, line 345: I'd like to avoid remote procedure call naming convention so let's go with a subresource. I can't tell if we want the subresource should be picklist or picklistItems. 
```
"/api/replenishments/$id/picklists"(parseRequest: true) {
    controller = { "replenishmentApi" }
    action = [POST: "createPicklist", PUT: "updatePicklist", DELETE: "deletePicklist"]
}
```

jmiranda: My guess would be picklistItems as the subresource and methods could be createPicklistItems, updatePicklistItems, deletePicklistItems.

grails-app/domain/org/pih/warehouse/picklist/Picklist.groovy
```diff
@@ -64,6 +64,8 @@ class Picklist implements Serializable {
     static constraints = {
         name(nullable: true)
         description(nullable: true)
+        requisition(nullable: true, validator: { value, obj -> value || obj.order })
+        order(nullable: true, validator: { value, obj -> value || obj.requisition })
```
jmiranda, line 68: Good idea.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/api/ReplenishmentApiController.groovy
awalkowiak: Each order item in the replenishment will be auto-picked at first, and then the user will have an option to change the pick or go back to the auto-pick (it is the same as currently existing picking for requisitions). These lines are commented out because the auto picking and picking is not yet implemented and it should be added in the mentioned ticket (or in OBAM-173).

grails-app/services/org/pih/warehouse/replenishment/ReplenishmentService.groovy
```diff
@@ -7,18 +7,25 @@
  * the terms of this license.
  * You must not remove this notice, or any other, from this software.
  **/
-package replenishment
+package org.pih.warehouse.replenishment
```
jmiranda, line 10: thank you. i just noticed this last night when i was browsing the code for something else.

jmiranda (approved):

PR #2640, opened 2021-09-09
OBKN-159 Validate all XML messages against XSD

npokharel:


jmiranda (commented):
@npokharel Can you check the failing unit tests on this PR. Otherwise, it looks good to me. I'll pull it down and work on the get resource from classpath aspect of this.

npokharel:
@jmiranda I re-ran the workflow and it reports all checks are passed. Maybe the error occurred as previously I canceled the workflow manually.

PR #2646, opened 2021-09-13
OBKN-55 Allow a user to scan a barcode to retrieve details for a product

jmiranda:
Global search of product from barcode


jmiranda (commented):
We'll leave this for now, but we'll want to use the default url mapping
```
/api/${resource}s -> ProductApiController.list() 
```

PR #2650, opened 2021-09-16
OBAM-172 First page of the bin replenishment wizard

awalkowiak:


grails-app/services/org/pih/warehouse/replenishment/ReplenishmentService.groovy
```diff
@@ -37,10 +37,26 @@ class ReplenishmentService {
     boolean transactional = true
 
     def getRequirements(Location location, InventoryLevelStatus inventoryLevelStatus) {
-        return Requirement.createCriteria().list() {
+        List<Requirement> requirements = []
+        Requirement.createCriteria().list() {
             eq("location", location)
             eq("status", inventoryLevelStatus)
+        }.collect { Requirement requirement ->
```
pmuchowski, line 44: why are you using collect here instead of findAll?

awalkowiak: I thought it will be an easier and cleaner way of getting rid of duplicates.

pmuchowski: in this case unique() should be fine

jmiranda: Agree with @pmuchowski. Seems like each (iterate) would be better than collect (transform), but unique seems like the better option.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/replenishment/ReplenishmentService.groovy
awalkowiak: fixed

pmuchowski (approved):

PR #2660, opened 2021-09-21
OBAM-173  Add Pick page for Bin Replenishment feature

awalkowiak:


jmiranda (commented):
Looks like a lot of copy-and-paste code which feels a little uncomfortable. Maybe we should consider moving some of the code to a PicklistService.

awalkowiak:
@jmiranda I was thinking about picklist service, but I wanted to get it working now and refactor it the improvement ticket. However, if you want this type of service now, then I can do it.

pmuchowski (approved):

PR #2662, opened 2021-09-22
OBPIH-3751 Fix ad hoc/stock indicator

dbienkowska:


grails-app/services/org/pih/warehouse/tableroapi/IndicatorDataService.groovy
```diff
@@ -961,8 +961,8 @@ class IndicatorDataService {
 
     GraphData getPercentageAdHoc(Location location) {
         Calendar calendar = Calendar.instance
-        // we need to get all requisitions that were created from the first day of the current month
-        calendar.set(calendar.get(Calendar.YEAR), calendar.get(Calendar.MONTH), 1, 0, 0, 0)
+        // we need to get all requisitions that were created from the first day of the previous month
+        calendar.set(calendar.get(Calendar.YEAR), calendar.get(Calendar.MONTH) -1, 1, 0, 0, 0)
```
awalkowiak, line 965: Will it work for January?

awalkowiak: plus there should also be added upper boundaries for the previous month's end. Currently, it's everything made after this date.

awalkowiak (changes requested):

awalkowiak (approved):

PR #2670, opened 2021-09-24
OBKN-186 Location dropdowns on create outbound stock movement page should provide options on click

npokharel:
Added for destination only.

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/CreateStockMovement.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/CreateStockMovement.jsx
@@ -88,21 +88,10 @@ const FIELDS = {
     defaultMessage: 'Destination',
     attributes: {
       required: true,
-      async: true,
       showValueTooltip: true,
-      openOnClick: false,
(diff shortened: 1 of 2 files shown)


src/js/components/stock-movement-wizard/outbound/CreateStockMovement.jsx
```diff
@@ -242,6 +242,23 @@ class CreateStockMovement extends Component {
       .catch(() => this.props.hideSpinner());
   }
 
+  fetchDestinations() {
+    const url = '/openboxes/api/locations?name=&direction=OUTBOUND';
+    return apiClient.get(url)
```
awalkowiak, line 247: This pulls all locations (around 1200) from the backend. I think, that when we were talking about this feature, we wanted just to show a few example locations for users to know what to type there and if the user types something in it will be pulled from the backend. Unless the requirements have changed since the tech huddle we had with @jmiranda.

awalkowiak (changes requested):

PR #2673, opened 2021-09-29
OBPIH-4047 Fix issues with invoice item candidate not appearing after…

dbienkowska:
… first not full invoice

### DIFF
--- a/grails-app/migrations/views/invoice-candidate-view.sql
+++ b/grails-app/migrations/views/invoice-candidate-view.sql
@@ -11,6 +11,7 @@ CREATE OR REPLACE VIEW invoice_item_candidate AS
         order_adjustment.description as description,
         1 as quantity,
         1 as quantity_to_invoice,
+        0 as quantity_from_invoice,
         order_item.quantity_uom_id as quantity_uom_id,
         order_item.unit_price as unit_price,
         order_item.quantity_per_uom as quantity_per_uom,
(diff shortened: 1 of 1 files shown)


grails-app/migrations/views/invoice-candidate-view.sql
```diff
@@ -11,6 +11,7 @@ CREATE OR REPLACE VIEW invoice_item_candidate AS
         order_adjustment.description as description,
         1 as quantity,
         1 as quantity_to_invoice,
+        0 as quantity_from_invoice,
```
awalkowiak, line 14: Is this used somewhere?

grails-app/migrations/views/invoice-candidate-view.sql
```diff
@@ -38,8 +39,11 @@ union
         product.gl_account_id as gl_account_id,
         product.product_code as product_code,
         product.name as description,
-        shipment_item.quantity/order_item.quantity_per_uom as quantity,
-        shipment_item.quantity/order_item.quantity_per_uom as quantity_to_invoice,
+        case when invoice_item.id is not null then (shipment_item.quantity/order_item.quantity_per_uom) - invoice_item.quantity
+        else shipment_item.quantity/order_item.quantity_per_uom end as quantity,
+	    case when invoice_item.id is not null then (shipment_item.quantity/order_item.quantity_per_uom) - invoice_item.quantity
+        else shipment_item.quantity/order_item.quantity_per_uom end as quantity_to_invoice,
+        invoice_item.quantity as quantity_from_invoice,
```
awalkowiak, line 46: same question as above

grails-app/migrations/views/invoice-candidate-view.sql
```diff
@@ -38,8 +39,10 @@ union
         product.gl_account_id as gl_account_id,
         product.product_code as product_code,
         product.name as description,
-        shipment_item.quantity/order_item.quantity_per_uom as quantity,
-        shipment_item.quantity/order_item.quantity_per_uom as quantity_to_invoice,
+        case when invoice_item.id is not null then (shipment_item.quantity/order_item.quantity_per_uom) - invoice_item.quantity
```
awalkowiak, line 42: @dbienkowska is it possible, that the shipment item will have more than two invoice items? then it should be sum of all `invoice_item.quantity`

awalkowiak: @dbienkowska Additionally, do we need both `quantity` and `quantity_to_invoice`? Seems like these values are the same.

jmiranda (approved):
LCTM - looks complicated to me

pmuchowski (approved):

PR #2687, opened 2021-10-12
OBAM-140 Add forecasting configuration to location edit page

dbienkowska:


dbienkowska:
@jmiranda In inventory level, product is not nullable constraint. Should it be changed to be nullable?

PR #2732, opened 2021-11-01
OBPIH-4108 Fixed bug with PersistenceContextJobListener configuration…

jmiranda:
…; use fatal log level on exceptions that occur within job context listener


jmiranda:
@awalkowiak @pmuchowski Don't merge this until you fix OBPIH-4081 or else the exception that causes the bug in that ticket will be swallowed by the job context listener.

PR #2738, opened 2021-11-03
OBPIH-4081 Fix product availability update when lot with wrong date a…

pmuchowski:
…dded

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
@@ -16,7 +16,7 @@ import util.ConfigHelper
 
 class ErrorsController {
 
-    def messageSource
+    def messageService
     MailService mailService
     def userService
     def grailsApplication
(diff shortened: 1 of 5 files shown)


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -95,9 +97,17 @@ class ErrorsController {
     def handleValidationErrors = {
         if (RequestUtil.isAjax(request)) {
             response.status = 400
+            def dateFormat = new SimpleDateFormat("MM/dd/yyyy")
             BeanPropertyBindingResult errors = request?.exception?.cause?.errors
             def errorMessages = errors.allErrors.collect {
-                return messageSource.getMessage(it.codes[0], it.arguments, it.defaultMessage, null)
+                def arguments = it.arguments?.collect {
```
awalkowiak, line 103: @pmuchowski please add a comment on why this is done here. It'd be good to have such information here for the future.

pmuchowski: @awalkowiak I needed to change this again in different PR, so I added comments there

jmiranda: Let's remove this here if it's modified somewhere else. Otherwise we'll need to fix the conflicts. And then I'll have to fix those same conflicts every time we rebase the OBKN and Grails 3 branches.

jmiranda: I'd also like to see if we could move this logic to the message service since it seems like a common use case.

pmuchowski: @jmiranda I removed those changes

PR #2740, opened 2021-11-04
OBPIH-4113: Remove local field from Location and fix Transaction vali…

pmuchowski:
…dation

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
@@ -101,7 +101,6 @@ class ErrorsController {
             }
             render([errorCode: 400,
                     errorMessage: "Validation error. " + request?.exception?.cause?.fullMessage,
-                    data: errors?.allErrors,
                     errorMessages: errorMessages
             ] as JSON)
             return
--- a/grails-app/domain/org/pih/warehouse/core/Location.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Location.groovy
@@ -51,7 +51,7 @@ class Location implements Comparable<Location>, java.io.Serializable {
 
     User manager                                // the person in charge of the warehouse
     Inventory inventory                            // each warehouse has a single inventory
-    Boolean local = Boolean.TRUE
+
     // indicates whether this warehouse is being managed on the locally deployed system
     Boolean active = Boolean.TRUE
     // indicates whether this warehouse is currently active
@@ -86,7 +86,6 @@ class Location implements Comparable<Location>, java.io.Serializable {
         })
         fgColor(nullable: true)
         logo(nullable: true, maxSize: 10485760) // 10 MBs
-        local(nullable: true)
         manager(nullable: true)
         inventory(nullable: true)
         active(nullable: false)
@@ -100,7 +99,7 @@ class Location implements Comparable<Location>, java.io.Serializable {
         cache true
     }
 
-    static transients = ["transactions", "events", "shipments", "requests", "orders"]
+    static transients = ["transactions", "events", "shipments", "requests", "orders", "managedLocally"]
 
     List getTransactions() { return Transaction.findAllByDestinationOrSource(this, this) }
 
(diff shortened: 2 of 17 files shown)


grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -96,12 +96,39 @@ class ErrorsController {
         if (RequestUtil.isAjax(request)) {
             response.status = 400
             BeanPropertyBindingResult errors = request?.exception?.cause?.errors
+            def validationErrors = errors?.fieldErrors?.collect {
```
jmiranda, line 99: I kind of understand what's happening here. But there's something I don't like about doing this block of code. Is there another way to achieve this without the hardcoded Location logic. And why is the marshaller being invoked on a validation error? Do we really send back the entire object on validation errors? 

@pmuchowski Could you remove the logic and record a video of the behavior you're seeing?

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -161,13 +160,17 @@ class Location implements Comparable<Location>, java.io.Serializable {
 
     }
 
+    Boolean isManagedLocally() {
```
jmiranda, line 163: This is a great idea. You could have just used isLocal() or getLocal() so you didn't have to change all "local" references to "managedLocally". However, if you made the change to be more readable (since the meaning of "local" is ambiguous) then I support that decision.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1879,32 +1879,27 @@ class InventoryService implements ApplicationContextAware {
...
         if (transaction?.transactionType?.id != Constants.TRANSFER_IN_TRANSACTION_TYPE_ID &&
                 transaction?.transactionType?.id != Constants.TRANSFER_OUT_TRANSACTION_TYPE_ID) {
-            return false
+            transaction.errors.rejectValue("transactionType", "transaction.localTransfer.invalidType", "Transaction have invalid type for local transfer")
         }
 
         // make sure we are operating only on locally managed warehouses
         if (transaction?.source) {
-            if (!(transaction?.source instanceof Location)) {
+            if (!(transaction?.source instanceof Location) || !transaction?.source?.managedLocally) {
```
jmiranda, line 1891: What weird logic. I don't remember why we had to do this. What else could source have been? :confused:

jmiranda:
@pmuchowski I'm fine with this getting merged if @awalkowiak thinks it looks ok. But I'd really like to think more about the validation error handler and would like to see a video showing the behavior before you added the Location logic. And maybe even what the JSON looks like if it works without a lazy init exception. I was not aware that we were passing back the entire object for validation errors. And my guess is that we shouldn't be doing that for any object. My preferred solution would be "return validation error JSON that works the same for all objects" rather than "make Location validation error JSON work in a special way".

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
pmuchowski: @jmiranda I'm not sure, do you think I should remove this?

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
pmuchowski: @jmiranda we're not using the validation error anywhere on the frontend, we're just using the error messages, do you think it's fine to just remove it from the json?

grails-app/domain/org/pih/warehouse/core/Location.groovy
pmuchowski: yes, I changed it to be more clear.

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
jmiranda: That would probably be ok. Although there are some cases where we'd like the client to know what the validation errors are. But if you're saying that we're already getting all validation errors as processed through the message service and not the raw components of validation errors, then yes I think we can remove the unprocessed validation errors.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
jmiranda: Added to https://trello.com/b/7uBWaaYi/sprint-75-wonderful-hound

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
pmuchowski: @jmiranda yes, we're using the message service to process all of the validation errors, so all of the validation errors are already displayed on the frontend.
In this case I will just remove the plain object, because it's not used anyway.

pmuchowski:
@jmiranda I updated the PR

jmiranda (approved):

PR #2746, opened 2021-11-08
OBPIH-4058 Formatted values in export and import PO

MagdaToczek:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -676,7 +676,7 @@ class OrderService {
                     if (quantity == "") {
                         throw new IllegalArgumentException("Missing quantity.")
                     }
-                    Integer parsedQty = Integer.valueOf(quantity)
+                    Integer parsedQty = java.text.NumberFormat.instance.parse(quantity)
```
pmuchowski, line 679: you can add an import instead of using package here, also you can use the getNumberInstance() instead of instance, like it's done in all other places

pmuchowski (changes requested):

pmuchowski (approved):

PR #2747, opened 2021-11-08
OBPIH-4127 Disabled and not displayed negative values on edit pick modal

MagdaToczek:


pmuchowski (changes requested):
The negative values should not be displayed on the frontend

pmuchowski (approved):

PR #2748, opened 2021-11-09
OBPIH-4119 Added improvements to stock card - pending outbound

MagdaToczek:


grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -320,14 +320,14 @@ class InventoryItemController {
         requisitionItems = requisitionService.getPendingRequisitionItems(location, product)
         requisitionItems.groupBy { it.requisition }.collect { k, v ->
             itemsMap.put(k, [
+                    requisitionItems: requisitionService.getListOfDetailsForPendingOutbound(v),
```
pmuchowski, line 323: I'm not sure if this is required, you can also access the picklist items by requsition.picklist.picklistItems.
In this case you can add here picklistItems group by lot and sum the quantity.
Also you should rename this field, those are not requisition items, it can be quantityPickedByLot

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -320,14 +320,14 @@ class InventoryItemController {
...
                     quantityRequested: v.quantity.sum(),
                     quantityRequired: v.sum() { RequisitionItem requisitionItem -> requisitionItem.calculateQuantityRequired() },
                     quantityPicked: v.sum() { RequisitionItem requisitionItem -> requisitionItem.calculateQuantityPicked() },
```
pmuchowski, line 326: if it's not needed anymore you can remove it

grails-app/views/inventoryItem/_showPendingOutboundStock.gsp
```diff
@@ -75,9 +72,22 @@
                     <td>
                         ${entry.value["quantityRequired"]} ${product?.unitOfMeasure}
                     </td>
-                    <td>
-                        ${entry.value["quantityPicked"]} ${product?.unitOfMeasure}
-                    </td>
+                    <g:set var="entryValue" value="${entry.value}"/>
```
pmuchowski, line 75: why you're getting entry value here, you can just take the quantityPickedByLot

pmuchowski (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
pmuchowski: or change the quantityPicked to quantityPickedByLot

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
MagdaToczek: Unfortunately, not every requisition has a picklist and in that case I'm retrieving this value from _requisitionItem.calculateQuantityPicked()_

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
MagdaToczek: It is sum up qty picked for requisition and is used in another place for summing this values at a bottom of a table

pmuchowski: ok

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
pmuchowski: In case requisition has no picklist the quantity picked will be 0 and there will be no picklist items

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -320,14 +320,17 @@ class InventoryItemController {
         requisitionItems = requisitionService.getPendingRequisitionItems(location, product)
         requisitionItems.groupBy { it.requisition }.collect { k, v ->
             itemsMap.put(k, [
+                    picklistItemsByLot: v.get(0).requisition.picklist ?
```
pmuchowski, line 323: I think you can use k instead of v.get(0).requisition,
also you don't need a null check you can use k?.picklist.? (it's null safe)

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -320,14 +320,14 @@ class InventoryItemController {
         requisitionItems = requisitionService.getPendingRequisitionItems(location, product)
         requisitionItems.groupBy { it.requisition }.collect { k, v ->
             itemsMap.put(k, [
+                    picklistItemsByLot: k?.picklist?.picklistItems?.findAll {it.quantity && it.inventoryItem.product == product}?.groupBy {it.inventoryItem.lotNumber},
```
pmuchowski, line 323: I think it may be safer to check if products have the same id

pmuchowski (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
jmiranda: Also, that feels like logic that would be useful in the Picklist class itself. 
```
List<PicklistItem> getPicklistItems(Product product) { 
    return picklistItems?.findAll { it.quantity && it.inventoryItem.product?.id == product?.id }
}

def getPicklistItemsByLot(Product product)  { 
    return getPicklistItems(product)?.groupBy {it.inventoryItem.lotNumber}
}
```
then in InventoryItemController
```
itemsMap.put(k, [
    picklistItemsByLot: k?.picklist?.getPicklistItemsGroupByLot(product),
    ...
])
```

MagdaToczek: Done :)

pmuchowski (approved):

PR #2754, opened 2021-11-10
OBPIH-4120 Refresh QATP on stock card after recall and record stock

pmuchowski:


pmuchowski:
@jmiranda are you fine with this approach?
The quantity ATP is fetched from Product availability table and displayed on the stock card.
I think the other solution could be to calculate it from the transaction like QOH, but it would be more complicated and we would have to duplicate the QATP calculation logic.
What do you think?

jmiranda (approved):

PR #2761, opened 2021-11-15
OBPIH-4123 Update Stock Transfer to Exclude Picked Items

awalkowiak:


jmiranda (approved):
LGTM. Could you just change "Quantity on Shelf" to "Quantity Available to Return"

awalkowiak:
@jmiranda but this is not related to any kind of return. You are making a stock transfer inside one location (between two bins). I'll confirm the labels with Kelsey in the ticket, but for now, I guess we could move it to QA.

awalkowiak:
@jmiranda I'll change it to "Quantity Available to Transfer" for now

pmuchowski (approved):

PR #2768, opened 2021-11-17
OBKN-99 Allow a warehouse user to print barcode labels for a bin location

chetanmaharishi:
OBKN-99 Allow a warehouse user to print barcode labels for a bin location


jmiranda:
This one was already handled by OBKN-266.

PR #2769, opened 2021-11-18
OBPIH-4139 Corrected showing wrong lot numbers on pack page after ref…

MagdaToczek:
…reshing pick page

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1025,7 +1025,13 @@ class StockMovementService {
      */
     void createPicklist(String id) {
         StockMovement stockMovement = getStockMovement(id)
+        for (RequisitionItem requisitionItem : stockMovement.requisition.requisitionItems) {
+            removeShipmentItemsForModifiedRequisitionItem(requisitionItem)
+        }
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1070,6 +1070,16 @@ class StockMovementService {
             log.info "Available items: ${availableItems}"
             List<SuggestedItem> suggestedItems = getSuggestedItems(availableItems, quantityRequired)
             log.info "Suggested items " + suggestedItems
+
+            def shipmentItems = ShipmentItem.findAllByRequisitionItem(requisitionItem)
```
pmuchowski, line 1074: it should not be done here, it may cause some issues and create the shipment items two times
move this to `void createPicklist(String id)` method (in line 1026), so it's done only on picklist refresh,
also you can use `removeShipmentItemsForModifiedRequisitionItem` method for this

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
MagdaToczek: Corrected

pmuchowski (approved):

PR #2795, opened 2021-11-25
OLMIS-4151 Added validation on pick page when after reload there is n…

MagdaToczek:
…othing to pick

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1027,7 +1027,7 @@ class StockMovementService {
 
     void createMissingPicklistItems(StockMovementItem stockMovementItem) {
         if (!stockMovementItem.requisitionItem?.picklistItems) {
-            createPicklist(stockMovementItem)
+            createPicklist(stockMovementItem, false)
         }
     }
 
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1063,6 +1063,16 @@ class StockMovementService {
                 createPicklist(stockMovementItem)
             }
         }
+        for (StockMovementItem stockMovementItem : stockMovement.lineItems) {
```
pmuchowski, line 1066: I think it may be best to move this to createPicklist(RequisitionItem requisitionItem), you can add a parameter to it and to createPicklist(StockMovementItem stockMovementItem), it could be named validateQtyAvailable and if it's true you could check if there are any available items for qiven requisition item and throw the validation error then

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1101,6 +1101,15 @@ class StockMovementService {
                             null)
                 }
             }
+            if (validateQtyAvailable && !suggestedItems) {
+                if (!requisitionItem.picklistItems && !(requisitionItem.status in [RequisitionItemStatus.CANCELED, RequisitionItemStatus.SUBSTITUTED, RequisitionItemStatus.CHANGED])) {
```
pmuchowski, line 1105: this part should not be needed anymore

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
MagdaToczek: Right, done.

pmuchowski (approved):

PR #2809, opened 2021-11-30
OBPIH-4058 Resolve some CSV round-trip headaches

mdpearson:
This PR should make CSV import and export more reliable.

On the import side, use `NumberFormat.parse()` and friends instead of `valueOf`, which allows things like commas in "1,000".

On the export side, use `NumberFormat.format()` and friends instead of grails's `numberFormat` which is optimized for screen readers and seems broken in French.

I didn't add any unit tests for this PR, but really should; my testing involved making pathological PO's with large numbers of extremely cheap ($0.0049 per 250 ct) supplies.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
@@ -24,6 +24,7 @@ import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Organization
 import org.pih.warehouse.core.User
 import org.pih.warehouse.core.ValidationCode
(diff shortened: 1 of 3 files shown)


awalkowiak:
@mdpearson I think this PR should target the `release/0.8.17` branch since ticket has `0.8.17` label.

cc @jmiranda

pmuchowski (approved):

MagdaToczek (approved):

awalkowiak:
I made a separate PR (https://github.com/openboxes/openboxes/pull/2815) targetting the release branch. It will be merged when Katarzyna gives the green light (to avoid QA interruption)

PR #2812, opened 2021-12-01
OBAM-184 Added generation of sales invoices to outbound movement and …

MagdaToczek:
…stock request


pmuchowski (changes requested):
I think you will need to change the getHasRegularInvoice method in Order, OrderItem and OrderAdjustment and default InvoiceTypeCode in InvoiceType too.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -398,7 +398,7 @@ class Order implements Serializable {
      * @return true if order has a regular invoice; false otherwise
      */
     Boolean getHasRegularInvoice() {
-        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.INVOICE }
+        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.PURCHASE_INVOICE }
```
pmuchowski, line 401: Should it be only purchase invoice or other types of invoices too?
I think it should be INVOICE, PURCHASE_INVOICE and SALES_INVOICE
@awalkowiak what do you think?

awalkowiak: I agree, previously these three types had code `INVOICE`, now after splitting this should be covered here too

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -107,6 +107,6 @@ class OrderAdjustment implements Serializable {
     }
 
     Boolean getHasRegularInvoice() {
-        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.INVOICE }
+        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.PURCHASE_INVOICE }
```
awalkowiak, line 110: Same as above

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -336,7 +336,7 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
     }
 
     Boolean getHasRegularInvoice() {
-        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.INVOICE }
+        return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.PURCHASE_INVOICE }
```
awalkowiak, line 339: Same as above

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
```diff
@@ -102,6 +111,37 @@ class InvoiceController {
...
+        Location currentLocation = Location.get(session.warehouse.id)
+        if (!currentLocation) {
+            throw new IllegalArgumentException("User must be logged into a location to create invoice")
+        }
+
+        StockMovement stockMovement = stockMovementService.getStockMovement(params.id)
+        Shipment shipment = stockMovement.shipment
+        orderService.createOrderFromShipment(shipment, currentLocation)
+
+        jsonObject.put("party", shipment?.destination?.organization)
```
awalkowiak, line 126: What's the point of using jsonObject here? Remove it and pass into the bindSalesInvoiceData the shipment and just assign appropriate properties.

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
```diff
@@ -102,6 +111,37 @@ class InvoiceController {
...
+        JSONObject jsonObject = new JSONObject()
+
+        Location currentLocation = Location.get(session.warehouse.id)
+        if (!currentLocation) {
+            throw new IllegalArgumentException("User must be logged into a location to create invoice")
+        }
+
+        StockMovement stockMovement = stockMovementService.getStockMovement(params.id)
+        Shipment shipment = stockMovement.shipment
+        orderService.createOrderFromShipment(shipment, currentLocation)
```
awalkowiak, line 124: Was it required because, in practice, we need the order behind each invoce? Because technically I think it is possible to have invoice associated only to the shipment without order.

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/order/Order.groovy
MagdaToczek: Done

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
MagdaToczek: Done

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
MagdaToczek: Done

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
MagdaToczek: Right, it is not needed anymore.

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
MagdaToczek: I think technically it is possible, I would have to modify method invoiceService.createFromShipmentItem() because it uses orderItems. Writing it I was following dev notes from ticket.

jmiranda: @MagdaToczek @awalkowiak We should probably create new methods for each object class because I've noticed cases where the data being returned needs to be changed. For example, inbound stock movements aren't represented properly so the data comes back with things like quantityPicked which doesn't make sense in that context. I think this needs a tech huddle, but please feel free to create your own methods to return this data rather than trying to patch an existing method with conditions (i.e. if it's an invoice return this data, if it's an order return this data).

awalkowiak:
This one is obsolete

PR #2825, opened 2021-12-06
OBPIH-4153 Corrected displaying wrong qty to return/transfer after mo…

MagdaToczek:
…ving to hold bin already picked qty

### DIFF
--- a/src/groovy/org/pih/warehouse/api/StockTransferItem.groovy
+++ b/src/groovy/org/pih/warehouse/api/StockTransferItem.groovy
@@ -79,10 +79,11 @@ class StockTransferItem {
         stockTransferItem.destinationBinLocation = productAvailability.binLocation
         stockTransferItem.quantity = productAvailability.quantityOnHand
         stockTransferItem.quantityOnHand = productAvailability.quantityOnHand
-        stockTransferItem.quantityNotPicked = productAvailability.quantityNotPicked > 0 ? productAvailability.quantityNotPicked : 0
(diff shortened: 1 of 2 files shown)


src/groovy/org/pih/warehouse/api/StockTransferItem.groovy
```diff
@@ -79,7 +79,7 @@ class StockTransferItem {
         stockTransferItem.destinationBinLocation = productAvailability.binLocation
         stockTransferItem.quantity = productAvailability.quantityOnHand
         stockTransferItem.quantityOnHand = productAvailability.quantityOnHand
-        stockTransferItem.quantityNotPicked = productAvailability.quantityNotPicked > 0 ? productAvailability.quantityNotPicked : 0
+        stockTransferItem.quantityNotPicked = productAvailability.quantityNotPicked > 0 && productAvailability?.binLocation?.isPickable() ? productAvailability.quantityNotPicked : 0
```
pmuchowski, line 82: You should also check if item is recalled, 
add isPickable method to ProductAvailability, like it was done in AvailableItem

pmuchowski (changes requested):

src/groovy/org/pih/warehouse/api/StockTransferItem.groovy
MagdaToczek: Is it fine now?

awalkowiak (approved):

PR #2832, opened 2021-12-08
OBPIH-4153 Corrected displaying wrong qty to return/transfer after mo…

MagdaToczek:
…ving to hold bin already picked qty

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -18,6 +18,7 @@ import org.apache.commons.lang.StringEscapeUtils
 import org.codehaus.groovy.grails.commons.ConfigurationHolder
 import org.hibernate.Criteria
 import org.pih.warehouse.api.AvailableItem
+import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.ApplicationExceptionEvent
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Location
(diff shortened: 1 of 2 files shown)


src/groovy/org/pih/warehouse/api/StockTransferItem.groovy
```diff
@@ -67,6 +67,7 @@ class StockTransferItem {
 
     static StockTransferItem createFromProductAvailability(ProductAvailability productAvailability) {
         StockTransferItem stockTransferItem = new StockTransferItem()
+        def picklistItemQuantityPicked = PicklistItem.findByInventoryItem(productAvailability.inventoryItem)?.quantity ?: 0
```
awalkowiak, line 70: I am not sure if we should consider something as picked if it is in on hold or recalled.

awalkowiak (changes requested):
@MagdaToczek please do changes according to our internal tech huddle, notes from it are in the OBPIH-4153 ticket

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -166,8 +166,24 @@ class ProductAvailabilityService {
...
+                    }
+                }
+                requisitionItem {
+                    isNotNull("id")
+                }
+                inventoryItem {
+                    ne("lotStatus", LotStatusCode.RECALLED)
+                }
+                binLocation {
+                    locationType {
```
awalkowiak, line 185: I think this is a bit more tricky than that, because `HOLD_LOCATION` is `supportedActivity` instead of type (take a look at `Location.isOnHold()`

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -166,8 +166,24 @@ class ProductAvailabilityService {
...
+                    or {
+                        requisitionItem {
+                            eq("product", product)
+                        }
+                        orderItem {
+                            eq("product", product)
+                        }
+                    }
+                }
+                requisitionItem {
```
pmuchowski, line 178: this will remove all returns, because for those lines requisition item is null
I think it should be order item not null or (requisition item not null and not recalled and not on hold)

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: true

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -166,8 +167,31 @@ class ProductAvailabilityService {
...
+                        orderItem {
+                            eq("product", product)
+                        }
+                    }
+                }
+                or {
+                    orderItem {
+                        isNotNull("id")
+                    }
                     requisitionItem {
```
pmuchowski, line 183: I'm not sure if criteria works like that, I think the `requisitionItem` should be inside the `and`

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
MagdaToczek: I'll try to rewrite it as a query.

PR #2833, opened 2021-12-08
OBAM-293 Create stock movement view to use for list pages

pmuchowski:


grails-app/migrations/views/stock-movement.sql
```diff
@@ -0,0 +1,131 @@
...
+                   count(*) as item_count
+            from requisition_item
+            group by requisition_id
+            ) ri on ri.requisition_id = r.id
+
+        left join shipment_reference_number sr on s.id = sr.shipment_reference_numbers_id
+        left join reference_number rn on rn.id = sr.reference_number_id and rn.reference_number_type_id = "10"
+    where r.is_template is false
+
+union all
```
jmiranda, line 44: does this need to be union all?

grails-app/migrations/views/stock-movement.sql
```diff
@@ -0,0 +1,131 @@
...
+           r.origin_id,
+           r.destination_id,
+           r.date_requested,
+           r.date_created,
+           r.last_updated,
+           r.requested_by_id,
+           r.created_by_id,
+           r.updated_by_id,
+           ri.item_count,
+           'OUTBOUND' as stock_movement_type,
```
jmiranda, line 16: We probably want to change OUTBOUND and INBOUND to something that describes the stock movement type rather than the direction. I'll get back to you.

grails-app/migrations/views/stock-movement.sql
```diff
@@ -0,0 +1,131 @@
...
+           s.origin_id,
+           s.destination_id,
+           s.date_created as date_requested,
+           s.date_created,
+           s.last_updated,
+           s.created_by_id as requested_by_id,
+           s.created_by_id,
+           s.updated_by_id,
+           si.item_count,
+           'INBOUND' as stock_movement_type,
```
jmiranda, line 60: I assume removing / refactoring this field will allow us to remove this subquery all together since all of these stock movements are already represented in the list above.

jmiranda: @pmuchowski can you commend on that?

Basically what I mean is that we should just get a list of all stock movements and whether they are inbound or outbound is a determined once we know the current location. So I could pass the current location as origin or destination and that would provide enough context to determine what records should be returned.

grails-app/migrations/views/stock-movement.sql
pmuchowski: union is more expensive, because is doing ordering and removing duplicates

grails-app/migrations/views/stock-movement.sql
pmuchowski: @jmiranda I think we can do that, but when you add ` and s.requisition_id is null` to this query and run it on obdev there will be some results, I'm not sure how those shipments were created, but you can find it in the list inbound movement page when you search by identifier

grails-app/views/stockMovement/list.gsp
```diff
@@ -53,13 +53,13 @@
 
         <div class="right">
             <div class="button-container">
-                <g:if test="${params.direction as StockMovementType == StockMovementType.INBOUND}">
+                <g:if test="${params.direction as org.pih.warehouse.api.StockMovementDirection == StockMovementDirection.INBOUND}">
```
jmiranda, line 56: I know this isn't on you @pmuchowski but this definitely a place where we probably could / should have create a separate GSPs for the different stock movement types / directions.

grails-app/domain/org/pih/warehouse/stockMovement/StockMovement.groovy
```diff
@@ -264,14 +259,47 @@ class StockMovement {
...
     static StockMovement createFromShipment(Shipment shipment) {
         return createFromShipment(shipment, Boolean.TRUE)
     }
 
     static StockMovement createFromShipment(Shipment shipment, Boolean includeStockMovementItems) {
 
-        String statusCode = (shipment.status.code == ShipmentStatusCode.PENDING) ?
-                RequisitionStatus.PENDING.toString() : RequisitionStatus.ISSUED.toString()
+        StockMovementStatusCode statusCode = (shipment.status.code == ShipmentStatusCode.PENDING) ?
+                StockMovementStatusCode.PENDING : StockMovementStatusCode.ISSUED
```
jmiranda, line 302: Let's use DISPATCHED here since that's what the stock movement state machine uses.

src/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
```diff
@@ -27,6 +27,7 @@ enum StockMovementStatusCode {
     PACKED(7, PENDING),
     REVIEWING(8, PENDING),
     DISPATCHED(9),
+    ISSUED(9),
```
jmiranda, line 30: Let's just use DISPATCHED instead of adding this. I know it's a bit confusing but if we're going to change this back to ISSUED I'd like to do that as a separate refactoring.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/stockMovement/StockMovement.groovy
```diff
@@ -264,14 +259,47 @@ class StockMovement {
         return name
     }
 
+    void createLineItems() {
```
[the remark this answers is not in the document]
jmiranda: Use the same prefix as in the StockMovement class something like buildLineItems()

jmiranda (approved):

jmiranda:
@pmuchowski Good work!

awalkowiak (approved):

PR #2843, opened 2021-12-13
Revert "OBPIH-4153 Corrected displaying wrong qty to return/transfer after mo…"

awalkowiak:
Reverts openboxes/openboxes#2825


awalkowiak:
I am reverting this one on the develop because the scope of this and a fix to it slightly changed, and this is made on the release branch.

PR #2846, opened 2021-12-13
OBPIH-4199 Disabling outbound returns for release 0.8.17

MagdaToczek:


grails-app/i18n/messages.properties
```diff
@@ -1183,6 +1183,7 @@ inventory.stockTransfers.transferredTo=Transferred To
 inventory.stockTransfers.editCompleted=Cannot edit completed order
 
 stockTransfer.label=Stock Transfer
+stockTransfer.temporaryDisabled.message=Creating Outbound Returns was temporary disabled
```
awalkowiak, line 1186: Should be `outboundReturns.temporaryDisabled.message` instead of stock transfer

MagdaToczek: Done

awalkowiak (approved):

PR #2847, opened 2021-12-13
OBPIH-4202 Look into automated security scans.

mdpearson:
(Not sure how much value this will add until it runs.)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,96 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ develop, master, release/0.8.* ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ develop, master, release/0.8.* ]
+  schedule:
+    - cron: '0 0 * * 3'
```
jmiranda, line 21: I can't cron, can you provide a comment on that this means? Every tuesday at midnight?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,96 @@
...
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ develop, master, release/0.8.* ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ develop, master, release/0.8.* ]
```
jmiranda, line 19: I think we should just use the release prefix and not qualify it as this will need to be maintained when we move to the next minor/major version.

jmiranda (commented):
This looks fine to me (sans a few minor changes), but I don't really know what this does beyond maybe automating dependency checks (doesn't github already do this for us). What was the impetus for this? What problem does this solve? And I noticed that this maybe doesn't actually work until we support Java 8 so perhaps it makes more sense to apply to the grails upgrade branch or wait until we're ready to move forward with that branch. 

IMPORTANT: I want to avoid adding more things our current develop branch that we need to migrate when for the Grails 3 upgrade so let's be sure this solves an actual problem and doesn't just add more work for us.

.github/workflows/codeql-analysis.yml
mdpearson: very close: every Wednesday. not sure what timezone but probably UTC

.github/workflows/codeql-analysis.yml
mdpearson: Hi Justin, those are good questions. The impetus and goals of this PR were/are:

1. Document what I did when looking into the log4j and other critical security issues I was talking to Shawn about;
2. Look into tools that would look for security issues in our code (as opposed to our dependencies);
3. It seemed easier to show the OpenMRS team how these tools worked via a PR than via a meeting, although that may simply be because I don't like meetings.

As to where this should be targeted, I think the grails 3 branch would be fine. It's already found quite a bit of low-hanging fruit in the grails 1 branch (OBPIH-4215 through OBPIH-4218) so unless we're planning big dependency changes there, it doesn't need to run again there any time soon.

You mentioned "doesn't GitHub already do this for us?" Not being an admin on the OB repo, I don't have a full view of what GitHub does, but I believe these tools do quite a bit more than dependabot.

In terms of cost to port -- They shouldn't be dependent on any grails version, and, I believe, should work just fine on any java or javascript codebase as configured. I think the only time we'd need to edit them would be if/when we move to containers.

mdpearson:
Hi Justin, those are good questions. The impetus and goals of this PR were/are:

Document what I did when looking into the log4j and other critical security issues I was talking to Shawn about;
Look into tools that would look for security issues in our code (as opposed to our dependencies);
It seemed easier to show the OpenMRS team how these tools worked via a PR than via a meeting, although that may simply be because I don't like meetings.
As to where this should be targeted, I think the grails 3 branch would be fine. It's already found quite a bit of low-hanging fruit in the grails 1 branch (OBPIH-4215 through OBPIH-4218) so unless we're planning big dependency changes there, it doesn't need to run again there any time soon.

You mentioned "doesn't GitHub already do this for us?" Not being an admin on the OB repo, I don't have a full view of what GitHub does, but I believe these tools do quite a bit more than dependabot.

In terms of cost to port -- They shouldn't be dependent on any grails version, and, I believe, should work just fine on any java or javascript codebase as configured. I think the only time we'd need to edit them would be if/when we move to containers.

.github/workflows/codeql-analysis.yml
mdpearson: resolved in c0ab81674

.github/workflows/codeql-analysis.yml
mdpearson: resolved in c0ab81674

mogoodrich:
@mseaton  @mdpearson  we should discuss potentially adding some of these to our other repos...

jmiranda:
Adding this to icebox to make sure we review this in the future.

PR #2852, opened 2021-12-15
OBPIH-4197 Fixed possibility to change location for user without defa…

MagdaToczek:
…ult location


awalkowiak (approved):
Nice solution, didn't know that we can do this that way

PR #2858, opened 2021-12-15
OBPIH-4217 Update lodash to fix a critical security vulnerability.

mdpearson:
While looking into the log4j issue, I ran a few scanners that pointed out other security vulnerabilities. This and a few others (forthcoming) were straightforward remedies. I don't think this needs to go out in the next release, so I've targeted develop. Let me know if I should aim it elsewhere. Thanks!


jmiranda (commented):
I think we get these dependency checks for free from dependabot. 
https://github.com/openboxes/openboxes/security/dependabot

I think we need a plan for how/when we apply these. My assumption is that we've postponed applying these patches until the release was done. But I think we should add this as a spike / topic to be tech huddled and figure out an appropriate strategy for this. @mdpearson can you add this as a Discussion ticket for retrospective.

awalkowiak (approved):

PR #2861, opened 2021-12-15
OBPIH-4218 Update to a (more) secure fork of log4j.

mdpearson:
Turns out log4j v1 also has two critical vulnerabilities. Grails uses SLF4J under the hood, and the documentation says "many Apache projects have 'commons-logging' as a transitive dependency, but [it shouldn't be included in a Grails project](https://grails.github.io/grails2-doc/1.3.7/guide/single.html#3.7%20Dependency%20Resolution)". So, I sprinkled a few `excludes` directives to remove all mention of log4j from `grails dependency-report`.


jmiranda (approved):
@mdpearson I always assumed that SLF4J was just a wrapper around log4j. See the log4j DSL in the Config.groovy. Meaning we are still susceptible to those vulnerabilities. Is that not true?

mdpearson:
> @mdpearson I always assumed that SLF4J was just a wrapper around log4j. See the log4j DSL in the Config.groovy. Meaning we are still susceptible to those vulnerabilities. Is that not true?

Shoot, you're right. I conflated commons-logging and log4j. The former does have some security holes, but nothing as bad as the latter. In response I've updated the code to remove commons-logging, and to use the latest log4j we can.

mdpearson:
@jmiranda per your discussion on the ticket, this PR now removes all trace of both commons-logging and log4j (verified with `grails dependency-report && egrep -rc 'logging|log4j' target/dependency-report/*.html`). Every log statement by OB and its dependencies should now route, via slf4j, to the [reload4j](https://reload4j.qos.ch) library.

jmiranda (approved):

awalkowiak (approved):

PR #2863, opened 2021-12-16
OBPIH-4213 Disabled link to baseline QOH report

MagdaToczek:


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -167,6 +167,9 @@ class InventoryController {
      *
      */
     def show = {
+        // OBPIH-4213: TEMPORARILY DISABLED
+        throw new UnsupportedOperationException("${warehouse.message(code: 'inventory.temporaryDisabled.message')}")
```
awalkowiak, line 171: On a second thought, please ensure that this does not break anything, because I see that there are couple redirects to this action in this controller

pmuchowski (approved):

PR #2896, opened 2022-01-10
OBPIH-4207,4208 Add option to show only errored items on the edit page

awalkowiak:


src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -38,7 +38,14 @@ class TableBodyVirtualized extends Component {
       }
     } else {
       _.forEach(fields.value, (field) => {
+        const dynamicAttr = getDynamicRowAttr ?
+          getDynamicRowAttr({ ...properties, rowValues: field }) : {};
         const subfields = field[subfieldKey];
+
+        if (dynamicAttr.hideRow) {
+          return;
```
pmuchowski, line 46: I think you should return 0 here

src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -88,6 +99,16 @@ class TableBodyVirtualized extends Component {
     const { totalCount } = properties;
 
     if (fields.value[index]) {
+      const dynamicRowAttr = fieldsConfig.getDynamicRowAttr ?
+        fieldsConfig.getDynamicRowAttr({
+          ...properties,
+          rowValues: _.filter(fields.value, v => !v.hideRow)[index],
```
pmuchowski, line 105: Why are you filtering the fields here?

pmuchowski (changes requested):

src/js/components/form-elements/TableBodyVirtualized.jsx
awalkowiak: @pmuchowski this is lodash's forEach version of `continue`

src/js/components/form-elements/TableBodyVirtualized.jsx
awalkowiak: I think this is something, that I forgot to remove. I'll check it.

PR #2900, opened 2022-01-11
OBPIH-4153 Corrected values ATP in returns and validation of returnSh…

MagdaToczek:
…impentItem in shipmentService

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -155,7 +155,7 @@ class ProductAvailabilityService {
             SELECT
                 pli.binLocation,
                 pli.inventoryItem,
-                sum(pli.quantity)
+                sum(pli.quantity)*((count(distinct pli.requisitionItem) + count(distinct pli.orderItem) )/ count(*))
             FROM Picklist pl
             INNER JOIN pl.picklistItems pli
             LEFT JOIN pl.requisition r
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -757,9 +757,11 @@ class ShipmentService {
 
             def quantityAvailableToReturn
             if (withBinLocation) {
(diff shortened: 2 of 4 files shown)


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,7 +155,7 @@ class ProductAvailabilityService {
             SELECT
                 pli.binLocation,
                 pli.inventoryItem,
-                sum(pli.quantity)
+                sum(pli.quantity)*(count(distinct pl.requisition) + count(distinct pl.order) / count(*))
```
pmuchowski, line 158: Why you did not use `pli.requisitionItem` and `pli.orderItem`?
Also I think you should add brackets before dividing by the count(*)

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
MagdaToczek: I checked that with `pli.requisitionItem` and `pli.orderItem` it works too, so I changed it.

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -757,9 +757,9 @@ class ShipmentService {
 
             def quantityAvailableToReturn
             if (withBinLocation) {
-                quantityAvailableToReturn = productAvailabilityService.getQuantityNotPickedInBinLocation(shipmentItem.inventoryItem, shipmentItem.binLocation)
+                quantityAvailableToReturn = productAvailabilityService.getQuantityNotPickedInBinLocation(shipmentItem.inventoryItem, shipmentItem.binLocation) + shipmentItem.quantity
```
pmuchowski, line 760: I think you should use quantity picked, not quantity,
right now in the shipmentItem there is method to get picked quantity for the requisition based shipment,
you could add similar method for the order

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
MagdaToczek: Done

pmuchowski (approved):

PR #2903, opened 2022-01-12
OBAM-144 Added forecast report

MagdaToczek:


grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -692,9 +692,11 @@ class ReportController {
                 def qtyOnHand = items.qtyOnHand.sum() ?: 0
                 def qtyExpiring = productExpiry.collect {it.quantity_on_hand}.sum() ?: 0
                 def qtyAvailable = qtyOnOrder + qtyOnHand - qtyExpiring ?: 0
-                def avgDemand = productExpiry.collect {it.average_daily_demand}.size() > 0 ? productExpiry.collect {it.average_daily_demand}.get(0).setScale(1, RoundingMode.HALF_UP) : 0
+                def avgDemand = productExpiry.collect {it.average_daily_demand}.size() > 0 ? productExpiry.collect {it.average_daily_demand}.get(0).setScale(1, RoundingMode.HALF_UP)*30 : 0
                 def monthsOfStock = ((replenishmentPeriodDays.toInteger() + leadTimeDays.toInteger())/30).setScale(1, RoundingMode.HALF_UP) ?: 0
                 def qtyNeeded = avgDemand * monthsOfStock ?: 0
+                def qtyToOrder = BigDecimal.ZERO.max(qtyNeeded - qtyOnHand)
+                def unitPrice = Product.findByProductCode(items[0].productCode).pricePerUnit ?: 0
```
awalkowiak, line 699: You have `...{ productId, items ->...` in the collect, use that instead of `items[0].productCode`

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
MagdaToczek: corrected

awalkowiak (approved):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -688,6 +688,31 @@ class ReportService implements ApplicationContextAware {
         return data
     }
 
+    List getOnOrderData(String locationId, String productId) {
```
awalkowiak, line 691: Why this was moved out from join to separate query?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
MagdaToczek: It is separated because of wrong data returned. On order data should be returned for a product and in previous version it was multiplied.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -665,7 +665,7 @@ class ReportController {
         params.origin = origin.name
 
         // Export as XLS
-        if (params.format == "text/csv") {
+        if (params.format == "text/csv" && params.print) {
```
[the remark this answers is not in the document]
MagdaToczek: It is required for preventing of printing report while page is rendered.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
awalkowiak: Ok got it, but I am afraid that this solution will result in very poor performance. I'll make some further change requests.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -760,15 +779,10 @@ class ReportService implements ApplicationContextAware {
 
             def results = dataService.executeQuery(query, params)
             data = results.collect {
```
awalkowiak, line 781: If you had to pull out the join into a separate query to fix the data, then instead of calling the second method for each product separately in the ReportController while collecting data you could do it here:

1. Remove product code and product name from select here - return only product ids (you are pulling the product in the controller anyway).
2. Instead of doing the collect here, call the getOnOrderData here and pass the list of product ids into it. 
3. In the getOnOrderData instead of doing `where product_id = :productId` do `where product_id in (:productIds)`.
4. Add a product id into the selected data in the getOnOrderData and return it here.
5. Do some small adjustments in the report controller according to the points above.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
MagdaToczek: ok, thanks.

MagdaToczek: done

awalkowiak (approved):
This looks way better now. Thanks.

PR #2909, opened 2022-01-16
OBDX-26 Use external product ID as mapping key between systems

chetanmaharishi:
OBDX-26 Use external product ID as mapping key between systems

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -1228,15 +1228,40 @@ class ProductService {
         return productInstance
     }
 
+    Attribute findExternalProductIdAttribute(String externalAttributeCode = null){
+        if (!externalAttributeCode) {
+            externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
+        }
+        if (!externalAttributeCode) {
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
+        Attribute attribute = Attribute.findByCode(externalAttributeCode)
+        return attribute
+    }
+
+    Product findProductByExternalId(String externalProductId){
+        String externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
+        if(!externalAttributeCode){
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
+        Attribute attribute = findExternalProductIdAttribute(externalAttributeCode)
+        if(!attribute){
+            attribute = new Attribute(code: externalAttributeCode, name: externalAttributeCode, allowOther: false)
+            attribute.save(flush: true)
+        }
+        ProductAttribute productAttribute = ProductAttribute.findByAttributeAndValue(attribute, externalProductId)
+        return productAttribute?.product
+    }
+
     Product createFromJson(JSONObject productJson){
         log.info "Product: ${productJson.toString(4)}"
 
-        String productId = productJson.getString("id")
-        Product productInstance = findProduct(productId)
+        String externalProductId = productJson.getString("id")
+        Product productInstance = findProductByExternalId(externalProductId)
+        Boolean isNewInstance = false
         if (!productInstance) {
             productInstance = new Product()
-            // FIXME Hack to handle external product ID from external system (see OBDX-26 for proposed solution)
-            productInstance.upc = productId
+            isNewInstance = true
         }
         productInstance.name = productJson.getString("name")
         productInstance.productCode = productJson.getString("productCode")
@@ -1255,6 +1280,7 @@ class ProductService {
             productInstance.productCode = generateProductIdentifier(productInstance.productType)
         }
         productInstance.save(flush: true)
+
 
         // Process attributes
         if(productJson.containsKey("attributes")){
@@ -1284,6 +1310,14 @@ class ProductService {
                 }
             }
         }
+        if(isNewInstance){ // Saving product id in product attribute
+            Attribute attribute = findExternalProductIdAttribute()
+            ProductAttribute productAttribute = new ProductAttribute(attribute: attribute, value: externalProductId)
+            productInstance.addToAttributes(productAttribute)
+            productInstance.save(flush: true, failOnError: true)
+        }
+
+
 
         // Process documents
         if (productJson.containsKey("documents")) {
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
     }
 
+    Attribute findExternalProductIdAttribute(String externalAttributeCode = null){
+        if (!externalAttributeCode) {
+            externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
+        }
+        if (!externalAttributeCode) {
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
+        Attribute attribute = Attribute.findByCode(externalAttributeCode)
```
jmiranda, line 1238: if we're not doing any special error handling let's just return from the query
```
return Attribute.findByCode(externalAttributeCode)
```

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
         return productInstance
     }
 
+    Attribute findExternalProductIdAttribute(String externalAttributeCode = null){
+        if (!externalAttributeCode) {
+            externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
```
jmiranda, line 1233: There should be a default setting in Config.groovy 
```
openboxes.product.externalProductId.attributeCode = "EXTERNAL_PRODUCT_ID"
```

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
+        }
+        if (!externalAttributeCode) {
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
```
jmiranda, line 1237: Remove 1235-1237. If we did want to do something like this in the code it should 
```
externalAttributeCode = grailsApplication.config.openboxes.product.externalProductId.attributeCode?:"EXTERNAL_PRODUCT_ID"
```
But we don't want that now since we'll have it configured in the Config.groovy or openboxes-config.properties or openboxes-config.groovy

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
+        Attribute attribute = Attribute.findByCode(externalAttributeCode)
+        return attribute
+    }
+
+    Product findProductByExternalId(String externalProductId){
+        String externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
```
jmiranda, line 1243: Same 
```
String externalAttributeCode = grailsApplication.config.openboxes.product.externalProductId.attributeCode
```

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
+        if(!externalAttributeCode){
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
```
jmiranda, line 1246: Remove these three lines

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
+            externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
+        }
+        if (!externalAttributeCode) {
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
+        Attribute attribute = Attribute.findByCode(externalAttributeCode)
+        return attribute
+    }
+
+    Product findProductByExternalId(String externalProductId){
```
jmiranda, line 1242: Make sure we check to see if this is null or empty. My guess is that we should throw an IllegalArgumentException?

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1228,15 +1228,40 @@ class ProductService {
...
+        String externalAttributeCode = grailsApplication.config.externalProductId.attribute.code
+        if(!externalAttributeCode){
+            externalAttributeCode = "EXTERNAL_PRODUCT_ID"
+        }
+        Attribute attribute = findExternalProductIdAttribute(externalAttributeCode)
+        if(!attribute){
+            attribute = new Attribute(code: externalAttributeCode, name: externalAttributeCode, allowOther: false)
+            attribute.save(flush: true)
+        }
+        ProductAttribute productAttribute = ProductAttribute.findByAttributeAndValue(attribute, externalProductId)
```
jmiranda, line 1252: This is fine for now, but what do you think we should do if we find more than one?

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1284,6 +1310,14 @@ class ProductService {
                 }
             }
         }
+        if(isNewInstance){ // Saving product id in product attribute
+            Attribute attribute = findExternalProductIdAttribute()
```
jmiranda, line 1314: Let's add this as a static method to Attribute if we can.
```
Attribute.getExternalProductIdAttribute()
```

test/integration/org/pih/warehouse/product/ProductNotificationTests.groovy
```diff
@@ -8,13 +8,24 @@ class ProductNotificationTests  extends GroovyTestCase {
 
     def productService
 
+    @Test
+    void testProductFindWithExternalId(){
+
+
```
jmiranda, line 14: ?

jmiranda (changes requested):
Looks pretty good. A few suggestions and maybe some things we need to discuss with Mihail during our call tomorrow.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
chetanmaharishi: this will return only one result. so if we have more than that is error, we may did something wrong. that why external id will be duplicate.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
chetanmaharishi: if we add static then we cannot use grailsApplication service.

PR #2915, opened 2022-01-18
OBDS-77 Added validations to location import

MagdaToczek:


grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,32 +31,44 @@ class LocationDataService {
 
         command.data.eachWithIndex { params, index ->
 
-            Location location = createOrUpdateLocation(params)
-
+            Location newLocation = Location.findByName(params.name)
+            Location location = new Location()
```
jmiranda, line 35: Is this just here to carry the errors?

jmiranda: I think we can just use the command object for that unless you are rejecting specific fields.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,32 +31,44 @@ class LocationDataService {
...
+            if (!locationType) {
+                location.errors.reject("Row ${index + 1}: Location Type '${params.locationType}' does not exist")
             }
 
-            if (!location.validate()) {
-                location.errors.each { BeanPropertyBindingResult error ->
-                    String errorMessage = messageSource.getMessage(error.fieldError, locale)
-                    command.errors.reject("Row ${index + 1}: ${errorMessage}")
-                }
+            if (locationType?.locationTypeCode in [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION, LocationTypeCode.ZONE] && !parentLocation) {
```
jmiranda, line 56: Let's move the list of LocationTypeCodes into a static method on LocationType
```
LocationType.getInternalLocationTypeCodes() { return [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION, LocationTypeCode.ZONE] }
LocationType.getFacilityLocationTypeCodes() { ... }
```
And/or something more DRY like a method a boolean method that returns whether the location is an internal location or facility level location
```
Boolean isInternalLocation() { return locationTypeCode in LocationType.getInternalLocationTypeCodes() }
Boolean isFacilityLocation() { return locationTypeCode in LocationType.getFacilityLocationTypeCodes() } 
```
to prevent us from having to copy that array all over the place.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,32 +31,44 @@ class LocationDataService {
...
-                location?.organization?.errors?.each { BeanPropertyBindingResult error ->
-                    String errorMessage = messageSource.getMessage(error.fieldError, locale)
-                    command.errors.reject("Row ${index + 1}: ${errorMessage}")
-                }
+            if (!(locationType?.locationTypeCode in [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION, LocationTypeCode.ZONE]) && parentLocation) {
+                location.errors.reject("Row ${index + 1}: Cannot assign a parent location to a location of type '${params.locationType}'")
             }
 
+            location.errors.getAllErrors().each {
+                command.errors.reject(it.getCode())
```
jmiranda, line 65: Oh now that I see this, that confirms that we should be using the command throughout.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,32 +31,44 @@ class LocationDataService {
...
+            }
         }
     }
 
     void importData(ImportDataCommand command) {
-        command.data.eachWithIndex { params, index ->
+        command.data.unique().eachWithIndex { params, index ->
```
jmiranda, line 71: Can you explain the need for this? If someone has accidentally included the same row twice it feels like we want to warn them about that because there's probably a reason behind the data error. But maybe I'm missing something. In any case, let's also add a comment since it's not clear what's happening here.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: I just wonna to prevent errors while importing, because I thought that all voulnerabilities should be corrected during the validation, but you are right that in imported data could be two same rows that are not pointed during validation.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: Sure, gonna correct it.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: You are right, it was useful in my previous approach but finally it is not necessary.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: right

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -30,30 +28,42 @@ class LocationDataService {
     Boolean validateData(ImportDataCommand command) {
 
         Locale locale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale)
+        def locationNamesToImport = command.data.collect {it.name}
 
         command.data.eachWithIndex { params, index ->
 
-            Location location = createOrUpdateLocation(params)
-
+            Location newLocation = Location.findByName(params.name)
```
awalkowiak, line 35: I think there is no need for this assignment, and the naming is confusing too. Just use the `Location.findByName(params.name)` in the if condition, where you check if it exists.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -91,4 +101,8 @@ class LocationDataService {
         }
         return location
     }
+
+    Boolean isInternalOrZoneLocation(LocationTypeCode locationTypeCode) {
```
awalkowiak, line 105: I'd prefer here separate `isInternal` and `isZone`, on top of that these should be a methods in the LocationType. And then call `locationType?.isIntenal() || locationType?.isZone()` in the `if` condition.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: Done

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: Done

PR #2924, opened 2022-01-22
OBDS-73 Decorate location, product and user for Swagger.

mdpearson:
Artur, I'm tagging you on this PR since we had a brief chat about it during a standup.

No rush on getting this merged since nothing depends on it, I've already published the spec it generates, and it's unclear whether we want to build on this work in the near term. Still, I thought you both might like seeing the sorts of things Swagger looks for and picks up. (You can imagine in a JAX-RS application the swagger-ification going pretty seamlessly. For something like OB which rolls its own API, it's a little less magical -- I'm basically using JAX-RS annotations as glorified structured comments, here.)


jmiranda (approved):

jmiranda:
@mdpearson in case I haven't mentioned it before i intend to switch to a less "roll your own" approach to our rest api once we get Grails 3 migration finished. For things like products it might not be that difficult but stock movements is going to be a heavy lift given all of the RPC style endpoints. So keep that in mind as we devise a plan.

PR #2927, opened 2022-01-24
OBPIH-4153 Fixed QATP for recalled and on hold bin in stock transfers

MagdaToczek:


grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -78,7 +78,7 @@ class PicklistService {
             picklist.save()
         }
 
-        productAvailabilityService.refreshProductsAvailability(orderItem?.order?.origin?.id, [orderItem?.product?.id], false)
+        productAvailabilityService.refreshProductsAvailability(orderItem?.order?.origin?.id, null, false)
```
awalkowiak, line 81: Instead of refreshing product availability for all products within a location, you could take products from the picklist before clearing.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
MagdaToczek: I've added refreshing while deleting items from picklist

awalkowiak (approved):

PR #2929, opened 2022-01-24
T4. Product availability integration

chetanmaharishi:
T4. Product availability integration

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -17,6 +17,7 @@ import org.codehaus.groovy.grails.web.json.JSONArray
 import org.pih.warehouse.core.ApiException
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Document
+import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Tag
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1358,4 +1361,37 @@ class ProductService {
...
+        products?.each {Product product ->
+            JSONObject jsonObject = new JSONObject()
+            String externalId = product.attributes.find{it.attribute == externalIdAttribute}.value
+            List productAvailableItem = availableItems.findAll {it.product == product}
+            jsonObject.put("productId", externalId)
+            jsonObject.put("productCode", product?.productCode)
+            jsonObject.put("locationNumber", location?.locationNumber)
+            jsonObject.put("quantityOnHand", quantityMap[product])
+            jsonObject.put("quantityAvailable", quantityAvailableMap[product])
+            jsonObject.put("earliestExpirationDate", productAvailableItem?.min {it?.inventoryItem?.expirationDate})
```
jmiranda, line 1389: We should be able to get quantity on hand, quantity available, and min expiration date in a single query. We can leave this for now, but I'll try to refactor when I get a chance.

jmiranda (approved):
We should be able to do m

PR #2933, opened 2022-01-25
OBAM-118 Added request from ward w/stocklist - data entry for inventory

MagdaToczek:


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -514,12 +655,23 @@ class AddItemsPage extends Component {
     }
   }
 
+  getIsRequestFromWard() {
+    this.setState({
+      isRequestFromWard: this.props.currentLocationId === this.state.values.destination.id && this.state.values.destination.type === 'WARD',
```
awalkowiak, line 660: This should be done in the `fetchAddItemsPageData` instead of in a separate method like this.

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
MagdaToczek: done

awalkowiak (approved):

pmuchowski (approved):

PR #2935, opened 2022-01-26
OBPIH-4218 Fix logging includes.

mdpearson:
Oops! The previous PR removed a log that was in use, and kept one that was a no-op.

Testing: I checked that this change restores the liquibase changelog files via `jar tf`, and verified that the resultant warfile launches successfully with Tomcat 7 on my machine.


awalkowiak (approved):

awalkowiak:
@mdpearson thanks for the fix. I did not notice it at first, and I reverted the previous one, but I reverted the revert (by force pushing :hand_over_mouth:) and merged this one. Unfortunately, the bamboo is pausing, so we have to wait for it to come back, to rebuild the obdev.

PR #2940, opened 2022-01-26
OBDX-9 T5. Add a flag stating "Sync with Stocks-Up" to the products

chetanmaharishi:
T5. Add a flag stating "Sync with Stocks-Up" to the products. In this way the WMS can be also used for products that have nothing to do with Stocks-Up and we'll be able to distinguish between them.


jmiranda:
I think we might have to go in a different direction, so I'm going to close this PR. Reassign the ticket to me and I'll try to come up with a solution over the weekend.

PR #2956, opened 2022-02-03
OBDS-79 Enabled location edit via import

MagdaToczek:


grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -112,15 +123,23 @@ class LocationDataService {
             location = new Location()
         }
 
+        def locationTypeToBeChanged = params.locationType ? LocationType.findByNameLike(params.locationType + "%") : null
```
awalkowiak, line 126: Can you name it simply `locationType`?

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -112,15 +123,23 @@ class LocationDataService {
...
         location.locationGroup = params.locationGroup ? LocationGroup.findByName(params.locationGroup) : null
         location.parentLocation = params.parentLocation ? Location.findByNameOrLocationNumber(params.parentLocation, params.parentLocation) : null
         location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
         location.address = createOrUpdateAddress(params, location?.address?.id)
 
+        //Do not allow to change internal type location to non-internal
+        if (((currentLocationType?.isInternalLocation() || currentLocationType?.isZone()) && (locationTypeToBeChanged?.isInternalLocation() || locationTypeToBeChanged?.isZone())) || !(currentLocationType?.isInternalLocation() || currentLocationType?.isZone())) {
+            location.locationType = locationTypeToBeChanged
+            location.supportedActivities = currentLocationType && locationTypeToBeChanged.equals(currentLocationType) ? location.supportedActivities : new HashSet(location.locationType.getSupportedActivities())
```
awalkowiak, line 140: If you want to add something to the supported activities, you should do `location.addToSupportedActivities(...)`, but is there a need to add something to the location's supported activities if it's already on the location type?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,12 +33,17 @@ class LocationDataService {
         def locationNumbersToImport = command.data.collect {it.locationNumber}
 
         command.data.eachWithIndex { params, index ->
+            Location locationToBeUpdated = params.id ? Location.findById(params.id) : null
```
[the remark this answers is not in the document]
MagdaToczek: yes we can :)

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,12 +33,17 @@ class LocationDataService {
...
             LocationGroup locationGroup = params.locationGroup ? LocationGroup.findByName(params.locationGroup) : null
             LocationType locationType = params.locationType ? LocationType.findByName(params.locationType) : null
             List<Organization> organizations = params.organization ? Organization.findAllByCodeOrName(params.organization, params.organization) : null
 
-            if (Location.findByName(params.name)) {
+            if (params.id && !locationToBeUpdated) {
+                command.errors.reject("Row ${index + 1}: Id '${params.id}' do not exists in the system.")
+            }
+
+            if (Location.findByName(params.name) && !(params.id && params.name.equals(locationFromGivenId?.name))) {
```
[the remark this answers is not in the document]
MagdaToczek: It should be locationToBeUpdated, I refactored that name and somehow it wasn't changed... Now it is just location.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -89,6 +94,10 @@ class LocationDataService {
             if (organizations?.size() > 1) {
                 command.errors.reject("Row ${index + 1}: '${organizations.size()}' records found for organization name '${params.organization}'. Please specify by entering the organization code instead")
             }
+
+            if (params.locationType && locationToBeUpdated && (locationToBeUpdated.locationType.isInternalLocation() || locationToBeUpdated.locationType.isZone()) && !(LocationType.findByNameLike(params.locationType + "%").isInternalLocation() || LocationType.findByNameLike(params.locationType + "%").isZone())) {
```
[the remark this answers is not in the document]
MagdaToczek: It is because here I'm checking if type of existing location isn't changed from internal to non-internal location type, so I have to check if existing location is internal and then check if new location type is also internal.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -103,7 +112,9 @@ class LocationDataService {
 
     Location createOrUpdateLocation(Map params) {
         Location location
-        if (params.locationNumber) {
+        if (params.id) {
+            location = Location.findById(params.id)
+        } else if (params.locationNumber) {
             location = Location.findByNameOrLocationNumber(params.name, params.locationNumber)
         } else {
             location = Location.findByName(params.name)
```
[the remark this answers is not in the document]
MagdaToczek: It shouldn't be null, because name is required field and in that place name should be already validated.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: sure

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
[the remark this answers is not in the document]
awalkowiak: Yeah, but instead of this `LocationType.findByNameLike(params.locationType + "%")` can't you use the `locationType` already defined? And why there is looking by name with wildcard `%`?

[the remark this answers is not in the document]
MagdaToczek: It was already done before, maybe it is because values are like "Supplier|fr:Fournisseurs" or "Pharmacy|fr:Pharmacy".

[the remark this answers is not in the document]
awalkowiak: @MagdaToczek Makes sense, but then should not you look for existing location type with the same way?

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: It occures that using `location.addToSupportedActivities(...)` is not possible because of lazy initialization

jmiranda: That probably just means that the location has validation errors (i.e. location.validate() is false) before we add to supported activities. But I would be really careful about changing the supported activities here. Is this really a requirement? I think only superusers should be able to override the supported activities of the location. It feels like the LocationType import should be responsible for that.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -18,6 +18,7 @@ import org.pih.warehouse.core.Organization
 import org.pih.warehouse.importer.ImportDataCommand
 
 import javax.annotation.Nullable
+import javax.xml.bind.ValidationException
```
jmiranda, line 21: wrong validation exception. should be grails validation exception.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -33,12 +34,17 @@ class LocationDataService {
...
             LocationGroup locationGroup = params.locationGroup ? LocationGroup.findByName(params.locationGroup) : null
             LocationType locationType = params.locationType ? LocationType.findByName(params.locationType) : null
             List<Organization> organizations = params.organization ? Organization.findAllByCodeOrName(params.organization, params.organization) : null
 
-            if (Location.findByName(params.name)) {
+            if (params.id && !location) {
+                command.errors.reject("Row ${index + 1}: Id '${params.id}' do not exists in the system.")
+            }
+
+            if (Location.findByName(params.name) && !(params.id && params.name == location?.name)) {
```
jmiranda, line 47: Holy smokes. Let's add a comment here to explain what's going on :)

jmiranda: I would imagine that we also need to validate that we cannot create a location with the same location number as another location. There's a unique constraint on locationNumber but this doesn't work when parentLocation is null. So we probably need to create our own custom validation.

Also I don't know if I like the idea of using the id in the import / export. I think the location number should be our unique key. The one case I'm thinking of is someone uploading a spreadsheet with new locations, then immediately making an edit to that spreadsheet and uploading it again. Good UX would to allow the second import attempt because we have populated the location number and that's enough to qualify uniqueness.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -50,7 +56,7 @@ class LocationDataService {
                 command.errors.reject("Row ${index + 1}: Active field is incorrectly filled. Please fill with TRUE or FALSE")
             }
 
-            if (params.locationNumber && Location.findByLocationNumber(params.locationNumber)) {
+            if (params.locationNumber && Location.findByLocationNumber(params.locationNumber) && !(params.id && params.locationNumber == location?.locationNumber)) {
```
jmiranda, line 59: Ok here's the location number part. Can you add a comment to explain what's happening?

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -112,15 +124,23 @@ class LocationDataService {
             location = new Location()
         }
 
+        def locationType = params.locationType ? LocationType.findByNameLike(params.locationType + "%") : null
+        def currentLocationType = location.locationType
+
```
jmiranda, line 129: Move this down to line 138 so the locationtype so we can keep it together with the actual processing.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -307,6 +307,7 @@ class Location implements Comparable<Location>, java.io.Serializable {
     }
 
     static PROPERTIES = [
+            "id"              : "id",
```
jmiranda, line 310: I don't love that we're adding ID here. We'd need good justification for including this as I think locationNumber should be used to ensure discoverability and uniqueness.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -89,6 +95,10 @@ class LocationDataService {
             if (organizations?.size() > 1) {
                 command.errors.reject("Row ${index + 1}: '${organizations.size()}' records found for organization name '${params.organization}'. Please specify by entering the organization code instead")
             }
+
+            if (params.locationType && location && (location.locationType.isInternalLocation() || location.locationType.isZone()) && !(LocationType.findByNameLike(params.locationType + "%").isInternalLocation() || LocationType.findByNameLike(params.locationType + "%").isZone())) {
```
jmiranda, line 99: Add a comment for readability.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -103,7 +113,9 @@ class LocationDataService {
 
     Location createOrUpdateLocation(Map params) {
         Location location
-        if (params.locationNumber) {
+        if (params.id) {
```
jmiranda, line 116: Again I don't love adding ID to the import unless it's absolutely necessary.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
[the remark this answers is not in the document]
MagdaToczek: Ok, I'll change it in existing location.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: I've got validation for location number a bit lower.

Adding id field was specified in a ticket. Right now while there is no id and location number exists in a system than validation will not let to create new location.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: Ok, I'll add a comment.

grails-app/domain/org/pih/warehouse/core/Location.groovy
MagdaToczek: Maybe I'll wait to tech huddle with this. It was added because it was specifically demanded in a ticket.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: OK

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: OK

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
MagdaToczek: Removed, since it is not used.

grails-app/domain/org/pih/warehouse/core/Location.groovy
jmiranda: Sounds good. In the meantime, I'll discuss with it Kelsey.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -115,12 +129,20 @@ class LocationDataService {
...
         location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
         location.address = createOrUpdateAddress(params, location?.address?.id)
 
+        def locationType = params.locationType ? LocationType.findByNameLike(params.locationType + "%") : null
+        def currentLocationType = location.locationType
+
+        //Do not allow to change internal type location to non-internal
+        if (((currentLocationType?.isInternalLocation() || currentLocationType?.isZone()) && (locationType?.isInternalLocation() || locationType?.isZone())) || !(currentLocationType?.isInternalLocation() || currentLocationType?.isZone())) {
+            location.locationType = locationType
+            location.supportedActivities = currentLocationType && locationType == currentLocationType ? location.supportedActivities : new HashSet(location.locationType.getSupportedActivities())
```
jmiranda, line 143: If we change location types then we should only clear the Location's supported activities. In other words, do not add the location type's supported activities because the location will inherit these by default if it's supported actitivies association is empty.

awalkowiak (approved):

jmiranda (approved):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -115,12 +130,20 @@ class LocationDataService {
...
         location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
         location.address = createOrUpdateAddress(params, location?.address?.id)
 
+        def locationType = params.locationType ? LocationType.findByNameLike(params.locationType + "%") : null
+        def currentLocationType = location.locationType
+
+        //Do not allow to change internal type location to non-internal
+        if (((currentLocationType?.isInternalLocation() || currentLocationType?.isZone()) && (locationType?.isInternalLocation() || locationType?.isZone())) || !(currentLocationType?.isInternalLocation() || currentLocationType?.isZone())) {
+            location.locationType = locationType
+            location.supportedActivities = currentLocationType && locationType == currentLocationType ? location.supportedActivities : location.supportedActivities?.clear()
```
jmiranda, line 144: @MagdaToczek This makes me a little nervous (the assignment of the supported activities in the default case might mark the association as "dirty" and we'd like to avoid extra db activity if possible given Hibernate's penchant for lazy init errors and the like). I would just add some logic like this.
```
if (currentLocationType && locationType != currentLocationType) { 
    location.supportedActivities?.clear()
}
```
With that said, we can leave this for now. Just keep an eye on this during testing.

jmiranda (approved):

PR #2965, opened 2022-02-07
OBPIH-4250 fix `from` email address on dev instance + OBS-1053 don't crash if smtp server is unavailable

mdpearson:
I'm pretty sure the dev failures were caused by invalid email addresses in the test database. "Miss Administrator" and "Mister Manager" now have `openboxes@pih.org` as their email address.

While I was at it, I consolidated disparate implementations in MailService, removing 150-200 lines of code, fixed a broken test, and added tests for what happens if an SMTP server is down (no exception is thrown) (OBS-1053). I also updated the tools we use to send and test email.

GitHub is showing a sea of red ink which is making review difficult. I preserved fifteen(!?) existing function signatures, removing only one that took a map as input. (This one was used only in AdminController and ErrorsController and was easy to refactor.) The rest all route to a new function called `sendMailImpl()`. The previous implementations differed somewhat in logging and error behavior and I thought it would be less of a support headache going forward if there were just one way to do it.

### DIFF
--- a/grails-app/conf/BuildConfig.groovy
+++ b/grails-app/conf/BuildConfig.groovy
@@ -72,12 +72,11 @@ grails.project.dependency.resolution = {
         // Required for barcode4j
         compile 'com.google.zxing:javase:2.0'
 
-        // Required by MailService (should replace with grails mail plugin)
-        compile 'org.apache.commons:commons-email:1.3'
-        compile 'org.apache.commons:commons-text:1.3'
+        compile "org.apache.commons:commons-email:1.5"
+        compile "org.apache.commons:commons-text:1.3"  // last Java 7-compatible release
         compile 'commons-lang:commons-lang:2.6'
         compile "org.jadira.usertype:usertype.jodatime:1.9"
-        compile 'org.apache.commons:commons-csv:1.6'
+        compile "org.apache.commons:commons-csv:1.6"  // last Java 7-compatible release
 
         // Required by LDAP
         compile "com.unboundid:unboundid-ldapsdk:2.3.6"
@@ -118,8 +117,11 @@ grails.project.dependency.resolution = {
         compile ("fr.opensagres.xdocreport:org.odftoolkit.odfdom.converter.pdf:1.0.6")
         compile "org.apache.xmlgraphics:batik-util:1.7"
 
-        // Fake SMTP server
-        test 'dumbster:dumbster:1.6'
+        /*
+         * This test SMTP client is the latest release that works with Grails 1,
+         * and Java 7, although it depends on a junit release we can't use (yet).
+         */
+        test("com.icegreen:greenmail:1.5.10") { excludes "junit" }
 
         // Required for GPars
         compile "org.codehaus.gpars:gpars:0.12"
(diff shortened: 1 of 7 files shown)


grails-app/services/org/pih/warehouse/core/MailService.groovy
```diff
@@ -53,353 +53,157 @@ class MailService {
...
-                // Override from user
-                if (fromUser) {
-                    email.setFrom(fromUser.email, fromUser.name)
-                }
-
-                // Authenticate
-                if (username && password) {
-                    email.setAuthentication(username, password)
-                }
+    Boolean sendMailImpl(
```
jmiranda, line 138: Minor nitpick but could we change this to. 
```
Boolean doSendMail()
```
I like that convention better than Impl since Impl is usually used for the a class implementing an interface, like
```
BasicMailServiceImpl implements MailService
```

mdpearson: hahaha not a nitpick at all! I originally called this method `actuallySendTheActualMail()` and tried to rename it to be more professional. `doSendMail()` is great. Fixed in ff47dc359.

grails-app/conf/BuildConfig.groovy
```diff
@@ -118,8 +117,11 @@ grails.project.dependency.resolution = {
         compile ("fr.opensagres.xdocreport:org.odftoolkit.odfdom.converter.pdf:1.0.6")
         compile "org.apache.xmlgraphics:batik-util:1.7"
 
-        // Fake SMTP server
-        test 'dumbster:dumbster:1.6'
+        /*
+         * This test SMTP client is the latest release that works with Grails 1,
+         * and Java 7, although it depends on a junit release we can't use (yet).
+         */
+        test("com.icegreen:greenmail:1.5.10") { excludes "junit" }
```
jmiranda, line 124: Was there a reason we changed the dependency?

jmiranda (approved):
Everything looks good. We should test whether there's a difference in behavior based on the exception that is being thrown (runtime vs checked). The apache email library has methods that throw runtime exception (e.g. IllegalStateException, IllegalArgumentException) and checked exception (e.g. EmailException). We should test to see if this code suppresses both exception types.

Also, we should consider looking into making email more asynchronous. I've started to do that using spring events i.e. emitting events that send emails rather than sending emails directly from the service method. See https://github.com/robfletcher/grails-spring-events.

grails-app/conf/BuildConfig.groovy
mdpearson: 1. dumbster doesn't seem to work in JDK 1.8 (it's commented out in obgm3)
2. dumbster hasn't seen development for six years so it probably never will
3. greenmail has a straightforward upgrade path to JDK 1.8 for grails 3+ migration
4. greenmail's better documented and (imho) has a cleaner API

mdpearson:
@jmiranda I've added tests for invalid credentials, quota violations and invalid port. These cover checked and unchecked exceptions. I couldn't find a non-disgusting way to get MailService to throw an `IllegalStateException`, but I will note that the exception handling in `doSendMail()` casts a wide net (it catches `Exception`) and, should something funky happen, it'll log the error to Sentry.

I like your idea of doing work more asynchronously, but I think that's a separate ticket, and am not sure I'd want to build it on something that's so specific to Grails 1–2 and isn't under development anymore. I'll copy bits of our conversation to a new ticket, perhaps to address once we're on a more modern stack. Sound good?

mdpearson:
Hi Justin, I filed a ticket (https://pihemr.atlassian.net/browse/OBPIH-4303) to cover making events for emails. Any other hurdles to merging the rest of this ticket as-is?

PR #2972, opened 2022-02-10
OBPIH-4284 Simplify view by removing tracking number and line item count.

mdpearson:
…unt.


awalkowiak (commented):
@mdpearson I think this is ok for the stock movement list page, but for the view page might be required. I am wondering if we should split it into two sql views.

PR #2976, opened 2022-02-11
OBPIH-4252 Adjusted location chooser to display same locations on gra…

MagdaToczek:
…ils and react pages


awalkowiak (changes requested):
You don't need the `getDefaultRoles` and `hasDefaultRoleRequestor`

awalkowiak (approved):

PR #2985, opened 2022-02-17
OBS-1077 Resolve a null.toImport() regression.

mdpearson:
I introduced this bug as part of OBPIH-4058 (Improve logging on PO CSV parse errors). When I edited this block, I inadvertently removed an `&&` that, by short-circuiting, served as a null guard.


jmiranda (approved):
Looks ok to me

PR #2986, opened 2022-02-18
OBS-1076 Resolve an assertion failure on batch import with consistent use of withNewSession().

mdpearson:
https://obnav.pih-emr.org/openboxes/batch/importData will generate a strange warning,
'collection [org.pih.warehouse.core.User.locationRoles] was not processed by flush()',
if you try to upload a ProductSupplier template with more than one new entry in it.
~~The long, winding stack trace falls within 2 lines of a save() call without flush=true,
and setting that parameter allows the import to succeed.~~

~~A possible concern of this change is that the resultant import is very slow (about 30
minutes for a \~4000-line .xls file: only two records per second). I'm not sure if the
flush parameter affects import speed negatively, but I'd rather have correct and slow
behavior than interrupt our colleagues' work.~~

**Update** I removed explicit calls to `flush()` and wrapped the `ProductSupplier`
instantiation within a `withNewSession` block. More details on the ticket, but this
remedy is much more performant, running in about 5 minutes on my local machine, 
of which 2 are spent in `validate()`.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -472,20 +472,14 @@ class ProductController {
 
         if (!packageInstance) {
             packageInstance = new ProductPackage(params)
-            ProductPrice productPrice = new ProductPrice()
-            productPrice.price = parsedUnitPrice?:0
-            packageInstance.productPrice = productPrice
             productInstance.addToPackages(packageInstance)
         } else {
             packageInstance.properties = params
-            if (packageInstance.productPrice) {
-                packageInstance.productPrice.price = parsedUnitPrice?:0
-            } else if (parsedUnitPrice) {
-                ProductPrice productPrice = new ProductPrice()
-                productPrice.price = parsedUnitPrice
-                packageInstance.productPrice = productPrice
-            }
         }
+        if (parsedUnitPrice) {
+            packageInstance.createOrGetProductPrice().price = parsedUnitPrice
+        }
+
 
         if (!productInstance.hasErrors() && productInstance.save(flush: true)) {
             flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'package.label', default: 'Product'), packageInstance.name])}"
@@ -1147,6 +1141,3 @@ class ProductController {
         render(view: "addDocument", model: [productInstance: productInstance, documentInstance: documentInstance])
     }
 }
-
-
-
--- a/grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
@@ -12,9 +12,10 @@ package org.pih.warehouse.product
 import grails.validation.ValidationException
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.PreferenceType
+import org.pih.warehouse.core.ProductPrice
+
 import java.math.RoundingMode
 import java.text.SimpleDateFormat
-import org.pih.warehouse.core.ProductPrice
 
 class ProductSupplierController {
 
@@ -110,10 +111,9 @@ class ProductSupplierController {
                 productSupplierInstance.contractPrice.price = parsedUnitPrice
                 productSupplierInstance.contractPrice.toDate = params.toDate ? dateFormat.parse(params.toDate) : null
             } else {
-                ProductPrice productPrice = new ProductPrice()
-                productPrice.price = parsedUnitPrice
-                productPrice.toDate = params.toDate ? dateFormat.parse(params.toDate) : null
-                productSupplierInstance.contractPrice = productPrice
+                ProductPrice contractPrice = productSupplierInstance.createOrGetContractPrice()
+                contractPrice.price = parsedUnitPrice
+                contractPrice.toDate = params.toDate ? dateFormat.parse(params.toDate) : null
             }
         }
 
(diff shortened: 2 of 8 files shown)


jmiranda (commented):
In general, flushing is a not a great solution, especially with imports. However we can probably live with it. 

I was hoping for more insight into what might be causing this issue. Any chance you feel comfortable walking through the code in a debugger to figure out why this happens?

mdpearson:
> In general, flushing is a not a great solution, especially with imports. However we can probably live with it.
> 
> I was hoping for more insight into what might be causing this issue. Any chance you feel comfortable walking through the code in a debugger to figure out why this happens?

I'd love to. Kelsey had flagged this as a blocker so I wanted to patch it reasonably quickly. I'll follow up with you on Slack about debugging advice.

mdpearson:
Any objection to merging this with a pinky-promise to investigate the root cause shortly after it's released?

jmiranda:
@mdpearson Can you reproduce this reliably in a non-prod environment? My guess is that the issue will disappear if we restart the production server.

mdpearson:
> @mdpearson Can you reproduce this reliably in a non-prod environment? My guess is that the issue will disappear if we restart the production server.

I was able to reproduce on dev. When I pushed a build containing this patch the issue went away -- so, that implies that either a restart or this patch fixed it. I'll push a build without this patch to dev and see if the import error recurs.

jmiranda:
I expect that the issue is related to data loaded into the Hibernate session. From what I've seen it occurs when a finder is invoked while you're persisting objects.

mdpearson:
bummer. I can reproduce on dev despite its having been restarted

jmiranda:
Be careful about removing the withNewSession - it can cause a StackOverflowError. 
https://github.com/grails/grails-core/issues/11294

http://people.apache.org/~pledbrook/grails-guide/guide/5.%20Object%20Relational%20Mapping%20(GORM).html#5.5.1%20Events%20and%20Auto%20Timestamping

> Notice the usage of withNewSession method above. Since events are triggered whilst Hibernate is flushing using persistence methods like save() and delete() won't result in objects being saved unless you run your operations with a new Session.
> 
> Fortunately the withNewSession method allows you to share the same transactional JDBC connection even though you're using a different underlying Session.

Also I noticed that you haven't added in the code to do a flush every 50 or 100 rows? Any reason for that? I really think it's necessary in this case given the number of rows we need to do insert / update. The only reason it wasn't added before was because I didn't expect 5000+ product supplier rows to be imported at once.

mdpearson:
> Also I noticed that you haven't added in the code to do a flush every 50 or 100 rows? Any reason for that?

It wasn't necessary in this case (at least, I was able to import a 3867-line .xls file without any explicit flushes).

mdpearson:
Hi Justin, thanks for those links! A few follow up questions.

> [from the docs] Since events are triggered whilst Hibernate is flushing using persistence methods like save() and delete() won't result in objects being saved unless you run your operations with a new Session.

None of the current `beforeUpdate`/`beforeInsert` methods call `save()` or `delete()`, so I'm not sure if this advice applies to our case?

As for the stack overflow warning I believe it applies to flushing within an event handler, which we also do not do.

Neither the `beforeInsert` nor `beforeUpdate` examples in the grails guide wrap their assignments in a `withNewSession`, for instance:

```
// Fired before an object is saved to the db
class Person {
   Date dateCreated
   def beforeInsert() {
       dateCreated = new Date()
   }
}
```

All that being said, I had another branch where I updated all event handlers to use withNewSession, and that worked fine too. So another approach would to standardize on something like this:

```
def beforeInsert = {
        User.withNewSession {
            def currentUser = AuthService.currentUser.get()
            if (currentUser) {
                createdBy = currentUser
                updatedBy = currentUser
            }
        }
    }
```

... perhaps with an explicit call to `save()` at the end. But I don't think either the save or the withNewSession closure is necessary. Let me know what you think.

jmiranda:
>  Since events are triggered whilst Hibernate is flushing using persistence methods like save() and delete() won't result in objects being saved unless you run your operations with a new Session.

Ah you may be right. I interpretted this differently, but based on the example they give I think you're correct.
```
def beforeDelete() {
      ActivityTrace.withNewSession {
         new ActivityTrace(eventName:"Person Deleted",data:name).save()
      }      
   }
```

jmiranda:
> It wasn't necessary in this case (at least, I was able to import a 3867-line .xls file without any explicit flushes).

So I'm confused as to what in this PR actually fixed the issue. Can you point it out to me? Just wrapping each create with a new session. 
```
ProductSupplier.withNewSession { ... }
```
That seems expensive since we're creating a new Hibernate session for each row / save. But if it work, it works I guess. I'd like to know what happens if we throw an intermittent exception in the middle of the iterator. Does everything rollback? Do some rows save.
```
    createOrUpdate(params)
    if (index % 50) throw new Exception("intermittent error")
}
```
Another thing to note is that the createOrUpdate is used to populate the domain, not persist it. I couldn't come up with a better name for this method, but it's supposed to populate the domain so we can validate and persist. In general, this method shouldn't be used to trigger the save because we sometimes want to do some things to the object before it's saved i.e. show it to the user to make sure the data is correct.

For example, we usually have three phases of the data import (i.e. upload -> validate / review -> import / persist) and we use createOrUpdate() to create the objects that are to be validated and reviewed by the user before performing the flush to the database. Right now it seems like we'd be saving to the database on both steps validate and import. 

The validate might not be using the createOrUpdate method here but just wanted to point out that it's usually what I use these methods for.

mdpearson:
@jmiranda thanks, this is all really helpful. A few quick responses.

> So I'm confused as to what in this PR actually fixed the issue. Can you point it out to me?

Yeah, it's the `withNewSession` closure.

> That seems expensive since we're creating a new Hibernate session for each row / save.

That's one of the reasons why I removed the withNewSession closures from the `beforeUpdate`/`beforeInsert` event handlers. But I think that change should be a separate PR, since it's disjoint from the problem at hand. I'll back it out of this PR. In terms of performance, `ProductSupplierDataService.process()` took about 5 minutes on my machine for the ~4,000-line .xls file, of which (roughly) 2 minutes were spent in `validate()`. The latter method, on quick glance, only hits the db once per record (`UnitOfMeasure.findByCode()`). So if the `save`-ing part of process is only 50% longer than that, the performance doesn't seem awful. But, I wish I had a 4,000 line xls file that _didn't_ crash to compare it to!

> I'd like to know what happens if we throw an intermittent exception in the middle of the iterator.

Oooh, me too! That's a great question. The difference between a hibernate session and a JDBC transaction is still fuzzy to me -- I'll look into this and get back to you.

> Another thing to note is that the createOrUpdate is used to populate the domain, not persist it. I couldn't come up with a better name for this method

Makes perfect sense now that you've explained it to me. I'll (de-)refactor so as to preserve this convention.

mdpearson:
I'm going to split this into two PR's.

PR #3004, opened 2022-02-24
WIP: OBKN-387 Order import needs to support XLSX

chetanmaharishi:
OBKN-387 Order import needs to support XLSX


jmiranda:
thanks @chetanmaharishi. I'll review early next week.

jmiranda:
@chetanmaharishi just pulled down this PR and it's working great. do you think there's an esay way to get this working without needing to create file on the filesystem i.e. using a byte array output stream?

mdpearson:
Hi folks, any reason we're adding a 2.5 mbyte jar file to the repo, instead of fetching it from e.g. maven? Also, 2.x has a pretty serious security vulnerability; if it's at all possible, let's try to use 3.x or later

https://mvnrepository.com/artifact/org.apache.xmlbeans/xmlbeans/2.3.0

jmiranda:
Thanks, @mdpearson. This was a proof of concept (WIP) that I asked Chetan to work on. I think he had issues with the maven dependency. My local version uses the maven dependency but I'll bump up the version to 3.x.

PR #3006, opened 2022-02-25
OBPIH-4260 Added accesses for requestor

MagdaToczek:


grails-app/views/dashboard/megamenu.gsp
```diff
@@ -243,7 +252,28 @@
         </g:hasRoleInvoice>
     </g:if>
 
+    <g:if test="${megamenuConfig.stockRequest.enabled && User.get(session?.user?.id)?.getHighestRole(session.warehouse).roleType == RoleType.ROLE_AUTHENTICATED}">
```
jmiranda, line 255: I don't like the database lookup being done here. Is there a way we could move this to a taglib?

grails-app/views/dashboard/megamenu.gsp
```diff
@@ -243,7 +252,28 @@
...
+                    <g:link controller="stockMovement" action="list" params="[direction:'INBOUND']">
+                        <warehouse:message code="default.list.label" args="[warehouse.message(code: 'stockMovements.inbound.label')]"/>
+                    </g:link>
+                </div>
+            </div>
+        </li>
+    </g:if>
+
     <g:if test="${megamenuConfig.inbound.enabled}">
+        <g:if test="${User.get(session?.user?.id)?.getHighestRole(session.warehouse).roleType != RoleType.ROLE_AUTHENTICATED}">
```
jmiranda, line 276: Same here. Also what are we doing here ... saying that the user can access this feature if they don't have ROLE_AUTHENTICATED? Could we do something like roleType > ROLE_AUTHENTICATED? 

Although now I'm seeing that SUPERUSER = 0 so perhaps we did this in the wrong direction. 

@awalkowiak @pmuchowski any thoughts?

grails-app/views/taglib/_displayLogo.gsp
```diff
@@ -1,8 +1,17 @@
+<%@ page import="org.pih.warehouse.core.RoleType" %>
+<%@ page import="org.pih.warehouse.core.User" %>
 <div class="logo">
     <g:if test="${attrs.includeLink}">
-        <a href="${createLink(uri: '/')}">
-            <img src="${attrs?.logoUrl}" />
-        </a>
+        <g:if test="${User.get(session?.user?.id)?.getHighestRole(session.warehouse).roleType == RoleType.ROLE_AUTHENTICATED}">
```
jmiranda, line 5: One more point about the need for this lookup. I assume you were getting lazy init errors when trying to retrieve an association from session.user. I never should have set the full user object in the session. Instead we should store the user ID in the session and fetch the object on every request. We do this with the thread locals in AuthService so we could/should use that object whenever we need the current user or current location. Or we should hydrate the sesion.user and session.warehouse on every request similar to how we populate the thread locals. See SecurityFilters.

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -149,10 +149,21 @@ class UserService {
         return false
     }
 
+    Boolean hasHighestRoleAuthenticated(User u, String locationId) {
```
jmiranda, line 152: I know this should be obvious, but I'm struggling with the naming of this method. Is this saying, the user at the given should at least by an authenticated user?

Also, this should take the roleType as an argument so it can used in other places.
```
hasHighestRole(User user, String locationId, RoleType )
```

src/groovy/org/pih/warehouse/core/RoleType.groovy
```diff
@@ -17,6 +17,7 @@ enum RoleType {
     ROLE_MANAGER('Manager', 2),
     ROLE_ASSISTANT('Assistant', 3),
     ROLE_BROWSER('Browser', 4),
+    ROLE_AUTHENTICATED('Authenticated', 5),
```
jmiranda, line 20: I was expecting a ROLE_AUTHENTICATED as well as a ROLE_REQUESTOR and/or ROLE_CUSTOMER. 

ROLE_AUTHENTICATED is what a user gets by default if they log in and have no other primary roles. But a user should never be assigned ROLE_AUTHENTICATED. 

ROLE_REQUESTOR (and ROLE_CUSTOMER) would now be primary roles that inherit from ROLE_AUTHENTICATED. 

@pmuchowski Is that what you had in mind as well?

jmiranda (changes requested):
@MagdaToczek I think there are a few details that we still need to work on, but really nice job with this one.

grails-app/views/dashboard/megamenu.gsp
jmiranda: Or something like this
```
AuthService.currentUser.get().isPrimaryRoleAtLeast(RoleType.ROLE_AUTHENTICATED)
```
Method naming is not my forte so if someone has a better name please let me know.

grails-app/services/org/pih/warehouse/core/UserService.groovy
MagdaToczek: Ok, I'll modify this one. It is saying that the user is not in a higher role than ... It is done like this to show simpler menu for user and to block going to dashboard from clicking logo etc.

src/groovy/org/pih/warehouse/core/RoleType.groovy
MagdaToczek: I do not think I understand it well, I'll wait for a tech huddle to talk about it.

grails-app/views/dashboard/megamenu.gsp
MagdaToczek: I would like to say here that user can access this feature if he has role higher than Authenticated.

grails-app/views/dashboard/megamenu.gsp
MagdaToczek: I would like to check if role is not higher than authenticated. I'm not interested in having here role higher than authenticated.

grails-app/conf/RoleFilters.groovy
```diff
@@ -86,13 +95,20 @@ class RoleFilters {
     }
 
     static Boolean needManager(controllerName, actionName) {
-        changeActions.any {
+        def changeActions = changeActions.any {
             actionName?.startsWith(it)
-        } || controllerName?.contains("Workflow") || changeControllers?.contains(controllerName) || managerActions[controllerName]?.contains(actionName)
+        }
+        def controller = controllerName?.contains("Workflow")
```
awalkowiak, line 101: Should be named as `isWorkflow`.

grails-app/conf/RoleFilters.groovy
```diff
@@ -86,13 +95,20 @@ class RoleFilters {
...
+        def changeController = changeControllers?.contains(controllerName)
```
awalkowiak, line 102: Should be named as `isChangeController`.

grails-app/conf/RoleFilters.groovy
```diff
@@ -86,13 +95,20 @@ class RoleFilters {
...
+        def manager = managerActions[controllerName]?.contains(actionName)
```
awalkowiak, line 103: Should be named as `isManagerAction` (or sth similar).

grails-app/conf/RoleFilters.groovy
```diff
@@ -86,13 +95,20 @@ class RoleFilters {
     }
 
     static Boolean needManager(controllerName, actionName) {
-        changeActions.any {
+        def changeActions = changeActions.any {
```
awalkowiak, line 98: Should be named as `isChangeAction`.

grails-app/conf/RoleFilters.groovy
```diff
@@ -57,13 +64,15 @@ class RoleFilters {
                 }
 
                 // Authorized users
+                def missAuthenticated = !userService.isUserInRole(session.user, RoleType.ROLE_AUTHENTICATED)
```
awalkowiak, line 67: How about changing naming of all these variables into:
`isNotAuthenticated`
`isNotBrowser`
`isNotManager`
`isNotAdmin`
`isNotSuperuser`
`hasNotRoleInvoice`
`isNotRequestor`

awalkowiak (changes requested):
On top of Justin's comments, I'd like to get some naming improvements.

grails-app/views/dashboard/megamenu.gsp
awalkowiak: @MagdaToczek if role is higher than Authenticated this means `role_type < ROLE_AUTHENTICATED` (because superuser = 0, admin = 1, ..., authenticated = 5).

awalkowiak: And `getHighestRole` can return `null`, hence you can get error trying access `roleType` on `null` object.

src/groovy/org/pih/warehouse/core/RoleType.groovy
awalkowiak: @jmiranda, @MagdaToczek I think this is what we agreed on the tech huddle.
Add a new role -> Authenticated
Add a new complimentary role -> Requestor
Hence it is missing: `ROLE_REQUESTOR('Requestor, 100)` after `ROLE_INVOICE`

grails-app/conf/RoleFilters.groovy
MagdaToczek: ok

grails-app/conf/RoleFilters.groovy
MagdaToczek: ok

grails-app/conf/RoleFilters.groovy
MagdaToczek: ok

awalkowiak (approved):

PR #3013, opened 2022-03-01
OBPIH-4253 Fixed assigning party_role while creating organization

MagdaToczek:


grails-app/migrations/0.8.x/changelog-2022-03-01-0850-add-unique-constraint-party-type-code.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="mtoczek" id="0103202208500-1">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <columnExists tableName="party_type" columnName="code"/>
+                <indexExists indexName="party_type_code"/>
+            </and>
```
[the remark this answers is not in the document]
jmiranda: Thank you for adding a database migration for this. I usually forget to do that and just add unique: true to the domain mapping. Which probably isn't the end of the world as we usually do most inserts through Grails, but having this constraint in the database is important.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -166,7 +166,34 @@ class IdentifierService {
...
             identifier = identifier.substring(0, maxSize)
         }
-        return identifier.toUpperCase()
+
+        // Checking for a duplicates
+        if (validateOrganizationIdentifier(identifier.toUpperCase())) {
+            return identifier.toUpperCase()
+        }
+
+        if (Character.isDigit(identifier.charAt(0))) {
```
jmiranda, line 175: Can you comment this code? From the looks of it we're adding a prefix to the organization code. 

So let's say we had an organization with code ABC and we tried to add a second one with the same identifier, the second one would be 0ABC? Is that correct? 

Are there requirements in the ticket for this? 

I think I remember Kelsey saying offhand something about appending a number to the code but not a prefix. To me, it seems like we'd want to go ABC, ABC2, ABC3.

jmiranda (changes requested):

grails-app/migrations/0.8.x/changelog-2022-03-01-0850-add-unique-constraint-party-type-code.xml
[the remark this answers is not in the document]
MagdaToczek: I'll modify it.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -39,11 +39,11 @@ class OrganizationService {
     }
 
     Organization findOrCreateOrganization(String name, String code, List<RoleType> roleTypes) {
-        Organization organization = Organization.findByCodeOrName(code, name)
+        Organization organization = !name ? Organization.findByCode(code) : Organization.findByName(name)
```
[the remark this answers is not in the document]
MagdaToczek: Hmm, I added it keeping in mind that code is generated incorrectly and than I modify it so I guess this is not needed anymore, so I revert this change.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -166,7 +166,41 @@ class IdentifierService {
...
+        println "Validating organization identifier " + identifier
+        def count = Organization.executeQuery( "select count(o.code) from Organization o where code = :identifier", [identifier: identifier] )
+
+        return count ? (count[0] == 0) : false
+    }
+
+    def getOrganizationIdentifierWithHighestSuffix(String identifier) {
+        def organizations = Organization.executeQuery( "select o.code from Organization o where code like :identifier", [identifier: identifier.substring(0, identifier.size() -1) + '%'] )
+        organizations.findAll { Character.isDigit(it.charAt(it.size() - 1)) }
+        return organizations.findAll { Character.isDigit(it.charAt(it.size() - 1)) }.sort().last()
```
pmuchowski, line 203: why do you use findAll two times?

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -166,7 +166,41 @@ class IdentifierService {
...
+        }
+
+        if (Character.isDigit(identifier.charAt(identifier.size() - 1))) {
+            String identifierWithHighestNumber = getOrganizationIdentifierWithHighestSuffix(identifier)
+            char suffix = identifierWithHighestNumber.charAt(identifierWithHighestNumber.size() - 1)
+            suffix++
+
+            return identifier.toUpperCase().substring(0, identifier.size() -1) + suffix
+        } else {
+            String identifierWithHighestNumber = getOrganizationIdentifierWithHighestSuffix(identifier)
```
pmuchowski, line 182: you are doing the substring in the getOrganizationIdentifierWithHighestSuffix even if there is no digit at the end

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
MagdaToczek: Oh, it should be deleted.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
MagdaToczek: yes, it is because there is possibility to create identifier without number, and there still can be another one in db, so I would like to check it and increase the highest number.

pmuchowski (approved):

awalkowiak (approved):

PR #3015, opened 2022-03-01
OBPIH-4305 Add clone-remote-database to source control.

mdpearson:
Justin,

I updated the script you sent me via Slack. It Works For Me™ copying the latest `prd` bits to my Mac, where I was able to reproduce OBS-1076 locally with it. Let me know if you have any trouble running it on Linux.

While I was working on this, I encountered the occasional `ERROR 2006 (HY000): MySQL server has gone away` error message, which I resolved by setting `max_allowed_packet` to its [legal maximum](https://dev.mysql.com/doc/refman/5.7/en/packet-too-large.html). The docs say "It is safe to increase the value of this variable because the extra memory is allocated only when needed," but if a 1G ceiling concerns you I can look into this more.

OpenBoxes takes about 10 minutes to start up after this script runs, with logs like
```
2022-03-01 11:10:51,970 [DefaultPluginScheduler_Worker-3] INFO  data.DataService  - Updated 0 rows in 2 ms
2022-03-01 11:10:51,973 [DefaultPluginScheduler_Worker-3] INFO  jobs.RefreshStockoutDataJob  - Refreshed stockout data in 1158 ms
2022-03-01 11:10:54,784 [ForkJoinPool-1-worker-6] INFO  inventory.ProductAvailabilityService  - Refreshing product availability location (BHI: SL Stock), forceRefresh (true) ...
...
...
...
2022-03-01 11:20:07,999 [DefaultPluginScheduler_Worker-5] INFO  inventory.ProductAvailabilityService  - Refreshed product availability in 557048ms
2022-03-01 11:20:08,017 [DefaultPluginScheduler_Worker-5] INFO  jobs.RefreshProductAvailabilityJob  - Finished refreshing product availability data in 557075 ms
```

Not a huge deal, since the script itself takes 30 minutes to fully clone `prd`. But it makes me wonder if I'm missing something when I did the copy, or if these services run automagically when the service starts and just run more slowly than I'm used to because the dataset is so much larger.

### DIFF
--- a/scripts/support/clone-remote-database
+++ b/scripts/support/clone-remote-database
@@ -1,0 +1,115 @@
+#!/bin/bash
+set -euo pipefail
+start_tm=$(date '+%s')
+export TIME="%C %E"
+
+usage()
+{
+	set +xv
+	echo 1>&2 "$0 - clone a remote mysql database to a local machine for testing"
+	echo 1>&2 ""
+	echo 1>&2 "Usage: $0 [-l <LOCAL_DB>] [-r <REMOTE_DB>] REMOTE_HOST"
+	echo 1>&2 "REMOTE_HOST -- name or ip of remote host (passed to ssh)"
+	echo 1>&2 "LOCAL_DB -- name of new database to create locally (default: 'openboxes')"
+	echo 1>&2 "REMOTE_DB -- name of database to clone from REMOTE_HOST (default: 'openboxes')"
+	echo 1>&2 ""
+	echo 1>&2 "Use with care -- this script will destroy \$LOCAL_DB if it exists!!"
+	exit 1
+}
+
+#
+# Pretty-print the difference between two timestamps, $1 and $2.
+#
+elapsed_tm()
+{
+	set +e
+	# guess whether we're running GNU date: thanks for nothing, POSIX
+	date -j >/dev/null 2>&-
+	if [ $? -ne 0 ]
+	then
+		# GNU/linux
+		date -ud@$(($2 - $1)) "+%H:%M:%S"
+	else
+		# macOS/BSD
+		date -juf "%s" $(($2 - $1)) "+%H:%M:%S"
+	fi
+	set -e
+}
+
+LOCAL_DB="openboxes"
+REMOTE_DB="openboxes"
+while getopts "hl:r:" o
+do
+	case "${o}" in
+	l)
+		LOCAL_DB=${OPTARG}
+		[ "$LOCAL_DB" ] || usage
+		;;
+	r)
+		REMOTE_DB=${OPTARG}
+		[ "$REMOTE_DB" ] || usage
+		;;
+	h)
+		usage
+		;;
+	*)
+		usage
+		;;
+	esac
+done
+shift $((OPTIND-1))
+
+[ "$1" ] && [ "$#" -eq 1 ] || usage
+REMOTE_HOST="$1"
+
+
+
+SQL_BASENAME="$REMOTE_HOST-$REMOTE_DB"
+
+echo "Clobbering local database $LOCAL_DB ..."
+mysql -e "drop database if exists \`$LOCAL_DB\`; create database \`$LOCAL_DB\` default charset utf8; grant all on \`$LOCAL_DB\`.* to 'openboxes'@'localhost' identified by 'openboxes';"
+mysql $LOCAL_DB -e 'select 1' > /dev/null
+
+# prevent "ERROR 2006 (HY000): MySQL server has gone away"
+# see also https://stackoverflow.com/questions/10474922/error-2006-hy000-mysql-server-has-gone-away
+mysql -e "set global max_allowed_packet=1024*1024*1024;"
+
+echo -n "Counting products in remote database $REMOTE_HOST:$REMOTE_DB ..."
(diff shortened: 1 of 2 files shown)


scripts/support/clone-remote-database
```diff
@@ -0,0 +1,115 @@
...
+{
+	set +xv
+	echo 1>&2 "$0 - clone a remote mysql database to a local machine for testing"
+	echo 1>&2 ""
+	echo 1>&2 "Usage: $0 [-l <LOCAL_DB>] [-r <REMOTE_DB>] REMOTE_HOST"
+	echo 1>&2 "REMOTE_HOST -- name or ip of remote host (passed to ssh)"
+	echo 1>&2 "LOCAL_DB -- name of new database to create locally (default: 'openboxes')"
+	echo 1>&2 "REMOTE_DB -- name of database to clone from REMOTE_HOST (default: 'openboxes')"
+	echo 1>&2 ""
+	echo 1>&2 "Use with care -- this script will destroy \$LOCAL_DB if it exists!!"
```
jmiranda, line 16: I would require a force argument if the database exists locally.

scripts/support/clone-remote-database
```diff
@@ -0,0 +1,115 @@
...
+
+[ "$1" ] && [ "$#" -eq 1 ] || usage
+REMOTE_HOST="$1"
+
+
+
+SQL_BASENAME="$REMOTE_HOST-$REMOTE_DB"
+
+echo "Clobbering local database $LOCAL_DB ..."
+mysql -e "drop database if exists \`$LOCAL_DB\`; create database \`$LOCAL_DB\` default charset utf8; grant all on \`$LOCAL_DB\`.* to 'openboxes'@'localhost' identified by 'openboxes';"
```
jmiranda, line 70: I would not do this until you've pulled down the remote database successfully.

scripts/support/clone-remote-database
```diff
@@ -0,0 +1,115 @@
...
+mysql $LOCAL_DB -e 'select 1' > /dev/null
+
+# prevent "ERROR 2006 (HY000): MySQL server has gone away"
+# see also https://stackoverflow.com/questions/10474922/error-2006-hy000-mysql-server-has-gone-away
+mysql -e "set global max_allowed_packet=1024*1024*1024;"
```
jmiranda, line 75: Why can't we just set to 1G to make it more clear?

scripts/support/clone-remote-database
```diff
@@ -0,0 +1,115 @@
...
+
+echo -n "Counting products in remote database $REMOTE_HOST:$REMOTE_DB ..."
+remote_product_cnt=$(ssh "$REMOTE_HOST" "mysql $REMOTE_DB -Nse 'select count(*) from product;'")
```
jmiranda, line 78: It probably doesn't matter but since we're potentially touching a production database, I'd like to reduce our footprint as much as possible.
```
select count(id) from product
```

jmiranda:
> OpenBoxes takes about 10 minutes to start up after this script runs, with logs like

OpenBoxes should be running within a minute or so. The ten minutes you're seeing is the time it takes to run background jobs to refresh data (i.e. the product availability table, produdct demand). There are other background jobs that get kicked off as soon as the server starts up and all of them together create a bit of load. But actually ten minutes seems high for those processes, so I would check to make sure you're allocating enough memory to Tomcat and MySQL. 

I use this for my development  environment.
```
GRAILS_OPTS=-XX:PermSize=512m -Xms2048m -Xmx2048m -server
```
I'm not doing anything special with MySQL but when MySQL is the bottleneck you can bump up the innodb_buffer_pool_size and innodb_log_file_size to several GBs to give MySQL room to work.

jmiranda:
Note: max_allowed_packet is set to 16M in my local MySQL config as well as on OBNAV. So not sure why you're running into that `MySQL server has gone away`. 
```
mysql> show variables like 'max_allowed%';
+--------------------+----------+
| Variable_name      | Value    |
+--------------------+----------+
| max_allowed_packet | 16777216 |
+--------------------+----------+
1 row in set (0.00 sec)
```
However, I also don't care about setting this value as long as we're not doing it on the remote server.

scripts/support/clone-remote-database
jmiranda: Also see comment below about my settings.

mdpearson: > Why can't we just set to 1G to make it more clear?

It needs to be an int, if you try a human-readable value you'll get `ERROR 1232 (42000) at line 1: Incorrect argument type to variable 'max_allowed_packet'`

scripts/support/clone-remote-database
mdpearson: Good point. Fixed in e4861fa5f.

scripts/support/clone-remote-database
mdpearson: Turns out this isn't needed anyway, see below.

mdpearson:
> Note: max_allowed_packet is set to 16M in my local MySQL config as well as on OBNAV. So not sure why you're running into that `MySQL server has gone away`.

I'm not 100% sure either; bumping the memory fixed the issue when I was tweaking the script, but when I removed it just now, the script worked fine. I may have added it before I caught a bug that was pulling down way too much from the server. Thanks for noting this! It's gone

PR #3019, opened 2022-03-03
OBDS-87 Categories Download page for products configuration wizard

awalkowiak:


awalkowiak:
@jmiranda I thought about adding a generic `downloadCsv` action to the `BatchController` (similar to the `downloadExcel`, but for csv), however i did not want to plug it directly in the `UrlMappings`. I think, that doing the clone of `BatchController` into `BatchApiController` was not an ideal solution to this, and having a redirect to `BatchController` from `ProductsConfigurationApiController` does not work in this case. So, for now I added a categories csv download directly into the `ProductsConfigurationApiController`.

pmuchowski (approved):

PR #3025, opened 2022-03-07
OBS-1089 Fix stocklist PDF export

mdpearson:
I couldn't reproduce [OBS-1089](https://pihemr.atlassian.net/browse/OBS-1089) on `develop`, but could on `master`; this one-line commit from @awalkowiak is the reason why. Any objection to a patch release to unblock a colleague on the pharmacy team?

argument in favor:
- this change has been soaking in develop since October

possible argument against:
- there may be other commits on develop that might need to come along for the ride

An alternative solution would be to target this patch to the specific template(s) affected by this bug. But I thought I'd go with the easiest fix first! :-)

### DIFF
--- a/grails-app/views/stocklist/_itemList.gsp
+++ b/grails-app/views/stocklist/_itemList.gsp
@@ -34,7 +34,7 @@
                 </g:if>
                 <tr>
                     <td class="b-t0 b-r0">${requisitionItem?.product?.productCode}</td>
-                    <td class="b-t0 b-r0">${requisitionItem?.product?.name}</td>
+                    <td class="b-t0 b-r0">${requisitionItem?.product?.name?.encodeAsHTML()}</td>
                     <td class="b-t0 b-r0">${requisitionItem?.productPackage ? requisitionItem.productPackage.uom?.code + "/" + requisitionItem.productPackage.quantity + " -- " + requisitionItem.productPackage.uom?.name : 'EA/1'}</td>
                     <td class="b-t0 b-r0">${requisitionItem.quantity}</td>
                     <td class="b-t0 b-r0"></td>


jmiranda:
> I couldn't reproduce [OBS-1089](https://pihemr.atlassian.net/browse/OBS-1089) on develop, but could on master; this one-line commit from @awalkowiak is the reason why.

Thanks, Matt. And that makes sense. The other way to deal with this is to escape the data on the way in. But i definitely like this approach better because I don't like changing data without the user knowing. 

> Any objection to a patch release to unblock a colleague on the pharmacy team?

Yes, I'd object because this affects every feature in the system and we already know that it's caused several regressions. So we'd need to do thorough testing which means we'd probably have to run through the full test script and fix a bunch of regressions. 

Instead of changing the codec, there are two options:
* Fix the data issues (for OBS-1089) on a case-by-case basis for now i.e. change > to &rsaquo; or whatever in the database. Or just remove the invalid characters altogether.
* Sprinkle encodeAsHtml() on any of the offending ${...} https://stackoverflow.com/questions/16444146/when-to-encode-as-html-in-grails

That way we don't have to run through the very time-intensive test script twice one for the hotfix and the other for the next release.

FWIW, you can also just change the codec in each deployed system's openboxes-config.groovy without requiring a hotfix. This will cause the regressions mentioned above but without the need of doing a hotfix and full deployment to each system.
```
grails.views.default.codec = "html"
```

awalkowiak:
@mdpearson this codec change can't be merged as is to the production. Basically, because of what @jmiranda said. This one might fix this particular issue, however, it breaks some other stuff. For example, you can check out on the Obdev's `Configuration > Data Migration > Product availability` table. There you can see the clear impact of this change. And there might be a bit more places like that scattered through the system which we are hoping to find during the test script.

mdpearson:
@awalkowiak @jmiranda Thanks for the additional information, that's very helpful!

I only needed to sprinkle one `.encodeAsHTML()` to resolve the issue, which should be far less disruptive. Moreover, it shouldn't cause a problem when @awalkowiak's broader change goes into effect. I'd still like to cut a hotfix for this so as to not play whack-a-mole with export nits. Let me know what you think.

jmiranda (approved):
Ahhh. That's more like it.

PR #3028, opened 2022-03-08
OBPIH-4260 Corrected access for user with browser role

MagdaToczek:


grails-app/conf/RoleFilters.groovy
```diff
@@ -63,6 +63,18 @@ class RoleFilters {
             'stockMovementItemApi': ['getStockMovementItems']
     ]
 
+    def static neutralActions = [
```
awalkowiak, line 66: The naming here is the only thing I do not like.

awalkowiak (changes requested):

grails-app/conf/RoleFilters.groovy
MagdaToczek: Changed to authenticatedActions

awalkowiak (approved):

PR #3031, opened 2022-03-08
OBS-1079, OBPIH-4301 - PDF bugfixes

mdpearson:
This PR updates the tools we use to generate PDF's, and removes a very old vendored dependency called `core-renderer-R8-isomorphic.jar`, which has namespace collisions with iText. It fixes [OBPIH-4301](https://pihemr.atlassian.net/browse/OBPIH-4301).

I can't guarantee it will resolve the intermittent `iTextFontResolver` failures (see [OBS-1079](https://pihemr.atlassian.net/browse/OBS-1079), etc.). However, there are reasons to be optimistic; a string dump of `core-renderer-R8-isomorphic.jar` has matches for:

- `Cp1252`
- `FontFamily`
- `FontResolver`
- Courier
- Helvetica
- Times

If our code, intending to call something in `org.xhtmlrenderer.*`, wound up in a shim library last updated in 2010, it's possible that the tools that reset the font cache did the same and wound up hitting an older iText API. At the very least, these changes should reduce the search space for PDF gremlins in the future.

I tested this against the cases listed in [OBS-1089](https://pihemr.atlassian.net/browse/OBS-1089) and [OBPIH-4301](https://pihemr.atlassian.net/browse/OBPIH-4301) and saw no regressions.


jmiranda:
Thanks, Matt. I like the idea of updating the dependencies if they don't break thing. But some of these changes are a little risky since we're updating dependencies for something that's not being tested here i.e. document template rendering. So I might recommend leaving this one out until after the release unless someone can thoroughly test document templates.

jmiranda (commented):
Approved, but proceed with caution.

mdpearson:
@jmiranda I agree 100%. No one's noticed the OBPIH-4301 issue, and OBS-1079 has an easy-enough remedy when it pops up. That being said, I'd love to learn more about what "thoroughly test document templates" would involve.

jmiranda:
@mdpearson Sounds good. Discuss the document template feature with Kelsey.

PR #3041, opened 2022-03-14
OBDS-117 added endpoints for organization and location group creation

MagdaToczek:


awalkowiak (commented):
@MagdaToczek have you tried using the `genericApi` for these?

MagdaToczek:
@awalkowiak what I do not like about generic api is an inconvenience with passing parameters. So maybe read actions could be used from generic API, but I think that create is better to save as it is now. 
But if you would like to use only generic api it is possible.

awalkowiak (approved):

PR #3042, opened 2022-03-14
OBDS-90 Add product import to Product Config Wizard

pmuchowski:


grails-app/conf/UrlMappings.groovy
```diff
@@ -530,6 +530,11 @@ class UrlMappings {
             action = [GET: "categoryOptions"]
         }
 
+        "/api/productsConfiguration/productOptions"(parseRequest: true) {
```
jmiranda, line 533: YES!!! It's REST-like, not RPC.

jmiranda (changes requested):
Small change requested but otherwise LGTM.

awalkowiak (commented):
I am merging as is to move it to QA. The controller naming refactor can be done together with message improvements that still needs to be done.

PR #3043, opened 2022-03-15
OBDS-122 Added endpoint to list zone or bin locations

MagdaToczek:


grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -128,4 +128,11 @@ class LocationApiController extends BaseDomainApiController {
 
         return location
     }
+
+    def show = {
```
awalkowiak, line 132: It should not be a `show` action, because `show` is more suitable for details of one location instead of a list of locations. Also I think you could just do a cherry-pick of this commit https://github.com/openboxes/openboxes/commit/2ba52a590a6b5b8b386917c9f5b4c1b0a75e4b06 I made of obkn feature branch.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3044, opened 2022-03-15
OBDS-102 Address page and form

kchelstowski:


pmuchowski (changes requested):

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -103,7 +103,12 @@ class LocationApiController extends BaseDomainApiController {
 
         bindLocationData(existingLocation, jsonObject)
 
-        if (!existingLocation.validate() || !existingLocation.save(failOnError: true)) {
+        if (existingLocation.validate() && !existingLocation.hasErrors()) {
+            if (existingLocation?.address?.validate() && !existingLocation?.address?.hasErrors()) {
+                existingLocation.address.save()
+            }
+            existingLocation.save()
```
jmiranda, line 110: Need to handle the case when address fails validation.

jmiranda: Also we should check whether the location validation handles both location and any associations.

jmiranda (changes requested):

pmuchowski (approved):

PR #3046, opened 2022-03-15
OBPIH-4325 Do not use withNewSession in event handlers.

mdpearson:
@jmiranda as discussed, I don't think the `withNewSession` closures are necessary in these domains' event handlers, and we don't use them consistently anyway.

I standardized on UnitOfMeasure's `beforeUpdate`/`beforeInsert` implementation, which matches GORM's handling of date stamps, where the insert handler does not set the updated field:

```
class Person {
   Date dateCreated
   Date lastUpdated
   def beforeInsert() {
       dateCreated = new Date()
   }
   def beforeUpdate() {
       lastUpdated = new Date()
   }
}
```

See also "[Automatic Timestamping](https://grails.github.io/grails2-doc/1.3.9/guide/single.html#5.5.1%20Events%20and%20Auto%20Timestamping)."

### DIFF
--- a/grails-app/domain/org/pih/warehouse/core/BudgetCode.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/BudgetCode.groovy
@@ -14,18 +14,11 @@ import org.pih.warehouse.auth.AuthService
 class BudgetCode implements Serializable {
 
     def beforeInsert = {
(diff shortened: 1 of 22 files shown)


mdpearson:
Will fix this as part of OBGM-366

PR #3050, opened 2022-03-16
OBDS-118 React-select refactor

awalkowiak:
I am pushing this to unblock some other related work. Remaining TODOs:
- fix styling
- find and fix existing bugs
- remove react-select-plus depenedencies

### DIFF
--- a/src/js/components/receiving/PartialReceivingPage.jsx
+++ b/src/js/components/receiving/PartialReceivingPage.jsx
@@ -188,7 +188,8 @@ const TABLE_FIELDS = {
               params.shipmentReceived || isReceived(false, params.fieldValue)}
               options={params.bins}
               onChange={value => params.setLocation(params.rowIndex, value)}
-              objectValue
+              valueKey="id"
+              labelKey="name"
               className="select-xs"
               clearable={false}
             />),
@@ -204,8 +205,9 @@ const TABLE_FIELDS = {
           hide: !hasBinLocationSupport,
         }),
         attributes: {
-          objectValue: true,
           clearable: false,
+          valueKey: 'id',
+          labelKey: 'name',
         },
       },
       recipient: {
@@ -218,6 +220,10 @@ const TABLE_FIELDS = {
           options: users,
           disabled: shipmentReceived || isReceived(true, fieldValue),
         }),
+        attributes: {
+          valueKey: 'id',
+          labelKey: 'name',
+        },
       },
       quantityShipped: {
         type: params => (params.subfield ? <LabelField {...params} /> : null),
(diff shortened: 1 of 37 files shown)


src/js/utils/Select.jsx
```diff
@@ -62,7 +61,7 @@ class Select extends Component {
       if (typeof value === 'string') {
         return { value, label: value };
       } else if (objectValue) {
-        option = { ...value, value: JSON.stringify(value.value) };
+        option = { ...value };
```
pmuchowski, line 64: this is not needed, you can remove this check

src/js/utils/Select.jsx
```diff
@@ -35,11 +34,11 @@ class Select extends Component {
   handleChange(value) {
     if (this.props.multi) {
       const values = _.map(value, val =>
```
pmuchowski, line 36: this map is not needed

src/js/utils/Select.jsx
```diff
@@ -35,11 +34,11 @@ class Select extends Component {
...
-        (this.props.objectValue ? JSON.parse(val.value) : val.value));
+        (this.props.objectValue ? val : val));
       this.setState({ value: values });
       this.props.onChange(values);
     } else if (value !== null && value !== undefined) {
-      const val = this.props.objectValue ? JSON.parse(value.value) : value.value;
+      const val = this.props.objectValue ? value.value : value;
```
pmuchowski, line 41: you don't have to check the objectValue here

src/js/utils/Select.jsx
```diff
@@ -77,14 +76,15 @@ class Select extends Component {
     let value = selectValue;
 
     if (objectValue) {
-      value = multi ? _.map(selectValue, val => JSON.stringify(val))
-        : JSON.stringify(selectValue);
+      value = multi ? _.map(selectValue, val => val)
```
pmuchowski, line 79: this map is not needed

src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -475,9 +475,30 @@ class PartialReceivingPage extends Component {
...
+          this.setState({
+            values: {
+              ...response,
+              containers: _.map(response.containers, container => ({
+                ...container,
+                shipmentItems: _.map(container.shipmentItems, shipmentItem => ({
+                  ...shipmentItem,
+                  binLocation: {
+                    ...shipmentItem.binLocation,
+                    value: shipmentItem.binLocation ? shipmentItem.binLocation.id : null,
```
pmuchowski, line 490: do we have to add value and label everywhere, for the select, I think it might be configurable, so you can specify the label field name and value field name in the select properties

pmuchowski (changes requested):

src/js/utils/Select.jsx
awalkowiak: yup, agree

src/js/utils/Select.jsx
awalkowiak: yup, agree

pmuchowski (approved):

PR #3053, opened 2022-03-16
OBDS-111 Add initial HelpScout JavaScript magic.

mdpearson:

PR #3059, opened 2022-03-18
OBDS-136 Add a HelpScout beacon to the Dashboard.

mdpearson:
Hi folks, I'm trying to embed two scripts in the react-based dashboard page, like I did for [OBDS-111 on gsp pages](https://github.com/openboxes/openboxes/pull/3053). `npm run bundle` doesn't complain about the syntax here, but when I run this, the dashboard page is blank. Any suggestions as to what I'm doing wrong?

MANY THANKS!


mdpearson:
@awalkowiak take a look at this and let me know what you think. As we discussed earlier today, I updated React to 16.8.x to use a handy glue component that lazy-loads the HelpScout beacon on the dashboard page. I browsed around the dashboard and didn't notice any ill effects of the update.

src/js/components/dashboard/Dashboard.jsx
```diff
@@ -445,6 +451,12 @@ Dashboard.propTypes = {
   dashboardConfig: PropTypes.shape({
     dashboards: PropTypes.shape({}),
     dashboardWidgets: PropTypes.shape({}),
+    helpscout: PropTypes.shape({
```
awalkowiak, line 454: These properties are not used for now. Can be removed.

grails-app/conf/Config.groovy
```diff
@@ -907,6 +907,12 @@ openboxes {
                 endpoint = "/${appName}/api/dashboard/openPurchaseOrdersCount"
             }
         }
+//        helpscout {
```
awalkowiak, line 910: Since this won't work like that you can get rid of this part.

package.json
```diff
@@ -64,17 +64,18 @@
     "lodash": "4.17.21",
     "moment": "2.22.2",
     "popper.js": "1.14.7",
-    "prop-types": "15.6.2",
+    "prop-types": "15.8.1",
     "query-string": "6.10.1",
-    "react": "16.4.2",
+    "react": "16.8.6",
```
awalkowiak, line 69: Were there any significant issues with that one?

src/js/components/dashboard/Dashboard.jsx
```diff
@@ -398,6 +399,11 @@ class Dashboard extends Component {
             />
           </div>
         </div>
+        {/* FIXME fetch providerKey from config instead of hard-coding it here */}
+        <LiveChatLoaderProvider provider="helpScout" providerKey="5eafcb3f-9e7e-4943-8ebb-669f60a696cd">
```
awalkowiak, line 403: This could be fetched in the similar way we get the `appContext`

awalkowiak (commented):
Later today I'll pull these changes locally, and give some further feedback.

grails-app/conf/Config.groovy
awalkowiak: Or the HelpScout config could be an additional config since we will probably pull these values via API

mdpearson:
@awalkowiak per your suggestion to use something similar to `appContext`, I only see it being fetched in one place, and in groovy code, not React. I'm sure I'm missing something...

```
grails-app/services/org/pih/warehouse/core/DocumentService.groovy
132:        def appContext = ApplicationHolder.application.parentContext
136:            file = appContext.getResource(archiveDirectory)?.getFile()
```

mdpearson:
I didn't notice any change in functionality when I bumped React from 16.4 to 16.8, other than the elimination of a dozen or so peer dependency warnings when running `nmp i`. (A number of other packages expect React 16.8+ already.)

mdpearson:
Hi @awalkowiak, let me know what you think. I moved the beacon definition to the header instead of the dashboard so that it will appear on any other React-based pages that generate a full page. If it works for you, feel free to merge at your leisure. I'll assign the ticket to you as a reminder. Thanks again.

package-lock.json
```diff
@@ -12059,9 +11998,9 @@
       }
     },
     "minimist": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
-      "integrity": "sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw=="
+      "version": "1.2.6",
```
awalkowiak, line 12001: Thanks for this (https://github.com/openboxes/openboxes/security/dependabot/93)

src/js/components/Layout/Header.jsx
```diff
@@ -71,29 +88,38 @@ class Header extends Component {
...
 export default connect(mapStateToProps)(Header);
 
 Header.propTypes = {
-  /** Active user's username */
-  username: PropTypes.string.isRequired,
-  /** Indicator if active user is impersonated */
+  helpScoutColor: PropTypes.string,
+  helpScoutKey: PropTypes.string,
+  highestRole: PropTypes.string.isRequired,
+  isHelpScoutEnabled: PropTypes.string,
```
awalkowiak, line 114: Should be `bool` instead of `string`.

src/js/components/Layout/Header.jsx
```diff
@@ -38,11 +38,28 @@ class Header extends Component {
   }
 
   render() {
+    let helpScoutBeacon = <div />;
+    if (this.props.isHelpScoutEnabled) {
```
awalkowiak, line 42: Please move it to a separate method. It should be: 
```
getHelpScoutBeacon() {
  if (this.props.isHelpScoutEnabled) {
    return (
       <LiveChatLoaderProvider provider="helpScout" providerKey={this.props.helpScoutKey}>
          <HelpScout
            color={this.props.helpScoutColor}
            horizontalPosition="right"
            icon="question"
            idlePeriod="0"
            zIndex="1050"
          />
        </LiveChatLoaderProvider>
    );
  }
  
  return null;
} 
```
and move it above `render` method and replace `{helpScoutBeacon}` with `{this.helpScoutBeacon()}`

awalkowiak (commented):
@mdpearson Looks good. I have two minor improvement-change requests in comments and a question: currently, you added it in the `Header` component. It is available in every place (where we have this header). If you want to have the beacon only on the dashboard, then it would be better to put it in the `Dashboard` component or if it can be in the entire app, then i think `Router` component might be better place (after `ClimbingBoxLoader` and `Alert` components)

mdpearson:
Hi @awalkowiak, thanks for catching my copy-pasta and for the suggestion to move the beacon to a separate method.

I don't want the beacon on only the dashboard, but I also don't want it appearing if React is just managing some small part of the page (I don't know if that ever happens in the app, so perhaps that's an unnecessary concern on my part).

Since Header.jsx already fetches data from the ApiController, which the beacon needs for config, I put it there. Taking a quick glance at the Router page, I think I'd need to add a few extra function calls to get that state, which isn't the end of the world (unless there's a performance hit to calling ApiController?) but is a little more complex.

I'm happy putting it wherever! Just let me know. Is there a difference between putting it in the Header vs. the Router vs. say the Footer?

awalkowiak (approved):

PR #3067, opened 2022-03-22
OBDS-110 Tooltip improvements

MagdaToczek:


src/js/components/locations-configuration/LocationDetails.jsx
```diff
@@ -61,6 +61,9 @@ const FIELDS = {
     attributes: {
       withTooltip: true,
       tooltip: 'react.locationsConfiguration.locationNumber.tooltip.label',
+      tooltipUrl: 'https://openboxes.atlassian.net/wiki/spaces/OBW/pages/1291452471/Configure+Organizations+and+Locations',
```
awalkowiak, line 64: These attributes are a bit odd. I think it would be better to add a `tootlipHtml` attribute, and then assign it to html param on Tooltip. So you could have there `html={tooltipHtml ? tooltipHtml : translate(tooltip, tooltip)}`

awalkowiak (changes requested):

awalkowiak:
@DaRoTP FYI, there are some eslint issues, but I guess after the requested refactor it should be cleaned up.

DaRoTP:
I removed the **tooltipHTML** attribute because it is unnecessary since **react-localize-redux** library already supports html elements in the translation, all you have to do is add the html element _(for example a link/anchor tag)_ in the _messages.properties_

grails-app/i18n/messages.properties
```diff
@@ -3414,7 +3414,7 @@ react.locationsConfiguration.skipConfiguration.label=Skip Configuration for now
 react.locationsConfiguration.additionalTitle.label=Fill in the details for your location. Click the question mark next to the field name to find out more.
 react.locationsConfiguration.details.label=Details
 react.locationsConfiguration.locationNumber.label=Location Number
-react.locationsConfiguration.locationNumber.tooltip.label=Enter a short code to identify the location. Often users choose an acronym, like “NDH” for “Neno District Hospital”.
+react.locationsConfiguration.locationNumber.tooltip.label=Enter a short code to identify the location. Often users choose an acronym, like “NDH” for “Neno District Hospital” <a href='https://openboxes.atlassian.net/wiki/spaces/OBW/pages/1291452471/Configure+Organizations+and+Locations'>Link to Something</a>.
```
awalkowiak, line 3417: Actually, links should not be specified here. I think we'd need to do it in a configurable way. In the config should be a property with map of links that we will be able to pull through an `/api/supportLinks` endpoint. For example:
```
openboxes.supportLinks = [
    locationName: 'https://openboxes.atlassian.net/wiki/spaces/OBW/pages/1291452471/Configure+Organizations+and+Locations',
]
```
and then on the front end side the tooltip content should be built in the `FIELDS` declaration based on a translated message with link and passed down to `Select` with attributes.

awalkowiak (changes requested):
See my comment above

src/js/components/locations-configuration/LocationDetails.jsx
```diff
@@ -42,7 +42,7 @@ function validate(values) {
   return errors;
 }
 
-const FIELDS = {
+const FIELDS = supportLinks => ({
```
awalkowiak, line 45: This should be passed into dynamic attributes.

src/js/utils/form-utils.jsx
```diff
@@ -64,16 +64,23 @@ export const renderFormFields = ({
...
-              { attr.withTooltip ?
+              {attr.withTooltip &&
                 <Tooltip
-                  title={translate(tooltip, tooltip)}
+                  interactive="true"
+                  arrow="true"
+                  trigger={trigger}
+                  hideOnClick="true"
+                  html={tooltipLinks
+                    ? ReactHtmlParser(translate(tooltip, tooltip, tooltipLinks))
```
awalkowiak, line 74: This `tooltipLinks` parameter should be named more generic because it will not always be the link.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3071, opened 2022-03-22
OBDS-105: Zone and Bin: Bin Table

kchelstowski:
As promised - I added a comment in the request inside `onFetchData` so we remember to change that parentLocation is not the session.location, but the location we are currently creating.


kchelstowski:
I think it would be better not to merge it yet, because there were some "big" changes in `ZoneAndBinLocations.jsx` while I was doing Zone Modal (OBDS-104) and it would be better for me to merge OBDS-104 first, so I will have less work while rebasing to develop with this branch, than if you merged this branch first.

PR #3075, opened 2022-03-24
OBDS-104 Zone and Bin: Zone modal

kchelstowski:


src/js/components/locations-configuration/EditZoneModal.jsx
```diff
@@ -0,0 +1,141 @@
...
+          ...acc,
+          [fieldName]: 'react.default.error.requiredField.label',
+        };
+      }
+      return acc;
+    }, {});
+};
+
+
+class EditZoneModal extends Component {
```
pmuchowski, line 56: you don't need a separate component for editing, it's almost identical to the addZoneModal

src/js/components/locations-configuration/ZoneAndBinLocations.jsx
```diff
@@ -1,26 +1,126 @@
...
+    this.addLocation = this.addLocation.bind(this);
+    this.setShowEditZoneModal = this.setShowEditZoneModal.bind(this);
+    this.handleZoneEdit = this.handleZoneEdit.bind(this);
+    this.editLocation = this.editLocation.bind(this);
+    this.deleteLocation = this.deleteLocation.bind(this);
     this.state = {
+      ...INITIAL_STATE,
+      showAddZoneModal: false,
+      showEditZoneModal: false,
+      valuesToEdit: {},
```
pmuchowski, line 42: I think you can find some better name for this, e.g. zoneToEdit or something similar

src/js/components/locations-configuration/ZoneAndBinLocations.jsx
```diff
@@ -1,26 +1,126 @@
...
+  }
+
+  handleZoneEdit(values) {
+    this.setState({
+      showEditZoneModal: !this.state.showEditZoneModal,
+      valuesToEdit: { ...values },
+    });
+  }
+
+  addLocation(data, entity) {
```
pmuchowski, line 63: why you have `entity` as a parameter here?

src/js/components/locations-configuration/ZoneAndBinLocations.jsx
```diff
@@ -1,26 +1,126 @@
...
+          },
+        },
+        {
+          label: this.props.translate('react.default.no.label', 'No'),
+        },
+      ],
+    });
+  }
+
+  fetchData(data, entity) {
```
pmuchowski, line 115: I think it will be best to create a separate method instead of using the entity parameter, also it's not used to fetch the data, so you should rename it, it can be updateZoneData

pmuchowski (changes requested):

src/js/components/locations-configuration/AddZoneModal.jsx
```diff
@@ -0,0 +1,93 @@
...
+      },
+    };
+  }
+
+  handleSubmit(values) {
+    this.props.showSpinner();
+    apiClient.post('/openboxes/api/locations/', {
+      ...values,
+      'parentLocation.id': `${this.props.locationId}`,
+      'locationType.id': 'ZONE',
```
awalkowiak, line 29: It should not be added like that. We do not have certainty what id this type will have in some other databases. You can take the id for this from the Location Type options, where locationTypeCode is 'ZONE".

awalkowiak (changes requested):
See my comment about zone location type id. Plus why there are so many changes in the package-lock.json?

src/js/components/locations-configuration/AddZoneModal.jsx
awalkowiak: (same will be for Bin Location in your next task)

src/js/components/locations-configuration/AddZoneModal.jsx
```diff
@@ -0,0 +1,94 @@
...
+      },
+    };
+  }
+
+  handleSubmit(values) {
+    this.props.showSpinner();
+    apiClient.post('/openboxes/api/locations/', {
+      ...values,
+      'parentLocation.id': `${this.props.locationId}`,
+      'locationType.id': `${this.props.zoneTypeId}`,
```
awalkowiak, line 29: As per discussion during standup: it might be safer to have a select since there might be multiple zone types in the system. You can make it preselected if there is only one location type with `ZONE` code in the options.

awalkowiak (changes requested):
One last thing to change according to our standup discussion.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3078, opened 2022-03-29
OBDS-129 Add import trigger for location groups, locations, organizat…

awalkowiak:
…ions, bin locations, categories and products with GET /api/loadData


pmuchowski (approved):

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
```diff
@@ -13,10 +13,41 @@ import grails.converters.JSON
...
         def listOfDemoData = grailsApplication.config.openboxes.configurationWizard.listOfDemoData
         
         render([data: listOfDemoData] as JSON)
     }
 
+
+
+    def load = {
+        loadDataService.importOrganizations(
+                new URL(grailsApplication.config.openboxes.configurationWizard.dataFiles.organizations)
```
awalkowiak, line 28: @hdsldv Perhaps it would be worth of adding the checks if the specific data loader is enabled or at least present in the config. So either:
```
if (grailsApplication.config.openboxes.demoData.organizations) {
    loadDataService.importOrganizations(new URL(grailsApplication.config.openboxes.demoData.organizations)
}
```
or url could be moved to the `url` parameter and we could add `enabled` prop:
```
if (grailsApplication.config.openboxes.demoData.organizations.enabled) {
    loadDataService.importOrganizations(new URL(grailsApplication.config.openboxes.demoData.organizations.url)
}
```
We could also include the token indexes for specific properties if we need. For example:
```
grailsApplication.config.openboxes.demoData = [
    organizations: [
        enabled: true.
        url: "urlToFile",
        nameIndex: 1, // 1 by default
        typeIndex: 2, // 2 by default
    ],
    (...)
]
```

cc @pmuchowski @jmiranda

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+
+    def importLocationGroups(URL csvURL) {
+        BufferedReader reader = csvURL.newInputStream().newReader();
+
+        CSVReader csvReader = CSVReaderUtils.toCsvReader(reader, [
+                'skipLines': 1
+        ]);
+
+        int nameIndex = 0;
```
awalkowiak, line 48: @hdsldv I think it would be worth keeping in the `demoData` config and it could be:
`int nameIndex = grailsApplication.config.openboxes.demoData.locationGroups.nameIndex ?: 0`

(side note: you don't have to add semicolons at the end of lines)

hdsldv: I changed it to parse column names automatically so I can access column values via names

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
hdsldv: Added enabling via config, column value are now accessed via names

grails-app/conf/Config.groovy
```diff
@@ -1209,7 +1209,7 @@ openboxes.typeahead.delay = 300
 openboxes.typeahead.minLength = 3
 
 // Allow system administrators to disable refresh on startup
-openboxes.refreshAnalyticsDataOnStartup.enabled = true
+openboxes.refreshAnalyticsDataOnStartup.enabled = false
```
jmiranda, line 1212: What is the reason for disabling this?

grails-app/conf/Config.groovy
```diff
@@ -1804,6 +1804,45 @@ openboxes.configurationWizard.listOfDemoData = [
                  "<ul>",
 ]
 
+openboxes.configurationWizard.dataInit = [
```
jmiranda, line 1807: Can we convert this to Groovy DSL syntax
```
openboxes.configuration.data { 
    locations { 
        enabled = true
        url = 
    }
    ...
}
```

grails-app/conf/UrlMappings.groovy
```diff
@@ -60,6 +60,11 @@ class UrlMappings {
         }
 
 
+        "/api/loadData"(parseRequest: true) {
```
jmiranda, line 63: Let's make this more REST-like And we need to do some thinking about naming convention here. 

Something like this maybe
```
GET /api/config/data/demo
```
But I'll do some more thinking about this tomorrow.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1345,4 +1345,25 @@ class ProductService {
             importProducts(products)
         }
     }
+
+    ProductCatalogItem createOrUpdateProductCatalogItem(Map params) {
```
jmiranda, line 1349: Move this to a new ProductCatalogService.

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -0,0 +1,187 @@
...
+**/
+package org.pih.warehouse.data
+
+
+import org.grails.plugins.csv.CSVMapReader
+import org.pih.warehouse.core.*
+import org.pih.warehouse.importer.ImportDataCommand
+import org.pih.warehouse.product.ProductCatalog
+
+class LoadDataService {
```
jmiranda, line 18: I think these methods should go into DataService.

jmiranda (changes requested):
Where did this data come from?

grails-app/conf/Config.groovy
awalkowiak: @jmiranda this was a suggestion for development purposes. This is the WIP version, and @hdsldv will remove it in the final version

grails-app/conf/Config.groovy
awalkowiak: Yup, agree. I gave the same suggestion.

grails-app/conf/Config.groovy
hdsldv: reverted

grails-app/conf/Config.groovy
hdsldv: done

grails-app/conf/UrlMappings.groovy
hdsldv: For now changed to /api/config/data/demo

grails-app/services/org/pih/warehouse/product/ProductService.groovy
hdsldv: done

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
hdsldv: I do not think it is possible. LoadDataService requires productService, which requires dataService. After moving to DataService there is creating bean exception, because of circular reference between DataService and ProductService

hdsldv:
> Where did this data come from?
@jmiranda If you have csv files in mind, they are linked in ticket description (except categories which is copy of UNSPSC_categories.csv). I converted them from excel to csv format

grails-app/conf/Config.groovy
hdsldv: I reverted change to DSL as I was not able to use it as key-value map

grails-app/conf/Config.groovy
jmiranda: @awalkowiak @hdsldv no worries. i was just wondering if there was some kind of conflict with the changes we were making. 

A better approach for local development environments would be to create ~/.grails/openboxes-config.groovy and add the property there.

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -0,0 +1,291 @@
...
+                if (!role) {
+                    role = new Role(
+                            roleType: roleType,
+                            name: roleType.toString(),
+                    )
+                }
+                userRoles.add(role)
+            }
+
+            if (User.findByUsername(attr.username)) {
```
jmiranda, line 257: This should be already handled through the unique constraint on the User domain.

jmiranda: You just need to check if the save was successful and throw an exception if not. 
```
// The hasErrors checks whether there were errors when data binding
if (user.hasErrors() || !user.save()) { 
    throw new ValidationException("invalid user", user.errors)
}
```

Or you can explcitly validate the domain before saving. 
```
user.validate() 
if (user.hasErrors() || !user.save()) { 
    throw new ValidationException("invalid user", user.errors)
}
```
I prefer the first approach.

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -0,0 +1,291 @@
...
+
+            Set<Role> userRoles = new HashSet<Role>();
+
+            for (RoleType roleType in roleTypes) {
+                Role role = Role.findByRoleType(roleType)
+
+                if (!role) {
+                    role = new Role(
+                            roleType: roleType,
+                            name: roleType.toString(),
```
jmiranda, line 251: You should use roleType.name() instead of roleType.toString() to prevent any weirdness with the implicit toString().

https://stackoverflow.com/questions/6667243/using-enum-values-as-string-literals

However, I don't think we want to be creating new system roles during import unless it's explicitly stated in the requirements. I'd rather throw an exception stating that a role with role type doesn't exist. There's also a case where one role type might be mapped to multiple roles. Not sure how we want to handle that. Probably throw an exception for that as well. 
 
Note, once you add a role to the userRoles set, nothing else happens so all of this role / user role logic is basically a no-op as the user roles won't be persisted to the database.

Instead, you probably want to use user.addToUserRoles(). 
```
for (RoleType roleType in roleTypes) {
    Role role = Role.findByRoleType(roleType)
    if (!role) throw new IllegalArgumentException("Role for role type ${roleType.name()} does not exist")
    user.addToUserRoles(new UserRole(user: user, role: role))
}
```
One caveat is that we probably need to save the user before we process the roles.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
hdsldv: Added failOnError when saving user

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
hdsldv: Now null role throws exception. I'm not sure what you mean by no-op - I'm assigning userRoles set to roles property and it works just fine - user and corresponding user_role records are created

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -0,0 +1,291 @@
...
+                userRoles.add(role)
+            }
+
+            if (User.findByUsername(attr.username)) {
+                throw new IllegalArgumentException('Duplicate username: ' + attr.username);
+            }
+
+            User user = new User(
+                    username: attr.username,
+                    password: attr.username, // FIXME: What is default password?
```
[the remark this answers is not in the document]
hdsldv: I added `active: false` for imported users as password has `blank: false` constraint

grails-app/conf/Config.groovy
hdsldv: It works, thanks

awalkowiak:
Since the majority of work was done and only fixes to the product supplier are the remaining part to be done here, I am gonna merge this one to unblock other work. I'll create a ticket for improvements and put it in the next sprint.

PR #3084, opened 2022-03-30
OBDS-109: Success Message

kchelstowski:


src/js/components/locations-configuration/SuccessMessage.jsx
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+
+import Translate from 'utils/Translate';
+
+import 'components/locations-configuration/SuccessMessage.scss';
+
+
+const SuccessMessage = ({ setShowSuccessMessage, history }) => (
+  <div className="success-main">
```
pmuchowski, line 12: I think you should use react-modal here, so all of this custom styling will not be needed

src/js/components/locations-configuration/Forecasting.jsx
```diff
@@ -109,9 +117,12 @@ class Forecasting extends Component {
     this.props.previousPage({ ...this.state.values, forecasting: values });
   }
 
+
   render() {
     return (
       <div className="d-flex flex-column">
+        {this.state.showSuccessMessage &&
```
pmuchowski, line 124: if you use Modal in the SuccessMessage component this condition will not be needed, you could just pass showSuccessMessage as a parameter

pmuchowski (changes requested):

pmuchowski (approved):

PR #3087, opened 2022-04-01
OBPIH-4310: Inventory summary indicator doesn't load when depot has b…

kchelstowski:
…een created by import

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
@@ -17,6 +17,7 @@ import org.pih.warehouse.core.LocationType
 import org.pih.warehouse.core.LocationTypeCode
 import org.pih.warehouse.core.Organization
 import org.pih.warehouse.importer.ImportDataCommand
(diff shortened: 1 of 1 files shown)


awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -134,7 +135,9 @@ class LocationDataService {
         location.parentLocation = params.parentLocation ? Location.findByNameOrLocationNumber(params.parentLocation, params.parentLocation) : null
         location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
         location.address = createOrUpdateAddress(params, location?.address?.id)
-
+        if (!location.inventory) {
```
awalkowiak, line 138: One thing @jmiranda suggested, it should check if the location have no parent
`!location.inventory && !location.parentLocation`

awalkowiak (changes requested):

awalkowiak (approved):

PR #3090, opened 2022-04-03
OBPIH-4345: User cannot continue work in location configuration after…

kchelstowski:
… page reload

### DIFF
--- a/src/js/components/locations-configuration/LocationDetails.jsx
+++ b/src/js/components/locations-configuration/LocationDetails.jsx
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import { Form } from 'react-final-form';
 import { getTranslate } from 'react-localize-redux';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
 import Alert from 'react-s-alert';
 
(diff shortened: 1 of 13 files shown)


awalkowiak (changes requested):

src/js/components/locations-configuration/LocationDetails.jsx
```diff
@@ -210,6 +214,55 @@ class LocationDetails extends Component {
...
 
+  fetchLocation() {
+    const url = `/openboxes/api/locations/${this.props.match.params.locationId}`;
+    apiClient.get(url).then((response) => {
+      const location = response.data.data;
+      if (!location) {
+        this.props.history.push('/openboxes/locationsConfiguration/create');
+        return;
+      }
+      const splitLocationTypeName = (locationType) => {
```
awalkowiak, line 225: @kchelstowski move this to utils please, so it could be used in other places. It could take a text parameter and a locale parameter. It always felt weird that we had to do this. 

@jmiranda why some names in the system are using these weird type of i18n? Could not it be in the description field when it's needed? Or perhaps there is a proper approach to such translations that we could implement.

PR #3091, opened 2022-04-04
OBPIH-4339, 4340 Product association improvements

awalkowiak:


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -492,8 +492,8 @@ class ProductAvailabilityService {
         if (products.size() != quantityMap.size()) {
             def missingProducts = products - quantityMap.keySet()
```
pmuchowski, line 493: This will not work, because products is a list of products and keySet is a list of ids, also I'm not sure it's needed, you can just go through the entire list (like it's doing right now)

pmuchowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: You right, I missed that.

awalkowiak: @pmuchowski fixed

pmuchowski (approved):

PR #3092, opened 2022-04-04
OBDS-101: Add location color pickers and logo upload

pmuchowski:


src/js/utils/ColorPicker.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+import PropTypes from 'prop-types';
+import { SliderPicker } from 'react-color';
+import { Tooltip } from 'react-tippy';
+
+import Input from 'utils/Input';
+
+const ColorPicker = ({ value, onChange }) => {
+  const handleChange = (color) => {
+    const hex = color.hex.split('#')[1];
```
awalkowiak, line 11: @pmuchowski is it possible that value will be empty or null?

awalkowiak (approved):

PR #3095, opened 2022-04-05
OBPIH-4345-new: User cannot continue work in location configuration after page reload

kchelstowski:
With the problems I had with the previous branch (OBPIH-4345) with conflicts and rebasing, I created a new one OBPIH-4345-new and making PR on that, because when trying to push changes from OBPIH-4345-new to OBPIH-4345, the conflicts were still appearing.

### DIFF
--- a/grails-app/conf/BootStrap.groovy
+++ b/grails-app/conf/BootStrap.groovy
@@ -56,6 +56,7 @@ import org.pih.warehouse.shipping.ContainerType
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentItem
 import org.pih.warehouse.shipping.ShipmentType
+import org.pih.warehouse.core.Address
 import util.LiquibaseUtil
 
 import javax.sql.DataSource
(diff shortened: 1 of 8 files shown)


grails-app/conf/BootStrap.groovy
```diff
@@ -157,7 +170,11 @@ class BootStrap {
                     backgroundColor : location?.bgColor,
                     zoneName : location?.zone?.name,
                     zoneId : location?.zone?.id,
-                    active : location?.active
+                    active : location?.active,
+                    organization : location?.organization,
+                    manager: location?.manager,
+                    address: location?.address,
+                    supportedActivities: location.supportedActivities ?: location.locationType?.supportedActivities,
```
awalkowiak, line 177: @jmiranda I am moving a question from previous pr to here:
> @jmiranda I am wondering if we want to include supportedActivities in this marshaller or if we want to keep it simpler and add a custom toJson mapper that will include these related objects like address and supportedActivities. What do you think about it?

For now, I am gonna merge it as-is and we can add a separate mapper later if needed.

awalkowiak (approved):

PR #3102, opened 2022-04-06
OBDS-132 Added Progress Screen for loading demo data

DaRoTP:


src/js/components/locations-configuration/LoadDataPage.jsx
```diff
@@ -90,31 +97,72 @@ const LoadDataContent = ({
       {ReactHtmlParser(summaryItemsList)}
     </div>
     <div className="d-flex justify-content-between m-3">
-      <button type="button" onClick={stepBackHandler} className="btn btn-outline-primary">
+      <button type="button" onClick={() => nextStep(null)} className="btn btn-outline-primary">
```
awalkowiak, line 100: Add a `prevStep` handler. (Bring back `stepBackHandler` but as a `prevStep`). Unless you want to unify it under one handler, then call it `goToStep`.

src/js/components/locations-configuration/LoadDataPage.jsx
```diff
@@ -90,31 +97,72 @@ const LoadDataContent = ({
...
       </button>
     </div>
   </React.Fragment>
 );
 
+const LoadDataProgressScreen = ({ nextStep }) => {
+  useEffect(() => {
+    apiClient.get('/openboxes/api/config/data/demo')
+      .then(() => {
+        nextStep(LOAD_DATA_STEPS.dataLoadedSuccess);
```
awalkowiak, line 114: I don't like the idea of `dataLoadedSuccess` being its own step, it does not feel right to me. I'd prefer it to be content that is displayed in `then` and the same for error content being displayed in `catch`.

awalkowiak (changes requested):

src/js/components/locations-configuration/LoadDataPage.jsx
DaRoTP: yes, I wanted to unify it as one handler and I agree **goToStep** make a lot more sens in that case than **nextStep**

awalkowiak (changes requested):
@DaRoTP what I realized is that this load data page should be in a separate directory. Please, move this from `locations-configuration` to the `load-demo-data`. On top of that please extract every stateful component to separate files.

awalkowiak (approved):

PR #3103, opened 2022-04-06
OBPIH-4354: Add zone to bin modal and table

kchelstowski:


src/js/components/locations-configuration/ZoneAndBinLocations.jsx
```diff
@@ -171,12 +183,21 @@ class ZoneAndBinLocations extends Component {
 
   handleLocationEdit(values) {
     this.props.showSpinner();
-    apiClient.post(`/openboxes/api/locations/${values.id}`, {
+    const payload = values.zoneLocation ? {
```
awalkowiak, line 186: Instead of doing it that way, could you do it the same as in `handleSubmit` from `AddBinModal`?

awalkowiak (changes requested):

awalkowiak (approved):

PR #3104, opened 2022-04-07
OBAM-310 Disable Stock Transfer and Replenishment workflows in locations without bin tracking

DaRoTP:
Disable Stock Transfer and Replenishment workflows in locations without bin tracking

### DIFF
--- a/grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
+++ b/grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
@@ -9,6 +9,7 @@
  **/
 package org.pih.warehouse
 
+import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.RoleType
 import org.pih.warehouse.core.User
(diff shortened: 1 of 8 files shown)


grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -18,9 +19,24 @@ class AuthTagLib {
     def userService
 
     def supports = { attrs, body ->
-        def location = Location.load(session.warehouse.id)
-        if (location && location.supports(attrs.activityCode)) {
```
awalkowiak, line 22: What with other existing usages of `g:supports activityCode`? Looks like you removed the check for activityCode, but you did not modify all usages. I'm fine with unifying `authorize` with `supports`, but then you need to make sure everything was covered. I suggest leaving the chack for single `activityCode` too.

@jmiranda are you fine with unifying `authorize` with `supports` tags, since those two seem to do the same?

jmiranda: I guess so?

awalkowiak (approved):

PR #3109, opened 2022-04-11
OBPIH-4354: Add zone to bin modal and table

kchelstowski:
Changed Zone Type to Zone Location in modals as Katarzyna suggested


awalkowiak (changes requested):
Please change the commit message. You modified only the header label here.

PR #3110, opened 2022-04-11
OBAM-316 Add ability to select products for replenishment

DaRoTP:


src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -26,6 +27,25 @@ const FIELD = {
...
+        defaultMessage: 'Select All',
+        flexWidth: 4,
+        getDynamicAttr: ({
+          rowIndex, isAllProductsSelected, selectAllCode, updateSelectedItems,
+        }) => ({
+          headerHtml: () => (
+            <input
+              type="checkbox"
+              className="mt-1"
+              checked={isAllProductsSelected}
```
awalkowiak, line 42: Should be `allRowsSelected`, since there might be multiple occurrences of the same product and the naming here might be misleading.

src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -139,6 +174,33 @@ class CreateReplenishment extends Component {
...
   }
 
+  isAllProductsSelected() {
+    const selectedProductCount = numberOfProductsSelected(this.state.values.requirements);
+    const productCount = this.state.values.requirements.length;
+    return selectedProductCount === productCount;
+  }
+
+  selectAllCode() {
+    const isAllSelected = this.isAllProductsSelected();
```
awalkowiak, line 184: I do not understand the idea behind this.

src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -92,6 +112,10 @@ const FIELD = {
   },
 };
 
+function numberOfProductsSelected(products) {
```
awalkowiak, line 115: You don't need this (see comment in line 128)

src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -101,6 +125,14 @@ function validate(values) {
       errors.requirements[key] = { quantity: 'react.replenishment.error.quantity.label' };
     }
   });
+  if (numberOfProductsSelected(values.requirements) < 1) {
```
awalkowiak, line 128: This can be as simple as: 
```
const anyRowSelected = _.find(values.requirements, row => row.checked);
if (!anyRowSelected) {
    ...
}
```
(or if you need an `anyRowSelected` helper function for other places, then do something similar as suggested in the comment on line 331)

src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -261,8 +325,11 @@ class CreateReplenishment extends Component {
               <div className="table-form">
                 {_.map(FIELD, (fieldConfig, fieldName) =>
                   renderFormField(fieldConfig, fieldName, {
-                    updateRow: this.updateRow,
                     values,
+                    updateRow: this.updateRow,
+                    updateSelectedItems: this.updateSelectedItems,
+                    isAllProductsSelected: this.isAllProductsSelected(),
```
awalkowiak, line 331: If you need a helper function `allRowsSelected` in other places, then:
`allRowsSelected: this.allRowsSelected(values.requirements)`
and
```
allRowsSelected(rows) {
    return !_.find(rows, row => !row.checked)
}
```
otherwise:
`allRowsSelected: !_.find(values.requirements, row => !row.checked)`

awalkowiak (changes requested):
@DaRoTP there are a couple of things that should be refactored/simplified, plus there are some conflicts after merging a few other PRs.

src/js/components/replenishment/CreateReplenishment.jsx
DaRoTP: the logic behind here was:
- if at the moment all rows are **selected** then do the opposite and **deselect** all rows
- else if at the moment not all rows are selected the **select** them all

PR #3111, opened 2022-04-11
OBPIH-4341 Add option to add two-way associations

awalkowiak:


grails-app/controllers/org/pih/warehouse/product/ProductAssociationController.groovy
```diff
@@ -73,6 +73,15 @@ class ProductAssociationController {
...
         def productAssociationInstance = new ProductAssociation(params)
         if (productAssociationInstance.save(flush: true)) {
+            if (params.hasMutualAssociation) {
+                def mutualAssociationInstance = new ProductAssociation()
+                bindMutualAssociationData(mutualAssociationInstance, params)
+
+                mutualAssociationInstance.mutualAssociation = productAssociationInstance
+                productAssociationInstance.mutualAssociation = mutualAssociationInstance
+
+                mutualAssociationInstance.save(flush: true, failOnError: true)
```
jmiranda, line 83: Again we should add something to undo these changes if the user unchecks the checkbox. Ask Kelsey what she thinks.

grails-app/views/productAssociation/create.gsp
```diff
@@ -75,6 +75,15 @@
 									<g:textArea class="text" name="comments" value="${productAssociationInstance?.comments}" />
 								</td>
 							</tr>
+
+							<tr class="prop">
+								<td class="name">
+									<label><warehouse:message code="productAssociation.mutualAssociation.label" default="Two-way Association"/></label>
```
jmiranda, line 81: "Two-way association" needs some work. Bi-directional is the only thing I can think of and that's not any better.

jmiranda (approved):

grails-app/views/productAssociation/create.gsp
jmiranda: Ok sounds good. My vote would be for Bi-directional association or Symmetric association. Or even something more laymen like "Do you want to create an inverse relationship for the associated product?"

awalkowiak:
@jmiranda I added delete handling for mutual associations according to Kelsey's request here: https://pihemr.atlassian.net/browse/OBPIH-4341?focusedCommentId=131002

PR #3117, opened 2022-04-13
OBDS-129 Fix product catalog items import, disable boston, chicago wa…

hdsldv:
…rehouse inventory and invetnory levels import

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1857,15 +1857,15 @@ openboxes {
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/mainWarehouseInventory.csv"
             }
             bostonWarehouseInventory {
-                enabled = true
+                enabled = false // INFO Disabled, Boston contains duplicates of main warehouse - product and lot
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/bostonWarehouseInventory.csv"
             }
             chicagoWarehouseInventory {
-                enabled = true
+                enabled = false // INFO Disabled, Chicago contains duplicates of main warehouse - product and lot
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/chicagoWarehouseInventory.csv"
             }
             inventoryLevels {
-                enabled = true
+                enabled = false // INFO Disabled, inventory relation is missing but not defined
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/inventoryLevels.csv"
             }
             users {
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1857,11 +1857,11 @@ openboxes {
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/mainWarehouseInventory.csv"
             }
             bostonWarehouseInventory {
-                enabled = false // INFO Disabled, Boston contains duplicates of main warehouse - product and lot
+                enabled = true
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/bostonWarehouseInventory.csv"
             }
             chicagoWarehouseInventory {
-                enabled = false // INFO Disabled, Chicago contains duplicates of main warehouse - product and lot
+                enabled = true
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/chicagoWarehouseInventory.csv"
             }
             inventoryLevels {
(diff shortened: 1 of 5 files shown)


grails-app/conf/templates/configuration/users.csv
```diff
@@ -1,5 +1,5 @@
 username,firstName,lastName,email,roleType
 browser,Mr,Browser,,"ROLE_BROWSER,ROLE_FINANCE"
-purchaser,Ms,Purchaser,,"ROLE_ADMIN,ROLE_APPROVER,ROLE_FINANCE"
-accountant,Mr,Accountant,,"ROLE_ADMIN,ROLE_APPROVER,ROLE_FINANCE,ROLE_INVOICE"
-superuser,Ms,Superuser,,"ROLE_SUPERUSER,ROLE_APPROVER,ROLE_FINANCE,ROLE_INVOICE"
+purchaser,Ms,Purchaser,,"ROLE_ADMIN,ROLE_FINANCE"
```
jmiranda, line 3: Is this because there's no role in the database with roletypecode = ROLE_APPROVER?

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
```diff
@@ -54,7 +54,7 @@ class LoadDataApiController extends BaseDomainApiController {
         }
 
         if (Boolean.valueOf(config.productCatalogItems.enabled)) {
-            loadDataService.importProductCatalogItems(new URL(config.productSuppliers.url))
+            loadDataService.importProductCatalogItems(new URL(config.productCatalogItems.url))
```
jmiranda, line 57: I'm confused. How were we encountering that error if we weren't even running the import on product catalog items?

grails-app/services/org/pih/warehouse/product/ProductCatalogService.groovy
```diff
@@ -17,6 +19,10 @@ class ProductCatalogService {
 
         if (params.productCode) {
             product = Product.findByProductCode(params.productCode)
+
+            if (!product) {
+                throw new ValidationException("Could not locate product with code: " + params.productCode, null);
```
jmiranda, line 24: Not sure if we can pass null here. If we can, then I guess it's fine. Otherwise it would be better to use a different exception like IllegalArgumentException or IllegalStateException.

One other issue is that if params.productCode is empty this will also lead to a weird exception like the one we saw today. So you might want to pull this out of that code block. In fact I would check to see if there's a valid product AND product catalog just before we try to find the product catalog item on line 32.

jmiranda (changes requested):
Glad you resolved the issue. What was it exactly?

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
hdsldv: importProductCatalogItems was receiving invalid csv. It was not possible to extract productCode which lead to the error with saving product association (because product was null)

grails-app/conf/templates/configuration/users.csv
hdsldv: That's right

grails-app/conf/Config.groovy
```diff
@@ -1857,15 +1857,15 @@ openboxes {
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/mainWarehouseInventory.csv"
             }
             bostonWarehouseInventory {
-                enabled = true
+                enabled = false // INFO Disabled, Boston contains duplicates of main warehouse - product and lot
```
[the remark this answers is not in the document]
hdsldv: You are right, changed to use inventory service

grails-app/conf/Config.groovy
```diff
@@ -1857,15 +1857,15 @@ openboxes {
...
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/bostonWarehouseInventory.csv"
             }
             chicagoWarehouseInventory {
-                enabled = true
+                enabled = false // INFO Disabled, Chicago contains duplicates of main warehouse - product and lot
```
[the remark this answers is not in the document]
hdsldv: done

grails-app/conf/Config.groovy
```diff
@@ -1857,15 +1857,15 @@ openboxes {
...
                 url = "https://raw.githubusercontent.com/openboxes/openboxes/develop/grails-app/conf/templates/configuration/chicagoWarehouseInventory.csv"
             }
             inventoryLevels {
-                enabled = true
+                enabled = false // INFO Disabled, inventory relation is missing but not defined
```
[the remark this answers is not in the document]
hdsldv: Seems like inventory is required when importing inventory level. In the ticket main warehouse is mentioned, but I do not see on what conditions inventory of specific warehouse should be used

grails-app/services/org/pih/warehouse/product/ProductCatalogService.groovy
hdsldv: Changed, now product and product catalog code are checked even if codes are empty

awalkowiak (approved):

awalkowiak:
I am merging as is to unblock the testing, the inventory level issue can be resolved later on.

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
jmiranda: Oh my. I didn't even think of that. Glad it wasn't any more complicated than that. :)

PR #3119, opened 2022-04-14
OBPIH-4041 Rearrange Product Menu

kchelstowski:


grails-app/conf/Config.groovy
```diff
@@ -1508,39 +1508,34 @@ openboxes {
...
-                        [label: "product.associations.label", defaultLabel: "Products Associations", href: "/${appName}/productAssociation/list"],
+                        [label: "categories.label", defaultLabel: "Categories", href: "/${appName}/category/tree"],
+                        [label: "product.catalogs.label", defaultLabel: "Catalogs", href: "/${appName}/productCatalog/list"],
                         [label: "product.tags.label", defaultLabel: "Tags", href: "/${appName}/tag/list"],
-                        [label: "unitOfMeasure.label", defaultLabel: "Unit of Measure", href: "/${appName}/unitOfMeasure/list"],
-                        [label: "unitOfMeasureClass.label", defaultLabel: "Uom Class", href: "/${appName}/unitOfMeasureClass/list"],
-                        [label: "unitOfMeasureConversion.label", defaultLabel: "Uom Conversion", href: "/${appName}/unitOfMeasureConversion/list"]
+                        [label: "attributes.label", defaultLabel: "Attributes", href: "/${appName}/attribute/list"],
+                        [label: "product.associations.label", defaultLabel: "Products Associations", href: "/${appName}/productAssociation/list"],
```
awalkowiak, line 1526: Should be only `Associations` in the label

grails-app/i18n/messages.properties
```diff
@@ -1693,7 +1694,7 @@ product.coldChain.label=Cold chain
 product.reconditioned.label=Reconditioned
 product.coldChain.message=Cold chain
 product.components.label=Components
-product.create.label=Create new product
+product.create.label=Create product
```
awalkowiak, line 1697: Change all other translation files too

awalkowiak (changes requested):
See comments above

grails-app/conf/Config.groovy
kchelstowski: nice catch - in the messages file it was "Associations" anyway, so only default message was set as Products Associations before, so I didn't realize the default message is wrong.

awalkowiak (approved):

PR #3122, opened 2022-04-15
OBDS-74 Support building swagger clients in Java and Python; fix warnings in swagger spec.

mdpearson:


mdpearson:
@jmiranda if you could give the README (https://github.com/openboxes/openboxes/pull/3122/files#diff-bdd6b6b7fec8b3d4b479cd52b20a989ddfc0cbdc2b3222fc9b41acd403c2a047) a once-over and let me know what our grantors would thing, I'd very much appreciate it. thank you!

swagger/docs_preamble.yml
```diff
@@ -0,0 +1,81 @@
...
+    cookie:
+      in: cookie
+      name: JSESSIONID
+      type: apiKey
+servers:
+- url: https://obdev.pih-emr.org/openboxes
+  description: Dev instance
+- url: https://obnavstage.pih-emr.org/openboxes
+  description: Staging instance
+- url: https://obnav.pih-emr.org/openboxes
```
jmiranda, line 74: Let's remove the production server since we don't actually want anyone accessing that server. In fact, I would say we should have this pointing at a neutral OB server instead. But we can leave dev and staging for now.

build.gradle
```diff
@@ -10,40 +10,49 @@
...
+        mavenCentral()
     }
     dependencies {
-        classpath "org.grails:grails-gradle-plugin:$grailsVersion"
-        classpath "org.grails.plugins:hibernate5:${gormVersion-".RELEASE"}"
-        classpath "com.bertramlabs.plugins:asset-pipeline-gradle:2.15.1"
-        classpath "gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:2.2.0"
-        classpath 'org.grails.plugins:database-migration:3.1.0.RC1'
-        classpath "gradle.plugin.com.virgo47.gradle:ClasspathJar:1.0.0"
-
```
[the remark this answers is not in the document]
jmiranda: Oh is this just for the build dependencies?

build.gradle
```diff
@@ -10,40 +10,49 @@
...
-//grails.project.test.reports.dir = "target/test-reports"
-//
-//grails.plugin.location.liquibase = "liquibase/"
+apply plugin: "com.bertramlabs.asset-pipeline"
+apply plugin: "io.swagger.core.v3.swagger-gradle-plugin"
+apply plugin: "org.grails.grails-gsp"
+apply plugin: "org.grails.grails-web"
 
-version "0.9.0-SNAPSHOT"
+version "0.9.0-SNAPSHOT"  // FIXME get this from git
```
jmiranda, line 49: I don't think that's true. this is a pre-release version. version tags don't exist until after we release.

build.gradle
```diff
@@ -10,40 +10,49 @@
...
 group "com.openboxes"
+sourceCompatibility = 1.8
+
+File secretsFile = file("${System.properties["user.home"]}/build.secrets")
+def secrets = new Properties()
+secrets.load(new FileInputStream(secretsFile))
```
jmiranda, line 55: Is this the right approach? Or do we need to add a ticket to look into a proper secrets / vault solution?

jmiranda (approved):
This is awesome. Approved with a few comments.

build.gradle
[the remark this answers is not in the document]
mdpearson: Hi Justin, yeah, it's somewhat confusing when rendered as a diff.

Basically, I used the new [Plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) wherever possible, for brevity (it's generally just one line, instead of several scattered about the file in the old `apply plugin:` syntax). I thought this would make merging changes from the v1 branch easier, but I could be wrong.

There are a few cases where I still use the old syntax, especially where the version number is parametrized. You can see this in e.g. the `swaggerCoreVersion` field which applies to both a plugin and a handful of dependencies. Using a variable here should prevent version skew and maybe even the occasional merge bomb.

No dependency was taken away or changed, just refactored.

build.gradle
mdpearson: I know in the python world there are plugins that if you're a descendant, of, say, 1.0.1, will generate build numbers like 1.0.2dev1 -- that's what I had in mind here

mdpearson: but it's a minor nit and I can remove the comment if you prefer

build.gradle
mdpearson: oh, I think a proper solution is worth looking into for sure, but it's probably not on the top-ten list. I just didn't want to commit any secrets to git and this was a quick and reasonably clean way to do so for the POC. Should I file a ticket?

PR #3129, opened 2022-04-19
OBAM-334 Add validation to replenishment pick page

kchelstowski:


kchelstowski:
The idea of having a helper "updating" parameter in the payload was for `validateRequirement` function, because without it, if you for example choose an item with qty to transfer = qty available and click "Next" on the check page, it will once again get the qty available from productAvailibilityService, so it would be 0 and the validation wouldn't pass. That's why I have put `+ item quantity` if you update (updating the replenishment happens if you are on either 2nd or 3rd page and go next).

grails-app/services/org/pih/warehouse/replenishment/ReplenishmentService.groovy
```diff
@@ -162,9 +162,15 @@ class ReplenishmentService {
         return order
     }
 
-    void validateRequirement(ReplenishmentItem item) {
-        if (item.quantity > productAvailabilityService.getQuantityAvailableToPromiseByProductNotInBin(item.location, item.binLocation, item.product)) {
-            throw new ValidationException("There is not available that quantity of the product with id: ${item.product.id}")
+    void validateRequirement(ReplenishmentItem item, Boolean updating = false) {
+        if (updating) {
+            if (item.quantity > productAvailabilityService.getQuantityAvailableToPromiseByProductNotInBin(item.location, item.binLocation, item.product) + item.quantity) {
```
awalkowiak, line 167: The `+ item.quantity` should rather be a sum of quantity from item.picklistItems and make it a bit more readable:
```
def quantityAvailableToPromise = productAvailabilityService.getQuantityAvailableToPromiseByProductNotInBin(item.location, item.binLocation, item.product)
def quantityPicked = sum of quantitiy from item.picklistItems
def quantityAvailableToPromiseWithPicked = quantityAvailableToPromise + quantityPicked
```

src/js/components/replenishment/ReplenishmentCheckPage.jsx
```diff
@@ -209,8 +209,12 @@ class ReplenishmentSecondPage extends Component {
     if (this.state.values.replenishment.status === 'PLACED') {
       this.props.showSpinner();
       const url = `/openboxes/api/replenishments/${this.props.match.params.replenishmentId}`;
-      const payload = { status: 'COMPLETED' };
-      apiClient.post(url, payload)
+      const payload = {
+        status: 'PLACED',
+        replenishmentItems: this.state.values.replenishment.replenishmentItems,
+        updating: true,
```
awalkowiak, line 215: no need for this

src/js/components/replenishment/ReplenishmentSecondPage.jsx
```diff
@@ -250,8 +250,12 @@ class ReplenishmentSecondPage extends Component {
     if (this.state.values.replenishment.status === 'PENDING') {
       this.props.showSpinner();
       const url = `/openboxes/api/replenishments/${this.props.match.params.replenishmentId}`;
-      const payload = { status: 'PLACED' };
-      apiClient.post(url, payload)
+      const payload = {
+        status: 'PLACED',
+        replenishmentItems: this.state.values.replenishment.replenishmentItems,
+        updating: true,
```
awalkowiak, line 256: no need for this

awalkowiak (changes requested):

PR #3131, opened 2022-04-19
OBDS-129 Fix inventory and inventory levels import

hdsldv:


grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
```diff
@@ -62,19 +63,31 @@ class LoadDataApiController extends BaseDomainApiController {
         }
 
         if (Boolean.valueOf(config.mainWarehouseInventory.enabled)) {
-            loadDataService.importInventory(new URL(config.mainWarehouseInventory.url))
+            loadDataService.importInventory(
+                    new URL(config.mainWarehouseInventory.url),
+                    Location.findByName('Main Warehouse')
```
awalkowiak, line 68: Let's move the names of warehouses to the config to be alongside URLs.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/LoadDataApiController.groovy
hdsldv: done

PR #3138, opened 2022-04-21
OBDS-129 Add boston stocklist import

hdsldv:


awalkowiak (approved):
It would be better to have it more generic as a list of stocklist to be imported, but as we are tight on schedule, we can live with this for now.

PR #3139, opened 2022-04-21
OBPIH-4376 Make liquibase migrations resilient when a transient table doesn't exist

mdpearson:


awalkowiak (approved):

jmiranda:
@awalkowiak @mdpearson This looks ok to me too. Are there circumstances where this would be bad? For example, let's say the view creation fails and that failure goes unnoticed by the user. Eventually if they go to view expiry data, the values will probably be 0, but the user won't know that this is due to a failed create view statement. And I'm not asking to solve that problem right now with this PR, but we need to identify and articulate what problems this may cause so we can remedy them in the future.

jmiranda (approved):

mdpearson:
@jmiranda I filed a ticket, OBPIH-4395, to look into downstream effects more carefully. For now, develop can't start from a clean database without this patch -- @awalkowiak, it should probably go in soon.

PR #3141, opened 2022-04-22
OBAM-326 Requests from ward should change based on push or pull stock list

DaRoTP:
Requests from ward should change based on push or pull stock list

### DIFF
--- a/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
@@ -26,6 +26,156 @@ import Translate, { translateWithDefaultMessage } from 'utils/Translate';
 
 import 'react-confirm-alert/src/react-confirm-alert.css';
 
+function addButton({
+  // eslint-disable-next-line react/prop-types
+  addRow, getSortOrder, newItemAdded, updateTotalCount,
+}) {
+  return (
+    <button
+      type="button"
+      className="btn btn-outline-success btn-xs"
+      onClick={() => {
+        updateTotalCount(1);
+        addRow({ sortOrder: getSortOrder() });
+        newItemAdded();
+      }}
+    ><span><i className="fa fa-plus pr-2" /><Translate id="react.default.button.addLine.label" defaultMessage="Add line" /></span>
+    </button>
+  );
+}
+
+
+const FIELDS = {
+  product: {
+    type: SelectField,
+    label: 'react.stockMovement.requestedProduct.label',
+    defaultMessage: 'Requested product',
+    headerAlign: 'left',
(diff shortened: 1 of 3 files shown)


DaRoTP:
I have noticed that there were a lot of duplicate fields in many variations of the fields objects like **product**, **quantityOnHand** _etc..._ so I moved most of the fields to a **FIELDS** object and then assign them appropriately  like
```
STOCKLIST_FIELDS_PUSH_TYPE = {
    product: FIELDS.product,
   ....
}
```

in case of adding/overwriting different propperites we can just use a **spread operator** and assign a new property

```
STOCKLIST_FIELDS_PUSH_TYPE = {
    product: { 
        ...FIELDS.product,
        headerAlign: 'left'
    },
   ....
}
```

DaRoTP:
Also it might be a good idea to refactor **AddItemsPage** component and split it into smaller sub-components because the component itself is already quite big and there are many **if conditions** in methods that make it quite hard to comprehend the logic

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -879,27 +595,39 @@ class AddItemsPage extends Component {
...
-              quantityRequested || demandPerReplenishmentPeriod - quantityOnHand;
+          const {
+            quantityRequested,
+            demandPerReplenishmentPeriod,
+            quantityOnHand,
+            quantityAllowed,
+          } = val;
+
+          let qtyRequested = 0;
+          if (qtyRequested) qtyRequested = quantityRequested;
```
awalkowiak, line 614: This part is weird. Will it be ever entered?
Why you can't do a simple:
` const qtyRequested = isPullType ? (demandPerReplenishmentPeriod - quantityOnHand) : quantityAllowed`?

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
DaRoTP: no, it seems that it will never enter, this is a mistake from my part.
Instead this is supposed to be:

```js
  let qtyRequested = 0;
  if (quantityRequested) qtyRequested = quantityRequested;
  else if (isPullType) qtyRequested = demandPerReplenishmentPeriod - quantityOnHand;
  else qtyRequested = quantityAllowed;
```

since previously there was a _two way_ condition  
```js
  const qtyRequested =
      quantityRequested || demandPerReplenishmentPeriod - quantityOnHand;
```
and now I also need to set **qtyRequested** to **quantityAllowed** if _request type_ is **Push**

awalkowiak (approved):

PR #3143, opened 2022-04-25
OBAM-331 Add reorder report

hdsldv:


grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -17,11 +17,7 @@ import groovy.time.TimeCategory
 import org.apache.commons.collections.FactoryUtils
 import org.apache.commons.collections.list.LazyList
 import org.apache.commons.lang.StringEscapeUtils
-import org.pih.warehouse.core.ActivityCode
-import org.pih.warehouse.core.Constants
-import org.pih.warehouse.core.Location
-import org.pih.warehouse.core.Tag
-import org.pih.warehouse.core.User
+import org.pih.warehouse.core.*
```
awalkowiak, line 20: Do not use wildcard imports

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -10,11 +10,7 @@
 package org.pih.warehouse.core
 
 import org.grails.plugins.csv.CSVWriter
-import org.pih.warehouse.inventory.InventoryItem
-import org.pih.warehouse.inventory.InventoryLevel
-import org.pih.warehouse.inventory.InventoryStatus
-import org.pih.warehouse.inventory.TransactionCode
-import org.pih.warehouse.inventory.TransactionEntry
+import org.pih.warehouse.inventory.*
```
awalkowiak, line 13: Do not use wildcard imports

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
hdsldv: done

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
hdsldv: done

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -565,6 +565,66 @@ class InventoryController {
...
+        def inventoryItems = dashboardService.getReorderReport(location)
+
+        String filename = "Reorder report - " + location.name + ".csv"
+
+        response.setHeader("Content-disposition", "attachment; filename=\"${filename}\"")
+
+        def hasRoleFinance = userService.hasRoleFinance(session.user)
+
+        def csv = ""
+        csv += '"' + "${warehouse.message(code: 'inventoryLevel.status.label')}" + '"' + ","
```
jmiranda, line 579: Probably would have been better to use the dataService.generateCsv() and use StringEscapeUtils.escapeCsv() for wrapping all values in double quotes. But I assume we were working off an older version that I wrote so I'll let it pass.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -554,6 +554,50 @@ class DashboardService {
...
+        ArrayList<Map> inventoryItems = getInventoryItems(location)
+
+        ArrayList<Map> reorderStock = inventoryItems.findAll { Map inventoryItem ->
+            def inventoryLevel = inventoryItem?.inventoryLevel as InventoryLevel
+
+            if (inventoryLevel?.binLocation != null) {
+                return false;
+            }
+
+            def quantityATP = inventoryItem.quantityAvailableToPromise as Number
```
jmiranda, line 568: Nitpick: I used "quantity available to promise" because it's the most precise naming for this value. However, we've standardized on quantityAvailable in most places, so I would continue that trend in order to avoid confusion. We should also consider converting all "quantityAvailableToPromise" references to "quantityAvailable". But we can leave that for now.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -554,6 +554,50 @@ class DashboardService {
         return reorderStock
     }
 
+    def getReorderReport(Location location) {
+        long startTime = System.currentTimeMillis()
+        ArrayList<Map> inventoryItems = getInventoryItems(location)
+
+        ArrayList<Map> reorderStock = inventoryItems.findAll { Map inventoryItem ->
+            def inventoryLevel = inventoryItem?.inventoryLevel as InventoryLevel
```
jmiranda, line 562: This is expensive, so I would consider doing this using a join in the getInventoryItems method. That's tricky because you have to do some domain mapping in the criteria query (or HQL).

Also getInventoryItems should probably be changed to getAvailableItems and should return the AvailableItem instead. Not a big deal for now, but we should clean that up when we get a chance.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -554,6 +554,50 @@ class DashboardService {
...
+
+            if (inventoryLevel?.binLocation != null) {
```
jmiranda, line 564: Add a comment (for future you's sake) to explain why we're doing this. It's probably not obvious to a develop coming in cold.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -554,6 +554,50 @@ class DashboardService {
...
+                return false;
+            }
+
+            def quantityATP = inventoryItem.quantityAvailableToPromise as Number
+
+            def minQuantity = inventoryLevel?.minQuantity
+            if (minQuantity && quantityATP >= minQuantity) {
```
jmiranda, line 571: Be careful. Please carefully consider the case where minQuantity = 0.If you intend the first expression to be minQuantity != null, that is not what you're getting and it could be short-circuiting on that expression. I don't have capacity for the mental gymnastics required here, but just know that it has bit me in the past.

PR #3145, opened 2022-04-27
OBAM-335 UI updates to ward pull/adhoc request

kchelstowski:
I spent a long time trying to figure out how to make the red asterisk for header - the same asterisk which is visible for "regular" fields with required: true. It was supposedly made and was used in many components (talking about fields, which have another nestled fields inside (ArrayField)), but it did not really work with current version, so even though it was used in many places, the asterisk didn't work.
With my changes, the red asterisk for headers are now "generic", which means they can be used anywhere (and will now finally work in places, where it was previously declared (e.g. stock-movement/combined-shipments/AddItemsPage.jsx) and many others, so I hope my changes will also improve other wizards.

### DIFF
--- a/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
@@ -379,6 +379,7 @@ const REQUEST_FROM_WARD_STOCKLIST_FIELDS_PULL_TYPE = {
       product: {
         ...FIELDS.product,
         fieldKey: 'disabled',
+        flexWidth: '2',
         getDynamicAttr: ({
           fieldValue, debouncedProductsFetch, rowIndex, rowCount, newItem,
         }) => ({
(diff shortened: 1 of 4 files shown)


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -617,9 +743,9 @@ class AddItemsPage extends Component {
 
           return {
             ...val,
+            quantityOnHand: '',
             disabled: true,
             quantityRequested: qtyRequested >= 0 ? qtyRequested : 0,
-            quantityOnHand: isPullType ? quantityOnHand : undefined,
```
awalkowiak, line 622: Why did you make this change?

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -716,6 +842,14 @@ class AddItemsPage extends Component {
       if (date.diff(dateRequested) > 0) {
         errors.lineItems[key] = { expirationDate: 'react.stockMovement.error.invalidDate.label' };
       }
+
+      if (this.state.isRequestFromWard) {
+        const isNumeric = n => !Number.isNaN(parseInt(n, 10));
```
awalkowiak, line 847: Why the need for this?

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -716,6 +842,14 @@ class AddItemsPage extends Component {
...
+        const isPositive = n => isNumeric(n) && n >= 0;
+        if (!item.quantityOnHand || !isPositive(item.quantityOnHand)) {
```
awalkowiak, line 849: Why you can't simply do `item.quantityOnHand >= 0`?

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
kchelstowski: It was said that the qoh shall be blank, would it be blank in any circumstances if I had left it as it was?

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -716,6 +842,12 @@ class AddItemsPage extends Component {
       if (date.diff(dateRequested) > 0) {
         errors.lineItems[key] = { expirationDate: 'react.stockMovement.error.invalidDate.label' };
       }
+
+      if (this.state.isRequestFromWard) {
+        if (!item.quantityOnHand || !(item.quantityOnHand >= 0)) {
```
awalkowiak, line 847: I think `item.quantityOnHand < 0` would be less confusing

awalkowiak (changes requested):

awalkowiak (approved):

PR #3146, opened 2022-04-27
OBPIH-4379, OBDS-137 Localize and restyle HelpScout

mdpearson:
* OBDS-137 Make HelpScout beacon's language track current locale.
* OBPIH-4379 Remove HelpScout's floating button from bottom-right of page.
* OBPIH-4379 Use color to highlight support button; style banner consistently.

(OBDS-137 depends on changes in OBPIH-4379 so I'm submitting them as one PR.)

I have a couple questions about how to wire messages.properties to a gsp page, would love your advice.


grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -58,7 +59,58 @@ class LocalizationApiController {
         Locale locale = localizationService.getLocale(languageCode)
         String message = messageSource.getMessage(params.id, params.list("args").toArray(), locale)
         render([code: params.id, message: message, currentLocale: locale] as JSON)
-
     }
 
+    static JSON localizeHelpScoutLabels(GrailsApplication app, Locale locale) {
```
mdpearson, line 64: I couldn't get grails to recognize this method from a gsp without making it static.

Not wild about that; the preceding functions can call e.g. `localizationService.getCurrentLocale()`, but the caller has the info immediately at hand so it's not the end of the world. Still, I'm hoping there's a more grails-y way to do this.

grails-app/views/common/_header.gsp
```diff
@@ -1,5 +1,43 @@
...
+      window.Beacon("config", {
+        "color": "${grailsApplication.config.openboxes.helpscout.widget.color}",
+        "enableFabAnimation": false
+      })
+      /*
+       * Use a scriptlet block to prevent html-escaping localization labels, which are JSON.
+       * More modern grails versions have a built-in raw() method to do this, but...
+       * https://stackoverflow.com/questions/1337464/overriding-grails-views-default-codec-html-config-back-to-none
+       */
+      window.Beacon("config", <%=LocalizationApiController.localizeHelpScoutLabels(grailsApplication, session?.user?.locale)%>)
```
mdpearson, line 36: The setting `grails.views.default.codec = "html"` in _Config.groovy_ html-escapes the json I need to send to `window.Beacon`, which totally breaks the page.

The linked SO article talks about using `<%=STATEMENT%>` to inject data without reformatting it. Localization information will never come from a user so I don't think there are any XSS, etc. issues with doing it in this way, but if there's a better way, I'd be all ears.

web-app/css/openboxes.css
```diff
@@ -2198,17 +2198,24 @@ img.photo_tiny {
 
 }
 
+.banner-menu,
+.banner-menu button,
+.banner-menu input {
+    height: 30px;
```
mdpearson, line 2204: this little bit of CSS properly aligns the search box and buttons on the banner.

web-app/css/openboxes.css
```diff
@@ -2873,6 +2881,79 @@ span.canceled, div.canceled {
...
+
+.button.helpscout:active i,
+.button.helpscout.active i {
+    filter: drop-shadow(-1px -1px 0 #226866); /* apply button's text-shadow to fontawesome */
+}
+
+/* HelpScout's default positioning wastes space and clobbers page content */
+.BeaconContainer {
+    right: 5px !important;
+    bottom: 5px !important;
```
mdpearson, line 2954: I hate using `!important` but for the life of me could not find another way to override HelpScout's JS-based positioning rules.

grails-app/conf/Config.groovy
```diff
@@ -1171,17 +1171,17 @@ openboxes {
...
+// Zopim widget
 openboxes.zopim.widget.enabled = false
 openboxes.zopim.widget.url = "//v2.zopim.com/?2T7RMi7ERqr3s8N20KQ3wOBRudcwosBA"
 
 // HelpScout beacon
-openboxes.helpscout.widget.color = "#ff9800"
+openboxes.helpscout.widget.color = "#3AB4B1"
 openboxes.helpscout.widget.enabled = true
-openboxes.helpscout.widget.key = "5eafcb3f-9e7e-4943-8ebb-669f60a696cd"
+openboxes.helpscout.widget.key = "44ee4f01-5334-4b93-ad25-03037903eb80"
```
jmiranda, line 1184: Does this expose us in any way?

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
jmiranda: I'm confused. Why couldn't we use ${g.message()} in the GSP or even here?

jmiranda: You can also use messageSource (see ApiController.getAppContext).

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -58,7 +59,58 @@ class LocalizationApiController {
...
+        return [
+            "labels": [
+                /* fields in the Contact Us pane */
+                "messageButtonLabel"     : app.mainContext.getMessage("helpscout.beacon.contact.messageButtonLabel.label", null, locale),
```
jmiranda, line 68: Is messageButtonLabel the name that helpscout is expecting in the config?

If so you could probably do this in a more systematic way.

message.properties
```
helpscout.messageButtonLabel.label = ...
helpscout.wereHereToHelp.label = ...
```

Code to build config that can be returned by the
```
def getHelpScoutConfig() { 
    Map configMap = [:]
    // LocalizationApiController could be refactored so we can move most of the code to messageService
    List messageProperties = messageService.getMessagePropertiesWithPrefix("helpscout")
    messageProperties.collect { it ->  
        // Property name is messageButtonLabel, wereHereToHelp, etc
        // You could also pass a regex with pattern matching instead of a prefix so that you can 
        // just pluck out the matching pattern out as the property name 
        String propertyName = extractPropertyName(it.key)   // remove helpscout and label
        configMap(propertyName, it.value)
    }    
    return configMap
```

grails-app/views/common/_header.gsp
jmiranda: Could this have been an API request that returns the labels as data?

jmiranda (approved):
I was confused why you needed to use a scriptlet. I feel like it would have been easier to make an API call to return the config data you needed. In fact, you could move the config to the config map returned by getAppContext() and have javascript make that request on every page. We're already doing that on the React side.

Or you could just build that config map in javascript. I don't love this but it at least doesn't require a workaround using scriptlets.
```
var config = whateveryoudotocreateamapinjavascript
config.put("labelHere", "${g.message('helpscout.myLabel.label')")
window.Beacon("config", config)
```

I am in no way going to keep this from being merged because of it.

awalkowiak (approved):

PR #3149, opened 2022-04-28
OBPIH-4378 &  OBPIH-4189 Landing page to see submitted requests

DaRoTP:
https://pihemr.atlassian.net/browse/OBPIH-4378
https://pihemr.atlassian.net/browse/OBPIH-4189


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -410,8 +410,14 @@ class StockMovementService {
             if(params.createdBefore) {
                 le("dateCreated", params.createdBefore)
             }
+            def orderBy = "dateCreated"
 
-            order("dateCreated", "desc")
+            if(params.orderBy == "requisition.dateRequested") {
```
DaRoTP, line 415: this might not be the best solution for ordering by a nested value but I couldn't find a better one.

awalkowiak: I think other option here would be:
```
if (params.orderBy == "requisition.dateRequested") {
    requisition {
        order("dateRequested", desc)
    }
} else {
    order("dateCreated", "desc")
}

```
I think it has a slightly better readability.

awalkowiak (changes requested):

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -43,6 +43,18 @@ class AuthTagLib {
         }
     }
 
+    def isNonInventoryManaged = {attrs, body ->
```
awalkowiak, line 46: Technically the sole `isNonInventoryManaged` is not true here because it needs `SUBMIT_REQUEST` too. `isNonInventoryManagedAndCanSubmitRequest` would be better in this case. However, I am thinking now that it might be useful for other places to add a tag `notSupports` and then in the template do:
```
<!-- Check if location does NOT support MANAGE_INVENTORY, but supports SUBMIT_REQUEST, then display no menu -->
<g:notSupports activityCode="${ActivityCode.MANAGE_INVENTORY}">
    <g:supports activityCode="${ActivityCode.SUBMIT_REQUEST}">
        <g:if test="${megamenuConfig.dashboard.enabled}">
            <g:isUserInRole roles="[RoleType.ROLE_BROWSER]">
                <li class="mm-item">
                    <g:link controller="dashboard" action="index" class="mm-item-link">
                        <warehouse:message code="dashboard.label" />&nbsp;
                    </g:link>
                </li>
            </g:isUserInRole>
        </g:if>
    </g:supports>
</g:notSupports>
```

awalkowiak (changes requested):

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
DaRoTP: yes, at first I thought about the same solution and it would have been easier if condition for **rendering** the menu was 

if does **NOT support** _MANAGE_INVENTORY_  [AND]()  supports _SUBMIT_REQUEST_ => **render menu**
else => don't render

but I have to consider the opposite, if this condition is met then don't render the menu which would mean that I have to negate the condition which results in

not( not support(_MANAGE_INVENTORY_) and  support(_SUBMIT_REQUEST_)  )
 => 
support(_MANAGE_INVENTORY_) or  not support(_SUBMIT_REQUEST_)  

and I am not entirely sure how to implement this **OR** logic for **supports or does not support**

that's why I chose to create a specific checker function for this case

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -43,6 +43,18 @@ class AuthTagLib {
         }
     }
 
+    def isNonInventoryManagedAndCanSubmitRequest = {attrs, body ->
```
awalkowiak, line 46: Please do one more rename here from `isNonInventoryManagedAndCanSubmitRequest` to `hideIfIsNonInventoryManagedAndCanSubmitRequest`

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -43,6 +43,18 @@ class AuthTagLib {
...
+        def warehouseInstance = Location.get(session.warehouse.id)
+        def authorized = true
+
+        authorized = authorized && !warehouseInstance.supports(ActivityCode.MANAGE_INVENTORY)
+        authorized = authorized && warehouseInstance.supports(ActivityCode.SUBMIT_REQUEST)
+
+        if (!authorized) {
```
awalkowiak, line 53: Add a comment for this line in the code, that `If the location does not have MANAGE_INVENTORY activity code and have SUBMIT_REQUEST activity code, then return empty body`

awalkowiak (changes requested):

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -43,6 +43,19 @@ class AuthTagLib {
         }
     }
 
+    def hideIfIsNonInventoryManagedAndCanSubmitRequest = {attrs, body ->
+        def warehouseInstance = Location.get(session.warehouse.id)
+        def authorized = true
+
+        authorized = authorized && !warehouseInstance.supports(ActivityCode.MANAGE_INVENTORY)
+        authorized = authorized && warehouseInstance.supports(ActivityCode.SUBMIT_REQUEST)
```
jmiranda, line 51: Why can't this be ...

```
boolean authorized = !warehouseInstance.supports(ActivityCode.MANAGE_INVENTORY) 
&& warehouseInstance.supports(ActivityCode.SUBMIT_REQUEST)?:true
```

PR #3157, opened 2022-04-29
OBPIH-4388 Prevent spreadsheet software from trimming leading 0's from ID's in PO downloads.

mdpearson:


awalkowiak (approved):

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -43,7 +43,7 @@ class CSVUtils {
         if (StringUtils.isBlank(s)) {
             throw new IllegalArgumentException("${fieldName} is empty or unset")
         }
-        DecimalFormat format = DecimalFormat.getNumberInstance(LocalizationUtil.localizationService.currentLocale)
+        DecimalFormat format = DecimalFormat.getNumberInstance(LocalizationUtil.localizationService.currentLocale) as DecimalFormat
```
jmiranda, line 46: I missed this the first time around, but this factory method might get expensive over a large data set. We can leave it for now, but we should figure out a better way of handling the repeated instantiation of a formatter (i.e. singleton instances with caching by locale?).

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -52,7 +52,7 @@ class CSVUtils {
          */
         try {
             return format.parse(s.replace("†", ""))
-        } catch (Exception e) {
+        } catch (Exception ignored) {
```
jmiranda, line 55: Why is this called ignored?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -63,11 +63,49 @@ class CSVUtils {
...
+             * https://superuser.com/questions/318420/formatting-a-comma-delimited-csv-to-force-excel-to-interpret-value-as-a-string
+             *
+             * Unfortunately, no other spreadsheet software does this :-(
+             *
+             * The following kludge plays nicely enough with Excel, Google
+             * Sheets, LibreOffice, and Numbers (Mac), but it could easily
+             * confuse other CSV parsers. (Not sure what we can do: Excel
+             * is our most important client software, by far.)
+             */
+            return "=\"${id}\"".toString()
```
jmiranda, line 99: I don't mind this, but I believe this is handled by the StringEscapeUtils 

https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html#escapeCsv(java.lang.String)

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
jmiranda, line 46: Can you explain why the coersion is necessary?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -140,7 +181,7 @@ class CSVUtils {
         // FIXME default value should be localized, but presently is "EA" everywhere
         def numerator = quantityUom ?: "EA"
         def denominator = formatInteger(quantityPerUom ?: 1)
-        return "${quantityUom}/${quantityPerUom}"
+        return "${numerator}/${denominator}"
```
jmiranda, line 184: Nitpick - this isn't really a case of numerator vs denominator. It's a standard uom notation. for example, "BX/10" is a more succinct version of "1 Box of 10 EA" or "1 Box = 10 EA".

jmiranda (commented):
Or if that doesn't work something else that preserves the value as a string. In Excel/Libreoffice that's as simple as prepending a single quote (').

I've ranted about similar issues in the past (Excel does whatever the ... it wants)
https://pihemr.atlassian.net/browse/OBPIH-1395 

My only concern with formatting these strings is that we need to make sure we preserve the string as-is, every character (even whitespace) because importing requires these numbers to be exact. So while we are solving for the

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
mdpearson: @jmiranda I'm afraid it's worse than it appears.

This returned string actually is intended as _input_ to the StringEscapeUtils, which the caller takes care of.

The ultimate contents of a csv record for a field that starts with a 0 ends up being `"=""00912"""`. If you just call escapeCSV, Excel will ignore your attempted escape, treat the ID as a number, and trim the leading 0's. :-(

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
mdpearson: Easy to change it to e.g. `return "${formattedUom}/${formattedQuantity}"`

mdpearson:
> Honestly, I think we just need to treat these fields as strings like I've done with other CSV export
> 
> ```
> StringEscapeUtils.escapeCsv(value.toString())
> ```
> 
> See dataService.generateCsv()

Unfortunately that approach doesn't work with Excel, which will dutifully try to parse out a number, even from a quoted string. (I wonder if we're talking about the same issue here, because escapeCsv won't generally add quotes for a numeric string.)

The most compatible approach I could find is what's presented here: wrap the value with quotes and prefix it with an equals sign. Despite not being able to find official documentation on the topic, Numbers, Excel, Google Sheets and LibreOffice all parse it correctly.

I took a look at the overhead of getting localized number instances per-call. 10,000 of those calls take roughly 3-700 ms. If you had a file with 10000 records, with, say, 5 numeric or currency fields per record, you'd probably waste a couple seconds. Let me see how yucky a cache would be.

mdpearson:
Hi Justin, [this](https://docs.groovy-lang.org/2.4.9/html/gapi/groovy/transform/Memoized.html) looks like just the ticket, but we can't use it till grails 3. Thoughts?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
mdpearson: resolved in dc92b52b2

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
mdpearson: resolved in 3ddb00d5c

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
mdpearson: Removed the name entirely in 3ddb00d5c.

mdpearson:
Closing this until I have the time and inclination (and ideally user complaint) to make OB correctly import data formatted in this way.

PR #3160, opened 2022-04-30
OBKN-602 Fix stock movement list endpoint parameters

awalkowiak:


jmiranda:
@awalkowiak I merged the one to develop. Was that ok or do we want to roll those changes back?

PR #3164, opened 2022-05-04
OBAM-321 Remove requisition templates from results

awalkowiak:
plus fix year intervals processing


jmiranda (commented):
LGTM - I'd like to refactor the year handling to hide it behind a method whose name provides enough context for us to know what's going on. But this is fine for now since it fixes the bug.

PR #3169, opened 2022-05-05
OBAM-314 Generate a transfer order PDF

DaRoTP:
https://pihemr.atlassian.net/browse/OBAM-314

- Implement the same PDF printout that is available for a stock transfer with picklist
- Add button for “Print Transfer Order” to pick page

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
@@ -9,6 +9,8 @@
  **/
 
 package org.pih.warehouse.inventory
+import org.pih.warehouse.order.Order
+import org.pih.warehouse.picklist.Picklist
 
 class ReplenishmentController {
 
@@ -18,5 +20,18 @@ class ReplenishmentController {
 
     def create = {
         render(template: "/common/react", params: params)
+    }
+
+    def print = {
+        Order transferOrder = Order.get(params.id)
+        def picklist = Picklist.findByOrder(transferOrder)
+
+        def lineItems = transferOrder.orderItems.findAll { !it.parentOrderItem }.sort { it.product.name }
+
+        def headerItems = [orderNumber: transferOrder.orderNumber,
+                           createdBy: transferOrder.createdBy,
+                           dateCreated: transferOrder.dateCreated]
+
+        [lineItems: lineItems, picklist: picklist, headerItems: headerItems]
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
@@ -28,7 +28,7 @@ class ReplenishmentController {
 
         def zoneNames = orderItems?.collect { it?.originBinLocation?.zone?.name }?.unique()?.sort { a, b -> !a ? !b ? 0 : 1 : !b ? -1 : a <=> b }
         def orderItemsByZone = orderItems?.groupBy { it?.originBinLocation?.zone?.name } ?: [:]
-        def allPickListItems = orderItems*.retrievePicklistItems()?.flatten()
+        def pickListItems = orderItems*.retrievePicklistItems()?.flatten()?.findAll { it.quantity > 0 }.groupBy { it?.orderItem }
 
         Map<Object, Map> itemsMap = [:]
         zoneNames.each { zoneName ->
@@ -38,13 +38,12 @@ class ReplenishmentController {
                     'hazardousMaterial'  : orderItemsByZone[zoneName].findAll { it?.product['hazardousMaterial'] },
                     'generalGoods'       : orderItemsByZone[zoneName].findAll { !it?.product['coldChain'] && !it?.product['controlledSubstance'] && !it?.product['hazardousMaterial'] },
             ]
-            itemsMap.put(zoneName, [lineItems: groupedLineItemsMap, pickListItems: allPickListItems]);
+            itemsMap.put(zoneName, [lineItems: groupedLineItemsMap, pickListItems: pickListItems]);
         }
-
         def headerItems = [
                 orderNumber: transferOrder.orderNumber,
                 createdBy  : transferOrder.createdBy,
-                dateCreated: transferOrder.dateCreated
+                dateCreated: transferOrder.dateCreated.format('MM/dd/yyyy')
         ]
 
         [itemsMap: itemsMap, headerItems: headerItems]
--- a/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
@@ -10,7 +10,6 @@
 
 package org.pih.warehouse.inventory
 import org.pih.warehouse.order.Order
-import org.pih.warehouse.picklist.Picklist
 
 class ReplenishmentController {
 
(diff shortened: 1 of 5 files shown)


DaRoTP:
Me and @awalkowiak thought about making this component somewhat generic for requesitions and for orders.
In both cases we pass [lineItems in the model](https://github.com/openboxes/openboxes/pull/3169/files#diff-94fbea89011b3f7d0c2794dee8056647ea4391623672fb8e76d34d0434868753R35) which are either **orderItems** or **requesitionItems** and the only difference would be the header that could also be passed in the model as a **headerTemplateName** and Map with appropriate parameters like ...
```groovy
[headerTemplateName: requisitionPrintHeader, headerParams: [ createdBy: createdBy,  orderNumber: orderNumber]]
```
and later render this template 

```groovy
<g:render template="${requisitionPrintHeader}" model="${headerParams}"/>
```
image



what do you think @jmiranda ?

grails-app/controllers/org/pih/warehouse/inventory/ReplenishmentController.groovy
```diff
@@ -19,4 +21,17 @@ class ReplenishmentController {
...
         render(template: "/common/react", params: params)
     }
+
+    def print = {
+        Order transferOrder = Order.get(params.id)
+        def picklist = Picklist.findByOrder(transferOrder)
+
+        def lineItems = transferOrder.orderItems.findAll { !it.parentOrderItem }.sort { it.product.name }
+
+        def headerItems = [orderNumber: transferOrder.orderNumber,
```
jmiranda, line 31: Minor nitpick ... move orderNumber to the next line and the trailing bracket to a new line. Then reformat.

grails-app/views/replenishment/_printPage.gsp
```diff
@@ -0,0 +1,103 @@
...
+                    <td colspan="10" class="middle center">
+                        <span class="fade">
+                            <warehouse:message code="default.none.label"/>
+                        </span>
+                    </td>
+                </tr>
+            </g:unless>
+            <g:each in="${lineItems}" status="i" var="lineItem">
+
+                <g:set var="picklistItems" value="${pickListItemsByOrder[lineItem.id]?.findAll { it.quantity > 0 }}"/>
```
jmiranda, line 36: This is what I was trying to avoid. I want the GSP to be as dumb as possible, so all of this data manipulation and transforming done here and in the next GSP should be handled in the controller or service layer. 

The GSP should be as simple as 
```
<g:each var="picklistItem" in="${picklistItems}">
    <tr>
        <td>${picklistItem.column1}</td>
        <td>${picklistItem.column2}</td>
        <td>${picklistItem.column3}</td>
    </tr>
```
And all the complex logic of creating the picklist items would be handled in the controller / service. That way we could have a print picklist action anywhere we need it, the GSP would be the same for each and the controller / service would transform the data to fit into the buckets we created for the GSP (header, groups, items, items grouped by groups). 

This probably would have required us creating our own Picklist and PicklistItem command classes that were mapped more cleanly to the GSP. For example, the handling of the zones would just be done as a group. 
```
PicklistCommand
    GroupByCommand sections 
    Map<GroupByCommand, List<PicklistItemCommand> getPicklistItemMap
```
Sections should actually be another command object

```
List<PicklistItemCommand> picklistItems = transformPicklistItems(picklist.picklistItems)
Map<GroupByCommand, List<PicklistItemCommand> picklistItemMap = picklistItems.groupBy { new GroupByCommand(it.whatever) }
PicklistCommand new PicklistCommand(sections: picklistItemMap.keySet(), picklistItemMap: picklistItemMap)
```
In this case we want to break it down by zone. In another case we might want to group it by categories.

However, I'm realizing that what I'm asking for might be too difficult so I'll let @awalkowiak decide if this meets the ticket's acceptance criteria and we can move it to QA.

jmiranda: The more I think about it, we could have gone even further and created generic reporting interfaces to deal with the elements of the report (i.e. Header, Groups, Rows) and then you could pass all of this data along with the column definitions as parameters to a generic "tabular report" GSP. Turning a very specific "print transfer order picklist" feature into a more generic "generate report" that is customizable based on the model / parameters passed to it.

jmiranda:
> what do you think @jmiranda ?

Ah, yes. I wrote the huge treatise for nothing. : You nailed it. 

> In both cases we pass [lineItems in the model](https://github.com/openboxes/openboxes/pull/3169/files#diff-94fbea89011b3f7d0c2794dee8056647ea4391623672fb8e76d34d0434868753R35) which are either orderItems or requesitionItems

The only point that was not explicit in your suggestion (and that I tried to make in mine) is that I think instead of trying to reuse RequisitionItem, PicklistItem, OrderItem we should consider adding a new command class (or even a List of Maps) that standardizes the elements of the report. Here we see two groups (zone, category) and then generic lines with row number, identifier, product name, lot number, exp date, from, to, total quantity, quantity to pick, comments.

image

Therefore I can transform anything (a sales order, requisition , replenishment order) into the list we need for this report. Even a hypothetical picklist based on product availability in the case that the requisition has not moved to the picking step yet (i.e. there's no picklist but we want to show what the allocation algorithm might generate).

The trickiest part here seems to be the total quantity vs quantity to pick since there's a rowspan required. But even that could be genericized in someway since we clearly know that 1) it could happen and 2) the value will either be 1:1 or 1:many (line item : picklist items).

cc @DaRoTP @awalkowiak

grails-app/views/replenishment/_printPage.gsp
```diff
@@ -0,0 +1,103 @@
...
+            <g:unless test="${lineItems}">
+                <tr>
+                    <td colspan="10" class="middle center">
+                        <span class="fade">
+                            <warehouse:message code="default.none.label"/>
+                        </span>
+                    </td>
+                </tr>
+            </g:unless>
+            <g:each in="${lineItems}" status="i" var="lineItem">
```
awalkowiak, line 34: Small improvement, but please change it to have a convention `each item in list` instead `each in list item` for better readability

grails-app/views/replenishment/_printPage.gsp
awalkowiak: @jmiranda I think we'll go with as generic as we can, and then we are gonna create a tech debt/improvement ticket to add commands and to refactor the existing print (for outbound sm) to use this new approach.

awalkowiak, line 36: Can we move this `?.findAll { it.quantity > 0 }}` to the logic in the controller?

grails-app/views/replenishment/_printPage.gsp
```diff
@@ -0,0 +1,103 @@
...
+
+                <g:set var="picklistItems" value="${pickListItemsByOrder[lineItem.id]?.findAll { it.quantity > 0 }}"/>
+                <g:set var="numInventoryItem" value="${picklistItems?.size() ?: 1}"/>
+
+                <g:set var="splitItems" value="${lineItem?.orderItems?.sort { a, b ->
```
awalkowiak, line 39: Could we make this part `lineItem?.orderItems?.sort` generic too? We might have requisition items for other picklist prints

grails-app/views/replenishment/print.gsp
```diff
@@ -0,0 +1,140 @@
...
+                <g:displayReportLogo/>
+            </div>
+        </td>
+        <td>
+            <div class="header">
+                <h1><warehouse:message code="order.transferOrder.label"/></h1>
+            </div>
+        </td>
+        <td class="top">
+            <table border="0">
```
awalkowiak, line 45: Please make this part as a loop on the `headerItems`

grails-app/views/replenishment/print.gsp
```diff
@@ -0,0 +1,140 @@
...
+</table>
+<g:set var="zoneNames" value='${itemsMap.keySet()}'/>
+<g:each var="zoneName" status="i" in="${zoneNames}">
+
+    <g:set var="lineItems" value='${itemsMap[zoneName].lineItems}'/>
+    <g:set var="pickListItems" value='${itemsMap[zoneName].pickListItems}'/>
+
+    <g:set var="showZoneName" value='${zoneName || zoneNames.size() > 1}'/>
+
+    <g:set var="pickListItemsByOrder" value='${pickListItems?.groupBy { it?.orderItem?.id } ?: [:]}'/>
```
awalkowiak, line 83: Can you move this grouping to the controller?

awalkowiak (changes requested):

awalkowiak (approved):

PR #3180, opened 2022-05-08
OBAM-355 Allow user to edit bin replenishment

kchelstowski:


grails-app/controllers/org/pih/warehouse/api/ReplenishmentApiController.groovy
```diff
@@ -107,7 +108,7 @@ class ReplenishmentApiController {
         }
 
         if (!replenishment.replenishmentNumber) {
-            replenishment.replenishmentNumber = identifierService.generateOrderIdentifier()
+            replenishment.replenishmentNumber = identifierService.generateOrderIdentifier() + Constants.REPLENISHMENT_CODE
```
awalkowiak, line 111: Call it REPLENISHMENT_PREFIX and add it in the beginning (use the property from config)

src/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -144,4 +144,7 @@ class Constants {
     static final String ROOT_CATEGORY_NAME = "ROOT"
 
     static final String DEFAULT_ORGANIZATION_CODE = "ORG"
+
+    // Order number suffix for bin replenishment case
+    static final String REPLENISHMENT_CODE = "_R"
```
awalkowiak, line 149: Should be REPLENISHMENT_PREFIX = "R-"
Add it also in the config as:
`openboxes.stockTransfer.binReplenishment.prefix = Constants.REPLENISHMENT_PREFIX`

grails-app/views/stockTransfer/_summary.gsp
```diff
@@ -45,6 +45,14 @@
                 <warehouse:message code="inventory.editStockTransfer.label" default="Edit Stock Transfer"/>
             </g:link>
         </g:elseif>
+        <g:elseif test="${orderInstance.orderNumber.contains(Constants.REPLENISHMENT_CODE)}">
```
awalkowiak, line 48: This should definitely not be `contains`, should be `startsWith`

awalkowiak (changes requested):

awalkowiak (approved):

PR #3188, opened 2022-05-11
OBPIH-4404 Shipment type seems not to be saved when save and exit fro…

kchelstowski:
…m stock movement

### DIFF
--- a/src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
+++ b/src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
@@ -422,10 +422,8 @@ class SendMovementPage extends Component {
             ...this.state.values,
             dateShipped: stockMovementData.dateShipped,
             shipmentType: {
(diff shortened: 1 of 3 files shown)


src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
```diff
@@ -420,7 +421,12 @@ class SendMovementPage extends Component {
           values: {
             ...this.state.values,
             dateShipped: stockMovementData.dateShipped,
-            shipmentType: _.get(stockMovementData, 'shipmentType.id'),
+            shipmentType: {
+              id: stockMovementData.shipmentType.id,
```
awalkowiak, line 425: I think you could here simply destructure `stockMovementData.shipmentType` and add only a label field since it is  made with `splitTranslation` (you might need to remove the property of `labelKey` from field configuration), and on top of that I think you don't have to add value field, since it's defined in the field config as `valueKey`.

(applies to all 3 files)

PR #3189, opened 2022-05-11
OBPIH-4403 Unable to receive inbound returns

kchelstowski:


grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -148,6 +148,13 @@ class ReceiptService {
         partialReceiptItem.lotNumber = shipmentItem.inventoryItem?.lotNumber
         partialReceiptItem.expirationDate = shipmentItem.inventoryItem?.expirationDate
         partialReceiptItem.quantityShipped = shipmentItem?.quantity ?: 0
+        if (!shipmentItem.inventoryItem) {
+            shipmentItem.inventoryItem =
+                    inventoryService.findOrCreateInventoryItem(
```
awalkowiak, line 153: This should be done in the StockTransferApiController and StockTransferService side. Debug why it is not working there.

awalkowiak (changes requested):

kchelstowski:
@awalkowiak after a long research it unexpectedly turned out that the changes I provided for OBPIH-4406 also resolve this task. The problem was that in StockTransferApiController the inventoryItem was created only if both lot and expirationDate weren't null. And if the inventoryItem was null, in OBPIH-4406 the product name and code weren't visible, because the data is taken from inventoryItem. 
If those changes seem to be OK for you, I think you can also merge OBPIH-4406 immedietaly as they provide the same change.

awalkowiak (approved):

PR #3190, opened 2022-05-11
OBPIH-4402 Add success message after submit request when logged to ward

DaRoTP:
https://pihemr.atlassian.net/browse/OBPIH-4402

### DIFF
--- a/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
@@ -10,6 +10,7 @@ import { confirmAlert } from 'react-confirm-alert';
 import { Form } from 'react-final-form';
 import { getTranslate } from 'react-localize-redux';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
 import Alert from 'react-s-alert';
 
 import { fetchUsers, hideSpinner, showSpinner } from 'actions';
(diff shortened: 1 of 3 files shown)


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1407,7 +1407,7 @@ class AddItemsPage extends Component {
           let redirectTo = `/openboxes/stockMovement/list?direction=INBOUND&movementNumber=${movementNumber}&submitted=true`;
 
           if (!this.props.supportedActivities.includes('MANAGE_INVENTORY') && this.props.supportedActivities.includes('SUBMIT_REQUEST')) {
-            redirectTo = '/openboxes/dashboard';
+            redirectTo = `/openboxes/dashboard?requestSubmitted=true&movementNumber=${movementNumber}`;
```
awalkowiak, line 1410: Can't you just display the `Alert.success` before redirecting?

DaRoTP: at the moment there was some kind of a problem with react-router and I couldn't figure it out so  I was redirecting using `window.location` (which refreshes the page and Alert would persist), but now I took another look at it and fixed the issue so the Alert message is displayed before redirect

awalkowiak (approved):

PR #3191, opened 2022-05-11
OBAM-271 Formating and monthly calculation fix

hdsldv:


grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
```diff
@@ -521,7 +519,7 @@ class ShowConsumptionCommand {
     }
 
     Float getNumberOfMonths() {
-        return numberOfDays / 30
+        return numberOfDays * 30
```
awalkowiak, line 522: The calculation does not correspond to the method name. Please add a separate method if there is a need for this type of calculation.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
hdsldv: fixed

awalkowiak (approved):

PR #3217, opened 2022-05-19
OBAM-367 Fix default bin product availability bug

DaRoTP:


DaRoTP:
After extensive investigation and testing I found that the problem was in getting the number of available products in `getQuantityAvailableToPromise`
the problem was that if **binLocation** was **null** it would sum up only quantityAvailableToPromise for all no null binLoations.
I wasn't sure if it was intentional or it was a bug but I couldn't find a case that it would use this logic.
This fixes the problem specified in the ticket.

awalkowiak (approved):

PR #3221, opened 2022-05-19
OBPIH-4374 Triage Java dependencies for Grails 3

mdpearson:
Hi folks,

This PR adds comments on a number of Java dependencies and removes a few that are never used. More details on the ticket. This should _definitely_ not go in before the release.


mdpearson:
I made a separate commit for each dependency I removed in case I was wrong -- should be easy to revert.

For the removal of our local copy of `core-renderer-R8-isomorphic.jar`, which was added in 2013, I put in _two_ commits. The first makes it so that the build actually refers to the local jar (current behavior ignores it entirely and fetches `core-renderer` from maven), while the second removes it altogether.

@jmiranda while these changes should not go in before the upcoming release, I'd like to do some more dependency cleanup to build off of this, and I'll be pushing another PR that removes dead code. Thoughts on where these should go? (A separate `feature/` branch?) Thanks!

mdpearson:
Superseded by OBPIH-4605.

PR #3222, opened 2022-05-20
OBPIH-4422 Invalid page redirect when exit from return's send page of…

kchelstowski:
… sent return


kchelstowski:
@awalkowiak You might be confused about the second (empty one) commit (with the title of another ticket) . It's just that I forgot that when I started working on OBPIH-4409 just for 30 minutes on the first day, I didn't realize next day, that I forgot to create a new branch for it, as for first day I was only "looking up" what is needed to be done.

awalkowiak (approved):

PR #3224, opened 2022-05-22
OBPIH-4409 After submitting request in ward it's details are visible …

kchelstowski:
…in breadcrumbs


kchelstowski:
@awalkowiak the label "don't merge" was there for a while, because I had to confirm some stuff with Katarzyna. It's open to review now.

PR #3252, opened 2022-05-30
Obpih 4457 unable to edit pending returns from stock movement lists

DaRoTP:


grails-app/views/stockMovement/_actions.gsp
```diff
@@ -26,14 +26,18 @@
                 </g:link>
             </div>
             <div class="action-menu-item">
-                <g:link
-                    controller="${stockMovement?.order ? "stockTransfer" : "stockMovement" }"
```
DaRoTP, line 30: It appears that conditional controller attribute in GSP doesn't work as expected and condition `stockMovement?.order` is always read as `false`.
I tried to investigate this issue but wasn't able to find the reason why it behaves like this.

so instead I used the `<g:if>` `<g:else>` approach

awalkowiak (approved):

PR #3258, opened 2022-05-31
OBGM-326 Remove barcode4j dependency.

mdpearson:
`barcode4j` has gone a decade without maintenance and we are using another equally-functional barcode generator (`zxing`) elsewhere in the codebase.


jmiranda:
I also wanted to move the product-specific barcode action to a generic barcode controller, but we can leave that for now since we don't want to do too much moving of code.

mdpearson:
@awalkowiak rebased this one; it's now ready to merge into Grails 3.

PR #3264, opened 2022-06-02
OBPIH-4472 Remove deprecated packages, resolve security warnings and fix eslint.

mdpearson:
This PR also fixes the build so that if webpack reports an error, the build reliably stops.

@awalkowiak I figure changes like this are best submitted at the beginning of a release cycle, and none of these files should change much between Grails 1 and Grails 3. Let me know if you have any questions about these changes, and thanks!


awalkowiak:
@mdpearson Woah, I am gonna need to review this one locally, there seem to be a lot of changes

PR #3270, opened 2022-06-03
OBPIH-4400.3 Use warehouse:message, not g:message, to localize helpscout button

mdpearson:


awalkowiak:
@mdpearson this should target the `release/0.8.18` branch, and when I changed the target it sees like there is other commit that wants to get it. Could you please recreate the PR with the proper base branch?

mdpearson:
@awalkowiak should be good to go now.

awalkowiak (approved):

PR #3273, opened 2022-06-06
OBPIH-4439 Mark order adjustment as "invoiced" only if there is purch…

awalkowiak:
…ase invoice


jmiranda (commented):
LGTM

Would have only made a nitpick about the purchaseInvoiceCode and that it should just be `invoiceTypeCode: InvoiceTypeCode.INVOICE` as the purchase . In addition, I think getPostedInvoiceItems() would have been better since there's nothing that is actually make this distinguishing these invoice items as "purchase invoice items" in the query.
cc @awalkowiak

PR #3285, opened 2022-06-09
OBPIH-4534 Lines with qty=0 are not removed after using save button i…

kchelstowski:
…n request workflow


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1287,16 +1285,31 @@ class AddItemsPage extends Component {
...
-          if (!this.props.supportedActivities.includes('MANAGE_INVENTORY') && this.props.supportedActivities.includes('SUBMIT_REQUEST')) {
-            redirectTo = '/openboxes/dashboard';
-          }
-          window.location = redirectTo;
-        },
-      });
+      const lineItems = _.filter(formValues.lineItems, item => !_.isEmpty(item));
+      const zeroedLines = _.some(lineItems, item => !item.quantityRequested || item.quantityRequested === '0');
+      if (zeroedLines || this.state.isRequestFromWard) {
+        this.confirmSave(() => this.saveRequisitionItemsInCurrentStepWithAlert({
```
awalkowiak, line 1291: I think you could use the `saveRequisitionItemsInCurrentStep` or at least rename `saveRequisitionItemsInCurrentStepWithAlert` if you want to use a callback and you don't want to refactor the `saveRequisitionItemsInCurrentStep`. The `saveRequisitionItemsInCurrentStepWithAlert` naming is no longer valid because there is no `alert` anymore inside.

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
awalkowiak: @kchelstowski actually you don't need it at all just make it as:
```
this.props.showSpinner();
    return this.saveRequisitionItemsInCurrentStep(lineItems)
      .then(res =>  {
        let redirectTo = '/openboxes/stockMovement/list?direction=INBOUND';
        if (!this.props.supportedActivities.includes('MANAGE_INVENTORY') && this.props.supportedActivities.includes('SUBMIT_REQUEST')) {
              redirectTo = '/openboxes/dashboard';
        }
        window.location = redirectTo;
      })
      .catch(() => {
        this.props.hideSpinner();
      });
```

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1284,19 +1267,41 @@ class AddItemsPage extends Component {
...
-            redirectTo = '/openboxes/dashboard';
-          }
-          window.location = redirectTo;
-        },
-      });
+      const lineItems = _.filter(formValues.lineItems, item => !_.isEmpty(item));
+      const zeroedLines = _.some(lineItems, item => !item.quantityRequested || item.quantityRequested === '0');
+      if (zeroedLines || this.state.isRequestFromWard) {
+        this.confirmSave(() => {
+          this.props.showSpinner();
```
awalkowiak, line 1278: Move this block to separate function, since you are duplicating it below

awalkowiak (approved):

PR #3288, opened 2022-06-10
OBPIH 4533 Fix parsing monthlyDemand

DaRoTP:


DaRoTP:
problem was that `monthlyDemand` was and integer and `.replace()` expects a string

PR #3291, opened 2022-06-13
OBPIH 4533 Fix request products not updating after save

DaRoTP:


DaRoTP:
The problem that was mentioned in  [this comment](https://pihemr.atlassian.net/browse/OBPIH-4533?focusedCommentId=133116)  seems to be connected to the fact that after the first save if one of the saved products had `quantityRequested <= 0 ` then it would remove this item from requisition items list and on next save _(since list of items has not updated on the fronted)_ it sends that removed item with the rest and causes an error.

One of the solutions was to update the list of products in the table on the fronted after save, and it looked like it was already implemented [here](https://github.com/openboxes/openboxes/pull/3291/files#diff-d8323bb26d8eaa563ed93c8a7a5dd18d79764f1d0e8528048dc50cf017fedc08L1216) , but there was a technical problem that cause the `setState` function not to be called.

After investigating - it appeared to be that calling  `_.find()` method was the culprit of this issue.
```
_.find(
    this.state.values.lineItems,
    lineItem => lineItem.sortOrder === val.sortOrder,
 )
```

 I am still not sure why it caused this problem, it might be connected to calling  `_.find()` method inside a  '_.map()' that causes some asynchronous side effects but another solution was - instead of setting updated products list from the response data after save we can call `fetchLineItems()` that refetches and sets updated items for us

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1211,46 +1210,14 @@ class AddItemsPage extends Component {
...
-
-          this.setState({
-            values:
-              { ...this.state.values, lineItems: lineItemsBackendData },
-            totalCount: lineItems.length,
-            currentLineItems: lineItemsBackendData,
-          });
-        })
         .then(() => this.fetchAddItemsPageData())
+        .then(() => this.setState({
```
awalkowiak, line 1214: These two next `thens` feels weird. `TableBodyVirtualized` should deal with item refetch through the `isFirstPageLoaded`. I think you should set the `isFirstPageLoaded: false` inside the FetchAddItemsPageData callback and then it will be pulled automatically. The `fetchLineItems` should be called only if the `this.props.isPaginated` is false.

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
DaRoTP, line 1214: setting empty lists for  `currentLineItems: []` and  `values: { ...this.state.values, lineItems: [] }` is necessary, because `fetchLineItems()` method concatenates previouse lineItems with the new fetched ones

awalkowiak: Move `this.fetchAddItemsPageData()` from a separate `then` to the setState's callback

awalkowiak (changes requested):

awalkowiak (approved):

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1029,6 +1029,9 @@ class AddItemsPage extends Component {
     this.fetchAddItemsPageData();
     if (!this.props.isPaginated) {
       this.fetchLineItems();
+    } else if (this.state.isFirstPageLoaded) {
+      // refetching update items when first page s already loaded
```
awalkowiak, line 1033: // Workaround for refetching items from scrach when the first page was already loaded and table is paginated

PR #3293, opened 2022-06-14
OBPIH-3083 Add handling icons to PO view and edit pages

kchelstowski:


grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -950,6 +950,7 @@
 	<td class="left middle" style="color: {{= product.color }}">
         {{= product.productCode }}
         {{= product.name }}
+        <g:renderHandlingIcons product="{{= product}}"/>
```
awalkowiak, line 953: @jmiranda do you know what magic trick needs to be done here to get this tag working properly in the `jquery-template`? The issue here is that in this version the product is passed as a `{{= product}}"` string, and if there would be: `product="{{= product.name}}"`, there would be info that `cannot get name on the null object`. However, I've seen that some other tags (eg. selects or text boxes) were used in the `jquery-template` without any issues with it.

jmiranda: @awalkowiak They probably worked because you were passing in the ID as an attribute. Here, I'd make it so the renderHandlingIcons could take a productId or product object and have the taglib handle both cases. 

But we're rendering HTML using javascript so I think it would be better to find a javascript-y way to render the icons.

Not sure if the product would return the list of icons, but if so I would iterate through the list and render each separately.

awalkowiak: I think we tried with Kacper the `productId` version already, but I'll double check that.

jmiranda: View Source and check to see what's rendered in the HTML at this point. 

In any case, I don't think we should be using these taglibs in javascript.

awalkowiak: @jmiranda here's the result (rendered with the `<g:textField ... />` to see the difference):
```
<td class="left middle" style="color: {{= product.color }}">
  {{= product.productCode }}
  {{= product.name }}

  <input type="text" name="productId" value="{{= product.id }}" id="productId" />
</td>
```
As you can see the text field label is rendered into input, but the handling icons are rendered as empty string because that's what is returned if the product is not found by `product.id`. I am not sure if that tells us anything. I believe the PO pages will be reworked in the near future I feel like there is no point in digging it further. I am gonna drop this part.

awalkowiak (approved):

PR #3294, opened 2022-06-14
OBPIH-4531 Fix Substitution reason code label

DaRoTP:


src/js/components/stock-movement-wizard/outbound/EditPage.jsx
```diff
@@ -367,10 +367,12 @@ class EditItemsPage extends Component {
           ...val.product,
           label: `${val.productCode} ${val.productName}`,
         },
-        reasonCode: this.props.reasonCodes.find(({ value }) => value === val.reasonCode),
+        // eslint-disable-next-line max-len
+        reasonCode: _.find(this.props.reasonCodes, ({ value }) => _.includes(val.reasonCode, value)),
```
DaRoTP, line 371: since there is a possibility for reason code to be `SUBSTITUTION(reason code)` we want to check if reason code includes a reason code string and then apply it on the table

awalkowiak (approved):

PR #3296, opened 2022-06-15
OBPIH-4549 PO shipment available items download has UOM formatting issue

kchelstowski:


grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -171,7 +171,7 @@ class CombinedShipmentItemApiController {
             def orderItems = OrderItem.findAllByOrderInList(orders)
             orderItems.findAll{ it.orderItemStatusCode != OrderItemStatusCode.CANCELED && it.getQuantityRemainingToShip() > 0 }.each {orderItem ->
                 String quantityUom = "${orderItem?.quantityUom?.code?:g.message(code:'default.ea.label')?.toUpperCase()}"
-                String quantityPerUom = "${g.formatNumber(number: orderItem?.quantityPerUom?:1, maxFractionDigits: 0)}"
+                String quantityPerUom = orderItem?.quantityPerUom?.toInteger()
```
awalkowiak, line 174: @kchelstowski Make sure that quantityPerUom cannot be decimal because the property itself has the BigDecimal type, but I have not found the case where it was decimal.

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
awalkowiak: @kchelstowski now I noticed that you have `String quantityPerUom = integer`, this should be fixed

PR #3298, opened 2022-06-15
OBPIH-4428 Link HelpScout Beacon to localized articles.

mdpearson:


awalkowiak (commented):
@mdpearson I am gonna merge it as is, but I am wondering if this is good approach with the `helpscout.beacon.key.label`s. I am wondering if this should not be defined in the config.groovy. Plus do we care about these keys being exposed?

PR #3302, opened 2022-06-19
OBPIH-4451 Issue with user permissions in location chooser

kchelstowski:
There are a lot of combinations of user/location permissions both user default permissions and location-based. I tested a bunch of them and it seemed to be working fine - if a user doesn't have any default permissions and doesn't have any location-based permissions, none locations are visible and if a user doesn't have any default permissions which were pointed out in the ticket, but has that permission in any location - that location is visible.
Worth pointing out is the fact that I had to be switching between develop branch and release branch to be able to edit a user, because while trying to edit the user on release branch I was getting an ugly error: 
```
"Exception Message: No enum constant org.pih.warehouse.core.RoleType.ROLE_PRODUCT_MANAGER "
```
It is probably due to the fact, that it was implemented on develop and wasn't implemented on release, but I have it in my database and it gives the error.

You might ask why I removed `else if (params.locationChooser && requestorInAnyLocation)` and have put it inside else - it's because if I had a requestor role in a location and had additionaly some other permissions to any other locations, it wouldn't show those locations, because the if would stop at that else if and wouldn't go further where the other locations would be loaded, so my thought was to put in inside else to make sure, that that case is covered.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
@@ -15,6 +15,7 @@ import org.codehaus.groovy.grails.web.json.JSONObject
 import org.hibernate.Criteria
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.LocationRole
 import org.pih.warehouse.core.LocationType
 import org.pih.warehouse.core.RoleType
 import org.pih.warehouse.core.User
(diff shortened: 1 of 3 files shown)


awalkowiak (changes requested):
Could you extract the common logic and move it to one method in the location service that both places could use?

kchelstowski:
As discussed there would be not option at the moment to split the logic into one method for both places as there are minor differences in both, e.g. the `getLoginLocationsMap` uses method `getLoginLocations` and the `LocationApiController` in analogous place uses `getLocations` with tons of params.
@awalkowiak statement was that it would need to be refactored in the future

awalkowiak (approved):

PR #3305, opened 2022-06-20
OBPIH-3810 Fix undescriptive error message on adjustment and item deletion in PO

DaRoTP:


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -402,9 +402,16 @@ class OrderController {
...
                     }
-                    orderInstance.removeFromOrderAdjustments(orderAdjustment)
-                    orderAdjustment.delete()
-                    if (!orderInstance.hasErrors() && orderInstance.save(flush: true)) {
+
+                    try {
+                        orderInstance.removeFromOrderAdjustments(orderAdjustment)
+                        orderAdjustment.delete()
+                        orderInstance.save(flush: true)
+                    } catch(Exception e) {
```
awalkowiak, line 410: I am not sure if catching 'vague' exception and throw `UnsupportedOperationException` is a good approach. What is the exception that you are getting here?

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
DaRoTP: yes, you are probably right
the exception that I am getting is 
```
Cannot delete or update a parent row: a foreign key constraint fails ('openboxes'.'order_adjustment_invoice', CONSTRAINT 'order_adjustment_invoice_ibfk_1' FOREIGN KEY ('order_adjustment_id') REFERENCES 'order_adjustment' ('id')
```

and so I am not sure how to deal with that then.

awalkowiak: On second thought your try-catch approach is definitely a wrong approach. You should check if the item or adjustment has an invoice associated and then throw `UnsupportedOperationException`.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -386,35 +386,25 @@ class OrderController {
...
-            }
         }
+
+        OrderAdjustment orderAdjustment = OrderAdjustment.get(params?.id)
+        if (!orderAdjustment) {
+            flash.message = "${warehouse.message(code: 'default.not.found.message', args: [warehouse.message(code: 'orderAdjustment.label', default: 'Order Adjustment'), params.id])}"
+            redirect(action: "show", id: params.order.id)
+        }
+
+        orderService.deleteAdjustment(order, orderAdjustment, user)
```
awalkowiak, line 403: I believe you don't have to pass order here, because you can make `orderAdjustment.order`

awalkowiak (changes requested):

awalkowiak (approved):

PR #3306, opened 2022-06-20
OBPIH-4440 Add item search to PO view page

kchelstowski:
After many different approaches, the dynamic filtering (hiding a row) worked the best here. I managed to recreate what was done probably in Stock History section. At the moment the filtering is done by product name which Kelsey admitted on one of the previous standups. The changes may seem weird (it might look like I changed the whole file) on Git, due to indents because of one additional <div> for everything, but the part I edited was just adding the <script> section and one if at the beginning of the first div.
The "if" was added to determine whether it is a putaway order or a PO order, because I assume it was not supposed to be done for putaways and as there is one view for both cases, it would work also for putaways if I hadn't added the "if".

### DIFF
--- a/grails-app/views/order/_itemDetails.gsp
+++ b/grails-app/views/order/_itemDetails.gsp
@@ -1,80 +1,122 @@
 <%@ page import="org.pih.warehouse.order.OrderItemStatusCode" %>
 <%@ page import="org.pih.warehouse.order.OrderType" %>
 <%@ page import="org.pih.warehouse.order.OrderTypeCode" %>
-<div id="tab-content" class="box">
-    <h2>
-        <warehouse:message code="order.itemDetails.label" default="Item Details"/>
-    </h2>
-    <g:if test="${orderInstance?.orderItems }">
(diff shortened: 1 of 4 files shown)


grails-app/views/order/_itemDetails.gsp
```diff
@@ -1,80 +1,122 @@
...
+        .val()
+        .toUpperCase();
+      filterTableItemDetails(filterCell, filterValue)
+    });
+
+  });
+  function filterTableItemDetails(cellIndex, filterValue) {
+    const tableRows = $("#order-items-details tr.dataRowItemDetails");
+    // Loop through all table rows, and hide those who don't match the search query
+    $.each(tableRows, function(index, currentRow) {
```
awalkowiak, line 21: Pass the `tableRows` to this function, and extract the rest, because it is common for all 3 filter functions.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3308, opened 2022-06-22
OBPIH-4451 Fixes after QA

kchelstowski:
The problem was that the methods `inRoleBrowser` etc were checking if a user has a permission in any or current location instead of checking if the user has those permissions globally. It is my fault that I haven't realized that, because:
1. it was working well while logging in (because you weren't logged in anywhere and the methods were giving false, and the locations were hidden then, which was good)
2. the "React" method (`getLocations` instead of `getLoginLocations` like it is anywhere else (GSP and logging in)) seemed to be working fine after logging in 
3. The problem turned out on GSP pages - the GSP pages were using the same stuff as while logging in, but inside any GSP page we were logged in to a location, so the methods `inRoleBrowser`, `isUserAdmin` etc were giving some weird stuff.

The fix was to create those methods to check if the user has those permissions globally instead of checking the permission in particular location. It seems to be working fine now in any case (logging in, react and gsp). Even though it should be refactored sometime as discussed, because it might provide to some mistakes in the future.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -149,6 +149,46 @@ class UserService {
         return false
     }
 
+    Boolean hasRoleAssistantGlobally(User u) {
+        if (u) {
+            def assistantRole = RoleType.ROLE_ASSISTANT
+            return u.roles.any { role -> role.roleType == assistantRole}
+        }
+        return false
+    }
+
+    Boolean hasRoleBrowserGlobally(User u) {
+        if (u) {
+            def browserRole = RoleType.ROLE_BROWSER
+            return u.roles.any { role -> role.roleType == browserRole}
+        }
+        return false
+    }
+
+    Boolean isManagerGlobally(User u) {
+        if (u) {
+            def managerRole = RoleType.ROLE_MANAGER
+            return u.roles.any { role -> role.roleType == managerRole}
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -149,6 +149,46 @@ class UserService {
         return false
     }
 
+    Boolean hasRoleAssistantGlobally(User u) {
```
awalkowiak, line 152: Make all these methods generic. Add it to the User domain as a method:
```
Boolean hasDefaultRole(RoleType roleType) {
    return roles?.any { Role role -> role.roleType == roleType} ?: false
}
```
and then in the `LocationService` and in the `LocationApiController` you can do:
```
def inRoleBrowser = user.hasDefaultRole(RoleType.ROLE_BROWSER)
def inRoleAssistant = user.hasDefaultRole(RoleType.ROLE_ASSISTANT)
def inRoleManager = user.hasDefaultRole(RoleType.ROLE_MANAGER)
def inRoleAdmin = user.hasDefaultRole(RoleType.ROLE_ADMIN)
def inRoleSuperuser = user.hasDefaultRole(RoleType.ROLE_SUPERUSER)
```

awalkowiak (changes requested):

awalkowiak (approved):

PR #3310, opened 2022-06-22
OBKN-469 Added a database migration to create product demand tables a…

jmiranda:
…fter view creation (to avoid issues with views that reference product demand tables)


jmiranda:
Don't actually need this because we merged it into master as a hotfix to 0.8.17 before merging the 0.8.18 release to master.

PR #3314, opened 2022-06-23
Merge master back to develop.

mdpearson:
This brings in all the stuff from master to develop, along with a bump in app.version to `0.8.19-SNAPSHOT`.


awalkowiak:
@mdpearson looks like this line is causing the failed test https://github.com/openboxes/openboxes/blob/master/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy#L156 
changing it to: `String localizedHelpScoutKey = helpScoutService?.localizedHelpScoutKey` will fix the issue

PR #3320, opened 2022-06-24
OBS-1176 Don't let LazyInitializationExceptions kill error pages.

mdpearson:
This PR adds a guard to `hasHighestRoleAuthenticated` and its close cousin `hasHigherRoleThanAuthenticated` so that, if they are called after the session is closed, they don't crash and don't leak sensitive information.

As I mentioned on Slack I'm not particularly happy with this approach, because I don't know what happened in the latest release to trigger it. Still, it can stanch the bleeding; and perhaps this PR or ticket is a better place to discuss than Slack.

Another option would be to eager-load the `roles` collection instead of doing so lazily: it certainly sees a lot of use, so perhaps there'd be a performance improvement along with crash prevention.

A third option would be to refactor the error page so as to not require these methods, and of course a fourth is to figure out the root cause of the issue. That latter one will be tough for me till Monday as I am on the road with limited internet access.

### DIFF
--- a/grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
+++ b/grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
@@ -9,12 +9,18 @@
  **/
 package org.pih.warehouse
 
+import org.hibernate.LazyInitializationException
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Location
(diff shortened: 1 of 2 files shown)


awalkowiak:
@mdpearson I thing this one is redundant, after this fix https://github.com/openboxes/openboxes/commit/6f856dc51a174adefcd672df1c277b28de9cdff2

PR #3321, opened 2022-06-24
OBPIH-4562 Items balance dialog in transaction report doesn't work

kchelstowski:


awalkowiak (changes requested):
@kchelstowski I think this might be another issue caused by the codec change. I am not sure how we could deal with it on the 1.3.9, but you could try with `encodeAsUrl()` or `raw()` on url in the controller

kchelstowski:
@awalkowiak I tried the solutions you asked me to check, but none of them works - the `encodeAsURL()` both tried in controller and at .gsp gave url like this:
```
%2Fopenboxes%2Fjson%2FgetTransactionReportDetails%3FstartDate%3D06%252F01%252F2020%26productCode%3D10001%26locationId%3D8a8a9e9665c4f59d0165c5505245002b%26endDate%3D06%252F27%252F2022
``` 
which was giving an error on the frontend side.
The `raw()` didn't work as well - gave the error - "No signature of method".

`toString()` - didn't change anything

I let you decide whether it is worth to figure it out further or we should accept my solution with replace

grails-app/views/report/_dataTableDialog.gsp
```diff
@@ -42,10 +42,11 @@
             "sPaginationType": "two_button",
             "sAjaxSource": "${url}",
             "fnServerData": function ( sSource, aoData, fnCallback ) {
+              const sSourceParsed = sSource.replaceAll("&amp;", "&");
```
awalkowiak, line 45: I am gonna merge this version as a short-term remedy, but I'll create a tech debt ticket to further investigate this issue.

PR #3322, opened 2022-06-25
OBKN-679 Disable validation for date requested field on outbound stoc…

jmiranda:
…k movement


jmiranda:
@awalkowiak Could you just check to see if I'm missing anything with this change? 

I know this seems kind of dumb but I am hoping that if this change somehow gets merged back to develop at some point in the future, the commented code will remind me that we don't actually want to do this. If anything we'd want to add a supported activity to enable / disable the validation, but I really don't want to spend time on that now.

PR #3330, opened 2022-06-30
OBPIH-4579 Add handling icons to receiving page

DaRoTP:


src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -157,7 +158,22 @@ const TABLE_FIELDS = {
         },
       },
       'product.name': {
-        type: params => (params.subfield ? <LabelField {...params} /> : null),
+        type: (params) => {
+          if (params.subfield) {
+            const fieldName = params.fieldName.replace('.name', '.handlingIcons');
```
DaRoTP, line 163: this might not be the best solution to getting handling icons property from a nested list but I couldn't find a better one

If you have any suggestions please let me know and I will modify it

awalkowiak (changes requested):

src/js/components/receiving/PartialReceivingPage.jsx
DaRoTP: since this view has a table with subfields - handling icons value i deeply nested in `values.containers[$containerIndex].shipmentItems[$itemIndex].product.handlingIcons`

DaRoTP: It seems that I have missed `parentIndex` parameter before that's why I came up with that sneaky `.fieldName.replace('.name', '.handlingIcons')` solution.

So now I fixed it into a more reliable way of getting this data

```
const { parentIndex, rowIndex } = params;
const fieldPath = `containers[${parentIndex}].shipmentItems[${rowIndex}].product.handlingIcons`;
const handlingIcons = _.get(params.values, fieldPath);
```

and I think there is no other logical way to get this nested data

awalkowiak (approved):

PR #3334, opened 2022-07-01
OBPIH-4590 Location group filtering not working in outbound

kchelstowski:


awalkowiak (approved):
I think there would be an option to improve the code a bit here, but since it is a hotfix then I am gonna merge it as is to avoid any potential regressions and I'll create a tech debt ticket for refactoring.

PR #3340, opened 2022-07-05
OBPIH-4580 Add supplier information to inbound packing list download …

kchelstowski:
…(excel)

Those empty strings, eg. 
```
                String supplierOutput = ""
```
are made and added to cell even if they are empty, because without at least empty string being added to the cell, the cell has no border, which would not be acceptable.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
@@ -52,6 +52,7 @@ import org.pih.warehouse.api.Stocklist
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.OrderType
 import org.pih.warehouse.order.OrderTypeCode
+import org.pih.warehouse.product.ProductSupplier
 import org.pih.warehouse.requisition.RequisitionItem
 import org.pih.warehouse.requisition.RequisitionItemSortByCode
 import org.pih.warehouse.shipping.ReferenceNumber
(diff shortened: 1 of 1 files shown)


awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -917,23 +917,17 @@ class DocumentService {
...
-                    manufacturerOutput += supplier.manufacturer
-                    manufacturerCodeOutput += supplier.manufacturerCode
-                }
-                row.createCell(CELL_INDEX).setCellValue(supplierOutput)
+
+                row.createCell(CELL_INDEX).setCellValue(supplier?.supplierCode ?: '')
                 row.getCell(CELL_INDEX++).setCellStyle(tableDataCenterStyle)
 
-                row.createCell(CELL_INDEX).setCellValue(manufacturerOutput)
+                row.createCell(CELL_INDEX).setCellValue(supplier?.manufacturer?.toString() ?: '')
```
kchelstowski, line 927: `toString()` was neccessary, because supplier.manufacturer is of `Organization` class and in this version it was said that 
```
setCellValue cannot be applied to '(java.io.Serializable & java.lang.Comparable<? extends java.io.Serializable & java.lang.Comparable<?>>)' 
```

awalkowiak (approved):

PR #3341, opened 2022-07-05
OBPIH-4589 Fix multiple availability refresh for the same product

awalkowiak:


jmiranda (approved):
Wow. I'm speechless. Good catch.

PR #3344, opened 2022-07-06
OBPIH-4446 Make already existing budget codes active by default

kchelstowski:


grails-app/migrations/0.8.x/changelog-2022-06-29-1200-alter-table-budget-code-add-column-active.xml
```diff
@@ -10,7 +10,7 @@
             </not>
         </preConditions>
         <addColumn tableName="budget_code">
-            <column name="active" type="BIT" />
+            <column name="active" type="BIT" defaultValueBoolean="true" />
```
awalkowiak, line 13: You have to add that in a separate migration. This won't work if this is already applied migration.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3346, opened 2022-07-07
OBPIH-4598 Add validation for inactive budget codes when adding items…

kchelstowski:
… to PO through import

### DIFF
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -731,9 +731,13 @@ class OrderService {
                         throw new IllegalArgumentException("Budget code is required.")
                     }
                     BudgetCode budgetCode = BudgetCode.findByCode(code)
-                    if (code && !budgetCode) {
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -731,9 +731,13 @@ class OrderService {
                         throw new IllegalArgumentException("Budget code is required.")
                     }
                     BudgetCode budgetCode = BudgetCode.findByCode(code)
-                    if (code && !budgetCode) {
-                        throw new IllegalArgumentException("Could not find budget code with code: ${code}.")
-
+                    if (code) {
```
awalkowiak, line 734: Reduce this to two ifs?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
awalkowiak: Actually, the cognitive complexity won't change, you can disregard it.

PR #3347, opened 2022-07-08
OBPIH-4597 Add validation for inactive product sources when adding it…

kchelstowski:
…ems to PO through import

### DIFF
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -653,10 +653,13 @@ class OrderService {
 
                     if (sourceCode) {
                         ProductSupplier productSource = ProductSupplier.findByCode(sourceCode)
-                        if (productSource && productSource.product != orderItem.product) {
-                            throw new ProductException("Wrong product source for given product")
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -653,10 +653,13 @@ class OrderService {
 
                     if (sourceCode) {
                         ProductSupplier productSource = ProductSupplier.findByCode(sourceCode)
-                        if (productSource && productSource.product != orderItem.product) {
-                            throw new ProductException("Wrong product source for given product")
-                        }
                         if (productSource) {
```
awalkowiak, line 656: Reduce this to two ifs

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
awalkowiak: Actually, the cognitive complexity won't change, you can disregard it.

awalkowiak (approved):

PR #3350, opened 2022-07-11
OBPIH-4444 allow deactivation of person records

DaRoTP:


grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -118,6 +118,10 @@ class UserController {
         log.info "attempt to save the user; show form with validation errors on failure"
         def userInstance = new User(params)
 
+        if (!params.hasProperty("active")) {
```
awalkowiak, line 121: Is this part required if there is a default value on it?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -269,6 +269,9 @@ class UserService {
                         ilike("email", "%" + term + "%")
                     }
                 }
+                if (params.status) {
```
awalkowiak, line 272: Why it's called `status` instead of `active`?

DaRoTP: this same logic of calling this parameter _status_ was already used in a method `def findUsers(...)`   [on this line ](https://github.com/openboxes/openboxes/blob/f98db19ecfa175b8830f724d635d6154e4d3c9d1/grails-app/services/org/pih/warehouse/core/UserService.groovy#L293) so I just followed the convention that was already used in this situtation.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
DaRoTP: the default value for _active_ on **Person**  is set to `True` which is inherited by **User** but when creating a new user default value should be `False`

awalkowiak: Could you add this as a comment above?

awalkowiak (approved):

PR #3358, opened 2022-07-14
OBPIH-4600 Reason code field disabled when undo substitution after re…

kchelstowski:
…turn to sm after save and exit

After long investigation I think the problem there was that `const isSubstituted` was considering `lineItems` instead of `editPageItems` which are taken to the table. After substituting, the item has `statusCode` === 'SUBSTITUTED', but when making the reproduction steps from the ticket, and undoing the item, this item in `lineItems`' array had still `statusCode` 'SUBSTITUTED' (it should have 'APPROVED') and it seemed like only editPageItems refetch after the undoing. 
Assuming that in above fields we don't use `lineItems` anywhere and we do use `editPageItems`, I think it should be the proper fix.

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/EditPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/EditPage.jsx
@@ -246,8 +246,7 @@ const FIELDS = {
         getDynamicAttr: ({
           fieldValue, subfield, reasonCodes, updateRow, values, rowIndex,
         }) => {
-          const isSubstituted = values.lineItems && values.lineItems[rowIndex] &&
-            values.lineItems[rowIndex].statusCode === 'SUBSTITUTED';
+          const isSubstituted = fieldValue && fieldValue.statusCode === 'SUBSTITUTED';
           return {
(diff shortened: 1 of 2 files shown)


src/js/components/stock-movement-wizard/outbound/EditPage.jsx
```diff
@@ -246,8 +246,7 @@ const FIELDS = {
         getDynamicAttr: ({
           fieldValue, subfield, reasonCodes, updateRow, values, rowIndex,
         }) => {
-          const isSubstituted = values.lineItems && values.lineItems[rowIndex] &&
-            values.lineItems[rowIndex].statusCode === 'SUBSTITUTED';
+          const isSubstituted = fieldValue && fieldValue.statusCode === 'SUBSTITUTED';
```
awalkowiak, line 249: @kchelstowski  Please take a look at `request/EditPage` if it is proper there too

kchelstowski: @awalkowiak nice catch, I've forgotten about requests - obviously there was the same issue - below I commited a fix for requests also

awalkowiak (approved):

PR #3367, opened 2022-07-18
OBPIH-4585 integration test fixes & OBPIH-4608 run integration tests in GHA

mdpearson:
Hi folks, this PR repairs the existing unit tests (OBPIH-4585) and runs them in GHA (OBPIH-4608). Integration tests are triggered identically to the existing unit test behavior and run in 5-15 minutes. By volume, most of this PR consists of tweaks to existing test setup and data; I'll flag the most interesting parts with comments. I also reduced copy-pasta test initialization by hoisting code to `DbHelper`. Thanks for your review!


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,137 @@
...
+        echo "dataSource.password=${{ env.OPENBOXES_PASSWORD }}" > ~/.grails/openboxes-config.properties
+        echo "dataSource.url=jdbc:mysql://localhost:3306/${{ env.DB_DATABASE }}?allowPublicKeyRetrieval=true&useSSL=false" >> ~/.grails/openboxes-config.properties
+        echo "dataSource.username=${{ env.OPENBOXES_USER }}" >> ~/.grails/openboxes-config.properties
+        echo "openboxes.fixtures.enabled=true" >> ~/.grails/openboxes-config.properties
+        echo "openboxes.jobs.dataCleaningJob.enabled=false" >> ~/.grails/openboxes-config.properties
+        echo "openboxes.jobs.refreshDemandDataJob.enabled=false" >> ~/.grails/openboxes-config.properties
+        echo "openboxes.jobs.refreshInventorySnapshotJob.enabled=false" >> ~/.grails/openboxes-config.properties
+        #
+        # FIXME Strangely, enabled = false doesn't disable refreshProductAvailabilityJob,
+        # FIXME but setting delayInMilliseconds to a high value has the same practical effect.
```
mdpearson, line 111: ^^^ I'm not sure why setting `enabled=false` doesn't work here; my uninformed guess is something something Docker character-set something something... the other changes work properly and the workaround is perfectly fine. Still, if either if you has insight into this, I'm curious and all ears!

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,137 @@
...
+      # OpenBoxes release. Technically, this step is optional in the sense that
+      # later steps would automatically run all the necessary migrations if it
+      # were removed. But those migrations would certainly take hours, and almost
+      # certainly fail with out-of-memory errors.
+      #
+    - name: preload empty database (v0.8.18-hotfix3)
+      run: |
+        gunzip -c ${{ github.workspace }}/test/data/preload.tgz | tar xvf -
+        mysql -p${{ env.DB_PASSWORD }} -u${{ env.DB_USER }} "${{ env.DB_DATABASE}}" < preload_schema.sql
+        mysql -p${{ env.DB_PASSWORD }} -u${{ env.DB_USER }} "${{ env.DB_DATABASE}}" < preload_data.sql
```
mdpearson, line 129: ^^^ @jmiranda by the time I read your comment on OBPIH-4608 I had already written this. Let me know if it's what you had in mind. (I generated these two files using the code in `OBPIH-4305-clone-remote-database`)

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      # if this step fails midway through, the next will pick up where it left off
```
mdpearson, line 131: This guards against `No such property: mergedConfig for class: org.codehaus.groovy.grails.commons.DefaultGrailsApplication`, which I ran into when developing this action, but which I can't reproduce anymore. Maddeningly intermittent Docker build failures are, well, maddening... I think `continue-on-error: true` can't possibly hurt in this context and with any luck will prevent build gotchas down the line

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -616,17 +616,29 @@ class Product implements Comparable, Serializable {
      *
      * @return
      */
-    String toString() { return "${name}" }
+    @Override
+    String toString() { return name }
```
mdpearson, line 620: no need for GString interpolation here

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -616,17 +616,29 @@ class Product implements Comparable, Serializable {
...
+        return name <=> obj?.name ?: id <=> obj?.id
+    }
 
-        def sortOrder =
-                name <=> obj?.name ?:
-                        id <=> obj?.id
-        return sortOrder
+    @Override
+    int hashCode() {
+        return id?.hashCode() ?: super.hashCode()
```
mdpearson, line 632: Some APIs use a product object as a key, and integration tests were failing to find the same object when looking up results.

In Grails 3 I'd love to make hashCode and equals a proper Trait and assign it to all domains to prevent bugs of this type, let me know what you think @jmiranda

grails-app/services/org/pih/warehouse/core/MailService.groovy
```diff
@@ -201,7 +202,7 @@ class MailService {
             email.send()
             return true
         } catch (Exception e) {
-            log.error("could not send ${summary}", e)
+            log.error("could not send ${summary}: ${ExceptionUtil.summarize(e)}")
```
mdpearson, line 205: Instead of dumping hundreds of lines of javax.mail traceback, we now just pull out the "Caused by" lines. Much easier for a human to read

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,137 @@
...
+        java-version: 7
+
+    - name: configure node 14
+      uses: actions/setup-node@v3
+      with:
+        node-version: 14
+
+    - name: initialize MySQL
+      run: |
+        sudo systemctl start mysql.service
```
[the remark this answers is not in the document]
jmiranda: Ok, we might definitely run into errors using MySQL 8 (https://github.com/openboxes/openboxes/issues?q=is%3Aissue+is%3Aopen+mysql) but we can address them as they come up.

.github/workflows/integration-tests.yml
jmiranda: If there's test data in preload_data.sql that is not in the Liquibase data migrations then this is probably fine and probably the preferred method as this would then be a test resource.

However if preload_schema.sql and preload_data.sql create the same database you'd get if you ran through the database migrations on an empty database, then I'd like for this to be more generic (it's no longer just a test resource). 

For example, for every new release (or after some other arbitrary milestone - every two months - it doesn't really matter) we should regenerate the baseline.sql and baseline-data.sql scripts (or just have one baseline.sql with everything). This would include the DATABASECHANGELOG table and include all of the database migrations that have been created up to this point. After installing this database, we start the application and the rest of the Liquibase migrations will be executed. And then the tests will be executed. 

The closer the baseline.sql is to the current data model, the less time will need to be spend running the migrations. But we could be on a very old baseline.sql and the outcome will be the same. The reason I want this to be generic is because it's not just applicable to testing. We have severs that would benefit from a clean build / deploy process as well and we could hook this script into that process.

With that said, we can keep this solution for now since it solves the immediate issue. But we should add a tech debt ticket to make this more generic.

.github/workflows/integration-tests.yml
jmiranda: Feel free to make suggestions if you have any thoughts about how to fix that issue (or any of these issues (https://github.com/openboxes/openboxes-docker/issues)

grails-app/services/org/pih/warehouse/core/MailService.groovy
jmiranda: I'm not sure I agree with optimizations like this. As an example, I just ran into an issue where java decided to disappear stacktraces of NPEs after the first occurrence in the log file. Since I didn't have access to the server or logs, I had no way of finding out what the stacktrace was so I had to do a ton of work just to figure it out. In general, I prefer verbosity over The Nothing so unless this was 100% guaranteed to remove obfuscation without potentially causing important bits to disappear, then we shouldn't do it.

https://pihemr.atlassian.net/browse/OBKN-668?focusedCommentId=133280

mdpearson: Resolved in e7740e183

awalkowiak (approved):

.github/workflows/integration-tests.yml
jmiranda: Doesn't look like this was resolved since we're still sending this to .properties file instead of .groovy.

test/integration/org/pih/warehouse/importer/InventoryExcelImporterTests.groovy
```diff
@@ -39,24 +26,24 @@ class InventoryExcelImporterTests extends GroovyTestCase {
...
-                [productCode:"00005", product:"Similac Advance low iron 400g", lotNumber:"HI12034", expirationDate:"15-09", binLocation:"Salle 3", quantityOnHand:11525.0, quantity:11525.0, comments:"Comment 14"],
-                [productCode:"00006", product:"Similac Advance + iron 365g", lotNumber:"GA12083", expirationDate:"15-07", binLocation:"Salle 3", quantityOnHand:90.0, quantity:90.0, comments:"Comment 15"],
-                [productCode:"00007", product:"MacBook Pro 8G", lotNumber:"CK12026", expirationDate:"15-10", binLocation:"Salle 3", quantityOnHand:7600.0, quantity:7600.0, comments:"Comment 16"],
-                [productCode:"00008", product:"Print Paper A4", lotNumber:"7516", expirationDate:"15-11", binLocation:"Salle 3", quantityOnHand:700.0, quantity:700.0, comments:"Comment 17"],
-                [productCode:"00008", product:"Print Paper A4", lotNumber:"7518", expirationDate:"16-04", binLocation:"Salle 3", quantityOnHand:6800.0, quantity:6800.0, comments:"Comment 18"]
+                [productCode:"00001", product:"Advil 200mg", lotNumber:"M18", expirationDate:"15-11", binLocation:"Salle 1", quantityOnHand:252.0, quantity:2520.0, comments:"Comment 1"],
```
jmiranda, line 29: Any reason why the quantityOnHand value was changed?

test/integration/org/pih/warehouse/inventory/InventoryServiceTests.groovy
```diff
@@ -63,11 +64,11 @@ class InventoryServiceTests extends GroovyTestCase {
         supplierLocationType = LocationType.get(Constants.SUPPLIER_LOCATION_TYPE_ID)
 
         // get or create a default location
-        acmeLocation = DbHelper.creatLocationIfNotExist("Acme Supply Company", supplierLocationType)
+        acmeLocation = DbHelper.getOrCreateLocation('Acme Supply Company', supplierLocationType)
```
jmiranda, line 67: For future reference, I prefer (and have tried to use) the convention findOrCreate since it matches what is available in later versions of Grails.

jmiranda (changes requested):

test/integration/org/pih/warehouse/importer/InventoryExcelImporterTests.groovy
mdpearson: Yeah, as the test data was originally written, QoH and quantity have the same value, which makes it hard to tell if the columns get swapped. While repairing this text fixture I made QoH 1/10th of quantity in every row so that the test will catch column shifts/swaps.

test/integration/org/pih/warehouse/inventory/InventoryServiceTests.groovy
mdpearson: That makes a lot of sense. Resolved in `b28afe136`

grails-app/domain/org/pih/warehouse/product/Product.groovy
mdpearson: OBGM-341

.github/workflows/integration-tests.yml
mdpearson: You bet. If it comes up again in Grails 3 I'll look at it a bit more closely.

test/integration/org/pih/warehouse/importer/InventoryExcelImporterTests.groovy
mdpearson: @jmiranda Let me know if the above explanation passes muster

.github/workflows/integration-tests.yml
mdpearson: Resolved in ed446e3de.

jmiranda (approved):

PR #3371, opened 2022-07-19
OBPIH-3859 Remove unnecessary groupings in Order Summary views

awalkowiak:


jmiranda (commented):
My only concern with removing the group by columns is tied to the fact that Mysql complains a lot about this unless you have set up a rule to ignore it.
https://stackoverflow.com/questions/41887460/select-list-is-not-in-group-by-clause-and-contains-nonaggregated-column-inc

I don't want to make it so users need to configure mysql.cnf to make this work.

awalkowiak:
@jmiranda I had to reinstate the grouping by `shipment.id` because i found a bug related to it on the PO view page

PR #3372, opened 2022-07-19
OBPIH-4601 Invalid # items on outbound list when revise and substitut…

kchelstowski:
…e items

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
@@ -9,6 +9,7 @@ import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.OrderItem
 import org.pih.warehouse.requisition.Requisition
 import org.pih.warehouse.requisition.RequisitionItem
+import org.pih.warehouse.requisition.RequisitionItemStatus
 import org.pih.warehouse.requisition.RequisitionSourceType
 import org.pih.warehouse.requisition.RequisitionStatus
 import org.pih.warehouse.requisition.RequisitionType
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
```diff
@@ -124,7 +125,16 @@ class OutboundStockMovementListItem implements Serializable {
     }
 
     Integer getLineItemCount() {
-        return requisition ? RequisitionItem.countByRequisition(requisition) : order ? OrderItem.countByOrder(order) : 0
+        if (requisition) {
+            RequisitionItem[] approvedItems = requisition.requisitionItems.findAll{ it ->
```
jmiranda, line 129: I'm pretty sure findAll would return a List if you invoked it on a List (I'm almost positive that requisitionItems is a List) Does this work actually work? Does it work because of type coercion?

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
```diff
@@ -124,7 +125,16 @@ class OutboundStockMovementListItem implements Serializable {
...
+                it.status == RequisitionItemStatus.APPROVED
```
jmiranda, line 130: What happens when a requisition item has been substituted or modified?

jmiranda (commented):
In general, I love the idea of getting rid of the extra countBy query so this is heading in the right direction.

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
kchelstowski: yeah, it works, because `requsitionItems` is a list of `RequsitionItem`'s elements. I attach a screen below:
[screen](https://i.imgur.com/6LKeqE5.png)

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
kchelstowski: According to the [screen](https://i.imgur.com/6LKeqE5.png) if an item is substituted or its `quantityRevised` changes - this item still appears in `requsitionItems` with status `SUBSTITUTED`/`CHANGED` and another one is 'created' with `APPROVED` status (for example take a look at `10045` in terms of substituting the item and at `10026` in terms of changing the `quantityRevised` (there are 'two' 10045 and 'two' 10026) and it was double counted, so according to [this screen](https://i.imgur.com/Xzlab9k.png) it was showing wrong amount of items as '10' (additional two because of substitutions [18586 and 10045] and additional two because of changing the `quantityRevised` [10026 and 11587])

awalkowiak (approved):

PR #3374, opened 2022-07-21
OBPIH-4277 Add option to make separate dashboard pages

kchelstowski:
Best to test locally.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -404,25 +404,23 @@ class UserService {
 
     def getDashboardConfig(User user, String id) {
         def fullConfig = grailsApplication.config.openboxes.dashboardConfig
+        def mainDashboardId = grailsApplication.config.openboxes.dashboardConfig.mainDashboardId
         def resultConfig = [
-                dashboards: fullConfig.dashboards[id ?: "mainDashboard"],
+                dashboard: fullConfig.dashboards[id ?: mainDashboardId],
                 dashboardWidgets: fullConfig.dashboardWidgets
         ]
         def userConfig = user.deserializeDashboardConfig()
-        Boolean configChanged = false
 
         if (userConfig != null) {
-            int userConfigSize = userConfig.size()
-            int resultConfigSize = resultConfig.dashboards.size()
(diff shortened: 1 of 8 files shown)


grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -402,27 +402,31 @@ class UserService {
...
         Boolean configChanged = false
 
         if (userConfig != null) {
             int userConfigSize = userConfig.size()
-            int configSize = config.dashboards.size()
+            int resultConfigSize = resultConfig.dashboards.size()
             // If the size is different, that mean that the config has changed
-            if (userConfigSize != configSize) {
-                return config
+            if (userConfigSize != resultConfigSize) {
```
awalkowiak, line 418: If the userConfig (which is based on the main dashboard) will have the same size as for example supplier dashboard stored in the resultConfig, then the user config will be returned instead of another dashboard by id.

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -402,27 +402,31 @@ class UserService {
         return false
     }
 
-    def getDashboardConfig(User user) {
-        def config = grailsApplication.config.openboxes.dashboardConfig
+    def getDashboardConfig(User user, String id) {
+        def fullConfig = grailsApplication.config.openboxes.dashboardConfig
+        def resultConfig = [
+                dashboards: fullConfig.dashboards[id ?: "mainDashboard"],
```
awalkowiak, line 408: This could be now singular (`dashboard`).

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -402,27 +402,31 @@ class UserService {
...
+                dashboardWidgets: fullConfig.dashboardWidgets
+        ]
         def userConfig = user.deserializeDashboardConfig()
         Boolean configChanged = false
```
awalkowiak, line 412: I think this boolean has no sense right now. Could you remove it?

awalkowiak (changes requested):
@kchelstowski Overall code looks good, I found one issue that needs to be resolved.

awalkowiak (approved):

PR #3380, opened 2022-07-25
OBPIH-3960 Fix managers with invoice user permission cannot delete item from invoice

drodzewicz:


grails-app/conf/RoleFilters.groovy
```diff
@@ -85,30 +85,37 @@ class RoleFilters {
...
 
-                    if (rule.access?.contains(RoleType.ROLE_ANONYMOUS) || (session.user && userService.isUserInRole(session.user, rule?.access))) {
-                        log.info "User has access to ${controllerName}.${actionName}"
+                    Boolean isUserInRole = true
+                    if (session.user && supplementalRoles.size() > 0) {
+                        isUserInRole = userService.isUserInRole(session.user, supplementalRoles)
                     }
-                    else {
+
+                    if (session.user && isMinimumRequiredRole && isUserInRole) {
```
jmiranda, line 105: Shoot. I think I either messed this up during the tech huddle or miscommunicated. We still need the check on whether the access rule allows anonymous access. That condition should also return true. If this is done in the isUserInRole() method then that would be fine. I'll need to go back and see what the logic looked like when we were discussing it on the tech huddle because I thought there was something different about the anonymous logic that I didn't like.

jmiranda: Nevermind you force pushed. If you can remember what the logic looked like last week, can you paste it here?

drodzewicz: the logic was 
`def isAnonymous = supplementalRoles?.contains(RoleType.ROLE_ANONYMOUS)`

and in an if statement we check `isAnonymous` or _other..._

```
if (isAnonymous || (session.user && isMinimumRequiredRole && isUserInRole)) {
    log.info "User has access to ${controllerName}.${actionName}"
    return true
}
```

the problem that you had with `isAnonymous ` from what I remember is that `RoleType.ROLE_ANONYMOUS` should be checked in `minimumRequiredRole ` and not in `supplementalRoles`
but then you said that it is probably unnecessary because if anonymous user has access it is checked here

```
 if (!rule) {
     log.info "No rule for ${controllerName}:${actionName} -> allow anonymous"
```

jmiranda: No I see the problem. We don't care if the user has the anonymous role. We just want to check if the access rule itself allows anonymous users. An anonymous user means that the user does not need to be authenticated. So restore the code that I originally had related to rules allowing anonymous access.

jmiranda (approved):

PR #3381, opened 2022-07-25
OBPIH-3787 Add amount outstanding on orders report

awalkowiak:


jmiranda (approved):
This is a monster; I don't have any major issues with it but we might want to localize the column names to save us from having to do that in the future. With that said, if we already have a bunch of CSV reports that will need to be localizable then it's probably fine.

grails-app/migrations/views/order-summary-view.sql
```diff
@@ -13,7 +13,7 @@ CREATE OR REPLACE VIEW order_item_status AS
             product.product_code        AS product_code,
             order_item.quantity * order_item.quantity_per_uom    AS quantity_ordered, -- to compare with shipped quantity which is already multiplied by qty per uom
             CASE
-                WHEN shipment.current_status IN ('SHIPPED', 'PARTIALLY_RECEIVED', 'RECEIVED') THEN SUM(shipment_item.quantity)
+                WHEN COALESCE(shipment.current_status) IN ('SHIPPED', 'PARTIALLY_RECEIVED', 'RECEIVED') THEN SUM(shipment_item.quantity)
```
jmiranda, line 16: What does COALESCE do here?

awalkowiak: You are right this is wrong approach. I added the shipment into the group by (but it will need testing on the multiple shipments per order item case).

PR #3387, opened 2022-08-02
OBGM-335 Fix "blank page" - broken SecurityInterceptor

awalkowiak:


grails-app/controllers/org/pih/warehouse/SecurityInterceptor.groovy
```diff
@@ -128,5 +128,7 @@ class SecurityInterceptor {
             redirect(controller: 'dashboard', action: 'chooseLocation')
             return false
         }
+
+        return true
```
awalkowiak, line 132: @mdpearson this is the issue that caused the blank page (not anything related to the bundle or context path as we expected). I think I accidentally removed it while rebasing. I fixed a few other issues I run into. 

cc @jmiranda

mdpearson: nice find! 🔍

mdpearson (approved):
@awalkowiak this looks good to me, and more to the point, works well with my forthcoming OBGM-340 work. Thank you

jmiranda (approved):

PR #3388, opened 2022-08-02
OBPIH-4001 Make admin, manager, assistant be able to delete pending POs

kchelstowski:
The main problem to fix was that the `Delete order` button was visible only for superusers and that I had to figure out how to prevent `<g:link>` of disabling the button after making is visible not only for superusers.
Deleting now works for role at least Assistant, so for roles: Superuser, Admin, Manager, Assistant.

### DIFF
--- a/grails-app/views/order/_actions.gsp
+++ b/grails-app/views/order/_actions.gsp
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.core.ActivityCode" %>
+<%@ page import="org.pih.warehouse.core.RoleType; org.pih.warehouse.core.ActivityCode" %>
 <%@ page import="org.pih.warehouse.core.Constants" %>
 <%@ page import="org.pih.warehouse.core.Location" %>
 <%@ page import="org.pih.warehouse.order.OrderStatus" %>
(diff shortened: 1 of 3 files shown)


grails-app/views/order/_actions.gsp
```diff
@@ -127,20 +129,9 @@
 								&nbsp;${warehouse.message(code: 'order.deleteOrder.label')}
 							</g:link>
 						</div>
-					</g:supports>
-				</g:if>
-				<g:if test="${orderInstance?.orderType == PUTAWAY_ORDER && orderInstance?.status != OrderStatus.COMPLETED}">
```
awalkowiak, line 132: @kchelstowski why this part was removed?

kchelstowski: I’m sorry, it was probably a missclick while refactoring if’s structure to look cleaner and by mistake I probably deleted this. Will add it back asap

grails-app/views/order/_actions.gsp
```diff
@@ -92,9 +92,9 @@
 					</g:link>
 				</div>
 			</g:elseif>
-			<g:isSuperuser>
-				<g:if test="${orderInstance?.orderType == PURCHASE_ORDER}">
-					<g:supports activityCode="${ActivityCode.PLACE_ORDER}">
+			<g:if test="${orderInstance?.orderType == PURCHASE_ORDER}">
+				<g:supports activityCode="${ActivityCode.PLACE_ORDER}">
+					<g:isSuperuser>
```
jmiranda, line 97: Since this was already there we can leave (and refactor later) but just wanted to point out that this could be a performance issue since _actions.gsp is likely being rendered for each order on the List Orders page. The max results parameter is usually set to 10, so this would be rendered at most 10 times. But it's still not great, especially since it's being called multiple times within the _actions.gsp and each time it's called there are likely database queries involved.

jmiranda: With that said, I don't know the best approach here. I think Spring Security uses a thread local to store the authenticated user and their authorities. We do something similar with AuthService but it's not widely used and as far as I can remember I don't think it's being invoked in g:isSuperuser.

jmiranda (approved):

PR #3389, opened 2022-08-02
OBPIH-3787 Improvements to the amount outstanding on order report

awalkowiak:


grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -1635,4 +1635,35 @@ class DocumentService {
...
+    void generateAmountOutstandingOnOrdersxcel(OutputStream outputStream, List<Map> data) {
+        try {
+            Workbook workbook = new HSSFWorkbook()
+            HSSFSheet sheet = workbook.createSheet("Sheet1")
+            createExcelHeader(sheet, 0, data.get(0).keySet().toList())
+            data.eachWithIndex { Map dataRow, index ->
+                createExcelRow(sheet, index + 1, dataRow)
+            }
+
+            sheet.setColumnWidth((short) 0, (short) ((50 * 4) / ((double) 1 / 20))) // Supplier
```
awalkowiak, line 1647: @jmiranda what is behind this width calculation? I used this as it is common within this service but I am curious about what it is based on.

jmiranda: I don't remember, but there was some reason for it or else I wouldn't have done it. I'd love to remove it though.

jmiranda: It would be great to either remove it or make it generic so we don't need more than one generateExcel() methods.

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -1635,4 +1635,35 @@ class DocumentService {
         return documents
     }
 
+    void generateAmountOutstandingOnOrdersxcel(OutputStream outputStream, List<Map> data) {
```
jmiranda, line 1638: If you do need to keep this fix the method name to include the E in Excel.

generateAmountOutstandingOnOrdersExcel

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
awalkowiak: I don't need this, I prefer the generic version. This was requested by QA. I'll ask Kelsey on the standup if this is something she wants.

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
awalkowiak: For now, I removed it. If Kelsey will want this, then I add it back.

PR #3393, opened 2022-08-04
OBPIH-4002 Admins cannot delete documents

kchelstowski:
Making admins be able to delete documents from PO

### DIFF
--- a/grails-app/views/order/_orderDocuments.gsp
+++ b/grails-app/views/order/_orderDocuments.gsp
@@ -26,11 +26,14 @@
                             <g:link action="editDocument" id="${documentInstance.id}" params="['order.id':orderInstance?.id]">
                                 <img src="${createLinkTo(dir:'images/icons/silk',file:'page_edit.png')}" alt="Edit" />
                             </g:link>
-                            <g:isUserAdmin>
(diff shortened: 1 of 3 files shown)


grails-app/views/order/_orderDocuments.gsp
```diff
@@ -26,9 +26,11 @@
                             <g:link action="editDocument" id="${documentInstance.id}" params="['order.id':orderInstance?.id]">
                                 <img src="${createLinkTo(dir:'images/icons/silk',file:'page_edit.png')}" alt="Edit" />
                             </g:link>
-                            <g:link action="deleteDocument" id="${documentInstance.id}" params="['order.id':orderInstance?.id]" onclick="return confirm('${warehouse.message(code: 'default.button.delete.confirm.message', default: 'Are you sure?')}');">
-                                <img src="${createLinkTo(dir:'images/icons',file:'trash.png')}" alt="Delete" />
-                            </g:link>
+                            <g:isUserAdmin>
```
jmiranda, line 29: As I mentioned in the previous PR, we should try not to use the static isUser* taglib any longer because these are configurable. The best approach would be to modify the LinkTagLib to handle this. Or always show the link and allow the auth mechanism to determine with the user has access.

jmiranda (changes requested):

awalkowiak (approved):

PR #3394, opened 2022-08-05
OBPIH-3860 Partially invoiced status in order item and order summary …

awalkowiak:
…views

### DIFF
--- a/grails-app/migrations/views/order-summary-view.sql
+++ b/grails-app/migrations/views/order-summary-view.sql
@@ -2,50 +2,46 @@ CREATE OR REPLACE VIEW order_item_status AS
     SELECT
         order_id,
         order_number,
-        product_code,
         quantity_ordered,
-        quantity_shipped,
+        SUM(quantity_shipped) AS quantity_shipped,
         order_item_id
     FROM (
-        SELECT order.id                 AS order_id,
+        SELECT
+            `order`.id                  AS order_id,
             `order`.order_number        AS order_number,
             order_item.id               AS order_item_id,
-            product.product_code        AS product_code,
-            order_item.quantity * order_item.quantity_per_uom    AS quantity_ordered, -- to compare with shipped quantity which is already multiplied by qty per uom
+            order_item.quantity         AS quantity_ordered,
             CASE
-                WHEN shipment.current_status IN ('SHIPPED', 'PARTIALLY_RECEIVED', 'RECEIVED') THEN SUM(shipment_item.quantity)
+                -- quantity divided by order_item.quantity_per_uom to match other quantities that are in uom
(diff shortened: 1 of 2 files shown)


grails-app/migrations/views/order-summary-view.sql
```diff
@@ -2,50 +2,46 @@ CREATE OR REPLACE VIEW order_item_status AS
...
 CREATE OR REPLACE VIEW order_receipt_status AS
     SELECT
         order_id,
         order_item_id,
         order_number,
-        quantity_ordered,
-        quantity_received,
-        quantity_canceled
+        SUM(quantity_received) AS quantity_received,
+        SUM(quantity_canceled) AS quantity_canceled
```
jmiranda, line 35: What's the reason behind removing quantity ordered. Are we just getting that from a different view?

grails-app/migrations/views/order-summary-view.sql
```diff
@@ -55,74 +51,65 @@ CREATE OR REPLACE VIEW order_receipt_status AS
...
         order_number,
         order_item_id,
-        quantity_ordered,
-        shipment_item_quantity_invoiced,
-        order_status,
-        (quantity_ordered > 0)                                AS shipment_payment_ordered,
-        (shipment_item_quantity_invoiced >= quantity_ordered) AS shipment_item_invoiced,
-        invoice_submitted
+        SUM(quantity_ordered)  				AS quantity_ordered,
+        IFNULL(SUM(quantity_invoiced), 0) 	AS quantity_invoiced
```
jmiranda, line 62: I wonder if it would be a good iea to use the IFNULL convention for all of these SUMs.

jmiranda (commented):
All of these views are fine for now because we don't have many orders or invoices. But eventually we might start to see performance degradation and it'll happen quickly. So we should have a plan of attack for creating materialized views or some other way to store / retrieve this data that doesn't require building a temporary table for every query.

grails-app/migrations/views/order-summary-view.sql
awalkowiak: I think I wanted to base the quantity ordered on the order item summary, and not have the same quantities in each of these views and keep each "minor" view tied to specific values:
- **order_item_status** for quantity_ordered and quantity_shipped,
- **order_receipt_status** for quantity_received and quantity_canceled,
- **order_item_payment_status** for quantity_invoiced for order items,
- **order_adjustment_payment_status** for quantity_ordered (which is used for adjustment items count), quantity_invoiced (which is used for adjustment items invoiced count)

grails-app/migrations/views/order-summary-view.sql
awalkowiak: I added it in some other PR, because it turned out it was a good idea

PR #3396, opened 2022-08-05
OBGM-342 Fight our way out of dependency hell, with comments

mdpearson:
This PR builds on OBGM-340's warfile fix to make unit and integration tests at least not crash. The main parts of it are (a) making sure that all jars from a product family are the same version and (b) enabling `failOnVersionConflict()` which is handy for detecting when two dependencies request different versions of the same underlying library. There are a dozen or so security patches in here as well, and I added a brief comment to each dependency explaining what it was used for. Hopefully that's helpful to future versions of ourselves.

@awalkowiak Java-dependency-wise this should be a pretty good baseline for the foreseeable future for Grails 3; you can merge it into the base branch at your leisure.

`grails run-app` - works
`grails war` - works
`grails test-app` - doesn't crash
`grails gradle projectReport` - works correctly

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -8,7 +8,7 @@
 
 ---
 grails:
-    profile: web
+    profile: react-webpack
     codegen:
         defaultPackage: org
     dbconsole:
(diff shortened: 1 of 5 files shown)


awalkowiak (approved):
I am giving an approve, but I have not went through it line by line. @jmiranda could you take a look at it too?

grails-app/conf/application.yml
```diff
@@ -8,7 +8,7 @@
 
 ---
 grails:
```
jmiranda, line 10: Need to create a new ticket to make sure we've migrated everything from Config.groovy over to application.yml.

jmiranda (approved):

PR #3401, opened 2022-08-08
OBKN-4150 Table 'openboxes.product_demand_details' doesn't exist

jmiranda:


grails-app/migrations/views/changelog.xml
```diff
@@ -5,6 +5,13 @@
 
   <changeSet id="1580848680306-1" author="jmiranda" runOnChange="true" runAlways="true" failOnError="true">
     <sqlFile path="views/product-demand-view.sql"/>
+  </changeSet>
+  <changeSet id="1580848680306-1b" author="jmiranda" runOnChange="true" runAlways="true" failOnError="true">
+    <preConditions onFail="MARK_RAN">
```
awalkowiak, line 10: @jmiranda how this check helps if later the issue is that somehow it does not exist?

awalkowiak: Or perhaps the issue was that it had two sqlFiles in one changeSet?

jmiranda: @awalkowiak It'll still take the same time the first time it's run, but won't run if the table already exists. This prevents an unnecessary execution of this script (which incidentally gets run in the background every hour or so).

awalkowiak (approved):

PR #3403, opened 2022-08-09
OBPIH-4419 ignore prepayment invoices

drodzewicz:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -704,7 +704,7 @@ class OrderService {
                     }
 
                     BigDecimal parsedUnitPrice = CSVUtils.parseNumber(unitPrice, "unitPrice")
-                    if (orderItem.hasInvoices && orderItem.unitPrice != parsedUnitPrice) {
+                    if (orderItem.hasInvoices && !orderItem.hasPrepaymentInvoice && orderItem.unitPrice != parsedUnitPrice) {
```
awalkowiak, line 707: What if there will be a prepayment invoice and a final invoice? Perhaps it would be better to check if `orderItem.hasRegularInvoice`

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -742,7 +742,7 @@ class OrderService {
                     }
                     BudgetCode budgetCode = BudgetCode.findByCode(code)
 
-                    if (orderItem.hasInvoices && orderItem.budgetCode != budgetCode) {
+                    if (orderItem.hasInvoices && !orderItem.hasPrepaymentInvoice && orderItem.budgetCode != budgetCode) {
```
awalkowiak, line 745: Same as above

awalkowiak (changes requested):

awalkowiak (approved):

PR #3406, opened 2022-08-10
OBPIH-4668 Fix Lag time when downloading US PO Template

drodzewicz:


drodzewicz:
After investigating this issue I found out that the problem was in fact in the template itself, more precisely in the defined variables in the template. 

The issue of lag time while downloading was caused not by amount of order items but by data in **order.origin.address.*** and **order.deliverto.address.*** , the fact that each field was nested deeply inside an order object _(like `order.deliverto.address.address2`)_ made it quite hard for the library `fr.opensagres.xdocreport to proccess` the data into a report. 

Solution to this problem that I found is to avoid defining nested variables inside these report templates and keep the template as simple as possible. In my solution in this pr I explicitly defined fields that are going to be used in the report so I moved this 4 level nested data `order.deliverto.address.city` into just 2 level `order['deliverto.city']`


| Original template | Updated Template  |
| ---- | ---- |
| image | image |

jmiranda (approved):

jmiranda (approved):

PR #3408, opened 2022-08-11
OBPIH-4701 Fixed import new items to PO

drodzewicz:


drodzewicz:
when working on [OBPIH-4419](https://pihemr.atlassian.net/browse/OBPIH-4419) I must have not checked import of new items in a PO and introduced a regression.
It appears that accessing `orderItem.hasRegularInvoice` on a new imported item throws a nullpointer exception which I did not expect.
My fix in this pr is to first check if `orderItem` has and `id` and if it does  all other transients will be available.
Do you think it would be a good idea? @awalkowiak

awalkowiak (approved):
I think it should be fine. The issue is that while looking for invoices we are passing the order item's id into the query.

PR #3412, opened 2022-08-12
OBPIH-4677 PO with canceled items only should not be visible on PO sh…

kchelstowski:
…ipment dropdown

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
@@ -47,12 +47,18 @@ class CombinedShipmentItemApiController {
         def vendor = Location.get(params.vendor)
         def destination = Location.get(params.destination)
         List<Order> orders = orderService.getOrdersForCombinedShipment(vendor, destination)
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -47,12 +47,18 @@ class CombinedShipmentItemApiController {
...
+                    it.orderItems.any { item -> item.getQuantityRemainingToShip() > 0 } &&
+                    !it.orderItems.every { OrderItem item -> item.canceled}
+                }
+                .collect {
+                    [
+                        id: it.id,
+                        orderNumber: it.orderNumber
+                    ]
+                }
+        ] as JSON)
```
jmiranda, line 61: I'd feel more comfortable if this was either 

``` 
    def data = ...
    render([data: data] as JSON)
```

jmiranda: In fact, the code here is complex enough that it probably deserves to be migrated to the service so we can write tests for it. Since you haven't changed much about the original code (aside from formatting), we can leave it as-is, but I'd like to refactor this a bit when we get a chance.

jmiranda (approved):
Approved, but we should probably think about making the code more readable (including a comment about what's happening).

PR #3413, opened 2022-08-12
OBPIH-4581 Allow autoincrement on default product type

awalkowiak:
… type


grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -261,9 +271,61 @@ class IdentifierService {
...
+     *
+     * */
+    def getNextSequentialProductCodeForDefaultType() {
+        def sequenceNumber = ConfigurationHolder.config.openboxes.identifier.sequenceNumber.format
+        def sequenceNumberRegex = sequenceNumber?.collect { getCharacterOptions(it) }?.join('')
+        def sequenceRegexPattern = ~/^${sequenceNumberRegex}$/
+        Product latestProduct = Product.findAllByProductTypeIsNull()?.findAll {
+            it.productCode =~ sequenceRegexPattern
+        }?.sort {a, b -> b.productCode <=> a.productCode }?.first()
+        def latestProductCode = latestProduct?.productCode?.toInteger()
```
awalkowiak, line 318: ?: ConfigurationHolder.config.openboxes.identifier.sequenceNumber.start

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -261,9 +271,61 @@ class IdentifierService {
...
+                characterOptions = ConfigurationHolder.config.openboxes.identifier.numeric
+                break
+            case 'L':
+                characterOptions = ConfigurationHolder.config.openboxes.identifier.alphabetic
+                break
+            case 'A':
+                characterOptions = ConfigurationHolder.config.openboxes.identifier.alphanumeric
+                break
+            default:
+                characterOptions = ''
```
awalkowiak, line 302: we should for now only allow `0, N, and D`, because we do not support alphabetic sequence increments
if we get anything else, then we probably (?) should throw an exception here

awalkowiak: L and A might be used when we get the "prefix" and delimeter into account

awalkowiak:
@jmiranda this one is ready for review. I gave up on the previous version because it was generating more "what ifs" than it made sense. I added the `DEFAULT` Product Type to the DB and added the option to change the current sequence number for the super users.

grails-app/conf/Config.groovy
```diff
@@ -1081,6 +1081,7 @@ openboxes.identifier.productCode.generatorType = IdentifierGeneratorTypeCode.SEQ
 openboxes.identifier.productCode.delimiter = Constants.DEFAULT_IDENTIFIER_SEPARATOR
 openboxes.identifier.productCode.format = "\${productTypeCode}\${delimiter}\${sequenceNumber}"
 openboxes.identifier.productCode.properties = ["productTypeCode": "code"]
+openboxes.identifier.defaultProductType.id = "DEFAULT"
```
jmiranda, line 1084: We should use the code property here, right?

grails-app/conf/templates/configuration/products.csv
```diff
@@ -1,58 +1,58 @@
...
-,QS823,,"Coat, Laboratory, Medium, Unisex",Lab,,,Each,Tertiary Care,20,,,,,,,,,,,,,,,,
-,SJ489,,"Coat, Laboratory, Small, Unisex",Lab,,,Each,Tertiary Care,20,,,,,,,,,,,,,,,,
-,PK000,,"Ethanol (ethyl alcohol), 95%, reagent grade, 16oz",Lab,,,Each,Tertiary Care,4.71,,,,TRUE,,,,,,,,,,,,
-,HG966,,"Tube, Blood collection, K2 EDTA, Lavender top, Plastic, 4mL",Lab,,,Each,Tertiary Care,0.068,TRUE,,,,,,,,,,,,,,,
-,DJ064,,"Tube, Blood collection, K2 EDTA, Whole blood, Lavender, 2mL (pediatric)",Lab,,,Each,Tertiary Care,0.068,TRUE,,,,,,,,,,,,,,,
-,TN940,,"Tube, Blood collection, Serum, Red, Plastic, 3mL (pediatric)",Lab,,,Each,Tertiary Care,0.0692,TRUE,,,,,,,,,,,,,,,
-,KY133,,"Tube, Blood collection, Serum, Red, Plastic, 4mL",Lab,,,Each,Tertiary Care,0.0845,TRUE,,,,,,,,,,,,,,,
+Id,ProductCode,ProductType,Name,Category,GLAccount,Description,UnitOfMeasure,Tags,UnitCost,LotAndExpiryControl,ColdChain,ControlledSubstance,HazardousMaterial,Reconditioned,Manufacturer,BrandName,ManufacturerCode,ManufacturerName,Vendor,VendorCode,VendorName,UPC,NDC,Created,Updated
+,QX039,Default,Morphine 10mg immediate release tablet,Pain,,,Each,Tertiary Care,0.25,TRUE,,TRUE,,,,,,,,,,,,,
```
jmiranda, line 2: Why is this lower case? I would have assumed we'd either use the id or code field, both values in those fields should be capitalized. If we're using the ProductType.name, that would be less ideal.

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -282,7 +282,7 @@ class Product implements Comparable, Serializable {
         productCode(nullable: true, maxSize: 255, unique: true)
         unitOfMeasure(nullable: true, maxSize: 255)
         category(nullable: false)
-        productType(nullable: true)
+        productType(nullable: false)
```
jmiranda, line 285: Reminder to self: Should see a database migration to set product type for all existing products.

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <property name="now" value="now()"/>
+
+  <changeSet author="awalkowiak" id="0209202212000-0">
+    <preConditions onFail="MARK_RAN">
+      <sqlCheck expectedResult="0">SELECT COUNT(*) FROM product_type WHERE id like "DEFAULT" OR code like "DEFAULT" OR name like "DEFAULT"</sqlCheck>
```
jmiranda, line 10: What's the difference between like and equality here?

jmiranda: Is like just case-insensitive equals?

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
```diff
@@ -0,0 +1,29 @@
...
+      <column name="product_type_code" value="GOOD"/>
+      <column name="product_identifier_format" value=""/>
+      <column name="sequence_number" value="0"/>
+      <column name="date_created" valueDate="${now}"/>
+      <column name="last_updated" valueDate="${now}"/>
+    </insert>
+  </changeSet>
+  <changeSet author="awalkowiak" id="0209202212000-1">
+    <sql>
+      UPDATE product SET product_type_id = (SELECT id FROM product_type WHERE id like "DEFAULT" OR code like "DEFAULT" OR name like "DEFAULT" LIMIT 1) WHERE product_type_id IS NULL;
```
jmiranda, line 26: We should have the same precondition for this changeset just to be extra cautious.

grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -382,6 +386,17 @@ class DataService {
         return product
     }
 
+    def findOrCreateProductType(String productTypeName) {
+        def productType = ProductType.findByName(productTypeName)
+        if (!productType) {
+            println("Could not find product type " + productTypeName)
```
jmiranda, line 392: i think we can use log.info or remove this

grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
```diff
@@ -17,4 +19,18 @@ class ProductTypeService {
 
         return productType.sequenceNumber
     }
+
+    def getDefaultProductType() {
```
jmiranda, line 23: This is fine as-is, but this could also be implemented as a named query on the ProductType domain.

http://docs.grails.org/latest/ref/Domain%20Classes/namedQueries.html

grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -382,6 +386,17 @@ class DataService {
         return product
     }
 
+    def findOrCreateProductType(String productTypeName) {
+        def productType = ProductType.findByName(productTypeName)
```
jmiranda, line 390: I think we should use code for lookup whenever it is available.

jmiranda (changes requested):

grails-app/conf/templates/configuration/products.csv
awalkowiak: The name is used here. It is unique too.

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
awalkowiak: Yup, I wanted to take whatever default product type someone already created

grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
awalkowiak: I initially added it in the domain, but when I was calling it in the gsp template I was getting weird exception, hence I moved it here. I'll try again with domain version.

grails-app/services/org/pih/warehouse/data/DataService.groovy
awalkowiak: Code is nullable. Name is unique and cannot be blank. So far there was looking for product type by name (for example in the importing)

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
awalkowiak: Ok

grails-app/services/org/pih/warehouse/data/DataService.groovy
awalkowiak: Ok

grails-app/conf/Config.groovy
awalkowiak: No. Code is nullable.

grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
awalkowiak: But I have not tried a named query then, perhaps it will be better. Thanks for the suggestion.

grails-app/conf/Config.groovy
jmiranda: I'm saying that code should be non nullable and unique. If we can't do it now it should be added as a future improvement.

grails-app/services/org/pih/warehouse/data/DataService.groovy
jmiranda: See my comment above.

jmiranda (approved):

PR #3414, opened 2022-08-15
OBPIH-4654 Outbound import doesnt work with lot tracking feature

kchelstowski:


src/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -275,8 +275,10 @@ class StockMovementItem {
             throw new IllegalArgumentException("Product '${productCode} ${productName}' could not be found")
         }
 
-        if (product.lotAndExpiryControl && (!expirationDate || !lotNumber)) {
-            throw new IllegalArgumentException("Both lot number and expiry date are required for the '${productCode} ${productName}' product.")
+        if (stockMovement.getStockMovementDirection(currentLocation) != StockMovementDirection.OUTBOUND) {
```
jmiranda, line 278: Something doesn't feel right about this but I don't know a better alternative. Perhaps you could leave createFromTokens() alone and add the validation to the ~importCsv closure~ stockMovementService.updateItems(). I'm not going to fight this since I think it's probably a "half a dozen of one, 6 of the other" problem but if you agree that this should be added to updateItems then please proceed with that solution.

jmiranda: And the nice thing about that solution is that you could probably just provide a custom validator on StockMovementItem.

kchelstowski: I do not think I should proceed with making it in `updateItems()` at this point, because we already had a validation for 

```
if (product.lotAndExpiryControl && (!expirationDate || !lotNumber)) {
                throw new IllegalArgumentException("Both lot number and expiry date are required for the '${productCode} ${productName}' product.")
}
```
so I only added additional `if` for something that already exists so I think it doesn't change much here, as we also have other validations inside `createFromTokens`.
Nevertheless it would be a nice idea to move all the validations inside `updateItems()`, but in my opinion there should be some kind of refactor ticket for that.
Let me know what you think about that.

jmiranda: Ok, I'll compromise. 
```
csv.toCsvReader(settings).eachLine { tokens ->
    Boolean validateLotAndExpiry = stockMovement.getStockMovementDirection(currentLocation) != StockMovementDirection.OUTBOUND
    StockMovementItem stockMovementItem = StockMovementItem.createFromTokens(tokens, validateLotAndExpiry)
    stockMovementItem.stockMovement = stockMovement
    stockMovementItem.sortOrder = sortOrder
    stockMovement.lineItems.add(stockMovementItem)
    sortOrder += 100
}
```
StockMovementItem
```
static StockMovementItem createFromTokens(String[] tokens, Boolean validateLotAndExpiry) {
  ...
  if (validateLotAndExpiry) {
    if (product.lotAndExpiryControl && (!expirationDate || !lotNumber)) {
        throw new IllegalArgumentException("Both lot number and expiry date are required for the '${productCode} ${productName}' product.")
    }
  }
  ...
}
```

jmiranda (changes requested):

jmiranda (approved):

PR #3415, opened 2022-08-16
OBPIH-4684 Allow admin/manager/assistant to delete pending putaways

drodzewicz:


grails-app/views/order/_actions.gsp
```diff
@@ -92,59 +92,59 @@
...
-										disabled="${orderInstance?.shipments || !isApprover}"
-										disabledMessage="${disabledMessage}">
-									<img src="${resource(dir: 'images/icons/silk', file: 'arrow_undo.png')}" />
-									&nbsp;${warehouse.message(code: 'order.rollbackOrderStatus.label', default: "Rollback order status" )}
-								</g:link>
-							</g:if>
-						</div>
-					</g:isSuperuser>
-					<g:isUserInRole roles="[RoleType.ROLE_SUPERUSER, RoleType.ROLE_ADMIN, RoleType.ROLE_ASSISTANT, RoleType.ROLE_MANAGER]">
+			<g:isUserInRole roles="[RoleType.ROLE_SUPERUSER, RoleType.ROLE_ADMIN, RoleType.ROLE_ASSISTANT, RoleType.ROLE_MANAGER]">
```
kchelstowski, line 95: As long as I remember @jmiranda didn't want to continue using this, when I was proceeding with this.
You should provide a solution which would also prevent user from deleting a putaway using url (it has the same URL as PO) and as long, as the permissions remain the same for PO and putaway it's ok, but I remember that @jmiranda wanted to divide PO and putaways, so you should add another rule for putaways and most probably create a new url for deleting putaways (e.g. "/putaway/delete/:id") (currently we are using /stockTransfer/delete/:id for both PO and putaway.
Anyway I think we should wait for @jmiranda to confirm my thoughts.
I'm open to help with this one if you have any troubles.

kchelstowski (changes requested):

grails-app/views/order/_actions.gsp
drodzewicz: Thanks @kchelstowski 
I have read comments from pr that you are probably talking about https://github.com/openboxes/openboxes/pull/3388 
I totally agree that this approach for checking user authorities with `<g:isUserInRole>`  is not the best idea considering performance, but maybe there should be a separate techdept  ticket created for that

drodzewicz: Also thanks for reminding me to check for the cases where user can delete an order via url `/openboxes/order/remove/:id`
I hope that checking for if OrderStatus is **PENDING** for both PO's and Putaways is going to be enough

kchelstowski: what I meant rather is not the status of order, but rather a case, where you have different permissions for putaway and PO:
for example: for deleting POs you have to be **at least** ADMIN and for putaways you have to be **at least** ASSISTANT - at the moment if we wanted to keep the approach to hide the button on the frontend side, you would just change this `<g:isUserInRole>` and the button wouldn't be visible for POs if you are for example MANAGER.
but since we are using the same url for deleting POs and putaways, what would happen if you pasted URL: `/stockTransfer/remove/:id` - we don't protect this case.
That's why we wanted to divide putaways and POs and create a different rule for putaways, which would require creating a new url, for example `/putaway/delete/:id` for putaways.

kchelstowski: This pr: #3390 is very similar, maybe the discussion there will help with understanding what I've written above?

grails-app/views/order/_actions.gsp
```diff
@@ -94,7 +94,8 @@
 			</g:elseif>
 			<g:if test="${orderInstance?.orderType == PURCHASE_ORDER}">
 				<g:supports activityCode="${ActivityCode.PLACE_ORDER}">
-					<g:isSuperuser>
+					<g:isUserInRole roles="[RoleType.ROLE_SUPERUSER, RoleType.ROLE_ADMIN, RoleType.ROLE_ASSISTANT, RoleType.ROLE_MANAGER]">
+						<g:isSuperuser>
```
jmiranda, line 98: 1. Why are we using both isUserInRole and isSuperuser?
2. We only need to specify the lowest role ROLE_ASSISTANT.

grails-app/views/order/_actions.gsp
jmiranda: @kchelstowski thanks for bring that up. I think we can ignore that for now since we don't have a requirement for separate permissions when deleting an PO vs deleting a putaway. Once that becomes a requirement we can add in the new URL mapping.

grails-app/views/order/_actions.gsp
```diff
@@ -132,19 +132,19 @@
 					</g:isUserInRole>
 				</g:supports>
 			</g:if>
-			<g:if test="${orderInstance?.orderType == PUTAWAY_ORDER && orderInstance?.status != OrderStatus.COMPLETED}">
-				<g:isSuperuser>
+			<g:elseif test="${orderInstance?.orderType == PUTAWAY_ORDER && orderInstance?.status != OrderStatus.COMPLETED}">
+				<g:isUserInRole roles="[RoleType.ROLE_SUPERUSER, RoleType.ROLE_ADMIN, RoleType.ROLE_ASSISTANT, RoleType.ROLE_MANAGER]">
```
jmiranda, line 136: Same as above here.

jmiranda (changes requested):

awalkowiak (approved):

PR #3418, opened 2022-08-16
OBPIH-4703 Make PO import not being able to assign inactive source (f…

kchelstowski:
…ixes after QA)

What wasn't working or rather what I missed, that it should be done is the case where we have for example two product sources for the same product - one inactive and one active - in this case it should match the active one instead of throwing an exception if the inactive was found as `.first()`.
I made it using sort by the `active` field and then returning the first, so I'm sure that the first returned will be active if there are active and inactive in list.
In my opinion what would make the code cleaner would be to just `findAll` active ones and don't even throw the exception, because the inactive sources wouldn't be even found - I will discuss it with Katarzyna and Kelsey, so we could fix it later while releasing (it is not a big thing, but would improve code complexity).

Second case which wasn't working is the rare case where we have a product source which has empty `supplierCode`, `manufacturerName` and `manufacturerCode`.  With my previous changes it would throw an exception if that product source were inactive, because it didn't assume that maybe a user didn't want to provide a product source and that's why he left those fields empty - I discussed it with Katarzyna and if there is such case it should find for the product source, but if it's inactive, then do NOT throw an exception and treat it as there would not be a product source for this product.
It's a release thing so you could trust in me that it should be working fine now and could be merged further not to block anything.

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
@@ -302,7 +302,8 @@ class ProductSupplierDataService {
                 'manufacturerCode': manufacturerCode,
                 'supplierCode': supplierCode,
         ])
-        def productSupplier = data ? ProductSupplier.get(data.first().id) : null
+        ArrayList<ProductSupplier> productSuppliers = data?.collect{ it -> ProductSupplier.get(it?.id) }?.sort{ !it?.active }
+        def productSupplier = !!productSuppliers ? productSuppliers?.first() : null
         return productSupplier
     }
 
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
@@ -302,7 +302,9 @@ class ProductSupplierDataService {
                 'manufacturerCode': manufacturerCode,
                 'supplierCode': supplierCode,
         ])
+        // Sort productSuppliers by active field, so we make sure that if there is more than one and one of then could be inactive, that the active ones are "moved" to the top of list
         ArrayList<ProductSupplier> productSuppliers = data?.collect{ it -> ProductSupplier.get(it?.id) }?.sort{ !it?.active }
+        // Double negation below is equivalent to productSuppliers.size() > 0 ? productSuppliers.first() : null (empty list is treated as true, so we can't do "productSuppliers ?")
         def productSupplier = !!productSuppliers ? productSuppliers?.first() : null
         return productSupplier
     }
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -679,12 +679,16 @@ class OrderService {
                                               supplier: supplier,
                                               sourceName: sourceName]
                         ProductSupplier productSupplier = productSupplierDataService.getOrCreateNew(supplierParams, false)
+                        def supplierParamFilled = supplierCode || manufacturerName || manufacturerCode
 
                         if (productSupplier) {
-                            if (!productSupplier.active) {
+                            if (!productSupplier.active && supplierParamFilled) {
                                 throw new ProductException("Product source ${productSupplier.code} for product ${productCode} is inactive")
                             }
-                            orderItem.productSupplier = productSupplier
+                            // If it matches a product source for empty params (rare case), but it's inactive, treat it as if there was not a product source
+                            if (productSupplier.active) {
+                                orderItem.productSupplier = productSupplier
+                            }
                         }
                     }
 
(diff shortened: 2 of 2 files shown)


grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -302,7 +302,8 @@ class ProductSupplierDataService {
                 'manufacturerCode': manufacturerCode,
                 'supplierCode': supplierCode,
         ])
-        def productSupplier = data ? ProductSupplier.get(data.first().id) : null
+        ArrayList<ProductSupplier> productSuppliers = data?.collect{ it -> ProductSupplier.get(it?.id) }?.sort{ !it?.active }
+        def productSupplier = !!productSuppliers ? productSuppliers?.first() : null
```
jmiranda, line 306: I don't think Groovy supports double exclamation. It'll handle the boolean without this. What you could say is 

```
!productSuppliers.empty ? productSuppliers.first() : null
```

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -302,7 +302,8 @@ class ProductSupplierDataService {
                 'manufacturerCode': manufacturerCode,
                 'supplierCode': supplierCode,
         ])
-        def productSupplier = data ? ProductSupplier.get(data.first().id) : null
+        ArrayList<ProductSupplier> productSuppliers = data?.collect{ it -> ProductSupplier.get(it?.id) }?.sort{ !it?.active }
```
jmiranda, line 305: Could you add a comment to describe what you're trying to accomplish here? Do we really want to sort by active? And in this case, what does that do? Sorts active = true first, then action = false? What about active = null?

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -679,12 +679,16 @@ class OrderService {
                                               supplier: supplier,
                                               sourceName: sourceName]
                         ProductSupplier productSupplier = productSupplierDataService.getOrCreateNew(supplierParams, false)
+                        def supplierParamFilled = supplierCode || manufacturerName || manufacturerCode
```
jmiranda, line 682: Let's come up with a better variable name here. What is the gist of this boolean value? The required fields for supplier are populated? Or something else?

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
kchelstowski: It actually does - @drodzewicz showed me "the magic" yesterday as he also tried to use it. It was helpful to shorten something like this: `productSuppliers.size() > 0 ? ... : ...`, because if a list is empty, `!!productSuppliers` returns false, and if it's not empty, it returns true, so it does work like `productSuppliers.size() > 0` but if you don't like this idea, I'll proceed with what you've written.

grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
kchelstowski: I'll add a comment - it just "pushes" an active one to the top if there are > 1 product sources where one or more of them could be inactive - I agree that anyway it can return inactive one if there are for example two inactive in the list, but the case here is to return active if there is for example: one active and one inactive instead of returning random one.
I didn't have any other idea how to "push" the active on top, because if I wanted to use `.find` by active, it wouldn't find inactive ones if there were the only one and the validation wouldn't appear.
As I wrote in one of the comments - I think the better idea here would be to use `.find` and NEVER find inactive ones, so we don't need that validation error, but just don't assign the product source if it's inactive, but it will need to be discussed with Katarzyna and Kelsey.

As for second part of question - with negation, it sorts in order: true, false - as for active = null I don't think it would ever happen since a product source is `true` by default: `Boolean active = Boolean.TRUE`. But answering - `null` is falsy, so anyway it would return the active one on the top instead of null.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
kchelstowski: yep, I also don't like this name, but I didn't have any other idea.
The gist is to determine, if a user filled any of those three fields, as by them we can search for a product source.
I had to do it, because there is a rare case where it could match a product source with all those fields empty, but what if a user didn't want to assign any product source (he left those three empty) and it matched a product source with all those three fields empty and it's an inactive product source? User would be punished for that, that this product source is inactive, but the user might think: "_What? I didn't even want to assign a product source, I left the fields empty, so what did it match_?" 
That's the reason I "mix the soup" below

kchelstowski: my only idea for name was something like this: `searchTermFieldsForSupplierFilled` but it was way too long and did not come up with anything else, but maybe this could help you with helping me to provide a better name? :laughing:

jmiranda (approved):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -679,12 +679,17 @@ class OrderService {
                                               supplier: supplier,
                                               sourceName: sourceName]
                         ProductSupplier productSupplier = productSupplierDataService.getOrCreateNew(supplierParams, false)
+                        // Check if any of search term fields for productSupplier are filled
```
jmiranda, line 682: In the future, provide the why as well as this doesn't clarify the code much.

PR #3419, opened 2022-08-17
OBPIH-4677 PO with canceled items (or at least one canceled) should n…

kchelstowski:
…ot be visible on PO shipment dropdown


jmiranda (approved):
This is why we need comments (and obviously tests). I just assumed the code was doing what it was supposed to be doing here since I wasn't sure what was expected.

PR #3427, opened 2022-08-24
OBPIH-4727 Return all default dashboard configs and saved personal dashboard config

drodzewicz:


drodzewicz:
This was something I didn't notice when testing locally because I had a fresh database, but it appears that when user is saving config on **My Dashboard** it is saving all of the dashboards and not just personal. 
Since the only dashboard that is _configurable/updatable_ is _My Dashboard/Personal_   it would make much more sense to store only personal dashboard config on user record in database and when returning the whole dashboard config via API to just append personal dashboard to the default config of all the other dashboard configs.

At the moment I came up with the solution, when returning  dashboard config to just extract **personal** dashboard config from user _(if user has one saved)_ and append it to the default configs of all other dashboards.

We should also consider that we should only save personal config in the database but I wasn't sure what to do with records that already have saved configs.

awalkowiak (approved):

PR #3432, opened 2022-08-24
OBPIH-4766 Don't implicitly close response when generating Excel files.

mdpearson:
DocumentService.generateExcel() closes the response provided to it, but 3 out of 4 callers attempt to flush the response after it returns, which won’t work. For clarity, I removed the close() call from generateExcel() and added a flush() call to the 1 caller that didn't already do so. Resolves OBENBOXES-OBNAV-Q25.


awalkowiak:
@mdpearson Is this something worth including in the 0.8.19? Or can wait for the 0.8.20?

awalkowiak (approved):

PR #3436, opened 2022-08-25
OBPIH-4764 bin import incorrect bin location type assigned

drodzewicz:


grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -476,7 +476,7 @@ class LocationService {
                 throw new ValidationException("location.cannotImportBinLocationsWithoutParentLocation.message")
             }
 
-            LocationType defaultLocationType = LocationType.findByLocationTypeCode(LocationTypeCode.BIN_LOCATION)
+            LocationType defaultLocationType = LocationType.findByLocationTypeCodeAndName(LocationTypeCode.BIN_LOCATION, "Bin Location")
```
awalkowiak, line 479: This is the wrong approach. We do not have certainty that in every instance of OB the naming will remain as it is here. I think we could try to take the oldest (date created) location type with code BIN_LOCATION. 

Additionally the same issue occurs within the regular location import in the `locationApiController.importCsv`, because there is `findByNameLike(name + '%')` which also takes the newest record. Perhaps there we could get rid of `like` with `%` wildcard option.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3438, opened 2022-08-25
OBPIH-3860 Show derived status on the orders list and details page

awalkowiak:


awalkowiak:
I am gonna merge, but I will revisit in case of any change request.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -73,8 +73,13 @@ class OrderController {
 
         def orders = orderService.getOrders(orderTemplate, statusStartDate, statusEndDate, params)
 
-        if (params.format && orders) {
+        def ordersDerivedStatus
```
jmiranda, line 76: In the future, you can just call something like this derivedOrderStatuses. Otherwise it's awkward to read.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1101,4 +1101,30 @@ class OrderService {
...
+    /**
+     * Gets map of derived status for orders (Order id as a key and derived status as a value)
+     * Done to improve performance of getting orders derived statuses on order list page
+     * */
+    def getOrdersDerivedStatus(List orderIds) {
+        if (!orderIds) {
+            return [:]
+        }
+
+        def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
```
jmiranda, line 1113: We can also use the injected messageSource for this. But this seems fine.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1101,4 +1101,30 @@ class OrderService {
...
+        def orderSummaryList =  OrderSummary.findAllByIdInList(orderIds)
```
jmiranda, line 1114: I assume this method is going to be slow due to the use of the order summary. Are we in for a rude awakening come testing? Or do you think this will perform ok?

grails-app/views/order/list.gsp
```diff
@@ -220,6 +223,19 @@
 					.datepicker('setDate', null);
 				});
 			});
+
+			function fetchOrdersDerivedStatus() {
```
jmiranda, line 227: Oh interesting. When I wrote the "rude awakening" comment I was thinking that one way to deal with this would be to pull the data in asynchronously. So I guess we won't be rudely awakened.

PR #3442, opened 2022-08-29
OBPIH-4746 Fix creating picklist and shipment items after revising

awalkowiak:


mdpearson:
Doesn't look like it'll cause problems elsewhere since there's only one caller of reviseItems. I'm curious as to _why_ the transactional context caused validation problems (since previously there was a save/flush right above the `createMissing*Items` calls). But, hibernate being what it is, if this works, then 👍 from me

mdpearson (approved):

awalkowiak:
>  I'm curious as to why the transactional context caused validation problems

@mdpearson That's my blind guess. The issue was that when removing old picklist items for revised items in the `stockMovementService.reviseItems` there is Product Availability refresh. Then while creating missing picklist items in `stockMovementService.createMissingPicklistItems`, there is a calculation of available items (taken from Product Availability). The values in available items were based on the `old` Product Availability, and after moving `createMissingPicklistItems` and `createMissingShipmentItems` values were `"fresh"`.

jmiranda (commented):
Let's put this into a tech debt ticket.

PR #3449, opened 2022-09-01
OBS-1253 Fix regression in updateProductAvailability and updateInventorySnapshots

mdpearson:
As part of restoring test coverage, I added guards to

- ProductAvailabilityService.updateProductAvailability()
- InventorySnapshotService.updateInventorySnapshots()

to prevent them from firing on objects that had no ID.
However, the change also made it impossible for these
methods to clear Product.productCode, which is an
optional field. I'm not sure if this change is causing
the performance problems, but I believe it is a bug.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
@@ -409,7 +409,7 @@ class InventorySnapshotService {
     }
 
     void updateInventorySnapshots(Product product) {
-        if (!product || !product.id || !product.productCode) {
+        if (!product?.id) {
             return
         }
         def results = InventorySnapshot.executeUpdate(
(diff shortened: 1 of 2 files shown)


awalkowiak (approved):

awalkowiak:
@mdpearson this one looks good, but I think it should not go to the master.

jmiranda (approved):
This looks fine to me. For all intents and purposes, product code is more or less required. We made it nullable so we could generate the identifier in the background. I don't exactly remember why this was done but I think there were some concerns with identifier collision, given the small universe of identifiers we could generate using the default format (AANN) which allowed for only 26 * 26 * 10 * 10 (or 67600) unique indentifiers. We probably weren't ever going to run out of product codes (SKUs), but given some probability axioms (pigeonhole principle, birthday paradox), the more identifiers we generate the more likely we'll start having collisions (eventually). I didn't know what the threshold was at the time, so I compromised. I just didn't want to prevent someone from creating a product due a validation constraint that was outside their control. Therefore, we allow the productCode to be null until it is automatically assigned an identifier about a minute after it has been saved. If the background job generates a collision, it tries again a few minutes later.

jmiranda (commented):
Agree with Artur that a change like this should only go to develop.

mdpearson:
I'd like to merge this into develop after the upcoming release.

PR #3450, opened 2022-09-01
Revert "OBPIH-4746 Fix creating picklist and shipment items after revising"

mdpearson:
This reverts commit aed8b65aab942c17f4f1ed5bbdffde2869d08cc7.

I'm wondering if pulling the following two calls

- stockMovementService.createMissingPicklistItems()
- stockMovementService.createMissingShipmentItems()

out of stockMovementService.updateRequisitionBasedStockMovementItems()'s
transactional context and into StockMovementApiController.reviseItems()
might have unanticipated performance implications? A total guess! But
perhaps worth a PR

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -157,13 +157,7 @@ class StockMovementApiController {
     def reviseItems = {
         StockMovement stockMovement = stockMovementService.getStockMovement(params.id)
         bindStockMovement(stockMovement, request.JSON)
-        // First revise the items
         stockMovementService.reviseItems(stockMovement)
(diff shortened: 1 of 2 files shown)


awalkowiak (changes requested):
@mdpearson Reverting this one introduces a bug that was found during regression. Plus as you wrote on slack, the performance issues were long before this PR (this one was one of the last in the test script cycle). If you want to try reverting this, then I'd suggest to first test it on the staging and trying to fix the issue this one was fixing in another way, and then if it is fine merge it to master directly 

cc @jmiranda

awalkowiak:
@mdpearson can this one be closed?

jmiranda (commented):
It will have negligible impact, but I think you're correct that these lines should go in reviseItems(). Unfortunately, until we solve the issue in Artur's comment, it'll probably have to stay in the controller. 

Edit: And this type of change should only go to develop. master should only get hotfixes that we're 99% sure will actually fix a particular problem.

PR #3456, opened 2022-09-09
OBPIH-4705 Fix Location chooser button

drodzewicz:
- fix long location name text overlay
- fix some padding/margin styling
- added ability to close modal on click outside

### DIFF
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -999,7 +999,7 @@ div.crossed-out .table-inner-row:before {
   }
 
   &::-webkit-scrollbar-thumb {
-    background-color: var(--blue-300);
+    background-color: var(--blue-301);
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/css/main.scss
```diff
@@ -999,7 +999,7 @@ div.crossed-out .table-inner-row:before {
   }
 
   &::-webkit-scrollbar-thumb {
-    background-color: var(--blue-300);
+    background-color: var(--blue-301);
```
awalkowiak, line 1002: Is the `301` a typo?

drodzewicz: It is not  a typo, it is a [different color](https://github.com/openboxes/openboxes/blob/feature/ui-redesign/src/css/colors.scss#L40)
Unfortunately this color was not specified on the color pallet on **figma** so I named it myself. 
Named it `blue-301`  because it is very similar to already defined color `blue-300`

awalkowiak (approved):

PR #3457, opened 2022-09-09
OBPIH-4706 Add click outside to close Global search

drodzewicz:


awalkowiak (changes requested):
Would it be enough to use `onBlur` (that's available on react-select props)?

drodzewicz:
> Would it be enough to use `onBlur` (that's available on react-select props)?

You are right, click outside hook is unnecessary and `onBlur` will be enough

awalkowiak (approved):

PR #3460, opened 2022-09-12
OBS-1253 Don't run Quartz jobs during DB migrations; consistently disable concurrent job execution.

mdpearson:
- some Quartz jobs have a `@DisallowConcurrentExecution` annotation;
- some declare `def concurrent = false`;
- some check whether a migration is happening before running;
- each one of the above is not done by at least one job :-( .

This PR makes *all* Quartz jobs perform all 3 checks and consolidates common code in a new `JobsUtil` class. (I also added comments to ease Grails 1/3 rebases going forward.)

I'm not sure whether the 0.8.19 deploy was actually affected by this; the migrations there took longer than they have on any other push so the window for error was wider -- but I think these bugs should be fixed even if they haven't bitten us in the last week or two. Let me know what y'all think.


jmiranda:
Added `do not merge` label as we should discuss this in a tech huddle.

PR #3462, opened 2022-09-13
OBPIH-4711 Add purchase order api

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/OrderApiController.groovy
```diff
@@ -0,0 +1,201 @@
...
+import org.pih.warehouse.order.OrderItemStatusCode
+import org.pih.warehouse.order.OrderStatus
+
+import java.math.RoundingMode
+
+class OrderApiController extends BaseDomainApiController {
+
+    def orderService
+
+    def list = {
```
awalkowiak, line 25: @jmiranda Initially, this was meant only for purchase orders, but I realized we currently can use the purchase orders list for looking for other types of orders too. I didn't want to lose this option after the PO list UI redesign, hence I left the option to look for other order types in this endpoint too. It is modeled based on the current `OrderController.list` action, but I took out the code related to downloading order and order items CSV into separate endpoints. For Purchase Orders, results are based on the OrderSummary (we might need to think a bit more about the materialized view before the next release), and for other Order types, it's based on the Order.

jmiranda (commented):
Oops forgot to submit this one.

PR #3468, opened 2022-09-15
OBPIH-4710 Create reusable filter component

kchelstowski:


src/js/components/Filter/FilterForm.jsx
```diff
@@ -0,0 +1,115 @@
...
+                </div>
+
+                <div className="d-flex">
+                  {!filtersHidden && _.map(filterFields, (fieldConfig, fieldName) =>
+                    renderFormField(fieldConfig, fieldName, formProps))}
+                </div>
+
+              </div>
+              <div className="submit-buttons">
+                <button
```
drodzewicz, line 85: Regarding the custom`Button` component that you would be able to use here

src/js/components/Filter/FilterForm.jsx
```diff
@@ -0,0 +1,115 @@
...
+        ...acc,
+        [curr]: '',
+      };
+    }
+    return acc;
+  }, { name: '' });
+  const [amountFilled, setAmountFilled] = useState(0);
+  const countFilled = (values) => {
+    // Calculate which object's values are not empty
+    setAmountFilled(Object.keys(values).filter(key => values[key]).length);
```
drodzewicz, line 40: A better solution would probably be to do `Object.values`
`Object.values(values).filter((val) => val).length`

drodzewicz: but also remember about `falsy` values like `0`.
If value of a filter will be `0` it will result as `false` during the filter and will not be counted

drodzewicz (changes requested):

src/js/components/form-elements/ButtonPrimary.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+const ButtonPrimary = ({ label, defaultLabel, disabled }) => (
```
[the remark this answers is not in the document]
kchelstowski: Yeah, I agree but wanted to split hem, but at one point thought that if they may be easier to use if they are separated.
As for second comment - they are global, I don't know why you think they aren't?

src/js/components/Filter/FilterForm.jsx
kchelstowski: Oh, I forgot to delete this section as in the beginning I was just copying form from some other component. It should be (and will be) deleted of course.

src/js/components/Filter/FilterForm.jsx
kchelstowski: that's nice point, I probably didn't consider this as I was only using string-like fields for testing. Thanks for that

kchelstowski:
@drodzewicz  I made `<Button>` instead of two separate `<ButtonPrimary>` `<ButtonTransparent>` and also fixed the issues you mentioned.
I also removed `<Button>` from `FilterForm.md` and added it to separate `.md` file inside `form-elements`.

drodzewicz (approved):

awalkowiak (approved):

PR #3469, opened 2022-09-15
OBPIH-4709 Redesign global search on gsp pages

drodzewicz:


awalkowiak (approved):

grails-app/views/taglib/_globalSearch.gsp
```diff
@@ -1,41 +1,93 @@
...
+      $("#global-search-input")
+        .blur(closeInputHandler)
+        .autocomplete({
+          delay: ${grailsApplication.config.openboxes.typeahead.delay},
+          minLength: ${grailsApplication.config.openboxes.typeahead.minLength},
+          position: { of: "#global-search" },
+          source: function (req, resp) {
+            $("#global-search-loading").get(0).classList.remove('d-none')
+            $.getJSON('${attrs.jsonUrl}', req, function (data) {
+              var suggestions = [];
```
kchelstowski, line 60: shouldn't we avoid using `var`? Wouldn't `const` work here?

kchelstowski (changes requested):

grails-app/views/taglib/_globalSearch.gsp
drodzewicz: I did not notice it. 
This was legacy code so I did not pay much attention to this but in context of this functions I don't think it makes much difference using **var** ,**const** or **let**

PR #3472, opened 2022-09-16
OBPIH-4734 Restyle date-picker

drodzewicz:
- restyle date picker popup
- restyle data picker input
- create a separate DateFilter component

### DIFF
--- a/src/js/components/form-elements/DateFilter/DateFilter.jsx
+++ b/src/js/components/form-elements/DateFilter/DateFilter.jsx
@@ -12,11 +12,19 @@ const CustomInput = React.forwardRef((props, ref) => {
   const {
     onClick, title, value, placeholder, onClear,
   } = props;
+
+  const onKeypressHandler = (event) => {
+    if (event.key === 'Enter') onClick();
+  };
+
   return (
-    <button
+    <div
       ref={ref}
+      tabIndex="0"
+      role="button"
       className="d-flex flex-row date-picker__input-wrapper"
       onClick={onClick}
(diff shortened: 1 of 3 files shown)


src/js/components/form-elements/DateFilter/index.jsx
```diff
@@ -0,0 +1 @@
+export { default } from './DateFilter';
```
kchelstowski, line 1: absolute import

src/js/components/form-elements/DateFilter/DateFilter.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+
+const DateFilter = (props) => {
+  const {
+    value, onChange, dateFormat, placeholder, label, timeFormat,
+  } = props;
+  const [isFocused, setIsFocused] = useState(false);
+
+  const onChangeHandler = (date) => {
+    const val = !date || typeof date === 'string' ? date : date.format(dateFormat);
```
kchelstowski, line 44: I assume it is working how it should be, but could you explain what it is exactly doing?

kchelstowski (changes requested):

src/js/components/form-elements/DateFilter/DateFilter.jsx
drodzewicz: I copied this from the **DateField** component  and looking at it again it looks unnecessary complicated so in the new commit I simplified it.

kchelstowski (approved):

awalkowiak (approved):

PR #3474, opened 2022-09-18
OBPIH-4774 Improve API endpoint for getting list of stock lists

awalkowiak:


grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -246,7 +246,7 @@ class RequisitionService {
                 } else {
                     eq("isTemplate", requisition.isTemplate)
                 }
-                if (requisition.isPublished) {
+                if (requisition.isPublished != null) {
```
jmiranda, line 249: Good catch

jmiranda (commented):
LGTM

PR #3477, opened 2022-09-20
OBPIH-4713 Create whole PO page with buttons, filters, table, API calls

kchelstowski:


grails-app/conf/BootStrap.groovy
```diff
@@ -194,13 +194,13 @@ class BootStrap {
                 name: orderSummary?.order?.name,
                 origin: orderSummary?.order?.origin?.name + (origOrgCode ? " (${origOrgCode})" : ""),
                 destination: orderSummary?.order?.destination?.name + (destOrgCode ? " (${destOrgCode})" : ""),
-                orderedOn: orderSummary?.order?.dateOrdered,
+                dateOrdered: orderSummary?.order?.dateOrdered,
```
kchelstowski, line 197: had to change it, because when fetching it was `orderedOn`, and then react-table when sorting was assigning it as `orderedOn`, but to sort by date we have to use `dateOrdered`, so not to make unnecessary ifs on the frontend side, I decided to change it here, as it is not a big 'game changer'

grails-app/conf/BootStrap.groovy
```diff
@@ -194,13 +194,13 @@ class BootStrap {
...
                 orderedBy: orderSummary?.order?.orderedBy?.getName(),
                 orderItemsCount: orderSummary?.order?.activeOrderItems?.size()?:0,
                 orderedOrderItemsCount: orderSummary?.order?.orderedOrderItems?.size()?:0,
                 shippedItemsCount: orderSummary?.order?.shippedOrderItems?.size()?:0,
                 receivedItemsCount: orderSummary?.order?.receivedOrderItems?.size()?:0,
-                total: "${orderSummary?.order?.total?.setScale(2, RoundingMode.HALF_UP)} ${orderSummary?.order?.currencyCode ?: defaultCurrencyCode}",
+                total: "${orderSummary?.order?.total ? orderSummary?.order?.total?.setScale(2, RoundingMode.HALF_UP) : 0} ${orderSummary?.order?.currencyCode ?: defaultCurrencyCode}",
```
kchelstowski, line 203: fixed bug that was not "closing" the JSON in the response. The problem was with orders with `total=0`, because `setScale` couldn't work here and the `total?` is only checking if it's not `null`, but 0 is not `null`, so it was proceeding.

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
```diff
@@ -83,7 +82,7 @@ class PurchaseOrderApiController {
 
     def statusOptions = {
         def options = OrderSummaryStatus.derivedStatuses()?.collect {
-            [ id: it.name(), value: it.name(), label: "${g.message(code: 'enum.OrderSummaryStatus.' + it.name())}" ]
+            [ id: it.name(), value: it.name(), label: "${g.message(code: 'enum.OrderSummaryStatus.' + it.name())}", variant: it.variant ]
```
kchelstowski, line 85: added `variant` property to `OrderSummaryStatus` to store variant of circle's color used in the table for statuses

src/groovy/org/pih/warehouse/order/OrderSummaryStatus.groovy
```diff
@@ -11,18 +11,22 @@ package org.pih.warehouse.order
...
+    PARTIALLY_SHIPPED('primary'),
+    SHIPPED('primary'),
+    PARTIALLY_RECEIVED('primary'),
+    RECEIVED('primary'),
+    PARTIALLY_INVOICED('primary'),
+    INVOICED('primary'),
+    COMPLETED('success'),
+    CANCELED('danger'),
+    REJECTED('danger')
+
```
kchelstowski, line 26: as commented above, added variants of circle's color for each status which will be the easiest to change in the future if added here in my opinion, so we do not have to provide any changes on the frontend side.

src/js/utils/form-utils.jsx
```diff
@@ -83,7 +84,7 @@ export const renderFormFields = ({
             :
             <FieldLabel />
         }
-        <div className="col-md-4 col-7 form-element-container">
+        <div className={`form-element-container ${!filterElement ? 'col-md-4 col-7' : 'flex-1 filter-element-container'}`}>
           {renderInput(input, attr)}
```
kchelstowski, line 88: changes which make the position of filter form fields work well without breaking already existing form modals

src/js/utils/option-utils.jsx
```diff
@@ -171,3 +171,20 @@ export const debounceLocationGroupsFetch = (waitTime, minSearchLength) =>
...
+            id: obj.id,
+            value: obj.id,
+            name: obj.name,
+            label: `${obj.name}`,
+          }
+        ));
+      }
+      return null;
+    });
+
```
kchelstowski, line 190: I saw we only had `debounceFetch` of organizations, so I added a new one, which is not debounced. I needed it for `Purchasing organization` fetch.

grails-app/conf/BootStrap.groovy
awalkowiak: This is no longer required (I added the `else 0.0` in total)

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+import PurchaseOrderStatus from 'components/purchaseOrder/PurchaseOrderStatus';
+import ActionDots from 'utils/ActionDots';
+import apiClient from 'utils/apiClient';
+import Translate, { translateWithDefaultMessage } from 'utils/Translate';
+
+import 'react-table/react-table.css';
+
+
+// Custom function to parse date to show like: May, 23, 2022
+const dateParser = (date) => {
```
awalkowiak, line 36: This is wrong approach. Add this parsing to the Bootstrap.groovy.

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+  ];
+  return `${monthNames[month]} ${day}, ${year}`;
+};
+
+
+const PurchaseOrderListTable = ({
+  filterParams, supportedActivities, highestRole, showTheSpinner, hideTheSpinner, translate,
+}) => {
+  // Temporary 'hard-coded' checking for role to display an action in dropdown or not
+  const hasMinimumRequiredRole = (role) => {
```
awalkowiak, line 52: Move this to the util function.

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+    // TODO: Figure out better way to check roles
+    const roles = ['Superuser', 'Admin', 'Manager', 'Assistant', 'Browser', 'Authenticated', 'Anonymous'];
+    return roles.indexOf(highestRole) <= roles.indexOf(role);
+  };
+
+  // Determine list of actions for each row individually
+  const findActions = (actionList, row) => {
```
awalkowiak, line 59: Seems like this also could be moved to some sort of `list utils`

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+      return true;
+    });
+  };
+
+  const [ordersData, setOrdersData] = useState([]);
+  const [loading, setLoading] = useState(true);
+  const [pages, setPages] = useState(-1);
+  const [totalPrice, setTotalPrice] = useState(0.0);
+  // Stored searching params for export case
+  const [currParams, setCurrParams] = useState({});
```
awalkowiak, line 87: Call it `currentParams`

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+      Header: 'Ordered On',
+      accessor: 'dateOrdered',
+      className: 'cell',
+      headerClassName: 'header text-align-left',
+      Cell: row => (
+        <span>{dateParser(row.original.dateOrdered)}</span>
+      ),
+      Footer: (
+        <span>
+          <Translate id="react.purchaseOrder.totalAmount.label" defaultMessage="Total amount" />: {totalPrice} USD
```
awalkowiak, line 262: Why it has hard coded `USD`? Add it in the DTO, taken from `grailsApplication.config.openboxes.locale.defaultCurrencyCode`

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+  ];
+
+
+  return (
+    <div className="purchase-order-list-list-section">
+      <div className="title-box d-flex justify-content-between align-items-center">
+        <span>
+          <Translate id="react.purchaseOrder.listOrders.label" defaultMessage="List Orders" />
+          &nbsp;
+          (<Translate id="react.purchaseOrder.totalAmount.label" defaultMessage="Total amount" />: {totalPrice} USD)
```
awalkowiak, line 276: Same as above

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -0,0 +1,385 @@
...
+        sortable
+        collapseOnSortingChange={false}
+        previousText={<i className="fa fa-chevron-left" aria-hidden="true" />}
+        nextText={<i className="fa fa-chevron-right" aria-hidden="true" />}
+        pageText=""
+        onFetchData={(state) => {
+          const offset = state.page > 0 ? (state.page) * state.pageSize : 0;
+          const sortingParams = state.sorted.length > 0 ?
+            {
+              sort: state.sorted[0].id,
```
awalkowiak, line 320: I think, that by default it should sort by `orderedOn` descending (or by OrderSummaryStatus if there would be a sortOrder defined there)

src/js/components/purchaseOrder/PurchaseOrderStatus.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const matchedStatus = statuses &&
+      statuses.length > 0 &&
+      statuses.find(stat => stat.id === statusProp);
+    if (matchedStatus && matchedStatus.variant) {
+      setCircle(matchedStatus.variant);
+    }
+  };
+  // Fetch all PO statuses
+  useEffect(() => {
+    apiClient.get('/openboxes/api/orderSummaryStatus')
```
awalkowiak, line 22: This should be fetched through actions and stored in the reducer.

src/js/utils/ActionDots.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+const ActionDots = ({ actions, id }) => (
+  <div className="btn-group">
+    <div
+      className="action-dots dropdown-toggle d-flex align-items-center justify-content-center"
+      data-toggle="dropdown"
+      aria-haspopup="true"
+      aria-expanded="false"
+    >
+        ...
```
awalkowiak, line 15: Use `RiMoreLine` or `RiMoreFill`

src/js/utils/option-utils.jsx
```diff
@@ -171,3 +171,20 @@ export const debounceLocationGroupsFetch = (waitTime, minSearchLength) =>
       callback([]);
     }
   }, waitTime);
+
+export const organizationsFetch = (roleTypes = ['ROLE_SUPPLIER'], active = false) =>
```
awalkowiak, line 175: Add action fetchSuppliers, and store it in the reducer. Why it has `active = false` by default?

awalkowiak (changes requested):

drodzewicz (approved):

src/js/utils/option-utils.jsx
kchelstowski: I believe when adding active column to party, we didn't want to create much regression and to set `active=true`, because that would make organizations not visible in many places where they should be visible. This conception is followed everywhere where fetching parties.

kchelstowski:
I realized that I added `fetchSuppliers` additionally, as I needed `fetchBuyers` more, but I think it should be anyway needed in some near future, so I decided not to delete it.

PR #3479, opened 2022-09-21
OBPIH-4733 Add new react-select styling for filters

awalkowiak:


drodzewicz (changes requested):
I would maybe create a separate style file for `FilterSelectField` and put all the inline styles and styles from `main.scss` into `FilterSelectField.scss`

drodzewicz (approved):

PR #3480, opened 2022-09-21
OBPIH-4686 Investigate why product availability issue requires intermittent reboots

jmiranda:
I broke this down into multiple commits in case we wanted to cherry-pick commits for a hotfix from these candidates. Given that the issue hasn't recurred in a week (or two?), we should test this on obdev for a few days before pushing a hotfix to obnavstage.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy
@@ -68,7 +68,9 @@ class PicklistItem implements Serializable {
     static transients = ['associatedLocation', 'associatedProducts', 'disableRefresh', 'pickable']
 
     String getAssociatedLocation() {
-        return requisitionItem ? requisitionItem?.requisition?.origin?.id : orderItem?.order?.origin?.id
+        return binLocation?.parentLocation?.id ?:
+                requisitionItem?.requisition?.origin?.id ?:
+                        orderItem?.order?.origin?.id
     }
 
     List getAssociatedProducts() {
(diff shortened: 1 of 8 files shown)


grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy
```diff
@@ -68,7 +68,9 @@ class PicklistItem implements Serializable {
     static transients = ['associatedLocation', 'associatedProducts', 'disableRefresh', 'pickable']
 
     String getAssociatedLocation() {
-        return requisitionItem ? requisitionItem?.requisition?.origin?.id : orderItem?.order?.origin?.id
+        return binLocation?.parentLocation?.id ?:
```
awalkowiak, line 71: @jmiranda What in case of default bin location? Is not it then null? I think the issue with refreshing for all locations will be still present for default bins.

awalkowiak (commented):
@jmiranda overall looks good to me, but what with the case when there is a default bin location?

grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy
jmiranda: There's nothing we can do except find a better way altogether. The idea here is that the associated location for the picklist item might be null, but the parent requisition item won't be null (unless we're deleting that too I guess).

If we see the null location case in other places we just have to use the same pattern (disable refresh through parent object, trigger manually at the end of the database transaction).

The one thing I don't like (but don't have the ability to deal with now) is that we're using the disable refresh / trigger manually pattern in seemingly random places. So within a single request we might execute that multiple times manually when once would have been enough.

PR #3486, opened 2022-09-22
OBPIH-4772 Create whole Invoice list page

kchelstowski:


grails-app/conf/UrlMappings.groovy
```diff
@@ -372,6 +372,16 @@ class UrlMappings {
...
+        "/api/invoices/getStatuses"(parseRequest: true) {
+            controller = { "invoiceApi" }
+            action = [GET: "statusOptions"]
+        }
+
+        "/api/invoices/getInvoiceTypeCodes"(parseRequest: true) {
+            controller = { "invoiceApi" }
+            action = [GET: "invoiceTypeCodes"]
+        }
+
```
kchelstowski, line 384: Endpoints for invoice statuses and for invoice type codes

src/css/main.scss
```diff
@@ -1097,3 +1097,112 @@ div.search-input:focus-within {
...
+
+        &:hover {
+          color: var(--red-ob) !important;
+        }
+      }
+    }
+  }
+}
+
+
```
kchelstowski, line 1208: I moved all the styling which were common for purchase order list and invoice list to one common, so we don't have to re-write the same styling for each page

src/groovy/org/pih/warehouse/invoice/InvoiceStatus.groovy
```diff
@@ -11,15 +11,19 @@ package org.pih.warehouse.invoice
...
+    PAID(40, 'primary') // Invoice has been paid
 
     int sortOrder
+    String variant
 
-    InvoiceStatus(int sortOrder) { [this.sortOrder = sortOrder] }
+    InvoiceStatus(int sortOrder, String variant) {
+        this.sortOrder = sortOrder
+        this.variant = variant
+    }
```
kchelstowski, line 26: as for PO statuses, I added variant to determine status circle on the frontend side

src/js/actions/index.js
```diff
@@ -457,6 +459,7 @@ export function fetchSuppliers(active = false) {
             type: FETCH_SUPPLIERS,
             payload: suppliers,
           });
+          return;
         }
```
kchelstowski, line 463: it missed return and the result was that it was an infinite loop

src/groovy/org/pih/warehouse/invoice/InvoiceStatus.groovy
```diff
@@ -11,15 +11,19 @@ package org.pih.warehouse.invoice
 
 enum InvoiceStatus {
 
-    PENDING(0),  // Drafted
-    INVOICED(10), // Invoice has been issued / received
-    SUBMITTED(20), // Invoice has been submitted for approval
-    POSTED(30),  // Invoice has been posted
-    PAID(40) // Invoice has been paid
+    PENDING(0, 'warning'),  // Drafted
```
awalkowiak, line 14: I think I don't like this. Could you add this as a StatusType enum at least? (and remember to use it on the OrderSummaryStatus too). And please add a comment in it to describe what it is for.

```
enum StatusType {
  PRIMARY,
  SUCCESS,
  WARNING,
  DANGER
}
```

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
```diff
@@ -82,7 +82,7 @@ class PurchaseOrderApiController {
 
     def statusOptions = {
         def options = OrderSummaryStatus.derivedStatuses()?.collect {
-            [ id: it.name(), value: it.name(), label: "${g.message(code: 'enum.OrderSummaryStatus.' + it.name())}", variant: it.variant ]
+            [ id: it.name(), value: it.name(), label: "${g.message(code: 'enum.OrderSummaryStatus.' + it.name())}", variant: "${g.message(code: 'enum.StatusType.' + it.variant)}"  ]
```
awalkowiak, line 85: This should not be translatable because it might not work properly on the frontend after other language translations are provided

awalkowiak (changes requested):

awalkowiak (approved):

PR #3489, opened 2022-09-23
OBPIH-4769 Add list endpoint in Product API

awalkowiak:


grails-app/conf/BootStrap.groovy
```diff
@@ -249,7 +249,20 @@ class BootStrap {
 
 
         JSON.registerObjectMarshaller(Product) { Product product ->
-            return product.toJson()
+            [
```
jmiranda, line 252: Why are we moving these out of the Product class? I am trying to move all of these into the domain classes.

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -24,6 +27,40 @@ class ProductApiController extends BaseDomainApiController {
...
+        params.unitOfMeasure = params.q
+
+        // If we specify a format=csv we want to download everything
+        if (params.format == 'csv') {
+            params.max = -1
+        }
+
+        def products = productService.getProducts(category, catalogs, tags, includeInactive, params)
+
+        if (params.format == 'csv') {
```
jmiranda, line 54: Do we need CSV handling on the API?

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -56,7 +93,7 @@ class ProductApiController extends BaseDomainApiController {
         render([data: data] as JSON)
     }
 
-    def list = {
+    def search = {
```
jmiranda, line 96: I think this might be a breaking change. The product API is already being used externally for things like the mobile app and other API clients so this might be a bad idea. In the future, please don't change existing endpoints unless we've had a discussion.

PR #3490, opened 2022-09-24
OBPIH-4773 create the page with all elements filters table for stock list

drodzewicz:


grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -393,8 +393,8 @@ class Requisition implements Comparable<Requisition>, Serializable {
             requestedBy: requestedBy?.name,
             createdBy: createdBy?.name,
             updatedBy: updatedBy?.name,
-            dateCreated: dateCreated,
-            lastUpdated: lastUpdated,
+            dateCreated: dateCreated?.format("MMM dd, yyyy"),
+            lastUpdated: lastUpdated?.format("MMM dd, yyyy"),
```
drodzewicz, line 397: is this is a good idea to set date format on backend or should I have formatted the date on frontend ?

src/js/components/form-elements/Button.jsx
```diff
@@ -5,32 +5,29 @@ import PropTypes from 'prop-types';
...
-    className={`${variant}-button d-flex justify-content-around align-items-center ${disabled && `${variant}-button-disabled`} ${variant === 'dropdown' && 'dropdown-toggle'}`}
-    disabled={disabled}
-    type={type}
-    onClick={onClickAction}
-    data-toggle={variant === 'dropdown' && 'dropdown'}
-    aria-haspopup={variant === 'dropdown' && 'true'}
-    aria-expanded={variant === 'dropdown' && 'false'}
-  >
-    {EndIcon ? (
+  label, defaultLabel, disabled, variant, type, onClick, EndIcon, isDropdown,
```
drodzewicz, line 8: Instead of having dropdown set by `variant` prop I added `isDropdown` prop for that and `variant` is responsible for setting button styling eg: **primary**, **transparent**, **secondary**

src/js/utils/list-utils.jsx
```diff
@@ -6,26 +6,32 @@ export const hasMinimumRequiredRole = (role, highestUserRole) => {
...
-  return filteredByActivityCode.filter((action) => {
+  // Filter by required user's role if provided
+  const filteredByMinimumRequiredRole = filteredByActivityCode.filter((action) => {
     if (action.minimumRequiredRole) {
-      return hasMinimumRequiredRole(action.minimumRequiredRole, highestUserRole);
+      return hasMinimumRequiredRole(action.minimumRequiredRole, highestRole);
     }
     return true;
   });
+  // Use custom filter callback
```
drodzewicz, line 31: filtering actions based on **supportedActivities** or **minimumRequiredUserRole** is a good but sometimes we need some other criteria so I added a custom callback that any condition can be applied

src/js/components/Filter/FilterForm.jsx
```diff
@@ -31,21 +33,18 @@ const FilterForm = ({
...
-  const [amountFilled, setAmountFilled] = useState(0);
+  let initialValues = Object.keys(filterFields)
+    .reduce((acc, key) => ({ ...acc, [`${key}`]: null }), {});
+  initialValues = { ...initialValues, ...defaultValues };
+
+  // Calculate which object's values are not empty
   const countFilled = (values) => {
-    // Calculate which object's values are not empty
-    setAmountFilled(Object.values(values).filter(value => !_.isEmpty(value)).length);
+    setAmountFilled(Object.values(values)
```
drodzewicz, line 42: the problem was that `_.isEmpety(...)` lodash function evaluates `boolean` values as `false`

src/js/utils/list-utils.jsx
```diff
@@ -6,26 +6,32 @@ export const hasMinimumRequiredRole = (role, highestUserRole) => {
 };
 
 
-export const findActions = (actionList, row, supportedActivities, highestUserRole) => {
-  // Firstly filter out by status if any provided
+export const findActions = (actionList, row, props) => {
+  const { supportedActivities = [], highestRole = [], customFilter } = props;
```
drodzewicz, line 10: I have made `supportedActivities`, `highestRole` and `customFilter` to be properties because we sometimes don't need to pass all of them.
So instead of passing `findActions(actions, row,  [], [], (action) => action.hasProperty)` 
we can do `findActions(actions, row,  { customFilter: (action) => action.hasProperty })`

src/js/components/purchaseOrder/PurchaseOrderStatus.jsx
```diff
@@ -30,17 +31,7 @@ const PurchaseOrderStatus = ({ status, allStatuses, fetchStatuses }) => {
...
-
-  return (
-    <div className="d-flex align-items-center justify-content-between">
-      <div className="d-flex justify-content-between align-items-center">
-        <span className={`${circle}-circle status-circle`} />
-        <span className="px-1">{statusToDisplay}</span>
-      </div>
-    </div>
-  );
+  return (<StatusIndicator status={status} variant={circle} />);
```
kchelstowski, line 34: maybe it's good idea to include this also in some short `.md` documentation if you have some time?

src/js/components/stock-list/StockListTable.jsx
```diff
@@ -0,0 +1,451 @@
...
+        setLoading(false);
+        setPages(Math.ceil(res.data.totalCount / state.pageSize));
+        setTotalData(res.data.totalCount);
+        setTableData(res.data.data);
+        // Store currently used params for export case
+        setCurrentParams(params);
+      });
+  };
+
+  // List of all actions for PO rows
```
kchelstowski, line 248: PO => stocklist

kchelstowski (changes requested):

src/js/components/Filter/FilterForm.jsx
```diff
@@ -31,21 +33,18 @@ const FilterForm = ({
...
-        ...acc,
-        [curr]: '',
-      };
-    }
-    return acc;
-  }, { ...defaultValues });
-  const [amountFilled, setAmountFilled] = useState(0);
+  let initialValues = Object.keys(filterFields)
+    .reduce((acc, key) => ({ ...acc, [`${key}`]: null }), {});
+  initialValues = { ...initialValues, ...defaultValues };
```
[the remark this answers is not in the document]
drodzewicz: It is a very simple code and basically works exactly the same as the original by initially setting `null` values to all of the filters and after overwriting them with `defaultValues` if the exist .
It's just in this case I saw the opportunity to make this code shorter without any `if(...)` statements.
But I see your point and I don't have any problem with bringing back the previous version.

src/js/components/stock-list/StockListTable.jsx
```diff
@@ -0,0 +1,451 @@
...
+      });
+  };
+
+  const customActionFilter = ({ isPublished }, row) => {
+    // skip actions that don't have isPublished property
+    if (isPublished === undefined) return true;
+    // negate XOR
+    // show actions that have same boolean value in row and in action
+    // eslint-disable-next-line no-bitwise
+    return !(row.original.isPublished ^ isPublished);
```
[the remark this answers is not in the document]
drodzewicz: Yes, this probably my "overengineering" mistake.
Looking at this after a good night sleep using `XOR` seems unnecessary

src/js/components/Filter/FilterForm.jsx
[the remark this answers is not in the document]
kchelstowski: my bad I haven't told you what the bug might be about - the problem is that you changed `''` to `null` - there is one case in PO list where I was checking if one of params was === `''` (couldn't do it any other way) so changing from `''` to `null` would break this case there.
We can agree that this `if` is for 0.0001% as the situation with two the same keys would probably never happen, so I would meet you halfway if you at least made the default `acc` as `{...defaultValues}`, because `.reduce` allows that and there is no point of using empty acc and having to use `let` to be able to assign the `defualtValues` then.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
kchelstowski: sure it is

grails-app/services/org/pih/warehouse/inventory/StocklistService.groovy
```diff
@@ -83,4 +91,140 @@ class StocklistService {
         }
         mailService.sendHtmlMailWithAttachment(to, subject, body, attachments)
     }
+
+    void clearStockListItems(String id) {
+        Requisition requisition = Requisition.get(id)
```
awalkowiak, line 96: Move this part to the stocklist controller and instead of throwing ObjectNotFoundException, you should return 404 status.

grails-app/services/org/pih/warehouse/inventory/StocklistService.groovy
```diff
@@ -83,4 +91,140 @@ class StocklistService {
...
+        if (!requisition) {
+            throw new ObjectNotFoundException(id, Requisition.class.toString())
+        }
+        requisitionService.clearRequisition(requisition)
+    }
+
+    void cloneStockList(String id) {
+        def requisition = Requisition.get(id)
```
awalkowiak, line 104: Same as above

grails-app/services/org/pih/warehouse/inventory/StocklistService.groovy
```diff
@@ -83,4 +91,140 @@ class StocklistService {
...
+        if (!requisition) {
+            throw new ObjectNotFoundException(id, Requisition.class.toString())
+        }
+        requisitionService.cloneRequisition(requisition)
+    }
+
+    void publishStockList(String id, Boolean publish) {
+        def requisition = Requisition.get(id)
```
awalkowiak, line 112: Same as above

grails-app/controllers/org/pih/warehouse/api/StocklistApiController.groovy
```diff
@@ -40,6 +54,16 @@ class StocklistApiController {
             throw new ObjectNotFoundException(params.id, Stocklist.class.toString())
         }
 
+        if(params.format == 'csv') {
```
awalkowiak, line 57: Why this was added to the read endpoint? Is this required?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -210,7 +210,29 @@ class RequisitionService {
      * @return
      */
     def getRequisitions(Requisition requisition, Map params) {
-        return getRequisitions(requisition, params, [], [])
+        List<Location> origins = [];
+        List<Location> destinations = [];
+
+        if (params.origin.getClass().isArray()) {
```
awalkowiak, line 216: Move it to the controller as:
```
def origins = params.origin ? Location.findAllByIdInList(params.list("origin")) : []
def destinations = params.destination ? Location.findAllByIdInList(params.list("destination")) : []
```
And then call the `getRequisitions` instead of `getAllRequisitionTemplates`:
`def requisitions = requisitionService.getRequisitions(requisition, params, origins, destinations)`

awalkowiak (changes requested):

awalkowiak (approved):

PR #3491, opened 2022-09-26
OBPIH-4772 Invoice list (fixes after QA)

kchelstowski:


drodzewicz (approved):

src/js/components/invoice/InvoiceListTable.jsx
```diff
@@ -130,16 +148,26 @@ const InvoiceListTable = ({
...
       className: 'text-left',
-      sortable: false,
     },
   ];
 
 
   const onFetchHandler = (state) => {
     const offset = state.page > 0 ? (state.page) * state.pageSize : 0;
+    const sortKeyParser = (key) => {
+      // Property is named invoiceTypeCode, but in marshaller was renamed to invoiceType
```
awalkowiak, line 158: Then rename it in the marshaller to the `invoiceTypeCode`

kchelstowski:
The problem with Search/Clear button if we wanted to refresh with the same filter values multiple times was that the state (filterParams) was not changed as after multiple submits we were providing the same values. 
Fix for that is to kind of "cheat" React and pass the same values, but in a object, that is different in terms of reference.

PR #3493, opened 2022-09-26
OBPIH-4712 DataTable component fixes

drodzewicz:
- created a table cell component that can receive tooltip and href props
- added tooltip to purchase order status cells
- fixed manual DataTable displaying wrong total data size


src/js/components/DataTable/TableCell.jsx
```diff
@@ -0,0 +1,49 @@
...
+  column, value, original, children, tooltip, tooltipLabel,
+}) => {
+  let cellElement = <div>{ children || value}</div>;
+
+  // interpolate string by inserting values with ":" at the beginning with values from the cell
+  if (column.link && typeof column.link === 'string') {
+    const href = column.link
+      .split('/')
+      .map(it => (it[0] === ':' ? original[`${it.substring(1)}`] : it))
+      .join('/');
```
kchelstowski, line 16: Isn't it a bit overthought? What about doing:
```
Cell: (row) => TableCell(row, {link: `/openboxes/order/show/${row.original.id}`}
```

so `TableCell` would get `row` and some `props` as params and inside props could be for example `link`?

src/js/components/invoice/InvoiceListTable.jsx
```diff
@@ -108,7 +109,8 @@ const InvoiceListTable = ({
       Header: 'Invoice Number',
       accessor: 'invoiceNumber',
       sortable: false,
-      Cell: row => (<a href={`/openboxes/invoice/show/${row.original.id}`}>{row.original.invoiceNumber}</a>),
+      link: '/openboxes/invoice/show/:id',
+      Cell: TableCell,
```
kchelstowski, line 113: comment about this is above

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -271,50 +312,55 @@ const PurchaseOrderListTable = ({
       className: 'active-circle',
       headerClassName: 'header',
       fixed: 'left',
-      maxWidth: 100,
-      Cell: row => (<PurchaseOrderStatus status={row.original.status} />),
+      width: 160,
+      tooltip: true,
+      Cell: row => (
+        <TableCell {...row} tooltipLabel={getStatusTooltip(row.original.status)}>
```
kchelstowski, line 318: Actually you could do the same with the `link` above, as you did here passing `tooltipLabel`

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -168,6 +168,47 @@ const PurchaseOrderListTable = ({
...
+        );
+      case 'COMPLETED':
+        return translate(
+          'react.purchaseOrder.status.completed.description.label',
+          'This order has been shipped and received. Related purchase invoices have been posted.',
+        );
+      default:
+        return status;
+    }
+  };
```
kchelstowski, line 211: what about keeping it somehow on backend, for example having another property for `OrderSummaryStatus` and inside `def statusOptions` in `PurchaseApiController` add another property behind `variant` called e.g. `tooltip`?
Let @awalkowiak decide what to do with this, it is just my preposition.

kchelstowski (changes requested):

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -168,6 +168,47 @@ const PurchaseOrderListTable = ({
     Alert.error(translate('react.default.featureNotSupported', 'This feature is not currently supported'));
   };
 
+  const getStatusTooltip = (status) => {
```
awalkowiak, line 171: Instead of doing this, just add messages like:
`react.enum.OrderSummaryStatus.PENDING.description=This Purchase Order is in progress and has not been finalized.`
and do:
`translate(``react.enum.OrderSummaryStatus.${status}.description``)`

awalkowiak (changes requested):

src/js/components/DataTable/TableCell.jsx
drodzewicz: I did it for convenience to not write the whole callback `(row) => ....` and pass just the component and  links or tooltips as attributes above.
Good point on the nested values like `origin.id` but I think it can be easily fixed by using lodash get function
like:
`
    const href = column.link
      .split('/')
      .map(it => (it[0] === ':' ? _.get(it.substring(1)) : it))
      .join('/');
`

awalkowiak (approved):

kchelstowski (approved):

PR #3495, opened 2022-09-26
OBPIH-4769 Improvements to the products list API

awalkowiak:


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -271,7 +271,9 @@ class ProductService {
      * @return
      */
     List<Product> getProducts(Category category, List<Tag> tags, boolean includeInactive, Map params) {
-        return getProducts(category, [], tags, includeInactive, params)
+        def categories = []
+        categories << category
```
jmiranda, line 275: I think you can shorten this to 
```
def categories = [category]
```

jmiranda (commented):
LGTM

grails-app/services/org/pih/warehouse/product/ProductService.groovy
awalkowiak: Yup, you're right

PR #3497, opened 2022-09-27
OBPIH-4770 Create new product list

kchelstowski:
I also found in the mean time some minor issues on PO list and changed some state names on invoiceList which were copied from PO list and not changed by mistake.


src/js/components/products/ProductsList.jsx
```diff
@@ -0,0 +1,43 @@
...
+import ProductsListHeader from 'components/products/ProductsListHeader';
+import ProductsListTable from 'components/products/ProductsListTable';
+
+
+const ProductsList = (props) => {
+  // Filter params are stored here, to be able to pass them to table component
+  const [filterParams, setFilterParams] = useState({});
+
+  useEffect(() => {
+    props.fetchTranslations(props.locale, 'purchaseOrder');
```
drodzewicz, line 17: wrong translation fetched
change to
`props.fetchTranslations(props.locale, 'productsList');`

src/js/components/products/ProductsListFilters.jsx
```diff
@@ -0,0 +1,115 @@
...
+    getDynamicAttr: ({ categories }) => ({
+      options: categories,
+    }),
+  },
+  includeCategoryChildren: {
+    type: CheckboxField,
+    label: 'react.productsList.includeSubcategories.label',
+    defaultMessage: 'Include all products in all subcategories',
+    attributes: {
+      filterElement: true,
```
drodzewicz, line 30: what does the attribute `filterElement: true` do for  `CheckboxField` ?

drodzewicz (changes requested):

src/js/components/products/ProductsListFilters.jsx
kchelstowski: it is added to filter form elements not to have to play around with position styles included in `form-utils`, so it is just to position filter elements, nothing functional.

drodzewicz: that's right I remember now. The naming implies that it has something to do with filtering, kind of misleading

drodzewicz (approved):

awalkowiak (approved):

PR #3498, opened 2022-09-27
OBPIH-4749 Improvements to the stock movement list API

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -24,25 +26,59 @@ import org.pih.warehouse.product.Product
...
+        if (params.format == 'csv') {
+            params.max = null
+            params.offset = null
+        }
+
+        def stockMovements = stockMovementService.getStockMovements(stockMovement, params)
 
-        def stockMovements = stockMovementService.getStockMovements(stockMovement, [max: max, offset: offset, orderBy: orderBy ])
+        if (params.format == 'csv' && stockMovements) {
+            def csv = getStockMovementsCsv(stockMovements)
```
jmiranda, line 77: I don't think we should be doing CSV from the API. Why not do this in the stock movement controller?

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -496,4 +532,195 @@ class StockMovementApiController {
...
+            def csv = getPendingRequisitionItemsCsv(pendingRequisitionItems)
+            def date = new Date()
+            response.contentType = "text/csv"
+            response.setHeader("Content-disposition", "attachment; filename=\"PendingShipmentItems-${new Date().format("yyyyMMdd-hhmmss")}.csv\"")
+            render(contentType: "text/csv", text: csv.out.toString())
+            return
+        } else {
+            def message = "No pending requisition items found"
+            response.status = 404
+            render([errorMessage: message] as JSON)
```
jmiranda, line 626: If the client is expecting CSV, why are we returning errors as JSON?

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -496,4 +532,195 @@ class StockMovementApiController {
...
+                shipmentItem.shipmentName,
+                shipmentItem.origin,
+                shipmentItem.destination,
+            )
+        }
+
+        return csv
+    }
+
+    def getPendingRequisitionItemsCsv(List pendingRequisitionItems) {
```
jmiranda, line 662: Here, these are just requisition items. No need for the pending descriptor. What if I end up decided to pass a list of picked or issued requisition items. It doesn't seem to be doing anything special because of the "pending" status so it's probably not necessary.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -496,4 +532,195 @@ class StockMovementApiController {
...
+            return
+        } else {
+            def message = "No pending requisition items found"
+            response.status = 404
+            render([errorMessage: message] as JSON)
+            return
+        }
+    }
+
+    def getShipmentItemsCsv(List shipmentItems) {
```
jmiranda, line 631: In general, I don't like these methods here in the controller. They're better off in a service. I also think all of these could be more generalizable. We could convert each of these objects into a map and pass the map to one renderCsv method that does something very simple like this.

```
def keys = data[0].keySet() 
csv.printRecord(keys)
data.each { row ->
    csv.printRecord(row.values)
}
```

jmiranda: And then all you need to do is convert the shipment or requisition item or whatever into a data map.
```
def data = shipments.collect { 
   return [
            "Code" : shipmentItem.productCode,
            "Product Name": shipmentItem.productName,
            "Quantity Incoming": shipmentItem.quantity,
            "Expected Shipping Date": shipmentItem.expectedShippingDate,
            "Expected Delivery Date": shipmentItem.expectedDeliveryDate,
            "Shipment Number": shipmentItem.shipmentNumber,
            "Shipment Name": shipmentItem.shipmentName,
            "Origin": shipmentItem.origin,
            "Destination": shipmentItem.destination,
  ]
}
```

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -496,4 +532,195 @@ class StockMovementApiController {
...
+        shipmentItems.each { shipmentItem ->
+            csv.printRecord(
+                shipmentItem.productCode,
+                shipmentItem.productName,
+                shipmentItem.quantity,
+                shipmentItem.expectedShippingDate,
+                shipmentItem.expectedDeliveryDate,
+                shipmentItem.shipmentNumber,
+                shipmentItem.shipmentName,
+                shipmentItem.origin,
```
jmiranda, line 654: I'm guessing the toString() is return the name, but you should be explicit here.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -496,4 +532,195 @@ class StockMovementApiController {
...
+            "Origin",
+            "Destination",
+        )
+
+        shipmentItems.each { shipmentItem ->
+            csv.printRecord(
+                shipmentItem.productCode,
+                shipmentItem.productName,
+                shipmentItem.quantity,
+                shipmentItem.expectedShippingDate,
```
jmiranda, line 650: Is there a preferred date format?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -193,7 +192,7 @@ class StockMovementController {
         def stockMovements
 
         try {
-            stockMovements = stockMovementService.getStockMovements(stockMovement, [max: max, offset: offset, createdAfter: createdAfter, createdBefore: createdBefore])
+            stockMovements = stockMovementService.getStockMovements(stockMovement, params)
```
jmiranda, line 195: Doesn't look like createdAfter and createdBefore are added back to the params map above.

jmiranda (commented):
There may be regressions added by this PR so test more than just the new list UIs.

PR #3499, opened 2022-09-28
OBS-1269 RequisitionItem.undoChanges() should remove dangling references from ShipmentItem

mdpearson:
This PR prevents the error reported in OBS-1269 where certain undo commands will crash on a requisition that has already been shipped. The cause is a foreign key constraint violation. This patch removes the requisitionItem(s) that the undo command will delete from any ShipmentItem object that points to them.

Since `ShipmentItem.requisitionItem` is explicitly nullable, this might be all we need? But I'm a little suspicious -- can this patch leave ShipmentItem objects in an inconsistent state? Is more recovery/undo logic needed here?

### DIFF
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -21,6 +21,7 @@ import org.pih.warehouse.product.Category
 import org.pih.warehouse.product.Product
 import org.pih.warehouse.product.ProductGroup
 import org.pih.warehouse.product.ProductPackage
+import org.pih.warehouse.shipping.ShipmentItem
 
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
```diff
@@ -680,7 +690,7 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
         }
     }
 
-    def getShipmentItems() {
+    Collection<ShipmentItem> getShipmentItems() {
```
jmiranda, line 693: I know there's a temptation to make these changes (I do it as well) but let's try to avoid doing things like this just so we don't have to chase regressions for things that weren't a problem. At some point I'm willing to go back through and make wholesale changes but let's leave working code alone.

jmiranda (commented):
Let's close this PR.

PR #3504, opened 2022-09-28
OBGM-353 warfile fixes and other improvements

mdpearson:
This is the code that's running on obnavtest1.

- resolve Gradle build warnings (OBGM-349)
- fix GSP/asset-pipeline conflicts (OBGM-353)
- prevent TransactionRequiredException on Dashboard load (OBGM-355)
- set up logging (OBGM-356)
- run hibernate migrations successfully from within Tomcat
- configure connection pooling
- remove stale Hibernate 3 references in favor of 5.2
- bump Grails to 3.3.16 and update other dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -297,6 +297,7 @@ springBoot {
 
 war {
     archiveName = 'openboxes.war'
+    zip64 = true
 }
 
 dependencies {
--- a/build.gradle
+++ b/build.gradle
@@ -202,7 +202,7 @@ configurations {
                 }
 
                 if (details.requested.group == 'org.springframework') {
-                    if (VersionNumber.parse(details.requested.version).major == 4) {
+                    if (details.requested.name != 'springloaded') {
                         details.because 'enforce consistent Spring versioning'
                         details.useVersion(springframeworkVersion)
                     }
--- a/build.gradle
+++ b/build.gradle
@@ -155,13 +155,6 @@ configurations {
                     details.useVersion(poiVersion)
                 }
 
-                if (details.requested.group.startsWith('org.apache.tomcat')) {
-                    if (details.requested.name != 'tomcat-embed-logging-log4j') {
-                        details.because 'enforce consistent Tomcat versioning'
-                        details.useVersion(tomcatVersion)
-                    }
-                }
-
                 if (details.requested.group == 'org.ow2.asm') {
                     details.because 'enforce a Java-8 compatible ASM release'
                     details.useVersion(asmVersion)
@@ -367,6 +360,16 @@ dependencies {
      */
     testCompile 'io.micronaut:micronaut-http-client:1.0.0.RC3'
 
+    /*
+     * Tomcat provides these libraries out of the box.
+     * https://tomcat.apache.org/tomcat-8.5-doc/RELEASE-NOTES.txt
+     */
+    provided 'javax.annotation:javax.annotation-api:1.3.2'
+    provided 'javax.el:javax.el-api:2.2.5'
+    provided "javax.servlet:javax.servlet-api:${servletVersion}"
+    provided "javax.servlet.jsp:javax.servlet.jsp-api:${jspVersion}"
+    provided "javax.websocket:javax.websocket-api:${websocketVersion}"
+
     assets 'junit:junit:4.13.2'
     testCompile 'junit:junit:4.13.2'
 
@@ -394,6 +397,11 @@ dependencies {
      * TODO do we need to specify an explicit version here?
      */
     runtime 'org.apache.tomcat:tomcat-jdbc'
+
+    // this library is provided by tomcat
+    provided "org.apache.tomcat:tomcat-juli:${minimumTomcatVersion}"
+
+    provided "org.apache.tomcat:tomcat-servlet-api:${minimumTomcatVersion}"
 
     // contents of this module are included in modern tomcat distributions
     provided 'org.apache.tomcat.embed:tomcat-embed-logging-log4j:8.5.2'
@@ -596,10 +604,15 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-logging'
 
     /*
-     * Prevent 'Unable to start EmbeddedWebApplicationContext
-     * due to missing EmbeddedServletContainerFactory bean.'
+     * Use `provided` per Grails 3's deployment guide, see
+     * http://docs.grails.org/3.3.16/guide/deployment.html
+     *
+     * We don't support application servers other than Tomcat.
+     *
+     * This ticket has a good discussion of `provided` gotchas.
+     * https://github.com/grails/grails-core/issues/10483
      */
-    compile 'org.springframework.boot:spring-boot-starter-web'
+    provided 'org.springframework.boot:spring-boot-starter-tomcat'
 
     // security patch
     compile 'org.yaml:snakeyaml:1.30'
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,6 @@ buildscript {
         classpath "com.bertramlabs.plugins:asset-pipeline-gradle:${assetPipelineVersion}"
         classpath "org.grails:grails-gradle-plugin:${grailsVersion}"
         classpath "org.grails.plugins:database-migration:${databaseMigrationVersion}"
-        classpath "org.grails.plugins:hibernate5:${gormVersion-'.RELEASE'}"
         classpath 'org.grails.plugins:views-gradle:1.3.0'
         classpath 'org.owasp:dependency-check-gradle:7.1.1'
     }
@@ -488,8 +487,8 @@ dependencies {
     // enable `import org.grails.plugins.web.taglib.*`
     compile 'org.grails.plugins:gsp'
 
-    // enable `import grails.orm, org.hibernate, org.springframework.orm`
-    compile "org.grails.plugins:hibernate5:${gormVersion}"
+    // prevent `No GORM implementations configured. Ensure GORM has been initialized correctly`
+    compile "org.grails.plugins:hibernate5:${gormVersion-'.RELEASE'}"
 
     /*
      * FIXME Configure this properly.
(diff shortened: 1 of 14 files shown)


grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -198,9 +198,16 @@ class DashboardController {
...
+
+                /*
+                 * Band-aid for OBGM-355, to prevent the following exception at dashboard launch:
+                 * javax.persistence.TransactionRequiredException: no transaction is in progress
+                 */
+                user.withTransaction {
+                    user.lastLoginDate = new Date()
+                    user.save(flush: true)
+                    session.user = user
+                }
```
mdpearson, line 210: @jmiranda @awalkowiak I wanted to highlight this band-aid -- let me know if there's a more performant way to avoid `TransactionRequiredException` than by wrapping the save call in a `withTransaction` block

grails-app/conf/application.yml
```diff
@@ -183,43 +190,65 @@ dataSource:
...
-            url: jdbc:mysql://localhost:3306/openboxes?serverTimezone=UTC&useSSL=false&createDatabaseIfNotExist=true
+            dumpQueriesOnException: true
+            formatSql: true
+            includeInnodbStatusInDeadlockExceptions: true
+            logAbandoned: true
+            logSql: false
+            logSlowQueries: true
+            username: "openboxes"
+            password: "openboxes"
+            url: jdbc:mysql://localhost:3306/openboxes3?serverTimezone=UTC&useSSL=false
```
mdpearson, line 230: I renamed the default database from `openboxes` to `openboxes3` to prevent accidental db mangling when switching between branches

grails-app/conf/application.yml
```diff
@@ -183,43 +190,65 @@ dataSource:
...
         maxWait: 10000
-        maxAge: 14400
+        maxAge: 10 * 60000
         timeBetweenEvictionRunsMillis: 5000
         minEvictableIdleTimeMillis: 60000
         validationQuery: SELECT 1
         validationQueryTimeout: 3
         validationInterval: 15000
-        testOnBorrow: true
+        testOnBorrow: false
```
mdpearson, line 212: testing a connection when we pull it from the pool adds a bit of latency; and since we ping unused connections every 15 seconds (testWhileIdle/validationInterval) the additional ping on checkout is unnecessary.

grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -467,43 +470,62 @@ class BootStrap {
...
 
-            /*
-             * FIXME This patch is unlikely to work in production. OBGM-333
-             *
-             * Previously, we used org.liquibase.grails.GrailsFileOpener here
-             * (from a plugin, https://github.com/liquibase/grails-liquibase/,
-             * that ceased development at Grails 1.3.6).
-             *
-             * The path manipulation here is probably dependent on a local install.
-             */
```
mdpearson, line 478: most of the changes here aren't strictly new, but rather pull code from a previous rebase

grails-app/migrations/changelog.groovy
```diff
@@ -6,8 +6,4 @@ databaseChangeLog = {
     //  - Add a new directory (e.g. 0.10.x) for new changelogs
     //  - Replace this changelog with the new version (e.g. 0.10.x/changelog.xml)
     include file: '0.9.x/changelog.xml'
-
-    // Recreate all views AFTER all changesets have been applied (just in case)
-    include file: 'views/changelog.xml'
-
```
mdpearson, line 12: I hoisted this to BootStrap.groovy, since it destroys views before doing any migrations, it seemed more clear to have it build them back after doing all migrations.

grails-app/conf/application.yml
jmiranda: I'd like to figure out a better solution before we merge this branch back to develop.

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -198,9 +198,16 @@ class DashboardController {
...
                 //userInstance.rememberLastLocation = Boolean.valueOf(params.rememberLastLocation)
-                user.lastLoginDate = new Date()
-                user.save(flush: true)
-                session.user = user
+
+                /*
+                 * Band-aid for OBGM-355, to prevent the following exception at dashboard launch:
+                 * javax.persistence.TransactionRequiredException: no transaction is in progress
+                 */
+                user.withTransaction {
```
jmiranda, line 206: This doesn't seem right. First of all, unless I;m I think it should be `User.withTransaction { }`. Secondly, we just need to annotate the method to be @Transactional.

grails-app/migrations/changelog.groovy
jmiranda: I don't want to run migrations programmatically if we don't have to. I would prefer if the drop / create views were done in the changelog.xml.

jmiranda (changes requested):
Looks good. I don't think I've made any blocking changes, but check through my comments, respond if necessary and if there's anything you think should change either make the change or create a ticket.

grails-app/conf/application.yml
```diff
@@ -183,43 +190,65 @@ dataSource:
...
         jmxEnabled: true
-        initialSize: 10
-        maxActive: 100
+        initialSize: 5
+        maxActive: 50
         minIdle: 5
         maxIdle: 25
         maxWait: 10000
-        maxAge: 14400
+        maxAge: 10 * 60000
```
jmiranda, line 206: Can you provide justification for these changes?

grails-app/conf/application.yml
jmiranda: I can live with that, but I'd probably rather have testOnBorrow over testWhileIdle. We can leave it as-is for now and come back to this if we encounter any issues related to the connection pool. This becomes more of a problem when the connection pool is on a separate server than the database and the cloud hosting provider is pretty cut throat when it comes to connections.

grails-app/conf/application.yml
```diff
@@ -183,43 +190,65 @@ dataSource:
...
         validationQueryTimeout: 3
         validationInterval: 15000
-        testOnBorrow: true
+        testOnBorrow: false
         testWhileIdle: true
         testOnReturn: false
-        jdbcInterceptors: ConnectionState
+        # https://tomcat.apache.org/tomcat-8.5-doc/jdbc-pool.html#JDBC_interceptors
+        jdbcInterceptors: "ConnectionState;StatementCache(max=200)"
         defaultTransactionIsolation: java.sql.Connection.TRANSACTION_READ_COMMITTED
```
jmiranda, line 217: I know this wasn't a change you made, but I think we might want to set this to the default MySQL isolation level.

> The default isolation level is [REPEATABLE READ](https://dev.mysql.com/doc/refman/5.6/en/innodb-transaction-isolation-levels.html#isolevel_repeatable-read). Other permitted values are [READ COMMITTED](https://dev.mysql.com/doc/refman/5.6/en/innodb-transaction-isolation-levels.html#isolevel_read-committed), [READ UNCOMMITTED](https://dev.mysql.com/doc/refman/5.6/en/innodb-transaction-isolation-levels.html#isolevel_read-uncommitted), and [SERIALIZABLE](https://dev.mysql.com/doc/refman/5.6/en/innodb-transaction-isolation-levels.html#isolevel_serializable). For information about these isolation levels, see [Section 14.7.2.1, “Transaction Isolation Levels”](https://dev.mysql.com/doc/refman/5.6/en/innodb-transaction-isolation-levels.html).

build.gradle
```diff
@@ -92,14 +96,49 @@ bootRun {
...
 
         resolutionStrategy {
+            /*
+             * Fail-fast if Gradle pulls in multiple versions of one
+             * dependency, which can cause deployment problems. Note,
+             * this won't differently-named modules that export the
+             * same API, which also can break the build. Fortunately,
+             * API collisions are less common than version collisions.
+             */
+            failOnVersionConflict()
```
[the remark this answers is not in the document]
mdpearson: If we run into any "works for me" issues I'd be happy to revisit, but until we see one in the flesh, I think this command will be a really helpful step towards a consistent and repeatable build.

build.gradle
```diff
@@ -563,30 +622,38 @@ dependencies {
...
+     * Use `provided` per Grails 3's deployment guide, see
+     * http://docs.grails.org/3.3.16/guide/deployment.html
+     *
+     * We don't support application servers other than Tomcat.
+     *
+     * This ticket has a good discussion of `provided` gotchas.
+     * https://github.com/grails/grails-core/issues/10483
      */
-    compile 'org.springframework.boot:spring-boot-starter-web'
+    provided 'org.springframework.boot:spring-boot-starter-tomcat'
```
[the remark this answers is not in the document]
mdpearson: My understanding is that `provided` declares that a dependency should be compiled against, but not included in the war file. All of the dependencies inside this dependency come "for free" when running on Tomcat (see https://tomcat.apache.org/tomcat-8.5-doc/RELEASE-NOTES.txt). Grails recommends using "provided" here if deploying to a different tomcat release than what you compile against, which we do (Spring boot includes tomcat 8.5.43, obnavtest1 uses Tomcat 9.0.16). (see  http://docs.grails.org/3.3.16/guide/deployment.html)

Using provided here prevents us from including jars that will be present on the class path anyway when deployed to a tomcat server, works correctly when using `grails run-app` and even starts up on jetty, because it provides the same APIs. So I'm almost certain this is the correct declaration here.

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
mdpearson: You're right on both counts, I think -- will fix

grails-app/conf/application.yml
mdpearson: `maxAge` is a few hundred milliseconds less than `validationInterval`, which, I believe, means that the db pool will never send a heartbeat ping to keep a connection alive. Instead, 4 times per minute, all unused db connections are destroyed and replaced with new ones. This seems like unnecessary churn, so I changed this to the value recommended by the Grails documentation. But, it's not strictly needed to get a clean deploy -- I think I'll pull all these config changes to a separate PR where we can discuss at leisure.

grails-app/migrations/changelog.groovy
jmiranda: This isn't resolved, but we can revisit this later so we can get this pushed to obnavtest1.

jmiranda (approved):

PR #3506, opened 2022-09-28
OBPIH-4751 Fix deleting Stock Movements endpoint

awalkowiak:
- plus fix filtering by dateCreated for inbounds
- plus fix OutboundStockMovementListItem toJson mapper


jmiranda (commented):
Ugh. Seems like an arduous task for something that should be relatively simple. Should we just rely on StockMovementService to return it for all three cases. Or continue to use 

```
stockMovementService.deleteStockMovement()
```
and hide the ugly details underneath.

I don't know.

PR #3511, opened 2022-09-30
OBPIH-4921 Improve list pages/filter component not to make unnecessar…

kchelstowski:
…y fetches of data

I've applied the changes to every existing new list page

### DIFF
--- a/src/js/components/Filter/FilterForm.jsx
+++ b/src/js/components/Filter/FilterForm.jsx
@@ -21,6 +21,7 @@ const FilterForm = ({
   defaultValues,
   allowEmptySubmit,
   hidden,
+  onClear,
 }) => {
   const [amountFilled, setAmountFilled] = useState(0);
   const [filtersHidden, setFiltersHidden] = useState(hidden);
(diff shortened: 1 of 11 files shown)


src/js/components/Filter/FilterForm.jsx
```diff
@@ -33,11 +32,10 @@ const FilterForm = ({
...
   };
-  // Create initialValues from filterFields as empty values
-  const initialValues = Object.keys(filterFields).reduce((acc, key) => {
-    if (!acc[key]) return { ...acc, [key]: '' };
-    return acc;
-  }, { ...defaultValues });
+
+  useEffect(() => {
+    updateFilterParams(defaultValues);
+  }, [defaultValues]);
```
awalkowiak, line 38: First I wrote that:
> I don't like the naming here. You have `defaultValues`, which I think should not change because they are `default` and you expect that it will change.

but then I read the rest of the files. Please add the comment on this `useEffect` that `// Default values can change based on the current location`

drodzewicz (approved):

PR #3517, opened 2022-10-03
OBPIH-4872 Fix fade class opacity styling bug

drodzewicz:
After upgrading `bootstrap` version - elements with class `.fade` additionally have `opacity: 0` applied to them.
The reason for that is that new version of bootstrap has additional styling for class `.fade` where:
```
.fade:not(.show) {
    opacity: 0;
}
```
so instead of adding in every instance where we use class `.fade` additional class `.show` I just overwritten the styling for `.fade:not(.show)`


drodzewicz:
additionally I have fixed success message for when submitting new request

awalkowiak (approved):

PR #3523, opened 2022-10-04
OBPIH-4868 menus for non admin users

drodzewicz:
Additionally I have fixed many warning that cluttered the console for Menu component proptypes

### DIFF
--- a/grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
+++ b/grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
@@ -50,6 +50,10 @@ class MegamenuService {
     List buildAndTranslateSubsections(List subsections, User user, Location location) {
         def builtSubsections = []
         subsections.each {
+            def minRole = it.minimumRequiredRole
+            if (minRole && !userService.isUserInRole(user, minRole)) {
+                return
+            }
             def roles = it.requiredRole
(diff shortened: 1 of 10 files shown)


grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
```diff
@@ -50,6 +50,10 @@ class MegamenuService {
     List buildAndTranslateSubsections(List subsections, User user, Location location) {
         def builtSubsections = []
         subsections.each {
+            def minRole = it.minimumRequiredRole
+            if (minRole && !userService.isUserInRole(user, minRole)) {
+                return
+            }
             def roles = it.requiredRole
             if (roles && !userService.hasAnyRoles(user, roles)) {
```
awalkowiak, line 58: Is the `requiredRole` still used in the config?

drodzewicz: it is used only on invoicing sections, where user is required to have `ROLE_INVOICE`

```
        invoicing {
            enabled = true
            requiredRole = [RoleType.ROLE_INVOICE]
            label = "react.invoicing.label"
            defaultLabel = "Invoicing"
            subsections = [ ... ]
        }
```

awalkowiak (approved):

PR #3524, opened 2022-10-04
OBPIH-4702 Add new ProductAssociationTypeCodes values

awalkowiak:


jmiranda (commented):
LGTM. The names of the newly added association types aren't sitting well with me but I can't come up with anything better at the moment.

PR #3526, opened 2022-10-04
OBPIH-4921 Improve list pages/filter component not to make unnecessar…

kchelstowski:
…y fetches of data (fixes after QA)

It has not been that easy straight forward as it was supposed to be, as it was not working propely on obnav and I had hard time to reproduce the issue locally, but it turned out not to be working (not always) when freshly running the project and having clean local storage. Many problems combined - the redux persist is not always working - it does not always "remember" the values - in this case it hardly ever remembers `currentLocation` when switching between React tabs and on the other hand if it does remember, it fetches the session data including `currentLocation` again, which makes the filters run again multiple times (multiple times if initially the redux persist REMEMBERED the value, so the fetch of data on list pages could continue, because currentLocation exists), but if it gets "fresh" (the same) `currentLocation`, it makes the fetch work again, so I had to come up with some solution to be able to "recognize" if:
1. redux persist remembers the currentLocation
2. To be able to determine if we initially fetched anything (so we do not fetch two times if redux persist remembered the currentLocatoin so the fetch could continue without "waiting" for another fetch of session info (FETCH_SESSION_INFO).
3. To be able to store the previous value of currentLocation, so if it is the same as previous one, it means we don't have to fetch data again

The code, especially those if's might seem complicated, but it was the only way to prevent each of those scenarios (you can just imagine the complexity of it, if anytime you run the project again, the redux persist gave different results of either storing the session info propely or not).

I know it's not possible to fetch the session data only once when running the project if we don't have 100% React , because in the "meantime" for example our user's role might change and we should be able to refresh that data ASAP, but I think we should consider dividing the session stuff which can't change "in the meantime" without your interference (and `currentLocation` might be one of many), so we can avoid those complicated solutions in some places.

### DIFF
--- a/src/js/MainRouter.jsx
+++ b/src/js/MainRouter.jsx
@@ -19,18 +19,17 @@ const TRANSLATION_PREFIXES = ['default', 'dashboard', 'combinedShipments', 'prod
 
 class MainRouter extends React.Component {
   componentDidMount() {
-    this.props.fetchSessionInfo().then(() => {
-      this.props.initialize({
-        languages: this.props.supportedLocales,
-        options: {
-          renderToStaticMarkup,
-          onMissingTranslation,
-        },
-      });
-      this.props.setActiveLanguage(this.props.locale);
-      this.props.fetchMenuConfig();
-      TRANSLATION_PREFIXES.forEach(prefix => this.props.fetchTranslations('', prefix));
+    this.props.fetchSessionInfo();
+    this.props.initialize({
+      languages: this.props.supportedLocales,
+      options: {
+        renderToStaticMarkup,
+        onMissingTranslation,
+      },
     });
+    this.props.setActiveLanguage(this.props.locale);
+    this.props.fetchMenuConfig();
+    TRANSLATION_PREFIXES.forEach(prefix => this.props.fetchTranslations('', prefix));
   }
 
   componentWillReceiveProps(nextProps) {
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -105,11 +105,14 @@ export function fetchUsers() {
 
 export function fetchSessionInfo() {
   const url = '/openboxes/api/getAppContext';
-  const request = apiClient.get(url);
-
-  return {
-    type: FETCH_SESSION_INFO,
-    payload: request,
+  return (dispatch) => {
+    apiClient.get(url)
+      .then((res) => {
+        dispatch({
+          type: FETCH_SESSION_INFO,
+          payload: res,
+        });
+      });
   };
 }
 
--- a/src/js/components/invoice/InvoiceListFilters.jsx
+++ b/src/js/components/invoice/InvoiceListFilters.jsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 
+import usePrevious from 'hooks/usePrevious';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
@@ -109,21 +110,29 @@ const InvoiceListFilters = ({
   fetchTypeCodes,
 }) => {
   const [defaultValues, setDefaultValues] = useState({});
+  const prevLocation = usePrevious(currentLocation);
+  const [initiallyFetched, setInitiallyFetched] = useState(false);
 
   useEffect(() => {
-    const initialEmptyValues = Object.keys(filterFields).reduce((acc, key) => {
-      if (!acc[key]) return { ...acc, [key]: '' };
-      return acc;
-    }, {});
-    setDefaultValues({
-      ...initialEmptyValues,
-      buyerOrganization: {
-        id: currentLocation?.organization?.id,
-        value: currentLocation?.organization?.id,
-        name: currentLocation?.organization?.name,
-        label: currentLocation?.organization?.name,
-      },
-    });
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
+    if ((!initiallyFetched && currentLocation?.id) || (prevLocation && prevLocation.id !== currentLocation?.id)) {
+      const initialEmptyValues = Object.keys(filterFields).reduce((acc, key) => {
+        if (!acc[key]) return { ...acc, [key]: '' };
+        return acc;
+      }, {});
+      setDefaultValues({
+        ...initialEmptyValues,
+        buyerOrganization: {
+          id: currentLocation?.organization?.id,
+          value: currentLocation?.organization?.id,
+          name: currentLocation?.organization?.name,
+          label: currentLocation?.organization?.name,
+        },
+      });
+      setInitiallyFetched(true);
+    }
   }, [currentLocation]);
 
   useEffect(() => {
--- a/src/js/components/products/ProductsListFilters.jsx
+++ b/src/js/components/products/ProductsListFilters.jsx
@@ -1,6 +1,8 @@
 import React, { useEffect, useState } from 'react';
 
+import usePrevious from 'hooks/usePrevious';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 
 import FilterForm from 'components/Filter/FilterForm';
 import CheckboxField from 'components/form-elements/CheckboxField';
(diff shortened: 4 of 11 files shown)


jmiranda:
@awalkowiak @kchelstowski Let's discuss this on the tech huddle tomorrow.

src/js/components/invoice/InvoiceListFilters.jsx
```diff
@@ -109,21 +110,29 @@ const InvoiceListFilters = ({
...
-        id: currentLocation?.organization?.id,
-        value: currentLocation?.organization?.id,
-        name: currentLocation?.organization?.name,
-        label: currentLocation?.organization?.name,
-      },
-    });
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
+    if ((!initiallyFetched && currentLocation?.id) || (prevLocation && prevLocation.id !== currentLocation?.id)) {
```
kchelstowski, line 120: it had to be done like this because: 
you might have a situation when you have currentLocation in store and go to another list page -
1.  if we left only 
```
if (currentLocation.id)
```
we would refetch twice - once having already stored location and second time when the getAppContext made a refetch.
2. if we left only
```
if (prevLocation.id !== currentLocation.id) 
```
we wouldn't make a fetch if we have currentLocation already stored, because the `prevLocation` would be the `currentLocation` (that was in store) and the `currentLocation` (being fetched from getAppContext) would be the same what `prevLocation` so we wouldn't make a single fetch.
That's why I added `initiallyFetched`, so if we have a situation where we have `currentLocation` in store, we make only **one** fetch, because we don't care about the `currentLocation` (the same) being fetched in the background from the `getAppContext`

jmiranda: We want to remove the complexity if we can just get to a place where the redux hook is not invoked unnecessarily when the current location has not changed. Our suspicion is that the the app context (being a complex object) might be too complex for whatever is detecting whether the current location hsa changed. Or maybe I misread that. 

Either way, one solution is adding a new API endpoint to retrieve just the current location and user IDs so that we can provide a simpler object to go through the useEffect hook to determine if the location changed.

For the issue mentioned above re: refetching twice vs not fetching at all ... I think I understand, but I have no idea how to do deal with that.

@awalkowiak mentioned that he will try to download the branch locally to test it out and see if there's an easy solution

src/js/components/purchaseOrder/PurchaseOrderListFilters.jsx
```diff
@@ -134,14 +133,9 @@ const PurchaseOrderListFilters = ({
...
   const isCentralPurchasingEnabled = supportedActivities.includes('ENABLE_CENTRAL_PURCHASING');
-  const prevLocation = usePrevious(currentLocation);
-  const [initiallyFetched, setInitiallyFetched] = useState(false);
 
   const determineDefaultValues = () => {
-    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
-    // but currentLocation doesn't change
-    // eslint-disable-next-line max-len
-    if ((!initiallyFetched && currentLocation?.id) || (prevLocation && prevLocation.id !== currentLocation?.id)) {
+    if (currentLocation?.id && buyers) {
```
kchelstowski, line 138: had to include `buyers` there as well, because it could also trigger refetch (you could have currentLocation.id and fetch, but the buyers fetch was still in progress and it caused another refetch then).

src/js/components/purchaseOrder/PurchaseOrderListFilters.jsx
```diff
@@ -170,12 +162,12 @@ const PurchaseOrderListFilters = ({
       fetchStatuses();
     }
 
-    if (!buyers || buyers.length === 0) {
+    if (!buyers) {
```
kchelstowski, line 165: I changed initialValue of `buyers` to `null`, because having it as `[]` was a bad idea, because if we had situation, when we would fetch none organization and made the `if` above like this: `buyers.length > 0` we would actually never fetch if we don't have buyers' organizations (even if we don't actually need them (we don't have central purchasing enabled).

src/js/reducers/organizationsReducer.jsx
```diff
@@ -6,7 +6,7 @@ const initialState = {
   data: [],
   fetched: false,
   suppliers: [],
-  buyers: [],
+  buyers: null,
```
kchelstowski, line 9: described above why I decided to change it

src/js/MainRouter.jsx
```diff
@@ -19,18 +19,17 @@ const TRANSLATION_PREFIXES = ['default', 'dashboard', 'combinedShipments', 'prod
...
-        options: {
-          renderToStaticMarkup,
-          onMissingTranslation,
-        },
-      });
-      this.props.setActiveLanguage(this.props.locale);
-      this.props.fetchMenuConfig();
-      TRANSLATION_PREFIXES.forEach(prefix => this.props.fetchTranslations('', prefix));
+    this.props.fetchSessionInfo();
+    this.props.initialize({
```
awalkowiak, line 23: On the initial load `this.props.supportedLocales` and `this.props.locale` will be empty and because you are not waiting for the data from `this.props.fetchSessionInfo()` you might wrongly initialize translations.

src/js/actions/index.js
```diff
@@ -105,11 +105,14 @@ export function fetchUsers() {
 
 export function fetchSessionInfo() {
   const url = '/openboxes/api/getAppContext';
-  const request = apiClient.get(url);
-
-  return {
-    type: FETCH_SESSION_INFO,
-    payload: request,
+  return (dispatch) => {
```
awalkowiak, line 108: Why you refactored this action?

src/js/components/invoice/InvoiceListFilters.jsx
```diff
@@ -111,20 +111,25 @@ const InvoiceListFilters = ({
...
-        id: currentLocation?.organization?.id,
-        value: currentLocation?.organization?.id,
-        name: currentLocation?.organization?.name,
-        label: currentLocation?.organization?.name,
-      },
-    });
-  }, [currentLocation]);
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
```
awalkowiak, line 116: Is it required? Looks like the next line is not that long.

src/js/components/products/ProductsListFilters.jsx
```diff
@@ -70,21 +71,27 @@ const filterFields = {
...
-      if (!acc[key]) return { ...acc, [key]: '' };
-      return acc;
-    }, {});
-    setDefaultValues({
-      ...initialEmptyValues,
-    });
-  }, []);
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
```
awalkowiak, line 84: Same as above

src/js/components/stock-movement/inbound/StockMovementInboundList.jsx
```diff
@@ -147,20 +147,25 @@ const StockMovementInboundList = (props) => {
...
-        id: props.currentLocation?.id,
-        value: props.currentLocation?.id,
-        name: props.currentLocation?.name,
-        label: props.currentLocation?.name,
-      },
-    });
-  }, [props.currentLocation]);
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
```
awalkowiak, line 152: Same as above

src/js/components/stock-movement/outbound/StockMovementOutboundList.jsx
```diff
@@ -161,21 +161,26 @@ const StockMovementOutboundList = (props) => {
...
-        value: props.currentLocation.id,
-        name: props.currentLocation.name,
-        label: props.currentLocation.name,
-      },
-      sourceType: props.isRequestsList ? 'ELECTRONIC' : null,
-    });
-  }, [props.currentLocation]);
+    // Avoid unnecessary re-fetches if getAppContext triggers fetching session info
+    // but currentLocation doesn't change
+    // eslint-disable-next-line max-len
```
awalkowiak, line 166: same as above

src/js/hooks/usePrevious.js
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from 'react';
+
+export default function usePrevious(value) {
```
awalkowiak, line 3: What is it for? Is is used anywhere?

webpack.config.js
```diff
@@ -120,6 +120,7 @@ module.exports = {
       templates: path.resolve(SRC, 'templates'),
       store: path.resolve(SRC, 'store'),
       css: path.resolve(ROOT, 'css'),
+      hooks: path.resolve(SRC, 'hooks'),
```
awalkowiak, line 123: Are we using this anywhere?

awalkowiak (changes requested):

src/js/actions/index.js
kchelstowski: because it is not how redux-thunk is supposed to work - I don't understand why we were passing request in dispatch instead of payload from this request - in some cases in turned out to be understood as "fetched" when I console.logged sessionReducer, but in reality the payload didn't come yet.

src/js/MainRouter.jsx
kchelstowski: I probably forgot to change it back, as I was trying to find any possible solution for my issue. Will bring it back

src/js/components/invoice/InvoiceListFilters.jsx
kchelstowski: Forgot to delete it after refactoring `if` to shorter version

src/js/hooks/usePrevious.js
kchelstowski: I also forgot to delete it after refactoring `if`. Should we consider keeping it anyway, as this hook might be useful in the future or should I delete it?

src/js/actions/index.js
awalkowiak: @kchelstowski makes sense, but then we have to refactor all remaining actions too. So I think there should be at least a TODO combined with a tech debt ticket to not forget about it.

drodzewicz (approved):

awalkowiak (approved):

PR #3530, opened 2022-10-06
OBS-1269 Remove dangling references when calling StockMovementService.reviseItems()

mdpearson:
Hi @jmiranda, I was able to reproduce OBS-1269 by moving back and forth a few times between Edit, Pick and Ship and editing the quantity field _without_ ever clicking the big red Undo button. More details are on the ticket.

The one-line summary of this change is that it basically presses the undo button for you if you don't press it yourself before editing the quantity.

In more detail, `reviseItems()`, as written, has guards around `removeShipmentAndPicklistItemsForModifiedRequisitionItem()` and `RequisitionItem::undoChanges()` which strikes me as both unnecessary and, at least in the case described in OBS-1269, incorrect. (Contrast the `revertItem()` implementation which calls them indiscriminately.)

In fact, nowhere else in this file do I see guards around either of these methods, while both of them have sufficient logic internally to prevent doing the wrong thing if they are called when there is in fact nothing to undo (I tested this assumption by making a separate build that called each method twice while testing this patch).

I was able to go back and forth cheerfully between states while making any number of edits with this patch in place. Is there anything else you'd like me to look into to make sure I didn't break something elsewhere?

### DIFF
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -21,8 +21,12 @@ import org.pih.warehouse.product.Category
 import org.pih.warehouse.product.Product
 import org.pih.warehouse.product.ProductGroup
 import org.pih.warehouse.product.ProductPackage
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
+
+    private static final transient Logger log = LoggerFactory.getLogger(RequisitionItem)
 
     def beforeInsert = {
         def currentUser = AuthService.currentUser.get()
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -271,7 +271,7 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
      */
     def changeQuantity(Integer newQuantity, ProductPackage newProductPackage, String reasonCode, String comments) {
 
-        println "Change quantity: " + newQuantity + " " + reasonCode + " " + comments
+        log.info "Change quantity: ${newQuantity} ${reasonCode} ${comments}"
         // And then create a new requisition item for the remaining quantity (if not 0)
         if (newQuantity == 0) {
             cancelQuantity(reasonCode, comments)
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -225,7 +225,7 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
             if (substitutionItem) {
                 removeFromRequisitionItems(substitutionItem)
                 requisition.removeFromRequisitionItems(substitutionItem)
-                modificationItem.disableRefresh = Boolean.TRUE
+                substitutionItem.disableRefresh = Boolean.TRUE
                 substitutionItem.delete()
             }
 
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1890,20 +1889,14 @@ class StockMovementService {
...
 
-                if (requisitionItem.shipmentItems || requisitionItem.picklistItems) {
-                    removeShipmentAndPicklistItemsForModifiedRequisitionItem(requisitionItem)
-                }
-
-                log.info "Item revised " + requisitionItem.id
-
-                // Cannot cancel quantity if it has already been canceled
-                if (requisitionItem.quantityCanceled) {
-                    requisitionItem.undoChanges()
```
jmiranda, line 1904: I can't think of under what circumstances this would occur, but there might be cases where we aren't canceling the quantity so we don't want to undo the changes. But I'll defer to @awalkowiak because this code has changed a lot since it was first written.

grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
```diff
@@ -21,9 +21,13 @@ import org.pih.warehouse.product.Category
 import org.pih.warehouse.product.Product
 import org.pih.warehouse.product.ProductGroup
 import org.pih.warehouse.product.ProductPackage
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
 
+    private static final transient Logger log = LoggerFactory.getLogger(RequisitionItem)
```
jmiranda, line 29: There should be an an implicit logger on all services, controller, and domain classes so you can probably remove this. Not sure why I used println here.

jmiranda (commented):
I'm going to defer to @awalkowiak on this one. It seems reasonable but I understand why these changes would fix the issue.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1890,20 +1889,14 @@ class StockMovementService {
                 }
 
                 if (!requisitionItem) {
-                    throw new IllegalArgumentException("Could not find stock movement item with ID ${stockMovementItem.id}")
+                    throw new IllegalArgumentException("Could not find requisition item for stock movement item with ID ${stockMovementItem.id}")
```
jmiranda, line 1892: This is unnecessary. A requisition item is *for all intents and purposes* a stock movement item. You cannot have a stock movement item without one.

grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
mdpearson: There should be, but the implicit logger doesn't come up during unit tests, which is why the previous PR failed tests (cf. f40f82b146). I have no idea why it doesn't under test (an implicit logger _is_ created when the application runs), but being explicit here fixes tests and also works properly when the app runs. That may be why you put a println here (there's only one other log statement in the whole domain/ directory)

PS this is me looking forward to grails 3

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
mdpearson: ok, good to know. Backed out in c4f461429

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
awalkowiak: @jmiranda I tried to be as nit picky in the review as I could, I pulled it locally too, and it seems to be fine. The `undo` button basically under the hood is doing the:
```
removeShipmentAndPicklistItemsForModifiedRequisitionItem(requisitionItem)
requisitionItem.undoChanges()
```
so we are simply doing the `revert item` first here. I think Katarzyna will find all the edge cases with it (for me it looks good).

Btw @jmiranda I found NPE in your recent changes in the `requisitionItem.undoChanges()` here: https://github.com/openboxes/openboxes/blob/develop/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy#L224. I am assuming, that it should be `substitutionItem.disableRefresh = Boolean.TRUE`, I'll push a fixing commit within this PR.

cc @mdpearson

awalkowiak (approved):

PR #3533, opened 2022-10-10
OBPIH-4895 add validation on create product type with empty code and product identifier

drodzewicz:


grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy
```diff
@@ -46,11 +46,14 @@ class ProductTypeController {
         def productTypeInstance = new ProductType(params)
         productTypeInstance.productTypeCode = ProductTypeCode.GOOD
         productTypeInstance.requiredFields = [ProductField.PRODUCT_CODE, ProductField.NAME, ProductField.CATEGORY, ProductField.GL_ACCOUNT]
-        if (productTypeInstance.save(flush: true)) {
+        if (!params.code && !params.productIdentifierFormat) {
+            productTypeInstance.errors.rejectValue("productIdentifierFormat","productType.codeOrIdentifierRequired.message")
+            productTypeInstance.errors.rejectValue("code", "")
```
kchelstowski, line 51: what is that line responsible for? the one with "code"?

drodzewicz: The idea was that when neither of two fields have values it should display a validation error at the top with a message: `productType.codeOrIdentifierRequired.message` and should highlight both fields _(red border)_ where the error is coming from.
To do that we need to assign errors to those fields with `rejectValue(fielName, message)`.
Since error is triggered when both fields are `null` I needed to display **one message** so I assigned that message to one of two field errors and other one set to empty string.

kchelstowski: my bad, I thought there was `||`, not `&&`.
Anyway, wouldn't it work if you left only first line? In the line below you write `!productTypeInstance.hasErrors()`, so even if you had only
 ```
productTypeInstance.errors.rejectValue("productIdentifierFormat","productType.codeOrIdentifierRequired.message")
``` 
it would trigger the error (because at least one error exists), wouldn't it? Or do I misunderstand something?

Comment (drodzewicz) [reply grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy:51]: yes, but I also wanted to mark the fields that the error is associated with.
On `create.gsp` there is checking for field errors on each field
```
<tr class="prop">
	<td valign="top" class="name">
		<label for="code"><warehouse:message code="productType.code.label" default="Code" /></label>
	</td>
	<td valign="top" class="value ${hasErrors(bean: productTypeInstance, field: 'code', 'errors')}">
		<g:textField class="text" size="80" name="code" value="${productTypeInstance?.code}" />
	</td>
</tr>
```
if field has errors, it applies class `errors` with red border styling

kchelstowski: ok, now sounds reasonable :)

kchelstowski (approved):

awalkowiak (approved):

PR #3536, opened 2022-10-11
OBPIH-4910 Add searchable supported activity to default product type

drodzewicz:


grails-app/migrations/0.8.x/changelog-2022-10-11-1200-add-default-product-type.xml
```diff
@@ -5,7 +5,7 @@
 
   <property name="now" value="now()"/>
 
-  <changeSet author="awalkowiak" id="0209202212000-0">
+  <changeSet author="awalkowiak" id="1110202212000-0">
```
awalkowiak, line 8: Why you made this change (and file name change)? If it is because you were getting an error that the changeset hash has changed, then you can add a `validCheckSum` property.

drodzewicz: yes, that was the error.
I didn't know that you can fix it by adding `validCheckSum` property.
Will fix it

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
```diff
@@ -31,4 +31,20 @@
       UPDATE product SET product_type_id = (SELECT id FROM product_type WHERE id like "DEFAULT" OR code like "DEFAULT" OR name like "DEFAULT" LIMIT 1) WHERE product_type_id IS NULL;
     </sql>
   </changeSet>
+  <changeSet author="awalkowiak" id="0209202212000-2">
+    <preConditions onFail="MARK_RAN">
+      <and>
+        <sqlCheck expectedResult="1">
+          SELECT COUNT(*) FROM product_type WHERE id="DEFAULT"
```
awalkowiak, line 38: Please make make the check here as it was done before, which is:
`<sqlCheck expectedResult="0">SELECT COUNT(*) FROM product_type WHERE id like "DEFAULT" OR code like "DEFAULT" OR name like "DEFAULT"</sqlCheck>`

grails-app/migrations/0.8.x/changelog-2022-09-02-1200-add-default-product-type.xml
```diff
@@ -31,4 +31,20 @@
...
+        </sqlCheck>
+        <sqlCheck expectedResult="0">
+          SELECT COUNT(*) FROM product_type_supported_activities WHERE product_type_id="DEFAULT" AND product_activity_code="SEARCHABLE"
```
awalkowiak, line 41: Please make here the part where you check product type id as
`product_type_id = (SELECT id FROM product_type WHERE id like "DEFAULT" OR code like "DEFAULT" OR name like "DEFAULT" LIMIT 1)`

awalkowiak (changes requested):

jmiranda:
@drodzewicz did you fix the validchecksum issue? if possible i would prefer to only add changesets, not modify them.

drodzewicz:
@jmiranda the validchecksum issue was something I had locally so everything is fine.

Current PR file changes only add new changeset of `id="0209202212000-2"` or did you meant to say that I should create new changelog file and not modify the old one ?

jmiranda:
> Current PR file changes only add new changeset of id="0209202212000-2" or did you meant to say that I should create new changelog file and not modify the old one ?

@drodzewicz I meant "never edit an existing changeset" unless it goes through Artur or me first. The only reason you would need to use validCheckSum is if you modified an existing changeset i.e. one that has already been executed in your database. 

But I think I understand what happened now. You were just modifying the changeset that you were working on in this PR. If you need to modify a changeset that you're actively working on (i.e. that hasn't been merged to develop already) you just need to remove the changeset record from DATABASECHANGELOG. And also revert any changes made to the database. 

If for some reason you need to modify a changeset that has been merged to develop (and has been deployed to any of our environments) the best case scenario is to create a new changeset that fixes the old one. If @awalkowiak or @jmiranda determine that we actually need to modify the existing changeset (i.e. it causes irreparable damage) then we can add a validCheckSum element to the changeset.

drodzewicz:
> > Current PR file changes only add new changeset of id="0209202212000-2" or did you meant to say that I should create new changelog file and not modify the old one ?
> 
> @drodzewicz I meant "never edit an existing changeset" unless it goes through Artur or me first. The only reason you would need to use validCheckSum is if you modified an existing changeset i.e. one that has already been executed in your database.
> 
> But I think I understand what happened now. You were just modifying the changeset that you were working on in this PR. If you need to modify a changeset that you're actively working on (i.e. that hasn't been merged to develop already) you just need to remove the changeset record from DATABASECHANGELOG. And also revert any changes made to the database.
> 
> If for some reason you need to modify a changeset that has been merged to develop (and has been deployed to any of our environments) the best case scenario is to create a new changeset that fixes the old one. If @awalkowiak or @jmiranda determine that we actually need to modify the existing changeset (i.e. it causes irreparable damage) then we can add a validCheckSum element to the changeset.

Understood. Will keep that in mind for the future.

jmiranda (approved):

PR #3537, opened 2022-10-11
OBPIH-4932 Unable to delete request as superuser

kchelstowski:
I'm wondering if it's somehow possible to "inject" `def userService` like we do everywhere, so just using `def userService` at the top of file?
I had hard time to find how to use the `UserService` here, because using method like `UserService.someMethod()` didn't want to work.
@awalkowiak ?

### DIFF
--- a/src/groovy/org/pih/warehouse/api/StockMovement.groovy
+++ b/src/groovy/org/pih/warehouse/api/StockMovement.groovy
@@ -2,6 +2,7 @@ package org.pih.warehouse.api
 
 import org.apache.commons.collections.FactoryUtils
 import org.apache.commons.collections.list.LazyList
+import org.codehaus.groovy.grails.commons.ApplicationHolder
 import org.codehaus.groovy.grails.commons.ConfigurationHolder
 import org.codehaus.groovy.grails.validation.Validateable
 import org.pih.warehouse.core.ActivityCode
@@ -10,6 +11,7 @@ import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Person
 import org.pih.warehouse.core.Role
 import org.pih.warehouse.core.User
+import org.pih.warehouse.core.UserService
 import org.pih.warehouse.inventory.StockMovementStatusCode
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.OrderItemStatusCode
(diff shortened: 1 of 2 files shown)


awalkowiak:
@kchelstowski IIRC you have to use `ApplicationHolder` here.

kchelstowski:
@awalkowiak Ok, so that's what I've actually done

drodzewicz:
is `package-lock.json` neccesary fro this PR ?

drodzewicz (changes requested):
is `packeg-lock.json` necessary for this PR ?

kchelstowski:
@drodzewicz I've just had to do `npm install` when changing to branch from `develop`'s base after a long time

src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -230,8 +232,8 @@ class StockMovement {
         if (electronicType) {
             User user = AuthService.currentUser.get()
             def accessRule = ConfigHelper.findAccessRule("stockRequest", "remove")
-            def userRoles = user.getEffectiveRoles(currentLocation)
-            if (!userRoles.any { Role role -> role.roleType == accessRule?.accessRules?.minimumRequiredRole }) {
+            def userService = ApplicationHolder.getApplication().getMainContext().getBean("userService")
```
jmiranda, line 235:
```
ApplicationHolder.application.mainContext.getBean("userService")
```

jmiranda: This doesn't feel right so we should add this to our tech debt pile to figure out a better way to centralize this type of custom authorization logic.

jmiranda (changes requested):

src/groovy/org/pih/warehouse/api/StockMovement.groovy
kchelstowski: @jmiranda what would you suggest at this point? I believe this one has to go to release and we already do have similar things implemented. Should we consider pushing it further and then refactor all this stuff after the tech debt or shouldn't we merge this one?

jmiranda: @kchelstowski Yeah I meant we should move forward with this but make sure we add this to our tech debt to discuss it further.

jmiranda: @kchelstowski "This doesn't feel right" was in reference to the use of the service in the StockMovement class, not my recommended change. I'd still like you to make the change suggested above.

jmiranda (approved):

PR #3539, opened 2022-10-11
OBPIH-4913 Add materialized view for the OrderSummary

awalkowiak:
~@jmiranda here are a few notes before you take a look at it:~

~1. I decided to leave the old views as they were before to have everything cleaner (and in one place) for doing the refresh updates. I named the "materialized" view as a `mv_order_summary` (I think we could/should[?] have a distinguisher if a table is a materialized view or a normal table).
2. The performance of the `create table as select * from view` is very fast (tested on about 755 POs, but it was less than 1s)
3. I wonder how to change all the `beforeDelete/afterDelete` for all item objects to not run into issues that something is already deleted and we can't get the id of the order or if it is not yet deleted and the values in the views are still obsolete. (For the order itself it works fine, so that's a good part).
4. I wonder if there should be an option to refresh for a list of orders, but I can't see the use case, because we are refreshing after items update.
5. I am not sure how I could make an improvement where I hold with an update on an order if there might be an update or deletion on multiple objects related to that one order.~

Done all improvements


grails-app/migrations/views/changelog.xml
```diff
@@ -75,7 +75,15 @@
     <sqlFile path="views/stocklist-item-list.sql"/>
   </changeSet>
   <changeSet id="1580848680306-23" author="awalkowiak" runOnChange="true" runAlways="true" failOnError="true">
-    <sqlFile path="views/order-summary-view.sql"/>
+    <sqlFile path="views/order-summary-helper-views.sql"/>
```
jmiranda, line 78: Does this change trigger a checksum validation?

grails-app/migrations/views/order-summary-materialized-view.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS mv_order_summary;
+CREATE TABLE mv_order_summary AS SELECT * FROM order_summary;
```
jmiranda, line 2: Careful here. There's going to be a time where this table is empty / missing and all queries will fail. We want that time to be as short as possible. So it would be better to create a temporary table mv_order_summary_temp) to load the data from the order_summary view and then replace mv_order_summary with that table i.e. either through a table rename or a similar CREATE TABLE SELECT or INSERT INTO SELECT.

jmiranda: Although I guess this is fine since it's just the initial load.

jmiranda (approved):
Looks good. The one thing to be careful with ... where / when we trigger the refresh needs to be somewhat strategic. If we update order and order items in the same request, then this will lead to lots of unnecessary refreshes.

grails-app/migrations/views/changelog.xml
awalkowiak: No complains found in the console about this

grails-app/migrations/views/order-summary-materialized-view.sql
awalkowiak: Fixed it

grails-app/domain/org/pih/warehouse/order/OrderSummary.groovy
```diff
@@ -29,6 +29,7 @@ class OrderSummary {
     static mapping = {
         version false
         cache usage: "read-only"
+        table "mv_order_summary"
```
jmiranda, line 32: order_summary_mv

Although I think we should take another look at our naming conventions. Part of me is thinking that I'd like this one to be order_summary and we can rename the view to be order_summary_v or something like that. But we can leave that discussion for another time.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1624,6 +1624,9 @@ class StockMovementService {
         shipment.destination = stockMovement.destination
         shipment.shipmentType = ShipmentType.get(Constants.DEFAULT_SHIPMENT_TYPE_ID)
 
+        // Save shipment before adding the items to avoid dead locks
+        shipment.save()
```
jmiranda, line 1628: Why would there be deadlocks here? Are you actually seeing deadlocks?

awalkowiak: Actually, it was not a `dead lock` issue, but `referencing an unsaved transient instance` and I simply moved this save from the 1648 line to be done once, instead of for each item. I'll fix the comment.

grails-app/domain/org/pih/warehouse/order/OrderSummary.groovy
awalkowiak: done

PR #3541, opened 2022-10-12
OBPIH-4896 Handle already existing product type with empty code and product identifier

drodzewicz:


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -196,6 +196,14 @@ class ProductController {
 
         updateTags(productInstance, params)
 
+        ProductType defaultProductType = ProductType.defaultProductType.list()?.first();
+        // Throw an error for product type with empty code and product identifier that is not a default product type
+        if (productInstance.productType != defaultProductType && !productInstance.productType?.code && !productInstance.productType?.productIdentifierFormat) {
```
kchelstowski, line 201: I'm just wondering if you can compare `productInstance.productType` to `defaultProductType` using equality comparator - even if they are of the same class and have the same properties and keys, will the reference be the same, so you can compare objects?
I believe in JS it wouldn't work (`{} === {}` would return false), but does it work in Groovy and you were aware of that?

drodzewicz: yes it was intentional and it does work as intended.

kchelstowski: cool, then JS just sucks in this case :D

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: This is a good point. And generally depends on the equals() implementation. 

https://stackoverflow.com/questions/28500091/grails-gorm-what-is-the-default-equals-implementation-in-gorm

In general, we're probably ok. But I've seen cases where this doesn't do the right thing so it's better to compare primitives (productType.id) or implement equals. 

We already have the "equals / hashCode" issue on our tech debt pile so I might do this to be on the safe side.
```
productInstance?.productType?.id != defaultProductType?.id
```

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
drodzewicz: yeah you guys have a good point.
Just to be on the safe lets compare primitive values to avoid potential bugs.

jmiranda: @drodzewicz I'll merge it for now since we want to get these obdev, but please remember to make and test that change when you get a chance.

PR #3547, opened 2022-10-13
OBPIH-4897 add autoincremental product code assignment when create products through import

drodzewicz:


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -198,7 +198,7 @@ class ProductController {
 
         ProductType defaultProductType = ProductType.defaultProductType.list()?.first();
         // Throw an error for product type with empty code and product identifier that is not a default product type
-        if (productInstance.productType != defaultProductType && !productInstance.productType?.code && !productInstance.productType?.productIdentifierFormat) {
+        if (productInstance.productType?.id != defaultProductType?.id && !productInstance.productType?.code && !productInstance.productType?.productIdentifierFormat) {
```
drodzewicz, line 201: addressed the primitive value comparison instead of the whole `ProductType` object 
based on the discussion from https://github.com/openboxes/openboxes/pull/3541#discussion_r994072420

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -938,7 +938,7 @@ class ProductController {
                 flash.message = "All ${command?.products?.size()} product(s) were imported successfully."
                 redirect(controller: "product", action: "importAsCsv", params: [tag: tags[0]])
             } catch (ValidationException e) {
-                command.errors = e.errors
+                command.errors.reject(e.message)
```
jmiranda, line 941: I don't know if this is what we want to do. Can you provide justification for this? The validation exception has an Errors object that includes an detailed error message for each invalid data element. Displaying just the message associated with the validation exception give us a generic error message without those details.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -731,6 +731,18 @@ class ProductService {
             addTagsToProduct(product, tags)
             addTagsToProduct(product, productProperties.tags)
 
+            ProductType defaultProductType = ProductType.defaultProductType.list()?.first();
+            // Throw an error for product type with empty code and product identifier that is not a default product type
+            if (product.productType?.id != defaultProductType?.id && !product.productType?.code && !product.productType?.productIdentifierFormat) {
+                throw new ValidationException("Could not save product '" + product.name + "'", product.errors)
```
jmiranda, line 737: Instead of throwing an exception here you should just add to the product's errors object. 
```
if (product.productType?.id != defaultProductType?.id && !product.productType?.code && !product.productType?.productIdentifierFormat) {
    product.errors.reject("...")
}
```

The save below should fail due to the error and throw the exception. If for some reason it doesn't you can call validate on your own.
```
if (!product.validate() || !product.save(flush:true)) {
```

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -731,6 +731,18 @@ class ProductService {
...
+            }
+
+            if (!product?.id || product.validate()) {
```
jmiranda, line 740: I was just about to comment that we should probably check if the product is valid here before calling generateProductIdentifier ... and then I saw this. Good work.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
drodzewicz: for some reason, that I wasn't able to figure out, assigning `e.errors` to `command.errors` didn't work.
It didn't display any error messages on `.gsp` page even thought error was was thrown.

I will try to do `product.errors.reject("...")` inseatd of throwing an exception, I was just following an example where few line below we have
```
 if (!product.save(flush: true)) {
    hrow new ValidationException("Could not save product '" + product.name + "'", product.errors)
}
```

drodzewicz: it was `importAsCsv.gsp` 
https://github.com/openboxes/openboxes/blob/develop/grails-app/views/product/importAsCsv.gsp#L18
even thought it has appropriate logic for checking if `command` has errors and siplay those errors but for some reason it didn't as if it lost those errors somewhere along the way

jmiranda: Ok, well make the suggested change(s) and if you still can't get it to work I can take a look.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
drodzewicz: Unfortunately doing `product.errors.reject("...")` and after that calling  `product.validate()` clears out previously assigned errors. I wasn't able to find any useful information on this topic but I assume that by calling  `product.validate()` it just resets `product.errors` and validates the object based on specified constrains in the class.

So instead I moved my validation logic into `validateProducts(csv)` _(where it probably should have been in the first place)_ and everything seems to work well there.

jmiranda: I still see the exception being thrown here. 

https://github.com/openboxes/openboxes/pull/3547/files#diff-c318547de40b750cb640136d100a6aee922f7aae15814cdc220741c3e33d4575R641

jmiranda: Oh nevermind. I see what you did. But just for clarification ... validate should not clear product.errors so we should try to figure out how to do this properly at some point.

jmiranda (approved):

PR #3550, opened 2022-10-13
OBPIH-4871 Comment tab displayed with smaller font than other tabs

drodzewicz:


drodzewicz:
the `.badge` is already in use by the newer boostrap version so I've just changed the class name to `.tab-badge`

jmiranda:
@kchelstowski what do you think?

kchelstowski (approved):
LGTM also. Probably the same kind of problem I had with `row` for adjust stock page

PR #3552, opened 2022-10-14
OBPIH-4924 fix putaway list redirects

drodzewicz:
- hide order list link button on show putaway show page
- fix redirect after delete pending putaway
- disable product type filter on putaway list

### DIFF
--- a/grails-app/views/order/_actions.gsp
+++ b/grails-app/views/order/_actions.gsp
@@ -12,12 +12,20 @@
 			<img src="${resource(dir: 'images/icons/silk', file: 'bullet_arrow_down.png')}" />
 		</button>
 		<div class="actions" style="min-width: 200px;">
-			<g:if test="${!request.request.requestURL.toString().contains('order/list')}">
+			<g:if test="${!request.request.requestURL.toString().contains('order/list') && orderInstance?.orderType != PUTAWAY_ORDER}">
 				<div class="action-menu-item">
-					<g:link controller="order" action="list">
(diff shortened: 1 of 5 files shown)


grails-app/views/order/_actions.gsp
```diff
@@ -14,10 +14,18 @@
...
-						<img src="${createLinkTo(dir:'images/icons/silk',file:'application_view_list.png')}" alt="View orders" style="vertical-align: middle" />
-						&nbsp;${warehouse.message(code: 'order.list.label')}
-					</g:link>
+					<g:if test="${orderInstance?.orderType == PURCHASE_ORDER}">
+						<g:link controller="purchaseOrder" action="list">
+							<img src="${createLinkTo(dir:'images/icons/silk',file:'application_view_list.png')}" alt="View orders" style="vertical-align: middle" />
+							&nbsp;${warehouse.message(code: 'order.list.label')}
+						</g:link>
+					</g:if>
+					<g:elseif test="${orderInstance?.orderType != PUTAWAY_ORDER}">
```
kchelstowski, line 23: what if orderType is equal to `PUTAWAY_ORDER`? I can't see where we handle this case?

drodzewicz: that was the point of not rendering this button for `PUTAWAY_ORDER` but I guess it is not too clear enough from the code.
So I guess I'll just make it a `<g:else>` and add checking for `orderInstance?.orderType != PUTAWAY_ORDER` to line 15

kchelstowski: ok, I see, it was hard to "read" from the code what it is about to do, but I get the point now

PR #3558, opened 2022-10-18
OBPIH-4914 Errors with loading / fetching redesigned list pages

kchelstowski:
I also fixed PO page, because when moving all the filter logic to `PurchaseOrderList`, the destination when changing location didn't change - it always stayed the same, because when we initially go to the PO page, we pass our current location as destination in the URL, but when changing location, the queryParams didn't change and this part:
```
 if (queryProps.destination) {
      defaultValues.destination = props.currentLocation.id === queryProps.destination
        ? currentLocationOption
        : await fetchLocationById(queryProps.destination);
    }
```
always took initially set destination from URl so it had to be changed to: 
`await fetchLocationById(props.currentLocation?.id);`

it was agreed with Katarzyna and me that I would fix it while doing this ticket, just wanted to explain it if you were wondering, why I changed this part also.

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -2928,7 +2928,7 @@ react.invoice.buyerOrganization.label=Buyer organization
 react.invoice.createdBy.label=Created by
 react.invoice.viewDetails.label=View Invoice Details
 react.invoice.edit.label=Edit Invoice
-react.invoice.error.fetching.label=Could not fetch list of invoices (most probably due to abort while changing location)
+react.invoice.error.fetching.label=Unable to fetch invoices
 
 # Stock Movement
 # Atributes
@@ -3131,8 +3131,8 @@ react.stockMovement.status.issued.description.label=This shipment has been shipp
 react.stockMovement.status.checking.description.label=This shipment is ready to be shipped.
 react.stockMovement.areYouSure.label=Are you sure you want to delete this Stock Movement?
 react.stockMovement.deleted.success.message.label=Stock Movement has been deleted successfully
-react.stockMovement.inbound.fetching.error=Could not fetch inbound movements (most probably due to abort while changing location)
-react.stockMovement.outbound.fetching.error=Could not fetch outbound movements (most probably due to abort while changing location)
+react.stockMovement.inbound.fetching.error=Unable to fetch inbounds movements
+react.stockMovement.outbound.fetching.error=Unable to fetch outbound movements
 
 react.StockMovementType.enum.RETURN_ORDER=Return
 react.StockMovementType.enum.STOCK_MOVEMENT=Request
@@ -3421,7 +3421,7 @@ react.stockTransfer.view.label=View stock transfer
 react.stockTransfer.delete.label=Delete
 react.stockTransfer.delete.confirm.label=Are you sure you want to delete this stock transfer?
 react.stockTransfer.delete.success.label=Stock transfer has been deleted successfully
-react.stockTransfer.fetch.fail.label=Could not fetch list of stock transfers (most probably due to abort while changing location)
+react.stockTransfer.fetch.fail.label=Unable to fetch stock transfers
 
 # Replenishment
 react.replenishment.label=Replenishment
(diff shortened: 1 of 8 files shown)


src/js/components/invoice/InvoiceListTable.jsx
```diff
@@ -36,7 +38,22 @@ const InvoiceListTable = ({
 
   // Util ref for react-table to force the fetch of data
   const tableRef = useRef(null);
+
+  // Cancel token/signal for fetching data
+  const { CancelToken } = axios;
+  const sourceRef = useRef(CancelToken.source());
+
+  // eslint-disable-next-line consistent-return
```
drodzewicz, line 46: remove this eslint warning and move the if condition inside return 
```
return () => {
    if (currentLocation?.id) sourceRef.current.cancel('Fetching canceled');
};
```

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -57,11 +58,27 @@ const PurchaseOrderListTable = ({
 
   // Util ref for react-table to force the fetch of data
   const tableRef = useRef(null);
+
+  // Cancel token/signal for fetching data
+  const { CancelToken } = axios;
```
drodzewicz, line 63: you can probably just do `import { CancelToken } from 'axios';`

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -57,11 +58,27 @@ const PurchaseOrderListTable = ({
...
+  const sourceRef = useRef(CancelToken.source());
+
   const fireFetchData = () => {
+    // Each time we fetch, we want to 'reset' the token/signal
+    sourceRef.current = CancelToken.source();
     tableRef.current.fireFetchData();
   };
   const isCentralPurchasingEnabled = supportedActivities.includes('ENABLE_CENTRAL_PURCHASING');
 
+  // eslint-disable-next-line consistent-return
```
drodzewicz, line 73: same as previouse, should probably move `if (...)` inside return

src/js/components/stock-movement/inbound/StockMovementInboundTable.jsx
```diff
@@ -45,7 +46,23 @@ const StockMovementInboundTable = ({
 
   const tableRef = useRef(null);
 
-  const fireFetchData = () => tableRef.current.fireFetchData();
+  // Cancel token/signal for fetching data
+  const { CancelToken } = axios;
```
drodzewicz, line 50: same import `CancelToken` straight from `axios`

drodzewicz (changes requested):

drodzewicz (approved):
everything else looks good

grails-app/i18n/messages.properties
```diff
@@ -2928,7 +2928,7 @@ react.invoice.buyerOrganization.label=Buyer organization
 react.invoice.createdBy.label=Created by
 react.invoice.viewDetails.label=View Invoice Details
 react.invoice.edit.label=Edit Invoice
-react.invoice.error.fetching.label=Could not fetch list of invoices
+react.invoice.error.fetching.label=Could not fetch list of invoices (most probably due to abort while changing location)
```
awalkowiak, line 2931: Where did this label come from?

awalkowiak (commented):
LGTM, but I strongly dislike the labels for the fetching errors or fails

grails-app/i18n/messages.properties
kchelstowski: obviously from my imagination :( what would you suggest?

jmiranda: Why do we need to add more context here? I would just say "Unable to fetch invoices" and leave it at that.

jmiranda: Or "Unable to fetch invoices. Try refreshing the page."

awalkowiak (approved):

PR #3559, opened 2022-10-18
OBPIH-4790 implement product association import

drodzewicz:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+                return
+            }
+
+            // Check for association duplicates
+            List<ProductAssociation> foundProductAssociations = ProductAssociation.findAllWhere([
+                    product             : productAssociationInstance.product,
+                    associatedProduct   : productAssociationInstance.associatedProduct,
+                    code                : productAssociationInstance.code,
+            ])
+            if (foundProductAssociations && foundProductAssociations.size() > 0) {
```
jmiranda, line 73: Is there a case where we want to update a product association?

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+            }
+            if (!product.active) {
+                command.errors.reject(
+                    "Row ${index + 1}: Product with code '${productAssociationInstance.product?.id}' is inactive"
+                )
+                return
+            }
+            if (!associatedProduct.active) {
+                command.errors.reject(
+                    "Row ${index + 1}: Product with code '${productAssociationInstance.associatedProduct?.id}' is inactive"
```
jmiranda, line 58: Use productCode instead of ID.

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+            }
+            if (!associatedProduct) {
+                command.errors.reject(
+                    "Row ${index + 1}: Product with code '${productAssociationInstance.associatedProduct?.id}' does not exist"
+                )
+                return
+            }
+            if (!product.active) {
+                command.errors.reject(
+                    "Row ${index + 1}: Product with code '${productAssociationInstance.product?.id}' is inactive"
```
jmiranda, line 52: Use productCode instead of ID.

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+                    associatedProduct   : productAssociationInstance.associatedProduct,
+                    code                : productAssociationInstance.code,
+            ])
+            if (foundProductAssociations && foundProductAssociations.size() > 0) {
+                command.errors.reject("Row ${index + 1}: Association already exists")
+                return
+            }
+
+
+            // find two-way association pairs
```
jmiranda, line 79: Can you provide a better comment here? I have no idea what is happening here.

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            if (twoWayAssociationPair) {
+                // mark this association as two-way association
+                params.hasMutualAssociation = true
+                if (twoWayAssociationPair['quantity'] * productAssociationInstance.quantity != 1) {
+                    command.errors.reject(
+                            "Row ${index + 1}: Quantity of Product association for Products with codes " +
+                            "'${productAssociationInstance.product?.id}' and " +
+                            "'${twoWayAssociationPair['associatedProduct.id']}' does not match"
```
jmiranda, line 93: Again ... productCode.

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+                    command.errors.reject(
+                            "Row ${index + 1}: Product Association with Product code " +
+                            "'${productAssociationInstance.product?.id}': ${error.getFieldError()}"
+                    )
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
jmiranda, line 110: Let's add more comments to make this method more readable. I'll review again once the comments are there.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,137 @@
...
+        command.data.eachWithIndex { params, index ->
+            ProductAssociation productAssociationInstance = new ProductAssociation(params)
+
+            // disable update product association on import
+            if (productAssociationInstance?.id) {
+                command.errors.reject("Row ${index + 1}: Cannot edit existing associations via import")
+                return
+            }
+
+            Product product = Product.get(productAssociationInstance.product?.id)
```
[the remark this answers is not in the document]
drodzewicz: yes it does but it also sets non existing products, for example if you pass a non existing product id, it resolves it into an "empty object" _(object of ProductAssociation class with all the fields of value null)_ so `if(thatEmptyObject)` results in `true`. 
So basically my problem was that I wanted to check if product exists in the DB and I wasn't sure how can I do this for 
```
params['product.id'] = 'nonExistingId123'
ProductAssociation thatAssiosaction = new ProductAssociation(params) 

if (thatAssiosaction.product) { ... }
```

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: No, as specified in the ticket product association import should not allow association updates, only creations

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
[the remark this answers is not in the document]
drodzewicz: actually scratch that I just found a way.
I can check if product exists by using `exists(...)` method
```
if (Product.exists(params['product.id'])) { ... }
```

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,141 @@
...
+class ProductAssociationDataService {
+
+    /**
+     * Validate product association
+     */
+    Boolean validateData(ImportDataCommand command) {
+        log.info "Validate data " + command.filename
+
+        command.data.eachWithIndex { params, index ->
+            ProductAssociation productAssociationInstance = new ProductAssociation(params)
```
drodzewicz, line 27: Now that I have replaced all instances where I used `product.id` with `product.productCode` and my params have 
```
params: [ 
    'product.prductCode': '16661',
    .... 
 ]
```

instead of 
```
params: [ 
    'product.id': '16661',
    .... 
 ]
```
now
` ProductAssociation productAssociationInstance = new ProductAssociation(params)` does not populate `product` and `associatedProduct` fields instead I have to explicitly do
```
productAssociationInstance.product = Product.findByProductCode(params['product.productCode'])
productAssociationInstance.associatedProduct = Product.findByProductCode(params['associatedProduct.productCode'])
```

awalkowiak: I'd say that's proper behavior. If there is a product code in the file, then you have to find it as you did, because IIRC it has to be an id in order to properly create object from params.

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,141 @@
...
+            productAssociationInstance.product = Product.findByProductCode(params['product.productCode'])
+            productAssociationInstance.associatedProduct = Product.findByProductCode(params['associatedProduct.productCode'])
+            params['product.id'] =  productAssociationInstance.product?.id
```
awalkowiak, line 30: @drodzewicz is there a need to assign an id to params if you already have a `product` and an `associatedProduct` on the `productAssociationInstance`?

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -0,0 +1,141 @@
...
+                command.errors.reject("Row ${index + 1}: Association already exists")
+            }
+
+
+            // find another product association that matches current associationInstance
+            // for two-way association relationship, where:
+            // - association.product = otherAssociation.associatedProduct
+            // - association.associatedProduct = otherAssociation.product
+            // - association.code = otherAssociation.code
+            def otherMatchingTwoWayAssociation = command.data.find {
```
awalkowiak, line 77: Should not this be searched in all associations in the system instead in the file data only?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: the reason for that is that I have `productAssociationInstance.product` and `productAssociationInstance.associatedProduct`  inside a  
`Boolean validateData(ImportDataCommand command) { ... }` method and after validating we move on to `importData(ImportDataCommand command) { ... }` method which is responsible for creating and saving those associations.

And so for 
`ProductAssociation firstProductAssociationInstance = new ProductAssociation(params)` 
to work properly
**params** has to have properties `product.id` and `associatedProduct.id` but instead it only has  `product.productCode` and `associatedProduct.productCode`

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: I think this might be a topic for a tech huddle and at the moment we should focus only on associations provided in the imported file.

This is also falls into a case of updating associations which is probably out of scope for this ticket

PR #3561, opened 2022-10-19
OBPIH-4922 Fix select dropdown on components using utils/Select

awalkowiak:


src/js/components/form-elements/SelectField.jsx
```diff
@@ -91,10 +11,7 @@ const SelectField = (props) => {
       {...attributes}
       className={`select-xs ${className}`}
       classNamePrefix="react-select"
-      customSelectComponents={{
-        Menu,
-        Option,
-      }}
+      customSelectComponents={{}}
```
drodzewicz, line 14: `customSelectComponents={{}}` empty object prop seems unnecessary since you already have defined 

```
Select.defaultProp = {
    customSelectComponents: {}
}
```

drodzewicz (changes requested):

src/js/components/form-elements/SelectField.jsx
awalkowiak: Yup, you're right, fixed

drodzewicz (approved):

drodzewicz (approved):

PR #3562, opened 2022-10-19
UI redesign of the menu, location chooser and list pages.

awalkowiak:


awalkowiak:
@jmiranda, @mdpearson this one is ready

grails-app/conf/Config.groovy
```diff
@@ -331,7 +331,7 @@ jqueryValidationUi {
 
 
 // Allow users to customize logo image url as well as label
-openboxes.logo.url = "https://openboxes.com/img/logo_30.png"
+openboxes.logo.url = "/openboxes/images/logo/openboxes_logo_40x40.jpg"
```
jmiranda, line 334: /openboxes should probably be replaced with ${appName}

jmiranda: With that said, we can push that change off until afterwards.

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -24,6 +27,40 @@ class ProductApiController extends BaseDomainApiController {
     def grailsApplication
     def productAvailabilityService
 
+    def list = {
```
jmiranda, line 30: We need to remember to include some warnings about this breaking API change.

jmiranda (approved):
LGTM. I got a quarter of the way through. I can't do any more. We'll fix the broken in the release branch.

PR #3563, opened 2022-10-20
OGBM-366 Prevent another missing-@Transactional exception.

mdpearson:
Before this patch, the grails 3 build would occasionally raise "No Session found for current thread" when navigating between gsp and react-based pages.


awalkowiak (approved):

jmiranda (commented):
@mdpearson See my comments in the ticket. cc @awalkowiak

mdpearson:
Will revisit as part of OBGM-363

PR #3564, opened 2022-10-20
OBPIH-4790 Product Association import Fix

drodzewicz:
- change product association template to match downloaded file
- change download data to return xls instea dof csv
- fix association validation errors


grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -24,16 +25,30 @@ class ProductAssociationDataService {
         log.info "Validate data " + command.filename
 
         command.data.eachWithIndex { params, index ->
+            DateFormat dateFormat = new SimpleDateFormat("yyyy-mm-dd hh:mm:ss")
+            if (params['dateCreated']) params['dateCreated'] = dateFormat.parse(params['dateCreated'])
+            if (params['lastUpdated']) params['lastUpdated'] = dateFormat.parse(params['lastUpdated'])
```
awalkowiak, line 30: Why you are importing these datetimes? Why it is not handled by the gorm as an auto timestamp? Is there a case where this makes sense to import these values?

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -24,16 +25,30 @@ class ProductAssociationDataService {
...
+
+            params['code'] = params['code']?.toUpperCase()
```
awalkowiak, line 32: Why there is `toUpperCase`?

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -24,16 +25,30 @@ class ProductAssociationDataService {
...
             params['product.id'] =  productAssociationInstance.product?.id
             params['associatedProduct.id'] =  productAssociationInstance.associatedProduct?.id
 
             // disable update product association on import
-            if (productAssociationInstance?.id) {
+            if (params['id']) {
                 command.errors.reject("Row ${index + 1}: Cannot edit existing associations via import")
             }
+
+            if (!productAssociationInstance?.code) {
```
awalkowiak, line 45: Turn it into:
```
if (!params['code']) {
  command.errors.reject("Row ${index + 1}: Association Type field can not be empty")
} else if (!productAssociationInstance?.code) {
  command.errors.reject("Row ${index + 1}: Association Type code '${params['code']}' does not exist")
}
```
to get rid of 1 `if`.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: as mentioned by Katarzyna after QA
_"when try to import association type "Substitute" or "substitute", works only for SUBSTITUTE - it shouldn't matter and can be confusing for users. "_

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: I thought about that users can set these dates themselves instead of it beeing automatically assigned, just like on manual create.

But now that I think abut it it doe snot make any sens for users to have any influence on setting `dateCreated` `lastUpdated` values, it should happen automatically, so I will remove it and ignore these fields

drodzewicz: additionally, even though on manual creation for product association we have date fields for **Date Created** and **Las updated** they don't work

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
awalkowiak: ok

awalkowiak (approved):

PR #3569, opened 2022-10-21
OBPIH-4943 Fix logo display on React pages and fix active menu sectio…

kchelstowski:
…n while on stock list page

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
@@ -155,7 +155,7 @@ class ApiController {
         String minimumExpirationDate = dateFormat.format(grailsApplication.config.openboxes.expirationDate.minValue)
         def logoLabel = grailsApplication.config.openboxes.logo.label
         def pageSize = grailsApplication.config.openboxes.api.pagination.pageSize
-        def logoUrl = "/openboxes/location/viewLogo/${session.warehouse?.id}"
+        def logoUrl = location?.logo ? "/openboxes/location/viewLogo/${session.warehouse?.id}" : null
         def locales = grailsApplication.config.openboxes.locale.supportedLocales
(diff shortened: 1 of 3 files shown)


drodzewicz (changes requested):
I think that `Logo.jsx` component needs some modifications.
Currently it is checking if logo `logoUrl` is not **null** and if it is not it renders the image with given logoURL and if it is **null** it renders default openboxes logo.

I think that a better approach would be to handle onError fallback on image 
```
        <img
          src={logoUrl}
          onError={e => e.target.src = OpenboxesLogo}
          alt="Openboxes"
          width="40"
          height="40"
          role="presentation"
          onClick={() => history.push('/openboxes')}
        />
```
The reason for that is that even if `logoURL` might not be **null**, we can always expect the link to be wrong/broken, so in case of handling onError it would set it to openboxes default logo if there is something wrong with the image

kchelstowski:
@drodzewicz it's nice to know - I hadn't known about `onError` before to be honest. Will try to refactor this. I assume, that instead of giving `alt` when it doesn't find the logo on theoretically good url (not null), it wouldn't give the `alt`, but provide us a default logo? 
And saying that, the `alt` would only display if it doesn't find the image on `logoUrl` and the default logo would not be found?

drodzewicz:
> @drodzewicz it's nice to know - I hadn't known about `onError` before to be honest. Will try to refactor this. I assume, that instead of giving `alt` when it doesn't find the logo on theoretically good url (not null), it wouldn't give the `alt`, but provide us a default logo? And saying that, the `alt` would only display if it doesn't find the image on `logoUrl` and the default logo would not be found?

yes, if `logoUrl` is wrong it would then display defaultLogo then if defaultLogo is wrong then alt text will appear

drodzewicz (approved):

awalkowiak (approved):

PR #3576, opened 2022-10-24
OBGM-20 Reenable external configuration and display it properly

mdpearson:
This PR updates OpenBoxes to look in `~/.grails` and `${catalina.base}` for configuration. It will read `openboxes-config.properties` and `openboxes-config.groovy` files, just like the Grails 2 build. (Hopefully this makes end-users' update process easy.) It also looks in `openboxes.yml` for config, and that's where I've been putting Grails-3-only configuration for my local development. Switching between Grails-1 and Grails-3 should be straightforward since Grails-1 doesn't inspect the `.yml` file.

I tested this by setting conflicting config in different config files and ensuring that overrides were done correctly, and by inspecting the footer on both React and Grails-based pages.

The `branch` field on Bamboo-based builds for some reason always says `feature/upgrade-to-grails-3.3.10` (the branch Bamboo initially checks out) and not the branch it switches to before building. That's an annoying cosmetic issue at this point; the sha and build date are correct. I filed a tech debt ticket (OBGM-368) for the misleading branch field, but the rest of this looks good to me.


jmiranda:
> I tested this by setting conflicting config in different config files and ensuring that overrides were done correctly, and by inspecting the footer on both React and Grails-based pages.

So does this mean that openboxes.yml > openboxes-config.groovy > openboxes-config.properties

> The branch field on Bamboo-based builds for some reason always says feature/upgrade-to-grails-3.3.10 (the branch Bamboo initially checks out) and not the branch it switches to before building. That's an annoying cosmetic issue at this point; the sha and build date are correct. I filed a tech debt ticket (OBGM-368) for the misleading branch field, but the rest of this looks good to me.

I don't understand this one. The automated bamboo builds *should* be building off feature/upgrade-to-grails-3.3.10. Are you talking about when you run a customized build plan for a specific branch? In that case, I understand the annoyance of that (am I actually running the right branch), but would say that's a very low priority so please label that as post-migration.

src/main/groovy/util/ConfigHelper.groovy
```diff
@@ -10,16 +10,37 @@
...
 
+    /**
+     * Return the version of the application, as specified in build.gradle.
+     */
+    static String getAppVersion() {
+        // build.app.version is readable by war files, app.version by local builds
+        return Holders.grailsApplication.metadata.getProperty(
+            'build.app.version',
+            String,
+            Holders.grailsApplication.config.getProperty('app.version')
```
jmiranda, line 26: My head is swimming in the details of what works and what has changed, but I'd like to follow conventions around the runtime and build properties as closely as possible.

I'm not sure if this is correct, but I saw application version retrieved in a more standardized way. Is there a reason we'd want to mess with that? I guess it's possible that we could set a different app version at build time, but do we actually do that?

If not then we should try to stick with the conventional
```
${g.meta(name: "info.app.version")}
```
and
```
Holder.grailsApplication.metadata.applicationVersion
```
https://docs.grails.org/3.3.10/guide/single.html#versioning

Also I think we can still use the g:meta tag lib, correct? I am cool with using the Holders.grailsApplication convention here since this is a utility, but in places where we can use taglibs (i.e. UI / controllers) I'd like to continue using the g:meta.

jmiranda: Ah I see that is still the case. 

https://github.com/openboxes/openboxes/pull/3576/files#diff-598db67b4f4d6a15ec400ad2c40de3da3771b0976985c82f7df2ee1c5c7fd810R8

src/main/groovy/util/ConfigHelper.groovy
```diff
@@ -10,16 +10,37 @@
...
 class ConfigHelper {
 
+    /**
+     * Return the version of the application, as specified in build.gradle.
+     */
+    static String getAppVersion() {
+        // build.app.version is readable by war files, app.version by local builds
+        return Holders.grailsApplication.metadata.getProperty(
+            'build.app.version',
+            String,
```
jmiranda, line 25: Does this need to be String.class? Or does String imply that?

grails-app/views/report/showInventoryReport.gsp
```diff
@@ -158,8 +158,8 @@
                                 </g:link>
                             </li>
                             <li>
-                                <g:link url="https://github.com/openboxes/openboxes/releases/tag/v${g.meta(name:'app.version')}" target="_blank">
-                                    <warehouse:message code="docs.releaseNotes.label" default="Release Notes"/> (${g.meta(name:'app.version')})
+                                <g:link url="https://github.com/openboxes/openboxes/releases/tag/v${ConfigHelper.getAppVersion()}" target="_blank">
```
jmiranda, line 161: Again, if possible it would be best to use info.app.version if that's the convention.

jmiranda (commented):
LGTM; just respond to my comments or make changes and then we can probably merge this one.

src/main/groovy/util/ConfigHelper.groovy
mdpearson: Doesn't need to be String.class anymore, welcome to the new groovy I guess

jmiranda (approved):
LGTM

mdpearson:
> So does this mean that openboxes.yml > openboxes-config.groovy > openboxes-config.properties

Yes, that seemed to make the most sense to me (that a Grails-3 only config file should beat files that Grails 1 reads, and that of the Grails 1 files the groovy file should override the properties file)

But that can be changed if it's not what you expect...

grails-app/views/report/showInventoryReport.gsp
mdpearson: fixed!

src/main/groovy/util/ConfigHelper.groovy
mdpearson: Yeah, `<g.meta/>` "just works" for app version and grails version. git branch info is a little more complicated under bamboo, so I kept the helper function for that one.

PR #3577, opened 2022-10-25
OBPIH-4943 Take default logo from config

kchelstowski:


grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -155,7 +155,8 @@ class ApiController {
         String minimumExpirationDate = dateFormat.format(grailsApplication.config.openboxes.expirationDate.minValue)
         def logoLabel = grailsApplication.config.openboxes.logo.label
         def pageSize = grailsApplication.config.openboxes.api.pagination.pageSize
-        def logoUrl = location?.logo ? "/openboxes/location/viewLogo/${session.warehouse?.id}" : null
+        def logoUrl = location?.logo ? "/openboxes/location/viewLogo/${session.warehouse?.id}" : grailsApplication.config.openboxes.logo.url
+        def defaultLogoUrl = grailsApplication.config.openboxes.logo.url
```
kchelstowski, line 159: it was needed in case `location.logo` exists and the url is "correct", but the image that this url contains does not load, we want to make sure, that we always have access to `defaultLogoUrl` from config, because if we don't have this `defaultLogoUrl`, we only return default logo, if the `location.logo` is null/falsy

src/js/components/Layout/Logo.jsx
```diff
@@ -4,19 +4,29 @@ import PropTypes from 'prop-types';
...
+    if (defaultLogoUrl) {
+      setDefaultLogo(defaultLogoUrl);
+    }
+  }, [defaultLogoUrl]);
+
   const handleError = () => {
-    setImgSrc(OpenboxesLogo);
+    if (defaultLogo) {
+      setImgSrc(defaultLogo);
+    }
```
kchelstowski, line 29: this is why we needed this additional `defaultLogoUrl` from `AppContext` - if the `img` handles an error (url is correct, but it does not find an image on this path, it will set default logo then. If we didn't have `defaultLogoUrl`, it would print `alt` from `img` in this case.

awalkowiak (changes requested):
I think the `deafultLogoUrl` property is not required. Plus please remove the icon that is not used anymore.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -155,7 +155,8 @@ class ApiController {
         String minimumExpirationDate = dateFormat.format(grailsApplication.config.openboxes.expirationDate.minValue)
         def logoLabel = grailsApplication.config.openboxes.logo.label
         def pageSize = grailsApplication.config.openboxes.api.pagination.pageSize
-        def logoUrl = location?.logo ? "/openboxes/location/viewLogo/${session.warehouse?.id}" : null
+        def logoUrl = location?.logo ? "/openboxes/location/viewLogo/${session.warehouse?.id}" : grailsApplication.config.openboxes.logo.url
```
awalkowiak, line 158: Additionally please use `createLink` here and use `contextPath` in the `grailsApplication.config.openboxes.logo.url`

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
awalkowiak: Since we have this logo in the data base, I think there should be no problem with not loading it.

awalkowiak (approved):

PR #3579, opened 2022-10-25
OBPIH-5004 Fix saving and canceling multiple receipt items

awalkowiak:
@jmiranda this line was throwing `collection [...] not processed by flush()` after doing `shipmentItem.addToReceiptItems(receiptItem)` when there was a multi-line cancel case.


jmiranda (approved):
LGTM

PR #3580, opened 2022-10-25
OBPIH-5005 Add option to trigger summary refresh for items and adjust…

awalkowiak:
…ments

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
@@ -33,6 +33,14 @@ import java.text.DecimalFormat
 
 class OrderItem implements Serializable, Comparable<OrderItem> {
 
+    def publishRefreshEvent = {
+        if (order?.isPurchaseOrder && !disableRefresh) {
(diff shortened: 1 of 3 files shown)


grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -74,6 +82,8 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
     Date dateCreated
     Date lastUpdated
 
+    Boolean disableRefresh = Boolean.TRUE
```
jmiranda, line 85: We usually set this to be false by default? Are there too many cases where we want to disable it by default?

jmiranda (approved):
If the answer is yes to my question below, then feel free to merge.

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
awalkowiak: Yeah, I think it is better to disable it for items and adjustments by default. Otherwise, every flush on related objects will trigger the `afterUpdate` for those and if there will be a lot of items, then a lot of duplicated triggers. So I think for now the answer is yes.

PR #3582, opened 2022-10-26
OBPIH-5008 Fix finding locationType by name on demo location data load

drodzewicz:


grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -142,7 +142,7 @@ class LocationDataService {
         if (!location.inventory && !location.parentLocation) {
             location.inventory = new Inventory(['warehouse': location])
         }
-        def locationType = params.locationType ? LocationType.findByName(params.locationType) : null
+        def locationType = params.locationType ? LocationType.findByNameLike(params.locationType + "%") : null
```
awalkowiak, line 145: @drodzewicz could you try this version:
```
LocationType.createCriteria().get {
  or {
    eq('name', params.locationType)
    like('name', params.locationType + "%")
  }
}
```
This would first look for an exact match and then look for a similar one.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3588, opened 2022-10-28
OBPIH-5024 inconsistent returns id for stock movement show page

drodzewicz:


drodzewicz:
I have also added a fix for menu item bug where dropdown sub section would sometimes disapear.
It was caused by additional `margin-top: 1px` which created a gap which caused that cursor to get out of element scope and close the menu.

I have added additional wrapper element that acts as a dropdown container that does not have any gaps and moved actuall menu inside the wrapper where we can have any amount of space between items and sections as we want

awalkowiak (approved):

PR #3592, opened 2022-10-31
OBPIH-5031 Clone dashboard config to not overwrite the default values

awalkowiak:
(cc @kchelstowski)

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -410,7 +410,8 @@ class UserService {
         def fullConfig = grailsApplication.config.openboxes.dashboardConfig
         def mainDashboardId = grailsApplication.config.openboxes.dashboardConfig.mainDashboardId
         def resultConfig = [
-                dashboard: fullConfig.dashboards[id ?: mainDashboardId],
+                // Clone to not overwrite the main config in the line 422
+                dashboard: fullConfig.dashboards[id ?: mainDashboardId].clone(),
                 dashboardWidgets: fullConfig.dashboardWidgets
         ]
         def userConfig = user.deserializeDashboardConfig()


grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -410,7 +410,8 @@ class UserService {
         def fullConfig = grailsApplication.config.openboxes.dashboardConfig
         def mainDashboardId = grailsApplication.config.openboxes.dashboardConfig.mainDashboardId
         def resultConfig = [
-                dashboard: fullConfig.dashboards[id ?: mainDashboardId],
+                // Clone to not overwrite the main config in the line 422
+                dashboard: fullConfig.dashboards[id ?: mainDashboardId].clone(),
```
jmiranda, line 414: I like the idea, but how would we overwrite. grailsApplication.config is stored in memory and can be overwritten? Oh shit, is that why some users see other user's dashboards? 

If so, yikes. Good catch.

jmiranda (approved):
LGTM, just respond to the question.

grails-app/services/org/pih/warehouse/core/UserService.groovy
awalkowiak: @jmiranda Yeah I was confused by this too, but seems that this line `resultConfig.dashboard.personal = personalDashboard` overwrites it for at least "some" time (I am not sure when or if it's refreshed). Long enough to log in on one user and then after logging to another user, that has no personal dashboard modified, to be able to see the previous one's dashboard pulled from the **default** config.

awalkowiak: So it might be a case when there is loggin in/out from one user to another involved. I have not tried this for two different users in two separate sessions at the same time.

kchelstowski:
@awalkowiak  That's nice catch, I'm surprised even though I have blind guess what it could happen, but I don't know how that stuff works in Groovy. 
I know, that at least in Python, if you do something like this with e.g. `List`:
```
x = [1,2,3]
y = x
y.append(4)
print(x)
```
the result of printing `x` is actually `[1,2,3,4]`. In order to create a shallow copy of any array, you can't just create a new variable and make it like in the example above `y = x`, because the **reference** is still to `x`.
Maybe similar thing happened there? But as I said, I don't know how this mechanism works in Groovy though, but it might be the case. 
Anyway, nice catch

PR #3593, opened 2022-10-31
OBPIH-5026 Change naming in the product association import

awalkowiak:


grails-app/domain/org/pih/warehouse/product/ProductAssociation.groovy
```diff
@@ -26,7 +26,7 @@ class ProductAssociation {
     String id
     ProductAssociationTypeCode code
     Product associatedProduct
-    BigDecimal quantity = 0
+    BigDecimal quantity = 1
```
jmiranda, line 29: This feels like a breaking change. What's the reason for this?

jmiranda (commented):
Everything else looks good, but please respond to the comment I made.

grails-app/domain/org/pih/warehouse/product/ProductAssociation.groovy
awalkowiak: To get rid of setting conversion by default to `1:0` (or `0:1`) which makes no sense

awalkowiak: I restored the 0 here, if it might be breaking

PR #3594, opened 2022-10-31
OBGM-363 Configure Sentry for Grails 3, with logback and servlets support.

mdpearson:
This PR enables Sentry support. Just set the `SENTRY_DSN` environment variable in Tomcat's `setenv.sh`.

The plugin we were using hasn't been updated in several years and uses a deprecated Sentry client. Sentry has documentation on interfacing with Grails 3 directly so I yanked the plugin.

- Send fine-grained release information to Sentry to better track when problems appear;
- Provide hints to make Sentry's "Most Relevant" filter work better;
- Provide context from previous log messages with each Sentry event, whenever possible;
- Stop using the abandoned org.grails.plugins:sentry plugin;
- Remove need to inject SentryClient bean;
- Simplify dependencies and update to latest Sentry libraries;
- Experiment with @CompileStatic annotation on a Grails interceptor.


build.gradle
```diff
@@ -522,16 +527,6 @@ dependencies {
...
-    // FIXME replace with com.google.zxing
-    implementation('org.grails.plugins:barcode4j:0.3') {
-        /*
-         * Old; relocated to `org.apache.avalon.framework`, which we
-         * pull in transitively via `org.apache.xmlgraphics:fop`.
-         */
-        exclude group: 'avalon-framework'
-        exclude group: 'org.apache.ant'  // not needed for gradle builds
-    }
-
```
mdpearson, line 534: This block was removed with the barcode4j refactor but somehow snuck back in.

mdpearson:
Here's an [example event](https://sentry.io/organizations/openboxes/issues/3707627817/?project=5221790&query=is%3Aunresolved&referrer=issue-stream). Note that the "most relevant" toggle now highlights code that we can edit -- you can click "full stack trace" to see everything. "Last Seen" and "First Seen" now show SHA's of the latest problematic build and the first one in which the error was caught.

The "breadcrumbs" portion of the page shows log events that happened before the error occurred. In this case, they're not super helpful since I started the app and then went to a page that I knew would immediately fail. With a non-made-up app crash this context should be more informative.

grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
```diff
@@ -13,7 +13,10 @@ import org.pih.warehouse.core.RoleType
  * */
 class RoleInterceptor {
     def userService
-    def dependsOn = [SecurityInterceptor]
```
mdpearson, line 16: `dependsOn` no longer works in Grails 3. the `order` declaration below is its replacement

mdpearson:
I added a fix for OBGM-366 to this PR, in a separate commit. (since this PR increases the work we do in Interceptor objects, it makes it more likely for OBGM-366 to strike.)

grails-app/conf/application.yml
```diff
@@ -34,11 +34,6 @@ grails:
             - ~/.grails/openboxes-config.groovy
             - ~/.grails/openboxes.yml
     plugin:
-        sentry:
-            active: false
-            dsn: https://{PUBLIC_KEY}@sentry.io/{PROJECT_ID}
```
mdpearson, line 39: we now get the dsn from an environment variable, `SENTRY_DSN`

mdpearson:
Superseded by OBGM-366

mdpearson:
Whoops, wrong window!

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
```diff
@@ -9,42 +9,60 @@
  **/
 package org.pih.warehouse
 
-
 import grails.util.Environment
-import io.sentry.SentryClient
-import io.sentry.event.UserBuilder
+import groovy.transform.CompileStatic
+import io.sentry.Sentry
+import io.sentry.protocol.User as SentryUser
```
jmiranda, line 15: I've never understand what these blocks are. Do you have any ideas?

image

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
```diff
@@ -9,42 +9,60 @@
...
 import org.pih.warehouse.core.User
 
+@CompileStatic
 class SentryInterceptor {
-    SentryClient sentryClient
 
-    public SentryInterceptor() {
-        matchAll().except(uri: '/static/**').except(uri: "/info").except(uri: "/health")
+    SentryInterceptor() {
+        matchAll().except(uri: '/static/**').except(uri: '/info').except(uri: '/health')
```
jmiranda, line 23: Can you turn this formatter off? These are the "minor" changes I want to avoid re: rebase conflicts. This isn't a problem here because interceptors are new to Grails 3, but I just keep seeing changes like this all over and knowing how painful rebasing is, this is just adding to our death by a thousand cuts rebasing problem.

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
```diff
@@ -9,42 +9,60 @@
...
+                String locationId = (session.getAttribute('warehouse') as Location)?.id
+                User user = User.get(userId)
+                SentryUser sentryUser = new SentryUser()
+                sentryUser.with {
+                    email = user?.email
+                    id = userId
+                    username = user?.username
+                    data = [
+                        environment: Environment.current.name,
+                        locale: user?.locale?.toString() ?: 'unset',
```
jmiranda, line 52: Is sentry expecting the unset value?

src/main/groovy/org/pih/warehouse/monitoring/SentryServletContainerInitializer.groovy
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.LoggerFactory
+
+import javax.servlet.ServletContainerInitializer
+import javax.servlet.ServletContext
+import javax.servlet.ServletException
+
+/**
+ * Configure the sentry-servlet library to send URL/user-agent data to Sentry.
+ *
+ * https://docs.sentry.io/platforms/java/guides/servlet/#configure
```
jmiranda, line 14: Can you elaborate on what this is for / why we need it? As a comment to the code if possible.

jmiranda (approved):

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
mdpearson: @jmiranda my guess is that GitHub simply doesn't like `import ... as ...`. Those spaces are no different than the other spaces on the line, so it doesn't look like a "hey, you're using a weird encoding here"

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
mdpearson: yeah, for sure, I'll clean this up and squash. I try to prevent "editor tidying" from creeping in and must have missed this one

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
mdpearson: `'unset'` is not a magic word, if that's what you're asking; but you can't specify a `null` value to the data dictionary. Another approach would be

```
if locale: user?.locale?.toString():
    data['locale'] = user?.locale?.toString()
```

i.e., simply don't provide the key if the value can't be ascertained. I could go either way. Thoughts?

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
mdpearson: Fixed in 23788a183.

grails-app/controllers/org/pih/warehouse/SentryInterceptor.groovy
mdpearson: I went with the alternate approach I described above. If we don't know a value, we just drop the key, instead of saying "unset". Fixed in 23788a183.

src/main/groovy/org/pih/warehouse/monitoring/SentryServletContainerInitializer.groovy
mdpearson: I added an explanatory comment as follows in 23788a183:

```
/**
 * Configure Sentry at application startup.
 *
 * Sentry integration occurs in three places. This class, which Sentry
 * finds via a file named javax.servlet.ServletContainerInitializer, sets
 * global parameters, including information about the release. The init
 * call also (magically) makes it possible for Sentry to retrieve url
 * and user-agent information from the application server.
 *
 * See also SentryInterceptor, which provides Sentry with additional
 * data at each page load, and logback.xml, for logging integration.
 *
 * https://docs.sentry.io/platforms/java/guides/servlet/#configure
 */
```

PR #3595, opened 2022-11-01
OBGM-366 Fix Interceptor session bugs by refactoring AuthService (also fixes OBPIH-4325).

mdpearson:
This PR is aimed against another PR to ease review. I'll merge it into the appropriate feature branch once approved.

I moved some of the session logic from `SecurityInterceptor` to `AuthService` so that grails does its usual transactional management for service methods.

Then I updated all consumers of AuthService to call getters instead of accessing thread local objects directly, which is a little cleaner, and while I was doing that I put in a fix for OBPIH-4325 (we don't need withNewSession when we get User info from a transactional service).

Finally I fixed a Grails-1-ism in RoleInterceptor and made sure that Interceptors that depend on AuthService.currentUser run after the two interceptors that set it.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
+++ b/grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
@@ -13,7 +13,10 @@ import org.pih.warehouse.core.RoleType
  * */
 class RoleInterceptor {
     def userService
-    def dependsOn = [SecurityInterceptor]
+
+    // this interceptor depends on SecurityInterceptor
+    int order = LOWEST_PRECEDENCE
+
     def static changeActions = ['delete', 'create', 'add', 'process', 'save',
                                 'update', 'importData', 'receive', 'showRecordInventory', 'withdraw', 'cancel', 'change', 'toggle', 'exportAsCsv']
     def static changeControllers = ['createProductFromTemplate']
--- a/grails-app/services/org/pih/warehouse/auth/AuthService.groovy
+++ b/grails-app/services/org/pih/warehouse/auth/AuthService.groovy
@@ -9,12 +9,41 @@
  **/
 package org.pih.warehouse.auth
 
+import grails.gorm.transactions.Transactional
+import groovy.transform.CompileStatic
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.User
 
+@CompileStatic
+@Transactional(readOnly = true)
 class AuthService {
 
-    static ThreadLocal<User> currentUser = new ThreadLocal<User>()
-    static ThreadLocal<Location> currentLocation = new ThreadLocal<Location>()
+    private static ThreadLocal<User> threadLocalUser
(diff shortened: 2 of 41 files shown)


grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
```diff
@@ -13,7 +13,10 @@ import org.pih.warehouse.core.RoleType
  * */
 class RoleInterceptor {
     def userService
-    def dependsOn = [SecurityInterceptor]
```
mdpearson, line 16: one of the things I don't like about conveniences like this is that Grails never tells you "I don't look at this magic field anymore." This is a filter parameter from Grails 1; interceptors have a global ordering, set immediately below.

grails-app/services/org/pih/warehouse/auth/AuthService.groovy
```diff
@@ -14,7 +14,32 @@ import org.pih.warehouse.core.User
...
+    private static ThreadLocal<User> threadLocalUser
+    private static ThreadLocal<Location> threadLocalLocation
 
+    void setCurrentUser(User user) {
+        if (!threadLocalUser) {
+            threadLocalUser = new ThreadLocal<User>()
+        }
+        // previous implementation had an explicit get call
+        // threadLocalUser.set(user?.id ? User.get(user.id) : null)
+        threadLocalUser.set(user)
```
mdpearson, line 26: I'm not sure if we need the `get(id)` call (is it to guard against accidentally referring to user from a different db session?). In my testing, not having it hasn't caused any problems, but if we called `get(id)` (see the original implementation, in red, at [SecurityInterceptor](https://github.com/openboxes/openboxes/pull/3595/files#diff-5072e24fd0145afc6923503785dde77c36cd285094e89513572c0f609e1579b7)) for a specific use case I didn't happen to check, I'd love to know about it.

mdpearson: Turns out we did!

grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
jmiranda: Exactly what i was thinking.

jmiranda (approved):
I think this look right. There are places where I felt we didn't need to inject authService (i.e. InventoryService and StockMovementService), but I guess it was fine since we would have needed to change the code from AuthService.currentUser.get() to AuthService.currentUser anyway.

However, in general we should try to fix bugs with the smallest number of code changes possible, even if it means keeping ugly code like AuthService.currentUser.get(). Eventually we'll be porting over to Spring Security so this code will change again after the migration but I want to make sure we don't create an even larger world of hurt when rebasing from Grails 1 to Grails 3 (in case that similar code is changed in Grails 1).

jmiranda:
@awalkowiak could you take a look at this as well.

awalkowiak (approved):

PR #3596, opened 2022-11-01
OBGM-65 Load static images consistently

mdpearson:
Some context in [this comment](https://pihemr.atlassian.net/browse/OBGM-65?focusedCommentId=137233) on the ticket.

### DIFF
--- a/grails-app/views/dashboard/_activitySummary.gsp
+++ b/grails-app/views/dashboard/_activitySummary.gsp
@@ -51,7 +51,7 @@
             },
             "oLanguage": {
                 "sZeroRecords": "No records found",
-                "sProcessing": "Loading ... <img alt='spinner' src=\"${resource(dir:'images/spinner.gif')}\" />"
+                "sProcessing": "Loading ... <img alt='spinner' src=\"${resource(dir: 'images', file: 'spinner.gif')}\" />"
             },
             "aLengthMenu": [
                 [5, 10, 25, 100, 1000, -1],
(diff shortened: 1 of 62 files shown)


jmiranda (approved):
LGTM; @awalkowiak can you take a look to make sure this looks ok to you.

grails-app/views/dashboard/_activitySummary.gsp
```diff
@@ -65,7 +65,16 @@
...
+                /*
+                 * We can't provide aData as a `file:` argument to resource() here,
+                 * as aData is undefined before the client browser executes this
+                 * file. The GSP parts of this file get rendered on the server
+                 * well before then, so we build the file's dirname in GSP and
+                 * its basename in JavaScript then concatenate them. :-( OBGM-65
+                 *
+                 * https://stackoverflow.com/questions/33325965/grails-gsp-accessing-variables-within-script#comment54459606_33330080
+                 */
+                $('td:eq(0)', nRow).html(`<img src="${resource(dir: '/images/icons/silk')}/` + aData['type'] + '.png" />');
```
jmiranda, line 77: Sorry you had to look into that. In the future, when you hit things like this let's make sure it's code we care about keeping.

awalkowiak: Since there is already used string interpolation, then it could be:
```
$('td:eq(0)', nRow).html(`<img src="${resource(dir: '/images/icons/silk')}/${aData['type']}.png" />`);
```
but I think the version with adding string looks a bit cleaner.

awalkowiak (approved):

PR #3603, opened 2022-11-04
OBPIH-5030 Ability to export all products tagged by a given tag

jmiranda:


grails-app/views/tag/create.gsp
```diff
@@ -19,66 +19,72 @@
...
+                                <tr class="prop">
+                                    <td valign="top" class="name">
+                                        <label for="tag"><warehouse:message code="tag.tag.label" default="Tag" /></label>
+                                    </td>
+                                    <td valign="top" class="value ${hasErrors(bean: tagInstance, field: 'tag', 'errors')}">
+                                        <g:textField name="tag" size="80" class="text medium" value="${tagInstance?.tag}" />
+                                    </td>
+                                </tr>
 
+%{--                                <tr class="prop">--}%
```
awalkowiak, line 41: Is the commented-out part still required?

jmiranda: Probably not. I just wanted to be able to quickly revert if someone complains about it.

PR #3606, opened 2022-11-04
OBPIH-5006 Add collapsable mobile menu for react and gsp pages 

drodzewicz:


kchelstowski (approved):

awalkowiak (changes requested):
Code-wise I think it looks ok, but locally for me, the initial megamenu from gsp template takes around 20s to build. Plus there is a tiny difference in both headers height.

kchelstowski (approved):

awalkowiak (approved):

PR #3611, opened 2022-11-09
OBPIH-5043 Validate for duplicate associations

drodzewicz:


grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -28,8 +29,32 @@ class ProductAssociationDataService {
...
             productAssociationInstance.associatedProduct = Product.findByProductCode(params['associatedProduct.productCode'])
             params['product.id'] =  productAssociationInstance.product?.id
+            params['product.name'] =  productAssociationInstance.product?.name
             params['associatedProduct.id'] =  productAssociationInstance.associatedProduct?.id
+            params['associatedProduct.name'] =  productAssociationInstance.associatedProduct?.name
 
+            params.hasMutualAssociation = Boolean.valueOf(params.hasMutualAssociation as String) || (params.hasMutualAssociation as String)?.equalsIgnoreCase("yes")
+
+            // check for duplicate association entries in the file
+            def indexOfDuplicate = listOfValidatedProductAssociationParams.findIndexOf{it ->
```
awalkowiak, line 39: Is the `findIndexOf` done on an empty list, or am I missing something here?

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -28,8 +29,32 @@ class ProductAssociationDataService {
...
+                if (params.hasMutualAssociation || it.hasMutualAssociation) {
+                    return  (
+                            params['product.id'] == it['product.id'] &&
+                            params['associatedProduct.id'] == it['associatedProduct.id'] &&
+                            params['code'] == params['code']
```
awalkowiak, line 44: Seems like this (`params['code'] == params['code']`) will be always true

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -65,16 +90,42 @@ class ProductAssociationDataService {
...
+                            eq("associatedProduct", productAssociationInstance.product)
+                            eq("code", productAssociationInstance.code)
+                        }
+                    }
+                }
+                if (existingAssociations) {
+                    if (existingAssociations.any{ it.mutualAssociation }) {
+                        command.errors.reject("Row ${index + 1}: Same two-way association already exists")
+                    } else {
+                        command.errors.reject("Row ${index + 1}: One-way of the association already exists")
```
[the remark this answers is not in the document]
kchelstowski: or nvm, now I realized the context - you wouldn't want to display this error from else if there is not any association

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -28,8 +29,32 @@ class ProductAssociationDataService {
...
+            def indexOfDuplicate = listOfValidatedProductAssociationParams.findIndexOf{it ->
+                if (params.hasMutualAssociation || it.hasMutualAssociation) {
+                    return  (
+                            params['product.id'] == it['product.id'] &&
+                            params['associatedProduct.id'] == it['associatedProduct.id'] &&
+                            params['code'] == params['code']
+                        ) || (
+                            params['product.id'] == it['associatedProduct.id'] &&
+                            params['associatedProduct.id'] == it['product.id'] &&
+                            params['code'] == params['code']
```
awalkowiak, line 48: Same here

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -28,8 +29,32 @@ class ProductAssociationDataService {
...
+            params['product.name'] =  productAssociationInstance.product?.name
             params['associatedProduct.id'] =  productAssociationInstance.associatedProduct?.id
+            params['associatedProduct.name'] =  productAssociationInstance.associatedProduct?.name
 
+            params.hasMutualAssociation = Boolean.valueOf(params.hasMutualAssociation as String) || (params.hasMutualAssociation as String)?.equalsIgnoreCase("yes")
+
+            // check for duplicate association entries in the file
+            def indexOfDuplicate = listOfValidatedProductAssociationParams.findIndexOf{it ->
+                if (params.hasMutualAssociation || it.hasMutualAssociation) {
+                    return  (
```
awalkowiak, line 41: This entire `or` looks weird are you sure it is proper?

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
```diff
@@ -28,8 +29,32 @@ class ProductAssociationDataService {
...
+                            params['code'] == params['code']
+                        ) || (
+                            params['product.id'] == it['associatedProduct.id'] &&
+                            params['associatedProduct.id'] == it['product.id'] &&
+                            params['code'] == params['code']
+                        )
+                }
+                return params['product.id'] == it['product.id'] &&
+                       params['associatedProduct.id'] == it['associatedProduct.id'] &&
+                       params['code'] == params['code']
```
awalkowiak, line 53: Same as above

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
awalkowiak, line 39: Add a space before and after the `{`

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: yep this is a typo

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: on line 128 I add visited/checked association to this list and later check for duplicates
```
listOfValidatedProductAssociationParams.add(params)
```

grails-app/services/org/pih/warehouse/data/ProductAssociationDataService.groovy
drodzewicz: It works properly but I agree it looks ugly.
the problem I had here was that I had to detect not only duplicates but also a case where we have two-way and one-way entries of the same association
```
Association_1(product: A, associatedProduct: B, type: Type1, hasMutualAssociation: false )
Association_2(product: A, associatedProduct: B, type: Type1, hasMutualAssociation: true )
```
```
Association_1(product: A, associatedProduct: B, type: Type1, hasMutualAssociation: false )
Association_2(product: B, associatedProduct: A, type: Type1, hasMutualAssociation: true )
```

jmiranda (commented):
I am not in a good mindset to really dig into this. My one comment would be ... add very descriptive comments about what we're doing here. Going so far as to describe the anatomy or examples the expected input so we can see what and why we're doing what we're doing. 

One of your comments below does a good job explaining the what or why. So that needs to go into the code as a comment. I would go even further and document all of the scenarios that might come up.

PR #3621, opened 2022-11-14
OBPIH-5078 Fix request feature for location without manage_inventory …

kchelstowski:
…and with submit_request, but not having type Ward

I additionally added "enum" for activity codes which I think we should consider using in other places instead of hard-coding supported activities everywhere. What do you think?


src/js/utils/supportedActivitiesUtils.js
```diff
@@ -0,0 +1,8 @@
+import ActivityCode from 'consts/activityCode';
+
+const isRequestFromWard = (currentLocationId, destinationId, supportedActivities) =>
```
awalkowiak, line 3: Technically now it can be not from `Ward` too

awalkowiak (changes requested):

src/js/utils/supportedActivitiesUtils.js
awalkowiak: @kchelstowski at least add a comment explaining that `Ward` in this context does not mean the location type but specific set of supported activities

awalkowiak: @kchelstowski nvm, I pushed the comment

awalkowiak (approved):

PR #3631, opened 2022-11-17
OBPIH-5122 Fix wrong inventory stockcard redirect after mark as expired

drodzewicz:


drodzewicz:
This was a very interesting/weird bug so I want to share it with you

## The problem
When user would mark product as expired, at last step instead of redirecting to an inventory stock card page _(in case of multiple products it would redirect to first product)_ it would redirect to dashboard
Later when  trying to enter product view page via global-search _(entering product code and clicking ENTER)_ it would do the same thing - instead of going to product stock card it would redirect to dashboard.

## What I managed to investigate
When using global search it hits endpoint with `controller: DashboardController, action: globalSearch`.
and if the provided `searchTerm` finds a product it redirects to inventoryItem -> showStockCard
```
redirect(controller: "inventoryItem", action: "showStockCard", id: product.id)
```
When I inspected the `product.id`  it was a normal string  _(eg. "10001")_

Then I followed to `inventoryItemController` and `showStockCard` and inspected `params` and I found out that value of `id` has extra brackets around the `id `which is why it would redirect to dashboard - because there is no product of id `"[10001]"`

##  My assumption 
In `saveDebitTransaction` which is cachable `@CacheFlush("inventoryBrowserCache")` after user marks product as expired it redirected to showStockCard with  `id = productIds`.

```
if (productIds.size() > 1) {
    redirect(controller: "inventoryItem", action: "showStockCard", id: productIds[0])
 } else {
    redirect(controller: "inventoryItem", action: "showStockCard", id: productIds)
 }
```
in case of `productIds.size() == 1` it passes the whole list of id's like:
```
    redirect(controller: "inventoryItem", action: "showStockCard", id: ["10001"])
```

I am not entirely sure how caching works in grails but it seems like later when using global-search it uses that cached link to redirect to showStockCard but this is just my assumption.

I understand why it would redirect to dashboard after ` redirect(controller: "inventoryItem", action: "showStockCard", id: productIds)` but I completely have no idea why it affect global-search methods

jmiranda:
The @Cacheable annotation here is to flush the cache (i.e. refresh it) whenever there's a new transaction. This may not be necessary any longer because we don't use the old dashboard. But we can create a separate tech debt issue for that (in fact it should be in the dead code ticket).

I recommend testing this change by expiring items for a single product 
https://obdev.pih-emr.org/openboxes/inventory/createExpired?product.id=10130

as well as multiple products. 
https://obdev.pih-emr.org/openboxes/inventory/createExpired?product.id=10130&product.id=10027

I don't think there's a way to get to a URL like the second example from within the app (we removed the functionality from the Inventory Browser) so you'll need to add the extra product.id request parameter yourself. You can include as many as you'd like.

jmiranda:
Nice work on the investigation / write up. It would be good to add that to the ticket. Or at least add a comment that links to your PR comment.

awalkowiak (approved):

PR #3635, opened 2022-11-18
OBPIH-5136  Enable deletion of a stock request

drodzewicz:
- when deleting request from ward redirect to dashboard
- set minimum required role for request deletion to Assistant

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
@@ -335,6 +335,11 @@ class StockMovementController {
                 return
             }
         }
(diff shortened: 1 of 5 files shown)


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -335,6 +335,11 @@ class StockMovementController {
                 return
             }
         }
+        if (currentLocation.isWard()) {
+            redirect(uri: "/dashboard")
+            return
+        }
+
```
kchelstowski, line 342: I believe you shouldn't check the location type, but the supported activities: has submit_request and doesn't have manage_inventory

kchelstowski (changes requested):

kchelstowski (approved):

awalkowiak (approved):

PR #3640, opened 2022-11-18
OBPIH-5138 HOTFIX: QOH on blur updating quantity requested to 0 over and over again

kchelstowski:
The problem was, that whenever we've gone from QOH, the quantityRequested (Needed Qty) was updated to 0 (if there was not any demand). I think the problem has not been seen before, because whenever we were adding a new line, we were not going from QOH from previous product to QOH of next product and we were always going from left to right.
The problem was happening both for keyboard and mouse, but when switching the cells by mouse, the quantity was not updating in the table until you added a new line (but when console.loging the values, they were 0).


src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -344,9 +344,13 @@ const REQUEST_FROM_WARD_STOCKLIST_FIELDS_PUSH_TYPE = {
         }) => ({
           onBlur: () => {
             const valuesWithUpdatedQtyRequested = values;
-            valuesWithUpdatedQtyRequested.lineItems[rowIndex].quantityRequested =
-              values.lineItems[rowIndex].quantityAllowed - fieldValue >= 0 ?
-                values.lineItems[rowIndex].quantityAllowed - fieldValue : 0;
+            const lineItem = valuesWithUpdatedQtyRequested.lineItems[rowIndex];
```
awalkowiak, line 347: Actually, these could be moved to a separate method and passed like the `updateRow`.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3642, opened 2022-11-22
OBPIH-5013 generate megamenu gsp from configuration

drodzewicz:


grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -149,13 +149,24 @@ class DashboardController {
...
+        }
+
+        Map menuConfig = grailsApplication.config.openboxes.megamenu;
+        menuConfig = menuConfig.findAll { key, val -> key != "configuration" }
+        User user = User.get(session?.user?.id)
+
+        if (userService.hasHighestRole(user, session?.warehouse?.id, RoleType.ROLE_AUTHENTICATED)) {
+            menuConfig = grailsApplication.config.openboxes.requestorMegamenu;
+        }
+        List translatedMenu = megamenuService.buildAndTranslateMenu(menuConfig, user, location).findAll { !it.isEmpty() }
```
jmiranda, line 165: Can you explain the filter here? And maybe add a comment.

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -149,13 +149,24 @@ class DashboardController {
         render results as JSON
     }
 
-    @Cacheable("megamenuCache")
     def megamenu = {
+        Location location = Location.get(session.warehouse?.id)
+        if (!location.supports(ActivityCode.MANAGE_INVENTORY) && location.supports(ActivityCode.SUBMIT_REQUEST)) {
```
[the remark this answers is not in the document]
drodzewicz: We still don't really have an idea on what it should be called so I believe we should create a separate tech debt ticket to refactor all instances in the application that use this logic and rename it into something more descriptive.

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
drodzewicz: There was a case where if menu section didn't have following attributes: `href`, `subsections`, `menuItems` it would return an empty object `return [:]` 
In my latest commit I figured that instead of filtering them out I should probably prevent the empty objects from being added to the megamenu list

**MegamenuService.groovy >  buildAndTranslateSections**
```
 def translatedSections = buildAndTranslateSections(value, key, user, location)
 if (translatedSections) {
    parsedMenuConfig << translatedSections
}
```

awalkowiak (approved):

PR #3643, opened 2022-11-22
OBPIH-5165 Adjust badge count total to count only selected fields that user can select

drodzewicz:


src/js/components/Filter/FilterForm.jsx
```diff
@@ -56,8 +56,14 @@ const FilterForm = ({
...
   // Calculate which object's values are not empty
   const countFilled = (values) => {
-    setAmountFilled(Object.values(values)
-      .filter((value) => {
+    setAmountFilled(Object.entries(values)
+      .filter(([key, value]) => {
+        // Ignore accounting for filter that is specified in "ignoreClearFilters"
+        if (ignoreClearFilters.includes(key)) return false;
+        // Ignore filter that is not specified in filterFields config
+        // and that is not a search field
```
kchelstowski, line 64: what are examples of filters that are not specified in `filterFields`? Because I can't remind myself of such case?

drodzewicz: This was the case where the badge counter would count filters such as `direction=INBOUND` on inbound pages or for requests `direction=OUTBOUND&sourceType=ELECTRONIC`. 
These filters are not specified in the `filterFields`

kchelstowski: Aren’t they added to the filterFields anyway when building defaultFilterValues? 
In the place, where you are looking on query params and adding them to filterValues?

drodzewicz: The query params filters are added to filterValues but filterFields stays unchanged.
FilterFields is like a configuration object that specifies what filters/components to load and we are never changing it.

kchelstowski: Oh, I see, I was thinking about filterValues all the time. I didn’t mean you had done something wrong, but just wanted to know, what is going on there. 
If it’s filterFields then I’m fine with it now

PR #3652, opened 2022-11-23
OBPIH-5163 Users with Browser permissions can't see the PO list page

kchelstowski:
Confirmed with Katarzyna, that delete should be still only for >= Assistants

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1134,7 +1134,7 @@ openboxes.security.rbac.rules = [
         [controller: 'glAccountType', actions: ['delete'], accessRules: [minimumRequiredRole: RoleType.ROLE_SUPERUSER]],
         [controller: 'preferenceType', actions: ['delete'], accessRules: [minimumRequiredRole: RoleType.ROLE_SUPERUSER]],
         [controller: 'purchaseOrderApi', actions: ['delete'], accessRules: [ minimumRequiredRole: RoleType.ROLE_ASSISTANT]],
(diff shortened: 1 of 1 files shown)


grails-app/conf/Config.groovy
```diff
@@ -1133,7 +1133,7 @@ openboxes.security.rbac.rules = [
         [controller: 'glAccount', actions: ['delete'], accessRules: [minimumRequiredRole: RoleType.ROLE_SUPERUSER]],
         [controller: 'glAccountType', actions: ['delete'], accessRules: [minimumRequiredRole: RoleType.ROLE_SUPERUSER]],
         [controller: 'preferenceType', actions: ['delete'], accessRules: [minimumRequiredRole: RoleType.ROLE_SUPERUSER]],
-        [controller: 'purchaseOrderApi', actions: ['*'], accessRules: [ minimumRequiredRole: RoleType.ROLE_ASSISTANT]],
+        [controller: 'purchaseOrderApi', actions: ['delete'], accessRules: [ minimumRequiredRole: RoleType.ROLE_ASSISTANT]],
```
awalkowiak, line 1136: Should `rollback` action also be available to everyone?

kchelstowski: Rollback depends on Approver role and we check it on the frontend side:
```
const rollbackHandler = (id) => {
    if (!isUserApprover) {
      Alert.error(translate(
        'react.default.errors.noPermissions.label',
        'You do not have permissions to perform this action',
      ));
      return;
    }
    (....)
 ```

awalkowiak: But through API you can now do this having any kind of role. So it also should also be here if you need a supplementary role Approver.

awalkowiak (approved):

PR #3659, opened 2022-11-25
OBPIH-5059 Newly created data are not visible on dropdowns in the app

kchelstowski:


grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -57,7 +57,6 @@ class SelectTagLib {
     def requisitionService
     def organizationService
 
-    @Cacheable("selectCategoryCache")
```
awalkowiak, line 60: I am not sure if this is a good idea. How looks the performance of this tag now?

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -130,7 +129,6 @@ class SelectTagLib {
         out << g.select(attrs)
     }
 
-    @Cacheable("selectTagsCache")
```
awalkowiak, line 133: Same as above

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -143,7 +141,6 @@ class SelectTagLib {
         out << g.select(attrs)
     }
 
-    @Cacheable("selectCatalogsCache")
```
awalkowiak, line 146: Same as above

grails-app/views/inventory/_filters.gsp
```diff
@@ -11,24 +11,39 @@
...
-				<g:selectTags name="tags" noSelection="['':'']"
-							  value="${commandInstance.tags*.id}"
-							  data-placeholder="Select a tag"
-							  class="chzn-select-deselect"/>
+				<g:selectTagsViaAjax id="tags"
+										 name="tags"
+										 class="ajaxSelect2"
+										 noSelection="['':'']"
+										 multiple="true"
+					                     data-placeholder="Select a tag"
```
kchelstowski, line 30: I don't know why the indents look so weird here, locally they are one under another

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -138,6 +138,8 @@ class InventoryController {
         command.location = command?.location ?: Location.get(session.warehouse.id)
         def category = params.categoryId ? Category.get(params.categoryId) : productService.getRootCategory()
         command.category = category?.id ? category : null
+        command.catalogs = params.catalogs ? command.catalogs : null
+        command.tags = params.tags ? command.tags : null
```
kchelstowski, line 142: it was needed to be added, because when you click "Delete all" (x) on this ajax select, the request is proceeding with for example: `catalogs=&....`, so an Array with one null element is passed, which made the response having no results - it got fixed just after another fetch, so if `params.catalogs` is empty string, we want to make the `command.catalogs` as `null`, not `[null]`

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
jmiranda: Yeah the better approach would be to flush the cache when we add new objects.

jmiranda: The easiest way to do that would be to add @CacheFlush annotation to methods used for CRUD operations on these objects (either on controller actions or service methods).

```
@CacheFlush("selectCategoryCache")
def saveCategory = {
    ...
}
```

Another (debatably better) approach would be to use the GORM / hibernate event handlers in the domain classes to publish a cache flush event.

```
class Category { 
    // Not sure if you can do this "multi variable declaration" but ... 
    def afterInsert, afterUpdate = { 
        publishEvent(FlushCacheEvent(this))
    }
}
```
The event class
```
class FlushCacheEvent extends ApplicationEvent { 
    FlushCacheEvent(Category source) { 
        super(source)
    }
    ...
}
```
And then the handler
```
class FlushCacheEventService {

    def springcacheService

    void onApplicationEvent(FlushCacheEvent event) { 
        if (event.source instanceOf Category) {
            springCacheService.flush("selectCategoryCache")
        } 
        // or more generically
        springCacheService.flush("select${event.source.className}Cache")

    }
}
```

jmiranda: I'm ok with either. If we're tight on time, go with the annotation approach.

jmiranda (commented):
As mentioned in the comments, we should use the springcache annotation approach if we're tight on time. The only problem with the CacheFlush annotation is that you may not catch all of the places where these objects are created / updated i.e. imported data, API calls, etc. I'm not too concerned about this as long as we capture most of the CRUD operations. One saving grace is that we can use the Flush Cache feature to flush all of the caches.

awalkowiak (approved):

PR #3660, opened 2022-11-25
OBPIH-5170 Validate quantity per uom on order item

awalkowiak:


grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -97,7 +98,7 @@ class ErrorsController {
             response.status = 400
             BeanPropertyBindingResult errors = request?.exception?.cause?.errors
             def errorMessages = errors.allErrors.collect {
-                return messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), null)
+                return messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), localizationService.getCurrentLocale())
```
awalkowiak, line 101: @jmiranda Passing `null` here was causing the NullpointerException in the spring's getMessage(). This is "temporary" fix. I add a topic regarding this to the Monday's TH.

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -750,19 +750,21 @@ class OrderController {
             orderItem.productSupplier = productSupplier
         }
 
+        if (!order.save(flush:true)) {
+            throw new ValidationException("Order is invalid", order.errors)
```
awalkowiak, line 754: @jmiranda I moved this out of the try-catch, to get the advantage of ErrorController's `handleValidationErrors` (which was missed within the `catch`)

PR #3662, opened 2022-11-25
OBPIH-5076 Add missing translations or labels

drodzewicz:


drodzewicz:
I have noticed that crwodin doesn't recognize labels that are rendered using `<g:message` only when I changed the tagLib to `<warehouse.message` it recognized it and highlighted the label

PR #3669, opened 2022-11-29
OBPIH-5171 Fix pressing enter when cursor placed in filter field triggers clear filter and page refresh

drodzewicz:


drodzewicz:
To prevent this bug from happening I had to somehow add `preventDefault()` on input when clicking enter but only when there are no options to choose from because otherwise if I just left `preventDefault()` without checking for `options.length` it would prevent user from selecting options using keyboard.
Other problem was how to get number of `options` and unfortunately since this is `FilterSelectField`  component is an async select  options in `props` always resulted in an empty list  so I had to extract a proper options list from one of the custom component which was `ValueContainer`.

awalkowiak (approved):

PR #3672, opened 2022-11-29
OBPIH-5076 Add missing translation labels on stock movement

drodzewicz:


awalkowiak (commented):
@jmiranda do we want to go with `g:message` -> `warehouse:message` change, or we're waiting with the `g:message` investigation?

jmiranda:
> @jmiranda do we want to go with `g:message` -> `warehouse:message` change, or we're waiting with the `g:message` investigation?

I would prefer to wait on the investigation so we don't needlessly change this all around.

drodzewicz:
I have rolled back my changes where I changed `g:message` into `warehouse:message` since we have kind of figured out why it wasn't translating labels.

grails-app/views/returns/show.gsp
```diff
@@ -130,7 +130,12 @@
                                 <g:message code="stockMovement.status.label"/>
                             </td>
                             <td class="value">
-                                <format:metadata obj="${stockMovement?.shipment?.status?:stockMovement?.statusCode}"/>
+                                <g:if test="${stockMovement?.shipment?.status}">
```
awalkowiak, line 133: @drodzewicz Is there a reason behind this change into `if-else`? You could add the `?.code` to the first part and have the same result:
```<format:metadata obj="${stockMovement?.shipment?.status?.code ?: stockMovement?.statusCode}"/>```

File: grails-app/views/stockMovement/show.gsp [hunk 2]
```diff
@@ -166,7 +166,12 @@
                                 <warehouse:message code="stockMovement.status.label"/>
                             </td>
                             <td class="value">
-                                <format:metadata obj="${stockMovement?.shipment?.status?:stockMovement?.requisition?.status }"/>
+                                <g:if test="${stockMovement?.shipment?.status}">
```
awalkowiak, line 169: Same question as above

PR #3673, opened 2022-11-30
OBPIH-5171 Fix filter refreshing on clicking enter when focused on select filter

drodzewicz:


src/js/components/form-elements/FilterSelectField.jsx
```diff
@@ -108,41 +108,21 @@ IndicatorsContainer.propTypes = {
 };
 
 const FilterSelectField = (props) => {
-  const renderInput = ({ className, ...attributes }) => {
```
drodzewicz, line 111: here I have rolled back my changes from previous pull requests that was supposed to fix this bug but apparently it didn;t work for some cases https://github.com/openboxes/openboxes/pull/3669
After further investigation I managed to find the exact problem which was having submit buttons in the form.

awalkowiak (changes requested):
@drodzewicz these changes should go to the `release/0.8.20-hotfix1` branch instead into develop

awalkowiak (commented):
lgtm

PR #3677, opened 2022-12-01
OBPIH-4940 Format number based on current locale for total amount on PO list pages

drodzewicz:


src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -437,13 +438,15 @@ const PurchaseOrderListTable = ({
     }
   };
 
+  const totalAmount = () => `${translate('react.purchaseOrder.totalAmount.label', 'Total amount')}: ${totalPrice.toLocaleString([locale, 'en'])} ${currencyCode}`;
```
jmiranda, line 441: Do we want to hard-code the locale here?

jmiranda: Oh sorry, that's probably a default value in the case that locale is null or undefined.

jmiranda (approved):

awalkowiak (approved):

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
drodzewicz: yes, the second argument is a fallback/default locale

PR #3678, opened 2022-12-01
OBPIH-5191 Refactor PDF internals

mdpearson:
This PR removes the `core-renderer` library (last updated in 2010) and `iText` (2008), replacing them with the goofily-named [flying-saucer](https://github.com/flyingsaucerproject/flyingsaucer) library and the less-goofily-named [OpenPDF](https://github.com/LibrePDF/OpenPDF).

`iText` has a number of bug fixes we can't use for licensing reasons, and `core-renderer` always shows up in the stack traces we see when fonts fail to load. With any luck, updating our dependencies will prevent PDF problems in the future, but -- since I never could reproduce the issue in the first place, I can't 100% guarantee that...

I updated the markup used for PDF generation because the newer libraries follow CSS more closely than the old ones did. The changes almost entirely concern changes to word-wrapping behavior. I also fixed some inconsistent rendering issues related to the `border-collapse` property.


grails-app/views/picklist/_page.gsp
```diff
@@ -88,26 +88,26 @@
                     <td class="middle">
                         <g:if test="${j==0}">
                             <g:if test="${requisitionItem?.parentRequisitionItem?.isSubstituted()}">
-                                <div class="canceled">
+                                <wordwrap:div class="canceled">
```
jmiranda, line 91: Why can't we handle this with CSS?

grails-app/views/picklist/_print.gsp
```diff
@@ -97,11 +90,11 @@
                 <td>
                     <g:displayReportLogo/>
                 </td>
-                <td>
+                <td style="white-space: nowrap;">
```
jmiranda, line 93: Same here. We should just add a CSS class.

jmiranda:
@awalkowiak Can you take a look at this one as well? 

@mdpearson Are these test build failures an actual problem or just an intermittent issue?

image

grails-app/taglib/org/pih/warehouse/WordWrapTagLib.groovy
```diff
@@ -0,0 +1,61 @@
...
+class WordWrapTagLib {
+
+    static String namespace = 'wordwrap'
+
+    /**
+     * Enclose `body` in an html element that wraps only on spaces.
+     *
+     * TIL you can't (currently) ask CSS to do this directly. Its default
+     * word-wrapping behavior allows breaks at slashes and hyphens,
+     * which can make a mess of tabular output containing dates and ID's.
```
mdpearson, line 24: @jmiranda I couldn't find a way to tell CSS to wrap only on spaces, the comments above and below explain why in more detail. If you or @awalkowiak know a way to do this in CSS I will gladly delete this taglib. Thanks!

jmiranda: Sorry, glossed right over the comment here.

mdpearson:
> @mdpearson Are these test build failures an actual problem or just an intermittent issue?
> 
> image

@jmiranda I'm afraid they're both. It looks like sdkman can't download Grails 1.3.9 anymore. I get a 403 trying it locally. Other builds don't seem to be affected because they use GHA's cache, but this will eventually cause problems elsewhere. https://github.com/openboxes/openboxes/pull/3685

grails-app/taglib/org/pih/warehouse/WordWrapTagLib.groovy
awalkowiak: @mdpearson I can't think of and find anything that would be useful here. I think we have to leave this version.

jmiranda (approved):

PR #3682, opened 2022-12-05
OBPIH-5090 adjust field length with field content in select fields of reusable filter component

drodzewicz:


drodzewicz:
filter input width is dependant on the width of the content that is displayed in the input or width of the placeholder

I have also cherry-picked a commit from OBPIH-5171 from hotfix-2 because I figured that it might have an impact on the behavior of the filterSelect component

src/js/components/Filter/FilterForm.jsx
```diff
@@ -130,14 +131,15 @@ const FilterForm = ({
...
+                      type="button"
                     />
                     <Button
                       defaultLabel="Search"
                       label="react.button.search.label"
+                      onClick={() => updateFilterParams(values)}
                       disabled={!allowEmptySubmit && _.every(values, value => !value)}
                       variant="primary"
-                      type="submit"
+                      type="button"
```
kchelstowski, line 142: I'm just wondering why it has been changed? Isn't `onSubmit` unnecessary now?
What were the sideeffects that `onClick={() => updateFilterParams(values)}` handles and submit doesn't, if `onSubmit` we are also calling the same (`updateFilterParams`)?

kchelstowski (commented):
Besides one doubt, which is rather just a question, good job :)

awalkowiak (approved):

PR #3686, opened 2022-12-06
OBPIH-5040 Fix crowdin not detecting translatable labels on wizard steps

drodzewicz:


src/js/components/invoice/InvoiceWizard.jsx
```diff
@@ -68,7 +68,7 @@ class InvoiceWizard extends Component {
     }
   }
 
-  getWizardTitle() {
+  get wizardTitle() {
```
awalkowiak, line 71: @drodzewicz Does this getter change help for crowdin?

drodzewicz: The main change that is important is to return `stepList` value from a function instead of  having it as a class property
```
  stepList = [
    this.props.translate('react.invoice.create.label', 'Create'),
    this.props.translate('react.invoice.addItems.label', 'Add items'),
    this.props.translate('react.invoice.confirm.label', 'Confirm'),
  ];
```
```
  get stepList() { [
    this.props.translate('react.invoice.create.label', 'Create'),
    this.props.translate('react.invoice.addItems.label', 'Add items'),
    this.props.translate('react.invoice.confirm.label', 'Confirm'),
  ];
};
```
I guess the reason for that is in the a component life where when `this.props.translate` executes it doe snot yet have access to fetched translations.
The `get stepList()` is just a little refactor that I thought made sense since these functions are already getters so we can use javascript built in class feature for exactly that.

awalkowiak (approved):

PR #3693, opened 2022-12-08
OBPIH-5133 Return exact match on product search API

jmiranda:


kchelstowski (approved):

awalkowiak (approved):
LGTM. I was wondering if we would want to find an exact match with case insensitive option, but considering that codes don't have to have only capital letters, I think case-sensitive is a better approach.

jmiranda:
@awalkowiak Sorry just saw this comment. I think that's a great point. In the case of a barcode scanner it won't matter since the barcode value should be exactly the same as the value in the database. If that's not the case, then that's a data issue.

But if someone is typing the value on their own then i think case insensitivity makes sense. We should mention that in the ticket, but let's let this breathe a little in QA / UAT and let Katarzyna and Kelsey decide.

PR #3696, opened 2022-12-09
OBPIH-4837 Transactions Report in csv makes credits and debits 0 for …

kchelstowski:
…locale in Fr

Like I told to @awalkowiak I don't feel good about this solution, but couldn't figure out anything better for now.
The problem was, that `transactionData` stores `transactionTypeName` always in English, for example: `Transfer in`, but `transactionTypeNames` example is: `Transfer in|fr:<french_translation>`, which resulted in comparing in line `534` for example `Transfer In == <its_french_translation>`, so the result was always `0`.
So the solution I figured out was just to split the `transationTypeName` by the delimiter (`|`) and take the first element (English), so we compare it propely.
Still, the translated version is needed in order to put it to the header of CSV, so I kept it

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
@@ -467,9 +467,9 @@ class InventorySnapshotService {
 
         def transactionData = getTransactionReportData(location, startDate, endDate)
 
-        def transactionTypeNames = TransactionType.createCriteria().list {
+        def transactionTypes = TransactionType.createCriteria().list {
             'in'("transactionCode", [TransactionCode.DEBIT, TransactionCode.CREDIT])
-        }.collect { it.name }
+        }
 
         // Get starting balance
         def balanceOpeningMap = getQuantityOnHandByProduct(location, startDate)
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -525,14 +525,15 @@ class InventorySnapshotService {
...
             transactionTypeNames.each { String transactionTypeName ->
+                String translatedTransactionTypeName = transactionTypeName
                 if (transactionTypeName.contains(Constants.LOCALIZED_STRING_SEPARATOR)) {
-                    transactionTypeName = LocalizationUtil.getLocalizedString(transactionTypeName)
+                    translatedTransactionTypeName = LocalizationUtil.getLocalizedString(transactionTypeName)
                 }
                 def quantity =
                         transactionData.find {
-                            it.productCode == product.productCode && it.transactionTypeName == transactionTypeName
+                            it.productCode == product.productCode && it.transactionTypeName == transactionTypeName.split("\\|")[0]
```
jmiranda, line 534: Why is this necessary? Shouldn't getLocalizedString be doing that for us?

kchelstowski: It does, but it compares `English == French` which always results in `false`. 
The `transactionData` in line `468` stores the `transactionTypeName` as English only, so that's why I made it

kchelstowski: I'll send some images from debugging in a moment to show what I mean

kchelstowski: So, the `transactionData` example object looks like this:
Transaction data

and the `transationTypeName` example looks like this:
Transaction data

so it was comparing `Consumption == Consommantion` for example, instead of comparing `Consumption == Consumption`

jmiranda: Ok got it. I am going to write a comment in the ticket to explain what I'd like to see in the long term. But for now, you have the right approach. Let's just move that logic to a convenience method on the TransactionType class if possible.

@awalkowiak Are we allowed to do something like this
```
Boolean compareTo(String transactionTypeName)
Boolean equals(String transactionTypeName)
```
or would it be better not to implement methods that are defined by common interfaces?

awalkowiak: @jmiranda I think I'd go with something like `Boolean compareTransactionType(String transactionTypeName)` because the string that might be passed not always will be the `transactionTypeName`, so for clarity, I would avoid using `compareTo` and `equals` and strictly let others know what I want to compare.
cc @kchelstowski

awalkowiak (approved):

PR #3700, opened 2022-12-12
OBPIH-4946 Refactor redux actions 

alannadolny:


src/js/actions/index.js
```diff
@@ -35,6 +34,7 @@ import {
   UPDATE_BREADCRUMBS_PARAMS,
 } from 'actions/types';
 import apiClient, { parseResponse } from 'utils/apiClient';
+// import { dispatch } from 'store';
```
awalkowiak, line 37: Remove if not required

awalkowiak (changes requested):

kchelstowski (approved):
I'd prefer async/await convention, but it's just about preference

PR #3701, opened 2022-12-12
OBPIH-5231 Add front-end unit tests for Wizard and Menu components

alannadolny:


jmiranda (commented):
Great idea @awalkowiak 

@alannadolny Can you explain why you deleted the jar file?

awalkowiak:
@jmiranda my blind guess: this is some weird issue after IDE setup. I locally created a fresh branch from develop and cherry-picked these commits and this lib removal change is no longer there. Plus, as you can see currently this `lib/something.jar` does not exists in the develop branch. We will sort this out.

drodzewicz (changes requested):

drodzewicz (approved):

PR #3703, opened 2022-12-12
OBGM-372 Build improvements made while looking for root cause

mdpearson:
This PR contains the stuff I tried before [fixing the root cause of the ClassNotFoundException](https://github.com/openboxes/openboxes/pull/3702), including:

1. Removing iText from Grails 3 (like OBPIH-5191 did for Grails 1),
2. Updating docx4j from 2.x to 8.x (which is actually supported, OBGM-380),
3. Removing build rules for old dependencies steps (1) and (2) removed from our dependency tree,
4. Cleaning up the logic we use to build standalone or tomcat-specific war files,
5. Fixing a number of namespace collisions (OBGM-392),
6. Updating some dependencies to get bug fixes and security patches.

I only tested #3702 with these changes, so the two should merge as a pair; I just separated them for ease of review and to separate a confirmed bug fix from general housekeeping.


awalkowiak:
@mdpearson I've seen you merged https://github.com/openboxes/openboxes/pull/3702, should this one be merged too?

PR #3707, opened 2022-12-13
OBGM-293 saving stockmovement leads to internal server error

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/StockTransferApiController.groovy
```diff
@@ -67,7 +67,7 @@ class StockTransferApiController {
         bindStockTransferData(stockTransfer, currentUser, currentLocation, jsonObject)
 
         Order order = stockTransferService.createOrUpdateOrderFromStockTransfer(stockTransfer)
-        if (order.hasErrors() || !order.save(flush: true)) {
```
awalkowiak, line 70: Why this save was removed?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -545,7 +545,15 @@ class StockMovementService {
                 throw new ObjectNotFoundException(id, StockMovement.class.toString())
             }
         }
+    }
+
+    StockMovement getShipmentBasedStockMovement(Shipment shipment) {
```
awalkowiak, line 550: Nice catch

grails-app/services/org/pih/warehouse/stockTransfer/StockTransferService.groovy
```diff
@@ -194,7 +195,7 @@ class StockTransferService {
         }
         order.save(failOnError: true)
 
-        def currentLocation = authService.currentLocation
+        def currentLocation = AuthService?.currentLocation
```
awalkowiak, line 198: Instead of replacing it with `AuthService`, please just inject service dependency as `def authService` at the beginning of this service class.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/StockTransferApiController.groovy
drodzewicz: This save throws `No transaction in progress error` which I wasn't sure why.
I looked inside `stockTransferService.createOrUpdateOrderFromStockTransfer()` method and found that `order.save()` is already being triggered there so I assumed that this might be the problem.

awalkowiak (approved):

PR #3712, opened 2022-12-15
OBPIH-5172 Add localization mode indicator to React pages

kchelstowski:


jmiranda (approved):
Looks good. What happens if both impersonate and translation mode are enabled at the same time?

kchelstowski:
@jmiranda they are displayed one under another

PR #3716, opened 2022-12-16
OBGM-387 Fix binLocation report export crashing

drodzewicz:
pass methods as parameters using method reference operator

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
@@ -415,7 +415,7 @@ class ReportController {
                     binLocations = binLocations.findAll { it.status == params.status }
                 }
 
-                String csv = ReportUtil.getCsvForListOfMapEntries(binLocations, binLocationCsvHeader, binLocationCsvRow)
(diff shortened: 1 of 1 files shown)


drodzewicz:
apparently in newer versions of groovy it doesn't let you to just simply pass a whole **class method** as an argument like before, you have to reference it using a  [method pointer ](https://groovy-lang.org/operators.html#method-pointer-operator)

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -415,7 +415,7 @@ class ReportController {
                     binLocations = binLocations.findAll { it.status == params.status }
                 }
 
-                String csv = ReportUtil.getCsvForListOfMapEntries(binLocations, binLocationCsvHeader, binLocationCsvRow)
+                String csv = ReportUtil.getCsvForListOfMapEntries(binLocations, this.&binLocationCsvHeader, this.&binLocationCsvRow)
```
awalkowiak, line 418: What does `this.&` do in that context?

drodzewicz: it's a  [method pointer](https://groovy-lang.org/operators.html#method-pointer-operator)
From what I understand if you want to pass a class method as argument in newer versions of groovy you have to pass it as a reference pointer

awalkowiak (approved):

PR #3718, opened 2022-12-17
OBPIH-5236 Add front-end unit tests for Util components and functions

alannadolny:


src/js/tests/.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  "parser": "@babel/eslint-parser",
```
awalkowiak, line 2: Is this required in the end?

awalkowiak (commented):
Overall LGTM, but I had one question

src/js/tests/.eslintrc
alannadolny: No, I forgot to remove that

kchelstowski (approved):

drodzewicz (approved):

PR #3720, opened 2022-12-19
OBGM-389 Increase ajax timeout configuration to 120000ms

drodzewicz:


drodzewicz:
The transaction report takes a long time to generate and since timeout for this request was set to _10000ms_ it would cancel it before it could return any data.
I checked with the development branch and `grailsApplication.config.openboxes.ajaxRequest.timeout` is set to _120000ms_ there.

jmiranda (approved):

jmiranda:
This reminds me that we probably have a ton of properties defined in Config.groovy that haven't been migrated to application.yml.

PR #3723, opened 2022-12-20
OBPIH-5120 Show that loading is in progress on substitution window

alannadolny:


grails-app/i18n/messages.properties
```diff
@@ -3036,6 +3036,7 @@ react.stockMovement.edit.label=Edit
 react.stockMovement.pick.label=Pick
 react.stockMovement.pack.label=Pack
 react.stockMovement.send.label=Send
+react.stockMovement.noSubstitutionsAvailable.message=There are no substitutions available
```
kchelstowski, line 3039: I think it won't work if it doesn't have `label` at the end

src/js/components/form-elements/TableBody.jsx
```diff
@@ -1,17 +1,41 @@
...
+    tableRef = () => {
+    },
     addRow = (row = {}) => fields.push(row),
   } = props;
   const RowComponent = properties.subfield ? TableRow : fieldsConfig.rowComponent || TableRow;
 
+  const {
+    emptySubstitutions,
+    loadingSubstitutions,
+  } = properties;
```
kchelstowski, line 26: I think you should add a comment about usage of those properties and also about the two ifs below, as `TableBody` is a component we use in many places and this particular case will only be used for substitution modal

src/js/components/form-elements/TableBody.jsx
```diff
@@ -1,17 +1,41 @@
...
+
+  if (loadingSubstitutions) {
+    return (<Spinner />);
+  }
+
+  if (emptySubstitutions && !fields.length) {
```
kchelstowski, line 32: what is that `fields.length` responsible for here? `fields` is probably an object, so I do not understand the usage of that? Isn't checking `emptySubstitutions` enough?

src/js/components/stock-movement-wizard/modals/SubstitutionsModal.jsx
```diff
@@ -147,17 +167,22 @@ const FIELDS = {
...
+      fieldConfig: {
+        attributes,
+        getDynamicAttr,
+      },
     } = props;
     const dynamicAttr = getDynamicAttr ? getDynamicAttr(props) : {};
     const attr = { ...attributes, ...dynamicAttr };
 
     this.state = {
+      loadingSubstitutions: false,
```
kchelstowski, line 185: I can't see `emptySubstitutions` being initialized here?

src/js/components/stock-movement-wizard/modals/SubstitutionsModal.jsx
```diff
@@ -264,12 +297,15 @@ class SubstitutionsModal extends Component {
...
+          emptySubstitutions: !substitutions.length,
+          loadingSubstitutions: false,
           formValues: {
             substitutions,
           },
           originalItem,
         });
-      }).catch(err => err);
+      })
+      .catch(err => err);
```
kchelstowski, line 308: don't we want to mark `loadingSubstitutions` as `true` if we get an error also? In my opinion if any error occurs, the spinner will still be visible.
You could use `.finally` in the end, so you don't have to hide the spinner twice

kchelstowski (changes requested):
Moreover I think you should avoid reformatting the code everywhere, because if there are bigger changes, it might be difficult to follow where you provided a change and where it is only a code reformatting

src/js/components/form-elements/TableBody.jsx
kchelstowski: I also think that such a request might come in some other places/modals, so maybe we could think about some "generic" naming here? @awalkowiak what do you think?

grails-app/i18n/messages.properties
drodzewicz: why? Do you mean crowding will not detect this label ?
I am pretty sure there are many labels that end with **.message** that we have in the project and it looks like all of them have translations

kchelstowski: I remember not a long time ago I didn't add `.label` and the translation did not work, but I do not rememeber if it was on gsp pages or react

kchelstowski: but it could be a random case that it didn't want to work for me, so if you are confident that it should work, I'm fine with it

src/js/components/form-elements/TableBody.jsx
alannadolny: Yes, fields is an object, but it has its length as a property. Checking emptySubstitutions isn't enough, because when you are clicking 'add custom substitution' emptySubsitutions is still 0, but fields.length will be > 0, so message with 'There are no substitutions available' will disapear. When it will be only emptySubstitutions there will be displayed this message and added substitution.

alannadolny:
Yes, I know that code formatting can be problematic and I am trying to avoid it, but this is my habit, so I forget about it sometimes. I will pay more attention to code formatting.

grails-app/i18n/messages.properties
drodzewicz: I think we should leave it as it is and if after merge it turns out that it doesn't work then we will know for sure :D

kchelstowski: I checked locally this branch and the translation seems to be working fine even if removing `.message` etc, so my case when it did not work without `.label` was probably a random one. Apologize for confusion

kchelstowski: what I meant at the beginning is that the label won't even be recognized, so that it'd take the `defaultMessage`, not that it will work, but will not be translatable with crowdin, but never mind, everything is fine

src/js/components/form-elements/TableBody.jsx
awalkowiak: @kchelstowski @alannadolny yes, this definitely has to be generic

src/js/components/form-elements/TableBody.jsx
```diff
@@ -1,17 +1,45 @@
...
+  // if substitution is still loading then spinner will be displayed within the table
+  if (loadingSubstitutions) {
+    return (<Spinner />);
+  }
+
+  // if there is no data to display then message will be displayed
+  // fields.length is checked due to possibility of adding custom substitution
+  if (emptySubstitutions && !fields.length) {
+    return (translate(
+      'react.stockMovement.noSubstitutionsAvailable.message',
```
awalkowiak, line 38: This should be passed from the modal component

kchelstowski (approved):

awalkowiak (approved):

PR #3724, opened 2022-12-20
OBGM-393 Catch data Integrity violation on product source deletion

drodzewicz:


drodzewicz:
I've spent some time trying to figure out the exact cause and the best solution for this problem but I wasn't able to and came up with just a quick band aid for this issue by replacing 
```productSupplierInstance.delete(flush: true)```
with 
```ProductSupplier.executeUpdate('delete ProductSupplier ps where ps.id = :id', [id: params.id])```

The main issue for this problem from what I've been able to gather is that when deleting a _productSource_ that is already refferenced for example in some **OrderItem** it should throw a `DataIntegrityViolationException` exception and the _try/catch_ block inside the `def delete() { ... }` should catch that exception and return an appropriate `flash.message` saying that you can't do that.
The problem was that for some reason when calling `productSourceInstance.delete()` it would not catch that exception inside of the _try/catch_ and would instead throw it elsewhere _(my guess is that i throws it somewhere at the end)_ 

I am not 100% satisfied with the solution and not exactly sure on what is the problem so if you have any ideas then let me know, but the current solution fixes the problem and delete productSource works as it used to.

**Review by kchelstowski — COMMENTED** [review]
@drodzewicz my blind guess reading the error from the ticket would be that you nest the transactions, so you have a situation, where you have a service marked as `@Transactional` and you call some method from another service which is also `@Transactional`.
You could try to set in a proper service (probably here it would be `ProductSupplierController` 
`@Transactional(rollbackFor = { org.springframework.dao.DataIntegrityViolationException })` and see what happens, but I also don't know if it's the best solution (or it is is even a solution :laughing: )
Another solution could be maybe to add `@Transactional(propagation = Propagation.REQUIRES_NEW)`

**Comment by kchelstowski** [comment]
@drodzewicz Some resource: https://stackoverflow.com/questions/25322658/could-not-commit-jpa-transaction-transaction-marked-as-rollbackonly

**Review by jmiranda — COMMENTED** [review]
This doesn't seem like a good solution. What is the actual data integrity error we're seeing?

**Comment by kchelstowski** [comment]
After long research and trying many different solutions, it seems like creating a `@Transactional` service and moving delete logic there and removing `@Transactional` annotation from controller is the only possible solution.
I know that @jmiranda you are not a fan of changing the Transactional logic for now, as it's "a bigger refactor", but I couldn't figure out any better workaround.
The problem was that even though we were getting to a point, where the exception `SQLIntegrityConstraintViolationException` is thrown, we couldn't see the `flash.message` on the actual view, but we were getting the `TransactionException` (the one that is attached as a screenshot in the ticket).
When moving the logic to the service and marking the service as `@Transactional` and keeping the `try/catch` in a controller seems to be working fine - we just catch all the exceptions before we start even interacting with the database.

It also depends on what we want to achieve - from the code perspective my blind guess is that we don't want to allow deleting a product source that is already associated with something - because if I wanted, another idea could be to just add:
```
productPackages cascade: 'all-delete-orphan'
```
to `ProductSupplier` mapping - but that would delete all the product packages that are associated with the product source, but I believe we don't want to allow deleting a product source that is already associated with something, so the preposition for now is to move the logic to the service and mark the service as transactional, not the controller, but since @jmiranda didn't want to do it now, because it would be a pain in the neck while rebasing, I think we will have to discuss it even further.

jmiranda:
Marked as `do not merge`. I'll try to take a look at this as soon as I can.

jmiranda (approved):

jmiranda:
I still want to pull this down and play around with it because I don't like that we removed the @Transactional for all of the other controller actions.

jmiranda:
One compromise is that we could make all of the service layer changes to develop and move them over in a rebase. That would make me feel more comfortable.

PR #3730, opened 2022-12-22
OBPIH-5195 fix displaying dropdowns

alannadolny:


src/js/components/Layout/menu/MenuSubsection.jsx
```diff
@@ -1,30 +1,47 @@
...
-              <a className="dropdown-item" key={menuItemKey} href={menuItem.href} target={menuItem.target}>
-                {menuItem.label}
-              </a>
-            ))}
-          </div>
-        ))}
+
+const DropdownMenu = ({ section, active }) => {
+  const dropdownRef = useRef(null);
+  const [droppedDown, setDroppedDown] = useState(false);
```
drodzewicz, line 10: I would call this state `[isOpen, setIsOpen] ` instead of `[droppedDown, setDroppedDown]`

src/js/components/Layout/menu/MenuSubsection.jsx
```diff
@@ -1,30 +1,47 @@
...
+
+  const shouldAlignLeft = (ref) => {
```
drodzewicz, line 12: I think that you don't need to have a `ref` as an argument here, you can just call the `dropdownRef` directly in the function body

src/js/components/Layout/menu/MenuSubsection.jsx
```diff
@@ -1,30 +1,47 @@
...
+    if (!ref.current) {
+      return false;
+    }
+    const elementCoordinates = ref.current.getBoundingClientRect();
+    if (droppedDown) {
+      return window.innerWidth > elementCoordinates.right + elementCoordinates.width;
+    }
+    return false;
+  };
```
drodzewicz, line 21: this method can be simplified to 
```
const shouldAlignLeft = () => {
    if (dropdownRef.current && isOpen) {
      const elementCoordinates = ref.current.getBoundingClientRect();
      return window.innerWidth > elementCoordinates.right + elementCoordinates.width;
    }
    return false;
}
```

other than that everything looks good to me, good job.

drodzewicz (changes requested):

kchelstowski (changes requested):
What about moving it to a custom hook, as it might be useful in the future?

src/js/hooks/useCheckRightSpace.jsx
```diff
@@ -0,0 +1,9 @@
+const useCheckRightSpace = (elementReference, isVisible) => {
+  if (elementReference.current && isVisible) {
+    const elementCoordinates = elementReference.current.getBoundingClientRect();
+    return window.innerWidth > elementCoordinates.right + elementCoordinates.width;
+  }
+  return false;
+};
+
+export default useCheckRightSpace;
```
kchelstowski, line 9: this is not a hook - what I meant is to move `ref` and `states` logic to a hook, so you could return them in the hook like this:
```
return { dropdownRef, isOpen, setIsOpen }
```
and you could use it in the component like this:
```
const { dropdownRef, isOpen, setIsOpen } = useCheckRightSpace();
```
remember to change the names in the hook to generic ones, but in the component you could do:
```
const { dropdownRef: <ref_name_from_hook> } 
```

kchelstowski (changes requested):

src/js/hooks/useCheckRightSpace.jsx
kchelstowski: remember also about moving your function `shouldAlignLeft` to the hook

kchelstowski (approved):

awalkowiak (approved):

PR #3731, opened 2022-12-22
OBPIH-5234 Return most recent Stock Movement when search returns multiple shipments with identical identifier

drodzewicz:


grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -69,14 +69,16 @@ class DashboardController {
             return
         }
 
-        def requisition = Requisition.findByRequestNumber(params.searchTerms)
-        if (requisition) {
+        List<Requisition> requisitions = Requisition.findAllByRequestNumber(params.searchTerms, [sort: 'dateCreated'])
+        if (requisitions.size() > 0) {
```
kchelstowski, line 73: you could just do `if (requisitions.size())`, as 0 is falsy

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -69,14 +69,16 @@ class DashboardController {
             return
         }
 
-        def requisition = Requisition.findByRequestNumber(params.searchTerms)
-        if (requisition) {
+        List<Requisition> requisitions = Requisition.findAllByRequestNumber(params.searchTerms, [sort: 'dateCreated'])
+        if (requisitions) {
```
kchelstowski, line 73: doesn't `findAllBy` return `[]` if there are no records? If so, empty list is truthy, so it will always go inside if and crash when doing `[].first()`
This is why we need to do `requisitions.size()`

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
drodzewicz: `findAllBy...()` does return an empty list when it does not match any records
empty list `[]` in groovy evaluates to `false` 
https://docs.groovy-lang.org/latest/html/documentation/#_list_as_a_boolean_expression

kchelstowski: oh ok, so it's different than in JS, but I remember at some point I got "destroyed" by Groovy when not adding this `.size()`, but I do not remember the exact case now

kchelstowski (approved):

awalkowiak (approved):

PR #3735, opened 2022-12-23
OBPIH-5252 Move filter form logic on list pages to a custom hook

alannadolny:


src/js/hooks/usePurchaseOrderFilters.jsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const fetchLocationById = async (id) => {
+    const response = await apiClient(`/openboxes/api/locations/${id}`);
+    return response.data?.data;
+  };
+
+  const fetchUserById = async (id) => {
+    const response = await apiClient(`/openboxes/api/generic/person/${id}`);
+    return response.data?.data;
+  };
```
kchelstowski, line 49: same as above

src/js/hooks/useProductFilters.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const fetchProductsCatalogs = async () => {
+    const response = await apiClient.get('/openboxes/api/catalogOptions');
+    return response.data.data;
+  };
+
+  const fetchProductsTags = async () => {
+    const response = await apiClient.get('/openboxes/api/tagOptions');
+    return response.data.data;
+  };
```
kchelstowski, line 36: I'm wondering if those three functions have to be inside the component and re-render each time? I know it's not you, who pulled it inside the component, but maybe it would be a good idea to move those functions to `utils/option-utils`, as those functions could be potentially useful in the future.

src/js/hooks/useStockListFilters.jsx
```diff
@@ -0,0 +1,111 @@
...
+  const [locations, setLocations] = useState([]);
+  const history = useHistory();
+
+  const { currentLocation } = useSelector(state => ({
+    currentLocation: state.session.currentLocation,
+  }));
+
+  const fetchLocationById = async (id) => {
+    const response = await apiClient(`/openboxes/api/locations/${id}`);
+    return response.data?.data;
```
kchelstowski, line 25: seems to be the same function that is used in purchase order list hook - that's why moving in to a util has sense

src/js/hooks/useStockTransferFilters.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const history = useHistory();
+  const { statuses, currentUser, currentLocation } = useSelector(state => ({
+    statuses: state.stockTransfer.statuses,
+    currentUser: state.session.user,
+    currentLocation: state.session.currentLocation,
+  }));
+
+  const fetchUserById = async (id) => {
+    const response = await apiClient(`/openboxes/api/generic/person/${id}`);
+    return response.data?.data;
```
kchelstowski, line 25: same as above

kchelstowski (changes requested):
Other than that, good job :)

jmiranda:
@kchelstowski @drodzewicz Given my limited knowledge of React, I'm going to rely on you to tell me when this is ready to merge.

kchelstowski (changes requested):
Everything looks good to me now, but would you mind changing the names of hooks from e.g. `useInboundFilters` => `useInboundListFilters`, and creating a structure for hooks, so we have it organized from the beginning? For example `hooks/list-pages/inbound` etc.

drodzewicz (approved):

kchelstowski (approved):
@jmiranda green light for merging this one

PR #3736, opened 2022-12-24
OBPIH-5237 Create a transaction for inventories with entries for both…

awalkowiak:
… products

This is required because of the issues with calculating current stock of primary product after merge. Previously when moving obsolete's most recent transaction entries to the primary's most recent transaction (with types PRODUCT_INVENTORY or INVENTORY), the stock might not reflect the current status. Especially when there were additional transaction entries (DEBITS and CREDITS) after the obsolete transaction (with types PRODUCT_INVENTORY or INVENTORY) which led to discrepancies.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
@@ -3371,14 +3371,10 @@ class InventoryService implements ApplicationContextAware {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jmiranda:
We should discuss this at the next tech huddle.

jmiranda (commented):
Needs discussion with @awalkowiak but merging it as-is so it can be tested while he's away.

PR #3738, opened 2022-12-27
OBPIH-5161+5246 Make user transition in/out of translation mode smoother

kchelstowski:
- getting rid of `useDebugLocale` logic, which Justin mentioned in the ticket 
- not allowing user to set her/his default locale to translation mode locale
- the localization mode is activated automatically when changing the language for Acoli
- the localization mode is deactivated when changing language to any other than Acoli when being in localization mode
- when disabling localization mode by button, we are brought back to the locale, that we had set before enabling localization mode

Bonus:

- changing the `webpack` config, so all babel's and eslint's stuff works also fine for `.js` files, not only `.jsx` - example: optional chaining or nullish coalescing operator, which were added by me not a long time ago, but were not working for `.js` files

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
@@ -57,12 +57,12 @@ class ApiController {
 
     @CacheFlush(["megamenuCache"])
     def chooseLocale = {
-        Locale locale = localizationService.getLocale(params.id)
-        if (!locale) {
+        Locale newLocale = localizationService.getLocale(params.id)
+        if (!newLocale) {
             throw new ObjectNotFoundException(params.id, Locale.class.toString())
         }
-        session.user.locale = locale
-        render([status: 200, text: "Current language is ${locale}"])
+        session.locale = newLocale
+        render([status: 200, text: "Current language is ${newLocale}"])
     }
 
     def getMenuConfig = {
@@ -88,7 +88,9 @@ class ApiController {
         def localizationMode
         def locale = localizationService.getCurrentLocale()
         Object[] emptyArgs = [] as Object []
-        if (session.useDebugLocale) {
+        // If current locale is equal to translation mode locale, we are in localization mode
+        boolean localizationModeEnabled = locale == new Locale(grailsApplication.config.openboxes.locale.translationModeLocale)
+        if (localizationModeEnabled) {
 
             localizationMode = [
                 "label"      : messageSource.getMessage('localization.disable.label', emptyArgs, 'Disable translation mode', locale),
@@ -164,7 +166,7 @@ class ApiController {
         String currencyCode = grailsApplication.config.openboxes.locale.defaultCurrencyCode
         String localizedHelpScoutKey = helpScoutService.localizedHelpScoutKey
         boolean isHelpScoutEnabled = grailsApplication.config.openboxes.helpscout.widget.enabled
-        boolean localizationModeEnabled = session.useDebugLocale ?: false
+        def translationModeLocale = grailsApplication.config.openboxes.locale.translationModeLocale
 
         render([
             data: [
@@ -197,7 +199,8 @@ class ApiController {
                 currencyCode         : currencyCode,
                 localizedHelpScoutKey: localizedHelpScoutKey,
                 isHelpScoutEnabled   : isHelpScoutEnabled,
-                localizationModeEnabled : localizationModeEnabled
+                localizationModeEnabled : localizationModeEnabled,
+                translationModeLocale : translationModeLocale
             ],
         ] as JSON)
     }
--- a/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
@@ -57,12 +57,12 @@ class ApiController {
 
     @CacheFlush(["megamenuCache"])
     def chooseLocale = {
-        Locale newLocale = localizationService.getLocale(params.id)
-        if (!newLocale) {
+        Locale locale = localizationService.getLocale(params.id)
+        if (!locale) {
             throw new ObjectNotFoundException(params.id, Locale.class.toString())
         }
-        session.locale = newLocale
-        render([status: 200, text: "Current language is ${newLocale}"])
+        session.locale = locale
+        render([status: 200, text: "Current language is ${locale}"])
     }
 
     def getMenuConfig = {
@@ -88,9 +88,7 @@ class ApiController {
         def localizationMode
         def locale = localizationService.getCurrentLocale()
         Object[] emptyArgs = [] as Object []
-        def translationModeLocale = grailsApplication.config.openboxes.locale.translationModeLocale
-        // If current locale is equal to translation mode locale, we are in localization mode
-        session.useDebugLocale = locale == new Locale(translationModeLocale)
+        def localizationModeLocale = grailsApplication.config.openboxes.locale.localizationModeLocale
 
         if (session.useDebugLocale) {
 
@@ -202,7 +200,7 @@ class ApiController {
                 localizedHelpScoutKey: localizedHelpScoutKey,
                 isHelpScoutEnabled   : isHelpScoutEnabled,
                 localizationModeEnabled : localizationModeEnabled,
-                translationModeLocale : translationModeLocale
+                localizationModeLocale : localizationModeLocale
             ],
         ] as JSON)
     }
--- a/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
@@ -88,9 +88,11 @@ class ApiController {
         def localizationMode
         def locale = localizationService.getCurrentLocale()
         Object[] emptyArgs = [] as Object []
+        def translationModeLocale = grailsApplication.config.openboxes.locale.translationModeLocale
         // If current locale is equal to translation mode locale, we are in localization mode
-        boolean localizationModeEnabled = locale == new Locale(grailsApplication.config.openboxes.locale.translationModeLocale)
-        if (localizationModeEnabled) {
+        session.useDebugLocale = locale == new Locale(translationModeLocale)
+
+        if (session.useDebugLocale) {
 
             localizationMode = [
                 "label"      : messageSource.getMessage('localization.disable.label', emptyArgs, 'Disable translation mode', locale),
@@ -166,7 +168,7 @@ class ApiController {
         String currencyCode = grailsApplication.config.openboxes.locale.defaultCurrencyCode
         String localizedHelpScoutKey = helpScoutService.localizedHelpScoutKey
         boolean isHelpScoutEnabled = grailsApplication.config.openboxes.helpscout.widget.enabled
-        def translationModeLocale = grailsApplication.config.openboxes.locale.translationModeLocale
+        boolean localizationModeEnabled = session.useDebugLocale ?: false
 
         render([
             data: [
--- a/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
@@ -29,6 +29,7 @@ class UserController {
     MailService mailService
     def userService
     def locationService
+    def localizationService
 
     /**
      * Show index page - just a redirect to the list page.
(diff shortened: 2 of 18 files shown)


src/js/components/Layout/Footer.jsx
```diff
@@ -64,15 +102,22 @@ const mapStateToProps = state => ({
...
   ipAddress: state.session.ipAddress,
   languages: getLanguages(state.localize),
+  localizationModeEnabled: state.session.localizationModeEnabled,
+  translationModeLocale: state.session.translationModeLocale,
 });
 
-export default connect(mapStateToProps, { changeCurrentLocale, setActiveLanguage })(Footer);
+const mapDispatchToProps = {
+  changeLocale: changeCurrentLocale,
+  setLanguage: setActiveLanguage,
```
alannadolny, line 111: What do you think about using `useDispatch()` instead of `mapDispatchToProps`? I found an interesting discussion on stackoverflow about this two approach [Should I use useselector/useDispatch instead of mapStateToProps](https://stackoverflow.com/questions/59748180/should-i-use-useselector-usedispatch-instead-of-mapstatetoprops) and i am curious what do you think

jmiranda:
I need a bit more time to review this one.

src/js/components/Layout/Footer.jsx
```diff
@@ -27,21 +39,47 @@ const Footer = ({
...
+        { _.map(languages, (language) => {
+          // When clicking on language that is a translation mode language, enable localization mode
+          if (language.code === translationModeLocale) {
+            return (
+              <a
+                className={`${locale === language.code ? 'selected' : ''}`}
+                key={language.code}
+                href="/openboxes/user/enableLocalizationMode"
+              >
+                <Translate id={`react.default.${language.name.toLowerCase()}.label`} /> {''}
```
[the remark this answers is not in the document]
kchelstowski: Probably not, but since it has not been code that I've written, I did not catch such details

src/js/components/Layout/Footer.jsx
kchelstowski: I don't see any pros/cons of using one versus another. I think it's rather what you prefer and what is more useful in each scenario (for example you can't use `mapDispatchToProps` inside a custom hook, so you are obligated to use `useDispatch()`. 
What I'm sure about is that we shouldn't mix both if possible and probably in "regular" components stick with `connect` approach as we've done for whole project, as it might be easier to read and debug, because we are used to this convention.
@drodzewicz any additional thoughts?

kchelstowski:
Marking it as a draft for a while, as I will probably want to bring back `useDebugLocale` for some cases, because creating vars with  `locale == translationModeLocale` for many GSP pages does not seem the best idea for me.

kchelstowski:
@jmiranda it's now ready for review. I've brought back `useDebugLocale` in the latest commit for some use cases, let me know if it's better to keep it like I did, or remove it completely and inject logic `locale == translationModeLocale` for each GSP page where we used to use `if (session.useDebugLocale)`. 
I personally believe that the latest version should be more readable and less complexed, but I leave it for you to decide :)

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -88,6 +88,10 @@ class ApiController {
         def localizationMode
         def locale = localizationService.getCurrentLocale()
         Object[] emptyArgs = [] as Object []
+        def translationModeLocale = grailsApplication.config.openboxes.locale.translationModeLocale
+        // If current locale is equal to translation mode locale, we are in localization mode
+        session.useDebugLocale = locale == new Locale(translationModeLocale)
```
jmiranda, line 93: Instead of sprinkling this logic everywhere, we should probably have it done in a filter. InitializationFilters seems to be the best candidate but I'm open to suggestions.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -57,12 +57,12 @@ class ApiController {
 
     @CacheFlush(["megamenuCache"])
     def chooseLocale = {
-        Locale locale = localizationService.getLocale(params.id)
-        if (!locale) {
+        Locale newLocale = localizationService.getLocale(params.id)
```
jmiranda, line 60: This seems unnecessary.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -197,7 +201,8 @@ class ApiController {
                 currencyCode         : currencyCode,
                 localizedHelpScoutKey: localizedHelpScoutKey,
                 isHelpScoutEnabled   : isHelpScoutEnabled,
-                localizationModeEnabled : localizationModeEnabled
+                localizationModeEnabled : localizationModeEnabled,
+                translationModeLocale : translationModeLocale
```
jmiranda, line 205: This is my fault since I kept referring to this as "translation mode", but let's either use the existing name "localization mode" or convert everything to "translation mode".

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -266,32 +267,31 @@ class UserController {
 
 
     def disableLocalizationMode = {
+        // If we are in localization mode and want to change the language through the footer on React side
+        // we want to make sure we set the language to the clicked one, not for the previous locale
+        // we want to set the prev locale only if we disable localization mode through "Disable localization mode" button
+        Locale localeToSet = params?.localeToSet ? new Locale(params.localeToSet) : session?.prevLocale
```
jmiranda, line 273: Let's just call this request parameter and variable "locale" (not localeToSet) since that's what it's been called. I'm worried that adding a new request parameter will cause confusion.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -266,32 +267,31 @@ class UserController {
...
-        } else {
-            redirect(controller: "dashboard", action: "index")
+            return
         }
+        redirect(controller: "dashboard", action: "index")
     }
 
     def enableLocalizationMode = {
+        // We want to store the previous locale, so we can go back to it when disabling localization mode through button
+        session.prevLocale = localizationService.getCurrentLocale()
```
jmiranda, line 286: Sorry to be nit-picky but let's avoid using abbreviations. `session.previousLocale` is more readable.

jmiranda: Also is there any difference between getting the current locale versus using `session.locale`?

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -302,40 +302,25 @@ class UserController {
...
-            if (!params.locale) {
-                redirect(controller: "dashboard", action: "index")
-            }
-
-            // convert the passed locale parameter to an actual locale
-            Locale locale = new Locale(params.locale)
+        // if no locale specified, do nothing
+        if (!params.locale) {
+            redirect(controller: "dashboard", action: "index")
+        }
```
jmiranda, line 308: I know this isn't code you wrote, but we should probably remove the redirects since that's handled at the end. And instead of handling the 
```
    if (params.locale) { 
        // convert the passed locale parameter to an actual locale
        Locale locale = new Locale(params.locale)
    
        if (locale) { 
            // Change the locale
            session.locale = locale
        
            // remove all redirects since they aren't going to do anything 
        
            // Check if chosen locale is equal to translation mode locale, if so, enable localization mode
            session.useDebugLocale = localizationService.isLocalizationModeEnabled() 
        }
    }

    log.info "Redirecting to " + params?.targetUri
    if (params?.targetUri) {
        redirect(uri: params.targetUri)
        return
    }
    // redirect to the dashboard
    redirect(controller: "dashboard", action: "index")
```

LocalizationService
```
isLocalizationModeEnabled() { 
    def localizationModeLocale = new Locale(grailsApplication.config.openboxes.locale.translationModeLocale)
    return currentLocale == localizationModeLocale
```

grails-app/taglib/org/pih/warehouse/MessageTagLib.groovy
```diff
@@ -28,7 +28,7 @@ class MessageTagLib {
         boolean databaseStoreEnabled = grailsApplication.config.openboxes.locale.custom.enabled
         if (!databaseStoreEnabled) {
             Locale defaultLocale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale)
-            attrs.locale = attrs.locale ?: session?.user?.locale ?: session.locale ?: defaultLocale
+            attrs.locale = attrs.locale ?: session?.locale ?: session?.user?.locale ?: defaultLocale
```
jmiranda, line 31: :+1:

grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
```diff
@@ -64,9 +64,15 @@ class LocalizationService {
      * Gets the current locale or return default locale.
      */
     Locale getCurrentLocale() {
-        // fetch the locale of the current user; if there isn't one, use the default locale
-        return (RequestContextHolder.currentRequestAttributes().getSession().user?.locale ?:
-                new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en"))
+        // First search for session locale, if none set, then search for user's default locale, otherwise get the default locale
+        def session = RequestContextHolder.currentRequestAttributes().getSession()
+        if (session?.locale) {
+            return session.locale
```
jmiranda, line 70: :+1:

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
kchelstowski: what do you mean as unnecessary here? I'm a bit confused, because I only changed the naming from `locale` to `newLocale` and I'm not sure if changing of name seems unnecessary or this part: `localizationService.getLocale(params.id)`

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
kchelstowski: sounds good to me, I haven't thought about such a solution

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
kchelstowski: For now it is only used for React part, as I didn't have an idea how to determine whether I disable localization mode through button "Disable localization mode" or I  just change the language.
When disabling the localization mode through button "Disable localization mode" we want to go back to previous chosen locale.
When changing the language while being in localization mode, we want to disable localization mode, but we don't want to go back to previous chosen locale, but to change it for the language we've just clicked, so that's why I had to figure out some way to determine this.
For GSP part it's unnecessary, because we always change the language through `updateUserAuthLocale`.
For React part I just add something like this when changing the language being in localization mode:
```
 href={`/openboxes/user/disableLocalizationMode?localeToSet=${language.code}`}
```
I know it may cause confusion, but I couldn't figure out anything better and that's why I make the checking if this param is provided and that's why I added a comment above this.
If you have any idea how I could deal with this case, I'd love to hear from you

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
kchelstowski: As for this case probably not, because to be in `disableLocalizationMode` we have to be first in `enableLocalizationMode`, and in `enableLocalizationMode` we set the `session.locale`.
It was just a bit of overthinking by me probably, as I wanted to be "safe" that we might not have `session.locale` yet, because when you log in and do not neither change the language nor enable localization mode, your locale is taken from `session.user.locale` as `session.locale` does not yet exist.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
jmiranda: I meant changing the variable name isn't necessary and I want to avoid unnecessary changes as much as possible since it complicates the rebase process for OBGM.

jmiranda:
chooseLocale should implement the same rules. but @kchelstowski says he's implemented the rules properly in the initialization filter. there's a weird case where the state is not set until we get to the controller and therefore if there's logic that depends on that state in the filters, then we'll have one request where the logic will be wrong. I'll pull this down and check it.

kchelstowski:
@jmiranda I played with it again and I reminded myself why I couldn't just stick with the `chooseLocale` for every case: changing language, disabling/enabling localization mode. The problem is that `api` requests don't refresh the page in this case - it just dispatches an action in "a React way" and the problem is that we inject the crowdin script in the `main.gsp` file - if we are in localization mode, it injects the script, if we are not, it doesn't and for `api` requests it doesn't trigger the `main.gsp` to "reload" - it only happens if you refresh the page (which happens when you click a link that redirects us to the enable/disable localization mode function). 
```
 <g:if test="${session.useDebugLocale}">
        <script type="text/javascript">
            var _jipt = [];
            _jipt.push(['project', 'openboxes']);
        </script>
        <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
    </g:if>
```
So I believe that if we'd like to stick with the `chooseLocale` for React for each of those cases, we should find a way to inject the crowdin script inside React, because now it is handled by GSP.
Saying that, I think for now it's ready to be merged unless you have some more questions

jmiranda (approved):

PR #3740, opened 2022-12-28
OBPIH-5202 Not translatable with crowdin in Consumption report

kchelstowski:


grails-app/views/consumption/_filters.gsp
```diff
@@ -20,64 +20,83 @@
         <g:hiddenField name="parametersHash" value="${command?.generateParametersHash()}"/>
 
         <div class="box">
-            <h2><warehouse:message code="consumption.parameters.label" default="Parameters"/></h2>
+            <h2><g:message code="consumption.parameters.label" default="Parameters"/></h2>
```
jmiranda, line 23: Why are we changing theses? I thought we agreed not to change the namespace.

kchelstowski: I remember at some point you wanted to get rid of `warehouse:message` so having an occasion I decided to change it to `g:message`. If you would like to stick with `warehouse:message` for now, I can bring it back

grails-app/i18n/messages.properties
```diff
@@ -516,6 +540,10 @@ default.success.message=Success
 default.summary.label=Summary
 default.support.label=Help
 default.systemError.label=System error
+default.dataTable.noData.label=No data available in table
+default.dataTable.showingEntries.label=Showing _START_ to _END_ of _TOTAL_ entries
```
jmiranda, line 544: I neglected to comment on this before. Why are we using this syntax instead of args i.e. {0} {1}? And where / when do these parameters get replaced?

grails-app/views/consumption/_filters.gsp
jmiranda: No, I don't want to change anything with these taglibs in the develop branch because it adds more work for the rebase on the Grails 3 branch.

jmiranda: So revert all of those. If there's an issue with warehouse:message we can discuss that but please refrain from making small refactorings like this. It can add a lot of complexity to the rebase so I want to limit the changes we make to existing code.

jmiranda (changes requested):

grails-app/i18n/messages.properties
kchelstowski: These are not the same args as `{0}`, `{1}`, these are specific `datatable` params that need to be provided like this and datatable "decodes" those params and substitute proper numbers into this.
Source:
https://datatables.net/examples/i18n/options.html

grails-app/views/consumption/_filters.gsp
kchelstowski: ok, sure

drodzewicz (approved):

grails-app/views/consumption/show.gsp
```diff
@@ -26,7 +26,7 @@
 
                 <div class="box">
                     <h2>
-                        ${warehouse.message(code:'consumption.label')} <small>(${command.rows?.keySet()?.size()} results)</small>
+                        ${warehouse.message(code:'consumption.label')} <small>(${command.rows?.keySet()?.size()} ${g.message(code: 'default.resultsLowerCase.label', default: 'results')})</small>
```
jmiranda, line 29: In the future, let's not add codes for lowercase vs uppercase. You should be able to the following
```
${g.message(code: 'default.resultsLowerCase.label', default: 'results').toLowerCase()}
```
or 
```
<%@ page import="import org.apache.commons.lang3.StringUtils" %>
${StringUtils.lowerCase(g.message(code: 'default.results.label', default: 'Results'))}
```

jmiranda (approved):

PR #3741, opened 2022-12-28
OBPIH-5264 Create a custom hook for fetching logic on new list pages

alannadolny:


src/js/hooks/list-pages/product/useProductsListTableData.jsx
```diff
@@ -0,0 +1,107 @@
...
+            totalCount: res.data.totalCount,
+            pages: Math.ceil(res.data.totalCount / tableState.pageSize),
+            currentParams: params,
+          });
+          setLoading(false);
+        })
+        .catch(() => {
+          setLoading(false);
+          return Promise.reject(new Error(translate('react.productsList.fetch.fail.label', 'Unable to fetch products')));
+        });
```
kchelstowski, line 73: I know you are not the author of this, but could we maybe add `finally`, so we `setLoading(false)` only once?
Apply for each hook if the same thing is there

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
```diff
@@ -0,0 +1,145 @@
...
+        .then((res) => {
+          setLoading(false);
+          setTableData({
+            data: res.data.data,
+            pages: Math.ceil(res.data.totalCount / state.pageSize),
+            totalCount: res.data.totalCount,
+            currentParams: params,
+          });
+        })
+        .catch(() => Promise.reject(new Error(translate('react.stockMovement.outbound.fetching.error', 'Unable to fetch outbound movements'))));
```
kchelstowski, line 96: I don't see `setLoading(false)` if an error occurs? Consider using `finally`

src/js/hooks/list-pages/useTableData.jsx
```diff
@@ -0,0 +1,34 @@
...
+  const tableRef = useRef(null);
+  // Cancel token/signal for fetching data
+  const sourceRef = useRef(CancelToken.source());
+
+  const { currentLocation } = useSelector(state => ({
+    currentLocation: state.session.currentLocation,
+  }));
+
+  const fireFetchData = () => {
+    sourceRef.current = CancelToken.source();
```
kchelstowski, line 17: bring back the comment that was here

src/js/hooks/list-pages/useTableData.jsx
```diff
@@ -0,0 +1,34 @@
...
+    tableRef.current.fireFetchData();
+  };
+
+  useEffect(() => fireFetchData(), [filterParams]);
+
+  useEffect(() => {
+    if (currentLocation?.id) {
+      sourceRef.current.cancel('Fetching canceled');
+    }
+  }, [currentLocation?.id]);
```
kchelstowski, line 27: I think you removed `return` from this `useEffect` - it was supposed to be a cleanup function (`return` statement inside `useEffect`)

kchelstowski (changes requested):

kchelstowski (approved):

drodzewicz (changes requested):
I feel like we can move quite a lot of stuff into **useTable** hook.
Since every TableHook has a _loading_ state a _tableData_ state with data, pages, totalCount and currentParams we can make it more generic and move it into _useTable_ hook

```
const [loading, setLoading] = useState(true);
const [tableData, setTableData] = useState({
    data: [],
    pages: -1,
    totalCount: 0,
    currentParams: {},
});
```

same with fetching table data,a ll it does it fetches the data and set it in tabelData state which means we can just pass the url and params to the useTable hook like:
```
const url = "openboxes/api/invoice
const parmas = { ... }
const {} = useTable({ url, params })
```
I see that each list page has it own unique error when fetch is unsuccessful so this can also be an optional parameter that is passed to the _useTable_ hook like an _errorMessage_
cc @kchelstowski

src/js/hooks/list-pages/product/useProductsListTableData.jsx
```diff
@@ -0,0 +1,79 @@
...
+import queryString from 'query-string';
+
+import useTableData from 'hooks/list-pages/useTableData';
+import apiClient from 'utils/apiClient';
+
+const useProductsListTableData = (filterParams) => {
+  const url = '/openboxes/api/products';
+  const messageId = 'react.productsList.fetch.fail.label';
+  const defaultMessage = 'Unable to fetch products';
+  const getSortingParams = tableState => (tableState.sorted.length > 0 ?
```
drodzewicz, line 12: I also feel like `getSortingParams` can be moved to **useTableData**.
The logic for this seems very generic across all of the _use*TableData_ hooks and you would only need to pass the `sortColumn: 'lasUpdated'` and  `sortOrder: 'desc'` as params

drodzewicz (changes requested):

drodzewicz (approved):

src/js/hooks/list-pages/inbound/useInboundListTableData.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+import { fetchShipmentStatusCodes, hideSpinner, showSpinner } from 'actions';
+import stockMovementApi from 'api/services/StockMovementApi';
+import { STOCK_MOVEMENT_API, STOCK_MOVEMENT_INCOMING_ITEMS } from 'api/urls';
+import useTableData from 'hooks/list-pages/useTableData';
+import exportFileFromAPI from 'utils/file-download-util';
+import { translateWithDefaultMessage } from 'utils/Translate';
+
+const useInboundListTableData = (filterParams) => {
+  const messageId = 'react.stockMovement.inbound.fetching.error';
```
kchelstowski, line 17: it should be named as `errorMessageId`

src/js/hooks/list-pages/inbound/useInboundListTableData.jsx
```diff
@@ -0,0 +1,123 @@
...
+              filterParams.receiptStatusCode?.map(({ id }) => id),
+    origin: filterParams?.origin?.id,
+    destination: filterParams?.destination?.id,
+    requestedBy: filterParams.requestedBy?.id,
+    createdBy: filterParams.createdBy?.id,
+    updatedBy: filterParams.updatedBy?.id,
+    ...sortingParams,
+  }, (value) => {
+    if (typeof value === 'object' && _.isEmpty(value)) return true;
+    return !value;
```
kchelstowski, line 33: I think that this function's arguments should be as `( { offset, currentLocation, state, sortingParams })`, because having so many arguments it might be difficult to follow the order of arguments. Making this change, you'll allow yourself not to pass them in the order that you have now.
Moreover I think that for better readability, it would be good to destructure `filterParams`, having:
```
const { receiptStatusCode, origin, destination, requestedBy, createdBy, updatedBy } = filterParams
```
and proceed with that, so we have:
```
origin: origin?.id
```

kchelstowski (changes requested):

src/js/hooks/list-pages/useTableData.jsx
```diff
@@ -0,0 +1,99 @@
...
+        cancelToken: sourceRef.current?.token,
+      })
+        .then((res) => {
+          setTableData({
+            data: res.data.data,
+            totalCount: res.data.totalCount,
+            pages: Math.ceil(res.data.totalCount / tableState.pageSize),
+            currentParams: params,
+          });
+          if (onFetchedData) onFetchedData(res.data);
```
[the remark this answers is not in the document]
alannadolny: I didn't know that optional chaining works with functions. Thanks for that

src/js/hooks/list-pages/inbound/useInboundListTableData.jsx
```diff
@@ -14,24 +14,32 @@ import exportFileFromAPI from 'utils/file-download-util';
 import { translateWithDefaultMessage } from 'utils/Translate';
 
 const useInboundListTableData = (filterParams) => {
-  const messageId = 'react.stockMovement.inbound.fetching.error';
+  const errorMessageId = 'react.stockMovement.inbound.fetching.error';
   const defaultMessage = 'Unable to fetch inbound movements';
```
kchelstowski, line 18: changing from `messageId` to `errorMessageId` as I requested, you should also change the `defaultMessage` naming to `defaultErrorMessage`

kchelstowski (changes requested):

src/js/hooks/list-pages/useTableData.jsx
```diff
@@ -77,9 +81,10 @@ const useTableData = ({
             pages: Math.ceil(res.data.totalCount / tableState.pageSize),
             currentParams: params,
           });
-          if (onFetchedData) onFetchedData(res.data);
+          // eslint-disable-next-line no-unused-expressions
```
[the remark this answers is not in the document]
alannadolny: Without this comment eslint shows that there is unused expression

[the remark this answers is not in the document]
alannadolny: Screenshot from 2023-01-05 10-54-34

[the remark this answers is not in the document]
kchelstowski: bring it back to `if` version then, but remember about brackets I mentioned earlier

kchelstowski (approved):

drodzewicz (approved):

PR #3743, opened 2022-12-30
OBPIH-5119 add exact product when type product code and press enter

drodzewicz:


src/js/components/form-elements/ProductSelectField.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import BaseField from 'components/form-elements/BaseField';
+import ProductSelect from 'components/product-select/ProductSelect';
+
+const ProductSelectField = (props) => {
+  // eslint-disable-next-line react/prop-types
```
kchelstowski, line 7: did you leave it on purpose?

src/js/components/product-select/ProductSelect.jsx
```diff
@@ -0,0 +1,110 @@
...
+  <strong style={{ color: option.color || 'black' }} className="d-flex align-items-center">
+    {option.label}
+  &nbsp;
+    {renderHandlingIcons(option.handlingIcons)}
+  </strong>);
+
+const SelectedValue = option => (
+  <span className="d-flex align-items-center">
+    <span className="text-truncate">{option.label}</span>
+    &nbsp;{renderHandlingIcons(option ? option.handlingIcons : [])}
```
kchelstowski, line 20: just want to show a different option:
```
option?.handlingIcons ?? []
```

src/js/components/product-select/ProductSelect.jsx
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  const onEnterPress = (event) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setIsExactMatch(true);
+  };
+
+  useEffect(() => {
+    if (isExactMatch && loadedOptions.length > 0) {
```
kchelstowski, line 50: could be simplified to 
```
isExactMatch && loadedOptions.length
```

src/js/components/product-select/ProductSelect.jsx
```diff
@@ -0,0 +1,110 @@
...
+      const exactMatchProduct = loadedOptions.find(product => product.exactMatch);
+
+      if (onExactProductSelected) {
+        onExactProductSelected({ isExactMatch, product: exactMatchProduct });
```
kchelstowski, line 54: what is that `isExactMatch` provided here for? From what I can see in `fields` config, we are only destructuring `product`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -1091,15 +1042,15 @@ class AddItemsPage extends Component {
...
+                  type="button"
                   disabled={invalid || showOnly}
                   onClick={() => this.previousPage(values, invalid)}
                   className="btn btn-outline-primary btn-form btn-xs"
                 >
                   <Translate id="react.default.button.previous.label" defaultMessage="Previous" />
                 </button>
                 <button
-                  type="submit"
+                  type="button"
```
kchelstowski, line 1053: Why do you change those? I just want to be in touch with this change, as it is not first time I see you changing those.
I guess that keeping it as `submit` doesn't just have any sense, since we provide `onClick` and our `onSubmit` does it: `() => {}`
Am I right?

src/js/utils/Select.jsx
```diff
@@ -276,7 +276,10 @@ class Select extends Component {
             isClearable={clearable}
             title=""
             delimiter={delimiter}
-            onMenuClose={multi ? () => this.sortOptionsByChecked(options, value) : null}
+            onMenuClose={() => {
+              if (multi) this.sortOptionsByChecked(options, value);
+              if (this.props.onMenuClose) this.props.onMenuClose();
```
kchelstowski, line 281: what is that change for?

kchelstowski (changes requested):

src/js/components/form-elements/ProductSelectField.jsx
drodzewicz: I don't think it's necessary anymore, I'll remove that

src/js/components/product-select/ProductSelect.jsx
drodzewicz: yeah sure, I just copied and pasted this chunk of code so didn't really notice that

src/js/components/product-select/ProductSelect.jsx
drodzewicz: I thought maybe this value might be useful in some way but I guess I didn't use it anywhere, so lets stick with the YAGNI principle  :D  and I'll remove that

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
drodzewicz: that's one part of it but also since I was implementing a on enter behavior on _product select input_ and this component is inside of a  _form_ tag it  would get that enter event and perform a click on button that is type submit

src/js/utils/Select.jsx
drodzewicz: I needed to perform some kind of an action _onMenuClose_ and currently Select component didn't have an option to pass my own callback for it to run on menu close.
This doesn't change anything except runs a `props.onMenuClose` if it is passed

jmiranda:
Let's discuss this one when I get back on Tuesday (or at the next tech huddle on Wednesday).

kchelstowski (approved):

jmiranda (approved):

PR #3744, opened 2022-12-30
OBPIH-5266 Create API architecture for React side and assign API urls to consts/json instead of using hardcoded api urls for PO list

alannadolny:


src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -125,17 +123,15 @@ const PurchaseOrderListTable = ({
...
-      });
+  const rollbackOrder = async (id) => {
+    const { status } = await purchaseOrderApi.rollbackOrder(id);
+    if (status === 200) {
+      Alert.success(translate(
+        'react.purchaseOrder.rollback.success.label',
+        'Rollback of order status has been done successfully',
+      ));
+      fireFetchData();
+    }
```
kchelstowski, line 134: I know that you are not the author, but I believe there also should be spinner added. Would you mind adding it?

kchelstowski (changes requested):
I'm really impressed by this approach. I'm glad we figured out such an idea. Can't wait to see it in more places

drodzewicz (approved):
looks good to me

kchelstowski (approved):

PR #3746, opened 2023-01-02
OBPIH-5267 Create API architecture for React side for rest list pages

alannadolny:


src/js/api/services/StockListApi.js
```diff
@@ -0,0 +1,17 @@
+import {
+  STOCK_LIST_API,
+  STOCK_LIST_CLEAR,
+  STOCK_LIST_CLONE,
+  STOCK_LIST_DELETE,
+  STOCK_LIST_PUBLISH, STOCK_LIST_UNPUBLISH,
```
kchelstowski, line 6: we should keep stocklist as `STOCKLIST` as it is a proper name.
so:
```
STOCKLIST_API
```
etc.

kchelstowski (changes requested):

jmiranda:
@drodzewicz can you review this PR?

jmiranda:
@drodzewicz nevermind, @kchelstowski says this one is ready to merge.

awalkowiak:
This is good, but I think this will be a pain in the ass while rebasing grails 3 on top of this (but I'll manage it, hopefuly). Definitely this refactor should be done before starting the grails 3.

PR #3747, opened 2023-01-02
OBPIH-4938 Add highlighting for the search input (fixes after QA)

alannadolny:


kchelstowski (changes requested):
LGTM, but one of the unit tests (that you had written not a long time ago) failed. Could you take a look?

alannadolny:
It's because of a little change in className of the input field (snapshot test). I added a new class which inform if input field has value. If there is no value the class name ends with additional space. I will fix it tommorrow.
obraz

kchelstowski (approved):

PR #3749, opened 2023-01-02
OBPIH-5209 Add missing crowdin translations for inventory details report

drodzewicz:


kchelstowski (approved):

jmiranda (approved):
Nice cleanup on the statuses list. 

But be careful with a refactoring like this as it unnecessarily complicates the OBGM rebase process. So only make these refactorings if they are 100% necessary.

PR #3750, opened 2023-01-03
OBPIH-5216 Add missing translations to EXPIRING stock raport

kchelstowski:


grails-app/views/inventory/listExpiringStock.gsp
```diff
@@ -171,7 +171,7 @@
                                                     ${quantityMap[inventoryItem]}
                                                 </td>
                                                 <td class="checkable center" >
-                                                    ${inventoryItem?.product?.unitOfMeasure?:"EA" }
+                                                    <g:message code="default.uom.${inventoryItem?.product?.unitOfMeasure?.toLowerCase()}.uppercase.label" default="EA"/>
```
drodzewicz, line 174: I am not sure that this is how we want to translate _unit of measure_ variables.
We still have yet to discuss the best approach for this on one of the upcoming tech huddles. 
We also have an open ticket for UOM translations **OBPIH-5187**

kchelstowski: thanks for pointing that out, I haven't known about this ticket before. So I'm gonna revert this change for now, and it will be covered later.

drodzewicz (approved):

grails-app/views/inventory/listExpiringStock.gsp
jmiranda: Thanks for catching that @drodzewicz. As I mentioned in a PR earlier today we should never have a label for uppercase vs lowercase. We can wrap these values with StringUtils.lowerCase() vs StringUtils.upperCase() or StringUtils.capitalize().

PR #3753, opened 2023-01-03
OBPIH-5087 Add validation on 0 and negative numbers in putaways and bin replenishment

alannadolny:


src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -131,21 +131,36 @@ class PutAwaySecondPage extends Component {
...
+        let disabledMessage;
+
+        if (props.value < 1) {
+          disabledMessage = this.props.translate(
+            'react.putAway.negativeQuantity.label',
+            'Quantity cannot be lower than 1',
+          );
+        }
+
+        if (props.original && props.value > props.original.quantityAvailable) {
```
kchelstowski, line 143: can be shorten to:
```
if (props.value > props.original?.quantityAvailable)
```

src/css/main.scss
```diff
@@ -445,6 +445,14 @@ body {
...
   color: $red-beta;
 }
+div.has-error {
+  input {
+    border: 2px solid $red-beta;
+    -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075);
+    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075);
+    border-radius: 4px;
+  }
+}
```
kchelstowski, line 455: why did you add those styles? Don't we have them already?

src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -131,21 +131,36 @@ class PutAwaySecondPage extends Component {
...
+              html={disabledMessage}
+              disabled={!disabledMessage}
               theme="transparent"
               arrow="true"
               delay="150"
               duration="250"
               hideDelay="50"
             >
-              <div className={props.value > props.original.quantityAvailable ? 'has-error' : ''}>
+              <div className={props.value > props.original.quantityAvailable ||
```
kchelstowski, line 161: I'd split this logic to a variables:
```
const isQuantityHigherThanOriginal = props?.value > props.original?.quantityAvailable
const isQuantityLowerThan1 = props?.value < 1
```
and use it above and also here inside `<div>`
```
<div className={(isQuantityHigherThanOriginal || isQuantityLowerThan1) ? 'has-error' : ''}>
```

src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -164,18 +179,15 @@ class PutAwaySecondPage extends Component {
...
+            html={disabledMessage}
+            disabled={!disabledMessage}
             theme="transparent"
             arrow="true"
             delay="150"
             duration="250"
             hideDelay="50"
           >
-            <div className={props.original && props.value > props.original.quantityAvailable ? 'has-error' : ''}>
+            <div className={(props.original && props.value > props.original.quantityAvailable) || props.value < 1 ? 'has-error' : ''}>
```
kchelstowski, line 190: apply the same logic from above here

src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -625,6 +638,8 @@ class PutAwaySecondPage extends Component {
         }
         <div className="submit-buttons">
           <button
+            disabled={_.some(this.state.putAway.putawayItems, putawayItem =>
+                (putawayItem.quantity > putawayItem.quantityAvailable) || putawayItem.quantity < 1)}
```
kchelstowski, line 642: I can see a similar logic here, but I assume it is not inside the `Cell`, so how about creating a function and pass there two arguments - the function that could be used for this case and above cases?

src/js/components/put-away/SplitLineModal.jsx
```diff
@@ -132,10 +151,18 @@ class SplitLineModal extends Component {
...
    */
-  isValid() {
+
+  isNegativeQuantity() {
+    return _.some(this.state.splitItems, items => _.toInteger(items.quantity) <= 0);
+  }
+
+  isQuantityHigherThanOriginalPutaway() {
     const qtySum = this.calculatePutAwayQty();
+    return qtySum > _.toInteger(this.props.putawayItem.quantity);
```
kchelstowski, line 161: just want to follow - do you have to use `_.toInteger`, because the quantity is in a string format?

src/js/components/put-away/SplitLineModal.jsx
```diff
@@ -132,10 +151,18 @@ class SplitLineModal extends Component {
...
+  }
 
-    return qtySum <= _.toInteger(this.props.putawayItem.quantity);
+  isValid() {
+    return !this.isNegativeQuantity() && !this.isQuantityHigherThanOriginalPutaway();
```
kchelstowski, line 165: do you use it anywhere? I see that in line `228` you delete `disabled={this.isValid()}` where it was used and I can't see any usage of this anymore?

kchelstowski (changes requested):

src/css/main.scss
alannadolny: We had them, but in the putaway modal (split line button) styles should be applied on .has-error > input (this styles didn't work before). I am adding this tag to styles below, but border-radius has to be added separately.

src/js/components/put-away/SplitLineModal.jsx
alannadolny: Yes, the quantity is in a string format

src/js/components/put-away/SplitLineModal.jsx
alannadolny: I used it in line 328 `disabled={!this.isValid() || !this.isBinSelected()}`

kchelstowski (approved):

grails-app/i18n/messages.properties
```diff
@@ -3322,7 +3322,9 @@ react.putAway.splitLine.label=Split line
 react.putAway.totalQty.label=Total QTY
 react.putAway.putAwayQty.label=Putaway QTY
 react.putAway.sumOfAll.label=Sum of all split items quantities cannot be higher than original putaway item quantity
+react.putAway.negativeSumOfAll.label=Items quantity cannot be lower than 1
```
jmiranda, line 3325: We don't have to do this in this PR, but we should change all language around higher vs lower to greater than vs less than.

src/js/components/put-away/SplitLineModal.jsx
kchelstowski: hard to say in this specific place why it is a string, but I know that in some places we have quantity as string. My blind guess is that initially the quantity is type of number (fetched from DB), but then whenever you edit the value, it becomes the string, that's why I believe in many places we have some kind of validation, if the value given is not a `NaN`. I think that @alannadolny will explain this specific case, as it is just my blind guess. Am I right @alannadolny ?

src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -24,6 +24,10 @@ import 'react-table/react-table.css';
 
 const SelectTreeTable = (customTreeTableHOC(ReactTable));
 
+function isInvalid(quantity, quantityAvailable) {
```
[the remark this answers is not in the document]
kchelstowski: yeah I agree, something didn't feel right to me about the naming also, but I didn't want to be nitpicky, but we should definitely follow the same convention everywhere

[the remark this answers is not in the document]
jmiranda: You only have to be nitpicky until everyone is on the same page. Then it becomes convention / second nature.

src/js/components/put-away/SplitLineModal.jsx
alannadolny: Yeah @kchelstowski you're right. The fetched quantity is type of number, after editing the value, it becomes the string:
Screenshot from 2023-01-04 23-53-20
Someone set types of this quantity in propTypes and set it to string and number before
Screenshot from 2023-01-05 00-00-04

drodzewicz (approved):

src/js/components/put-away/SplitLineModal.jsx
```diff
@@ -64,6 +64,25 @@ class SplitLineModal extends Component {
     }
   }
 
+  getErrorMessage() {
+    if (this.isQuantityHigherThanOriginalPutaway()) {
```
drodzewicz, line 68: should probably be `isQuantityGreaterThanOriginalPutaway`

drodzewicz (changes requested):

kchelstowski (approved):

drodzewicz (approved):

PR #3754, opened 2023-01-04
OBPIH-4903 Allow filter by Product Code in PO View Page

kchelstowski:


drodzewicz (approved):

grails-app/i18n/messages.properties
```diff
@@ -1577,7 +1577,7 @@ order.pending.label=Pending
 order.pendingOrderLog.label=Pending orders
 order.placeOrder.label=Place order
 order.print.label=Print order
-order.filterByName.label=Filter by product name
+order.filterByNameOrCode.label=Filter by product name or code
```
jmiranda, line 1580: Just so we don't have to keep changing it, let's call this filterByProduct.

kchelstowski: honestly, I made this change on purpose to see, if our crowdin's PR will be triggered, but ok, I'll revert that

alannadolny (approved):

jmiranda (approved):

PR #3756, opened 2023-01-04
OBPIH-5234 return most recent shipment when performing a global search

drodzewicz:


grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -78,6 +78,10 @@ class DashboardController {
 
         List<Shipment> shipments = Shipment.findAllByShipmentNumber(params.searchTerms, [sort: 'dateCreated', order: 'desc'])
         if (shipments) {
+            // sort in descending order list of shipments based on a shipment dateCreated value or returnOrder dateCreated of that shipments
+            shipments = shipments.sort {a,b ->
+                (b.returnOrder?.dateCreated ?: b.dateCreated) <=> (a.returnOrder?.dateCreated ?: a.dateCreated)
```
jmiranda, line 83: Should this be 
```
shipments = shipments.sort {a,b ->
    b.returnOrder?.dateCreated <=> a.returnOrder?.dateCreated ?:
        b.dateCreated <=> a.dateCreated
}

Comment (drodzewicz) [reply grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy:83]: But if I am not mistaken that would mean that shipments with _returnOrders_ would always end up higher than shipment _without returnOrder_

given this list of shipments
```
[
    shipmentA: {
        dateCreated: '2022-12-03',
        returnOrder: { dateCreated: '2022-12-04' },
    },
    shipmentB: {
        dateCreated: '2022-12-02',
        returnOrder: { dateCreated: '2022-12-03' },
    },
    shipmentC: {
        dateCreated: '2022-12-04',
        returnOrder: null,
    },
    shipmentD: {
        dateCreated: '2022-12-05',
        returnOrder: null,
    },
]
```

when comparing _shipmentB_ with _shimpmentC_ this logic would look something like that:
assuming that **a** is _shipmentB_ and **b** is _shimpmentC_

```
// b.returnOrder?.dateCreated <=> a.returnOrder?.dateCreated
null <=> '2022-12-03' results in -1
```
that would probably sort them like that: _shipmentA, shipmentB, shipmentD, shipmentC_

I guess one of the key questions is what is the correct "most recent stockMovement "

Comment (jmiranda) [reply grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy:83]: Sorry I was just following your original code not really thinking through the logic. So you could just switch it around, I guess.
```
shipments = shipments.sort {a,b -> 
    b.dateCreated <=> a.dateCreated ?:
        b.returnOrder?.dateCreated <=> a.returnOrder?.dateCreated
}
```

But actually I'd prefer we just use shipment.dateCreated since dateCreated should not be null. 

Therefore we can make the following change and call it a day.
```
List<Shipment> shipments = Shipment.findAllByShipmentNumber(params.searchTerms, [sort: 'dateCreated', order: 'asc'])
```

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
drodzewicz: > 
> But actually I'd prefer we just use shipment.dateCreated since dateCreated should not be null.
> 
> Therefore we can make the following change and call it a day.
> 
> ```
> List<Shipment> shipments = Shipment.findAllByShipmentNumber(params.searchTerms, [sort: 'dateCreated', order: 'asc'])
> ```
This is how I initially have done it and it is the current behavior so I guess this PR can be closed.

I have informed Katarzyna that when testing this ticket in case of stock movements of type return order she should compare the dates created of the shipments.
She has already tested it again and has approved it, so no changes are needed anymore.

Sorry for the late reply on this PR it kinda forgot about it

PR #3760, opened 2023-01-05
OBPIH-5202 Fix consumption, bin location, transation, in stock reports

kchelstowski:
What had to be done for the "Showing X to Y of entries" was unfortunately to concat a few strings together, because `dataTable` got crazy when its arguments `_START_` etc, were returned by `g.message`, so those have to be hardcoded, this is why I had to translate each string of this message separately.
As for table not fitting the screen, it was just that crowdin codes are very long and as on consumption report we have a lot of columns, they couldn't fit the screen. What I've done is just simply setting the `max-width` of the column and wrapping the word if it is wider than `100px`.

I applied the fix for the "Showing X to Y of entries" also for other reports that were in other tickets, but the same thing was not working there.


drodzewicz (approved):

jmiranda (approved):
I don't really like that we have to do this, but I don't feel like spending any more time on these translation issues.

PR #3767, opened 2023-01-10
OBPIH-5207 Fix crowdin translations on Dashboard

drodzewicz:
- Add missing translations for Dashboards sidebar items
- Add missing translation For labels on Archive indicators, save config etc.
- Fix truncated label causing crowding not to translate it properly
- Modify indicator filter component option label rendering to instead accept arguments in the label for more control on the order of words


src/css/main.scss
```diff
@@ -487,6 +487,10 @@ a.disabled {
   color: $gray-delta;
 }
 
+.text-align-left {
+  text-align: left;
+}
+
```
kchelstowski, line 493: don't we have `text-left` in bootstrap?

kchelstowski (changes requested):

kchelstowski (approved):

PR #3768, opened 2023-01-10
Updated grant command to include create user command, as some version…

trs998:
…s of MySQL grant cannot create users.

Mysql command grant could be used to create users, but this use was deprecated and now doesn't work on current versions of MySQL - updated command twill work in them all.

### DIFF
--- a/docs/installation/ubuntu1804/mysql.md
+++ b/docs/installation/ubuntu1804/mysql.md
@@ -17,7 +17,8 @@ $ mysql -u root -p -e 'create database openboxes default charset utf8;'
 
 ## Grant permissions to new database user
 ```
-$ mysql -u root -p -e 'grant all on openboxes.* to openboxes@localhost identified by "<password>";'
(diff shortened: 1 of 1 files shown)


jmiranda (commented):
These are commands that work in MySQL 8. Need to confirm that these same commands work for MySQL 5.7.

docs/installation/ubuntu1804/mysql.md
```diff
@@ -17,7 +17,8 @@ $ mysql -u root -p -e 'create database openboxes default charset utf8;'
 
 ## Grant permissions to new database user
 ```
-$ mysql -u root -p -e 'grant all on openboxes.* to openboxes@localhost identified by "<password>";'
+$ mysql -u root -p -e 'CREATE USER openboxes@localhost identified by "<password>";'
```
mdpearson, line 20: You may want to use `CREATE USER IF NOT EXISTS` here

awalkowiak:

PR #3769, opened 2023-01-10
Minor bugs in documentation

trs998:
Updated file name for log file - /opt/tomcat is a symlink to whatever version of tomcat got installed, so /opt/tomcat/tomcat-<version> does not make sense - corrected by removing extraneous component.
Also fixed documentation bug where java path ended up incorrect.


jmiranda:
Thank you @trs998

awalkowiak:

PR #3770, opened 2023-01-10
OBPIH-5273 Dashboard is underlined in menu on some pages

alannadolny:


grails-app/views/dashboard/megamenu.gsp
```diff
@@ -99,14 +99,43 @@
 
 <script type="text/javascript">
   $(document).ready(function () {
-      const path = window.location.pathname
-      const menuConfigValues = $(".menu-config-value").toArray();
-      const matchingSection = menuConfigValues.find(it => it.value.includes(path))
+    function getHrefs(value) {
```
drodzewicz, line 102: this function seems a little too complicated which makes it very hard to read and understand what is going on.
I think that we should make it more explicit and straight forward. 
My proposition on how to deal with certain pages that are not underlined in the menu for GSP pages is to add a script in eg. show.gsp for stocklists
```
<script type="text/javascript">
  $(document).ready(function () {
    const matchingMenuSection = $("#requisitionTemplate')
    const matchingMenuSectionCollapsable = $("#requisitionTemplate-collapsed").get(0);
    if (matchingMenuSection) matchingMenuSection.classList.add('active-section');
    if (matchingMenuSectionCollapsable) matchingMenuSectionCollapsable.classList.add('active-section');
  })
</script>
``` 
that way when this show page is going to be rendered the script is going to add an active class to the appropriate section

of course you would have to implement this script in all of the pages that require it but I don't think this will be a big problem since it's a simple logic and I would prefer some duplicate chunks of  code instead of a complicated hard to read function

grails-app/conf/Config.groovy
```diff
@@ -1340,7 +1340,7 @@ openboxes {
                         [label: "inventory.import.label", defaultLabel: "Import Inventory", href: "/${appName}/batch/importData?type=inventory&execution=e1s1"],
                         [label: "inventory.createStockTransfer.label", defaultLabel: "Create Stock Transfer", requiredActivitiesAll: ActivityCode.binTrackingList(), href: "/${appName}/stockTransfer/create"],
                         [label: "inventory.listStockTransfers.label", defaultLabel: "List Stock Transfers", requiredActivitiesAll: ActivityCode.binTrackingList(), href: "/${appName}/stockTransfer/list"],
-                        [label: "inventory.createReplenishment.label", defaultLabel: "Create Replenishment", requiredActivitiesAll: ActivityCode.binTrackingList(), href: "/${appName}/replenishment/index"]
+                        [label: "inventory.createReplenishment.label", defaultLabel: "Create Replenishment", requiredActivitiesAll: ActivityCode.binTrackingList(), href: "/${appName}/replenishment/create"]
                     ]
```
drodzewicz, line 1344:
```
[label: "order.createPurchase.label", defaultLabel: "Create Purchase Order", href: "/${appName}/purchaseOrder/index", requiredActivitiesAny: [ActivityCode.PLACE_ORDER]],
```
Create purchase order should also probably be fixed 
from `purchaseOrder/index` to `purchaseOrder/create`

src/js/utils/menu-utils.jsx
```diff
@@ -0,0 +1,75 @@
...
+        if (!pathnameWithoutParams.includes(sectionPath.replace(/\/index$/, ''))) {
+          return false;
+        }
+        // if found matching pathname
+        // then check if all parameters of section path match with current path parameters
+        if (sectionSearch) {
+          const {
+            direction,
+            ...otherParams
+          } = mapQueryParamsToObject(search.substring(1, search.length));
```
drodzewicz, line 36: use query-string parse instead
```
import queryString from 'query-string';

const parsed = queryString.parse(search);
```

grails-app/views/order/addComment.gsp
```diff
@@ -78,5 +78,6 @@
 			</div>
 		</div>
 	</div>
+	<g:render template="/dashboard/activeSection" model="[section: orderInstance.orderType.name == 'Putaway Order' ? 'inbound' : orderInstance.orderType.name == 'Outbound' ? 'outbound' : 'purchasing']"/>
```
kchelstowski, line 81: Rather use:
```
orderInstance?.orderType?.code == Constants.PUTAWAY_ORDER
```
or even:
```
orderInstance?.orderType?.isPutawayOrder()
```
By the way I'm also concered about comparing order to `Outbound` - we don't have `orderType` 'Outbound'.

grails-app/views/order/addDocument.gsp
```diff
@@ -111,5 +111,6 @@
 			</div>
 		</div>
 	</div>
+	<g:render template="/dashboard/activeSection" model="[section: orderInstance.orderType.name == 'Putaway Order' ? 'inbound' : orderInstance.orderType.name == 'Outbound' ? 'outbound' : 'purchasing']"/>
```
kchelstowski, line 114: same as above

grails-app/views/order/show.gsp
```diff
@@ -303,5 +303,6 @@
               });
             }
         </script>
+    <g:render template="/dashboard/activeSection" model="[section: orderInstance.orderType.name == 'Putaway Order' ? 'inbound' : orderInstance.orderType.name == 'Outbound' ? 'outbound' : 'purchasing']"/>
```
kchelstowski, line 306: same as above

kchelstowski (changes requested):

src/js/components/Layout/menu/Menu.jsx
```diff
@@ -32,7 +35,7 @@ const Menu = ({ menuConfig, location }) => {
                 <MenuSection
                   section={section}
                   key={`${section.label}-menu-section`}
-                  active={activeSection === section.label}
+                  active={activeSection === section.id}
```
kchelstowski, line 35: nice catch

kchelstowski (approved):

jmiranda:
I need a bit more time on this one.

jmiranda (commented):
I don't think we want to be passing in this section information on every GSP when we already have it available in the Config (i.e. search href by the current URL to get the active section). Let's discuss this one on the tech huddle.

jmiranda:
Tech Huddle discussion
* Make as many work implicitly without having to inject the active section
    * inventoryItem -> inventory
    * product -> product
    * order -> purchasing
* Stock movement can be hard-coded / injected for now 
* All other exceptions can wait to another release
* activeSection.gsp should be included in megamenu.gsp or custom.gsp 
* Consider using sitemesh to handle the active section value

When you need to override, use the following convention
stockMovement/show.gsp (add to <head>)
```
<g:set var="activeSection">$stockMovement?.stockMovementType?.toLowecase()</g:set>
<content tag="activeSection">${activeSection?:controllerName}</content>
```

grails-app/views/dashboard/megamenu.gsp
```diff
@@ -101,14 +102,33 @@
...
     const menuConfigValues = $(".menu-config-value").toArray();
-    const matchingSection = menuConfigValues.find(it => it.value.includes(path))
-    // Assign active-section class to matched section
-    if (matchingSection) {
-      const matchingMenuSection = $("#" + matchingSection?.name).get(0);
-      const matchingMenuSectionCollapsable = $("#" + matchingSection?.name + "-collapsed").get(0);
-      if (matchingMenuSection) matchingMenuSection.classList.add('active-section');
-      if (matchingMenuSectionCollapsable) matchingMenuSectionCollapsable.classList.add('active-section');
-    }
+    const urlPartsWithSectionConfig = "${(grailsApplication.config.openboxes.menuSectionsUrlParts as JSON)}";
```
kchelstowski, line 105: I feel like it could be better to return it from the backend, from the `def megamenu` just adding another property to return after `menu`, `configurationSection`, `megamenuConfig`:
```
menuSectionsUrlParts: grailsApplication.config.openboxes.menuSectionsUrlParts as JSON
```

and to create a hidden field with id/class, that would allow you to access it from the js script side

kchelstowski (changes requested):
Don't do any changes until @awalkowiak approves my requests

grails-app/views/dashboard/megamenu.gsp
awalkowiak: @kchelstowski Yup, good suggestion

src/js/utils/menu-utils.jsx
```diff
@@ -0,0 +1,80 @@
+import _ from 'lodash';
+import queryString from 'query-string';
+
+export const checkActiveSection = ({
```
awalkowiak, line 4: @alannadolny could this util also be simplified to use the new `sections config`?

awalkowiak (commented):
One question about the util on the react side plus take a look at Kacpers suggestion

grails-app/views/dashboard/megamenu.gsp
```diff
@@ -99,33 +100,49 @@
...
+      const exactMatch = menuConfigValues.find(it => it.value.includes(path))
+
+      if(exactMatch) {
+        return exactMatch;
       }
 
+      // check if current section exists in object with wrongly underlined urls
+      const sectionFromJson = Object.keys(parsedUrlPartsWithSection).find(sectionName => {
+        return !!parsedUrlPartsWithSection[sectionName].some(section => path.includes(section));
+      })
```
drodzewicz, line 119: Alternatively if we want to avoid this parsing of LinkedHasMap to JSON we can do everything using groovy.
By everything I mean finding the matched section based on `config.openboxes.menuSectionUrlParts`

So instead of doing it with javascript we can do it with groovy

```
      <g:set
        var="matchedSection"
        value="${grailsApplication.config.openboxes.menuSectionsUrlParts.find{ key, val -> val.any{ request?.forwardURI?.contains(it)} }}"
     />
      return menuConfigValues.find(it => it.name === "${matchedSection?.key}");
```

@alannadolny @kchelstowski  @awalkowiak  what do you think ?

src/js/utils/menu-utils.jsx
alannadolny: I don't think so, because this util has to compare sectionSearch, not only sectionPath. There are some tricky places where the sectionSearch can change the section which should be underlined.

grails-app/views/dashboard/megamenu.gsp
alannadolny: The solution that Kacper propsed, fixed the problem with replacing &quotes to ". 
So now it looks like: 
`   const urlPartsWithSection = $("#menuSectionUrlParts").val();
    const parsedUrlPartsWithSection = JSON.parse(urlPartsWithSection);`
I think it is much cleaner than doing it in groovy

kchelstowski: @drodzewicz I love that idea, but it depends on if we want Alan to spend some more time on this one

awalkowiak (approved):

kchelstowski (approved):

PR #3771, opened 2023-01-10
OBPIH-5303 Fix double fetch on list pages after changing the location

kchelstowski:
The problem occured, because two `useEffect` actions were triggering itself, but as both of them were doing something else in addition and both of them need to be triggered after changing a location (one to clear the form values and one to clear all the states, params, query params) I needed to come up with some idea how to make it common.
What I've done is adding a filterFormReducer and some kind of triggering based on the boolean flag set in this reducer called `shouldRebuildFilterValues` - I decided to go into reducer, because I can use one logic for all list pages instead of doing the same logic on each list page.
The only list page that had some differences in `ifs` for building filter values logic was the PO list, so I created a separate hook for this list as we are tight on time and I didn't want to spend much time figuring out how to include the PO logic in a common hook I've created.


drodzewicz (approved):
From what see it makes sense to me so this solution should work fine.
I have also tested it locally and all of the list behavior look right to me.
Maybe for the future we should look into this separate `usePurchaseOrderFilterCleaner` to maybe generalize it into one hook `useCommonFilterCleaner`

jmiranda:
Added @awalkowiak as a reviewer for when he gets back since this is a refactoring that he should approve.

PR #3774, opened 2023-01-11
Updated documentation to recommend adding a uploads folder

trs998:
Added a line to the suggested configuration file to put the uploads folder in a functional location and a note that this can be changed. Without this line the default is /uploads which tomcat cannot write to (and while this could be created manually, this is a poor location for segregation of multiple services on a system)


jmiranda:
Discuss on tech huddle. Should these documentation changes go to master, release, or develop?

awalkowiak:

PR #3781, opened 2023-01-17
OBPIH-5307 Pagination issue on browse inventory list page

kchelstowski:
The problem is more complexed. I tried many different approaches including even pure SQL, but the problem is that I was using `product_availibility` table as a source, but it turned out, that in this table there might not be all of the products.
The reason I tried to use `product_availibility` table is that we need the information of QOH. 
We can't access the QOH through the `createCriteria` of `Product` and on the other hand we can't access **ALL** products when trying to do the query via `ProductAvailibility`.
The results were filtered after being paginated and that's why the problem occured.
For now I pushed a version where unfortunately we don't paginate the results in the `createCriteria()`, but just get all of the items and we cover the pagination "by hand".

To be able to access QOH, Artur's advice was to create a view for browse inventory which would have QOH and Product and see the performance of that in comparison to accessing it via:
```
 def quantityMap = products ? getQuantityOnHandByProduct(command.location, products) : []
```
The `getQuantityOnHandByProduct` method:
```
Map<Product, Integer> getQuantityOnHandByProduct(Location location) {
        def quantityMap = [:]
        if (location) {
            def results = ProductAvailability.executeQuery("""
						select pa.product, sum(pa.quantityOnHand)
						from ProductAvailability pa
						where pa.location = :location
						group by pa.product
						""", [location: location])
            results.each {
                quantityMap[it[0]] = it[1]
            }
        }

        return quantityMap
    }
```

The perfomance is still bad even if I added the view, but I believe it could be better if we were able to access the `product` properties from the `createCriteria()` for `BrowseInventory` domain - I explained it in the code that I couldn't access the properties of product there. If we could access them, we could get rid of the `Product.createCriteria(...)` and the `BrowseInventory.createCriteria(..)` would handle all the logic from the product's query.


jmiranda:
Tech Huddle: Ignore pagination in the short term. And then let's completely redesign this feature maybe in React.

kchelstowski:
@awalkowiak @jmiranda as you suggested, I removed the pagination. I must say it's super slow guys - it took me **84 seconds** for this list to render

jmiranda (approved):
Now that is a beautiful PR.

jmiranda:
> @awalkowiak @jmiranda as you suggested, I removed the pagination. I must say it's super slow guys - it took me **84 seconds** for this list to render

I missed this comment earlier. So yeah if that's the case we can't merge this one without some more work being done.

jmiranda:
@kchelstowski @awalkowiak @drodzewicz Can you review this when you get a chance?

jmiranda:
@awalkowiak Also this PR was originally targeted to the release, but I think we can send it to develop since it's probably not a showstopper bug. Thoughts?

awalkowiak (approved):
This LGTM, but please @kchelstowski and @drodzewicz review it too.

src/groovy/org/pih/warehouse/product/ProductTypeEditor.groovy
```diff
@@ -0,0 +1,19 @@
...
+ * By using this software in any fashion, you are agreeing to be bound by
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.product
+
+
+import java.beans.PropertyEditorSupport
+
+class ProductTypeEditor extends PropertyEditorSupport {
```
awalkowiak, line 15: @jmiranda Why it is required?

jmiranda: Because I added the product type filter to the inventory browser search filters.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -776,80 +782,84 @@ class ProductAvailabilityService {
...
-                        }
-                    }
-                }
-            }
-        }.unique()
+        def innerProductIds = searchTerms ?
+                productService.searchProducts(searchTerms.toArray(), [])?.collect { it.id } : []
+
+        // Retrieve all product types with SEARCHABLE and SEARCHABLE_NO_STOCK activity codes
+        String productTypeQuery = "select pt from ProductType pt left join pt.supportedActivities sa where sa=:productActivityCode"
```
kchelstowski, line 789: nice job, this is what I needed and didn't know how to get it to be able to use it in the `createCriteria`

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -776,80 +782,84 @@ class ProductAvailabilityService {
...
-            items << [
+            // This is pretty inefficient if the previous query does not narrow the results
+            // if the inner products list is empty, but there are search terms then return empty results
+            if (innerProductIds || searchTerms) {
+                'in'("id", innerProductIds ?: [null])
+            }
+            order("productCode")
+        }
+
+        def quantityMap = products ? getQuantityOnHandByProduct(command.location, products) : [:]
```
kchelstowski, line 853: if I understand the query correctly - I assume that in the `createCriteria` we already count the `quantityOnHandByProduct` - so do we have to do it also here at this point?

drodzewicz (approved):

PR #3783, opened 2023-01-18
OBPIH-5268 Completely remove itext and core-renderer from build process.

mdpearson:
This PR makes sure there is one and only one implementation of `com.lowagie.*` in the class path. No guarantees it'll fix the recurring Cp1252 warnings, but it _will_ make it easier to follow the stack traces. At present it's not clear whether the issue is coming from `itext` or `openpdf`. (I suspect the former)


mdpearson:
To be perfectly honest I think this bogeyman may persist until the Grails 3 migration, so I don't want to spend too much time on it after this - let me know what you think

PR #3784, opened 2023-01-18
OBS-1375 Do not crash ProductAvailabilityService threads if deadlock occurs

mdpearson:
I figured it was easier to talk about this head-scratcher over a PR.

Current `saveProductAvailability` behavior is to log and then re-raise the exception, which can cause the GPars worker running it to crash. Since the blast radius for this issue is fairly small, for debugging purposes I made `saveProductAvailability` return a boolean if it succeeded, then call log.info() if it works, or log.error() if it doesn't. What I really want to know is what the other GPars threads are doing, but they stop logging after the deadlock exception is raised.

Not sure what the "right" behavior here is. Is `saveProductAvailability()` idempotent? If so, we could retry it (perhaps after an incremental delay) since the deadlock conditions likely pass a few seconds after the error is raised. Or we could resolve the root of the issue, but I think that'll need cleaner logging.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -39,7 +39,6 @@ class ProductAvailabilityService {
     def persistenceInterceptor
     def locationService
     def inventoryService
-    def dataService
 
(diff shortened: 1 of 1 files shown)


jmiranda:
@jmiranda to review the deleted code, but everything looks ok.

jmiranda (commented):
Deleted code needs to be reviewed by @jmiranda

jmiranda (approved):

PR #3786, opened 2023-01-19
Merge 0.8.19 hotfix2 into 0.8.21

awalkowiak:
This is the work, that was done for the Dominica implementation as a 0.8.19-hotfix2. We should move it to the master (release), as to not run into these changes being lost when updating Dominica to the newest version in the future.


kchelstowski (approved):
LGTM. Speaking for the part that I was working on, everything seems to be covered as it should

PR #3787, opened 2023-01-19
OBPIH-5329 Improve performance after deleting location

awalkowiak:
plus fixed deleting bin location with zone
plus fixed finding location type by name during import

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
@@ -241,7 +241,15 @@ class LocationApiController extends BaseDomainApiController {
         }
         def parentLocation = existingLocation.parentLocation
         if (parentLocation) {
+            existingLocation.parentLocation = null
             parentLocation.removeFromLocations(existingLocation)
+            parentLocation.save(flush: true)
+        }
+        def zone = existingLocation.zone
+        if (zone) {
+            existingLocation.zone = null
+            zone.removeFromLocations(existingLocation)
+            zone.save(flush: true)
         }
         existingLocation.delete(flush: true)
 
--- a/grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
@@ -144,12 +144,8 @@ class LocationDataService {
         }
         def locationType = null
         if (params.locationType) {
-            locationType = LocationType.createCriteria().get {
-                or {
-                    eq('name', params.locationType)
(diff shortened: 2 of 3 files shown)


grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -241,7 +241,15 @@ class LocationApiController extends BaseDomainApiController {
...
         if (parentLocation) {
+            existingLocation.parentLocation = null
             parentLocation.removeFromLocations(existingLocation)
+            parentLocation.save(flush: true)
+        }
+        def zone = existingLocation.zone
+        if (zone) {
+            existingLocation.zone = null
+            zone.removeFromLocations(existingLocation)
+            zone.save(flush: true)
```
jmiranda, line 252: This shouldn't be necessary since there's no association between zone and internal locations so this is basically a no-op. I assume this is in response to an issue you had (something like "deleted object would be re-saved"). In the future, can you add any issues you encounter to the ticket a) so there's an audit trail of the issue and b) so we can discuss it.

grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -144,12 +144,8 @@ class LocationDataService {
...
         def locationType = null
         if (params.locationType) {
-            locationType = LocationType.createCriteria().get {
-                or {
-                    eq('name', params.locationType)
-                    like('name', params.locationType + "%")
-                }
-            }
+            // Find by name with like, but value more the exact match
+            locationType = LocationType.findByNameOrNameIlike(params.locationType, params.locationType + "%")
```
jmiranda, line 148: Could you explain why this was necessary?

src/groovy/org/pih/warehouse/inventory/RefreshProductAvailabilityEvent.groovy
```diff
@@ -34,7 +34,8 @@ class RefreshProductAvailabilityEvent extends ApplicationEvent {
         this.locationId = source?.isInternalLocation() ? source?.parentLocation?.id : source?.id
         this.productIds = []
         this.forceRefresh = forceRefresh
-        this.disableRefresh = disableRefresh
+        // Avoid refreshing PA for all location, when there is no locationId
+        this.disableRefresh = this.locationId ? disableRefresh : Boolean.TRUE
```
jmiranda, line 38: This should probably be the only commit for this ticket. The other issues should have separate bug tickets that include the stacktrace and/or steps to reproduce.

jmiranda (commented):
Sorry for not getting to this sooner.

PR #3792, opened 2023-01-23
OBPIH-5273 Dashboard is underlined in menu on some pages (fixes after qa)

alannadolny:


src/js/components/stock-movement-wizard/StockMovementVerifyRequest.jsx
```diff
@@ -197,6 +197,9 @@ class StockMovementVerifyRequest extends Component {
             },
           };
 
+          const { pathname } = this.props.history.location;
+          this.props.history.replace({ pathname, search: `direction=${resp?.origin?.id === this.props.currentLocation?.id ? 'OUTBOUND' : 'INBOUND'}` });
```
awalkowiak, line 201: I do not get why this needs to be done. I think `create request` is always inbound and `verify request` is always outbound or if there is a way to enter verify request as `inbound` then why not add it as `injectedDirectly`?

src/js/utils/menu-utils.jsx
```diff
@@ -5,6 +5,7 @@ export const checkActiveSection = ({
   menuUrls,
   path,
   params,
+  menuSectionsUrlParts,
```
awalkowiak, line 8: If you are passing this here, then this entire util should be reworked to be based on this config and to get it working in the same way as on the grails side

grails-app/views/order/addComment.gsp
```diff
@@ -78,5 +78,13 @@
 			</div>
 		</div>
 	</div>
+<script>
+	$(document).ready(function() {
+		const menuConfigValues = $(".menu-config-value").toArray();
+		const stockMovementDirection = ${orderInstance.orderType.isPutawayOrder()} ? 'inbound' : 'purchasing';
+		const foundSection = menuConfigValues.find(it => stockMovementDirection === it.name)
+		applyActiveSection(foundSection)
```
awalkowiak, line 86: Two of the above lines could be moved to the `applyActiveSection` and only the `stockMovementDirection` could be passed instead of `foundSection`. That would reduce the amount of lines copy pasted.

awalkowiak (changes requested):

awalkowiak (approved):

PR #3793, opened 2023-01-24
OBPIH-5333 Fix wrapping zone name for stock transfer/replenishment

kchelstowski:


grails-app/views/inventoryItem/_showStockHistory.gsp
```diff
@@ -456,10 +456,12 @@
                                     <g:if test="${stockHistoryEntry?.isInternal}">
                                         <span>&nbsp;&rsaquo;&nbsp;</span>
                                         <g:if test="${stockHistoryEntry?.destinationBinLocation}">
-                                            <span>
+                                            <span class="line-base" title="${stockHistoryEntry?.destinationBinLocation?.zone}">
```
awalkowiak, line 459: Should the title have `?.name`? or `zone` is expected here?

kchelstowski: Yeah, by mistake I copied it from the `if` below that doesn't have the `.name`. I added it, but anyway this tooltip was working fine without `.name`, so I assume if you don't provide any property of the class and you just call the class (`zone`), it takes `name` by default.

awalkowiak: It worked because of ```String toString() { return this.name }``` which is in the `Location`

awalkowiak (approved):

PR #3795, opened 2023-01-24
OBPIH-5329 Improve deleting location with assigned zone

awalkowiak:


grails-app/services/org/pih/warehouse/data/LocationDataService.groovy
```diff
@@ -144,8 +144,7 @@ class LocationDataService {
         }
         def locationType = null
         if (params.locationType) {
-            // Find by name with like, but value more the exact match
-            locationType = LocationType.findByNameOrNameIlike(params.locationType, params.locationType + "%")
+            locationType = LocationType.findByIdOrName(params.locationType, params.locationType)
```
kchelstowski, line 147: if we want to search by name also, don't we want to keep `"%"`? I see it being removed and can't guess the reason?

kchelstowski: nvm, it became `findByIdOrName` and was `NameIlike`

kchelstowski (approved):

PR #3801, opened 2023-01-26
OBPIH-5342 Fix HelpScout not opening when changing locales

drodzewicz:


drodzewicz:
After investigating this issue it appeared to be that the fault of it was that there were two instances of  `<HelpScout /> ` being rendered at the same time.
During one of the previous releases when I was working on the collapsible mobile menu I went with the approach of rendering all of the navItems in the megamenu twice  _(one for default megamenu and another for mobile collapsible)_ and hide/show them using bootstrap breakpoint classes like `d-md-none`.
Since these navbarItems were hidden using classes that meant that they are still available in the DOM hence why there were two insatnces of the `<HelpScout />` button.
With my current solution all I do is instead of hiding these navbarItems using classes I use javascript instead which removes these elements from the DOM.

drodzewicz:
The `NavbarIcons.jsx` file looks like it has a lot of changes starting on [this line](https://github.com/openboxes/openboxes/pull/3801/files#diff-e8cae2f35da21cb9f1e6aac507f988860e66607858d7651c738dbc060439edb0R139) but actually I just wrapped these elements in braces with `isMenuCollapsed &&` to render the appropriate navitems

awalkowiak (approved):

PR #3803, opened 2023-01-27
OBPIH-5217 Add missing translations for some OrderTypes

drodzewicz:


grails-app/i18n/messages.properties
```diff
@@ -1602,6 +1602,9 @@ order.ordersInto.label=Orders into {0}
 order.ordersPlacedBy.label=Orders placed <b>by</b>
 order.ordersPlacedWith.label=Orders placed <b>with</b>
 order.orderTypeCode.label=Order Type
+order.orderType.PUTAWAY_ORDER=Putaway Order
+order.orderType.RETURN_ORDER=Return Order
+order.orderType.TRANSFER_ORDER=Transfer Order
```
kchelstowski, line 1607: we already specify some of orderTypes, including `TRANSFER_ORDER`
```
enum.OrderTypeCode.TRANSFER_ORDER=Transfer order
```
I don't mind your solution, as you probably didn't know about it, but shouldn't we follow that convention?

awalkowiak: Yes, we should use the enum here

awalkowiak: Actually, PUTAWAY_ORDER and RETURN_ORDER are not parts of the OrderTypeCode enum, so i think this could be added as a separate option

awalkowiak: @drodzewicz lets leave it as it is, but add remaining values of `orderType.code` plus rename begining of it as:
`order.orderType.code.<each of the codes>`

awalkowiak (changes requested):

PR #3804, opened 2023-01-27
OBPIH-5305 No validation on empty qty to transfer in bin replenishments 

alannadolny:


src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -145,17 +145,22 @@ const FIELD = {
...
-    if (item.quantity) {
-      if (item.quantity < 1) {
-        errors.requirements[key] = { quantity: 'react.replenishment.error.quantity.label' };
-      }
-      if (item.quantity > item.quantityAvailable) {
-        errors.requirements[key] = { quantity: 'react.replenishment.error.quantity.greaterThanQATP.label' };
+    // error shouldn't be added to fields which are not checked
+    if (item.checked) {
+      if (item.quantity) {
+        if (item.quantity < 1) {
```
awalkowiak, line 152: I would do this as:
```
if (item.quantity && item.quantity < 1) {...}
else if (item.quantity && item.quantity > item.quantityAvailable) {...}
else if (!item.quantity && item.checked) {...}
```
or 
```
if (item.quantity) {
   if (item.quantity < 1) { ... }
   if (item.quantity > item.quantityAvailable) { ... }
} else {
   if (item.checked) { ... }
}
```
I think that would look a bit better than
```
if () {
    if () {
        if () {   
...
```

kchelstowski: if I can propose:
We don't have to do nested `ifs` for 
```
if (item.quantity) {
  if (item.quantity < 1) {
    ...
  }
}
```
we could simplify it to `if (item?.quantity < 1)`

kchelstowski (approved):

src/js/components/replenishment/CreateReplenishment.jsx
```diff
@@ -145,17 +145,24 @@ const FIELD = {
...
   const errors = {};
   errors.requirements = [];
-
   _.forEach(values.requirements, (item, key) => {
+    // If item not checked, we don't want to validate
+    if (!item.checked) {
+      return;
+    }
+    // If item checked, validate
     if (item.quantity) {
```
drodzewicz, line 154: I think we can go even one step further and do something like:
```
if (!item.checked) {
  return;
}
if (!item.quantity) {
  errors.requirements[key] = { quantity: 'react.replenishment.error.emptyQtyToTransfer.label' };
  return;
}
if (item.quantity < 1) {
  errors.requirements[key] = { quantity: 'react.replenishment.error.quantity.label' };
}
if (item.quantity > item.quantityAvailable) {
  errors.requirements[key] = { quantity: 'react.replenishment.error.quantity.greaterThanQATP.label' };
}
```

We could also short circuit _(add **return** inside of each **if** to not go further because error is already found)_ but it is optional in this case

drodzewicz (changes requested):

awalkowiak (approved):

PR #3805, opened 2023-01-27
OBPIH-5313 Improve synonyms feature for products translation

kchelstowski:


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -992,10 +992,15 @@ class ProductController {
      */
     def addSynonymToProduct = {
         println "addSynonymToProduct() " + params
-        def product = Product.get(params.id)
-        if (product) {
-            product.addToSynonyms(new Synonym(name: params.synonym, locale: RCU.getLocale(request)))
-            product.save(flush: true, failOnError: true)
+        Product product = null
+        // Getting locale from RCU, because unable to inject localizationService to productService (session scope)
+        Locale locale = RCU.getLocale(request)
```
jmiranda, line 997: The user should be choosing which locale, so the locale will be passed in as a parameter.

grails-app/domain/org/pih/warehouse/core/Synonym.groovy
```diff
@@ -11,6 +11,7 @@ package org.pih.warehouse.core
 
 import org.pih.warehouse.auth.AuthService
 import org.pih.warehouse.product.Product
+import org.pih.warehouse.product.SynonymClassification
```
jmiranda, line 14: Synonym is in the core package (i.e. we could add a synonyms association to other objects as well). So this should also be in the core package. To be more like other enums it should probably be SynonymTypeCode. We may also want to create a SynonymType but that might be overkill at the moment.

grails-app/domain/org/pih/warehouse/core/Synonym.groovy
```diff
@@ -50,6 +52,7 @@ class Synonym implements Serializable {
     static constraints = {
         name(nullable: false, maxSize: 255)
         locale(nullable: true)
+        classification(nullable: false)
```
jmiranda, line 55: Should be synonymTypeCode or synonymType if we end up creating that class.

jmiranda (changes requested):
Looks good just a few changes. 

@awalkowiak We always end up adding the *Type class to wrap the enum *TypeCode. Do you think we should do that here? It would save us the trouble from having to refactor in the future but I'm ok leaving it out for now.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
kchelstowski: damn, I must've removed it somehow - thanks for vigilance

kchelstowski: it was obviously meant to be, if the user doesn't provide the locale, it will take it from the current locale, so it was meant to be:
```
Locale locale = params.locale ? new Locale(params.locale) : RCU.getLocale(request)
```

jmiranda: Awesome. Now that we ask the user for a locale, if the user doesn't provide one let's throw a validation error. My instinct to use the current locale was a bad one.

kchelstowski:
@jmiranda @awalkowiak I changed the `classification` to `synonymTypeCode` as you suggested - if it might have an impact in complexity of any refactor in the future, I would like to reduce that complexity if I am able to :) 
I didn't go for creating another class `SynonymType` as you said, it might be an overkill at this moment.
I also added a validation if a user doesn't provide a locale, so I do not look for user's current locale anymore.

grails-app/migrations/0.8.x/changelog-2023-01-27-1015-add-synonym-classification-column.xml
```diff
@@ -0,0 +1,16 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="kchelstowski" id="270120231015-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="synonym" columnName="synonym_type_code" />
+            </not>
+        </preConditions>
+        <addColumn tableName="synonym">
+            <column name="synonym_type_code" type="VARCHAR(100)" defaultValue="DISPLAY_NAME"/>
```
drodzewicz, line 13: shouldn't defaultValue be `ALTERNATE_NAME` instead ? _(as specified in the ticket)_

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1371,4 +1372,20 @@ class ProductService {
...
         }
     }
+
+    Product addSynonymToProduct(String productId, String synonymTypeCodeName, String synonymValue, String localeName) {
+        Product product = Product.get(productId)
+        if (!localeName) {
+            throw new IllegalArgumentException("You must provide a locale")
+        }
+        Locale locale = new Locale(localeName)
+        SynonymTypeCode synonymTypeCode = synonymTypeCodeName ? SynonymTypeCode.valueOf(synonymTypeCodeName) : SynonymTypeCode.DISPLAY_NAME
```
drodzewicz, line 1382: probably same here. Default value should probably be `ALTERNATE_NAME`

grails-app/views/product/_productSynonyms.gsp
```diff
@@ -45,11 +60,35 @@
 
                                       url="[controller: 'product', action:'addSynonymToProduct']">
                             <input name="id" type="hidden" value="${productInstance?.id}" />
-                            <input id="synonym" type="text" name="synonym" value="" size="80" class="medium text"/>
-                            <button  class="button">
-                                <img src="${createLinkTo(dir:'images/icons/silk', file:'add.png')}" />&nbsp;
-                                ${warehouse.message(code:'default.button.add.label')}
-                            </button>
+                            <div class="d-flex w-75 gap-8">
```
drodzewicz, line 63: It would be a good idea to already add translations for the input placeholders 
_Select a locale_
_Select a classification_
_Type the synonym here_

drodzewicz (changes requested):

grails-app/migrations/0.8.x/changelog-2023-01-27-1015-add-synonym-classification-column.xml
kchelstowski: good catch - it was done first time round, before I consulted it with Manon and I've forgotten to change it then, thanks for viligance

grails-app/views/product/_productSynonyms.gsp
kchelstowski: yeah, also forgot to add it immediately, then you obviously forget to go back to such minor things :rofl:  thanks

awalkowiak (approved):
LGTM.

@jmiranda I don't see any real value added with adding the `type code class`. In the future, we could add `type classification` within the enum itself, if we're gonna need it (like we are doing with `sortOrder`, `displayStatusCode` or `variant`)

PR #3809, opened 2023-01-27
OBS-1385 Update c3p0; resolve gpars dependency conflict; add default Quartz config

mdpearson:
This PR resolves a few issues I found while testing OBS-1360 Ansible deployments.

config-wise, I reduced the number of Quartz threads and "niced" them to prevent thrashing the CPU when a bunch of jobs kick off.

build-wise, the following nits are resolved:

- `gant` now uses the same `GPars` release across all build phases,
- We had duplicate symbols between a 15-year-old release and a 5-year-old one. `c3p0` was renamed in 2007 (`c3p0:c3p0` -> `com.mchange:c3p0`) and one of our old dependencies was pulling in the old version. At least once I saw Grails come up using the 2007 version, although I think the runtime dependency resolution was more or less arbitrary.

(See also #3810.)

### DIFF
--- a/grails-app/conf/BuildConfig.groovy
+++ b/grails-app/conf/BuildConfig.groovy
@@ -67,10 +67,10 @@ grails.project.dependency.resolution = {
         runtime "org.slf4j:jul-to-slf4j:1.7.33"
 
         // Required by database connection
-        compile 'mysql:mysql-connector-java:5.1.47'
+        compile 'mysql:mysql-connector-java:5.1.49'
 
         // Required by database connection pool
-        compile 'com.mchange:c3p0:0.9.5.3'
+        compile 'com.mchange:c3p0:0.9.5.5'
 
         // Required by docx4j functionality
         compile('org.docx4j:docx4j:2.8.1') {
@@ -131,10 +131,15 @@ grails.project.dependency.resolution = {
          */
         test("com.icegreen:greenmail:1.5.10") { excludes "junit" }
 
-        // Required for GPars
+        build 'org.codehaus.gpars:gpars:0.12'  // otherwise early build chain uses 0.9
         compile "org.codehaus.gpars:gpars:0.12"
+        // Required for GPars
         compile "org.codehaus.jsr166-mirror:jsr166y:1.7.0"
         compile "org.codehaus.jsr166-mirror:extra166y:1.7.0"
+
+        compile('org.quartz-scheduler:quartz:2.1.6') {
+            exclude 'c3p0'  // otherwise Quartz pulls in an ancient release from 2007
+        }
 
         // Unknown
         build('org.jboss.tattletale:tattletale-ant:1.2.0.Beta2') { excludes "ant", "javassist" }


grails-app/conf/Config.groovy
```diff
@@ -132,6 +132,12 @@ grails.validateable.packages = [
         'org.pih.warehouse.shipment',
 ]
 
+org.quartz.scheduler.instanceName = 'DefaultQuartzPool'
+org.quartz.scheduler.instanceId = 'Config.groovy'
```
jmiranda, line 136: need clarification on instanceId

grails-app/conf/Config.groovy
```diff
@@ -132,6 +132,12 @@ grails.validateable.packages = [
...
+org.quartz.scheduler.skipUpdateCheck = true
+org.quartz.threadPool.threadCount = 2
+org.quartz.threadPool.threadPriority = Thread.MIN_PRIORITY
```
jmiranda, line 139: Do we actually know how this affects the quartz threads?

grails-app/conf/BuildConfig.groovy
```diff
@@ -125,11 +125,16 @@ grails.project.dependency.resolution = {
          */
         test("com.icegreen:greenmail:1.5.10") { excludes "junit" }
 
-        // Required for GPars
+        build 'org.codehaus.gpars:gpars:0.12'  // otherwise early build chain uses 0.9
```
jmiranda, line 128: Am I misunderstanding the scopes? Does scope work upwards or downwards? i.e. compile -> runtime or vice versa?

jmiranda: In other words, do we need boht build and compile? Or just build?

mdpearson: My understanding is that everything in the `compile` class path is also included in the `runtime` class path, but the `build` class path is disjoint from both.

In other words:
- I need the dependency at runtime only -> `runtime`
- I need the dependency at compile time and runtime -> `compile`
- I only need the dependency at compile time -> `implementation` (Gradle 4+)
- I need the dependency at compile time and runtime and Grails also silently uses it -> `build` _and_ `compile`.

If you run `grails dependencyReport` before and after this PR you can see that the build step will use GPars 0.9 while the others use 0.12. That seemed like a recipe for trouble, but I can't point to a specific problem it was causing.

grails-app/conf/Config.groovy
mdpearson: - `instanceId` can be any string you like. I set it to the name of the config file to make it easier to understand from log output where the config was specified, but I can happily strike it.
- `skipUpdateCheck` means Quartz shouldn't phone home and tell you you're running an old version. We already know that, and they recommend disabling it for production instances. I believe our plugin does this for us, so it could be removed if you like
- `threadCount` affects how many threads Quartz will allocate in its internal pool. The default is 10 which is a lot for a 1- or 2-core machine.
- `threadPriority` affects how "nice" (in the POSIX sense) the Quartz threads are. I set them to lowest priority since as I understand it we use Quartz exclusively for background tasks that are not time-sensitive.

You can see Quartz picking up these settings at application startup:

```
2023-02-01 16:42:33,893 [localhost-startStop-1] INFO  core.QuartzScheduler  - Scheduler meta-data: Quartz Scheduler (v2.1.6) 'LegacyQuartzPool' with instanceId '/home/tomcat/.grails/openboxes-config.groovy'
  Scheduler class: 'org.quartz.core.QuartzScheduler' - running locally.
  NOT STARTED.
  Currently in standby mode.
  Number of jobs executed: 0
  Using thread pool 'org.quartz.simpl.SimpleThreadPool' - with 4 threads.
  Using job-store 'org.quartz.simpl.RAMJobStore' - which does not support persistence. and is not clustered.
```

grails-app/conf/Config.groovy
mdpearson: thanks for pointing this out -- see big comment below

grails-app/conf/Config.groovy
jmiranda: I would love to see these as comments in the configuration if possible.

jmiranda (changes requested):
Add comments to config

jmiranda (changes requested):
@mdpearson I'm having second thoughts about the default settings here. I think we should set the values according to the default specified by Quartz and then override them in our configuration (i.e. ansible, runtime config). The Config.groovy / application.yml could/should include a discussion of what the default is and any caveats based on our knowledge of those settings. But we should not be too opinionated because we don't know how the software will be deployed and don't want to be dealing with issues related to our own biases on these matters.

mdpearson:
> @mdpearson I'm having second thoughts about the default settings here. I think we should set the values according to the default specified by Quartz and then override them in our configuration (i.e. ansible, runtime config). The Config.groovy / application.yml could/should include a discussion of what the default is and any caveats based on our knowledge of those settings. But we should not be too opinionated because we don't know how the software will be deployed and don't want to be dealing with issues related to our own biases on these matters.

Yeah, that's a perfectly reasonable approach. I removed the config changes from this PR.

jmiranda:
> Yeah, that's a perfectly reasonable approach. I removed the config changes from this PR.

Should we add a ticket to include them in the Ansible code? Or at least in documentation / knowledge base? Or do you think this should be left as an implementation detail that is left to the implementing org? 

I suppose, moving the configuration and discussion to the ticket (https://pihemr.atlassian.net/browse/OBS-1385) would be the path of least resistance.

jmiranda (approved):
LGTM

PR #3810, opened 2023-01-27
OBS-1385 Improve Quartz configuration and dependency management under Grails 3

mdpearson:
This PR resolves a few issues I found while testing OBS-1360 Ansible deployments.

config-wise, I reduced the number of Quartz threads and "niced" them to prevent thrashing the CPU when a bunch of jobs kick off.

build-wise, I cleaned up the way we pull in Quartz dependencies and added a clarifying comment to our future selves should we see connection-pooling problems in Grails 3. (See also #3809.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -295,7 +295,6 @@ configurations {
             force 'org.checkerframework:checker-qual:3.12.0'  // use consistent version
             force 'org.dom4j:dom4j:2.1.3'  // security patch
             force 'org.objenesis:objenesis:2.1'  // openjdk support
-            force 'org.quartz-scheduler:quartz:2.3.2'  // security patch
         }
     }
 }
@@ -627,10 +626,9 @@ dependencies {
     // use a release optimized for Grails 3.3.x for this `excel-import` sub-dependency
     implementation 'org.grails.plugins:joda-time:2.1.0'
 
-    // enable `import org.quartz.*`
+    // enable `import grails.plugins.quartz.Job*`
     compile('org.grails.plugins:quartz:2.0.13') {
-        exclude group: 'com.mchange'
-        exclude group: 'com.zaxxer'
+        transitive = false  // we pull in quartz-scheduler directly, below
     }
 
     // list all scheduled jobs at e.g. http://localhost:8080/openboxes/quartz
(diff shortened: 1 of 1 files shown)


jmiranda:
@mdpearson will recreate a new pull request off the newly rebased OBGM branch.

grails-app/conf/application.yml
```diff
@@ -60,6 +60,14 @@ quartz:
         showTriggerNames: true
         showCountdown: true
         showTickingCloud: true
+    scheduler:
+        instanceId: application.yml
```
jmiranda, line 64: Let's add a comment here to explain why we're using this value vs others.

jmiranda:
Tech Huddle:
* RefreshProductAvailabilityJob and RefreshOrderSummaryJob are more important than any other activity in the system so those need high priority equal to database transactions. 
* @mdpearson to write up some considerations of other job scheduling technologies for post-migration work so we can horizontally scale.

grails-app/conf/application.yml
```diff
@@ -60,6 +60,14 @@ quartz:
...
+        instanceName: DefaultQuartzPool
+        skipUpdateCheck: true
+    threadPool:
+        threadCount: 2
```
jmiranda, line 68: We want to just make sure that this thread count is high enough for a default system. When we start up the system, there are a number of scheduled tasks that startup immediately. At the same time, these job instances check whether the database migrations are being run. I want to make sure that there are enough threads in each of the different pools to not cause weird blocking during startup.

grails-app/conf/application.yml
```diff
@@ -60,6 +60,14 @@ quartz:
...
+        # java.lang.Thread.MIN_PRIORITY == 1
+        threadPriority: 1
```
jmiranda, line 70: Should set this according to the different jobs. Refresh product availability and order summary jobs should at least have regular priority, if not elevated.

jmiranda (changes requested):
Need to recreate onto the newly rebased branch and add some comments where necessary.

jmiranda (changes requested):
Same here
https://github.com/openboxes/openboxes/pull/3809#pullrequestreview-1293530922

jmiranda:
@mdpearson I just changed the target branch of this PR. Could you double check that everything that you wanted to include is here?

@awalkowiak Can you give this another review when you get a chance?

jmiranda (approved):
LGTM

PR #3814, opened 2023-01-30
OBPIH-5233 Fix decimal format of a given locale not working for unit price on PO

drodzewicz:
I was hopping that by giving an input tag `<input type="number" />` an attribute `lang="fr"` it would format it's numbers according to the given locale but unfortunately it doesn't work like that, which is why I had to change the input type to "text".
What I mean by format is for example a number **202021.13**
in **English** is equivalent to **202021.13** or **202,021.13**
in **French** is equivalent to **202021,13**
in **Spanish** is equivalent to **202021,13** or **202.021,13**
[more formats](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html)

Then on the backend side I would need to parse this string of numbers to a normal decimal format based on selected locale because otherwise when casting it to BigDecimal _(unitPrice is of type BigDecimal)_ it would ignore the decimal values.

### DIFF
--- a/grails-app/views/order/_orderItemForm.gsp
+++ b/grails-app/views/order/_orderItemForm.gsp
@@ -45,7 +45,7 @@
         <input type="number" id="quantityPerUom" name="quantityPerUom" class="text required" placeholder="${g.message(code: 'default.qtyPerUom.label', default: 'Qty per UoM')}" style="width: 100px"/>
     </td>
     <td class="center middle">
-        <input type="text" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
+        <g:decimalNumberField id="unitPrice" name="unitPrice" required="true" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px" />
     </td>
     <td class="center middle">
     </td>
--- a/grails-app/views/order/_orderItemForm.gsp
+++ b/grails-app/views/order/_orderItemForm.gsp
@@ -45,7 +45,7 @@
         <input type="number" id="quantityPerUom" name="quantityPerUom" class="text required" placeholder="${g.message(code: 'default.qtyPerUom.label', default: 'Qty per UoM')}" style="width: 100px"/>
     </td>
     <td class="center middle">
-        <input type="number" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
+        <input type="text" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
     </td>
     <td class="center middle">
     </td>
--- a/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
@@ -27,11 +27,9 @@ import org.pih.warehouse.product.Product
 import org.pih.warehouse.product.ProductSupplier
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentItem
-import org.pih.warehouse.util.LocalizationUtil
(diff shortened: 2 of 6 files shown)


grails-app/views/order/_orderItemForm.gsp
```diff
@@ -45,7 +45,7 @@
         <input type="number" id="quantityPerUom" name="quantityPerUom" class="text required" placeholder="${g.message(code: 'default.qtyPerUom.label', default: 'Qty per UoM')}" style="width: 100px"/>
     </td>
     <td class="center middle">
-        <input type="number" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
+        <input type="text" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
```
awalkowiak, line 48: What happens now if in fr locale someone will use dot `.` as a decimal point separator? Plus what happens if someone enter text here instead of number?

drodzewicz: If a user is in a french locale and types in a number with a dot decimal separator like **2.22** it will parse it as **2,00**, so if you switch the language you have to use the appropriate number formatting based on the rules of the given locale.

As for typing in text, it results with a parse exception that is displayed in the error notification like
`{"errorCode":500,"cause":"java.text.ParseException","errorMessage":"Unparseable number: \"dwdw\""}`
so this should probably be fixed.

Do you maybe have a suggestion on what could be done different from the current version ?

jmiranda:
Tech Huddle
@jmiranda to research best practices around number format in non-US locales. Need to check to see what the formatName attribute does (https://docs.grails.org/3.1.1/ref/Tags/formatNumber.html).

```
default.number.label = 
default.currency.label = 
default.percent.label = 
```

> formatName (optional) - Look up format from the default MessageSource / ResourceBundle (i18n/.properties file) with this key. Look up format from the default MessageSource / ResourceBundle (i18n/.properties file) with this key. If format and ? formatName are empty, format is looked up with 'default.number.format' key. If the key is missing, '0' formatting pattern is used.
> type (optional) - The type of formatter to use: 'number', 'currency' or 'percent' . format or formatName aren't used when type is specified.

jmiranda (changes requested):
So I think we can make this work without doing anything drastic. 

_orderItemForm.gsp
```
<g:textField type="number" id="unitPrice" required name="unitPrice" size="2" class="text required" placeholder="${g.message(code: 'default.pricePerUom.label', default: 'Price per UoM')}" style="width: 100px"/>
```
image

image




_orderItemFormDialog.gsp
```
<g:textField type="number" id="dlgUnitPrice" name="unitPrice" value="${fieldValue(bean: orderItem, field: 'unitPrice')}" size="10" class="text" />
```

image

jmiranda:
The only case that doesn't seem to work is using a mix of grouping / decimal characters. So if you enter a value like 1.920.280,23 then I believe the data binding mechanism would convert that to 1.92. I'm guessing if it was really necessary, we could add a custom editor to handle this special case. But I'd like to find out if users would actually do something like this first.

drodzewicz:
I think that initially my approach to this problem was wrong. At first I thought that to solve this issue I needed to make sure that the backend always received a decimal number in the same format like **2.32** no matter which locale is chosen. 
But I guess grails is smarter than that and when we have selected a locale it expects a decimal number in a format of a that selected locale.
One case that confused me was when I set a number to **1.999,2** in french locale and it converted it to **1.00** but that was probably because there is no **dot** separator in french. I expected it to escape this character but instead it cut the number at the dot. 

After analyzing all the comments and discussions around this topic I cam up with the following solution.
My solution is to have a controlled text input that only lets users to type or paste in numbers in a format of the chosen locale.
That means that 
given I am in an english locale, it would only let me input decimal numbers with **dot** decimal and prevent me from inputting a **comma** 
and in case of a french locale it lets me input **comma** and prevents me from inputting a **dot**.

Contrary to the previous method of how I extracted the decimal separator from `.ToLocaleStrint(locale)` using regex, this time I am not using any "hacky way" to do it, I simply access this from the built in `java.text.DecimalFormat` that we use across the whole project to format numbers

cc @awalkowiak  @jmiranda

jmiranda (approved):

PR #3817, opened 2023-01-30
OBS-1253 Don't run Quartz jobs during DB migrations; consistently disable concurrent job execution

mdpearson:
This one is from the icebox, but since I was dinking around with Quartz config last week I figured I'd bring it back to life for a tech huddle.

- some Quartz jobs have a `@DisallowConcurrentExecution` annotation;
- some declare `def concurrent = false`;
- some check whether a migration is happening before running;
- each one of the above is not done by at least one job :-( .

This PR makes all Quartz jobs perform all 3 checks and consolidates common code in a new JobsUtil class. (I also added comments to ease Grails 1/3 rebases going forward.)

I'm not sure whether the 0.8.19 deploy was actually affected by this; the migrations there took longer than they have on any other push so the window for error was wider -- but I think these bugs should be fixed even if they haven't bitten us in the last week or two. Let me know what y'all think.

### DIFF
--- a/grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
+++ b/grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
@@ -1,28 +1,20 @@
 package org.pih.warehouse.jobs
 
-import org.codehaus.groovy.grails.commons.ConfigurationHolder as CH
 import org.quartz.DisallowConcurrentExecution
-import util.LiquibaseUtil
 
 @DisallowConcurrentExecution
 class AssignIdentifierJob {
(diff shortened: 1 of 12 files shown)


jmiranda (approved):
LGTM; @awalkowiak Take a look at these two PRs from @mdpearson and let me know if they will cause any issues with the OBGM rebase when this comes through after the next release. Can't remember how much we've touched Quartz on the OBGM side.

awalkowiak (approved):
@jmiranda this one looks good to me in the context of rebase (I think it even will be rebased automatically)

jmiranda (commented):
Approved

PR #3818, opened 2023-01-31
OBPIH-5315 Display product name translation on stock card header and in the global search

kchelstowski:
Added:

- separate getter for translated product name - I think it's a better approach instead of overriding the `name` getter, so that we have better control of that, especially that we don't cover the translations for all places at once
- if a synonym for user's current locale exists, we want to display it and have a tooltip with the original name of the product. If the synonym doesn't exist, we don't need to have the tooltip
- for now implemented only for stock card header and the global search, so that I do not block other tickets. The rest of cases for stock card (all the tabs and modals) will be covered in a separate, already existing ticket

What we might consider thinking about in the future (consulted with Manon) is that what to do if someone creates a product with e.g. Spanish name - so that the "original" name of the product is not in English.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/JsonController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/JsonController.groovy
@@ -1079,8 +1079,10 @@ class JsonController {
                     type : product.class,
                     url  : request.contextPath + "/" + type + "/redirect/" + product.id,
                     value: product.name,
-                    label: product.productCode + " " + product.name + " " + quantity,
-                    color: product.color
+                    label: product.productCode + " " + (product.translatedName ?: product.name) + " " + quantity,
+                    color: product.color,
+                    // Do not remove double negation even if IDE recommends this - we want to convert string/null to boolean
+                    hasTranslatedName: !!product.translatedName
             ]
         }
         render json.findAll { it != null } as JSON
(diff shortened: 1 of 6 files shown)


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1079,8 +1079,10 @@ class JsonController {
                     type : product.class,
                     url  : request.contextPath + "/" + type + "/redirect/" + product.id,
                     value: product.name,
-                    label: product.productCode + " " + product.name + " " + quantity,
-                    color: product.color
+                    label: product.productCode + " " + (product.translatedName ?: product.name) + " " + quantity,
+                    color: product.color,
+                    // Do not remove double negation even if IDE recommends this - we want to convert string/null to boolean
+                    hasTranslatedName: !!product.translatedName
```
awalkowiak, line 1085: Why you cant return the translated name and then check if it is there if you need it? This one looks like a redundant field.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
kchelstowski: because the one we are actually using to display the name is `label`, and it would be impossible to determine on the frontend side if inside this `label` we have `translatedName` or the original name

awalkowiak: ok, makes sense

awalkowiak (approved):

jmiranda (approved):

PR #3822, opened 2023-02-01
OBPIH-5220 Send more information to New Relic, as recommended by New Relic

mdpearson:


jmiranda:
@mdpearson Can you provide some more details (in the ticket) to help me understand what this does and why we need/want it? Since we're using NewRelic I'm all for anything that provides more details on the APM side of things. But other than the "transaction name" component I don't really have a basis for judging this change. In addition, do we need to consider making this configurable? For example, if someone is hosting the application on a LAN without internet will this have any negative impact?

mdpearson:
I'll check with New Relic themselves about this as part of https://pihemr.atlassian.net/browse/OBS-1454

mdpearson:
@jmiranda when we were looking at the dashboards today with New Relic -- this code was running on the host we were looking at. I don't think it necessarily needs to go out with the next release if there's a code freeze, but I did want to bring it to your attention.

jmiranda:
@mdpearson As mentioned on the tech debt call I thought we were missing tags from the React page, but it looks like i was mistaken. The grails-template.html file (under src/assets) is actually using the main.gsp template so we should be ok there.

mdpearson:
@jmiranda any other edits required for this PR?

jmiranda (approved):
@mdpearson Nope. LGTM.

jmiranda:
@awalkowiak If you feel comfortable with these changes can you merge them with the other hotfix issues.

PR #3823, opened 2023-02-02
OBPIH-4949 Disable next button when no items added to all workflows 

alannadolny:


src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -504,7 +504,9 @@ class AddItemsPage extends Component {
                     }
                   }}
                   className="btn btn-outline-primary btn-form float-right btn-xs"
-                  disabled={!_.some(values.returnItems, item => !_.isEmpty(item))}
+                  disabled={
+                  !_.some(values.returnItems, item => item.product && _.parseInt(item.quantity))
```
kchelstowski, line 508: It's rather a suggestion (or my own preference), not a change request, but imho more readable would be in this case:
```
_.some(values.returnItems, item => !item.product || !_.parseInt(item.quantity))
```

alannadolny: It doesn't work in all cases. For example, in create inbound workflow "next" button should be enabled when there is at least one line with entered product and quantity. In your suggestion the "next" button is disabled when there are lines with product and quantity and one additional empty line.

kchelstowski: ok, got it now. I understood it like you explained. Now it makes sense to me what you did

kchelstowski (approved):

awalkowiak (changes requested):
Please fix indentation, it looks off a bit

PR #3824, opened 2023-02-02
OBPIH-5326 add translations to product names via import

drodzewicz:


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -992,12 +1013,10 @@ class ProductController {
      */
     def addSynonymToProduct = {
         println "addSynonymToProduct() " + params
-        Product product = null
+        Product product = Product.read(params.id)
```
kchelstowski, line 1016: it can't stay like this - as discussed, it will cause `LazyInitializationExceptions` when trying to add a wrong synonym (when the validation should appear)

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -1106,6 +1125,49 @@ class ProductController {
...
+        if (request.method == "POST") {
+            File localFile = null
+            MultipartHttpServletRequest mpr = (MultipartHttpServletRequest) request
+            CommonsMultipartFile uploadFile = (CommonsMultipartFile) mpr.getFile("file")
+
+            if (uploadFile?.empty) {
+                flash.error = "Please upload a non-empty file"
+                redirect(controller: 'product', action: 'edit', id: params['product.id'])
+                return
+            } else {
```
kchelstowski, line 1143: `else` is not required when you make a `return` above

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -0,0 +1,46 @@
...
+    Boolean validateData(ImportDataCommand command) {
+        log.info "Validate data " + command.filename
+
+        command.data.eachWithIndex { params, index ->
+            Product product = Product.findByProductCode(params['product.productCode'])
+            if (!product) {
+                command.errors.reject("Row ${index + 1}: Product with code '${params['product.productCode']}' does not exist")
+            }
+            try {
+                SynonymTypeCode.valueOf(params['synonymTypeCode']);
```
kchelstowski, line 29: what is this line supposed to be doing? I do not see it being assigned anywhere

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
drodzewicz: it is in a **try** **catch** block, if `params['synonymTypeCode']` has a string/value that does not exists in the SynonymTypeCode enum it will throw an exception where in catch block I assign a validation error that such a SynonymTypeCode does not exist

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
drodzewicz: Thanks for pointing that out, this didn't look like a problem to me at first but I see it now

awalkowiak (approved):

PR #3827, opened 2023-02-06
OBPIH-5316 Add ability to search products by translated synonym of type DISPLAY_NAME

drodzewicz:


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1272,6 +1280,7 @@ class ProductService {
             where product.active = 1 and (${terms.collect {"""
                 lower(product.name) like '%${it}%' 
                 or lower(product.product_code) like '${it}%' 
+                or (synonym.synonym_type_code = 'DISPLAY_NAME' and synonym.name like '%${it}%')
```
kchelstowski, line 1283: Could we inject here the enum value instead of hard-coding the `DISPLAY_NAME`?

kchelstowski (changes requested):

drodzewicz:
it is a bit unclear to me why we have 
```
def searchProductDtos(String[] terms) { ... }
```
and a pretty similar method
```
List<Product> searchProducts(String[] terms, List<Category> categories) { ... }
```

I also assume that we are interested in synonyms of type **DISPLAY_NAME** when searching for a product

kchelstowski:
@drodzewicz I agree it's a bit unclear - from what I can see, the `def search` in the `ProductApiController` 
```
if(params.availableItems) {
            products = productService.searchProducts(terms, [])
            def location = Location.get(session.warehouse.id)
            def availableItems = productAvailabilityService.getAvailableBinLocations(location, products).groupBy { it.inventoryItem?.product?.productCode }
            products = []
            availableItems.each { k, v ->
                products += [
                    productCode: k,
                    name: v[0].inventoryItem.product.name,
                    id: v[0].inventoryItem.product.id,
                    product: v[0].inventoryItem.product,
                    quantityAvailable: v.sum { it.quantityAvailable },
                    minExpirationDate: v.findAll { it.inventoryItem.expirationDate != null }.collect {
                        it.inventoryItem?.expirationDate
                    }.min()?.format("MM/dd/yyyy"),
                    color: v[0].inventoryItem.product.color
                ]
            }

            products = products.unique()
        } else {
            products = productService.searchProductDtos(terms)
        }
```

We are using the `searchProductDtos` if the `params.availableItems` is falsy, but I do not know the reason.
We can see that if the `params.availableItems` is true, the DTO includes product's `quantityAvailable` etc - as for `ProductSearchDto` - it doesn't include the properties such as `quantityAvailable`.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
drodzewicz: yeah I wasn't really sure what would be the best choice since this query expects a string so either a 
`DISPLAY_NAME` or `${SynonymTypeCode.DISPLAY_NAME.name()}`
but looking at it again I think the `enum.*.name()` makes a bit more sense
so I'll update it

drodzewicz: actually I don;t even need to `*.name()` since we already have 
```
   String toString() {
        return name()
    }
```
so that's great 
thanks @kchelstowski

kchelstowski: since `toString()` is overrided I don't think you even have to add `.name()`

kchelstowski (approved):

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -327,6 +327,14 @@ class ProductService {
             }
             or {
                 if (params.name) ilike("name", "%" + params.name.replaceAll(" ", "%") + "%")
```
kchelstowski, line 329: shouldn't this be removed at this point?

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/product/ProductService.groovy
drodzewicz: probably not because this query first matches by the product name and if it does not find any then matches by synonym name

kchelstowski: yeah, but I feel like one of the `ifs` is redundant or maybe I misunderstand something - 
what about doing:
``` 
or { 
  if (params.name) {
    ilike("name", "%" + params.name.replaceAll(" ", "%") + "%")
    synonyms {
          and {
            ilike("name", "%" + params.name.replaceAll(" ", "%") + "%")
            eq("synonymTypeCode", SynonymTypeCode.DISPLAY_NAME)
          }
    }
  }
}

drodzewicz: Oh I see, I misunderstood what you meant by "removing"
I interpreted it as if you wanted me to remove the if and whats inside it  
like the `ilike("name", "%" + params.name.replaceAll(" ", "%") + "%")`

but I guess you meant that we can put all that inside one if instead of two.
Sounds good I can  do that.

awalkowiak (approved):

PR #3828, opened 2023-02-07
OBPIH-5317 Export files with product translations (second part)

alannadolny:


kchelstowski (commented):
Have you made sure that those do not impact the HTML reports or could you at least name those, that impact? The HTML reports are done in another ticket that has already its own PR, so it would be good to know what might impact on each other.

Other than that LGTM

alannadolny:
It has impact on: on order report summary, request detail report and transaction report

awalkowiak (approved):

PR #3834, opened 2023-02-09
OBPIH-5367 Add validation to prevent multiple display names per locale

kchelstowski:
I tried to make it as generic as I could, so that the methods I created, e.g. `getSynonymsOfTypeAndLocale` or `validateMultipleSynonymsOfType` are "ready" to be used in the future if we for example add some new synonym type codes that would have to be unique.

What is done:

- validation not to allow multiple synonyms of type `DISPLAY_NAME` per locale (included both - adding manually and through import)
- preventing from clearing the form values if an error occurs (so that the user can correct him/herself without having to re-write everything again)
- I also added a potential fix for another `NullPointerException` that Katarzyna has discovered today at inventory level dialog (probably the source is the same as yesterday - I found only one potential place in the dialog file which could cause it here)

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -1141,7 +1141,7 @@ class ProductController {
     def importProductSynonyms = { ImportDataCommand command ->
 
         log.info params
-        log.info command.location
+        log.info command?.location
 
         if (request.method == "POST") {
             File localFile = null
@@ -1167,12 +1167,13 @@ class ProductController {
             command.data = excelImporter.data
 
             command.errors = null
-            excelImporter.importData(command)
+            excelImporter.validateData(command)
 
             if (command.errors.allErrors) {
                 flash.errors = command.errors
             } else {
                 flash.message = "Succesfully imported product synonyms"
+                excelImporter.importData(command)
             }
         }
 
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -1017,13 +1017,19 @@ class ProductController {
         def inputValues = null
         try {
             product = productService.addSynonymToProduct(params.id, params.synonymTypeCode, params.synonym, params.locale)
-        } catch (IllegalArgumentException e) {
+        } catch (Exception e) {
             // If adding a synonym fails, we still want to return the product to the view
             product = Product.read(params.id)
             // If a validation error occurs, we want to return those values to the view,
             // and set them as initialValues of the form, so a user can correct him/herself
             inputValues = params
             flash.error = e.message
+            if (e instanceof ValidationException) {
+                // If the ValidationError occurs, we don't want to display it as a flash error,
+                // but render it as productInstance's errors using <g:renderErrors>
+                flash.error = null
+                product.errors = e.errors
+            }
         }
         render(template: 'productSynonyms', model: [productInstance: product, inputValues: inputValues])
     }
@@ -1161,13 +1167,12 @@ class ProductController {
             command.data = excelImporter.data
 
             command.errors = null
-            excelImporter.validateData(command)
+            excelImporter.importData(command)
 
             if (command.errors.allErrors) {
                 flash.errors = command.errors
             } else {
                 flash.message = "Succesfully imported product synonyms"
-                excelImporter.importData(command)
             }
         }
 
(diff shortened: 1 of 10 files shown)


grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -56,6 +60,16 @@ class ProductSynonymDataService {
                 } else {
                     command.errors.reject("Row ${index + 1}: Locale '${params['locale']}' is not a supported locale")
                 }
+
+                // Validation not to allow having more than one synonym of DISPLAY_NAME type for one locale
+                if (!productService.validateMultipleSynonymsOfType(
```
awalkowiak, line 65: I think you could possibly do a custom validator on the domain side. Something like this:
```
static constraints = {
    name(nullable: false, maxSize: 255)
    locale(nullable: true, validator: { val, obj ->
        if (obj.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME) {
            def duplicates = obj.product?.synonyms?.findAll {Synonym synonym ->
                synonym.locale == val && synonym.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME
            }
            return duplicates.size() > 1 ? "error message" : true
        }
    })
    synonymTypeCode(nullable: false)
    updatedBy(nullable: true)
    createdBy(nullable: true)
}
```
This worked for me in this way but still requires some minor improvements and some further testing if some "lazy" won't occur (so far looks good).

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -1014,14 +1014,18 @@ class ProductController {
...
         } catch (IllegalArgumentException e) {
             // If adding a synonym fails, we still want to return the product to the view
             product = Product.read(params.id)
+            // If a validation error occurs, we want to return those values to the view,
+            // and set them as initialValues of the form, so a user can correct him/herself
+            inputValues = params
             flash.error = e.message
         }
-        render(template: 'productSynonyms', model: [productInstance: product])
+        render(template: 'productSynonyms', model: [productInstance: product, inputValues: inputValues])
```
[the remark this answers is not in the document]
kchelstowski: If I returned `params`, I would also return them in case if the synonym was successfully added - I want to return them only if an error occurs, that's why I created this additional variable.
Do you have any other idea of handling this?

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
kchelstowski: Sure, I will test it out - have never seen such a solution, thanks

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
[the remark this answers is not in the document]
awalkowiak: Ok, I guess it makes sense

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -56,16 +59,18 @@ class ProductSynonymDataService {
...
-    }
 
-    void importData(ImportDataCommand command) {
-        log.info "Import data " + command.filename
+            // If none validation error occured yet, try to add a synonym
+            if (!command.errors.allErrors) {
+                Synonym synonym = new Synonym(name: params['name'], locale: new Locale(params['locale']), synonymTypeCode: params['synonymTypeCode'])
+                product.addToSynonyms(synonym)
+                if (!synonym.validate() || !product.save(flush: true, failOnError: true)) {
+                    command.errors.reject("Row ${index + 1}: Validation error occured. Most probably you are trying to add multiple DISPLAY_NAME synonym for one locale")
```
awalkowiak, line 70: I don't like the "most probably" type of messages. Normally, I think you should have access to `synonym.errors`, however, in this case, please bring back the `validateData` and do rejects based on the real cause, not the "most probable" one.

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -16,10 +16,10 @@ import org.pih.warehouse.product.Product
 
 class ProductSynonymDataService {
 
-    def productService
     def grailsApplication
 
-    Boolean validateData(ImportDataCommand command) {
```
awalkowiak, line 22: Why you get rid of this validation?

awalkowiak (changes requested):

awalkowiak (approved):

PR #3835, opened 2023-02-10
OBPIH-5411 See product name translation based on "product locale" (Scope: outbounds & requests)

alannadolny:


src/js/components/stock-movement-wizard/request/EditPage.jsx
```diff
@@ -97,12 +97,19 @@ const AD_HOCK_FIELDS = {
...
+            >
+              <span className="d-flex align-items-center">
+                <span className="text-truncate">
+                  {value.translatedName ?? value.name}
+                </span>
+                {renderHandlingIcons(value ? value.handlingIcons : null)}
               </span>
-              {renderHandlingIcons(value ? value.handlingIcons : null)}
-            </span>
+            </Tooltip>
```
kchelstowski, line 112: generally I don't mind it, since it is not a big deal here, but be aware that making the `Tooltip` disabled instead of just making:
```
if (value.translatedName) { 
 return <Tooltip>...</Tooltip>
}
return <span>...</span>
```
you still render to DOM some additional `Tooltips` that are never actually used

src/js/utils/Select.jsx
```diff
@@ -107,9 +107,16 @@ class Select extends Component {
...
+      return (
+        <div className="p-1">
+          {value.name}
+        </div>);
+    }
+
     if (showLabelTooltip) {
       const valueMapped = multi && value ? this.props.value.map(v => v && v.label) : [];
-      const valueLabel = multi ? valueMapped.join(', ') : (value.label || value.name);
+      const valueLabel = multi ? valueMapped.join(', ') : value.label || value.name;
```
kchelstowski, line 119: are you sure those `( )` are not needed here? Nothing will break due to lack of it? Is there a reason you deleted them?

src/js/utils/Select.jsx
```diff
@@ -254,11 +261,18 @@ class Select extends Component {
       return null;
     };
 
+    const isTooltipEnabled = () => {
+      if (this.props.value.translatedName) {
+        return true;
+      }
+      return !(!showLabelTooltip || (showValueTooltip && this.props.value));
```
kchelstowski, line 268: this statement feels so complicated to me, I will try to think about a better statement and will let you know

src/js/utils/option-utils.jsx
```diff
@@ -84,7 +84,8 @@ export const debounceProductsFetch = (waitTime, minSearchLength, locationId) =>
             productCode: obj.productCode,
             handlingIcons: obj.handlingIcons,
             lotAndExpiryControl: obj.lotAndExpiryControl,
-            label: `${obj.productCode} - ${obj.name}`,
+            label: `${obj.productCode} - ${obj.translatedName ?? obj.name}`,
+            originalLabel: `${obj.productCode} - ${obj.name}`,
```
kchelstowski, line 88: more common way of doing it (as it was already done in some place, I do not exactly remember where) would be to replace `originalLabel` with `hasTranslatedName: !!obj.translatedName` and taking as an example one of your Tooltips from above:

Changing from:
```
<Tooltip
    html={<div className="text-truncate">{option.originalLabel}</div>}
    theme="dark"
    disabled={option.originalLabel === option.label}
    position="top-start"
  >
```
to:
```
<Tooltip
    html={<div className="text-truncate">{option.label}</div>}
    theme="dark"
    disabled={!option.hasTranslatedName}
    position="top-start"
  >
```

kchelstowski (changes requested):

src/js/utils/Select.jsx
kchelstowski: in my opinion to simplify it we could name the function as `isTooltipDisabled` since the `<Tooltip>`'s component property is named as `disabled`.
So we could have:
```
const isTooltipDisabled = () => {
    if (!this.props.value.translatedName) {
      return true;
    }
    return !showLabelTooltip || (showValueTooltip && this.props.value);
```

I know we should be avoiding negated boolean expression, but since the `<Tooltip>`'s component property is negated itself (it is `disabled` not `enabled`), I think the function should be named basing on Tooltip's component property.

It's up to code convention, so I'd like to hear from you @awalkowiak @jmiranda as we've had some discussions around it lately, what you think about it.
In my opinion leaving it as it is, the last statement is very hard to read

src/js/components/stock-movement-wizard/outbound/EditPage.jsx
```diff
@@ -83,12 +84,19 @@ const FIELDS = {
         defaultMessage: 'Product name',
         attributes: {
           formatValue: value => (
-            <span className="d-flex">
-              <span className="text-truncate">
-                {value.name || ''}
+            <Tooltip
```
drodzewicz, line 87: use the showValueTooltip and tooltipValue attributes instead _(see comment below)_

src/js/components/stock-movement-wizard/modals/PackingSplitLineModal.jsx
```diff
@@ -40,12 +41,19 @@ const FIELDS = {
         defaultMessage: 'Product name',
         attributes: {
           formatValue: value => (
-            <span className="d-flex">
-              <span className="text-truncate">
-                {value.name}
+            <Tooltip
```
drodzewicz, line 44: use the showValueTooltip and tooltipValue attributes instead _(see comment below)_

src/js/components/stock-movement-wizard/modals/SubstitutionsModal.jsx
```diff
@@ -87,24 +88,38 @@ const FIELDS = {
               style={{ color: option.color ? option.color : 'black' }}
               className="d-flex align-items-center"
             >
-              {option.label}
-              &nbsp;
-              {renderHandlingIcons(option.value ? option.value.handlingIcons : [])}
+              <Tooltip
```
drodzewicz, line 91: this Product field should probably be replaced by `<ProductSelectFiel />` instead of `<Select />`
I think that when I was working o implementing exact match I might have missed this modal but I think it would be a good idea to replace this with ProductSelectFiel

drodzewicz (changes requested):

drodzewicz (approved):

awalkowiak (approved):

PR #3836, opened 2023-02-10
OBPIH-5413 Add product translations for invoices

kchelstowski:
Added product translations for:

- invoice view page
- invoice add items page, confirm page
- included for prepayment invoice, regular invoice, final invoice

### DIFF
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
@@ -173,6 +173,9 @@ class InvoiceItem implements Serializable {
                 unitPrice: unitPrice,
                 totalAmount: totalAmount,
                 totalPrepaymentAmount: totalPrepaymentAmount,
+                orderAdjustment: orderAdjustment,
+                productName: product?.name,
+                translatedProductName: product?.translatedName
         ]
     }
 }
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
@@ -93,10 +93,7 @@ class InvoiceItemCandidate {
     }
 
     Boolean isAdjustment() {
-        if (adjustmentAmount || adjustmentPercentage) {
-            return true
-        }
-        return false
+        return adjustmentAmount || adjustmentPercentage
     }
 
     Map toJson() {
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
@@ -15,6 +15,7 @@ import org.pih.warehouse.core.GlAccount
 import org.pih.warehouse.core.Organization
 import org.pih.warehouse.core.UnitOfMeasure
 import org.pih.warehouse.order.Order
+import org.pih.warehouse.product.Product
 import org.pih.warehouse.shipping.Shipment
 
 class InvoiceItemCandidate {
@@ -58,7 +59,7 @@ class InvoiceItemCandidate {
     static constraints = {
     }
 
-    static transients = ['unitOfMeasure', 'candidateUnitPrice']
+    static transients = ['unitOfMeasure', 'candidateUnitPrice', 'associatedProduct', 'adjustment']
 
     String getUnitOfMeasure() {
         if (quantityUom) {
@@ -84,6 +85,20 @@ class InvoiceItemCandidate {
         return unitPrice
     }
 
+    Product getAssociatedProduct() {
+        if (productCode) {
+            return Product.findByProductCode(productCode)
+        }
+        return null
+    }
+
+    Boolean isAdjustment() {
+        if (adjustmentAmount || adjustmentPercentage) {
+            return true
+        }
+        return false
+    }
+
     Map toJson() {
         return [
             id: id,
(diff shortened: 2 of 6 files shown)


grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
```diff
@@ -84,6 +85,20 @@ class InvoiceItemCandidate {
...
+        }
+        return null
+    }
+
+    Boolean isAdjustment() {
+        if (adjustmentAmount || adjustmentPercentage) {
+            return true
+        }
+        return false
+    }
```
alannadolny, line 100: I am not sure, but I think it can be simplified to: `return adjustmentAmount || adjustmentPercentage`. Correct me if I'm wrong.

kchelstowski: it would then return not a Boolean, but the actual value of either `adjustmentAmount` or `adjustmentPercentage`

drodzewicz: If you want you can then cast this value to boolean using double bang operator 
```
return !!(adjustmentAmount || adjustmentPercentage)
```

kchelstowski: I know, but I also know that Justin is not a big fan of it, so I wanted to do it in a more common way

drodzewicz: Actually I think that since you have defined the function return type as _Boolean_ it should cast its return value implicitly.
I am not 100% sure about this but I think I have noticed this behavior before, so even if you would have
`return adjustmentAmount || adjustmentPercentage` 

By the way I am totally fine with how the function look right now and it is not a change request

kchelstowski: You guys were right - it is working fine - I expected it to work as it is in JS, so if you have something like:
`return "string" || null`, I'd expect it to return `"string"`, but it seems like Groovy is cleverer than I thought and if you write `||` or `&&` it parses to `Boolean` by itself, so in the case above I'd get `true` as the result, not `"string"`.
It's nice to know, as it's another Groovy's feature that works in a different way than in JS.

jmiranda:
@awalkowiak Don't merge this. I have some changes/comments, but need more time to review.

grails-app/views/invoice/_invoiceItems.gsp
```diff
@@ -24,7 +24,12 @@
                         ${invoiceItem?.product?.productCode?:g.message(code:'default.all.label')}
                     </td>
                     <td>
-                        ${invoiceItem?.description}
+                        <g:if test="${invoiceItem?.orderAdjustment}">
+                            ${invoiceItem?.description}
+                        </g:if>
+                       <g:else>
+                           <g:translatedProductName product="${invoiceItem?.product}" />
+                       </g:else>
```
jmiranda, line 32: Check out the FormatTagLib. I'd prefer to use something like that. We could even add a tag for InvoiceItem. But we at least want to use the following tag 
```
<format:product product="${product}"/>
```
but rewritten to use synonyms instead of the pipe-delimited convention.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -173,6 +173,9 @@ class InvoiceItem implements Serializable {
                 unitPrice: unitPrice,
                 totalAmount: totalAmount,
                 totalPrepaymentAmount: totalPrepaymentAmount,
+                orderAdjustment: orderAdjustment,
+                productName: product?.name,
+                translatedProductName: product?.translatedName
```
jmiranda, line 178: I'm debating how to handle this so I think we should discuss this on a tech huddle next week.

jmiranda: Options: 
1. As-is
1. Return a representation of the product that includes the translated name(s) by type  
    1. Product.toJson() - This could get expensive if we have to render this 100s of times
    2. Product.toAnotherJson() that is used less often 
1. Create a new ProductName DTO that takes the product and synonyms and renders an object in a particular locale that conforms to our naming rules (i.e. one display_name per locale). 
    2. productName: [displayName: "english display name", brandNames: ["english brand name 1", "english brand name 2"]   
1. There might be other options as well

jmiranda (commented):
discuss on tech huddle

grails-app/domain/org/pih/warehouse/invoice/InvoiceItemCandidate.groovy
jmiranda: As far as I know the following would definitely return a boolean
```
Boolean isAdjustment() { 
    return (adjustmentAmount || adjustmentPercentage)
}
```

jmiranda: Oh sorry, just read @drodzewicz's comment. And yeah this would be the only way to go.

grails-app/views/invoice/_invoiceItems.gsp
jmiranda:
```
<format:name product="${product}"/>
<format:displayName product="${product}"/>
<format:alternateName product="${product}"/>
...
<format:name object="${invoiceItem.product}"/>
<format:displayName object="${invoiceItem.product}"/>
<format:alternateName product="${product.product}"/>
```

jmiranda (approved):
Since we already have tons of PRs doing the same thing so we'll merge this as-is and we'll create a new ticket for the refactoring.

PR #3839, opened 2023-02-13
OBPIH-5338 Fix picklist pdf not displaying translated product names

drodzewicz:


jmiranda:
@drodzewicz @awalkowiak @kchelstowski This one is ready for review.

jmiranda:
@awalkowiak please use squash & merge if you're the one to merge it.

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -20,6 +20,7 @@ class PicklistController {
 
     def picklistService
     def pdfRenderingService
+    def localizationService
```
kchelstowski, line 23: is it still needed?

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -55,7 +56,6 @@ class PicklistController {
 
         renderPdf(
                 template: "/picklist/print",
-                //locale:locale,
```
kchelstowski, line 58: should be removed or uncommented?

grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
```diff
@@ -64,15 +64,7 @@ class LocalizationService {
...
-        // First search for session locale, if none set, then search for user's default locale, otherwise get the default locale
-        def session = RequestContextHolder.currentRequestAttributes().getSession()
-        if (session?.locale) {
-            return session.locale
-        }
-        if (session?.user?.locale) {
-            return session.user.locale
-        }
-        return new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
+        return LocaleContextHolder.locale ?: new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
```
kchelstowski, line 67: as I wrote in the ticket - I will check it locally, since I want to understand the need of that change and why it was not working and if it didn't cause any regression issues related to crowdin

kchelstowski (changes requested):

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -678,22 +680,44 @@ class Product implements Comparable, Serializable {
...
     }
 
     String getTranslatedNameWithLocaleCode() {
         String localeTag = LocalizationUtil.localizationService.getCurrentLocale().toLanguageTag().toUpperCase()
         return "${name}${translatedName ? " (${localeTag}: ${translatedName})" : ''}"
     }
 
-    String getTranslatedName() {
-        Locale currentLocale = LocalizationUtil.localizationService.getCurrentLocale()
+    Synonym getSynonymByLocale(SynonymTypeCode synonymTypeCode, Locale locale) {
```
drodzewicz, line 691: If nothings has changed I believe that only **DISPLAY_NAME** has a validation for having one **DISPLAY_NAME** per locale and other typeCodes can have multiple.

So if we are making this `getSynonymByLocale()` function more generic, that it can return synonyms of other SynonymTypeCodes and not just **DISPLAY_NAME**, it should probably return a `List<Synonym>` because there might be cases where a product has multiple `ALTERNATIVE_NAME` or `BRAND_NAME` in one locale.

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
jmiranda: That's what I'm talking about! This is the kind of nitpicky review comments I'm looking for.

grails-app/domain/org/pih/warehouse/product/Product.groovy
jmiranda: Good point. Thanks for bringing that up. I'll try to figure out how to handle that.

jmiranda: This would actually be a great place for some unit tests.

kchelstowski (changes requested):
As I explained in the ticket, we should consider adding the `LocaleContextHolder.setLocale()` to the `InitializationFilters`

awalkowiak (approved):

PR #3843, opened 2023-02-15
OBPIH-5412 See product name translation based on "product locale" (Scope: PO and inbounds)

drodzewicz:


jmiranda (commented):
Just out of curiosity, what necessitates the need for a new component (ProductSelectField) here? 

Is it the fact that we're translating the name? If so, can we maybe make this more generic LocalizedSelectField? Or make sure that all data passed to the SelectField has a localized name / label. Just a thought.

drodzewicz:
> Just out of curiosity, what necessitates the need for a new component (ProductSelectField) here?

the **ProductSelectField** component is not a new component in the context of this PR, it was created for the case of _"select exact match product when pressing enter"_.
That case kind of required another custom select component because it was quite hard to implement selection of the the _exactMatch_ by using props provided by the `react-select` library

awalkowiak (approved):

PR #3845, opened 2023-02-16
OBPIH-5400 Allow search by Budget Code

alannadolny:


grails-app/controllers/org/pih/warehouse/core/BudgetCodeController.groovy
```diff
@@ -10,14 +10,21 @@
...
     def list = {
         params.max = Math.min(params.max ? params.int('max') : 10, 100)
-        [budgetCodes: BudgetCode.list(params), budgetCodesTotal: BudgetCode.count()]
+
+        if (params.q) {
+            def budgetCodeList = budgetCodeService.findBudgetCodes(params.q)
+            return [budgetCodes: budgetCodeList, budgetCodesTotal: budgetCodeList.size()]
+        }
+
+        return [budgetCodes: BudgetCode.list(params), budgetCodesTotal: BudgetCode.count()]
```
kchelstowski, line 27: 1. You could move this `if` directly to the `createCriteria`, so you would only have here:
```
List<BudgetCode> budgetCodes = budgetCodeService.list(params)
return [budgetCodes: budgesCodes, totalCount: budgetCodes.totalCount()]
```

2. Doing this:
```
budgetCodesTotal: budgetCodeList.size()
```
is wrong, because you won't get the `totalCount` for pagination by doing so, but only the size of actual results, so if your page has 10 results per page, the size would always be <= 10

grails-app/services/org/pih/warehouse/core/BudgetCodeService.groovy
```diff
@@ -0,0 +1,21 @@
...
+
+class BudgetCodeService {
+
+    boolean transactional = true
+
+    def findBudgetCodes(String term) {
+        return BudgetCode.createCriteria().list {
+                ilike("code", "%" + term + "%")
+        }
+    }
```
kchelstowski, line 20: I don't see the pagination being done here

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/core/BudgetCodeService.groovy
```diff
@@ -0,0 +1,25 @@
...
+* the terms of this license.
+* You must not remove this notice, or any other, from this software.
+**/
+package org.pih.warehouse.core
+
+import org.codehaus.groovy.grails.web.servlet.mvc.GrailsParameterMap
+
+class BudgetCodeService {
+
+    boolean transactional = true
```
awalkowiak, line 16: Is it required?

kchelstowski (approved):

PR #3846, opened 2023-02-16
OBPIH-5397 Revert date formatting on product.lastUpdated property and…

kchelstowski:
… a few more

I also found that for "fresh" PO list, Stock transfer list and its marshallers the same formatting was done as for product, so I also reverted it there.

I had a doubt about naming this constant, so I'd be grateful for any advice from you - for now I named it as `tableDataFormat`, as this format was initially implemented for new, redesigned list pages.

As discussed - I return this format to the AppContext, so that we have access to it from the `sessionReducer`.
I also added a custom component for the `react-table` cell, named `DateCell` not to repeat the same logic and formatting for all the list pages I've edited.
I also added a few unit tests for this component, so you have a better understanding of what is going on.


src/js/components/DataTable/DateCell.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+const DateCell = ({ tableDateFormat, ...row }) => (
+  <TableCell
+    {...row}
+    value={row.value && moment(row.value).format(tableDateFormat)}
+    defaultValue="-"
+  />
+);
+
+const mapStateToProps = state => ({
```
drodzewicz, line 17: is there a reason why are not using `useSelector` hook to access redux store values ?

src/js/components/DataTable/DateCell.jsx
```diff
@@ -0,0 +1,26 @@
...
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import TableCell from 'components/DataTable/TableCell';
+
+const DateCell = ({ tableDateFormat, ...row }) => (
+  <TableCell
+    {...row}
+    value={row.value && moment(row.value).format(tableDateFormat)}
+    defaultValue="-"
```
drodzewicz, line 13: shouldn't **defaultValue** also be customizable instead of hardcoding a **"-"**?

src/js/components/DataTable/DateCell.jsx
kchelstowski: `mapStateToProps` and `useSelector` are replaceable, and it's just my own preference to use `mapStateToProps` inside the components and the `useSelector` for e.g. hooks. If it bothers you, I could change it, but answering the question in a nutshell - it's just a personal preference

src/js/components/DataTable/DateCell.jsx
kchelstowski: it could be, why not, I didn't pay that much attention to `defaultValue` since it was only copied from the already existing code, but makes sense

src/js/components/DataTable/DateCell.jsx
drodzewicz: ok, I understand
I marked it as a comment and not a change request, just wanted to know if there is a technical reason for not using hooks
but I have nothing against the current version

src/js/tests/utils/DateCell.test.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { render, screen, within } from '@testing-library/react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import store from 'store';
+
+import DateCell from 'components/DataTable/DateCell';
+
+import '@testing-library/jest-dom';
+
+let dateCell;
```
alannadolny, line 11: Is it necessary to declare dateCell as a let? As i can see you are assigning value to this variable in each "it" function, so I think you can use something like: `const dateCell = ...` in each function

alannadolny (commented):
Other than that LGTM

src/js/tests/utils/DateCell.test.jsx
kchelstowski: @alannadolny it is working fine, so I changed it

alannadolny (approved):

drodzewicz (approved):

awalkowiak (approved):
I think this one is ok, @jmiranda can you take a look if config properties look good to you?

src/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -34,6 +34,8 @@ class Constants {
     static final String EXPIRATION_DATE_FORMAT = "MM/dd/yyyy"
     static final String DELIVERY_DATE_FORMAT = "MM/dd/yyyy HH:mm XXX"
     static final String EUROPEAN_DATE_FORMAT = "dd/MM/yyyy"
+    static final String TABLE_MONTH_YEAR_DATE_FORMAT = "MMM DD, yyyy"
+    static final String TABLE_DATE_DEFAULT_VALUE = "-"
```
jmiranda, line 38: Something is bothering me about the use of "TABLE" here. I understand the need to add context and I'm aware that naming is hard. But TABLE seems arbitrary since we might actually display this outside of a table. So let's try to come up something generic, but not restricting. I think DISPLAY_DATE_FORMAT or DEFAULT_DISPLAY_DATE_FORMAT would be fine for now.

src/js/components/DataTable/DateCell.jsx
jmiranda: Just out of curiosity, I would like to know more about this. Can you provide a basic ELI5 for why you would use mapStateToProps over the useSelector hook?

I've read the following post on StackOverflow so just want to hear your thoughts.
https://stackoverflow.com/questions/59748180/should-i-use-useselector-usedispatch-instead-of-mapstatetoprops

kchelstowski:
@jmiranda as I explained to Dariusz - it's rather my personal preference and a habit - I was taught this way, as `mapStateToProps` is an older way of injecting redux state to the component. They both do the same, but in a different approach. 
For me having `mapStateToProps` in components is more readable, because we separate the redux state logic from the component, so we always know where to find it (at the end of the file, after the component definition), whereas `useSelector` might be placed anywhere in the component.
The advantage of using `mapStateToProps` for our project, where we do not have TypeScript is also that using the pieces of state as the props, the ESLint will force us to write the `propTypes` for those, whereas with the `useSelector`, we don't inject those pieces of state through the component's props.

1. `mapStateToProps` approach (a stupid, not existing example, because we probably won't have the quantity in the state, but that came to my mind - I just wanna show the typing difference I explained above):
```
const MyComponent = ({ quantity }) => {
    return (
      <div>{quantity}</div>
    );
}

const mapStateToProps = (state) => ({
  quantity: state.productReducer.quantity,
})

export default connect(mapStateToProps)(MyComponent);


MyComponent.propTypes = {
  quantity: PropTypes.number.isRequired,
}
```

You can see, that by using `mapStateToProps` approach here, I know that the `quantity` prop is type of **number**, not string, as it let's say, might happen in some places (a stupid example, but that came to my mind at first thought)

2. `useSelector` approach:
```
const MyComponent = () => {
    const { quantity, somethingElse } = useSelector(state => ({
       quantity: state.productReducer.quantity,
       somethingElse: state.somethingElseReducer.something,
    }));

    return (
      <div>{quantity}</div>
    );
}

export default MyComponent;
```

Here you can see that I do not know the type of those and the ESLint doesn't force me to check it and write propTypes for it, so if a person who didn't work on this component comes here to do something with it, he/she can't look at the `propTypes` to know, what type of those are, but he/she needs to `console.log` it first to get to know that.
With the TypeScript we wouldn't have such a problem, because it would force us to create an `interface` or `type` for this `useSelector`'s state

I hope I gave my point of view now more clearly, if there is anything you would want to know more, let me know.

PR #3847, opened 2023-02-16
OBPIH-5411 See product name translation based on "product locale" (Scope: outbounds & requests) fixes after QA

alannadolny:


awalkowiak (changes requested):
Other than one minor thing it looks good to me

awalkowiak (approved):

PR #3848, opened 2023-02-16
OBPIH-5237 Fix merging inventory items and PA when products had null …

awalkowiak:
…or empty lot number

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -996,15 +996,10 @@ class ProductAvailabilityService {
                 existingPA.quantityAllocated += remainingPA.quantityAllocated
                 existingPA.quantityNotPicked += remainingPA.quantityNotPicked
                 existingPA.quantityOnHold += remainingPA.quantityOnHold
-                existingPA. quantityAvailableToPromise += remainingPA.quantityAvailableToPromise
+                existingPA.quantityAvailableToPromise += remainingPA.quantityAvailableToPromise
                 existingPA.save(flush: true)
 
-                remainingPA.quantityOnHand = 0
-                remainingPA.quantityAllocated = 0
-                remainingPA.quantityNotPicked = 0
-                remainingPA.quantityOnHold = 0
-                remainingPA. quantityAvailableToPromise = 0
-                remainingPA.save(flush: true)
+                remainingPA.delete(flush: true)
 
                 return
             }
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -365,6 +359,19 @@ class ProductMergeService {
...
 
+    /**
+     * Finds inventory item from primary product, that has the same lot and expiry date as obsoleteInventoryItem
+     * (null lotNumber and empty string lotNumber are treated equally)
+     * */
+    def getPrimaryInventoryItem(Set<InventoryItem> primaryInventoryItems, InventoryItem obsoleteInventoryItem) {
+        def obsoleteLotNumber = obsoleteInventoryItem?.lotNumber
+        def obsoleteExpirationDate = obsoleteInventoryItem?.expirationDate
+        return primaryInventoryItems?.find {
+            def exactLotNumber = obsoleteLotNumber ? (it.lotNumber == obsoleteLotNumber) : (it.lotNumber == null || it.lotNumber == "")
```
jmiranda, line 370: Don't forget about the whitespace case i.e. if someone enters more than one space character or worse ... tabs, returns, etc.

For readability, it might be better to move this method to Product.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -996,15 +996,10 @@ class ProductAvailabilityService {
                 existingPA.quantityAllocated += remainingPA.quantityAllocated
                 existingPA.quantityNotPicked += remainingPA.quantityNotPicked
                 existingPA.quantityOnHold += remainingPA.quantityOnHold
-                existingPA. quantityAvailableToPromise += remainingPA.quantityAvailableToPromise
+                existingPA.quantityAvailableToPromise += remainingPA.quantityAvailableToPromise
```
jmiranda, line 999: I know it's kind of long, but I'd prefer to spell existingPA out as existingProductAvailability for readability. Same with remainingPAs, et al. 

Remind me why we need to do this instead of just rerunning the product availability once all transactions have been "fixed".

jmiranda (changes requested):
Is there a compelling reason to have methods like this 
```
PicklistService.updatePicklistItemsOnProductMerge()
```
rather than
```
ProductMergeService.updatePicklistItems()
```

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: @jmiranda I did all the requested changes. I do not remember why I went this route. I only remember that I initially did with the refresh, but then I changed to this approach and pushed this version in the first commit/PR for this feature.

PR #3852, opened 2023-02-17
OBPIH-5367 Fix validation for multiple DISPLAY_NAME synonyms in a sin…

kchelstowski:
…gle import

It is my bad, because I haven't checked adding two `DISPLAY_NAME` synonyms in a single import, but was only checking it for the case, where product already had a `DISPLAY_NAME` synonym.

The idea of the fix is that when we loop over the rows and whenever we add a `DISPLAY_NAME` synonym, I'm adding this `productCode` to a List, and then in the `if` statement I check not only if the product already has the synonym of `DISPLAY_NAME` (in the database), but also "in cache" (in cache = we want to add the `DISPLAY_NAME` for this product in any line above for this single import).

I also fixed some weird indents in the lines from yesterday, so actually the only lines I added is `25-27` and `76`, `80`.

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
@@ -68,12 +68,12 @@ class ProductSynonymDataService {
             // Check if none error occures to be sure, that we have correct product, locale and synonymTypeCode
             // before we check the duplicates
             if (!command.errors.allErrors && synonymTypeCode == SynonymTypeCode.DISPLAY_NAME) {
-                Set<Synonym> duplicates = product?.synonyms?.findAll { Synonym synonym ->
+                Synonym duplicate = product?.synonyms?.find { Synonym synonym ->
                     synonym.locale == new Locale(params['locale']) && synonym.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME
                 }
                 // If the product already has a synonym of type DISPLAY_NAME and a locale
                 // or we are trying to add it in any of the line above for this single import, throw a validation error
-                if (duplicates || productsWithDisplayName.find{ it?.code == product?.productCode && it?.locale == params['locale'] }) {
+                if (duplicate || productsWithDisplayName.find{ it?.code == product?.productCode && it?.locale == params['locale'] }) {
                     command.errors.reject("Row ${index + 1}: You cannot add multiple display names in the same language. Edit the existing synonym instead.")
                     return
                 }
--- a/grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
@@ -22,6 +22,9 @@ class ProductSynonymDataService {
     Boolean validateData(ImportDataCommand command) {
         log.info "Validate data " + command.filename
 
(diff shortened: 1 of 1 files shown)


alannadolny (approved):

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -65,12 +68,16 @@ class ProductSynonymDataService {
...
+                Set<Synonym> duplicates = product?.synonyms?.findAll { Synonym synonym ->
+                    synonym.locale == new Locale(params['locale']) && synonym.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME
+                }
+                // If the product already has a synonym of type DISPLAY_NAME
+                // or we are trying to add it in any of the line above for this single import, throw a validation error
+                if (duplicates || productsWithDisplayName.find{ it == product?.productCode }) {
+                    command.errors.reject("Row ${index + 1}: You cannot add multiple display names in the same language. Edit the existing synonym instead.")
+                    return
+                }
+                productsWithDisplayName.add(product?.productCode)
```
awalkowiak, line 80: I think this also should be tied with `locale`, since you can have multiple display names in different locales.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
kchelstowski: Oh yeah, that’s a bummer. Thanks for catching that

grails-app/services/org/pih/warehouse/data/ProductSynonymDataService.groovy
```diff
@@ -65,12 +68,16 @@ class ProductSynonymDataService {
             // Check if none error occures to be sure, that we have correct product, locale and synonymTypeCode
             // before we check the duplicates
             if (!command.errors.allErrors && synonymTypeCode == SynonymTypeCode.DISPLAY_NAME) {
-               Set<Synonym> duplicates = product?.synonyms?.findAll { Synonym synonym ->
-                   synonym.locale == new Locale(params['locale']) && synonym.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME
-               }
-               if (duplicates) {
-                   command.errors.reject("Row ${index + 1}: You cannot add multiple display names in the same language. Edit the existing synonym instead.")
-               }
+                Set<Synonym> duplicates = product?.synonyms?.findAll { Synonym synonym ->
```
drodzewicz, line 71: This is probably a non-problem, but I'll just mention it.
We probably don't need to do `synonyms.findAll{}` to check for duplicate DISPLAY_NAME synonym. 
If it returns at least one item we can be sure that this product already has a displayName synonym in this locale, so I guess doing this should be enought
```
Synonym displayNameSynonym = product?.synonyms?.find { Synonym synonym ->
    synonym.locale == new Locale(params['locale']) && synonym.synonymTypeCode == SynonymTypeCode.DISPLAY_NAME
}
if (displayNameSynonym || productsWithDisplayName.find{ it?.code == product?.productCode && it?.locale == params['locale'] }) { 
   ... 
}
```

kchelstowski: you are right, thanks for catching that

awalkowiak (approved):

PR #3855, opened 2023-02-17
OBPIH-5422 Show translated name and name in one column for Export pro…

kchelstowski:
…duct demand report

By misunderstanding I made a separate column for `name` and separate for the `translatedName`.
Following the convention for other reports, we want to have it in one column with: `<name> (<locale>: <translatedName>)`

Unfortunately since we use pure SQL to get this data, I had to do collect in order to achieve this effect.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
@@ -23,7 +23,7 @@ import org.pih.warehouse.product.Product
 import org.pih.warehouse.report.ChecklistReportCommand
 import org.pih.warehouse.report.InventoryReportCommand
 import org.pih.warehouse.report.MultiLocationInventoryReportCommand
-import org.pih.warehouse.util.LocalizationUtil
+
 import org.quartz.JobKey
 import org.quartz.impl.StdScheduler
 import util.ReportUtil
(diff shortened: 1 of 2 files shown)


grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -188,6 +188,31 @@ class ReportController {
         Location location = Location.get(session.warehouse.id)
         def data = forecastingService.getDemandDetails(location, null)
         if (params.downloadFormat == "csv") {
+            String currentLocaleTag = LocalizationUtil.localizationService.getCurrentLocale().toLanguageTag().toUpperCase()
+            data = data.collect {
```
awalkowiak, line 192: I don't like this approach since this report can be very big. Is there a way to achieve it through SQL query in the `getDemandDetails` method?

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
kchelstowski: At the moment I don’t have any clue. We’ve followed this way for Alan’s reports as well (in each of this report the collect was done). I can dig it on Monday to try it out, but I don’t feel like I will be able to merge two columns in one using the sql query

kchelstowski:
@awalkowiak I managed to do it using pure SQL - using `CONCAT` + `IFNULL` functions - thanks for pointing that out, since it might be useful in the future, e.g. if we face some performance issues, where this case was handled by `.collect` for other reports.

awalkowiak (approved):

PR #3857, opened 2023-02-20
OBPIH-5393 Edit product synonym

drodzewicz:
- add edit button on actions column of synonym table
- add synonym dialog for editing product synonym


grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -98,7 +98,7 @@ class ErrorsController {
             response.status = 400
             BeanPropertyBindingResult errors = request?.exception?.cause?.errors
             def errorMessages = errors.allErrors.collect {
-                return messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), localizationService.getCurrentLocale())
+                return g.message(error: it, args: it.arguments, default: it.defaultMessage, locale: localizationService.getCurrentLocale())
```
kchelstowski, line 101: If we don't want to cause any regression related to this, we can always add some additional arguments and do an `if` whether to return what you did, or the `messageService` thing.
Moreover - I think you could just simply catch the exception yourself, just doing `try/catch` inside `editProductSynonym` action, so you would not be dependent on this handler.

If we want to proceed with this change, we should at the very least change the `default` from:
```
default: it.defaultMessage
```
to what it was:
```
default: (it.defaultMessage ?: it.codes[0])
```

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -1028,6 +1028,20 @@ class ProductController {
...
+     */
+    def editProductSynonym = {
+        println "editProductSynonym() " + params
+        productService.editProductSynonym(params['synonym.id'], params.synonymTypeCode, params.synonym, params.locale)
+        render (status: 200, text: "successfully edited synonym")
+    }
+
+    def editProductSynonymDialog = {
+        Synonym synonym = Synonym.read(params.id)
+        render(template: 'productSynonymsEdit', model: [product: synonym.product, synonym: synonym])
```
kchelstowski, line 1042: I know this comment and below might be nitpicky - but just in case shouldn't we add as much null checks as we can? (here, `synonym?.product` - I don't know, in case someone deleted the product "in the mean time")

grails-app/views/product/_productSynonyms.gsp
```diff
@@ -39,8 +39,15 @@
                             ${g.message(code: 'enum.SynonymTypeCode.' + synonym.synonymTypeCode)}
                         </td>
                         <td class="middle">
+                            <a href="javascript:void(0);" class="button btn-show-dialog"  data-width="900"
+                               data-title="${warehouse.message(code:'inventoryLevel.edit.label', default: 'Edit synonymss')}"
```
kchelstowski, line 43: should the code stay `inventoryLevel...`? I can also see a typo - "Edit synonyms**s**"

grails-app/views/product/_productSynonyms.gsp
```diff
@@ -91,9 +98,9 @@
                                     class="medium text"
                                     placeholder="${g.message(code: 'synonym.typeSynonym.placeholder.label', default: 'Type the synonym here')}"
                                 />
-                                <button  class="button">
+                                <button class="button">
                                     <img src="${createLinkTo(dir:'images/icons/silk', file:'add.png')}" />&nbsp;
-                                    ${warehouse.message(code:'default.button.add.label')}
+                                    <g:message code="default.button.add.label"/>
```
kchelstowski, line 103: is there a reason for this change? I know @jmiranda is not a big fan of changing those without a reason in terms of OBGM rebases

grails-app/views/product/_productSynonymsEdit.gsp
```diff
@@ -0,0 +1,127 @@
+<%@ page import="org.pih.warehouse.core.SynonymTypeCode" %>
+<g:form name="editSynonym" action="editProductSynonym" >
+    <g:hiddenField name="product.id" value="${product.id}" />
+    <g:hiddenField name="synonym.id" value="${synonym.id}" />
```
kchelstowski, line 4: as above - I'd consider adding null check here and in the below lines - `synonym?.locale?.language` etc

grails-app/views/product/_productSynonymsEdit.gsp
```diff
@@ -0,0 +1,127 @@
...
+      url:'${g.createLink(controller:'product', action:'editProductSynonym')}',
+      data: data,
+      success: function() {
+        window.location.href = "${request.contextPath}/product/edit/${product.id}";
+      },
+      error: function(jqXHR, textStatus, errorThrown) {
+        if (jqXHR.responseText) {
+          try {
+            let data = JSON.parse(jqXHR.responseText);
+            if (data.errorMessages.length > 0) {
```
kchelstowski, line 100: no need of checking `length > 0`, could be just `if (data.errorMessages.length)` - (I know it was probably copied from somewhere)

kchelstowski (changes requested):
The core LGTM, as for this error messages parsing - I'd consider trying to catch it by yourself using `try/catch` if we are afraid of causing any regression related to this change in the handler.

grails-app/views/product/_productSynonyms.gsp
drodzewicz: I didn't notice that, thanks

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
drodzewicz: Yes I am a bit wary of any potential regressions,  so I guess the safest option for me in this situation is to just do the `g.message(error: e.error )` inside of a catch block for edit product synonym

kchelstowski (approved):
If you had made sure that the injection of `localizationService` works fine inside the `ProductController`, LGTM

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: It wouldn't hurt (in case someone removes the checks above it), but it's unnecessary in the current context. But again, it doesn't hurt and I like to be defensive assuming it's not decreasing readability.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -1028,6 +1029,34 @@ class ProductController {
...
+     * Edit existing product synonym
+     */
+    def editProductSynonym = {
+        println "editProductSynonym() " + params
+        try {
+            productService.editProductSynonym(params['synonym.id'], params.synonymTypeCode, params.synonym, params.locale)
+            render (status: 200, text: "successfully edited synonym")
+        } catch (ValidationException e) {
+            def errorMessages = e.errors.allErrors.collect {
+                g.message(error: it, locale: localizationService.currentLocale)
```
jmiranda, line 1042: @drodzewicz Is this the code you were talking about earlier today? You're overriding how you handle the validation exception instead of letting ErrorsController handle it? 

I don't mind going this direction but we definitely should 
1. document (maybe in API docs) both approaches (specifically what's different with the JSON response for each)
2. consider adopting a consistent approach 
3. unit test each approach

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1411,4 +1411,15 @@ class ProductService {
         }
         return product
     }
+
+    Synonym editProductSynonym(String synonymId, String synonymTypeCodeName, String synonymValue, String localeName) {
+        Synonym synonym = Synonym.get(synonymId)
+        Locale locale = localeName ? new Locale(localeName) : null
+        SynonymTypeCode synonymTypeCode = synonymTypeCodeName ? SynonymTypeCode.valueOf(synonymTypeCodeName) : SynonymTypeCode.ALTERNATE_NAME
```
jmiranda, line 1418: Can you provide some rationale for this? Something like "it's in the requirements, dummy" would suffice. 

If this was just an assumption you made, I might prefer throwing an exception if the synonym type code is not specified. But would like to hear from others as well.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1411,4 +1411,15 @@ class ProductService {
...
+        synonym.properties = [ locale: locale, synonymTypeCode: synonymTypeCode, name: synonymValue ]
+        if (!synonym.validate() || !synonym.save(flush: true, failOnError: true)) {
+            throw new ValidationException("Invalid synonym", synonym.errors)
```
jmiranda, line 1421: `failOnError:true` seems like overkill here since both validate() and save() should lead us to a validation exception. failOnError will probably throw it's own ValidationException so we just need to make sure that both lead to the same outcome. I'm not sure how to test that out though.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: Note: you don't have to do any of those things. We just need to get them into a tech debt task list.

drodzewicz: Yes it is.
Initially I wanted to avoid manually catching the exception and let the error handler on `ErrorController` handle it for me, but as I have mention on the tech huddle, the way the current errorHandler is implemented it takes just the first exception code and translates it and omits the others.
So my choice was either refactor the error handler by replacing `messageService.getMessage(code: it.codes[0], ...)` with `g.message(error: it)` or catch the exception and handle the validation code translations manually.
I chose the latter just because I wasn't sure if replacing the messageService with g.message would introduce any regression

grails-app/services/org/pih/warehouse/product/ProductService.groovy
drodzewicz: No ti was not in the requirements, this was just my assumption based on the **addSynonymToProduct** which defaults to ALTERNATE_NAME.
I agree, in case of an edit it should probably throw an exception

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: When you say "the first exception code" what are you expecting? Can you provide more details, maybe an example?

I'm trying not to ask a leading question here so as not to bias your response.

jmiranda: TIL there's a "History for Selection" feature that allows you to see the history of a single line (or presumably multiple lines). 

https://watch.screencastify.com/v/DAxx4F6CQDAX34uuy7rg

drodzewicz: Currently the error handler looks like this 
**ErrorController.groovy**
```
    def handleValidationErrors = {
            response.status = 400
                return messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), localizationService.getCurrentLocale())
            }
            render([errorCode: 400,
                    errorMessage: "Validation error. " + request?.exception?.cause?.fullMessage,
                    errorMessages: errorMessages
            ] as JSON)
            return
        }
        render(view: "/error")
    }

```
From my understanding when it does the _collect_ of the errors list
```errors.allErrors.collect{ return messageService.getMessage(it.codes[0], ...) }```
it essentially just tries to translate all of the errors and return the appropriate label _(if it exists)_ as an error message.

You can see on the image below, how does the error instance look inside. 
_(the highlighted code is a code that I am expecting it to return because it is defined in messeges.properties file)_

image

So the problem in our implementation of this error handler is, is that we only translate the first Error code `it.code[0]`, for which label in **messages.properties** does not exist and it just returns a default message instead.

What I am expecting for it to return is an existing error message that we have defined in the messages.properties file _(highlighted on the screenshot )_

**messages.properties**
```
synonym.locale.validator.multipleError=You cannot add multiple display names in the same language. Edit the existing synonym instead.
```

if we use `g.message(error: it)` instead of `messageService.getMessage(code: it.codes[0])` what I have observed is that it goes through the whole list of `error.codes` and tries to match the first code that is defined in the messages.properties file and return it as an error message

**Comment by drodzewicz** [comment]
I have pushed a new commit that should resolve all the comments that you @jmiranda have previously left.
Other than the discussion that we currently have about the error handler I think this PR can be reviewed again and maybe if everything looks good can be merged 
cc @awalkowiak

File: grails-app/controllers/org/pih/warehouse/product/ProductController.groovy (hunk 7 above)
Comment (jmiranda) [reply grails-app/controllers/org/pih/warehouse/product/ProductController.groovy:1042]: @drodzewicz Perfect. That is how I was hoping you would describe the problem. So this just became more complicated and probably needs a tech huddle. 

For context, the reason we made that change was because the previous version just rendered an Errors object as JSON, which put all of the onus of translation on the client. 

image

At the time, we didn't care about the error message being localized because no one was translating those error messages. So we just grabbed the first one because it was convenient and we likely were just going to return the default error message anyway. 

But now that we are localizing error messages, we definitely should refactor to error handling in a more generic way like you did. You can keep your error handling for now (as a proof of concept) and we can add a ticket to migrate that solution to ErrorsController when we have time.

Comment (jmiranda) [reply grails-app/controllers/org/pih/warehouse/product/ProductController.groovy:1042]: By the way, multipleError should just be "unique" to avoid confusion.

Change
```
synonym.locale.validator.multipleError
```
to something like 
```
synonym.locale.unique.message
```
This doesn't really capture the issue, so perhaps there's a better code like
```
synonym.displayName.unique.message
```

@awalkowiak #3849

jmiranda (changes requested):
See comment(s)

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: We need to add a new ticket to refactor ErrorsController to use g.message(error:error). 

But we need to take into consideration the issue we encountered with the MessageSource throwing an NPE if one of the args is a Date. See Line 25 of MessageService.

I was trying to figure out why we went down the path we did and I think it was because I glossed over the fact that we could pass an error object to g.message. I think I had looked at how renderErrors was working and came to the conclusion that we needed to roll our own solution. 

So thank you @drodzewicz for finding that. That'll make error handling a lot cleaner.
https://docs.grails.org/3.1.1/ref/Tags/message.html

drodzewicz:
> See comment(s)

There are so many comment under this PR that I am getting lost in here.
If your only request was to change the label name to `synonym.displayName.unique.message` then I have pushed the code already for that.
Thanks for the review!

jmiranda:
Yeah, I just saw the change. Thanks. 

I was also referring to the minor issues that @kchelstowski pointed out related to null safety. But I'm fine merging if you don't think those are important.

jmiranda (approved):

PR #3859, opened 2023-02-20
OBPIH-5428 Items are duplicated on product list when they have synonyms added 

alannadolny:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -335,6 +336,7 @@ class ProductService {
                             eq("synonymTypeCode", SynonymTypeCode.DISPLAY_NAME)
                         }
                     }
+                    setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)
```
jmiranda, line 339: I'm interested to hear more about how testing (re: pagination) went. And while it doesn't matter, I usually put this at the top of the query closure to make it clear that we're doing sneaky shit.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -11,6 +11,7 @@ package org.pih.warehouse.product
 
 import grails.validation.ValidationException
 import groovy.xml.Namespace
+import org.hibernate.Criteria
```
jmiranda, line 14: This can probably be removed.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
kchelstowski: I was also afraid of this, so I'm gonna test this one locally today

alannadolny: Kacper showed me when the pagination doesn't work (I thought about another case so i didn't spotted this one). I am converting this pr to draft and I'm continuing working on it.

kchelstowski (commented):
As supposed - the pagination didn't work correctly. Since we've always struggled with such pagination/distinct, I decided to help Alan and play with it a bit. 
The issue with `setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY)` is that it returns distinct products, but it includes the duplicates in the `totalCount`.

The only solution I figured out for now could be to make (unfortunately) another, almost the same query, but to return only the `totalCount` of distinct products and then, override the `totalCount` returned from the main query with the `countProducts`. The reason I decided to override it, instead of returning e.g. a Map containing `[results: products, totalCount: countProducts]` is that I didn't want to cause any regression, as `getProducts()` is used in many places.

I know it might not be the best solution, but it might be a good start for Alan to keep working on this ticket or to test this solution more deeply.

As we might've expected some performance issues by doing two queries instead of one, I measured the avg time with/without my change out of 5 tries:

Without my change: avg is **76,8 ms**
With my change: avg is **119 ms**

grails-app/services/org/pih/warehouse/product/ProductService.groovy
jmiranda: Thanks @kchelstowski. 

@alannadolny :clap: Let us know if you need some guidance. I'll need to do some research to figure out the right approach but we can work on this together if you hit a wall.

alannadolny:
I tried to simplify Kacper solution but I noticed that I removed some necessary part of code, so I reverted my changes.

jmiranda:
@alannadolny I removed the duplicate query and I'm not seeing the pagination issue. Can you provide steps to reproduce?

jmiranda:
@alannadolny @kchelstowski @awalkowiak Finally got something working. Please take a look and let me know if I missed anything.
https://watch.screencastify.com/v/W5If8Kt3Qzx3czgsvtoB

jmiranda (approved):

kchelstowski (approved):
@jmiranda I'm not sure what you mean by "I removed the duplicate query", because I can see that you did the same what I, but in a more cleaner way (but still two queries are being run). 
If you just meant, that you removed the duplicate **closure**, but the two queries are still needed and you agree with my solution, then LGTM and nice job of merging this closure to one :) 

I tested this locally again with the data I created for this yesterday, and it is still working fine after your changes :)

jmiranda:
> If you just meant, that you removed the duplicate **closure**, but the two queries are still needed and you agree with my solution, then LGTM and nice job of merging this closure to one :) 

Yes I meant query definition. As far as I know there's no way to get rid of the two query executions. Even Grails does two query executions every time we have a paginated criteria query. So in fact I believe there are three query executions here. I'll include the query log when I get back to my office later.

PR #3862, opened 2023-02-22
OBGM-397 error message when opening add items to po sm dialog

drodzewicz:


drodzewicz:
There were 3 separate issues in the scope of this ticket.
1. extracting payload data from `request.JSON` instead of `params` like we do in grails 1
2.  Column type for _date_ordered_ on **Order** table was of datatype _varchar(255)_ which caused a bit of trouble when querying the Order data. Basically since _date_ordered_ was of type _varchar(255)_ it had some records in a wrong timestamp format like:
 _2020-04-23 00:00:00.0000000_ which I assume hibernate does not approve of and when trying to parse it as a Date throws an Exception 
```
java.sql.SQLDataException: Cannot convert string '2020-04-23 00:00:00.0000000' to java.sql.Timestamp value
```
changing the column datat type to datetime fixed this issue.
3. Apparently GORM throws an error when trying to use `*.findBy*inList()` with an empty or nullish list passed as argument.
From what I can tell this was a known issue in the grails community and based on the discussions that I have found surrounding this issue, it seems it has been fixed in  [grails ver. 6](https://github.com/grails/grails-data-mapping/pull/1220)

jmiranda:
I started to review this but need more time.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -60,21 +60,21 @@ class CombinedShipmentItemApiController {
 
     def findOrderItems() {
         List<Order> orders
-        if (params.orderIds) {
-            orders = Order.findAllByIdInList(params.orderIds)
+        if (request.JSON.orderIds) {
```
jmiranda, line 63: Larger topic that needs to be added to our tech debt / tech huddle topics would be how we want to handle params <=> request.JSON in a backwards compatible way.

https://github.com/openboxes/openboxes/labels/tech-huddle.

awalkowiak (approved):

PR #3863, opened 2023-02-22
OBPIH-5148 Can't see full item description on add items to PO shipment page

alannadolny:
I added a props `forceTooltip`, because if i just set `showValueTooltip=true` the tooltip doesn't work. When i changed the condition in `isTooltipDisabled()` then i found some places when two tooltips were displayed (because of using Tooltip inside `FIELDS`)

### DIFF
--- a/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
@@ -105,6 +105,7 @@ const FIELDS = {
         flexWidth: '4',
         required: true,
         attributes: {
+          forceTooltip: true,
           disabled: true,
         },
       },
--- a/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
@@ -105,7 +105,7 @@ const FIELDS = {
         flexWidth: '4',
         required: true,
         attributes: {
-          forceTooltip: true,
+          showValueTooltip: true,
           disabled: true,
         },
       },
--- a/src/js/utils/Select.jsx
+++ b/src/js/utils/Select.jsx
@@ -262,11 +262,13 @@ class Select extends Component {
     };
 
     const isTooltipDisabled = () => {
-      if (this.props.value?.translatedName) {
+      const { forceTooltip, value: fieldValue } = this.props;
+
+      if (fieldValue?.translatedName || forceTooltip) {
         return false;
       }
 
-      return !showLabelTooltip || (showValueTooltip && this.props.value);
+      return !showLabelTooltip || (showValueTooltip && !fieldValue);
     };
 
     return (
@@ -398,9 +400,11 @@ Select.propTypes = {
   classNamePrefix: PropTypes.string,
   translate: PropTypes.func.isRequired,
   defaultPlaceholder: PropTypes.string,
+  forceTooltip: PropTypes.bool,
 };
 
 Select.defaultProps = {
+  forceTooltip: undefined,
   value: undefined,
   onChange: null,
   onMenuClose: null,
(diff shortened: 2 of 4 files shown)


kchelstowski (changes requested):

alannadolny:
I changed the contition to: `!(fieldValue?.translatedName || showLabelTooltip || (showValueTooltip && this.props.value))`
So now, if we have at least one of this properties set to true then the tooltip will be visible. Otherwise, if all of the properties are falsy then the tooltip won't be visible.

src/js/utils/Select.jsx
```diff
@@ -398,9 +398,11 @@ Select.propTypes = {
   classNamePrefix: PropTypes.string,
   translate: PropTypes.func.isRequired,
   defaultPlaceholder: PropTypes.string,
+  forceTooltip: PropTypes.bool,
```
awalkowiak, line 401: I think you are not using this anymore

src/js/utils/Select.jsx
```diff
@@ -398,9 +398,11 @@ Select.propTypes = {
...
 };
 
 Select.defaultProps = {
+  forceTooltip: undefined,
```
awalkowiak, line 405: I think you are not using this anymore

src/js/utils/Select.jsx
```diff
@@ -262,11 +262,11 @@ class Select extends Component {
     };
 
     const isTooltipDisabled = () => {
-      if (this.props.value?.translatedName) {
-        return false;
-      }
+      const { value: fieldValue } = this.props;
 
-      return !showLabelTooltip || (showValueTooltip && this.props.value);
+      return !(
```
awalkowiak, line 267: how about:
```
if (this.props.value?.translatedName || showLabelTooltip) {
    return false;
}

return !(showValueTooltip && this.props.value);
```

awalkowiak (changes requested):

kchelstowski (changes requested):
LGTM, but could you please add comments what is going on here, if anyone ever faces this function?

src/js/utils/Select.jsx
```diff
@@ -107,7 +107,7 @@ class Select extends Component {
       multi, placeholder, showLabelTooltip, value, defaultPlaceholder,
     } = this.props;
 
-    if (value?.translatedName) {
+    if (value?.translatedName || value?.hasTranslatedName) {
```
kchelstowski, line 110: Could you show an exact example that this change is needed? I think we should rather stick to only `value?.hasTranslatedName`. 
Since `hasTranslatedName` is defined as this:
```
hasTranslatedName: !!obj.translatedName,
```
there should be no way that the `hasTranslatedName` is `true` and the `value.translatedName` is false, because if the `value.translatedName` would be `false`, the `hasTranslatedName` should also be `false`.

src/js/utils/Select.jsx
```diff
@@ -261,12 +261,21 @@ class Select extends Component {
       return null;
     };
 
+
+    /* We would like to see the tooltip when an item
+      have translatedName or when the showLabelTooltip
```
kchelstowski, line 266: `have` => `has`

kchelstowski (changes requested):

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -105,6 +105,7 @@ const FIELDS = {
         flexWidth: '4',
         required: true,
         attributes: {
+          forceTooltip: true,
```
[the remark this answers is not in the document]
jmiranda: I love the overall sentiment. No entendií el resto.

src/js/utils/Select.jsx
jmiranda: Agreed. I'm also thinking we might benefit from having these convenience properties (hasDisplayName, hasBrandNames, hasAlternateNames, etc) returned by the API but I think that needs a bit more thinking.

awalkowiak (approved):

PR #3864, opened 2023-02-23
OBGM-396 Unable to create PO

kchelstowski:
There were a few things, because fixing one of the errors, I got to another one :D 
Generally the error I was getting was different than the error in the ticket - I was getting an error that 
```
groovy.lang.MissingMethodException
No signature of method: org.pih.warehouse.order.Order.publishEvent() is applicable for argument types: 
```
and this is because in Grails 1.3.9 we have access to event publisher "globally" - that means it is injected automatically for each bean, but in Grails 3 we have to inject it ourselves with `@Autowired` annotation.

Then I was getting an error, that `addItems` method is not recognized - this was because in Grails 3 the methods in the redirects have to be as string.
And then finally the error was that `unexpected token ||` - it's weird, because we have the same line working on Grails 1.3.9 - after wrapping the statements into `( )` it suddenly works fine.

I was wondering about `webapp/webpack` directory being created when running the project:
Screenshot from 2023-02-23 08-35-15

I assumed it should be removed before I push the code, but it is not in `.gitignore` - should we consider adding it? cc @mdpearson 

cc @awalkowiak @jmiranda @drodzewicz

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
@@ -86,7 +86,7 @@ class PurchaseOrderController {
             return
         }
 
-        redirect(action: addItems, id: order.id)
+        redirect(action: "addItems", id: order.id)
     }
 
 
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -18,8 +18,13 @@ import org.pih.warehouse.invoice.InvoiceTypeCode
 import org.pih.warehouse.picklist.Picklist
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentStatusCode
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.ApplicationEventPublisher
 
 class Order implements Serializable {
+
+    @Autowired
+    ApplicationEventPublisher applicationEventPublisher
 
     def beforeInsert = {
         createdBy = AuthService.currentUser
@@ -31,13 +36,13 @@ class Order implements Serializable {
 
     def publishRefreshEvent = {
         if (isPurchaseOrder) {
-            publishEvent(new RefreshOrderSummaryEvent(this))
+            applicationEventPublisher?.publishEvent(new RefreshOrderSummaryEvent(this))
         }
     }
 
     def publishRefreshEvenBeforeDelete = {
         if (isPurchaseOrder) {
-            publishEvent(new RefreshOrderSummaryEvent(this, true))
+            applicationEventPublisher?.publishEvent(new RefreshOrderSummaryEvent(this, true))
         }
     }
 
@@ -122,7 +127,8 @@ class Order implements Serializable {
             "canceled",
             "completed",
             'activeOrderAdjustments',
-            'activeOrderItems'
+            'activeOrderItems',
+            'applicationEventPublisher'
     ]
 
     static hasMany = [
--- a/grails-app/views/purchaseOrder/_showOrderItems.gsp
+++ b/grails-app/views/purchaseOrder/_showOrderItems.gsp
@@ -132,8 +132,8 @@
                     </g:form>
                 </div>
                 <div id="add-adjustment" class="box">
-                    <g:set var="canManageAdjustments" value="${order?.status >= OrderStatus.PLACED && isApprover
-                            || order?.status == OrderStatus.PENDING}"/>
+                    <g:set var="canManageAdjustments" value="${(order?.status >= OrderStatus.PLACED) && (isApprover
+                            || order?.status == OrderStatus.PENDING)}"/>
                     <h2 style="display: flex; align-items: center; justify-content: space-between;">
                         <warehouse:message code="default.add.label" args="[g.message(code: 'orderAdjustment.label')]"/>
                     </h2>


grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
```diff
@@ -86,7 +86,7 @@ class PurchaseOrderController {
             return
         }
 
-        redirect(action: addItems, id: order.id)
+        redirect(action: "addItems", id: order.id)
```
jmiranda, line 89: Wow I had to look this up. I know in Grails 1.3.9 you can pass a closure to "action" but I had never seen it done and wasn't sure of the 

> action - The action to redirect to, either a string name or a reference to an action within the current controller

https://grails.github.io/grails2-doc/1.3.9/ref/Controllers/redirect.html

So in grails 3.3.10, where actions are no longer controllers (but methods) ... I suspect this isn't ok any longer. The docs for the redirect's action parameter suggest it is no longer ok.

> action (optional) - the name of the action to use in the link, if not specified the default action will be linked

But the full documentation implies that it is still ok. 
https://docs.grails.org/3.3.10/guide/single.html#_chaining

jmiranda: So @kchelstowski did you change this because it caused an issue or because it didn't look right.

If we change it here we should probably consider changing it in develop as well. 

@awalkowiak Thoughts?

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -31,13 +36,13 @@ class Order implements Serializable {
 
     def publishRefreshEvent = {
         if (isPurchaseOrder) {
-            publishEvent(new RefreshOrderSummaryEvent(this))
+            applicationEventPublisher?.publishEvent(new RefreshOrderSummaryEvent(this))
```
jmiranda, line 39: Marking this ticket for https://github.com/openboxes/openboxes/labels/tech-huddle.

jmiranda: I want to make sure this is the right approach and then backport this to develop.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -122,7 +127,8 @@ class Order implements Serializable {
             "canceled",
             "completed",
             'activeOrderAdjustments',
-            'activeOrderItems'
+            'activeOrderItems',
+            'applicationEventPublisher'
```
jmiranda, line 131: Are you getting a warning about this? I don't know if we want to mark this as transient because it's actually a property that has an implicit getter and setter. Most transients are derived properties that only have a getter.

grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
jmiranda: @kchelstowski Ignore, just reread the summary and saw you addressed this.

grails-app/domain/org/pih/warehouse/order/Order.groovy
jmiranda: To clarify what I mean, I think the autowired approach is fine and probably the right approach here but I just wanted to point out that the reason "publishEvent" works on its own is not standard injection but rather groovy magic.

For every controller/domain/service class, the spring events plugins adds the publishEvent as a closure.
https://github.com/robfletcher/grails-spring-events/blob/master/SpringEventsGrailsPlugin.groovy#L107

And the publishEvent closure invokes the publishEvent method on the asynEventPublisher.  
https://github.com/robfletcher/grails-spring-events/blob/master/SpringEventsGrailsPlugin.groovy#L113

We could do something similar here just to make it work out of the box and not require a ton of new service injections. We'd just need to figure out where that magic goes in a Grails 3 application (Bootstrap.groovy). 

I don't know exactly why this magic is no longer happening in Grails 3. Are we no longer using the spring events plugin? Are we using the standard Spring Events API? Anyway, I'm open to suggestions on the right approach. 

And for the record, in Grails 3 (post-migration) and beyond we will be looking into other technologies for asynchronous programming (https://async.grails.org/latest/guide/index.html) and event publishing (https://async.grails.org/latest/guide/index.html#events), namely 

* Reactor 
* RxJava 
* GPars  

or a message queue approach using 

* RabbitMQ https://guides.grails.org/grails-rabbitmq/guide/index.html
* Redis https://github.com/grails/grails-redis
* Kafka https://guides.grails.org/grails-micronaut-kafka/guide/index.html

You can read more about that here  https://docs.grails.org/latest/guide/async.html

Exciting times lie ahead.

grails-app/controllers/org/pih/warehouse/order/PurchaseOrderController.groovy
kchelstowski: @jmiranda wanted to make screenshots for you from the IntelliJ, but it doesn't want to capture this, so:
I changed it, because as I wrote - in Grails 3 it is supposed to be as string. If it is not as string, even IDE has it in gray and says: `No candidates found for method call addItems.` - which doesn't happen in Grails 1.3.9 - it recognizes the method  both when being in string and not.

grails-app/domain/org/pih/warehouse/order/Order.groovy
kchelstowski: from what I've read - in Grails 1.3.9 `publishEvent` was a member of `GrailsApplication` object and you could refer to it in any controller/service/domain and in Grails 3 it got updated to use `ApplicationEventPublisher` which is a spring bean from spring context, so the only way I see it working is to inject it into the domain using `@Autowired`.

Had some part of Redis at the university and also saw some usage of Reactor, so I'm also excited if we ever go this way :)

grails-app/domain/org/pih/warehouse/order/Order.groovy
kchelstowski: Exactly - when I hadn't marked it as transient, it wanted to be persisted to DB and the application didn't want to run.

jmiranda (approved):
This is ready to go.

PR #3865, opened 2023-02-23
OBS-1412 Do not call toBoolean() on an uninitialized Config object

mdpearson:


jmiranda (commented):
The problem with this solution is that the type of enabled is different based on which config file it is retrieved from (openboxes-config.properties returns a String, openboxes-config.groovy returns a Boolean). AFAIK there's nothing we can do about this so the best solution was to explicitly coerce the value to Boolean. 

What I assume we're seeing is that our Ansible playbook is not explicitly setting grails.mail.enabled in either openboxes-config.properties or openboxes-config.groovy

So I think the solution should probably be 

* Be more defensive and add null safety
       
        grailsApplication.config.grails.mail.enabled ? 
            grailsApplication.config.mail.enabled.toBoolean : false

* Add a default value to Config.groovy
         
        grails.mail.enabled = false

mdpearson:
@jmiranda this issue can occur if either of the following happens:

1. `grails.mail.enabled` is not set;
2. `grails.mail` is not configured at all.

In either case you get an object that doesn't implement `toBoolean()` and you get this error.

The _real_ reason this happened is over at https://github.com/openboxes/openboxes-devops/pull/14; this PR just makes our code a bit more defensive.

jmiranda:
> In either case you get an object that doesn't implement toBoolean() and you get this error.

@mdpearson I was wrong about my initial premise that grails.mail.enabled was null. I just forgot that grails.mail.enabled is in DSL form in Config.groovy. So you're right ... those conditions would cause this but there's almost no way (unless someone did something like this in openboxes-config.groovy)
```
grails.mail = {}
```

> The real reason this happened is over at https://github.com/openboxes/openboxes-devops/pull/14; this PR just makes our code a bit more defensive.

Yes, that's what I was thinking. Thank you for the explanation.

jmiranda (commented):
Ok, so I just noticed that I was wrong about Config.groovy not setting a default value for `grails.mail.enabled`. Obviously, that was incorrect I just had trouble finding it because it was defined in the form of a DSL.
```
// Default mail settings
grails {
    mail {
        enabled = false
        ...
    }
}
```
So I'm confused why this issue even came up, but assume this is probably on the Ansible side (perhaps Ansible is creating config files that are clobbering the grails or grails.mail DSL)

jmiranda:
Sorry for the disjointedness of these two comments. The first one was written AFTER the second one. I just hadn't posted one before the other and then was editing them both at the same time. 

I'm still confused why this line would have caused the issue

https://github.com/openboxes/openboxes-devops/pull/14/files#diff-5ea833942977d0cb6d75cad91d7d73eac60513e249f2e54c22b9df41a86b6ac3R26

Is this template initializing the value to null (essentially overriding Config.groovy)

mdpearson:
> Sorry for the disjointedness of these two comments. The first one was written AFTER the second one. I just hadn't posted one before the other and then was editing them both at the same time.
> 
> I'm still confused why this line would have caused the issue
> 
> https://github.com/openboxes/openboxes-devops/pull/14/files#diff-5ea833942977d0cb6d75cad91d7d73eac60513e249f2e54c22b9df41a86b6ac3R26
> 
> Is this template initializing the value to null (essentially overriding Config.groovy)

Before that PR went in, the line in the devops repo evaluated to either `grails.mail.enabled = False` or `grails.mail.enabled = True` (it's a Jinja2 template, and Jinja is Python-based, and Python capitalizes booleans).

If you try to evaluate `grails.mail.enabled = False` in Groovy, you'll get `groovy.lang.MissingPropertyException: No such property: False` -- note that False is a symbol, not a string.

When an error like that happens in a config file, the enclosing stanza seems to get clobbered. Given that this PR is still open, and the thing that causes it is resolved over in the devops repo, I'm going to close this out.

PR #3866, opened 2023-02-24
OBPIH-5455 Add filter for GL account

alannadolny:


grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -287,4 +289,12 @@ class ProductApiController extends BaseDomainApiController {
...
         render([data: tags] as JSON)
     }
+
+    def glAccountOptions = {
+        def glAccounts = GlAccount.list().collect {
+            [id: it.id, label: "${it.code} - ${it.name}"]
+        }
+
+        render([data: glAccounts] as JSON)
+    }
```
kchelstowski, line 299: this should be in the `GlAccountController`, not in the `ProductApiController` and would be good to add this fetching stuff to the service and here leave:
```
def glAccounts = glAccountService.getGlAccountsOptions()
...
```

src/js/utils/option-utils.jsx
```diff
@@ -218,3 +218,8 @@ export const fetchProductsTags = async () => {
   const response = await apiClient.get('/openboxes/api/tagOptions');
   return response.data.data;
 };
+
+export const fetchProductsGlAccounts = async () => {
+  const response = await apiClient.get('/openboxes/api/glAccountOptions');
+  return response.data.data;
+};
```
kchelstowski, line 225: maybe we could follow, since you've refactored some of our api calls to use `api/services`, to use this approach here?

kchelstowski (changes requested):

src/js/api/urls.js
```diff
@@ -27,3 +27,6 @@ export const STOCKLIST_CLEAR = id => `${STOCKLIST_API}/${id}/clear`;
 export const STOCKLIST_CLONE = id => `${STOCKLIST_API}/${id}/clone`;
 export const STOCKLIST_PUBLISH = id => `${STOCKLIST_API}/${id}/publish`;
 export const STOCKLIST_UNPUBLISH = id => `${STOCKLIST_API}/${id}/unpublish`;
+
+// GL ACCOUNTS
+export const GL_ACCOUNTS_OPTION = '/openboxes/api/glAccountOptions';
```
kchelstowski, line 32: having an occassion - could you please specify at the top of the file 
```
const API = '/openboxes/api'
```
and use it in this endpoint?
You don't have to change it the lines above, but just to have less refactor in the future and use this constant since now

kchelstowski (changes requested):

src/js/utils/option-utils.jsx
```diff
@@ -218,3 +219,8 @@ export const fetchProductsTags = async () => {
   const response = await apiClient.get('/openboxes/api/tagOptions');
   return response.data.data;
 };
+
+export const fetchProductsGlAccounts = async () => {
+  const { data } = await glAccountApi.getGlAccountOptions();
+  return data.data;
+};
```
kchelstowski, line 226: I'm adding one additional comment, which should be rather answered by @awalkowiak - I'm wondering if we should wrap it inside `try/catch` and have somewhat user-friendly error message if any happens, or let the handler handle that, but we could have some "ugly" error message in terms of user perspective?

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -282,10 +282,11 @@ class ProductService {
      * @param category
      * @param catalogs
      * @param tags
+     * @param glAccounts
      * @param params
      * @return
      */
-    List<Product> getProducts(List<Category> categories, List<ProductCatalog> catalogsInput, List<Tag> tagsInput, boolean includeInactive, Map params) {
+    List<Product> getProducts(List<Category> categories, List<ProductCatalog> catalogsInput, List<Tag> tagsInput, List<GlAccount> glAccounts, boolean includeInactive, Map params) {
```
awalkowiak, line 289: I am starting to dislike these parameters more with each new one. I think we could think about refactoring this one.

awalkowiak (changes requested):

grails-app/conf/UrlMappings.groovy
```diff
@@ -65,6 +65,12 @@ class UrlMappings {
             action = [GET: "tagOptions"]
         }
 
+        // Gl account options for filters on product list page
+        "/api/glAccountOptions"(parseRequest: true) {
+            controller = { "glAccount" }
```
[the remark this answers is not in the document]
jmiranda: @awalkowiak Why weren't these just added as straight-up REST endpoints? 
```
/api/categories
/api/catalogs
/api/tags
/api/glAccounts
```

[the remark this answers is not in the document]
jmiranda: As a compromise maybe we could come up with a new endpoint API controller (i.e. SelectOptionsApiController) to have a single place to implement all of these global-y things.

jmiranda (changes requested):
Need to create a new API for the options endpoints.

src/js/utils/option-utils.jsx
jmiranda: Discussion for standardizing error messages #3849

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -0,0 +1,38 @@
...
+
+import grails.converters.JSON
+import org.pih.warehouse.core.SelectOptionsService
+
+class SelectOptionsApiController {
+
+    SelectOptionsService selectOptionsService;
+
+    def glAccountOptions = {
+        def glAccounts = selectOptionsService.getGlAccountsOptions()
```
jmiranda, line 20: @awalkowiak What do you think about using the generic API service here?

```
def glAccountOptions = { 
    def glAccounts = genericApiService.getList(GlAccount.class, [:]).collect {
            [id: it.id, label: "${it.code} - ${it.name}"]
        }
    }
    render([data: glAccounts] as JSON)
}
```

jmiranda (approved):
LGTM

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
awalkowiak: Yes, if we can get rid of adding one another redundant service, then this should be good. cc: @alannadolny

grails-app/services/org/pih/warehouse/api/GenericApiService.groovy
```diff
@@ -137,4 +141,27 @@ class GenericApiService {
         }
     }
 
+    def getGlAccountsOptions() {
```
awalkowiak, line 144: You should not put this here since in that way this is in no way generic. You should use the exact code that Justin suggested.

awalkowiak (changes requested):

awalkowiak (changes requested):

PR #3868, opened 2023-02-27
OBPIH-5355 Restyling components for notifications

kchelstowski:
I tried to make it as much configurable as I could - the usage is pretty simple and described in the `notification.md` file.

I made `notification()` as **curried** function, so it takes as an argument `type` of the notification from enum `NotificationType` and then returns a function which expects arguments: `message` (required), `details` (the description of the error, optional), `startIcon` (optional, each type of the notification has its default `startIcon`).

I also tried to cause as little regression as I could, so I also took care of already existing notifications (which are called by `Alert.success(...)` directly), and managed to make them look as close as to the new design:
Screenshot from 2023-02-27 11-24-19
Manon has said it looked good, and said, that we would have in the future some tickets to change the existing alerts and their messages.
I also refactored our error handler in the `apiClient` to make it use our new notification function. An example below:
Screenshot from 2023-02-27 13-46-59

Overall I've managed to come up with this:

Screenshot from 2023-02-27 13-40-45

@drodzewicz to create a notification for lost connection you just have to do this:
```
notification(NotificationType.INFO)({
      message: 'Lost connection',
      details: 'You are now offline. The changes you made will not be saved',
      startIcon: <RiWifiOffLine />,
});
```

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1300,6 +1300,9 @@ openboxes.expirationDate.format = Constants.EXPIRATION_DATE_FORMAT
 openboxes.display.date.format = Constants.DISPLAY_DATE_FORMAT
 openboxes.display.date.defaultValue = Constants.DISPLAY_DATE_DEFAULT_VALUE
 
+// Notifications configuration
+openboxes.notification.disappear.time.ms = 8000
+
 // Global megamenu configuration
 
 openboxes.menuSectionsUrlParts = [
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -1301,7 +1301,9 @@ openboxes.display.date.format = Constants.DISPLAY_DATE_FORMAT
 openboxes.display.date.defaultValue = Constants.DISPLAY_DATE_DEFAULT_VALUE
 
 // Notifications configuration
-openboxes.notification.disappear.time.ms = 8000
+
+// delay is in ms
+openboxes.client.notification.autohide.delay = 8000
 
 // Global megamenu configuration
 
--- a/src/js/components/dashboard/CustomAlert.jsx
+++ b/src/js/components/dashboard/CustomAlert.jsx
@@ -1,0 +1,65 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { RiAlertLine, RiCheckboxCircleLine, RiCloseLine, RiErrorWarningFill, RiInformationLine } from 'react-icons/ri';
+
+import NotificationType from 'consts/notificationTypes';
+
+const defaultStartIcon = {
+  [NotificationType.SUCCESS]: <RiCheckboxCircleLine />,
+  [NotificationType.INFO]: <RiInformationLine />,
+  [NotificationType.ERROR]: <RiErrorWarningFill />,
+  [NotificationType.WARNING]: <RiAlertLine />,
+};
+
+const CustomAlert = ({
+  message, classNames, customFields, id, styles, condition, handleClose,
+}) => {
+  // Show default start icon for already existing Alerts,
+  // which are not called by notification(), but Alert
+  const getStartIcon = () => {
+    if (customFields?.startIcon) {
+      return customFields.startIcon;
+    }
+    return defaultStartIcon[condition ?? NotificationType.INFO];
+  };
+
+  return (
+    <div className={`${classNames} ${!customFields?.details ? 'no-details' : ''}`} id={id} style={styles}>
+      <div className="s-alert-box-inner">
+        <div className="alert-start-icon">
+          {getStartIcon()}
+        </div>
+        <span className="alert-title">{message}</span>
+        {customFields?.details && <span className="alert-details">{customFields?.details}</span>}
+        <div className="alert-close-icon">
+          <span role="presentation" onClick={handleClose}>
+            <RiCloseLine />
+          </span>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+
+export default CustomAlert;
+
+
+CustomAlert.propTypes = {
+  classNames: PropTypes.string.isRequired,
+  /** Condition is a built-in prop in the Alert and the types of it can be:
+    success, error, warning, info.
+  */
+  condition: PropTypes.string.isRequired,
+  /** customFields can contain anything in the future - for now we use details and startIcon */
+  customFields: PropTypes.shape({
+    details: PropTypes.string.isRequired,
+    startIcon: PropTypes.element.isRequired,
+  }).isRequired,
+  handleClose: PropTypes.func.isRequired,
+  id: PropTypes.string.isRequired,
+  /** message (title) is the first param passed to the Alert.(..) */
+  message: PropTypes.string.isRequired,
+  styles: PropTypes.shape({}).isRequired,
+};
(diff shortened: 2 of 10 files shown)


grails-app/conf/Config.groovy
```diff
@@ -1300,6 +1300,9 @@ openboxes.expirationDate.format = Constants.EXPIRATION_DATE_FORMAT
 openboxes.display.date.format = Constants.DISPLAY_DATE_FORMAT
 openboxes.display.date.defaultValue = Constants.DISPLAY_DATE_DEFAULT_VALUE
 
+// Notifications configuration
+openboxes.notification.disappear.time.ms = 8000
```
jmiranda, line 1304: We have to be careful with the word "notification" as its overloaded and can be used to mean a lot of different things (email notifications, push notifications, etc). So let's try to be more specific here. 
```
openboxes.toast...
```
Or if we need to clarify the context ...
```
openboxes.react.toast...
```

And also `disappear` and `ms` are a bit confusing to me. Disappear is actually a fairly descriptive word, I just haven't seen it used in software before. It might just be me though so feel free to state your case with some examples.

We have a convention (for good or bad) in which we include the property description and unit of measure in the property name. Something like this

- autoCloseInMilliseconds 
- autoHideInMilliseconds
- disappearInMilliseconds

I've never loved that but it was something I came up with to avoid confusion. We could also just add a descriptive comment above the property to explain what it's used for and include the UoM. Or we could choose to always standardize on milliseconds. In any of those cases, the following might be better

- openboxes.toast.autoClose
- openboxes.toast.autoHide
- openboxes.toast.autoDisappear

And instead of "time.ms" I prefer a more descriptive word like

- duration
- delay
- timeout

One other thing that we might want to configure here is whether the "toast" is actually allowed to disappear after some period of time. We probably don't need a boolean config property for the API should support enabling / disabling the auto hide so we can keep some messages on screen until the user closes them.

```
openboxes.toast.autoHide.enabled = true
openboxes.toast.autoHide.delayInMilliseconds = 5000
```
or with a comment
```
// Enable autohide of toasts by default and delay autohide in 5 seconds
openboxes.toast.autohide.enabled = true
openboxes.toast.autohide.delay = 5000
```

FWIW I'm not saying we shouldn't use the `time.ms` I just haven't personally seen that myself so I would just like to see that used commonly out in the wild in order to justify its use.

In general, I usually go to a library that we're trying to mimic and find out what their API looks like before I start naming things. 

Bootstrap uses `delay` and `autohide`
https://getbootstrap.com/docs/4.2/components/toasts/#options

React Toast uses `autoClose`
https://fkhadra.github.io/react-toastify/introduction/

jQuery Toast uses `hideAfter`
https://github.com/kamranahmedse/jquery-toast-plugin

jmiranda (changes requested):
Should we discuss this on the next tech huddle or just discuss it here?

grails-app/conf/Config.groovy
jmiranda: Another example, VueJS uses just `duration` to signify how long the toast should be displayed, implying that all toasts auto-hide unless (I assume) they have `dismissible` set to true. 
https://www.npmjs.com/package/vue-toast-notification#user-content-available-options

Not sure if that's what we want for our toast API but it would possibly remove the need for the `enabled` config property discussed above. Plus I guess the decision to make a message "always visible, but closeable" vs "visible for a certain duration" is probably a developer concern, not something that a sysadmin should be in the business of overriding. That is, unless we set the auto hide duration too low I guess.

kchelstowski: 1. I don't feel like there is a need to keep it until tech huddle, since it is a blocker for Dariusz's ticket
2. I also didn't feel comfortable with the naming, but when naming it, I felt like I've forgotten whole English, if you know what I mean :laughing: 
3. I couldn't make it `alert()` as the library is named, because the `alert()` is already a reserved word and a built-in alert in the browser, so I discussed it with the guys and decided to go with `notification()` for now and if anyone suddenly has their own opinion going further, just point it out in the PR, so thanks for your compehensive comment about this one.
4. It's not my "invention" of having this auto hide - it was specified in the mockup that the notifications should hide after 8 seconds.
5. I will tell the guys to read your comment and we can agree on final naming for the `notification()` - as for "disappear" - I love the "duration" here, so thanks again for that research and for showing your process of dealing with such stuff.

jmiranda: Great points. I'll let @awalkowiak and others (including @mdpearson) ping pong this around for a few hours tomorrow. But I agree - if this is a blocker we can merge a version we can live with for now and then make a final decision later.

Also, I hate naming. The only way I survive naming exercises is by finding someone else who has named something in a way that I either like or (at least) don't hate as much as the names I come up with on my own.

src/js/components/dashboard/CustomAlert.jsx
```diff
@@ -0,0 +1,65 @@
...
+  [NotificationType.ERROR]: <RiErrorWarningFill />,
+  [NotificationType.WARNING]: <RiAlertLine />,
+};
+
+const CustomAlert = ({
+  message, classNames, customFields, id, styles, condition, handleClose,
+}) => {
+  // Show default start icon for already existing Alerts,
+  // which are not called by notification(), but Alert
+  const getStartIcon = () => {
```
drodzewicz, line 20: the **startIcon** naming kinda bothers me for some reason.
Since we are not dealing with multiple icons in this component I don't think it is necessary to specify the position in the prop name.
I think just an `icon` by it self should work well.

drodzewicz (changes requested):
Overall looks very good.
Good job @kchelstowski

src/js/utils/apiClient.jsx
```diff
@@ -55,9 +56,12 @@ export const handleSuccess = response => response;
 export const handleError = (error) => {
   switch (error.response.status) {
     case 400: {
-      const errorMessages = _.map(_.get(error, 'response.data.errorMessages', ''), errorMessage => `<div>${errorMessage}</div>`);
+      const errorMessages = _.map(_.get(error, 'response.data.errorMessages', ''), errorMessage => `${errorMessage}, `);
```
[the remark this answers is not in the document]
kchelstowski: you are right it could be moved out - I haven't gone into the details here, because anyway we'll have to revisit it probably while having tickets for refactoring existing alerts, but thanks for pointing that out

src/js/components/dashboard/CustomAlert.jsx
kchelstowski: you are right - I just thought about it like the close icon is "endIcon", but you are right it's unchangeable at the moment and we could stick to the `icon` if it's confusing for you

grails-app/conf/Config.groovy
drodzewicz: My naming suggestion is
```
openboxes.client.notification.autohide.delay.milliseconds = 5000
```
At first I thought that we could specify the unit of measure in the comment above but I think that it would be better to explicitly have the units in the variable name to avoid any potential confusion.
That way a developer doesn't have dig through the config file looking for comments wondering if it is seconds or milliseconds.

Also I think that we should stick with name `notification` instead of `toast` since everyone has been calling it that way and I specify a prefix that it is a **client** notification

awalkowiak: @jmiranda @kchelstowski I like the Dariusz suggestion with `openboxes.client.notification.(...)` because I don't like using the keyword `react`.

awalkowiak (approved):

PR #3869, opened 2023-02-28
OBPIH-5320 create listener for connection status

drodzewicz:


drodzewicz:
Initially I wanted to create a `useConnectionListener` hook which made more sens, but since all of our pages are written as class components I couldn't use hooks and decided on implementing this listener logic in a Higher Order Component.
If you have any better suggestions where to implement this listener, then please let me know.

kchelstowski (commented):
Before I start reviewing the code, could you explain why it has to be class component? Even when using this HOC, I think the `ConnectionListenerHOC` doesn't have to be a class component.

To use hook I think you could just place it in the `Router` since `MainRouter` renders only `<Router>` and place it here and that should be enough.

Another preposition could be in the `index.jsx` to wrap the `<MainRouter>` inside the `<ConnectionListener>`:
```
<ConnectionListener>
  <MainRouter />
</ConnectionListener>
```
and to render `children` inside the `<ConnectionListener>`
You could even then play around with `useContext` if you wanted to use the values of `<ConnectionListener>` component inside the `<MainRouter>` or wherever.

Tell me your thoughts and lets discuss if further.

Anyway nice job of visualization

drodzewicz:
> To use hook I think you could just place it in the Router since MainRouter renders only <Router> and place it here and that > > should be enough.

I feel I bit stupid right now that I have not noticed that `Router.jsx` is a functional component...
Thanks @kchelstowski for pointing that out. Fortunately it is not going to be a hard task to move the logic to the hook from a HOC, so give me a moment to do that.

kchelstowski (approved):

grails-app/conf/Config.groovy
```diff
@@ -1081,6 +1081,9 @@ grails.plugin.raven.dsn = "https://{PUBLIC_KEY}:{SECRET_KEY}@app.getsentry.com/{
 // Default Ajax request timeout
 openboxes.ajaxRequest.timeout = 120000
 
+// connection timeout milliseconds
+openboxes.connection.timeout = 8000
```
jmiranda, line 1085: Can you explain the rationale behind the 8 second timeout? Rather than something round like 5 or 10 seconds?

Also let's be a little more explicit about what this property is for. "connection" is an overloaded term - in the context of our application it could mean database, URL, and someday maybe websocket, etc

So let's go with something more descriptive.
```
// Timeout used with the navigator.onLine event listener
openboxes.browser.connection.status.timeout = 8000
```
Or if that's still too vague, we really only care about this from the perspective that it tells us whether the application is online or offline (whichever makes more sense)

* openboxes.online.status.listener.timeout
* openboxes.offline.status.listener.timeout
* openboxes.browser.online.listener.timeout

Or we could map it to the specific API we're using. I don't love this since the API could change in the future, but it's slightly better than "openboxes.connection.timeout"

* openboxes.navigator.online.timeout

jmiranda (changes requested):
> but since we are dealing with the online state of the browser which works like a socket that immediately informs us of the change in online status exponential backoff seems unnecessary.

To clarify, I wasn't saying that we'd use exponential backoff to retry the connection. It was more of a strategy to prevent a steady stream of events to the user i.e. online-offline-online-offline-online-offline

offlineonline

> when online state is changed, do not immediately inform the user that the connection is lost/restored

And I actually think we do want to change the status to offline as soon as we get the offline status event in order to prevent the user from taking an action that will be detrimental to their current session. What we don't necessarily want to do is always respond to online events as quickly as offline. But since we've never seen this in action I thought it would be good to first see how the listener behaves in responding to all events. And this is why I was asking whether we could initially set the timeout to 0 on whatever server we use for testing and have that give us an experience that was more fine-grained.

Once we've seen that behavior we can adapt to the flood of events (that occur in an unstable network situation) by incorporating something like "exponential backoff" to retry the status. Since we're receiving all of the status events, the "retry" will be one in which we are testing whether the connection is stable. In other words, we'll keep track of all online/offline events that occur over a duration (the timeout value) and if there < X offline events in that period of time then we can say that the browser online status is stable. If not, we retry less often until it is. Similar to Gmail, we'd probably have a "Check your connection and try again" message with a way for the user to manually retry.

But I see the online/offline event listener as being something that contributes to the state that helps make a decision about whether our connection is stable. It's not actually determining that itself.

grails-app/conf/Config.groovy
drodzewicz: You are right, looking at it again I should have picked a better name for it, I was inspired by the variable name above for `ajaxRequest.timeout`.

`openboxes.browser.connection.status.timeout` sounds good

As for the reasoning behind _8000 ms_ for timeout I don't have a strong opinion about it and we can discuss what is the most optimal value.

jmiranda (approved):
LGTM; might need to consider shorter thresholds and add a feature flag

jmiranda:
p.s. I added the https://github.com/openboxes/openboxes/labels/tech-huddle label to indicate that this was discussed in a tech huddle, not that it needs another one.

awalkowiak (approved):

PR #3870, opened 2023-02-28
OBPIH-5446 Add product_family_id column to Product

kchelstowski:


kchelstowski:
I've added some minor fixes - added a validation not to allow assigning more than one product family for product - GORM doesn't allow it in this way, that it doesn't add the duplicate when triggering `addToSiblings`, it just replaces the product with itself (or even not), but I wanted to add some user-friendly info if a user tries to do it and doesn't see anything happening.

Also an existing problem in the old feature was that when mapping the products that are in this `productGroup`, if a `productGroup` had 4 products, we were not displaying those 4 products, but 4 times we displayed the main product.

Except some minor fixes, everything seems to be working fine 

Thanks for comprehensive research about what I've done with `mappedBy` @jmiranda and any support I've had from you in this ticket @jmiranda @awalkowiak

grails-app/controllers/org/pih/warehouse/product/ProductGroupController.groovy
```diff
@@ -239,34 +234,38 @@ class ProductGroupController {
...
+                product.removeFromProductGroups(productGroup)
+                productGroup.removeFromProducts(product)
+            }
             product.save(flush: true)
         } else {
             response.status = 404
         }
 
-        render(template: 'products', model: [productGroup: productGroup, products: productGroup.products])
+        render(template: 'products', model: [productGroup: productGroup, products: isProductFamily?productGroup?.siblings:productGroup.products])
```
awalkowiak, line 268: Please add spaces around `?` and `:`. For a brief moment, I thought that there should be a dot after `?` and could not understand why.

grails-app/domain/org/pih/warehouse/product/ProductGroup.groovy
```diff
@@ -24,15 +24,14 @@ class ProductGroup implements Comparable, Serializable {
     Date dateCreated
     Date lastUpdated
 
-    static belongsTo = Product
-    static hasMany = [products: Product]
+    static hasMany = [products: Product, siblings: Product]
+    static mappedBy = [products: "productGroups", siblings: 'productFamily']
```
awalkowiak, line 28: Please unify quotes.

grails-app/services/org/pih/warehouse/product/ProductGroupService.groovy
```diff
@@ -0,0 +1,28 @@
...
+
+    def grailsApplication
+
+    def addProductToProductGroup(String productGroupId, String productId, boolean isProductFamily) {
+        def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+        ProductGroup productGroup = ProductGroup.get(productGroupId)
+        Product product = Product.get(productId)
+        if (productGroup && product) {
+            if (product.productFamily) {
+                throw new IllegalArgumentException(g.message(code: "productGroup.productFamily.duplicate.error", default: "This product already has a product family"))
```
awalkowiak, line 22: You will throw this even if the `productGroup` is not `productFamily` type, but product already has the product family assigned.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/product/ProductGroupService.groovy
kchelstowski: good point, thanks for catching that

kchelstowski:
Additionally to the fixes after review I fixed one bug related to tabs/cookies - sometimes after refreshing the page, when it was getting straight into Products/Siblings tab, it didn't have the `isProductFamily` param with the correct value -> on Siblings tab it should have `true`, but when adding a product to the Siblings, it was suddenly `false`.
I removed the jQuery cookie from `.tabs`, so after refresh we are always at Product Group tab and then after going to Products/Siblings tabs, the `isProductFamily` param has correct value then.

cc @jmiranda @awalkowiak

awalkowiak (approved):

PR #3871, opened 2023-02-28
OBPIH-5323 Save line by line in outbound add items step

alannadolny:


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -676,12 +695,33 @@ class AddItemsPage extends Component {
...
               ...val,
+              disabled: true,
               product: {
                 ...val.product,
                 label: `${val.productCode} ${val.product.translatedName ?? val.product.name}`,
               },
             }),
           );
+          // In autosave, we don't modify the state, because
+          // lines which not passed the validation will be
```
kchelstowski, line 706: nitpicky, but -> `which not passed` -> `which have not passed`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -676,12 +695,33 @@ class AddItemsPage extends Component {
...
+          // deleted during users work
+          if (!withStateChange) {
+            // We want to disable saved line, so I am looking for product with the same
+            // code and quantity higher than 0 in response
+            // (to avoid disabling new line with the same productCode)
+            const savedItemsProductCodes = lineItemsBackendData.map(item => item.productCode);
+            const lineItemsAfterSave = this.state.values.lineItems.map((item) => {
+              if (
+                _.includes(savedItemsProductCodes, item.product.productCode)
+                && parseInt(item.quantityRequested, 10)
```
kchelstowski, line 716: I assume we don't expect negative quantity here (it would evalute to `true` then)? If so, I'm ok with that. Probably we don't, since it's data from backend.

kchelstowski (changes requested):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: I am not sure what do you mean by data from backend. Of course, products are coming from backend, but this lineItems are new added lines.

It's doesn't matter if this is negative or positive. This is not data from backend, it is new added line. This is a check if there is quantity entered, because if quantity is empty or zero and we enter quantity to product with the same productCode both lines will be disabled, but we want to disable only line with entered quantity.

kchelstowski: My point was to understand what will happen if the quantity is negative - negative qty evalutes to true and my point was to understand if we will ever go to this function if having negative qty

alannadolny: Oh, you are right. I started to explain how it works and then I realized that I was wrong and it should be > 0

kchelstowski (approved):

drodzewicz (approved):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -676,12 +695,33 @@ class AddItemsPage extends Component {
...
           );
+          // In autosave, we don't modify the state, because
+          // lines which have not passed the validation will be
+          // deleted during users work
+          if (!withStateChange) {
+            // We want to disable saved line, so I am looking for product with the same
+            // code and quantity higher than 0 in response
+            // (to avoid disabling new line with the same productCode)
+            const savedItemsProductCodes = lineItemsBackendData.map(item => item.productCode);
+            const lineItemsAfterSave = this.state.values.lineItems.map((item) => {
```
awalkowiak, line 713: Technically there is option to enter the same product twice (or even more). You should first look into `lineItem.id` in case the specific row was updated, and then if there is no `item.id` you can look into `product.id` or `product.productCode`

awalkowiak (approved):

PR #3875, opened 2023-03-02
OBPIH-5237 Validate if merged products have pending stock movements

awalkowiak:
and do not allow to make a merge before finishing the pending stock movements

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -1118,12 +1118,9 @@ class ProductController {
             throw new IllegalArgumentException("Merge products feature is not enabled")
         }
 
-        productMergeService.validateProducts(params.primaryProduct, params.obsoleteProduct)
-
         productMergeService.mergeProduct(params.primaryProduct, params.obsoleteProduct)
 
-        flash.message = "${params.obsoleteProduct} Product merge to ${params.primaryProduct} has succeeded. " +
-            "There are currently running refresh data jobs in the background, that can take some time to process."
+        flash.message = "${warehouse.message(code: 'product.mergeProducts.success.message', args: [params.primaryProduct, params.obsoleteProduct])}"
         redirect(controller: "inventoryItem", action: "showStockCard", id: params.primaryProduct)
     }
 
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -1117,23 +1117,12 @@ class ProductController {
         if (!enabled) {
             throw new IllegalArgumentException("Merge products feature is not enabled")
         }
-        Product primaryProduct = Product.get(params.primaryProduct)
-        Product obsoleteProduct = Product.get(params.obsoleteProduct)
 
-        if (!primaryProduct) {
-            throw new IllegalArgumentException("No Product found with ID ${params.primaryProduct}")
-        }
-        if (!obsoleteProduct) {
-            throw new IllegalArgumentException("No Product found with ID ${params.obsoleteProduct}")
-        }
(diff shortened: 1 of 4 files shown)


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -1117,23 +1117,12 @@ class ProductController {
...
-        if (primaryProduct == obsoleteProduct) {
-            throw new IllegalArgumentException("Cannot merge the product with itself")
-        }
+        productMergeService.validateProducts(params.primaryProduct, params.obsoleteProduct)
 
-        productMergeService.mergeProduct(primaryProduct, obsoleteProduct)
+        productMergeService.mergeProduct(params.primaryProduct, params.obsoleteProduct)
 
-        flash.message = "${obsoleteProduct.productCode} Product merge to ${primaryProduct.productCode} has succeeded. " +
+        flash.message = "${params.obsoleteProduct} Product merge to ${params.primaryProduct} has succeeded. " +
```
jmiranda, line 1125: Probably needs to be localized. We should use the same pattern as with saved, created, deleted, updated, etc just in case we eventually want to merge other domain objects.
```
default.created.message={0} {1} created
```
So ...

```
default.merged.message = "{0} {1} was successfully merged into {0} {2}"
```
and then flash.message would look like this
```
flash.message=${g:message(code: "default.merged.label" args="[g.message(code: "product.label"), params.obsoleteProduct, params.primaryProduct]")}"
```

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -77,6 +78,9 @@ class ProductMergeService {
          *
          * */
 
+        Product primary = Product.get(primaryId)
+        Product obsolete = Product.get(obsoleteId)
+
```
jmiranda, line 83: Do we need to validate the existence of these products again? What happens if someone deletes between the validation and merge process?

Also can we add a tech debt ticket to refactor the mergeProduct method to separate each operation into a separate method.

```
def mergeProduct(String primaryId, String obsoleteId) {
    mergeProductSuppliers(primary, obsolete)
    mergeProductComponents(...)
    ... 
    mergeTransaction() 
}
```

We should be able to independently test each of these merge operations.

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -437,4 +441,30 @@ class ProductMergeService {
...
+        Product obsolete = Product.get(obsoleteId)
+        if (!obsolete) {
+            throw new IllegalArgumentException("No Product found with ID ${obsoleteId}")
+        }
+
+        if (primary == obsolete) {
+            throw new IllegalArgumentException("Cannot merge the product with itself")
+        }
+
+        def primaryRequisitionItems = requisitionService.getPendingRequisitionItems(primary)
```
jmiranda, line 460: Same with validation
```
validateProducts(String primaryId, String obsoleteId) { 
    validateProducts([primaryId, obsoleteId])
    validateOutboundStockMovementItems([primary, obsoleteId])
    ...
```

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
awalkowiak: done

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
awalkowiak: done

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
awalkowiak: done

PR #3876, opened 2023-03-02
Obpih 5281 add columns and filters to list product page

drodzewicz:


grails-app/i18n/messages.properties
```diff
@@ -4096,7 +4096,7 @@ react.stocklists.fetch.fail.label=Unable to fetch stock lists
 
 # Products list
 react.productsList.header.label=Product List
-react.productsList.includeSubcategories.label=Include all products in all subcategories
+react.productsList.includeSubcategories.label=Include all subcategories
```
jmiranda, line 4099: I think we could get away with just "Include children" or "Include subcategories" since "all" is implied by the fact that we don't have more options. This is minor but important since the less words the user sees the less they need to process. 

We might need to do some more usability around this checkbox / label because the behavior that I expected would be "include subcategories" by default. Right now we exclude subcategories by default.
image

And the user is required to check the box in order to see them. 

So perhaps we really want to include the subcategories by default and allow the user to "Exclude subcategories" if they only want to see the products in the current category. 

Unfortunately, there's probably enough nuance between "Include" and "Exclude" in terms of the processing time required by the user (i.e. to calculate what products will be returned) so it would be better if this was researched a little more extensively.

jmiranda: I remembered mint.com doing something around this but so far have only seen it with respect to their NLP-ish search query feature. 

image

NOTE: This is a screenshot from Google. I have way less than 5890 in my account.

jmiranda: I found what I was looking for ... the first two options represent (parent and children vs just parent). 

image

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -361,6 +363,10 @@ class ProductService {
                 if (params.unitOfMeasure) ilike("unitOfMeasure", "%" + params.unitOfMeasure + "%")
                 if (params.createdById) eq("createdBy.id", params.createdById)
                 if (params.updatedById) eq("updatedBy.id", params.updatedById)
+                and {
+                    if (params.createdAfter) ge("dateCreated", dateCreatedAfter)
+                    if (params.createdBefore) le("dateCreated", dateCreatedBefore)
+                }
```
jmiranda, line 369: What happens if both are null? Should we have an if (params.createdAfter or params.createdBefore) around this?

jmiranda (changes requested):

grails-app/i18n/messages.properties
drodzewicz: > I think we could get away with just "Include children" or "Include subcategories" since "all" is implied by the fact that we don't have more options. This is minor but important since the less words the user sees the less they need to process.

You are right, "all" is probably unnecessary, I'll change it to _Include subcategories_

grails-app/services/org/pih/warehouse/product/ProductService.groovy
drodzewicz: If both are `null` from what I can tell it doesn't effect the query nor does it throw any errors, but if you have concerns that maybe this can resurface as a problem on grails 3 branch then I guess wrapping this inside of an if statement could be much safer.

grails-app/conf/BootStrap.groovy
```diff
@@ -267,6 +270,36 @@ class BootStrap {
...
+            ]
+        }
+
+        JSON.registerObjectMarshaller(GlAccount) { GlAccount glAccount ->
+            [
+                id              : glAccount.id,
+                name            : glAccount.name,
+                code            : glAccount.code,
+                description     : glAccount.description,
+                glAccountType   : glAccount.glAccountType.glAccountTypeCode.name(),
```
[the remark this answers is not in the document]
drodzewicz: Yes, you have a good point, I think I'll just add a new Marshaler for `GlAccountType`
```
JSON.registerObjectMarshaller(GlAccountType) { GlAccountType glAccountType ->
    return [
        id : glAccountType.id,
        name : glAccountType.name,
        code : glAccountType.code
        accountTypeCode : glAccountType.glAccountTypeCode.name(),
    ]
}
```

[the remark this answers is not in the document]
drodzewicz: There is one more thing that occurred to me right now, but I guess  we'll talk about it on the tech huddle.
I added a bunch of new marshallers to the `Product` domain which increased the size of the Product object. What I am afraid of is that I am introducing an issue of over-fetching in some places like the [product search](https://github.com/openboxes/openboxes/blob/develop/grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy#L101) which returns a list of **Product** instances when `availableItems = true`.
Do you think that this might be a problem ? Because if so, we can look into implementing a separate Product DTO for when `availableItems = true`.

kchelstowski (approved):

awalkowiak (approved):

PR #3883, opened 2023-03-06
OBPIH-5454 add filter for product family to products list page

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -273,6 +275,14 @@ class ProductApiController extends BaseDomainApiController {
...
 
+    def productGroupOptions = {
+        def productGroups = ProductGroup.list(sort: "name").collect {
+            [id: it.id, label: "${it.name}"]
+        }
+
+        render([data: productGroups] as JSON)
+    }
+
     def categoryOptions = {
```
alannadolny, line 286: I think there is a new controller `SelectOptionsApiController` for this kind of methods.

alannadolny (changes requested):

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
drodzewicz: Thanks, I totally forgot about that

drodzewicz: I have also moved some of the other _get select options_ methods from `ProductApiController` to `SelectOptionsApiController` like tags, catalogs and categories

alannadolny (approved):

kchelstowski (approved):

awalkowiak (approved):

PR #3884, opened 2023-03-06
OBPIH-5325 Lines colours indications based on saved status response

alannadolny:
If someone will add a new indicator in the future, you have to add `showStatusIndicator` property, and each lines should have its own status. I created a simple enum with all of available statuses to make it easier to use and to gather all of the available statuses in one place. 

There is a thing that needs a little bit more thinking. For now I am adding to an item property called status `{ ...item, status: ... }`, it is not send within the request, but I am not sure if it is the appropriate name of the property, because we are using this in some other places in another feature.


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -709,49 +735,97 @@ class AddItemsPage extends Component {
...
+    // I can't check !!rowIndex, because it can be 0,
+    // so there is possibility to return false when the
+    // rowIndex is present
+    const isEdited = rowIndex !== undefined;
+    const itemsWithStatuses = values.lineItems.map((item) => {
+      if (isEdited && rowIndex === values.lineItems.indexOf(item)) {
+        return { ...fieldValue, status: SaveStatus.PENDING };
+      }
+
+      if (item.product && parseInt(item.quantityRequested, 10) <= 0) {
```
kchelstowski, line 813: is `quantityRequested` the only field that is validated in this workflow?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -709,49 +735,97 @@ class AddItemsPage extends Component {
...
+  // if rowIndex is passed, it means that we are editing row
+  // not adding new one
+  saveProgress = ({ values, rowIndex, fieldValue }) => {
+    // I can't check !!rowIndex, because it can be 0,
+    // so there is possibility to return false when the
+    // rowIndex is present
+    const isEdited = rowIndex !== undefined;
+    const itemsWithStatuses = values.lineItems.map((item) => {
+      if (isEdited && rowIndex === values.lineItems.indexOf(item)) {
+        return { ...fieldValue, status: SaveStatus.PENDING };
```
kchelstowski, line 810: doesn't the row (an item) has also a `status` property itself, that might be useful somewhere else? If it does and you are having `status` in here, you override it.
If I'm correct, we should think about a different name for this

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: There are 3 fields: requested product, quantity and recipient. If we don't have product and quantity, the line will be `pending`. Recipient is not necessary, so there is no need for validating this field. Correct me if I am wrong.

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: As I can see there is a property `statusCode` not `status`, but maybe something like `saveStatus` or `saveProgress` will be better to avoid future complication?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
kchelstowski: this is what I remembered, but got confused about the `validate` function which also contains checking for `item.boxName` and `item.palletName`. 

By the way thing that came to my mind now is that maybe instead of checking it manually here and in the `validate` function, we could here maybe just call the `validate` function and just check if returned errors are an empty object? 
What do you think? We'd obviously have to refactor `validate` function a bit or just call it with a proper argument (not item, because it expects something like:
```
{ 
  values: {
    lineItems: [item]
  }
}
```

but by doing so, if we wanted to refactor/add/remove any validation, we wouldn't have to do it in a few places, but just one.

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
kchelstowski: I think it could be a better option, but I'm not an expert in naming stuff

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: Validating function checks if field is empty or if there is a wrong value. If the field is empty, line should have `pending` status, so if i would like to use this function I have to check which type of error it is, so i think there will be some additional unnecessary code. I just need only one exact error from the validating function, so I think this will be overcomplicated if I use it.

kchelstowski: Ok, I get the point now

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
drodzewicz: @kchelstowski is right, having this value named `status` might lead into some complications.
We need to stay consistent so my suggestion is to name it `saveStatus` just like you named the _enum SaveStatus_

alannadolny: I changed the name of this property to `saveStatus`.

src/css/main.scss
```diff
@@ -1350,3 +1350,20 @@ div.search-input:focus-within {
   }
 }
 
+// Save indicators styling
+
+.indicator {
```
awalkowiak, line 1355: Let's name it a bit more descriptive. For example: `line-item-indicator` or even `line-item-save-indicator`

src/js/components/form-elements/TableRow.jsx
```diff
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import { Tooltip } from 'react-tippy';
 
 import { renderFormField } from 'utils/form-utils';
+import SaveIndicator from 'utils/SaveIndicator';
```
awalkowiak, line 8: Maybe let's call it `LineItemSaveIndicator` or `RowSaveIndicator`

src/js/components/form-elements/TableRow.jsx
```diff
@@ -93,6 +94,8 @@ class TableRow extends Component {
             hideDelay="50"
           >
             <div className="d-flex flex-row border-bottom table-inner-row">
+              {fieldsConfig.showSaveIndicator &&
+                <SaveIndicator lineStatus={rowValues.saveStatus} />}
```
awalkowiak, line 98: `lineItemSaveStatus`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -55,6 +56,7 @@ const NO_STOCKLIST_FIELDS = {
     type: ArrayField,
     arrowsNavigation: true,
     virtualized: true,
+    showSaveIndicator: true,
```
awalkowiak, line 59: `showLineItemSaveIndicator` or `showRowSaveIndicator`

src/js/consts/saveStatus.js
```diff
@@ -0,0 +1,7 @@
+const SaveStatus = {
```
awalkowiak, line 1: `LineItemSaveStatus` or `RowSaveStatus`

src/js/utils/SaveIndicator.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import SaveStatus from 'consts/saveStatus';
+
+const SaveIndicator = ({ lineStatus }) => <div className={`${lineStatus?.toLowerCase()} indicator`} />;
```
awalkowiak, line 7: `LineItemSaveIndicator` or `RowSaveIndicator`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
awalkowiak: @alannadolny as I said on our "ad hoc code review meeting" words `status` and `statusCode` are "reserved".

awalkowiak (changes requested):

awalkowiak (approved):

PR #3896, opened 2023-03-08
OBPIH-5488 Filter out inactive GlAccounts on Product list filter

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -21,35 +21,38 @@ class SelectOptionsApiController {
     GenericApiService genericApiService;
 
     def glAccountOptions = {
-        def glAccounts = genericApiService.getList(GlAccount.class.simpleName, [:]).collect {
-            [id: it.id, label: "${it.code} - ${it.name}"]
-        }
+        List<GlAccount> glAccounts = genericApiService.getList(GlAccount.class.simpleName, [:])
+                .findAll {
```
kchelstowski, line 25: I might agree it's not a big case here, since probably we don't expect too many `glAccounts`, but I don't feel like it's the right approach, since you first fetch whole list and then filter it out using `.findAll`.
If you and @awalkowiak agree with me, I (unfortunately) think this should be done in `GlAccountService` using `createCriteria`.

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
drodzewicz: I agree, also wanted to start a discussion around this topic.
I thought about moving this to it's own `GlAccountService` but I just wasn't sure if creating a whole new service just for this one case is justified, we already handle many cases this way  in our code at the moment.

I also thought about that maybe in the nearest future, after we deal with the grails migration we can start looking into refactoring a lot of the controllers by moving the logic into services, that way a lot of the code from `GlAccountController` will have to be moved to `GlAccountService` and etc.

What do you think about this @awalkowiak ?

kchelstowski: I honestly don't feel like doing some costly operations is better than following the standards (avoid creating services). 
I remember we didn't want to create services ONLY if there is already something (a service), that we could use to handle given case. 
As you wrote, anyway we will most probably have to refactor a bunch of stuff and do it, as it should be, so I feel like if you create this service now, this would reduce the amount of the work in the future :)
But as I'm not the one, to decide on such things, I'd love to hear from you and I agree with whatever decision you make, since as I wrote, there are not many `GlAccounts` and the operation doesn't cost that much, but just wanted to point out, that "generally" this is not a right approach in my opinion.

drodzewicz: I think you convinced me enough, I'll create a `GlAccountService` and will move the logic there, thanks.

grails-app/services/org/pih/warehouse/glAccount/GlAccountService.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+
+class GlAccountService {
+
+    List<GlAccount> getGlAccounts (Map params) {
+        def max = params.max ? params.int("max") : null
+        def offset = params.offset ? params.int("offset") : null
+
+        return GlAccount.createCriteria().list(max: max, offset: offset) {
+            if (params.active != null) {
```
kchelstowski, line 22: I'm wondering if there is a need to do `!= null`?
Isn't 
```suggestion
            if (params.active) {
```
enough?

kchelstowski: nevermind, I think I get it now - if we had `false` we wouldn't "go" into this `if`

kchelstowski (approved):

awalkowiak (approved):

PR #3897, opened 2023-03-08
OBPIH-5287 Add product family to Request Detail Report

alannadolny:


grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1846,6 +1849,9 @@ class JsonController {
                             destination             : it.destination,
                             productCode             : it.productCode,
                             productName             : it.productName,
+                            productFamily           : it.productFamily ?: '',
+                            category                : it.category ?: '',
+                            productCatalogs         : it.productCatalogs ?: '',
```
kchelstowski, line 1854: shouldn't catalogs be done using `.join` here?
```suggestion
                            productCatalogs         : it.productCatalogs?.join(", "),
```

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
alannadolny: I will add this in `ForecastingService` as I did it with `productCatalogs`

alannadolny: Oh, I misunderstood your comment. I thought you said about `category`. I think there is no need to add `.join` here, because it is already added in `ForecastingService`

kchelstowski: Nevermind, I thought you were doing some two reports and haven't realized that the `data` is what is returned from the `ForecastingService` where it is already done. I'm sorry for confusion.

kchelstowski: So if that's the case, I think a better option would be to do whole "logic" inside the service, so to move `?: ''` to the service and here have only 
```
productFamily        : it.productFamily,
category             : it.category,

Comment (kchelstowski) [reply grails-app/controllers/org/pih/warehouse/JsonController.groovy:1854]: + if you do join and the list is empty, the result you get is an empty string, not null, so there is not need to do `?: ''` for `productCatalogs`

**Review by kchelstowski — APPROVED** [review]

File: grails-app/controllers/org/pih/warehouse/JsonController.groovy [hunk 2]
```diff
@@ -1846,12 +1849,15 @@ class JsonController {
...
                             productName             : it.productName,
-                            unitPrice               : it.unitPrice ?: '',
+                            productFamily           : it.productFamily,
+                            category                : it.category,
+                            productCatalogs         : it.productCatalogs,
+                            unitPrice               : it.unitPrice,
                             qtyRequested            : it.quantityRequested,
                             qtyIssued               : it.quantityIssued,
                             qtyDemand               : it.quantityDemand,
-                            reasonCode              : it.reasonCode ?: '',
```
awalkowiak, line 1853: Why were "Elvis operators" removed in this file?

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
kchelstowski: I think Alan followed my suggestion about `productFamily` and moved them (elvis operators) to the service layer for every property. 
Keeping it in both  - `JsonController` and `ForecastingService` you just do twice the same, so you might run into controller having: `'' ?: ''`

awalkowiak (approved):

PR #3899, opened 2023-03-08
OBPIH-5218 Add basic routes tracking with GA

awalkowiak:


package.json
```diff
@@ -82,6 +82,7 @@
     "react-dropzone": "4.3.0",
     "react-final-form": "3.6.5",
     "react-final-form-arrays": "1.0.6",
+    "react-ga": "3.3.1",
```
mdpearson, line 85: @awalkowiak I believe the GA release supported by this plugin will be turned off July 1. Should we use the following instead?

- https://www.npmjs.com/package/react-ga4
- https://support.google.com/analytics/answer/11583528

mdpearson (changes requested):
I (still) don't speak React, so I can't comment on most of this code. The webpack changes look reasonable. My main concern about this PR is that we are doubling down on a version of Google Analytics that is going away. My vote would be to bite the bullet and use this ticket as an excuse to migrate to a supported version. Thoughts?

package.json
awalkowiak: @mdpearson yeah I am aware of this. I added the "next step / to do" to the ticket (https://pihemr.atlassian.net/browse/OBPIH-5218?focusedCommentId=140355) to replace it with the ga4 version when we will be forced to do it because what I understood from @jmiranda's comment - we want to wait until it "will be ~forced on~ done by google for us" (https://pihemr.atlassian.net/browse/OBPIH-5218?focusedCommentId=140307). Replacing the old react-ga with the new one will be easy in that case.

I will recheck tomorrow morning if react-ga4 is backward compatible with od Universal Analytics tracking id. If it is then I'll swap the lib in this PR.

webpack.config.js
```diff
@@ -42,6 +42,10 @@ module.exports = {
       templateParameters: compilation => ({
         jsSource: `\${createLinkTo(dir:'/js', file:'bundle.${compilation.hash}.js')}`,
         cssSource: `\${createLinkTo(dir:'css/', file:'bundle.${compilation.hash}.css')}`,
+        // eslint-disable-next-line no-template-curly-in-string
+        gaEnabled: '${grailsApplication.config.google.analytics.enabled}',
+        // eslint-disable-next-line no-template-curly-in-string
+        webPropertyID: '${grailsApplication.config.google.analytics.webPropertyID}',
```
jmiranda, line 48: Oh shit we do actually do this at build time. :confused:

jmiranda: That scares me a bit since someone would conceivably want to change the config properties at runtime. In fact, this is a bigger issue and they would try to change them, but changing them would have no effect.

src/js/components/Router.jsx
```diff
@@ -1,226 +1,25 @@
...
-            <AsyncResetInstancePage />
-          </Route>
-          <MainLayoutRoute path="/**/purchaseOrder/list" component={AsyncPurchaseOrderList} />
-          <MainLayoutRoute path="/**/requisitionTemplate/list" component={AsyncStockList} />
-          <MainLayoutRoute path="/**/product/list" component={AsyncProductsList} />
-          <MainLayoutRoute path="/**/stockTransfer/list" component={AsyncStockTransferList} />
-          <MainLayoutRoute path="/**/dashboard/:configId?" component={Dashboard} />
-          <MainLayoutRoute path="/**/" component={Dashboard} />
-        </Switch>
+        <SwitchWrapper />
```
jmiranda, line 22: Could you explain this component and why we needed to refactor this here?

jmiranda (changes requested):

package.json
awalkowiak: I'll investigate it further.

PR #3900, opened 2023-03-09
OBPIH-5497 Refactor product display name to use displayName instead translatedName transient variable

drodzewicz:


alannadolny (approved):

grails-app/views/deliveryNote/_printPage.gsp
```diff
@@ -90,10 +90,10 @@
                             <td class="middle" rowspan="${numInventoryItem}">
                                 <g:if test="${requisitionItem?.parentRequisitionItem?.isSubstituted()}">
                                     <div class="canceled">
-                                        <g:translatedProductName product="${requisitionItem?.parentRequisitionItem?.product}" />
+                                        ${requisitionItem?.parentRequisitionItem?.product?.displayNameOrDefaultName}
```
kchelstowski, line 93: By doing so, you got rid of the tooltip. I would avoid doing changes where `<g:translatedProductName>` is used, so it will be easier to track it and replace it with the refactored version of the taglib.

kchelstowski (changes requested):

grails-app/views/deliveryNote/_printPage.gsp
kchelstowski: so do you say that previously you used `<g:translatedProductName>` here, but the tooltip wasn't needed there then? Is that right?

drodzewicz: yes

kchelstowski (approved):

grails-app/views/deliveryNote/_printPage.gsp
kchelstowski: if so, LGTM

awalkowiak (approved):

PR #3901, opened 2023-03-09
OBPIH-5482 Add validation for unique product group name 

kchelstowski:


alannadolny (approved):

grails-app/controllers/org/pih/warehouse/product/ProductGroupController.groovy
```diff
@@ -151,6 +151,9 @@ class ProductGroupController {
                 redirect(controller: "productGroup", action: "list")
             } else {
                 println productGroupInstance.errors
+                // Refresh the instance from db, to avoid returning to the view productGroupInstance
+                // that is persisted in Hibernate session with the binded properties that didn't pass the validation
+                productGroupInstance.refresh()
```
awalkowiak, line 156: I think this is a common thing. If something does not pass the validation then you show a validation exception without reverting to the good state, but the user has to correct it.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/product/ProductGroupController.groovy
kchelstowski: The case was that by leaving it as it was, the name was also updated in the header while it should display the actual name of the product group, not the one that you tried to edit it for:
Screenshot from 2023-03-10 10-19-45

If I wanted to edit it for `test_group` (I already have `test_group`), if the validation occurs, I would see the `test_group` in the header, not the `test_group888`

awalkowiak (approved):

PR #3902, opened 2023-03-09
OBPIH-5323 Save line by line in outbound add items step (Fix for duplicating items)

alannadolny:
In this pr I added:

- Fix for `rowSaveStatus: RowSaveStatus.PENDING` after adding new line using tab or arrows
- Fix for duplicating items when they are being added too fast
- Fix for showing previously added item in next line (connected to previous point)
- Enable next button after using save button when item is not autosaved (i.e when we don't have internet connection and we click save button when connection is restored)

I fixed duplicating item by adding new property in `state`. In this variable we are storing items which were sent and then we are excluding it from checking items to save during next trigger of autosave.

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -355,7 +355,7 @@ class AddItemsPage extends Component {
       !item.statusCode &&
       parseInt(item.quantityRequested, 10) > 0 &&
       item.product);
-    // Here I am changing rowSaveStatus from EDITED to SAVING
+    // Here I am changing rowSaveStatus from PENDING to SAVING
     // because all of these lines were sent to save
     this.setState(previousState => ({
       values: {
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -299,6 +299,7 @@ class AddItemsPage extends Component {
       totalCount: 0,
       isFirstPageLoaded: false,
       isSaveCompleted: true,
+      sentToSave: [],
     };
 
     this.props.showSpinner();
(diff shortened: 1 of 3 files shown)


kchelstowski (commented):
I need more time to review this one (I'll do it tomorrow). If it looks good to you guys @awalkowiak @drodzewicz, I don't mind going further with this one

kchelstowski (approved):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -355,6 +356,13 @@ class AddItemsPage extends Component {
       !item.statusCode &&
       parseInt(item.quantityRequested, 10) > 0 &&
       item.product);
+    // Here we have adding sent items to state, because we want to avoid
+    // sending a request with the same items few times
+    // (this problem occurs only during adding new items, because
+    // items which are already saved have theirs own IDs)
+    this.setState({
+      sentToSave: [...this.state.sentToSave, ...lineItemsToBeAdded],
```
awalkowiak, line 364: Instead of doing this, can you simply mark new rows (or edited) as `PENDING` before triggering the save?

kchelstowski (approved):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -379,6 +392,20 @@ class AddItemsPage extends Component {
         key => key !== 'product',
       );
 
+      if (newQty === oldQty && oldItem.rowSaveStatus === RowSaveStatus.ERROR) {
```
awalkowiak, line 395: I don't understand this part. Does it mean, that if the newly entered quantity equals the previous state quantity and if it was errored in the previous state, then now we want to mark it as saved?

awalkowiak (commented):
Overall LGTM, but I have one question

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: It means that, if we set wrong quantity (less than 0) our row has `ERROR` status, when we set the same quantity as we had before the `ERROR` status, the request isn't send, so I am checking here if we have the same quantity as before (because this value is currently saved - that's the reason why I set `SAVED` status). When we changed the value to other than the previous one the request will be send, and we will have `SAVED` status after getting response.

awalkowiak (approved):

PR #3905, opened 2023-03-10
OBPIH-5504, OBPIH-5506 Improve localization in GSP files and taglibs

mdpearson:
This PR has fixes for OBPIH-5504 and OBPIH-5506. Some screenshots, test cases and discussion are in OBPIH-5223.


grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -926,7 +926,7 @@ class InventoryItemController {
                     log.error("Sorry, unable to add new item to shipment.  Please try again.")
                     flash.message = "${warehouse.message(code: 'inventoryItem.unableToAddItemToShipment.message')}"
                 } else {
-                    def productDescription = format.product(product: productInstance) + (inventoryItem?.lotNumber) ? " #" + inventoryItem?.lotNumber : ""
+                    def productDescription = format.product(product: productInstance).decodeHTML() + (inventoryItem?.lotNumber) ? " #" + inventoryItem?.lotNumber : ""
```
mdpearson, line 929: My understanding of Grails taglibs is that they are primarily intended to be called via, e.g., `<format:product product=...>` from within a GSP page. In that context, any reserved HTML characters must be escaped.

In Grails 1 there's no way to disambiguate between calling a taglib from a GSP page vs from a controller. Controllers do _not_ want html escapes in their strings. So, when we call a taglib from a controller, we need to stick a `.decodeHTML()` at the end. Nits like this are scattered throughout this PR; I'm calling this one out because it sorts highest in the code review.

grails-app/controllers/org/pih/warehouse/product/CategoryController.groovy
```diff
@@ -57,7 +57,7 @@ class CategoryController {
         }
 
         if (!categoryInstance.hasErrors() && categoryInstance.save()) {
-            flash.message = "${warehouse.message(code: 'category.saved.message', arg: [format.category(category: categoryInstance)])}"
+            flash.message = "${warehouse.message(code: 'category.saved.message', arg: [format.category(category: categoryInstance).decodeHTML()])}"
```
mdpearson, line 60: Here's another interesting case. We're taking the output of one taglib (`format.category`) and applying it to another (`warehouse.message`). In this case, the latter will html-encode the string it returns. If we don't call `decodeHTML()` on the argument to it, that part of the string will be double-escaped.

We don't need to call `decodeHTML()` on the outer taglib, because we're assigning it to flash.message, which must be html-escaped.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -183,9 +183,7 @@ class ProductController {
 
         if (!productInstance.hasErrors() && productInstance.save(flush: true)) {
             log.info("saved product " + productInstance.errors)
-            def warehouseInstance = Location.get(session.warehouse.id)
-            def inventoryInstance = warehouseInstance?.inventory
```
mdpearson, line 187: neither of these definitions is used

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -183,9 +183,7 @@ class ProductController {
...
-            flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'product.label', default: 'Product'), format.product(product: productInstance)])}"
+            flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'product.label', default: 'Product').decodeHTML(), format.product(product: productInstance).decodeHTML()])}"
```
mdpearson, line 186: We need exactly one `decodeHTML()` here for the same reason as in `CategoryController`, q.v.

grails-app/taglib/org/pih/warehouse/FormatTagLib.groovy
```diff
@@ -85,18 +69,17 @@ class FormatTagLib {
...
             // use the locale specified in the tag if it exists, otherwise use the user locale if it exists, otherwise use the system default locale
             // (note that we explicitly do a containsKey test because it is possible that the locale attribute has been specified but has been set to null--which means show the default locale)
             Locale defaultLocale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale)
             Locale locale = attrs.containsKey('locale') ? attrs.locale : session?.user?.locale ?: defaultLocale
-            def value = ""
-            value += LocalizationUtil.getLocalizedString(attrs.product.name.encodeAsHTML(), locale)
-
-            // default format is to display the localized name of the product
-            out << value
+            out << LocalizationUtil.getLocalizedString(attrs.product.name, locale).encodeAsHTML()
```
mdpearson, line 82: `LocalizationUtil.getLocalizedString()` won't find matches for an escaped string.

grails-app/taglib/org/pih/warehouse/FormatTagLib.groovy
```diff
@@ -120,13 +106,13 @@ class FormatTagLib {
...
             // default format is to display the localized name of the catergory
-            def value = LocalizationUtil.getLocalizedString(attrs?.category?.name?.encodeAsHTML(), locale)
+            def value = LocalizationUtil.getLocalizedString(attrs?.category?.name, locale)
             if (attrs.shorten) {
                 if (value.length() > 35) {
                     value = value.substring(0, 35) + "..."
                 }
             }
-            out << value
+            out << value.encodeAsHTML()
```
mdpearson, line 115: Here we call `encodeAsHTML()` on the string _after_ it's been localized, not before

grails-app/taglib/org/pih/warehouse/MessageTagLib.groovy
```diff
@@ -9,74 +9,29 @@
...
-
-            Locale defaultLocale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale)
-            attrs.locale = attrs.locale ?: session?.locale ?: session?.user?.locale ?: defaultLocale
-
-            def image = (!hasOthers) ? 'decline' : 'accept'
-
-            message = messageSource.getMessage(attrs.code, null, attrs.default, request.locale)
-            def resolvedMessage = "${defaultTagLib.message.call(attrs)}"
-            out << """${resolvedMessage}"""
-        }
```
mdpearson, line 79: There were a lot of unused assignments in this function

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -823,19 +795,10 @@ class SelectTagLib {
...
     }
-
-    private String optionValueToString(def el, def optionValue) {
-        if (optionValue instanceof Closure) {
-            return optionValue(el).toString().encodeAsHTML()
-        }
-
-        el[optionValue].toString().encodeAsHTML()
-    }
-
```
mdpearson, line 840: OK, if you've made it this far, you've read all the taglib code. Taglibs don't listen to `grails.views.default.codec` and need `encodeAsHTML()` sprinkled throughout. GSP pages, on the other hand, do follow this config setting, and (almost) every `encodeAsHTML()` call within them is incorrect.

This stuff changes (for the better) in Grails 3.

grails-app/views/email/_shipmentItemReceived.gsp
```diff
@@ -80,7 +80,7 @@
                             ${receiptItem?.inventoryItem?.product?.productCode}
                         </td>
                         <td>
-                            ${format.product(product: receiptItem?.inventoryItem?.product) }
+                            <format:product product="${receiptItem?.inventoryItem?.product}" />
```
mdpearson, line 83: The `${format.product ...}` call will escape the contents of the GString due to the default codec. But! this call invokes a taglib, and, taglibs need to call `encodeAsHTML()` for correctness. You'll double-escape the product name here, leading to confusing presentation.

Fortunately the fix also increases readability: `<format:product ...>` will get escaped exactly once

grails-app/views/inventory/listExpiredStock.gsp
```diff
@@ -71,6 +71,7 @@
                             </h2>
 		          			<div class="filter-list-item">
 		           				<label><warehouse:message code="category.label"/></label>
+		           				%{-- not sure what the ${{...}} in optionValue means: it may need to be de-escaped. OBPIH-5506 --}%
```
mdpearson, line 74: I added this comment because I have no idea what is going on in line 77. What does `${{...}}` do here? I either need to test it or read the spec or ask a reviewer.

grails-app/views/inventoryItem/_showCurrentStock.gsp
```diff
@@ -89,7 +89,7 @@
                     <td colspan="6">
                         <div class="fade empty center">
                             <warehouse:message code="inventory.noItemsCurrentlyInStock.message"
-                                               args="[format.product(product:commandInstance?.product)]"/>
+                                               args="[format.product(product:commandInstance?.product).decodeHTML()]"/>
```
mdpearson, line 92: `format.product()` will escape html characters, but warehouse.message takes an unescaped string as argument.

At this point I'm wondering if using taglib output as input to another taglib is an anti-pattern when special characters may be involved

grails-app/views/inventory/listExpiredStock.gsp
kchelstowski: @mdpearson you can use **closures** if you need to make some additional operations on `optionValue` instead of just using e.g. `optionValue="name"`

we can write for example:
```groovy
 optionValue="${{ it?.name?.toUpperCase() }}"
```

awalkowiak:
@mdpearson some conflicts have occured. Plus a question from my side: is it possibile that this is default codec's fault `grails.views.default.codec = "html"` (however this was specifically changed to not allow XSS)

mdpearson:
> @mdpearson some conflicts have occured. Plus a question from my side: is it possibile that this is default codec's fault `grails.views.default.codec = "html"` (however this was specifically changed to not allow XSS)

@awalkowiak Conflicts are fixed. The default codec you mention is a Good Thing. It saves us the trouble of having to call `encodeAsHTML()` throughout our GSP pages the way we need to in our tag libs.

grails-app/views/inventory/listExpiredStock.gsp
mdpearson: @kchelstowski thanks, that's very helpful!

kchelstowski (approved):

drodzewicz (approved):

jmiranda:
Do not Squash and Merge.

awalkowiak (approved):

PR #3906, opened 2023-03-10
OBPIH-5502 Refactor product translations to use displayName instead of translatedName (scope: invoice, requests, receiving) [React]

kchelstowski:
Scope:

- invoice
- request
- receiving
- product select global refactor

Those include only changes on React side, as GSP view pages will be done in other ticket and also it was blocked by the refactor of `<format:displayName>` ticket.

### DIFF
--- a/src/groovy/org/pih/warehouse/product/ProductSearchDto.groovy
+++ b/src/groovy/org/pih/warehouse/product/ProductSearchDto.groovy
@@ -8,7 +8,7 @@ class ProductSearchDto {
 
     String name
 
-    String translatedName
+    String displayName
 
     String productCode
 
@@ -45,7 +45,7 @@ class ProductSearchDto {
                 id                 : id,
                 productCode        : productCode,
                 name               : name,
-                translatedName     : translatedName,
+                displayName        : displayName,
                 color              : productColor,
                 handlingIcons      : handlingIcons,
                 lotAndExpiryControl: lotAndExpiryControl,
--- a/src/js/components/receiving/PartialReceivingPage.jsx
+++ b/src/js/components/receiving/PartialReceivingPage.jsx
@@ -166,7 +166,7 @@ const TABLE_FIELDS = {
         attributes: {
           className: 'text-left ml-1',
           showValueTooltip: true,
-          formatValue: value => getReceivingItemValue(value),
+          formatValue: getReceivingItemValue,
         },
         getDynamicAttr: ({ fieldValue }) => ({
           tooltipValue: fieldValue?.name,
--- a/src/js/components/receiving/PartialReceivingPage.jsx
+++ b/src/js/components/receiving/PartialReceivingPage.jsx
@@ -19,7 +19,7 @@ import EditLineModal from 'components/receiving/modals/EditLineModal';
 import apiClient, { flattenRequest, parseResponse } from 'utils/apiClient';
 import Checkbox from 'utils/Checkbox';
 import { renderFormField } from 'utils/form-utils';
-import renderHandlingIcons from 'utils/product-handling-icons';
+import { getReceivingItemValue, getReceivingPayloadContainers } from 'utils/form-values-utils';
 import Select from 'utils/Select';
 import Translate, { translateWithDefaultMessage } from 'utils/Translate';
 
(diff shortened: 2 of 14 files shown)


src/groovy/org/pih/warehouse/product/ProductSearchDto.groovy
```diff
@@ -45,7 +45,7 @@ class ProductSearchDto {
                 id                 : id,
                 productCode        : productCode,
                 name               : name,
-                translatedName     : translatedName,
+                displayName        : displayName,
```
kchelstowski, line 48: it was not straight forward to include whole `displayNames` here, since the data for `ProductSearch` is taken from pure SQL query.
It could be doable by fetching in the `ProductSearchDto` the `Product` by creating a method:

```groovy
Map getDisplayNames() {
  Product product = Product.get(id)
  return product.displayNames
}
```

But I think it would be an overkill and N+1 query if we wanted to fetch the whole `Product` inside the DTO that was supposed to shorten the `Product`....
thoughts?

src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -166,11 +166,7 @@ const TABLE_FIELDS = {
         attributes: {
           className: 'text-left ml-1',
           showValueTooltip: true,
-          formatValue: value => (
-            <div className="d-flex">
-              {value?.translatedName ?? value?.name}
-              {renderHandlingIcons(value?.handlingIcons)}
-            </div>),
+          formatValue: value => getReceivingItemValue(value),
```
drodzewicz, line 169: in this case you can just pass the function
```suggestion
          formatValue: getReceivingItemValue,
```

src/js/utils/option-utils.jsx
```diff
@@ -89,8 +89,8 @@ export const debounceProductsFetch = (waitTime, minSearchLength, locationId) =>
             productCode: obj.productCode,
             handlingIcons: obj.handlingIcons,
             lotAndExpiryControl: obj.lotAndExpiryControl,
-            label: `${obj.productCode} - ${obj.translatedName ?? obj.name}`,
-            hasTranslatedName: !!obj.translatedName,
+            label: `${obj.productCode} - ${obj.displayName ?? obj.name}`,
+            hasDisplayName: !!obj.displayName,
```
drodzewicz, line 93: I am wondering why  `hasDisplayName` has to be a boolean.
What I mean is that we could just pass 
```
displayName: obj.displayName
```
and then we can use this variable the same way we use `hasDisplayName` 
```
if (!!option.deisplayName) {
    ...
}
```
and furthermore we will have access to value of `displayName` if we ever gonna need that.

src/js/utils/form-values-utils.jsx
```diff
@@ -0,0 +1,52 @@
...
+      /** We have to omit product.displayNames, due to an error
+       *  while binding bindData(partialReceiptItem, shipmentItemMap)
+       *  it expects product.displayNames to have a setter, as we pass
+       *  product.displayNames.default: XYZ, to the update method, but it's not a
+       *  writable property.
+       *  With deprecated product.translatedName it was not the case, because
+       *  it was recognizing the transient and we didn't access product.translatedName.something
+       *  but product.translatedName directly
+       * */
+      return _.omit(item, 'product.displayNames');
```
drodzewicz, line 50: I will try and check this out locally because I am not sure what exactly seems to be the problem.
But if there won't be a better solution for this I think this approach should be fine.

drodzewicz (changes requested):

src/js/utils/option-utils.jsx
kchelstowski: obviously it doesn't have to be. I was also considering your point and called it explicity as boolean, because it was more readable for me at that moment, but I agree it does not really matter and if we wanted to use it somehow in the future, it would be better to make it as `displayName` and do the negations wherever it is needed or even it wouldn't be needed and I could even shorten your `if` to:

```javascript
if (option.displayName) {
 ...
}
```

Thanks for pointing that out, I will change it

awalkowiak (approved):

PR #3908, opened 2023-03-10
OBPIH-5365 Update Next functionality on add items step of outbound workflow

alannadolny:
In this ticket I finished:

- Getting rid of unnecesary performing saving current state after triggering `next` button
- Adding notification when user clicks on `next` button during the autosave

I had to disable `next` button when rows contains line with `qty == 0`, because when we have only items with qty 0, and we go to the next page we can see infinite spinner. (I can't add this validation in validating function, because it will disable `save` button also)

In function triggered by `save` button I added filtering out rows with qty 0, because this function removes these rows after performing save. There is no need to refetch the data because of the filtering.

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -314,6 +314,9 @@ class AddItemsPage extends Component {
     this.loadMoreRows = this.loadMoreRows.bind(this);
     this.updateTotalCount = this.updateTotalCount.bind(this);
     this.updateRow = this.updateRow.bind(this);
+    this.debouncedSave = _.debounce(() => {
+      this.saveRequisitionItemsInCurrentStep(this.state.values.lineItems, false);
+    }, 1000);
   }
 
   componentDidMount() {
@@ -824,11 +827,11 @@ class AddItemsPage extends Component {
     // We don't want to save the item during editing or
     // when there is an error in line
     if (isEdited) {
-      _.debounce(() => {
-        this.saveRequisitionItemsInCurrentStep(itemsWithStatuses, false);
-      }, 1000)();
+      this.debouncedSave();
       return;
     }
+
+    this.debouncedSave.cancel();
 
     this.saveRequisitionItemsInCurrentStep(itemsWithStatuses, false);
   };
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -553,8 +553,10 @@ class AddItemsPage extends Component {
   confirmTransition(onConfirm, items) {
     confirmAlert({
       title: this.props.translate('react.stockMovement.confirmTransition.label', 'You have entered the same code twice. Do you want to continue?'),
-      message: _.map(items, item =>
-        <p key={item.sortOrder}>{item.product.label} {item.quantityRequested}</p>),
+      message: _.map(items, item => (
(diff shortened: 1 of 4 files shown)


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -854,7 +829,14 @@ class AddItemsPage extends Component {
     const lineItems = _.filter(formValues.lineItems, item => !_.isEmpty(item));
 
     if (_.some(lineItems, item => !item.quantityRequested || item.quantityRequested === '0')) {
-      this.confirmSave(() => this.saveItems(lineItems));
+      this.confirmSave(() => {
+        this.saveItems(lineItems);
+        const filteredLineItems = lineItems.filter(item => item.quantityRequested > 0);
```
kchelstowski, line 834: you could also do:
```suggestion
        const totalCount = lineItems.reduce((acc, item) => (item.quantityRequested > 0 ? acc + 1 : acc), 0);
```

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -854,7 +829,14 @@ class AddItemsPage extends Component {
...
+        this.setState({
+          values: { lineItems: filteredLineItems },
+          totalCount: filteredLineItems.length,
```
kchelstowski, line 837: and here respecitvely:
```suggestion
          totalCount,
```

kchelstowski (changes requested):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: Under this line I am using `FilteredLineItems`, so why should I change this to totalCount?

```        
this.setState({
      values: { lineItems: filteredLineItems },
      totalCount: filteredLineItems.length,
  });
```

kchelstowski: oh, I haven't seen, that you are using `filteredLineItems` anywhere else other than `totalCount`. I thought you are doing this `.filter` only to get the `totalCount` of it.
Then my preposition has no sense, since it adds additional, unnecessary operation

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -658,38 +659,14 @@ class AddItemsPage extends Component {
...
     }
   }
 
-  /**
-   * Saves current stock movement progress (line items) and goes to the next stock movement step.
-   * @param {object} formValues
-   * @param {object} lineItems
-   * @public
-   */
-  saveAndTransitionToNextStep(formValues, lineItems) {
```
awalkowiak, line 675: I am not sure if this is a good idea in terms of "last line dilemma", that we discussed on slack. Please take a look at that case.

awalkowiak (commented):
Overall LGTM, but please make sure you want to remove the `saveAndTransitionToNextStep`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: I solved this by adding debounced save. I removed `saveAndTransitionToNextStep`, because we don't want to performing save if all of our lines are autosaved. Additionally we have `transitionToNextStep` which has the same responsibility as this deleted function, but without performing save.

awalkowiak (approved):

PR #3912, opened 2023-03-13
OBPIH-5501 replace translated name with display name in stock movements, returns and combined shipment

drodzewicz:


src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -260,14 +260,9 @@ class AddItemsPage extends Component {
...
           const inboundReturn = parseResponse(resp.data.data);
-          const returnItems = inboundReturn.stockTransferItems.length > 0 ?
-            _.map(inboundReturn.stockTransferItems, item => ({
-              ...item,
-              product: {
-                ...item.product,
-                label: `${item.product.productCode} ${item.product.translatedName ?? item.product.name}`,
-              },
-            })) : new Array(1).fill({ sortOrder: 100 });
+          const returnItems = inboundReturn.stockTransferItems.length > 0
```
kchelstowski, line 263: I know it's not your code, but if we already "touch" some part of the code that could be simplified, just a short suggestion, which I'm sure you are aware of, but just didn't see it
```suggestion
          const returnItems = inboundReturn.stockTransferItems.length
```

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -314,10 +311,6 @@ class AddItemsPage extends Component {
       val => ({
         ...val,
         disabled: true,
-        product: {
-          ...val.product,
-          label: `${val.productCode} ${val.product.translatedName || val.product.name}`,
-        },
```
kchelstowski, line 320: :+1: for catching all of those

src/js/utils/option-utils.jsx
```diff
@@ -114,8 +114,8 @@ export const debounceAvailableItemsFetch = (waitTime, minSearchLength) =>
             quantityAvailable: obj.quantityAvailable,
             minExpirationDate: obj.minExpirationDate,
             handlingIcons: obj.product.handlingIcons,
-            label: `${obj.productCode} - ${obj.translatedName ?? obj.name}`,
-            hasTranslatedName: !!obj.translatedName,
+            label: `${obj.productCode} - ${obj?.product.displayNames?.default ?? obj.name}`,
+            hasTranslatedName: !!obj?.product?.displayNames?.default,
```
kchelstowski, line 118: shouldn't it be renamed to:
```suggestion
            hasDisplayName: !!obj?.product?.displayNames?.default,
```

or even as we've discussed in my pr, just keep:
```javascript
displayName: obj?.product?.displayNames?.default,
```

src/js/utils/option-utils.jsx
```diff
@@ -136,8 +136,8 @@ export const debounceProductsInOrders = (waitTime, minSearchLength, vendor, dest
             name: obj.name,
             productCode: obj.productCode,
             handlingIcons: obj.handlingIcons,
-            label: `${obj.productCode} - ${obj.translatedName ?? obj.name}`,
-            hasTranslatedName: !!obj.translatedName,
+            label: `${obj.productCode} - ${obj.displayNames?.default ?? obj.name}`,
+            hasTranslatedName: !!obj.displayNames?.default,
```
kchelstowski, line 140: same as above:
```suggestion
           displayName: obj.displayNames?.default,
```

kchelstowski (changes requested):

awalkowiak (approved):
Overall LGTM, please take a look at Kacper's suggestions.

src/js/utils/option-utils.jsx
drodzewicz: From what I can see, it appears to be that `hasTranslatedName` from **debounceAvailableItemsFetch** and **debounceProductsInOrders** is not being used anywhere, so I think it's safe to remove it.

src/js/components/product-select/ProductSelect.jsx
```diff
@@ -8,15 +8,19 @@ import { debounceProductsFetch } from 'utils/option-utils';
 import renderHandlingIcons from 'utils/product-handling-icons';
 import Select from 'utils/Select';
 
+const productOptionLabel = ({
+  productCode, displayName, displayNames, name,
+}) => (`${productCode} - ${displayName || displayNames?.default || name}`);
+
```
kchelstowski, line 14: :+1:

kchelstowski (approved):

PR #3913, opened 2023-03-13
OBPIH-5499 refactor translated product name tag lib for the display name

drodzewicz:


drodzewicz:
After refactoring productTranslatedNameTagLib to format:displayName I have replaced those used instances in following places:

- Edit product > inventory levels > edit stock level (header)
- Show stock card > Product asssociations > Name column
- edit Product > Product asssociations > Associated product column
- edit Product > Product asssociations > Create product association dialog > Product row
- edit product > packages tab > edit > Product row
- show stock card > in stock tab > edit item >  Product row
- show stock card > record stock > Product row
- invoice sow page > invoice items > Name column
- requisition show page > product column
- stock movement show page > Receipt > Product column
- stock movement show page > Packing List > Product column
- list expired stock > Product column
- list expiring stock > Product column
- inventory list > product column
- stocklist show page > Product column
- consumption report

kchelstowski (approved):
I personally like the `type="html"`, because it clearly says, that we are adding some additional html `<span>` etc. while without having it, we just return "clear" string.
But I don't mind going both ways, as I also like `showTooltip="${true}"`

alannadolny (approved):

awalkowiak (approved):

PR #3919, opened 2023-03-14
OBPIH-5496 Replace translatedName with displayName name on gsp pages and reports

drodzewicz:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -681,7 +681,6 @@ class ReportService implements ApplicationContextAware {
             [
                     productCode          : it.productCode,
                     productName          : product?.displayNameWithLocaleCode,
-                    translatedProductName : product?.translatedName,
```
kchelstowski, line 684: is there a reason it has been removed, not replaced by the `displayName`?

grails-app/views/report/showOnOrderReport.gsp
```diff
@@ -158,9 +158,9 @@
                 } );
             },
             "fnRowCallback": function( nRow, aData ) {
-                $('td:eq(1)', nRow).text(aData?.translatedProductName ?? aData?.productName);
+                $('td:eq(1)', nRow).text(aData?.displayName ?? aData?.productName);
```
kchelstowski, line 161: doesn't my comment from above refer to this? I can't see you having `displayName` in the json

grails-app/services/org/pih/warehouse/report/ReportService.groovy
drodzewicz: Yes, it should be replaced, thanks for spotting that.
Not sure how I missed that :confused:

kchelstowski (approved):

PR #3924, opened 2023-03-16
OBPIH-5466 Cannot move stock to hold bin

drodzewicz:
- add ability to include/exclude locations by providing a list of ActivityCodes
- return hold bins on Add items page on outbound returns workflow
- return hold bins on create Putaway workflow
- return hold bins on stock transfer workflow
- return hold bins on stock receiving workflow

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
@@ -25,9 +25,14 @@ class InternalLocationApiController {
             throw new IllegalArgumentException("Must provide location.id as a request parameter")
         }
 
-        ActivityCode[] activityCodes = params.activityCode ? params.list("activityCode") : null
         LocationTypeCode[] locationTypeCodes = params.locationTypeCode ? params.list("locationTypeCode") : [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION]
-        List<Location> locations = locationService.getInternalLocations(parentLocation, locationTypeCodes, activityCodes)
+
+        Map activityCodeParams = [:]
+        activityCodeParams.allActivityCodes = params.allActivityCodes ? params.list("allActivityCodes") as ActivityCode[]: []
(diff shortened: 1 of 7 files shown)


drodzewicz:
So unfortunately I wasn't able to fix this issue with just passing parameters to the current Api like you @jmiranda  suggested in the ticket.
Let me go over the issue again here in the PR so that everyone is on the same page.
## The problem
The main issue for this hotfix was that many dropdowns of locations on pages such as _TransferStock_, _Create Putaway_, _Receiving_ did not include Hold Bins in them. 
The reason for that was that enpoints that were called on those pages returned a filtered list of locations that were **LocationTypeCode** = _BIN_LOCATION_ and since Hold bins are of  **LocationTypeCode_** = _INTERNAL_ that's why the hold bins were not included in the response.

## What I have tried
At first, it looked obvious to just include both parameters for locationTypeCode **[BIN_LOCATION, INTERNAL]** but when doing that it also returned some locations that we don't necessary need like _Receiving locations_  or any locations with ActivityCodes _RECEIVE_STOCK_

## My solution
I needed a way to filter out some locations based on provided ActivityCodes which is why I introduced 3 parameters _allActivityCodes_, _anyActivityCodes_, _ignoreActivityCodes_ instead of a single _ActivityCodes_ that we had before so on pages like TransferStock, Create Putaway where we don't want to see any receiving locations we pass parameters
```
    locationTypeCode: ['BIN_LOCATION', 'INTERNAL'],
    ignoreActivityCodes: ['RECEIVE_STOCK'],
```

jmiranda:
I need more time to review this one, so perhaps we can discuss during the tech huddle on Monday. Also, this should probably target master, correct?

awalkowiak:
@jmiranda it correctly targets the `0.8.21-hotfix1` branch

jmiranda:
Looks good, just need to go through a few scenarios in my head (putaway vs stock transfer, etc)

awalkowiak (approved):

awalkowiak:
@jmiranda would be good to move it forward to QA (if you won't merge it today, then I will merge it tomorrow morning)

jmiranda (approved):

PR #3926, opened 2023-03-16
OBPIH-5324 Track unsaved changes when falling online to offline

alannadolny:
I prepared reducer for autosave, if there will be a need to add tracking unsaved changes for more workflows, you just need to pass name of the workflow to an action and add this workflow to `initialState` in `autosaveReducer`. I added `lastUpdated` to `JSON` from outbound and `lastUpdated` to appropriate workflow in state. Comparision of dates is indicator of that if we need a draft option. If the date in state is newer than the date from API, it means that our changes wasn't saved.


src/js/actions/index.js
```diff
@@ -25,7 +26,7 @@ import {
   FILTER_FORM_PARAMS_BUILT,
   HIDE_SPINNER,
   REBUILD_FILTER_FORM_PARAMS,
-  REMOVE_FROM_INDICATORS,
+  REMOVE_FROM_INDICATORS, REMOVE_LINES,
```
kchelstowski, line 29: it should be in another line

kchelstowski:
```suggestion
  REMOVE_FROM_INDICATORS,
  REMOVE_LINES,
```

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -70,21 +70,30 @@ const NO_STOCKLIST_FIELDS = {
...
       // eslint-disable-next-line react/prop-types
-      addRow, getSortOrder, showOnly, updateTotalCount,
+      addRow, getSortOrder, showOnly, updateTotalCount, getStockMovementDraft, isDraftAvailable,
     }) => (
-      <button
-        type="button"
-        id="addButton"
-        className="btn btn-outline-success btn-xs"
-        disabled={showOnly}
+      <>
```
kchelstowski, line 75: Have you made sure it is working without having to write
```javascript
 <React.Fragment>
```
I remember we had to do it like this, since `<>` didn't work, but probably after I bumped up ESLint version, it might be working now, so confirm please

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -300,6 +320,8 @@ class AddItemsPage extends Component {
       newItem: false,
       totalCount: 0,
       isFirstPageLoaded: false,
+      isDraftAvailable: false,
+      workflow: 'outbound',
```
kchelstowski, line 324: could we make it as enum, something like
```javascript
 const StockMovementDirection = {
   OUTBOUND: 'outbound',
   INBOUND: 'inbound',
 }
```

so it is "prepared" for other workflows and we don't hardcode strings if those will potentially be used in more places?

src/js/actions/index.js
```diff
@@ -542,3 +543,19 @@ export const setShouldRebuildFilterParams = (flag = true) => (dispatch) => {
     type: FILTER_FORM_PARAMS_BUILT,
   });
 };
+
+export const addLines = (workflow, lines, id) => dispatch => dispatch({
```
kchelstowski, line 547:
```suggestion
export const addLines = ({ workflow, lines, id }) => dispatch => dispatch({
```
If you have many arguments it is in my opinion better to wrap it inside an object, so you don't need to follow the order while passing the arguments.
Treat it as an advice for the future or a titbit, not a change request if you don't wanna deal with refactoring it now.

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -729,6 +769,12 @@ class AddItemsPage extends Component {
     };
 
     if (payload.lineItems.length) {
+      this.props.addLines(
+        this.state.workflow,
+        itemCandidatesToSave,
+        this.state.values.stockMovementId,
+      );
+
```
kchelstowski, line 777: and here respecitvely:
```javascript
 this.props.addLines({
   workflow: this.state.workflow,
   lines: itemCandidatesToSave,
   id: this.state.values.stockMovementId,
 })

Comment (kchelstowski) [reply src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx:777]: By the way - don't we want to store those only if we lose connection? My visualization of this ticket was that you'd do something like this (pseudocode):
```javascript
  useEffect(() => {
    if (connection === offline) {
      create a draft (addLines to the redux)
    }
  }, [connectionStatus])
```

For now I can see that you add it to the reducer each time you save.
Obviously I might be wrong what the expectations were, but at the moment I can see that you do a lot of unnecessary redux operations even if you have connection.

kchelstowski (changes requested):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: Yes, it's working now

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: We have to store each change after losing connection, so you are right that there is a lot of unnecessary redux operations. I have to add an `if` statement before every adding lines to draft to reduce those operations. This approach with `useEffect` won't work because those lines have to be added to state in `then()`, `finally()`, etc. when we don't have internet connection, not when we are losing connection.

alannadolny: Right, you can only change / add quantity, because other actions needs api calls (I don't remember if changing recipient needs an api call). If you edit quantity being offline the request is not sent, so it triggers the catch function and a row has `RowSaveStatus.ERROR` and then we are saving our lines as a draft version.

src/js/reducers/autosaveReducer.jsx
```diff
@@ -0,0 +1,26 @@
+import { ADD_LINES, REMOVE_LINES } from 'actions/types';
+
+const initialState = {
```
awalkowiak, line 3: Add a TODO here: 
`TODO: Consider an option to store multiple stock movements`

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -623,13 +658,17 @@ class AddItemsPage extends Component {
...
+    const url = `/openboxes/api/stockMovements/${stockMovementId}`;
     apiClient.get(url)
       .then((resp) => {
-        const { hasManageInventory } = resp.data.data;
-        const { statusCode } = resp.data.data;
+        const { hasManageInventory, statusCode, lastUpdated } = resp.data.data;
         const { totalCount } = resp.data;
+        // if data from backend is older than the version from local storage
+        // we want to allow users use their version
+        const isDraftAvailable = (stockMovementId === id) && (lastUpdated < lastSaved);
```
awalkowiak, line 671: Perhaps you should also compare here also the statuses (`statusCode`)

awalkowiak (changes requested):

alannadolny:
I tried to add storing multiple drafts at the same time, but I found problem connected with `redux-persist`. For now we have an reducer which is ready for storing multiple stock movements and delete only one saved draft. Drafts are stored in object, each one is identified by its id
```
{
[stockMovementId]: { saved stock movement },
...
}
```
The issue occurs when you are losing internet connection, making some changes in more than one stock movement to create drafts, coming back online and refresh. After refreshing you should be able to working on more than one drafts, but only one is available. It is because of `redux-persist` behaviour. Our `react-redux` reads persisted data only after reloading page, so when we are making changes in more than one card, we are overriding previously persisted data. I tried changing an default option in our `redux-persisted` config from `stateReconciler: hardSet` (default) to `stateReconciler: autoMergeLevel1`, to avoid overriding actually saved drafts. It was working in some cases, but after: going offline -> create multiple stock movements -> refresh -> going online -> refresh, we had only one saved draft in out persisted reducer, so the previous state got overrided again. I didn't found any method for reading the currently persisted reducer to add it to our state before overriding. For now, i fixed all of the requested changes, but i didn't manage to solve this one problem.

cc @awalkowiak @kchelstowski @drodzewicz

awalkowiak (approved):

PR #3928, opened 2023-03-16
OBPIH-5153 Add shipment type filter to shipment list pages

kchelstowski:
Do not merge yet - it's ready for an initial review, but still want to refactor/shorten a few things.
Making it translatable was a pain in the neck - I forgot that we are still having a ticket/spike ticket in the backlog for handling shipment type translations - handling it on the frontend side considering all async actions getting onto each other (translate function, translations fetch, shipmentTypes fetch) added many additional, unnecessary `useEffects` and stuff.
As the result, the `shipmentTypes` translations do not work as they should - if you refresh the page, the tooltip in the select is wrong - it's the defaultMessage from the **enum**.

My conclusion is that when we work on the shipmentTypes translation ticket, we should consider maybe adding additional column to the `ShipmentType` table called e.g. `translationKey` and return it from the API already translated, as we do with the statuses.
The only difference is that statuses were enums, not a DB table, like `ShipmentType`.

FYI: I'm aware there might be some not deleted `console.logs` etc, so I will potentially look at it ASAP tomorrow, but just wanted to create a PR so that it is ready for initial review.


grails-app/i18n/messages.properties
```diff
@@ -3358,6 +3358,12 @@ react.stockMovement.deleted.success.message.label=Stock Movement has been delete
 react.stockMovement.inbound.fetching.error=Unable to fetch inbounds movements
 react.stockMovement.outbound.fetching.error=Unable to fetch outbound movements
 react.stockMovement.inbound.filters.receiptStatus.label=Receipt Status
+react.stockMovement.filters.shipmentType.label=Shipment type
```
jmiranda, line 3361: do we need a separate message for the filter. i think we should just use the same label. If one changes, the other one needs to change. 
```
react.stockMovement.shipmentType.label
```

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -375,6 +375,7 @@ class StockMovementService {
         def offset = params.offset ? params.int("offset") : null
         Date createdAfter = params.createdAfter ? Date.parse("MM/dd/yyyy", params.createdAfter) : null
         Date createdBefore = params.createdBefore ? Date.parse("MM/dd/yyyy", params.createdBefore) : null
+        List<ShipmentType> shipmentTypes = params.list("shipmentType")?.collect{ ShipmentType.read(it) } ?: null
```
jmiranda, line 378: I don't mind this, just make sure that we don't get 2N queries for this. In other words, for the normal way to do this would be to check whether the parameter exists and then do the lookup.
```
params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
```
whereas your code might be doing the same queries twice once to check if the list is null and again as part of the return of the elvis operator. 

I've never understood what happens here (whether it evaluates again or caches and returns the value from the null check).

src/js/actions/index.js
```diff
@@ -18,7 +18,7 @@ import {
   FETCH_REASONCODES,
   FETCH_REQUISITION_STATUS_CODES,
   FETCH_SESSION_INFO,
-  FETCH_SHIPMENT_STATUS_CODES,
+  FETCH_SHIPMENT_STATUS_CODES, FETCH_SHIPMENT_TYPES,
```
jmiranda, line 21: This should be on the next line for readability.

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -37,6 +38,7 @@ class OutboundStockMovementService {
         def offset = params.offset ? params.int("offset") : null
         Date createdAfter = params.createdAfter ? Date.parse("MM/dd/yyyy", params.createdAfter) : null
         Date createdBefore = params.createdBefore ? Date.parse("MM/dd/yyyy", params.createdBefore) : null
+        List<ShipmentType> shipmentTypes = params.list("shipmentType")?.collect{ ShipmentType.read(it) } ?: null
```
jmiranda, line 41: same as below.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: I honestly do not understand the point or I missed something.
From my understanding and knowledge how it works, if I do this:
```groovy
params.list("shipmentType")?.collect{ ShipmentType.read(it) } ?: null
```
The `.collect` doesn't get evaluted if the `params.list("shipmentType")` is `null` and "goes further" if it is not `null`.

So summing up, in my opinion both - my version and your version do the same thing.
But I love to hear from you a further explanation/thoughts, because I've probably missed something, but this was my understanding of those expressions.

src/js/actions/index.js
kchelstowski: yeah, this was done automatically - this is one of the "not deleted console.logs" I mentioned yesterday - I was just pushing the code late at night and didn't have the energy to fix those then, but wanted to push the code so that it gets your initial review guys, before Artur is out today.

src/js/components/stock-movement/common/ShipmentIdentifier.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { RiInformationLine } from 'react-icons/all';
+
+import ShipmentType from 'consts/shipmentType';
+
+const ShipmentIdentifier = ({ shipmentType, identifier }) => {
+  const getShipmentTypeIcon = () => ShipmentType[shipmentType?.enumKey ?? 'Default']?.icon ?? <RiInformationLine />;
```
alannadolny, line 9: As I saw in the enum all of the shipmentTypes has their own icon, so what is the case for `?? <RiInformationLine />`?

src/js/components/stock-movement/inbound/StockMovementInboundList.jsx
```diff
@@ -26,7 +29,10 @@ const StockMovementInboundList = (props) => {
         defaultValues={defaultFilterValues}
         setFilterParams={setFilterValues}
         filterFields={filterFields}
-        formProps={{ shipmentStatuses: props.shipmentStatuses }}
+        formProps={{
+          shipmentStatuses: props.shipmentStatuses,
+          shipmentTypes: props.shipmentTypes,
+      }}
```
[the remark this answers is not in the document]
kchelstowski: what do you mean by it looking odd?

src/js/components/stock-movement/inbound/StockMovementInboundTable.jsx
```diff
@@ -143,13 +145,24 @@ const StockMovementInboundTable = ({
...
       minWidth: 100,
       Cell: (row) => {
-        const { isReturn, id, order } = row.original;
+        const {
+          isReturn, id, order, shipmentType,
+        } = row.original;
         const stockMovementId = isReturn ? order?.id : id;
         return (
-          <TableCell{...row} link={`/openboxes/stockMovement/show/${stockMovementId}`} />);
+          <TableCell{...row} link={`/openboxes/stockMovement/show/${stockMovementId}`}>
```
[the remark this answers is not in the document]
kchelstowski: sure I'll add it - it was already existing like that, so I didn't notice that

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
```diff
@@ -36,6 +43,7 @@ const useOutboundListTableData = (filterParams) => {
       requestedBy: requestedBy?.id,
       createdBy: createdBy?.id,
       updatedBy: updatedBy?.id,
+      shipmentType: shipmentType?.map?.(({ id }) => id),
```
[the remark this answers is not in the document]
kchelstowski: Yes, I am sure it is working - it was not working, because you were doing it like this:
```javascript
 shipmentType?.map(({ id }) => id),
```
as `.map` is a function, it also needs a `?` after calling `map`, so it doesn't get executed if anything before that is `undefined` or `null`

src/js/components/stock-movement/common/ShipmentIdentifier.jsx
kchelstowski: it is in case we add some additional `shipmentType` in the future and someone forgets to add it the icon to the enum, so to be sure that if we forget to provide an icon, it fallbacks to default's icon

src/js/components/stock-movement/inbound/StockMovementInboundList.jsx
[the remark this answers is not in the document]
alannadolny: I mean it should be::
```        
          formProps={{
             shipmentStatuses: props.shipmentStatuses,
             shipmentTypes: props.shipmentTypes,
          }}
```
(This closing brackets should be at the same height as formProps)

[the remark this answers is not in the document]
kchelstowski: oh, ok, got it and agree. Sharp eye

src/js/components/stock-movement/common/ShipmentIdentifier.jsx
alannadolny: ok, so that's clear for me now

grails-app/conf/BootStrap.groovy
```diff
@@ -104,7 +104,8 @@ class BootStrap {
             [
                     id         : shipmentType.id,
                     name       : shipmentType.name,
-                    description: shipmentType.description
+                    description: shipmentType.description,
+                    label      : shipmentType.displayName,
```
kchelstowski, line 108: Added this to DTO to avoid some costly, complexed logic on the frontend side, which hasn't even't worked in 100% like it should due to async conflicts happening.

grails-app/domain/org/pih/warehouse/shipping/ShipmentType.groovy
```diff
@@ -37,4 +42,10 @@ class ShipmentType implements java.io.Serializable {
 
     String toString() { name }
 
+    String getDisplayName() {
+        def g = ApplicationHolder.application.mainContext.getBean( 'org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib' )
+        String defaultName = LocalizationUtil.getDefaultString(name)
+        return g.message(code: "react.stockMovement.shipmentType.${defaultName.toLowerCase()}.label", default: name)
+    }
+
```
kchelstowski, line 50: this handles the translations of shipment types that crowdin will "understand". There has been a ticket for that in the backlog that has been there for a long time, because we didn't have an idea how to handle that, since those were translated in an old way using `:fr` separators.

grails-app/conf/BootStrap.groovy
drodzewicz: To me currently this looks like a good solution but still maybe requires a tech huddle discussion since we are dealing with translating values from the database.
My general question is why was the **shipmentType** a DB record value instead of an enum to begin with?
Correct me if I am wrong but currently I don't see a workflow that would allow us to add new or delete shipment types so maybe we should look into a potential refactor to follow our convention of defining static types/values as enums, that way we can just follow our method of translation using crowdin like for every other type.

kchelstowski: yeah, I agree it probably shouldn't have been created as a DB table, but maybe there was some argument for that to be a table instead of an enum. 
I think this is only it, that is not an enum (like statuses etc), but the table. I think that this refactor should be maybe handled after grails migration, since there is a lot of associations between a domain and `ShipmentType` table, so it might be an overkill to handle it now, but I agree with you.

kchelstowski: I also think that another approach could be to just store a translation code having an additional column in the `ShipmentType` table, but I feel like my solution is more "dynamic", so whenever we decide to change any prefix or anything in the translation code, we do not have to do migrations etc. But instead of even doing so, we should probably in the future think about doing what you've said - migrating it to an enum

src/js/consts/shipmentType.jsx
```diff
@@ -0,0 +1,36 @@
...
+    messageId: `${TRANSLATION_PREFIX}.sea.label`,
+    defaultMessage: 'Sea',
+  },
+  DEFAULT: {
+    icon: <RiInformationLine />,
+    messageId: `${TRANSLATION_PREFIX}.default.label`,
+    defaultMessage: 'Default',
+  },
+};
+
```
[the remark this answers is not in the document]
kchelstowski: fyi: I agreed with @alannadolny that it didn't have any sense anymore, since after refactoring the translations, I got rid of a few usages of those and currently I'm using this call only once, so it doesn't require a function.

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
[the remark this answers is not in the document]
alannadolny: I love that explanation, thanks 😁

alannadolny (approved):

grails-app/conf/BootStrap.groovy
jmiranda: > My general question is why was the shipmentType a DB record value instead of an enum to begin with?

Because it predated our convention of using type table to wrap enums. The shipment domains (Shipment, ShipmentType, and ShipmentItem) were literally the first set of domains after Product and InventoryItem and ShipmentType was likely the first "type" domain. It hasn't been upgraded because we haven't made any significant changes to the Shipment data model since we started.

We should definitely refactor ShipmentType to include the enum, but to also incorporate a more complex real-world data model (i.e. carriers, shipping methods, etc) around it. The user should not ever be selecting a shipment type of Air, Sea, etc. They should be choosing a specific shipping option available from a carrier UPS Ground, FedEx Priority Overnight®, DHL International Direct. And those shipping options would be tied to a shipment type. 

However, it's not always that simple since it's likely (for both international and domestic shipments) there will be multiple modes of transportation involved i.e. sea container received at port of entry, truck picks up container and delivers to customs clearance dropoff, another truck picks up container and drives to train depot to be loaded onto a train. 

You get the idea. The current shipment type domain is not very useful as it is currently constituted so this needs more discussion. The good news is that PIH has no interest in changing it so this might be something I need to do on my own.

grails-app/domain/org/pih/warehouse/shipping/ShipmentType.groovy
jmiranda: Just to be completely accurate, I think you mean the pipe-delimited convention (e.g. sometext|fr:somefrenchtext).

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: I am just wondering if under the covers the elvis operator (`?:`) 
```
params.list("shipmentType").collect{ ShipmentType.read(it) } ?: null
```
does this
```
def shipmentTypes = params.list("shipmentType").collect{ ShipmentType.read(it) }
if shipmentTypes != null 
    return params.list("shipmentType").collect{ ShipmentType.read(it) }
else 
    return null
```
or this 
```
def shipmentTypes = params.list("shipmentType").collect{ ShipmentType.read(it) }
if shipmentTypes != null 
    return shipmentTypes
else 
    return null
```

The former would likely cause [negligible] performance degradation since it's doing all of those database reads twice. Obviously, we can probably live with that, but I don't want to get in the habit of doing that unless we know which it is.

grails-app/conf/BootStrap.groovy
jmiranda, line 108: I might prefer this just being called displayName since we've already used that for Product. Adding a new `label` property would just add confusion.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: It does neither I believe.
What happens behind this: `params.list("shipmentType")?.collect{ ShipmentType.read(it) }` is that:
```groovy
def shipmentType = params.list("shipmentType")
if (shipmentType != null) {
  return shipmentType.collect { ShipmentType.read(it) }
}
return null
```
so the `.collect.` doesn't get evaluated if `params.list("shipmentType")` is `null`, so the first scenario you've given cannot happen, because null safe operator prevents us from doing an operation if something before it is `null`, so the `.collect` can't be evaluted it the `params.list("shipmentType")` is `null`.

> "In Groovy we can do this shorthand by using the null safe operator (?.). If the variable before the question mark is null it will not proceed and actually returns null for you."

So, again, in my opinion both: my version and this version:
```groovy
params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
```
do the same thing, but I believe my version does it even more "efficiently", since in worst case scenario the above might call `params.list("shipmentType")` twice, wheras my version should just use "one" reference to `params.list("shipmentType")`.


Now I even actually believe, that I could just leave this:
```groovy
params.list("shipmentType")?.collect{ ShipmentType.read(it) }
```
without having to do the elvis operator.

grails-app/domain/org/pih/warehouse/shipping/ShipmentType.groovy
kchelstowski: Yeah, exactly, I apologize for the shorthand

grails-app/conf/BootStrap.groovy
kchelstowski: sure, I can change it. I've done it as `label` to avoid having to do some additional mapping on the frontend side for the `Select` to work fine, but no worries about that.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: After a quick check, I need to agree with your version - it isn't because of some 2N queries, because all I've explained is right, so I hope everyone could have some value from it, but basically why I brought back your version has a different reason:
```groovy
params.list("shipmentTypes")
```
even if there is no `params.shipmentTypes`, it evaluates to an empty list, so with my version:
```groovy
params.list("shipmentType")?.collect{ ShipmentType.read(it) } 
```
I was doing unnecessary collect on empty list, because `?` only checks if it is `null`, not if the value is truthy. I just thought `params.list` doesn't evalute do `[]`, but to `null`.

Anyway, thanks for that discussion, as there was an occasion to double check some things and consolidate some knowledge
```

PR #3929, opened 2023-03-17
OBPIH-5503 Refactor translatedName to displayName for the GSP autosug…

kchelstowski:
…gest (product select)

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
@@ -134,8 +134,7 @@ class ProductApiController extends BaseDomainApiController {
                     minExpirationDate: v.findAll { it.inventoryItem.expirationDate != null }.collect {
                         it.inventoryItem?.expirationDate
                     }.min()?.format("MM/dd/yyyy"),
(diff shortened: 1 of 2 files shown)


grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -134,8 +134,7 @@ class ProductApiController extends BaseDomainApiController {
                     minExpirationDate: v.findAll { it.inventoryItem.expirationDate != null }.collect {
                         it.inventoryItem?.expirationDate
                     }.min()?.format("MM/dd/yyyy"),
-                    color: v[0].inventoryItem.product.color,
-                    translatedName: v[0].inventoryItem.product.translatedName,
```
kchelstowski, line 138: it is not necessary, because on `ProductSelect` we get get the `displayNames` from the `product.displayNames`. Product instance is returned a few lines above and on the frontend side we have this mapping after fetching:
```javascript
{
  ...
  displayNames: obj?.product?.displayNames,
}

alannadolny (approved):

jmiranda (approved):

PR #3932, opened 2023-03-17
OBPIH-5504, OBPIH-5506 Improve localization in GSP files and taglibs (second attempt)

mdpearson:
This PR is identical to #3905 except for a correction to the refactoring in `grails-app/taglib/org/pih/warehouse/MessageTagLib.groovy`. @jmiranda I verified that it works correctly when `openboxes.locale.custom.enabled = true`.


grails-app/taglib/org/pih/warehouse/MessageTagLib.groovy
```diff
@@ -9,79 +9,31 @@
...
-            attrs.locale = attrs.locale ?: session?.locale ?: session?.user?.locale ?: defaultLocale
-            out << defaultTagLib.message.call(attrs)
-            return
-        }
+        attrs.locale = attrs?.locale ?: session?.locale ?: session?.user?.locale ?: defaultLocale
 
         // Checks the database to see if there's a localization property for the given code
-        if (session.user) {
+        if (databaseStoreEnabled && session.user) {
             def localization = Localization.findByCodeAndLocale(attrs.code, session?.user?.locale?.toString())
```
mdpearson, line 30: For a long time this code has had two slightly different ways of determining the locale. Is this intentional? Contrast line 26, which cascades through four different locale properties (1. attrs passed to the taglib 2. locale of the session 3. preferred locale of the user 4. defaultLocale of the application itself), to line 30, where we only check the user's preferred locale when looking for a localization.

This looks like it could be a subtle bug, but I don't know enough about this code to be sure.

Most of the red ink in this file is setting variables that are never used.

kchelstowski: @mdpearson currently we use `getCurrentLocale()` method from `LocalizationService` to determine the current locale. You can check this method out along with the locale logic in the `InitializationFilters`.
The following is correct for now:
session.locale > session.user.locale > defaultLocaleFromConfig > en

After the migration we'll refactor this to rely only on Spring's LCH (LocaleContextHolder) and SessionLocaleResolver.
For now, we "manipulate" this by setting it manually in the `InitializationFilters` like this:
```groovy
Locale locale = session?.locale ?: session.user?.locale ?: new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
...
LocaleContextHolder.setLocale(locale)
```

So summing up I believe in the line 30th, there should be:
```groovy
Localization.findByCodeAndLocale(attrs.code, LocalizationUtil.currentLocale)
```

mdpearson: @kchelstowski Thank you very much, your comment was very, very helpful. `LocalizationUtil.currentLocale` covers three of the four locale properties the code currently checks. To catch the remaining one (`attrs.locale`, I think I'll use the following pattern:

```
attrs.locale = attrs?.locale ?: LocalizationUtil.currentLocale  // set a locale if none is provided
...
Localization localization = Localization.findByCodeAndLocale(attrs.code, attrs.locale.toString())
```

awalkowiak (approved):

PR #3934, opened 2023-03-20
OBPIH-5159 Product Description on Amount Outstanding on Orders Report 

alannadolny:


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -967,7 +967,7 @@ class ReportService implements ApplicationContextAware {
                 "PO Number"                                         : orderItem?.order?.orderNumber,
                 "Type"                                              : "Item",
                 "Code"                                              : orderItem?.product?.productCode,
-                "Description"                                       : orderItem?.product?.description,
+                "Description"                                       : orderItem?.product?.description ?: orderItem?.product?.displayNameOrDefaultName,
```
kchelstowski, line 970: I can see that in the ticket you were supposed to replace description with the product name whereas you do fallback to product name if the description is `null`.

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
alannadolny: Oh, you're right. I was convinced that I had to do fallback to product name.

awalkowiak: Good catch @kchelstowski

kchelstowski (approved):

PR #3936, opened 2023-03-20
OBPIH-5471 Unable to add/edit inventory levels on new staging

kchelstowski:
The issue is not reproducible for me locally. The issue is not even (don't know why) reproducible on obstg, but is still reproducible on obdev1.
I went through the code and "potential candidates" that might cause this `NullPointerException`, but what worries me that on the obdev1 you can only see:

Screenshot from 2023-03-20 17-25-01

Whereas when I was trying to "manipulate" code a bit, to cause all possible NullPointers, I was always getting a detailed message:
Screenshot from 2023-03-20 16-14-05

So I'm not really convinced if the issue is strictly connected to any of those "candidates" or to some maybe mismatching between database and code, but anyway - @mdpearson - the issue seems to have fixed itself on the obstg, if it might help you, but is still visible on the obdev1.

What I tried:

- went through the `inventoryLevel/_form.gsp` and manipulated the code a bit to cause NullPointers, e.g. in 2nd line changed the value to `null`, same in line 3rd, 11 (`productInstance: null`)
- in `InventoryLevelController` in `def dialog`:
```groovy
render(template: "form", model: [inventoryLevelInstance: inventoryLevelInstance, productInstance: productInstance])
```
I changed `inventoryLevelInstance` to `null` - it caused `NullPointerException`, but I got a detailed message why.
As for `productInstance: null` - I got an infinite spinner inside the dialog.

### DIFF
--- a/grails-app/views/inventoryLevel/_form.gsp
+++ b/grails-app/views/inventoryLevel/_form.gsp
@@ -64,7 +64,7 @@
                         <g:hiddenField name="location.id" value="${locationInstance?.id}"/>
                         <g:select name="displayLocation.id" from="${[locationInstance]}"
                             optionKey="id" optionValue="${{format.metadata(obj:it)}}" class="chzn-select-deselect"
-                                                  value="${locationInstance.id}" disabled="${true}"/>
+                                                  value="${locationInstance?.id}" disabled="${true}"/>
                     </td>
                 </tr>
                 <tr class="prop">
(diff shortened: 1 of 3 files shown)


mdpearson:
@kchelstowski I commented on the ticket with some thoughts on diagnosis. The patch here looks reasonable to me. Thank you!

jmiranda:
While this PR looks fine, I don't want to merge this until we actually know what's going on.

jmiranda (commented):
While this PR looks fine, I don't want to merge this until we actually know what's going on.

awalkowiak:
Closing this one, since it was resolved by @jmiranda

PR #3939, opened 2023-03-21
OBPIH-5189 Color code formulary (fixes after QA)

alannadolny:


src/js/utils/Select.jsx
```diff
@@ -221,7 +222,7 @@ class Select extends Component {
     const SingleValue = props => (
       <components.SingleValue {...props}>
         {this.props.valueRenderer ? (
-          this.props.valueRenderer(props.data)
+          this.props.valueRenderer({ ...props.data, showSelectedOptionColor })
```
kchelstowski, line 225: What hides under the `props.data`?
I'm wondering why you have to do it this way if e.g. `showLabelTooltip` is passed in the same way (using attributes) and it isn't provided here like this.

src/js/utils/Select.jsx
```diff
@@ -433,6 +435,7 @@ Select.defaultProps = {
   onEnterPress: null,
   optionRenderer: null,
   valueRenderer: null,
+  showSelectedOptionColor: null,
```
kchelstowski, line 438: If you had made prop type as `bool`, let's stick to it also for the default value.
```suggestion
  showSelectedOptionColor: false,
```

kchelstowski (changes requested):

src/js/utils/Select.jsx
alannadolny: props.data is a selected product, for example:
```
{
    "value": "ff8081816e1e6731016e1e6782610003",
    "id": "ff8081816e1e6731016e1e6782610003",
    "name": "Advil 200mg",
    "productCode": "00001",
    "handlingIcons": [],
    "lotAndExpiryControl": null,
    "displayName": null,
    "color": null,
    "exactMatch": true
}
```
I need this property in `SelectedValue` (in `ProductSelect` file), `showLabelTooltip` is not being used here. It is used within the `Select` component, but I need `showSelectedOptionColor` one level deeper.

awalkowiak (approved):

PR #3944, opened 2023-03-22
OBPIH-5525 match exactly product code issue in inbound returns

drodzewicz:


src/js/components/form-elements/FieldArrayComponent.jsx
```diff
@@ -21,15 +21,12 @@ class FieldArrayComponent extends Component {
...
-  focusField(index, fieldName, options = {}) {
+  focusField(index, fieldName) {
     const field = _.get(this.fieldRefs, `[${index}].${fieldName}`);
     // 8 - the amount of rows shown in the table on 1360x786 resolution
     const fieldToScroll = _.get(this.fieldRefs, `[${index - 8 > 0 ? index - 8 : 0}].${fieldName}`);
 
     if (field) {
-      if (options.enable) {
-        field.disabled = false;
-      }
```
drodzewicz, line 32: This option was something I added when working on this feature for selecting product on exact match.
It's purpose was to force enable a field so it can be focusable.
Unfortunately it cause a bug on outbound workflow which made the selected product disappear.

src/js/components/product-select/ProductSelect.jsx
```diff
@@ -76,13 +76,12 @@ const ProductSelect = ({
...
-      if (onExactProductSelected) {
-        onExactProductSelected({ product: exactMatchProduct });
-      }
       if (exactMatchProduct) {
         selectRef.current.select.select.setValue(exactMatchProduct);
       }
-
+      if (onExactProductSelected) {
+        setTimeout(() => onExactProductSelected({ product: exactMatchProduct }), 0);
+      }
```
drodzewicz, line 84: Wrapping `onExactProductSelected` in a **setTimeout** with **0 ms** is kind of a workaround/hack to make this function act asynchronously.
The issue was that on outbound workflow `quantityRequested` field is initially disabled and thus is unfocusable. It turns to enabled only after product is selected so that means this callback would need to be executed after the state change.
Wrapping this callback inside of a **setTimeout** puts it in the **event que** just like the `setState` unlike the regular code that is beeing executed immediately.

src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -190,13 +192,8 @@ class AddItemsPage extends Component {
...
 
-  updateRow(values, index) {
-    const item = values.returnItems[index];
-    this.setState({
-      formValues: update(values, {
-        returnItems: { [index]: { $set: item } },
-      }),
-    });
+  updateFormValues(values) {
+    this.setState({ formValues: values });
```
drodzewicz, line 196: I have refactored this whole `updateRow` function since to me the logic that is inside seems redundant.
Maybe I am not seeing something so please point it out to me but it looks like it extracts `item` from `values` and later modifies `values` bu inserting `item` to the same index that it was previously extracted from.

src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -74,24 +76,24 @@ const FIELDS = {
         flexWidth: '4',
         required: true,
         getDynamicAttr: ({
-          updateRow, rowIndex, values, originId, focusField,
+          updateFormValues, rowIndex, values, originId, focusField,
         }) => ({
           locationId: originId,
           onExactProductSelected: ({ product }) => {
             if (focusField && product) {
               focusField(rowIndex, 'quantity');
```
kchelstowski, line 84: just to understand - you say that your workaround with `setTimeout` has to be there, because in this place we don't have `product` yet?
What actually "updates" the product after clicking enter? Is it `onBlur`?
Those are only clarification questions so that I could understand what exactly is going on

kchelstowski: Thanks for that explanation. I think it is what I thought, but I'm gonna read it in details tomorrow and give a review. Anyway I rely on you in this feature, so if others feel okay with those changes, go ahead, I just wanted to understand the case more if I ever fall to something similar.

awalkowiak (commented):
It's weird, but I guess ok as a workaround. Could you document it on the code?

kchelstowski (approved):
I think we all agree that the solution with the `setTimeout` is quite unique :smile: but as @awalkowiak if you add a comment about that, it would be great. Anyway, I think there should be a way to do it without having to do such tricks, but I don't think it's worth to spend time to figure that out.

PR #3945, opened 2023-03-22
OBPIH-5437 Add search bar in Pack page step of outbound creation

kchelstowski:
The ticket was supposed to be straight forward, but the problem occured, because there is not any place yet, where we'd have a search/filter bar and not having `subfieldKey` property for the field.
As the result, the table's height was not counted properly then - it was only making the rows blank instead of "deleting" the row.
This is why I had to provide some changes to the calculation of height - this method was difficult to understand for me, so I didn't wanna do much refactor there and just refactored in the place that was necessary.
I did a round of "smoke tests" if I hadn't broken anything due to those changes, but everything seems to be working fine.

### DIFF
--- a/src/js/components/form-elements/FilterInput.jsx
+++ b/src/js/components/form-elements/FilterInput.jsx
@@ -1,0 +1,44 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { getTranslate } from 'react-localize-redux';
+import { connect } from 'react-redux';
+
+import { translateWithDefaultMessage } from 'utils/Translate';
+
+const FilterInput = ({
+  itemFilter, onChangeFunc, clearFunc, translate,
+}) => (
+  <div className="d-flex mr-auto justify-content-center align-items-center">
+    <input
+      value={itemFilter}
+      onChange={onChangeFunc}
+      className="float-left btn btn-outline-secondary btn-xs filter-input mr-1 mb-1"
+      placeholder={translate('react.stockMovement.searchPlaceholder.label', 'Search...')}
+    />
+    {itemFilter &&
+      <i
+        role="button"
+        className="fa fa-times-circle"
+        style={{ color: 'grey', cursor: 'pointer' }}
+        onClick={() => clearFunc()}
+        onKeyPress={() => clearFunc()}
+        tabIndex={0}
+      />
+    }
+  </div>
+);
+
+const mapStateToProps = state => ({
+  translate: translateWithDefaultMessage(getTranslate(state.localize)),
+});
+
+export default connect(mapStateToProps)(FilterInput);
+
+
+FilterInput.propTypes = {
+  itemFilter: PropTypes.string.isRequired,
+  onChangeFunc: PropTypes.func.isRequired,
+  clearFunc: PropTypes.func.isRequired,
+  translate: PropTypes.func.isRequired,
+};
--- a/src/js/components/form-elements/TableBodyVirtualized.jsx
+++ b/src/js/components/form-elements/TableBodyVirtualized.jsx
@@ -28,18 +28,20 @@ class TableBodyVirtualized extends Component {
 
   getHeight() {
     const { fieldsConfig: { subfieldKey, getDynamicRowAttr }, fields, properties } = this.props;
-    const { totalCount } = properties;
     let height = 0;
     const maxTableHeight = window.innerHeight < 900 ?
       // 0.35 * window.innerHeight = 35vh from table-content class in StockMovement.scss
       0.35 * window.innerHeight : 0.40 * window.innerHeight;
 
     if (!subfieldKey) {
-      if (totalCount * 28 > maxTableHeight) {
-        height = maxTableHeight;
-      } else if (totalCount > 0) {
-        height = totalCount * 28;
-      }
+      height = fields.value.reduce((acc, field) => {
+        const dynamicAttr = getDynamicRowAttr ?
+          getDynamicRowAttr({ ...properties, rowValues: field }) : {};
+        if (dynamicAttr.hideRow || height > maxTableHeight) {
+          return acc;
+        }
+        return acc + 28;
+      }, height);
     } else {
       _.forEach(fields.value, (field) => {
         const dynamicAttr = getDynamicRowAttr ?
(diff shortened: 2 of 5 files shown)


src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -59,31 +61,31 @@ class TableBodyVirtualized extends Component {
         }
       });
     }
-
     return height || 28;
   }
 
   getRowHeight({ index }) {
     const { fieldsConfig: { subfieldKey, getDynamicRowAttr }, fields, properties } = this.props;
+    const rowValues = fields.value ? fields.value[index] : null;
```
alannadolny, line 69: Can't you use null safe operator here?

src/js/components/stock-movement-wizard/outbound/PackingPage.jsx
```diff
@@ -463,6 +470,11 @@ class PackingPage extends Component {
             <AlertMessage show={this.state.showAlert} message={this.state.alertMessage} danger />
             { !showOnly ?
               <span className="buttons-container">
+                <FilterInput
+                  itemFilter={itemFilter}
+                  onChangeFunc={e => this.setState({ itemFilter: e.target.value })}
+                  clearFunc={() => this.setState({ itemFilter: '' })}
```
alannadolny, line 476: Are you sure it is working correctly in all cases? `setState` is asynchronous function, so I am thinking about situation when you input a value, then clear the value and you can see that the things that you inputted appears in the field again, because of that `clearFunc` finished its changing state earlier than `onChangeFunc`

kchelstowski: I get your point, but I think you are overcomplicating simple events with the same, not complexed logic applied. In "normal world" it is not possible, and this is how everyone does the clearing of form/inputs. 
Have you ever experienced anything like this and have any proposed solution? You could also pull my branch and try to achieve that :smile: I couldn't achieve that.
Answering your question - yes, I made sure it is working correctly.

src/js/components/form-elements/TableBodyVirtualized.jsx
kchelstowski: sure I could, but I was just copying the code from bottom to the top, so didn't even much care what was there in details. 
But it also depends if the `rowValues` could be `undefined` in this case, because if I used null safe operator, it would evaluate to `undefined`, so I gotta check if there is none strict checking if `rowValues` is `=== null`.

src/js/components/form-elements/FilterInput.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { getTranslate } from 'react-localize-redux';
+import { connect } from 'react-redux';
+
+import { translateWithDefaultMessage } from 'utils/Translate';
+
+const FilterInput = ({
+  itemFilter, onChangeFunc, clearFunc, translate,
```
drodzewicz, line 10:
```suggestion
  itemFilter, onChange, onClear, translate,
```
onClick, onClear etc.. is a more conventional name for listeners in React.

src/js/components/form-elements/FilterInput.jsx
```diff
@@ -0,0 +1,44 @@
...
+      className="float-left btn btn-outline-secondary btn-xs filter-input mr-1 mb-1"
+      placeholder={translate('react.stockMovement.searchPlaceholder.label', 'Search...')}
+    />
+    {itemFilter &&
+      <i
+        role="button"
+        className="fa fa-times-circle"
+        style={{ color: 'grey', cursor: 'pointer' }}
+        onClick={() => clearFunc()}
+        onKeyPress={() => clearFunc()}
```
drodzewicz, line 25: Could you explain why you have both `onClick` and `onKeyPress` assigned a same `clearFunction` ?

src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -28,18 +28,20 @@ class TableBodyVirtualized extends Component {
...
-      } else if (totalCount > 0) {
-        height = totalCount * 28;
-      }
+      height = fields.value.reduce((acc, field) => {
+        const dynamicAttr = getDynamicRowAttr ?
+          getDynamicRowAttr({ ...properties, rowValues: field }) : {};
+        if (dynamicAttr.hideRow || height > maxTableHeight) {
+          return acc;
+        }
+        return acc + 28;
```
drodzewicz, line 43: this `28` value can be seen in many places in the file, I think it would be a good idea to maybe move it to a named variable. It will be much easier if we ever want to change this value.
Also a name for the variable should provide an explanation to what this value stands for, because at the moment I don't know what `28` is _(margin, height, padding ?)_

drodzewicz (changes requested):

src/js/components/form-elements/TableBodyVirtualized.jsx
kchelstowski: As you can see, `height =`, so obviously it is height.
Sure I can move it to a variable, as I said in the comment, I hadn't wanted to make much refactor to this, since anyway it was a pain in the neck to understand what is going on here, so I didn't want to cause any regression in case I don't understand any case.
But of course moving this value to a variable isn't "risky", so I'll do that, thanks for the advice.

src/js/components/stock-movement-wizard/outbound/PackingPage.jsx
alannadolny: Maybe you're right that I am overcomplicating this 😄 User has to be faster than light to be able to reproduce that

awalkowiak (commented):
Oof. I am gonna have to review this one thoroughly.

src/js/components/form-elements/FilterInput.jsx
kchelstowski: The answer is that the ESLint shouts if you don't add at least one key listener for a static element such as `<i>` or `<div>`, so that is why it was probably added back in the past.

> Visible, non-interactive elements with click handlers must have at least one keyboard listener.

src/js/components/form-elements/TableBodyVirtualized.jsx
```diff
@@ -28,18 +31,20 @@ class TableBodyVirtualized extends Component {
...
       // 0.35 * window.innerHeight = 35vh from table-content class in StockMovement.scss
       0.35 * window.innerHeight : 0.40 * window.innerHeight;
 
     if (!subfieldKey) {
-      if (totalCount * 28 > maxTableHeight) {
-        height = maxTableHeight;
-      } else if (totalCount > 0) {
-        height = totalCount * 28;
-      }
+      height = fields.value.reduce((acc, field) => {
```
awalkowiak, line 40: @kchelstowski have you tested this against cases where there were more items on the page (around 20+)? Plus, have you tested if there is no regression on substituting items, split picking or split packing?

awalkowiak (commented):
Overall LGTM, but I am scared of some regressions on the virtualized table component

kchelstowski:
@awalkowiak the only issue I've found and have just pushed was for table with many rows - the scroll didn't appear, so I couldn't go to the top. Thanks for reminding me of that case, as I must've not checked it before

drodzewicz (approved):

PR #3946, opened 2023-03-22
OBGM-419 Update node plugin to, among other things, build cleanly on remote agents

mdpearson:
Minor tweak to support remote agents on the new Bamboo server, since the old one seems dead


mdpearson:
Merging this without review because the build is broken without it.

awalkowiak (commented):
lgtm

PR #3947, opened 2023-03-23
OBPIH-5433 Add Validation for Stock Movement deletion. 

alannadolny:
When stock movement is deleted there is a redirect to react pages with param `deleted: true`. After redirect we check if there is this param passed, if yes we trigger notification.

There were errors validation before, but because of redirect to react pages, we didn't saw the appropriate messages.


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -211,12 +211,17 @@ class StockMovementController {
                 try {
                     stockMovementService.deleteStockMovement(params.id)
                     flash.message = "Successfully deleted stock movement with ID ${params.id}"
+                    params.deleted = 'true'
```
kchelstowski, line 214: I'd add a comment above why it is needed and also can't it be boolean here instead of the string?

src/js/components/stock-movement/inbound/StockMovementInboundList.jsx
```diff
@@ -22,6 +23,10 @@ const StockMovementInboundList = (props) => {
 
   useTranslation('stockMovement', 'reactTable');
 
+  useEffect(() => {
+    triggerDeletionNotification(props.location?.search);
+  }, []);
+
```
kchelstowski, line 29: I know we were discussing it yesterday, but now when I see the code, I see a possibility to move it to a custom hook, called maybe: `useDeleteNotificationListener`

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz, line 214: This is not a change request but rather a discussion point.

do you think that maybe it would be better instead of passing this `deleted` flag to pass

```groovy
params.notificationMessage = g.message(
'react.stockMovement.deleted.success.message.label',
'Stock Movement has been deleted successfully',
)
params.notificationType='SUCCESS'

```

because at the moment your `notification-utils` are very specific to the case of deleting a stock Movement and if we ever want to add an exact same logic for example a success message of invoice or something like that we would need to add an additional utils function for that specific case

kchelstowski (changes requested):
other than that LGTM

src/js/components/stock-movement/outbound/StockMovementOutboundList.jsx
```diff
@@ -22,6 +23,10 @@ const StockMovementOutboundList = (props) => {
 
   useTranslation('stockMovement', 'StockMovementType', 'reactTable');
 
+  useEffect(() => {
+    triggerDeletionNotification(props.location?.search);
```
drodzewicz, line 27: I think you should also consider clearing that `deleted="true"` param from search query after extracting the value and displaying the notification.

src/js/utils/notification-utils.jsx
```diff
@@ -0,0 +1,17 @@
...
+  const parsedQueryParams = queryString.parse(queryParams);
+  if (parsedQueryParams.deleted) {
+    notification(NotificationType.SUCCESS)({
+      message: translate({
+        id: 'react.stockMovement.deleted.success.message.label',
+        defaultMessage: 'Stock Movement has been deleted successfully',
+      }),
+    });
+  }
+}
```
drodzewicz, line 17: As I have mentioned in my first comment about generalizing this utils function it would look  something like that
```js
export default function triggerNotification(queryParams) {
  const parsedQueryParams = queryString.parse(queryParams);
  const { notificationMessage, notificationType } = parsedQueryParams
  if (notificationMessage) {
    notification(notificationType || NotificationType.SUCCESS)({
      message: notificationMessage,
    });
  }
}
```

drodzewicz (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
alannadolny: I think it is a good idea. I am just worry about the length of the params. If there is nothing against passing this kind of param, I will do it.

kchelstowski: if we were about to go this way, I'd consider adding an enum for the notification type on the backend side similar to the one on the frontend side. I don't think you have to worry about "the length the param", because "inside" it will be only a simple string after the g.message executes.

drodzewicz: The max url length is 2048 characters, so on that part it should be ok.
Also just as I mentioned in one of my comments you could clear that param once you display the notification

src/js/components/stock-movement/outbound/StockMovementOutboundList.jsx
alannadolny: I didn't do that, because the param was deleted on list pages somehow, but to make this function more reusable, I will add this.

kchelstowski: it was cleared probably while fetching the list data, because there we also manipulate the query params, so maybe it was cleared then

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
alannadolny: I did this accidentally, in other places i use boolean instead of string.

src/js/components/stock-movement/inbound/StockMovementInboundList.jsx
alannadolny: I am more willing to do it as a hook than keeping the logic connected with `useEffect` here.

grails-app/controllers/org/pih/warehouse/inventory/StockTransferController.groovy
```diff
@@ -99,18 +100,26 @@ class StockTransferController {
 
         try {
             stockTransferService.deleteStockTransfer(params.orderId ?: params.id)
-        } catch (IllegalArgumentException e) {
-            flash.message = "${warehouse.message(code: 'default.not.found.message', args: [warehouse.message(code: 'inventory.stockTransfer.label', default: 'Stock Transfer'), params.id])}"
+            params.notificationMessage = g.message(
+                    code: 'react.stockMovement.deleted.success.message.label',
+                    body: 'Stock Movement has been deleted successfully',
+            )
+            params.notificationType = NotificationType.SUCCESS
```
kchelstowski, line 107: Going further what I thought about is creating some util class that would have properties `message` and `type`, so it would be easier to control if we ever consider adding additional properties or anything, but it's just a thought which I wanted to bring out.

kchelstowski (approved):

src/js/hooks/useDeleteNotificationListener.jsx
```diff
@@ -0,0 +1,30 @@
+import { useEffect } from 'react';
+
+import _ from 'lodash';
+import queryString from 'query-string';
+import { useHistory, useLocation } from 'react-router-dom';
+
+import notification from 'components/Layout/notifications/notification';
+import NotificationType from 'consts/notificationTypes';
+
+const useDeleteNotificationListener = () => {
```
drodzewicz, line 10: `useDeleteNotificationListener` seems a bit too specific, shouldn't it be more generic like `useQueryMessageNotificationListener` ?

src/js/components/stock-movement/outbound/StockMovementOutboundList.jsx
```diff
@@ -21,6 +22,7 @@ const StockMovementOutboundList = (props) => {
   } = useOutboundFilters(props.isRequestsList);
 
   useTranslation('stockMovement', 'StockMovementType', 'reactTable');
+  useDeleteNotificationListener();
```
drodzewicz, line 25: I feel like now this hook is generic enough that it can be placed somewhere near the root element like for example on `Router.jsx` where we have the `useConnectivityListener`.
That way this feature would be available across all the react pages

drodzewicz (changes requested):

awalkowiak (commented):
I am gonna review it tomorrow

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -210,13 +211,21 @@ class StockMovementController {
             if (stockMovement?.shipment?.currentStatus == ShipmentStatusCode.PENDING || !stockMovement?.shipment?.currentStatus) {
                 try {
                     stockMovementService.deleteStockMovement(params.id)
-                    flash.message = "Successfully deleted stock movement with ID ${params.id}"
+                    params.notificationMessage = g.message(
```
jmiranda, line 214: params.notificationMessage doesn't seem necessary. Just use the flash.message. Objects placed in flash scope should persist through a redirect. 

https://docs.grails.org/3.1.1/ref/Controllers/flash.html

src/groovy/org/pih/warehouse/api/NotificationType.groovy
```diff
@@ -0,0 +1,14 @@
+package org.pih.warehouse.api
+
+enum NotificationType {
```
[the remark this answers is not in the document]
jmiranda: :) yeah this doesn't feel right to me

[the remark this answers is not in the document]
jmiranda: this might be an ok solution, but I don't even think I understand the problem. Could someone upload a video to the ticket that demonstrates the problem and all of the places it occurs? Are we basically just trying to solve a communication issue between GSP and React? Or is there something else going on?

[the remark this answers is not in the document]
jmiranda: So I reviewed the rest of the code and I think understand the problem a little better. It does seem to be a "how do we get some bits of information from the previous request to the next request within the confines of a redirect". Which for grails / spring is the flash scope. But I've never had to think about this from Reacts perspective. 

I'm less hesitant about moving forward with this solution so I'll leave it up to you guys.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -235,7 +239,8 @@ class StockMovementController {
         params.direction = (currentLocation == stockMovement.origin) ? StockMovementDirection.OUTBOUND :
                 (currentLocation == stockMovement.destination) ? StockMovementDirection.INBOUND : "ALL"
 
-        redirect(action: "list", params:params)
+
+        redirect(action: "list", params: params << ["flash": flash as JSON])
```
awalkowiak, line 243: I think this `params << ["flash": flash as JSON]` should happen in the `list` action while rendering the template.

grails-app/controllers/org/pih/warehouse/inventory/StockTransferController.groovy
```diff
@@ -99,18 +97,26 @@ class StockTransferController {
...
+                    code: 'react.stockMovement.deleted.success.message.label',
+                    body: 'Stock Movement has been deleted successfully',
+            )
+        } catch (Exception e) {
+            flash.message = "${warehouse.message(code: 'stockMovement.delete.error.message')}"
+            redirect(controller: "stockMovement", action: "show", id: params.id)
+            return
         }
 
+        params = params << ["flash": flash as JSON]
```
awalkowiak, line 110: I think this `params << ["flash": flash as JSON]` should happen in the `list` action or in the `list` action of the stock movement controller before rendering the template.

src/js/hooks/useFlashScopeListener.jsx
```diff
@@ -0,0 +1,36 @@
...
+    history.replace({
+      search: parsedClearedQuerySearch,
+    });
+  };
+
+  useEffect(() => {
+    if (!flash) {
+      return;
+    }
+    const { message } = JSON.parse(flash);
```
awalkowiak, line 28: You can add also an option for `error` message.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
alannadolny: It cannot be in the `list` action, because there are two redirects:

- from remove action to list action
- from list action to react pages

Flash message is persisted only during one redirect

awalkowiak (approved):

PR #3950, opened 2023-03-23
OBPIH-5229 allow filter by payment terms in po list page

drodzewicz:


awalkowiak (approved):

src/js/actions/index.js
```diff
@@ -424,6 +426,17 @@ export function fetchPurchaseOrderStatuses() {
   };
 }
 
+export function fetchPaymentTerms() {
+  return (dispatch) => {
+    purchaseOrderApi.getPaymentTerms().then((res) => {
```
kchelstowski, line 431: for this simple case it doesn't matter, but IMO async/await convention is generally cleaner

src/js/hooks/list-pages/purchase-order/usePurchaseOrderListTableData.jsx
```diff
@@ -27,14 +27,15 @@ const usePurchaseOrderListTableData = (filterParams) => {
...
-      status, origin, orderedBy, createdBy, destinationParty, destination,
+      status, origin, orderedBy, createdBy, destinationParty, destination, paymentTerm,
     } = filterParams;
     return {
       ..._.omitBy({
         offset: `${offset}`,
         max: `${state.pageSize}`,
         ...sortingParams,
         ...filterParams,
+        paymentTerm: paymentTerm && paymentTerm.map(term => term.id),
```
kchelstowski, line 38: could be simplified to:
```javascript
paymentTerm: paymentTerm?.map?.(({ id }) => id),
```

src/js/reducers/purchaseOrderReducer.jsx
```diff
@@ -1,4 +1,4 @@
-import { FETCH_PURCHASE_ORDER_STATUSES } from 'actions/types';
+import { FETCH_PAYMENT_TERMS, FETCH_PURCHASE_ORDER_STATUSES } from 'actions/types';
 
 const initialState = {
   statuses: [],
```
kchelstowski, line 4: you haven't added an `initialState` for `paymentTerms`. I hope it won't break anything and won't cause any null pointers

kchelstowski (commented):
"post-factum" review

src/js/reducers/purchaseOrderReducer.jsx
drodzewicz: Thanks for spotting that, I'll make sure to add it in one of the other PR's tat I'l be working on

PR #3955, opened 2023-03-28
OBPIH-5229 Add option to filter purchase orders by empty payment term

drodzewicz:


src/js/actions/index.js
```diff
@@ -426,16 +426,21 @@ export function fetchPurchaseOrderStatuses() {
...
-      dispatch({
-        type: FETCH_PAYMENT_TERMS,
-        payload: res.data.data,
-      });
-    });
+export const fetchPaymentTerms = ({ translate }) => async (dispatch) => {
+  const blankPaymentTermOption = {
+    id: 'null',
+    value: 'null',
+    label: translate?.(
```
kchelstowski, line 433: unfortunately I'm afraid that it will most of the times fallback to the default value, due to differences in `translate` being already properly set while running into this function + due to translations not being fetched yet (I've run into the same issue with the shipment types when trying to do it like this).
I'd consider adding this to the backend where you return the JSON.
Plus I'm wondering why you have both `id` and `value` as `'null'` (as string).

kchelstowski (changes requested):

src/js/actions/index.js
```diff
@@ -426,16 +426,21 @@ export function fetchPurchaseOrderStatuses() {
...
+      'react.purchaseOrder.column.blankPaymentTermOption.label',
+      'Blank Payment Term',
+    ),
   };
-}
+  const response = await purchaseOrderApi.getPaymentTerms();
+  return dispatch({
+    type: FETCH_PAYMENT_TERMS,
+    payload: [blankPaymentTermOption, ...response?.data?.data],
```
alannadolny, line 441: If one part from the chaining: ` ...response?.data?.data` will be null your payload will look like: `[blankPaymentTermOption, '']`. Is it intended?

src/js/hooks/list-pages/purchase-order/usePurchaseOrderFilters.jsx
```diff
@@ -25,7 +27,9 @@ const usePurchaseOrderFilters = () => {
     paymentTerms,
     currentUser,
     currentLocale,
+    translate,
   } = useSelector(state => ({
+    translate: translateWithDefaultMessage(getTranslate(state.localize)),
```
alannadolny, line 32: We are using `translateWithDefaultMessage(getTranslate(state.localize))` really often (I took this function just for example). What do you think about creating a place for all of ours 'selectors'? (Maybe just as a future improvement)
After that we can do something like:
```
translate: getTranslation()
```
If there will be a need for only one selector we can skip the `state` argument and do only:
```
const translate = useSelector(getTranslation())
```
cc @kchelstowski @awalkowiak

kchelstowski: @alannadolny could you expand on that and show a more detailed example? Because for now I don’t see how you want to get rid of the state argument?

src/js/actions/index.js
drodzewicz: you are right about the fallback to default, the only solution for that would be to add this as an option on the API.
Thanks for catching that.

`'nulll'` is something I chose as an option value to filter Purchase orders that don't have a paymentTerm _( that is where `paymentTerm = null`)_, do you have any better suggestions for that ?

src/js/actions/index.js
drodzewicz: if I return `blankPaymentTermOption` from the api like @kchelstowski suggested, it shouldn't be an issue, but I guess as it is currently it might have been an issue so thanks.

src/js/actions/index.js
kchelstowski: how about doing just `null`? Does it throw any null pointers then? If so, maybe you tell where and we could figure out how to handle that together?
Basically "I'm into" doing it as `null`, but was just wondering why it has to be as string, not directly `null`.

src/js/hooks/list-pages/purchase-order/usePurchaseOrderFilters.jsx
alannadolny: *An example from one of my projects*
If I created a selector like this:
```
export const getReviewsByExerciseId = (id) => (state) =>
  state.reviews.reviews.filter((review) => review.exercise === id);
```
There is no need to pass the `state` argument, because it will be applied automatically. In this example we have to pass the `id` only.
So, it will be used in this way:
```
const reviews = useSelector(getReviewsByExerciseId(id));
```
In the code fragment which I selected for comment in this review, we have to pass the `state` argument. It can be done:
```
const supportedActivities = useSelector(getSupportedActivities());
const buyers = useSelector(getBuyers());
const currentLocation = useSelector(getCurrentLocation());
```

drodzewicz: I quite like the idea, but I can only see it being useful in cases like `translate` where we can replace the whole boilerplate of having to import   `getTranslate` and `translateWithDefaultMessage`

drodzewicz: I think that using multiple `useSelector` in a component is possible so it shouldn't be problem to do something like
```js
const translate = useSelector(getTranslateSelector);
const { locale, buyers } = useSelector((state) => ({ locale: state.session.locale, buyers: state.buyers  }));

```

kchelstowski: yeah, I know this is possible and we currently do it like this, but I was wondering if @alannadolny has a follow-up preposition of his idea to  get rid of passing state and doing it for multiple values.

src/js/actions/index.js
drodzewicz: having it being a string instead of a null value in my opinion is safer because having it as a `null` value can result in unexpected behavior since null value is sometimes treated as an empty string or may not be included at all.

For the current example, when passing a parameter `paymentTerm: null` it sends a following request to the server
```
openboxes/api/purchaseOrdersdestination=destID&max=10&offset=0&order=desc&sort=dateOrdered&paymentTerm
```
which is being received on the backend as
```groovy
params = [
    paymentTerm: "",
    max: 10,
    offset: 0,
    ...
]
```
and later when performing a search in the `paymenTerm` param list on 
https://github.com/openboxes/openboxes/blob/3399abec63f43369d63795b3d9471522004c626c/grails-app/services/org/pih/warehouse/order/OrderService.groovy#L111-L113

the `if(...)` result in `false` even if the option is found because empty string is **falsey**
I could just do 
```groovy
if (params.list("paymentTerm").find{ it.equalsIgnoreCase("")} != null) { 
     isNull("paymentTerm") 
 } 
```
but I feel it just complicates the readability and behaviour of the code

src/js/hooks/list-pages/purchase-order/usePurchaseOrderFilters.jsx
kchelstowski: this is what I was afraid of

src/js/actions/index.js
kchelstowski: thanks for that explanation, this is what I expected! So, I'm fine with that, as I don't see a more reasonable and cleaner solution for that.

src/js/hooks/list-pages/purchase-order/usePurchaseOrderListTableData.jsx
```diff
@@ -35,7 +35,7 @@ const usePurchaseOrderListTableData = (filterParams) => {
         max: `${state.pageSize}`,
         ...sortingParams,
         ...filterParams,
-        paymentTerm: paymentTerm && paymentTerm.map(term => term.id),
+        paymentTerm: paymentTerm && paymentTerm.map(({ id }) => id),
```
kchelstowski, line 38: just FYI:
could be shorten to:
```suggestion
        paymentTerm: paymentTerm.map?.(({ id }) => id),
```

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -65,7 +65,12 @@ class SelectOptionsApiController {
         List<PaymentTerm> paymentTerms = genericApiService.getList(PaymentTerm.class.simpleName, [sort: "name"]).collect {
             [id: it.id, label: it.name, value: it.id ]
         }
-        render([data: paymentTerms] as JSON)
+        Map blankPaymentTermOption = [
+                id: 'null',
```
awalkowiak, line 69: I think it could be empty string

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -65,7 +65,12 @@ class SelectOptionsApiController {
...
+                value: 'null',
```
awalkowiak, line 70: I think it could be empty string

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -65,7 +65,12 @@ class SelectOptionsApiController {
         List<PaymentTerm> paymentTerms = genericApiService.getList(PaymentTerm.class.simpleName, [sort: "name"]).collect {
             [id: it.id, label: it.name, value: it.id ]
         }
-        render([data: paymentTerms] as JSON)
+        Map blankPaymentTermOption = [
```
awalkowiak, line 68: I think we tend to add an empty option on the front-end side. Maybe would it be good to add an empty option to the select component that could accept the configurable label property?

awalkowiak (commented):
Overall LGTM, but I have minor questions

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
kchelstowski: I like this idea, even though this is the only place for now that needs this, this might be very useful for the future.

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
drodzewicz: unfortunately it is quite hard to implement this with an empty string since there are many places in our current filter logic where we check for a ''truthy" value of a filter.
Additionally I think having this option a `'null'` is more explicit and much more self explanatory.

src/js/utils/Select.jsx
```diff
@@ -115,9 +133,26 @@ class Select extends Component {
...
+              if (label) {
+                return label;
+              }
+              // if there are no labels on value item
+              // then try to extract these labels from select options
+              const option = this.options.find(it => it?.id && (it?.id === v?.id));
+              return option?.[labelKey] ?? option.label;
+            })
+            .join(', ');
+        } else {
```
alannadolny, line 151: I prefer doing `return` in `if` statement instead of `else`, but it's just my personal preference

src/js/utils/Select.jsx
```diff
@@ -128,6 +163,29 @@ class Select extends Component {
...
+      }
+
+      if (option.options) {
+        option = { ...option, options: this.mapOptions(this.props.options) };
+      }
+
+      return option;
+    }));
+  }
+
```
alannadolny, line 188: As i can see it's not your code, but maybe return option directly from `if` instead of assigning to option and then return.
I mean:
```js
 mapOptions(values) {
    return (_.map(values, (value) => {
      }

      if (value && this.props.valueKey && !option.value) {
        return { ...value, value: option[this.props.valueKey] };
      }

      if (value && this.props.labelKey && !option.label) {
        return { ...value, label: option[this.props.labelKey] };
      }

      if (option.options) {
        return { ...value, options: this.mapOptions(this.props.options) };
      }

      return value;
    }));
  }
```

src/js/utils/Select.jsx
```diff
@@ -115,9 +133,26 @@ class Select extends Component {
     }
 
     if (showLabelTooltip) {
-      const valueMapped = multi && value ?
-        this.props.value.map(v => v?.[labelKey] ?? v?.label) : [];
-      const valueLabel = multi ? valueMapped.join(', ') : (value.label || value.name);
+      let valueLabel = '';
+      if (value) {
```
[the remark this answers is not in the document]
drodzewicz: I understand your remark and I also prefer to use returns in `if(...)` statement, to short circuit the logic, but this is a little different case since in the end we need to return a html tag.

All I am doing in this code is assigning a  `valueLabel` that later will be used in the return below

https://github.com/openboxes/openboxes/blob/82af9198c5ca243772f846f9de3ff5723493def6/src/js/utils/Select.jsx#L156-L160

src/js/utils/Select.jsx
drodzewicz: yes makes sense, although I would need to reorder those actions

src/js/hooks/list-pages/purchase-order/usePurchaseOrderListTableData.jsx
```diff
@@ -35,8 +35,8 @@ const usePurchaseOrderListTableData = (filterParams) => {
         max: `${state.pageSize}`,
         ...sortingParams,
         ...filterParams,
-        paymentTerm: paymentTerm && paymentTerm.map(term => term.id),
-        status: status && status.map(statusElement => statusElement.value),
+        paymentTerm: paymentTerm.map?.(({ id }) => id),
+        status: status.map?.(statusElement => statusElement.value),
```
kchelstowski, line 39:
```suggestion
        status: status?.map?.(statusElement => statusElement.value),
```

src/js/hooks/list-pages/purchase-order/usePurchaseOrderListTableData.jsx
```diff
@@ -35,8 +35,8 @@ const usePurchaseOrderListTableData = (filterParams) => {
         max: `${state.pageSize}`,
         ...sortingParams,
         ...filterParams,
-        paymentTerm: paymentTerm && paymentTerm.map(term => term.id),
-        status: status && status.map(statusElement => statusElement.value),
+        paymentTerm: paymentTerm.map?.(({ id }) => id),
```
kchelstowski, line 38:
```suggestion
        paymentTerm: paymentTerm?.map?.(({ id }) => id),
```

src/js/utils/Select.jsx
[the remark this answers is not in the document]
kchelstowski: we could always separate "label building" from this function and then just get in the `getTooltipHtml()` already the proper label like:
```javascript
const valueLabel = getValueLabel() ?? '';
```
this way we could inject the "return logic" proposed by Alan, but this comment is just as "a bonus".

src/js/utils/Select.jsx
```diff
@@ -128,6 +163,24 @@ class Select extends Component {
...
+      if (value.options) {
+        return { ...value, options: this.mapOptions(value.options) };
+      }
+      if (value && this.props.labelKey && !value.label) {
+        return { ...value, label: value[this.props.labelKey] };
+      }
+      if (value && this.props.valueKey && !value.value) {
+        return { ...value, value: value[this.props.valueKey] };
+      }
+      return value;
```
kchelstowski, line 180: is there a reason you've changed the order of those statements?

kchelstowski (changes requested):

src/js/utils/Select.jsx
drodzewicz: yep there is,
let me give you a quick example.

Given we have a function with 
``` js
const variable = '';
if (condition A) {
   variable = A;
}
if (condition B) {
   variable = B;
}
if (condition C) {
   variable = C;
}
return variable
```
and if I have conditions **A** = `true` and **C** = `true`, it should return result **C** 

If I want to refactor this function to a _"short-circuit"_ (returning a value once the condition is met)
``` js
if (condition A) {
   return A;
}
if (condition B) {
   return B;
}
if (condition C) {
   return C;
}
```
for the same conditions **A** = `true` and **C** = `true` I get **A**
that's why these `if` statements have to be reordered

kchelstowski: I'm sorry, I haven't noticed that in the previous version we were not returning, but just re-assigning values

awalkowiak (approved):

PR #3956, opened 2023-03-28
OBPIH-5493 Ability to enable/disable autosave feature in configs

alannadolny:


src/js/reducers/sessionReducer.jsx
```diff
@@ -121,6 +123,8 @@ export default function (state = initialState, action) {
         displayDateDefaultValue: _.get(action, 'payload.data.data.displayDateDefaultValue', '-'),
         notificationAutohideDelay: _.get(action, 'payload.data.data.notificationAutohideDelay', 8000),
         browserConnectionTimeout: _.get(action, 'payload.data.data.browserConnectionTimeout', 0),
+        isAutosaveEnabled: _.get(action, 'payload.data.data.isAutosaveEnabled') &&
+          _.includes(_.get(action, 'payload.data.data.supportedActivities'), ActivityCode.AUTOSAVE),
```
kchelstowski, line 127: I'd add a helper function outside a reduce to help readability:
```javascript
const isAutosaveEnabled = (payload) => {
   const { isAutosaveEnabled, supportedActivities } = payload?.data?.data
   return isAutosaveEnabled && supportedActivities.includes(ActivityCode.AUTOSAVE);
}
```

and in the reducer just use:
```javascript
isAutosaveEnabled: isAutosaveEnabled(action.payload)
```

grails-app/conf/Config.groovy
```diff
@@ -1308,6 +1308,9 @@ openboxes.display.date.defaultValue = Constants.DISPLAY_DATE_DEFAULT_VALUE
 // delay is in ms
 openboxes.client.notification.autohide.delay = 8000
 
+// Autosave configuration
+openboxes.autosave.enabled = true
```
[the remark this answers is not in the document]
alannadolny: If I understand it properly, it should be disabled by default for location, not for whole application. `openboxes.autosave.enabled` disabling the autosave option for whole application.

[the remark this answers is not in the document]
kchelstowski: Please, double check that or consult it with anyone. In my opinion reading the ticket's description and remembering what was discussed, in my opinion it was supposed to be disabled by default for whole instance, but I also might be wrong.

grails-app/views/location/list.gsp
```diff
@@ -122,7 +122,7 @@
                                                 </g:else>
 
                                             </td>
-                                            <g:each var="activity" in="${org.pih.warehouse.core.ActivityCode.list()}">
+                                            <g:each var="activity" in="${grailsApplication.config.openboxes.autosave.enabled ? ActivityCode.listWithAutosave() : ActivityCode.list()}">
```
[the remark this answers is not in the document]
alannadolny: For my understanding, if we disable it for openboxes, any location shouldn't have option to use it. No excuses 😅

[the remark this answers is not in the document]
kchelstowski: I know you shouldn't be able then to set it, but my point is, if the instance has it enabled and then you disable it, while a location has this supported activity, if it should be hidden in the location view page (even though it's still set in the database), or it should be visible, but somehow crossed out so that user would understand, that it's disabled, but "physically" the location has it.

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -132,9 +132,9 @@ class LocationApiController extends BaseDomainApiController {
     }
 
     def supportedActivities = {
-        def data = ActivityCode.list().collect { it.name() }
-
-        render ([data:data] as JSON)
+        boolean isAutosaveSupported = grailsApplication.config.openboxes.autosave.enabled
+        def data = isAutosaveSupported ? ActivityCode.listWithAutosave() : ActivityCode.list()
```
awalkowiak, line 136: I still don't like this approach, but I am not sure what would be better. I think we should tech huddle this and make an improvement/tech debt ticket for this to make it more generic and configurable for other activities. What I imagine here would be a configurable supported activities property and a filter that would add an activity to the list if it is globally enabled.

@alannadolny for now let's move this logic to the ActivityCode.list(), so it is checked only once.

cc @jmiranda

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
awalkowiak, line 1: Lets import:
`import org.codehaus.groovy.grails.commons.ConfigurationHolder as CH`

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -93,6 +94,10 @@ enum ActivityCode {
         ]
     }
 
+    static listWithAutosave() {
```
awalkowiak, line 97: Instead of this check if the autosave activity code is enabled in the `list()` as:
```groovy
static list() {
    def activityCodes = [
            MANAGE_INVENTORY,
            ADJUST_INVENTORY,
            APPROVE_ORDER,
            APPROVE_REQUEST,
            PLACE_ORDER,
            PLACE_REQUEST,
            FULFILL_ORDER,
            FULFILL_REQUEST,
            SEND_STOCK,
            RECEIVE_STOCK,
            CONSUME_STOCK,
            CROSS_DOCKING,
            PUTAWAY_STOCK,
            PICK_STOCK,
            EXTERNAL,
            ENABLE_NOTIFICATIONS,
            PACK_SHIPMENT,
            PARTIAL_RECEIVING,
            REQUIRE_ACCOUNTING,
            ENABLE_CENTRAL_PURCHASING,
            HOLD_STOCK,
            SUBMIT_REQUEST,
            DYNAMIC_CREATION
    ]

    return CH.config.openboxes.autosave.enabled ? activityCodes + [AUTOSAVE] : activityCodes
}
```

awalkowiak (changes requested):

awalkowiak (approved):

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -91,6 +94,8 @@ enum ActivityCode {
                 SUBMIT_REQUEST,
                 DYNAMIC_CREATION
         ]
+
+        return CH.config.openboxes.client.autosave.enabled ? activityCodes + [AUTOSAVE] : activityCodes
```
jmiranda, line 98: I don't like this. A user should be able to set AUTOSAVE on any location at any time. If we disable autosave globally, that overrides the location setting.

jmiranda (commented):
Everything looks good except the logic in ActivityCode. Unless you can convince me otherwise, I'd like to revert the logic that removes autosave (as an option).

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
awalkowiak: I think you're right, but won't it be confusing to users, that they set the `AUTOSAVE`, but it does not work for them?. 
@alannadolny will you make/prepare this small change (allow always set `AUTOSAVE`, but needs the global config to be working)?

alannadolny: I prepared this small change [here](https://github.com/openboxes/openboxes/pull/3982/files)

PR #3957, opened 2023-03-30
OBPIH-5438 Number format issues in prices when importing PO items

drodzewicz:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -908,23 +905,20 @@ class OrderService {
...
-        // FIXME it's not entirely clear why we reformat strings here like this
-        orderItems.each { orderItem ->
-            if (orderItem.unitOfMeasure) {
-                String[] uomParts = orderItem.unitOfMeasure.split("/")
-                def quantityUom = CSVUtils.parseNumber(uomParts[1], "unitOfMeasure")
-                orderItem.unitOfMeasure = "${uomParts[0]}/${quantityUom}"
-            }
-            orderItem.unitPrice = orderItem.unitPrice ? CSVUtils.parseNumber(orderItem.unitPrice, "unitPrice").setScale(4, RoundingMode.FLOOR).toString() : ''
-        }
-
```
drodzewicz, line 927: The FIXME comment above is right, it is not clear why such an operation was required but this reformatting of **unitPrice** and **unitOfMeasure** has introduced a bug when trying to import OrderItems in different locales.

The reason for the bug is that by using `CSVUtils.parseNumber(...)` it parses the number to english format with a comma which does not work in other locales than english and that caused the imported prices to be wrongly formatted in other locales.

While discussing this issue we came to the conclusion that we won't support a cross language export/import.
We do not expect a user to **export** a file in **english** locale the **switch** to **french** locale and try to import that file.

alannadolny (approved):

kchelstowski (approved):

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -149,4 +150,16 @@ class CSVUtils {
     static CSVPrinter getCSVPrinter() {
         return new CSVPrinter(new StringBuilder(), CSVFormat.DEFAULT)
     }
+
+    /**
+    * Return a CSV separator character based on provided data and expected number of column
```
awalkowiak, line 155: Can you add a comment here about when you expect the `;` as a separator and why.

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -149,4 +150,16 @@ class CSVUtils {
...
+    * */
+    static char getSeparator(String text, Integer columNumber) {
+        def otherSeparator = ';'
```
awalkowiak, line 158: Perhaps you could add a const for this? Beside the `DEFAULT_COLUMN_SEPARATOR`, as a `CUSTOM_COLUMN_SEPARATOR`?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -149,4 +150,19 @@ class CSVUtils {
...
+    /**
+    * Return a CSV separator character based on provided data and expected number of columns.
+    *
+    * Depending on configuration of users preferred csv editor like Excel or Libre office,
+    * when saving a csv file it might reformat the file with a different data separator.
+    * Using this utils function we want to figure out which character is used as a data separator in the file.
+    * */
+    static char getSeparator(String text, Integer columNumber) {
+        String firstLine = text.split("\\r?\\n").first()
+        if (firstLine.contains(Constants.CUSTOM_COLUMN_SEPARATOR) && firstLine.split(Constants.CUSTOM_COLUMN_SEPARATOR).size() == columNumber) {
```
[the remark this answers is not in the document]
drodzewicz: Unfortunately my previous version of `getSeparator()` function would fail in this scenario, so after thinking a bit about it I figured out probably a better solution which involves "sample parsing" the first line and checking if the amount of columns match the expected number

https://github.com/openboxes/openboxes/blob/0179626d25ca5806b4124b8ba39086ed1e96e295/src/groovy/org/pih/warehouse/importer/CSVUtils.groovy#L168-L180

The idea behind this solution is to take the first line as a sample for test, usually the first line is a header columns row.
We check if the row contains "the other" separator character, for example a `;`
if it does then we try to parse this line and generate field keys using `initFieldkEys()` function that CSVMapReader plugin provides us with.
If the amount of generated field keys match the expected amount of columns passed as an argument to the function `columnCount` then we can assume that we found the data separator otherwise we fallback on the default `,`

awalkowiak (approved):

jmiranda (commented):
@awalkowiak I'd like to get a TLDR on this one just so we're on the same page.

awalkowiak:
@drodzewicz can you give a TLDR - a brief summary of these changes?

drodzewicz:
a short TLDR; of this is that we focus on single language workflows.
After discussing this issue with Manon we came to the conclusion that we don't want to handle a scenario where a user exports csv in English then changes locale to French  and tries to import that previously exported template.

An acceptable scenario that we want to support would be user export file in _(example locale 1)_ then imports that file in _(example locale 1)_

PR #3959, opened 2023-03-31
OBPIH-5222 / OBPIH-5560 Import and Export files for products with special characters

alannadolny:
There is solution for two tickets - OBPIH-5222 and OBPIH-5560

OBPIH-5222:
For this ticket I added a BOM to exporting files to allow users on Windows and Mac open files with special characters without problems with displaying them.

OBPIH-5560:
In this PR I fixed importing products with special characters. The main problem was that files from Windows / MacOS have different encodings, so the char codes used in this encodings had different meaning in UTF-8.
To solve this problem I created a function `detectCSVCharset` which takes file as an argument an return its encoding (It have to be a file, because most of the information about encoding is in [file signature](https://profilbaru.com/article/List_of_file_signatures). It is possible to guess the encoding basing only on file content, but it will have lower accuracy). The library which I used for guessing the encoding has two main encoding which we need (Windows-1252, UTF-8), but encoding for MacOS is missing here. ([List of all available encodings](https://github.com/albfernandez/juniversalchardet/blob/main/src/main/java/org/mozilla/universalchardet/Constants.java)). In case when the encoding is not recognized the function for detecting returns `null`. It means that it can be Mac encoding, so I am using this encoding it in this case. This detected encoding is passed as an argument in `getText()` method. Because of that imported products with a different encoding will be saved in database and showed in the import preview correctly.

### DIFF
--- a/grails-app/conf/BuildConfig.groovy
+++ b/grails-app/conf/BuildConfig.groovy
@@ -173,6 +173,7 @@ grails.project.dependency.resolution = {
         compile 'org.xhtmlrenderer:flying-saucer-core:9.1.15'
         compile 'org.xhtmlrenderer:flying-saucer-pdf-openpdf:9.1.15'
         compile 'com.github.librepdf:openpdf:1.2.0'
+        compile 'com.github.albfernandez:juniversalchardet:2.4.0'
     }
     plugins {
 
--- a/grails-app/controllers/org/pih/warehouse/JsonController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/JsonController.groovy
@@ -1862,7 +1862,7 @@ class JsonController {
                 }
 
                 response.setHeader("Content-disposition", "attachment; filename=\"Request-Detail-Report.csv\"")
-                render(contentType: "text/csv", text: CSVUtils.prependBOMToCSVString(sw.toString()), encoding: "UTF-8")
+                render(contentType: "text/csv", text: CSVUtils.prependBomToCsvString(sw.toString()), encoding: "UTF-8")
                 return
             }
             render([aaData: data] as JSON)
--- a/grails-app/controllers/org/pih/warehouse/JsonController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/JsonController.groovy
@@ -1862,7 +1862,7 @@ class JsonController {
                 }
 
                 response.setHeader("Content-disposition", "attachment; filename=\"Request-Detail-Report.csv\"")
-                render(contentType: "text/csv", text: CSVUtils.addBOMToCSVString(sw.toString()), encoding: "UTF-8")
+                render(contentType: "text/csv", text: CSVUtils.prependBOMToCSVString(sw.toString()), encoding: "UTF-8")
                 return
             }
             render([aaData: data] as JSON)
--- a/grails-app/controllers/org/pih/warehouse/JsonController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/JsonController.groovy
@@ -26,6 +26,7 @@ import org.pih.warehouse.core.Person
 import org.pih.warehouse.core.Tag
 import org.pih.warehouse.core.User
 import org.pih.warehouse.core.ValidationCode
+import org.pih.warehouse.importer.CSVUtils
 import org.pih.warehouse.inventory.InventoryItem
 import org.pih.warehouse.inventory.InventoryStatus
 import org.pih.warehouse.inventory.Transaction
@@ -1861,7 +1862,7 @@ class JsonController {
                 }
 
                 response.setHeader("Content-disposition", "attachment; filename=\"Request-Detail-Report.csv\"")
-                render(contentType: "text/csv", text: sw.toString(), encoding: "UTF-8")
+                render(contentType: "text/csv", text: CSVUtils.addBOMToCSVString(sw.toString()), encoding: "UTF-8")
                 return
             }
             render([aaData: data] as JSON)
--- a/src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
+++ b/src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
@@ -179,15 +179,20 @@ class CSVUtils {
         return Constants.DEFAULT_COLUMN_SEPARATOR;
     }
 
-    static String addBOMToCSVString(String csvString) {
+    static String prependBOMToCSVString(String csvString) {
         return '\uFEFF' + csvString
     }
+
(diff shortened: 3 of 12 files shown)


drodzewicz (approved):

grails-app/conf/BuildConfig.groovy
```diff
@@ -173,6 +173,7 @@ grails.project.dependency.resolution = {
         compile 'org.xhtmlrenderer:flying-saucer-core:9.1.15'
         compile 'org.xhtmlrenderer:flying-saucer-pdf-openpdf:9.1.15'
         compile 'com.github.librepdf:openpdf:1.2.0'
+        compile 'com.github.albfernandez:juniversalchardet:2.4.0'
```
mdpearson, line 176: nice find!!

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -149,4 +150,16 @@ class CSVUtils {
     static CSVPrinter getCSVPrinter() {
         return new CSVPrinter(new StringBuilder(), CSVFormat.DEFAULT)
     }
+
+    static String addBOMToCSVString(String csvString) {
```
mdpearson, line 154: nit: what do you think of renaming this method to `prependBOMToCSVString()`?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -149,4 +150,16 @@ class CSVUtils {
...
+        return '\uFEFF' + csvString
+    }
+
+    static String detectCSVCharset(File file) {
```
mdpearson, line 158: In `ProductController`, there are two invocations of this method where, if it returns nothing, a default of `'MacRoman'` is applied. I wonder if that default should be specified here. Also, I imagine that most OB users are _not_ formatting their files on a Mac; would `'WINDOWS-1252'` be a better default?

test/integration/org/pih/warehouse/product/ProductServiceIntegrationTests.groovy
```diff
@@ -327,7 +327,7 @@ class ProductServiceIntegrationTests extends GroovyTestCase {
 
 		// FIXME Export code appends column delimiter for every column (even the last)
 		def expectedHeader = Constants.EXPORT_PRODUCT_COLUMNS.join(",").replace("\n", "") + ","
-		def actualHeader = lines[0]
+		def actualHeader = lines[0].replace("\uFEFF", "")
```
mdpearson, line 330: This line and the one below it makes me wonder if we need/want a method in `CSVUtils` called something like `stripBOMIfPresent()`?

mdpearson (changes requested):
I'm not sure that `'MacRoman'` is the default we want to use -- can you confirm?

I also wonder if we might have missed a place or two where we need to add/remove a BOM, but that can wait for the larger CSV refactor. If this is supporting accented characters in the places we most care about, I say, "many thanks!" -- Also, great find on the encoding-detector tool. Slick!

kchelstowski (approved):

drodzewicz (approved):

grails-app/controllers/org/pih/warehouse/JsonController.groovy
```diff
@@ -1861,7 +1862,7 @@ class JsonController {
                 }
 
                 response.setHeader("Content-disposition", "attachment; filename=\"Request-Detail-Report.csv\"")
-                render(contentType: "text/csv", text: sw.toString(), encoding: "UTF-8")
+                render(contentType: "text/csv", text: CSVUtils.prependBOMToCSVString(sw.toString()), encoding: "UTF-8")
```
jmiranda, line 1865: make this camelcase for readability
prependBomToCsvString

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -177,4 +178,21 @@ class CSVUtils {
         }
         return Constants.DEFAULT_COLUMN_SEPARATOR;
     }
+
+    static String prependBOMToCSVString(String csvString) {
```
jmiranda, line 182: camelCase

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -177,4 +178,21 @@ class CSVUtils {
...
+        return '\uFEFF' + csvString
+    }
+
+    static String stripBOMIfPresent(String csvString) {
```
jmiranda, line 186: camelCase

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -177,4 +178,21 @@ class CSVUtils {
...
+        return csvString.replace("\uFEFF", "")
+    }
+
+
+    static String detectCSVCharset(File file) {
```
jmiranda, line 191: camelCase

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -177,4 +178,21 @@ class CSVUtils {
...
+        def detector = new UniversalDetector(null)
```
jmiranda, line 192: Default constructor should be ok here, no?
https://github.com/albfernandez/juniversalchardet/blob/main/src/main/java/org/mozilla/universalchardet/UniversalDetector.java#L105

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -177,4 +178,21 @@ class CSVUtils {
...
+        byte[] fileBytes = file.bytes
+        detector.handleData(fileBytes, 0, fileBytes.length - 1);
+        detector.dataEnd();
+        return detector.getDetectedCharset() ?: 'MacRoman';
```
jmiranda, line 196: I don't love MacRoman being used as a default here but let's discuss during tech huddle if possible.

jmiranda (commented):
Nice work. Let's discuss the details on the tech huddle tomorrow. But if we need to get this merged, I'm ok with it.

grails-app/controllers/org/pih/warehouse/JsonController.groovy
awalkowiak: I suggested `prependBOMToCSVString` to match the "convention" we had in the `CSVUtils` (-> `CSV` in the class name and another method), but if you want a camel case, then I'm fine with that.

PR #3961, opened 2023-04-03
OBPIH-5437 Add validation for hidden lines while going to next outbound step

kchelstowski:
Added after UAT:

- ability to filter by name (and displayName) of product
- add validation (confirm alert) when trying to go to the next step while having some lines hidden
- focus on filter input when clicking "No" in the confirm alert


kchelstowski:
Marking it as "do not merge" for now, to confirm with Manon if the validation should be included in all three steps where the filter input exists, or only pack page. 
As soon as I will get to know that, I will either add it to remaining steps or remove the label.

alannadolny (approved):

src/js/utils/form-values-utils.jsx
```diff
@@ -52,3 +52,12 @@ export const getReceivingPayloadContainers = formValues =>
       return _.omit(item, 'product.displayNames');
     }),
   }));
+
+/** Expects to have filterValue already in lower case */
```
awalkowiak, line 56: Would not it be clearer if everything was lowered in one place?

awalkowiak (changes requested):

src/js/utils/form-values-utils.jsx
```diff
@@ -52,3 +52,12 @@ export const getReceivingPayloadContainers = formValues =>
...
+export const matchesProductCodeOrName = ({
+  productCode, productName, displayName, filterValue,
+}) =>
+  (productCode?.toLowerCase()?.includes(filterValue) ||
+    productName?.toLowerCase()?.includes(filterValue) ||
+      displayName?.toLowerCase()?.includes(filterValue)
+  );
```
drodzewicz, line 63: Do you think that maybe we could just pass a single **product** object and extract those values inside of this function ?

```suggestion
export const matchesProductCodeOrName = ({ product, filterValue }) => {
const productCode = product?.productCode;
const productName = product?.name;
const displayName = product?.displayName ?? product?.displayName?.default;

return (productCode?.toLowerCase()?.includes(filterValue) ||
    productName?.toLowerCase()?.includes(filterValue) ||
    displayName?.toLowerCase()?.includes(filterValue)
  );
}  
```

src/js/utils/form-values-utils.jsx
kchelstowski: When I look at it now, you are right - I just wanted to avoid writing three times `filterValue.toLowerCase()` here, but I will just create a new const on the top of the function which will make `filterValue.toLowerCase()`.
Thanks for the advice

src/js/utils/form-values-utils.jsx
kchelstowski: sure I can do it, I just used the fact, that I had access directly to the `productCode` and `productName` from the DTO, not having to access it via `product.productCode`, but it doesn't matter and it's cleaner doing it the way you suggested. Thanks

awalkowiak (approved):

PR #3962, opened 2023-04-03
OBPIH-5538 Fix requested by select at create request page

kchelstowski:


kchelstowski:
> I understand this was just a typo, right ?

@drodzewicz yeah, exactly. I knew it would appear somewhere - fortunately it was the only place found by Katarzyna.

PR #3963, opened 2023-04-03
OBPIH-5560 Can't import products with special characters

alannadolny:
In this PR I fixed importing products with special characters. The main problem was that files from Windows / MacOS have different encodings, so the char codes used in this encodings had different meaning in UTF-8.
To solve this problem I created a function `detectCSVCharset` which takes file as an argument an return its encoding (It have to be a file, because most of the information about encoding is in [file signature](https://profilbaru.com/article/List_of_file_signatures). It is possible to guess the encoding basing only on file content, but it will have lower accuracy). The library which I used for guessing the encoding has two main encoding which we need (Windows-1252, UTF-8), but encoding for MacOS is missing here. ([List of all available encodings](https://github.com/albfernandez/juniversalchardet/blob/main/src/main/java/org/mozilla/universalchardet/Constants.java)). In case when the encoding is not recognized the function for detecting returns `null`. It means that it can be Mac encoding, so I am using this encoding it in this case. This detected encoding is passed as an argument in `getText()` method. Because of that imported products with a different encoding will be saved in database and showed in the import preview correctly.

### DIFF
--- a/grails-app/conf/BuildConfig.groovy
+++ b/grails-app/conf/BuildConfig.groovy
@@ -173,6 +173,7 @@ grails.project.dependency.resolution = {
         compile 'org.xhtmlrenderer:flying-saucer-core:9.1.15'
         compile 'org.xhtmlrenderer:flying-saucer-pdf-openpdf:9.1.15'
         compile 'com.github.librepdf:openpdf:1.2.0'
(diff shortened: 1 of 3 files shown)


alannadolny:
Moved to this [PR](https://github.com/openboxes/openboxes/pull/3959)

PR #3964, opened 2023-04-03
OBPIH-5494 Use user attributes in the Hotjar and enable it on react

awalkowiak:


grails-app/views/common/_hotjar.gsp
```diff
@@ -11,5 +11,11 @@
         r.src=t+h._hjSettings.hjid+j+h._hjSettings.hjsv;
         a.appendChild(r);
     })(window,document,'https://static.hotjar.com/c/hotjar-','.js?sv=');
+
+    // Hotjar user attributes
+    // https://help.hotjar.com/hc/en-us/articles/4402892526487-What-are-User-Attributes-
+    // https://help.hotjar.com/hc/en-us/articles/360038394053-How-to-Set-Up-User-Attributes
+    var userId = '${session?.user?.id}' || null;
+    window.hj('identify', userId, {});
```
awalkowiak, line 19: @jmiranda The `{}` is a custom user attribute config. Do we need to track now something else than only the user id? Examples can be seen here: https://help.hotjar.com/hc/en-us/articles/360038394053-How-to-Set-Up-User-Attributes

grails-app/views/common/_hotjar.gsp
```diff
@@ -11,5 +11,11 @@
         r.src=t+h._hjSettings.hjid+j+h._hjSettings.hjsv;
         a.appendChild(r);
     })(window,document,'https://static.hotjar.com/c/hotjar-','.js?sv=');
+
+    // Hotjar user attributes
+    // https://help.hotjar.com/hc/en-us/articles/4402892526487-What-are-User-Attributes-
+    // https://help.hotjar.com/hc/en-us/articles/360038394053-How-to-Set-Up-User-Attributes
+    var userId = '${session?.user?.id}' || null;
```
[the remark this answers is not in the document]
jmiranda: Yes, username seems like it would work better.

grails-app/views/common/_hotjar.gsp
jmiranda: At the moment I don't think we know what we'll need. Let's take it one step at a time though. Once we get the username set we'll be able to do some testing to see what else would be useful.

jmiranda (changes requested):
See comments

PR #3973, opened 2023-04-05
OBPIH-3186 Merge Products feature

awalkowiak:


kchelstowski (approved):
I can't see any "anomaly"

drodzewicz (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -3370,4 +3370,21 @@ class InventoryService implements ApplicationContextAware {
...
+                or {
+                    eq("product", product)
+                    inventoryItem {
+                        eq("product", product)
+                    }
+                }
+            }
+        }
+
+        return transactions?.collect { it.inventory }?.unique()
```
jmiranda, line 3388: Can we not do this as a projection inside the query? Feels like this could be very slow with older products with a long stock history.

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -0,0 +1,516 @@
...
+
+    def inventoryService
+    def invoiceService
+    def productAvailabilityService
+    def requisitionService
+
+    /**
+     * Preforms product swapping for a product pairs passed as params
+     * */
+    def mergeProduct(String primaryId, String obsoleteId) {
```
jmiranda, line 36: At some point, we talked about refactoring this into smaller methods. Since that's probably not going to happen before the merge, we should create a tech debt ticket to do that refactoring in the future.

If it's not too big of a task I would recommend we do it just to make this more readable and maintainable. 
```
validateProducts()
logProductMergeData()
mergeProductSuppliers()
mergeProductComponents()
mergeTransactions()
mergeRequisitions()
mergeShipments()
...
```

Even the validate method could be broken up into smaller methods.

jmiranda (commented):
I'm going to leave this as a Comment (as opposed to Request changes) and let you decide the proper approach (i.e. create tickets for refactoring vs just doing them now).

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
awalkowiak: There's already a comment with the ticket number inside. I'll move it here, for visibility.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -887,4 +888,101 @@ class ProductAvailabilityService {
...
+     * Change product and inventory to primary for rows with given obsolete inventory item
+     * */
+    void updateProductAvailabilityOnMergeProduct(InventoryItem primaryInventoryItem, InventoryItem obsoleteInventoryItem, Product primaryProduct, Product obsoleteProduct) {
+        if (!primaryProduct?.id || !primaryInventoryItem?.id || !obsoleteInventoryItem?.id) {
+            return
+        }
+
+        // First update records that won't violate product_availability_uniq_idx (location_id, product_code, lot_number, bin_location_name)
+        String updateStatement = """
+            UPDATE IGNORE product_availability
```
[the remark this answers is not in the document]
awalkowiak: As we discussed on the tech huddle, I added a ticket for that to the backlog (OBPIH-5608)

PR #3978, opened 2023-04-06
OBPIH-5544 if dynamic creation disabled for all location types add location option should not appear in outbound

drodzewicz:


src/js/components/stock-movement-wizard/modals/AddDestinationModal.jsx
```diff
@@ -67,64 +68,50 @@ const FIELDS = {
...
-          return { ...locationType, label: this.props.locale === 'fr' && fr ? fr : en };
-        });
-
-        this.setState({ locationTypes });
-      });
+  mapToOptionsList(list) {
+    return _.map(list, (locationType) => {
+      const [en, fr] = _.split(locationType.name, '|fr:');
+      return { ...locationType, label: this.props.locale === 'fr' && fr ? fr : en };
+    });
```
kchelstowski, line 114: there is a util function to handle that:
https://github.com/openboxes/openboxes/blob/678dba1019dd668f9f6ec0db8684af51216f0b06/src/js/utils/translation-utils.jsx#L3-L8

kchelstowski (changes requested):

kchelstowski (approved):

awalkowiak (approved):

PR #3980, opened 2023-04-07
OBPIH-5583 Fix stocklist csv export bug when stocklist name has quotations in the string

drodzewicz:


drodzewicz:
This PR fixes a bug that occurred while exporting a stocklist csv file which had stocklist names inside of quotations.
This problem seem to occur only in libre office, other editors like _google sheets_ and _excel_ must have been smarter and knew how to deal with such data.

An example for this bug was when we have a stocklist name which contains commas like: _"one, two, three"_ it generates such columns like
```
Product code, Product description, Category, "one, two thee" [Published], other stocklist name [Published], etc...
```

notice that only column with stocklist name that had commas has quotations, because we do
```
StringEscapeUtils.escapeCsv(requisition?.name)
``` 
So because those quotations are not spread across the whole column name, libre office had some trouble parsing this csv.
So a solution to that was to include the whole column name string inside of `StringEscapeUtils.escapeCsv(...)` which resulted in generated columns like
```
Product code, Product description, Category, "one, two thee [Published]", other stocklist name [Published], etc...
```

alannadolny (approved):

kchelstowski (approved):

awalkowiak (approved):

PR #3981, opened 2023-04-07
OBPIH-5546 Validate for duplicate location names when creating via outbound

alannadolny:
Validation for duplicate names works:

   -  location names should be unique within the same parent
   
   - location names should be unique when there is no parent on that location - in this case parent will be null, so I am also looking for the same parent, but with `null`.
   
I will mark this PR as a draft, because I have problem with validation messages on react side - they are low readable. All of the previous validation messages are low readable also.
For example:
Screenshot from 2023-04-07 16-49-36
If anyone have idea how to make it more readable, I am open for suggestions. Apart from that, I think all of the work is done.


kchelstowski:
@alannadolny I investigated the issue with the "ugly message" you mentioned and feels like there are two separate problems and two separate solutions:
1st problem: when we **throw** the error in the controller, like you have now in the `LocationApiController`: https://github.com/openboxes/openboxes/blob/a90b94b704c61cebd966897bb22c14078321d164/grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy#L140-L160

we are "redirected" to the `ErrorsController` to the `handleValidationErrors`:
https://github.com/openboxes/openboxes/blob/a90b94b704c61cebd966897bb22c14078321d164/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy#L96-L110

and the reason why the error is returned in an ugly way is that this part:
```groovy
messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), localizationService.getCurrentLocale())
```
works incorrectly. I don't know what was the purpose of that `messageService.getMessage`, so maybe either @jmiranda or @awalkowiak might know, but feels like calling `it.codes[0]` makes no sense, since those "codes" are something like this:
Screenshot from 2023-04-07 20-54-15

What works is changing this to:
```groovy
g.message(error: it, locale: localizationService.currentLocale)
```
so we pass an Error instance the `g.message` and Grails handles it properly and returns a beautiful error for us:
Screenshot from 2023-04-07 20-54-02


2nd problem:
In the `apiClient` we display errorMessage vs errorMessages in a reversed way than we should probably:
https://github.com/openboxes/openboxes/blob/a90b94b704c61cebd966897bb22c14078321d164/src/js/utils/apiClient.jsx#L56-L66
and since the "ugly" error has also its source in the `errorMessage` in this case:
Screenshot from 2023-04-07 21-06-56

it has higher priority than the `errorMessages`, so we have to revert the order of `errorMessage || errorMessages` to `errorMessages || errorMessage`.

By doing both, we might also fix/make prettier other validation errors throughout the app, which is fascinating!



A second solution would be to move the logic of `.validate()` to the **service** and in the service throw the `ValidationException` and catch it in the controller like this:
```groovy
} catch (ValidationException e) {
            List<String> errorMessages = e.errors.allErrors.collect {
                g.message(error: it, locale: localizationService.currentLocale)
            }
            response.status = 400
            render([ errorMessages: errorMessages ] as JSON)
        }
```

but on the other hand I think it's better to risk and fix the `ErrorsController` instead of having to do try/catch around the controllers.
We definetely need to test other `ValidationExceptions` thrown on the React side with/without my suggestion to see, if we have only improved the error messages or if we had maybe broken something.
I will push suggested changes in a commit below.

kchelstowski:
@alannadolny by the way 26 integration tests have failed. Could you investigate them? They are probably connected to the validator, not my changes, because they were failed even before my commit. If you hit the wall, let us know, we can try to help you.

kchelstowski:
@awalkowiak @jmiranda If we decide to merge this, please, do NOT use squash & merge here in case the error controller refactor could cause an issue, so we can just easily revert that commit, but I'm pretty confident it won't be necessary.

drodzewicz:
@kchelstowski just to quickly address the first issue you have mentions with error handler using only first message code. I have run in to this problem not too long ago, and there was a little discussion about it https://github.com/openboxes/openboxes/pull/3857#discussion_r1112603162

kchelstowski:
@drodzewicz thanks, I had a feeling that you've run into the same issue not a long time ago, but I was too lazy to find it out. So having you and Alan running into the same issue, I think we should not hesitate to refactor the `ErrorsController`. I have a feeling that we can fix more errors throughout the project than destroy with my change.

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -66,7 +66,13 @@ class Location implements Comparable<Location>, java.io.Serializable {
     static mappedBy = [locations: "parentLocation"]
 
     static constraints = {
-        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation')
+        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation', validator: { name, obj ->
+            Boolean isLocationDuplicated = obj.parentLocation ?
+                    Location.findAllByNameAndParentLocation(name, obj.parentLocation).size() > 1 :
+                    Location.findAllByNameAndParentLocationIsNull(name).size() > 1
+
+            return isLocationDuplicated ? ['duplication.error'] : true
```
jmiranda, line 74: Let's use the same error code as the standard unique constraint error. 
```
default.not.unique.message=Property [{0}] of class [{1}] with value [{2}] must be unique
```
image

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -66,7 +66,13 @@ class Location implements Comparable<Location>, java.io.Serializable {
     static mappedBy = [locations: "parentLocation"]
 
     static constraints = {
-        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation')
+        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation', validator: { name, obj ->
+            Boolean isLocationDuplicated = obj.parentLocation ?
+                    Location.findAllByNameAndParentLocation(name, obj.parentLocation).size() > 1 :
+                    Location.findAllByNameAndParentLocationIsNull(name).size() > 1
```
jmiranda, line 72: Since our validators are getting complicated we should move them to a static method, closure, or named query that can be passed in as a validator. 

Something like this
```
    static constraints = {
        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation', validator: uniqueNameValidator)
        ...
    }

    static uniqueNameValidator = { String value, Location obj ->
        Boolean isNameDuplicated = obj.parentLocation ?
                    Location.findAllByNameAndParentLocation(name, obj.parentLocation).size() > 1 :
                    Location.findAllByNameAndParentLocationIsNull(name).size() > 1

            // not sure if validator.unique is correct so check the docs to see what the default unique validation error code is 
            return isNameDuplicated ? ['validator.unique'] : true
    }
```
NOTE: we should do the same for the other two properties that use the custom validator, but leave that for later. We can use this as an opportunity to create a reference implementation.

test/integration/testutils/DbHelper.groovy
```diff
@@ -29,11 +29,16 @@ class DbHelper {
...
+        Location location = Location.findByName(name) ?: new Location(
             locationType: locationType ?: LocationType.get(Constants.WAREHOUSE_LOCATION_TYPE_ID),
             name: name,
             organization: findOrCreateOrganization(organizationName ?: "DbHelper's Dummy Organization")
-        ).save(failOnError: true, flush: true)
+        )
+        if(location.validate() && location.hasErrors()) {
+            println "TEST1"
+            println location?.errors
+        }
```
jmiranda, line 40: Remember to remove this.

jmiranda: In fact, let's revert this whole method once you're done testing.

test/integration/testutils/DbHelper.groovy
```diff
@@ -70,12 +75,19 @@ class DbHelper {
...
             productType: findOrCreateProductType("Default"),
-        ).save(failOnError: true, flush: true)
+        )
+
+        if(product.validate() && product.hasErrors()) {
+            println "TEST1"
+            println product?.errors
+        }
+
+        return product.save(failOnError: true, flush: true)
```
jmiranda, line 90: Same. Revert this method when you're done testing.

test/integration/testutils/DbHelper.groovy
```diff
@@ -129,6 +141,14 @@ class DbHelper {
...
         def inventory = new Inventory(warehouse: location)
         location.inventory = inventory
+        if(inventory.validate() && inventory.hasErrors()) {
+            println "TEST1"
+            println inventory?.errors
+        }
+        if(location.validate() && location.hasErrors()) {
+            println "TEST1"
+            println location?.errors
+        }
```
jmiranda, line 151: And again

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -97,8 +97,8 @@ class ErrorsController {
         if (RequestUtil.isAjax(request)) {
             response.status = 400
             BeanPropertyBindingResult errors = request?.exception?.cause?.errors
-            def errorMessages = errors.allErrors.collect {
-                return messageService.getMessage(it.codes[0], it.arguments, (it.defaultMessage ?: it.codes[0]), localizationService.getCurrentLocale())
+            List<String> errorMessages = errors.allErrors.collect {
+                return g.message(error: it, locale: localizationService.currentLocale)
```
jmiranda, line 101: I think we discussed this before and I don't think we should be changing this right now unless we know for sure this is not going to break something.

jmiranda: If you're running into an issue please raise it in the ticket and mark the ticket with the tech huddle label so we can discuss and try to resolve it.

grails-app/domain/org/pih/warehouse/core/Location.groovy
jmiranda: This is getting pretty complicated, so we should consider writing some unit tests to show all of the different validations. Not something for @alannadolny to do, but maybe @awalkowiak or me @jmiranda.

jmiranda (changes requested):
Good work @alannadolny. Just a few comments and change requests.

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
jmiranda: Oops, just saw the extensive discussion above. I'll review that discussion and provide some feedback tomorrow.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -66,7 +66,10 @@ class Location implements Comparable<Location>, java.io.Serializable {
     static mappedBy = [locations: "parentLocation"]
 
     static constraints = {
-        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation')
+        name(nullable: false, blank: false, maxSize: 255, unique: 'parentLocation', validator: { name, obj ->
+            Boolean isLocationDuplicated = Location.findAllByNameAndParentLocation(name, obj.parentLocation).size() > 1
```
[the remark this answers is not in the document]
jmiranda: I'm nervous about using finders in the constraint. Are we sure this is safe? I can't remember if we do this elsewhere.

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
jmiranda: > so maybe either @jmiranda or @awalkowiak might know, but feels like calling it.codes[0] makes no sense, since those "codes" are something like this

Just to elaborate on this ... it did make sense at a time when we weren't translating error messages because all of those codes should point to the same localization message. So we just needed any one of the codes to display the error message properly. Now that we're translating messages, we need to be more exact with which error code we use.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -95,6 +95,24 @@ class Location implements Comparable<Location>, java.io.Serializable {
...
+        Location.withNewSession {
+            List<Location> otherLocations
+            if (obj.parentLocation) {
+                otherLocations = Location.findAllByNameAndParentLocation(name, obj.parentLocation)
+            } else {
+                otherLocations = Location.findAllByNameAndParentLocationIsNull(name)
+            }
+
+            // Exclude edited location from otherLocations (since it is validated withNewSession)
+            if (obj.id) {
```
awalkowiak, line 108: @jmiranda, @alannadolny I had to do this and then check `otherLocations?.size() > 0`, because:
- on create new (no other location with this name existing) `otherLocations.size()` was 0,
- on create new (1 other location with this name existing) `otherLocations.size()` was 1,
- on edit (no other location with this name existing) `otherLocations.size()` was 1,
- on edit (1 other location with this name existing) `otherLocations.size()` was 2,

kchelstowski (approved):

PR #3983, opened 2023-04-11
OBPIH-5545 Add ability to delete all supported activities from a location

kchelstowski:


grails-app/i18n/messages.properties
```diff
@@ -1442,6 +1443,7 @@ location.zoneAssigned.message=You cannot delete zone that is assigned to a bin l
 location.bgColor.label=Background color
 location.fgColor.label=Foreground color
 location.bgColor.invalid.matchingcolor=Colors cannot be the same
+location.supportedActivities.invalid.supportedActivities=You cannot set supported activity NONE in combination with other supported activities
```
kchelstowski, line 1446: I'm waiting for suggestions, as I struggled with figuring out a reasonable validation message for that

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -82,6 +82,10 @@ class LocationController {
             if (selectedOrganization && (selectedOrganization?.id != currentOrganization?.id) && !selectedOrganization?.active) {
                 throw new IllegalArgumentException("The organization ${selectedOrganization?.name} is inactive, you can't assign it to the location")
             }
+            // If none supported activities are chosen, assign "None"
+            if (locationInstance?.id) {
```
kchelstowski, line 86: @jmiranda @awalkowiak this is the point I was trying to make on the tech huddle when I was saying 

> "I'm scared of how to distinguish whether I create or edit a location inside an update method"

because without this `if`, I'd assign `None` to the location while creating it from the scratch, not editing.

Fortunately it was as simple as adding this `if (locationInstance?.id)`

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -167,6 +167,12 @@ class LocationApiController extends BaseDomainApiController {
         if (!existingLocation) {
             throw new IllegalArgumentException("No Location found for location ID ${params.id}")
         }
+        // We only want to assign "None" if we go next from the Details step
+        // Without having this if, we would assign "None" when going from Address step to Details (previous step)
+        // because the jsonObject wouldn't contain supportedActivities
+        if (params.step == "1") {
```
jmiranda, line 173: What is params.step? 

Aside: If you're basing this off of the use of step numbers for the stock movement wizard please know that I made a mistake there. We didn't really have a good understanding of how that should be done so that was admittedly kind of a hack. I would hate to repeat that mistake, so let's not do it here if there's a more explicit way of handling what we need. 

Feels like a good candidate for a tech huddle but in the meantime let's come up with something straightforward/explicit.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -93,6 +93,13 @@ class Location implements Comparable<Location>, java.io.Serializable {
         dateCreated(display: false)
         lastUpdated(display: false)
         sortOrder(nullable: true)
+        supportedActivities(validator: { Set<String> activities, Location location ->
```
jmiranda, line 96: I'm having trouble imagining how this is going to work. Does this handle the case of supported activities inherited from location type? For example, in that case the location's supported activities would be empty but the location type support activities would not be.

src/js/components/locations-configuration/LocationDetails.jsx
```diff
@@ -362,7 +369,7 @@ class LocationDetails extends Component {
         fgColor: values.fgColor || '',
       };
 
-      apiClient.post(locationUrl, payload)
+      apiClient.post(locationUrl, payload, { params: { step: 1 } })
```
jmiranda, line 372: Not a huge fan of this params thing. Should probably be part of the payload, right?

@awalkowiak Any thoughts?

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
jmiranda: Just so it's clear why this bothers me. Imagine someone developing a new client for OpenBoxes. A mobile application or god forbid, an Electron desktop application. They want to allow users to create new locations same as we do within the webapp. So they go to the API docs to see what they need to pass as a payload and they see an example.

> To create a location you'll want to post a request to the Locations API like this:

```
POST /openboxes/api/locations
{
        "name": "My New Location",
        "active": "true",
        "locationNumber": "MNL",
        "organization.id": null,
        "locationType.id": "2",
        "supportedActivities": null
}
```
> Oh and if you want to set the supported activities for this new location to empty, you need to pass a random parameter to the API like this:
```
POST /openboxes/api/locations?step=1
{
        "name": "My New Location",
        "active": "true",
        "locationNumber": "MNL",
        "organization.id": null,
        "locationType.id": "2",
        "supportedActivities": null
}
```
What I'm trying to say is we should be trying to think of our API endpoints from the perspective of other developers. Not what will work fine internally for us.

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
jmiranda: Actually, the logic I'd like to see here would be something like this since it's more maintainable/readable. The same thing happens with your code but you need to think it through to see 
```
if (locationInstance?.id && params.list("supportedActivities").empty) {
    params.supportedActivities = [ActivityCode.NONE.id]
}
```

Please test to make sure empty works there under all scenarios 
* params.supportedActivities not present as a request parameter
* params.supportedActivities is present but set to null or empty (supportedActivities=null, supportedActivities=)
* params.supportedActivities is present and set to None (supportedActivities=NONE)
* params.supportedActivities is present and set to multiple activities

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
jmiranda: I just walked through the Location Configuration wizard, set the required properties, unchecked use default activities, cleared the supported activities, and clicked Next.

image

And the logic for the step seemed to have been ignored. 

image

Here's the request as a cURL just so you can see what is being passed to the API by our react API client.

```
curl 'https://openboxes.ngrok.io/openboxes/api/locations?useDefaultActivities=false&step=1' \
  -H 'authority: openboxes.ngrok.io' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'content-type: application/json' \
  -H 'cookie: JSESSIONID=F3279E29D67D0DC79CC152D26033C3C1; __utmc=180537977; __utmz=180537977.1586446259.1.1.utmcsr=(direct)|utmccn=(direct)|utmcmd=(none); JSESSIONID=2E25CB9439F5309BF0EF32D5B0F894B3; G_ENABLED_IDPS=google; uvts=c3af64d0-a29c-4dd6-5f32-6c516e52989b; __utma=180537977.890378389.1586446259.1679791756.1679802048.1740; _ga=GA1.3.890378389.1586446259; _gid=GA1.3.1054352685.1681247566; _gat_gtag_UA_37396941_1=1' \
  -H 'origin: https://openboxes.ngrok.io' \
  -H 'referer: https://openboxes.ngrok.io/openboxes/locationsConfiguration/create' \
  -H 'sec-ch-ua: "Chromium";v="112", "Google Chrome";v="112", "Not:A-Brand";v="99"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Linux"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-origin' \
  -H 'user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36' \
  --data-raw '{"name":"Test Location","active":true,"locationNumber":"TEST","organization.id":"5097","locationGroup.id":"","manager.id":"","locationType.id":"2","supportedActivities":[],"bgColor":"","fgColor":""}' \
  --compressed
```

jmiranda: But regardless, what I think we should be doing is figuring out the most straightforward way to make the API do something explicit (i.e. override supported activities to be [None]) and require the UI and/or API client to figure out how to evaluate the current state, figure out whether conditions have been met and then pass that explicit requirement back to the API (make this ).

Note, that I say UI "or" API client. It might be that the API client can infer what we need from the current state without changing the UI. If not, then we need to consider changing the UI so the user is explicitly stating ... "i want this supported activities thing to be empty". It feels like we should be able to infer, but I haven't played around with the Location Configuration enough yet.

kchelstowski: > I just walked through the Location Configuration wizard, set the required properties, unchecked use default activities, cleared the supported activities, and clicked Next.

> And the logic for the step seemed to have been ignored.

That's weird and shouldn't have happened (should have None). Thanks for catching that and I'll double check that.

grails-app/domain/org/pih/warehouse/core/Location.groovy
kchelstowski: If you "inherit" from the location type, your (location's) supported activities are empty.
This validator just checks if you try to add `None` together with any other activities, so it blocks having e.g.: `[None, Submit request]`, but allows: `[None]` or any combination without none, so: `[Submit request, Manage Inventory, Hold Stock]`

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
kchelstowski: According to the `params.step`, this is a first idea that came to my mind for handling this and also had in mind that it is used throughout the project. I didn't know you didn't like it and you would like to avoid that. Thanks for the explanation why it is wrong. I also have an idea how to handle that on the frontend side, but I'll probably have to write a bit more code.

To get you on the same page - there was a problem while going back to the details step from the address step, because when clicking "Previous" we are also saving the data, and because on the Address step we don't set the supportedActivities, it is not sent to the API while going back to the previous step, so the API overrode my supportedactivities to `["None"]` while doing so.

kchelstowski: @jmiranda but I can see that you've tested it on an environment where there are not my changes yet? Or am I mistaken?

kchelstowski: never mind, I haven't noticed the branch, I just saw: `https://openboxes.ngrok.io/openboxes/api/locations?useDefaultActivities=false&step=1` so I immedietaly assumed that there is not my branch deployed

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
kchelstowski: It was not working for the case when the `params.supportedActivities = ""`, because it was creating: `[""]`, so it was not empty anymore. A workaround for this was to just make additional `.findAll{ it }` to filter out falsy values from the array, which `""` is.
If the `params.supportedActivities` is `null`, it was returning `[]`, so it's fine.
For the rest, so `params.supportedActivities = ["NONE"]` etc. it was obviously working fine.

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -192,6 +192,11 @@ class LocationApiController extends BaseDomainApiController {
     }
 
     Location bindLocationData(Location location, JSONObject jsonObject) {
+        // We only want to set the supportedActivities if we provide it in the payload
+        if (jsonObject.containsKey("supportedActivities")) {
+            jsonObject.supportedActivities = jsonObject.supportedActivities ?: [ActivityCode.NONE.id]
+        }
```
kchelstowski, line 198: @jmiranda I've used this instead of adding additional `step` param to handle the edge case I've mentioned earlier about going "from the previous step".

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -121,6 +126,8 @@ class LocationController {
                     return
                 }
             } else {
+                // Refresh to avoid saving binded, not validated data to the persisted object
+                locationInstance.refresh()
```
kchelstowski, line 130: If you don't understand a purpose of it, I can explain it, but let me know before, as it might be a longer comment/discussion, but simply:
```groovy
locationInstance.properties = params
```
don't validate the data, so it assumes that they are correct.
If we do this at the beginning:
```groovy
def locationInstance = inventoryService.getLocation(params.id)
```
this `Location` object is already persisted in the Hibernate session, so whenever we change any of properties of a persisted object, we don't have to explicity do `save()` after, because the data is saved after the session ends.
What happened was that even though I was getting a `ValidationException` having `["NONE", "SUBMIT_REQUEST"]`, it got saved due to that, so before the session ends, I want to refresh the instance to its "old" version if any validation exception occurs.

kchelstowski (commented):
FYI @jmiranda @awalkowiak

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -82,6 +82,11 @@ class LocationController {
             if (selectedOrganization && (selectedOrganization?.id != currentOrganization?.id) && !selectedOrganization?.active) {
                 throw new IllegalArgumentException("The organization ${selectedOrganization?.name} is inactive, you can't assign it to the location")
             }
+            // If none supported activities are chosen, assign "None"
+            // [""].empty would evaluate to false, so we want to filter out falsy values with findAll{it}
+            if (locationInstance?.id && params.list("supportedActivities").findAll{ it }?.empty) {
```
jmiranda, line 87: Let's move this above 
```
def supportedActivities = params.list("supportedActivities").findAll{ it }
if (locationInstance?.id && supportedActivities?.empty) {
    params.supportedActivities = [ActivityCode.NONE.id]
}
```

grails-app/i18n/messages.properties
jmiranda: Manon says this is good.

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
jmiranda:
```
Location.groovy

Boolean isSupportedActivitiesEmpty() { 
   return supportedActivities.empty || (supportedActivities.contains(ActivityCode.NONE) && supportedActivities.size() == 1)) 
```

kchelstowski:
@jmiranda I've moved the logic of `params.list("supportedActivities")` to a variable above, as you suggested. Let me know if there is anything more I should fix. I'm gonna be OOO since Friday, so in ideal scenario I would want to have some space tomorrow to cover any QA fixes.

awalkowiak (approved):

PR #3984, opened 2023-04-12
OBPIH-5474 Unable to run on order summary report on new staging

alannadolny:
I tried to put all of the qty calculactions in sql query, but finally it gave me worse executing time than it had previously (Maybe it was worse time for my laptop 😅). For now (on my local machine, about 90 records in report) executing time is reduced from ~1.5s to ~0.5s

### DIFF
--- a/grails-app/services/org/pih/warehouse/report/ReportService.groovy
+++ b/grails-app/services/org/pih/warehouse/report/ReportService.groovy
@@ -17,6 +17,7 @@ import org.apache.http.impl.client.DefaultHttpClient
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Organization
+import org.pih.warehouse.core.SynonymTypeCode
 import org.pih.warehouse.inventory.Inventory
 import org.pih.warehouse.inventory.Transaction
 import org.pih.warehouse.inventory.TransactionEntry
@@ -27,6 +28,7 @@ import org.pih.warehouse.order.OrderItem
 import org.pih.warehouse.product.Category
 import org.pih.warehouse.product.Product
 import org.pih.warehouse.reporting.DateDimension
+import org.pih.warehouse.util.LocalizationUtil
 import org.springframework.context.ApplicationContext
 import org.springframework.context.ApplicationContextAware
 import org.xhtmlrenderer.pdf.ITextRenderer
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -677,15 +699,15 @@ class ReportService implements ApplicationContextAware {
             def qtyOnHand = it.qtyOnHand ? it.qtyOnHand.toInteger() : 0
             def qtyOrderedNotShipped = it.qtyOrderedNotShipped ? it.qtyOrderedNotShipped.toInteger() : 0
             def qtyShippedNotReceived = it.qtyShippedNotReceived ? it.qtyShippedNotReceived : 0
-            Product product = Product.findByProductCode(it.productCode)
+            def displayNameWithLocaleCode = "${it.productName}${it.displayName ? " (${locale?.toUpperCase()}: ${it.displayName})" : ''}"
```
kchelstowski, line 702: For now I don't mind it, but be aware, that if we ever consider changing the `getDisplayNameWithLocaleCode()` method in the `Product` domain, we'd have to cover this place as well.
I think for now it could stay, but if we once again have to do it anywhere else, I'd consider creating a static method in the `Product` domain which would take the arguments: `name`, `displayName` and `locale` and place it below the already existing method, so we'd always remember to change both.

kchelstowski (approved):
other than that LGTM, gj

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -659,15 +661,35 @@ class ReportService implements ApplicationContextAware {
...
                 oos.quantity_ordered_not_shipped as qtyOrderedNotShipped,
                 oos.quantity_shipped_not_received as qtyShippedNotReceived, 
-                ps.quantity_on_hand as qtyOnHand
+                ps.quantity_on_hand as qtyOnHand,
+                product_group.name as productFamilyName,
+                category.name as productCategoryName,
+                (
+                select s.name from synonym s
+                where s.product_id = product.id
+                and s.synonym_type_code = '${SynonymTypeCode.DISPLAY_NAME}'
```
awalkowiak, line 678: Let's keep one convention and pass `SynonymTypeCode.DISPLAY_NAME` and `locale` as query params (same as `locationId`)

awalkowiak (changes requested):

awalkowiak (approved):

PR #3985, opened 2023-04-12
OBPIH-5591 Create generic logic and components for the new feature bars

kchelstowski:
Generic component and logic for the new feature bars.
In the future to add a new one, we only need to add a constant in the `newFeatures.js`:

```javascript
export const NewFeature = {
  ...
  FUTURE_FEATURE: 'FUTURE_FEATURE',
};

export const NewFeatureBarConfigs = {
  ...
  [NewFeature.FUTURE_FEATURE]: {<config here>}
  },

};
```

and then just use generic redux actions to manipulate the visibility by providing the proper name, e.g:
```javascript
showNewFeatureBar(NewFeature.FUTURE_FEATURE)
```


grails-app/i18n/messages.properties
```diff
@@ -4153,3 +4153,8 @@ react.productsList.column.dateCreated.label=Created on
 # React table common
 react.reactTable.pagination.of.label=of
 react.reactTable.pagination.rows.label=rows
+
+# React new features bar
+react.newFeature.autosave.version.beta.label=BETA
+react.newFeature.readMore.label=Read more
+react.newFeature.autosave.title.label=New autosave feature is here!
```
alannadolny, line 4160: This `newFeature` sounds a little bit different for me. Additionally I think that `BETA` label is not only for `autosave`, so maybe we should remove this part of label? My suggestion is something similar to `react.bar.version.beta.label` What do you think @awalkowiak?

src/css/main.scss
```diff
@@ -1382,3 +1382,43 @@ div.search-input:focus-within {
...
+    padding: 4px 8px;
+    color: var(--blue-primary);
+    display: flex;
+    align-items: center;
+    gap: 4px;
+
+    svg {
+      width: 18px;
+      height: 18px;
+    }
```
alannadolny, line 1417: I am not sure if this `svg` styles is not inherited from the `svg` above. If not you can also add a className to it to remove this additional styles.

grails-app/i18n/messages.properties
kchelstowski: yeah, you are right it shouldn't have `.autosave`, it should be `react.newFeature.version.beta.label`.
As for `newFeature` I will leave it for you, I have not managed to come up with anything more reasonable.

src/css/main.scss
kchelstowski: yeah, I just needed it in the `.version-box` at the beginning and then it turned out to be needed also "globally", so I just forgot to remove this part. Thanks.

drodzewicz:
Overall PR looks good but one thing that bothers me is Naming of this component.
**NewFeatureBar** - I don't think this is the right name for this component, I think we should go for something more generic.
Usually such bars/Banners appear on other websites as an alternative to Toast notifications.
These banners are not just for alerting user with a **new feature** they can also be used to display that a _current page_ is getting **deprecated** or maybe  there is an issue that users should be wary of.

So my suggestion is to rename this to something like an **InfoBar** or **InfoBanner**

awalkowiak (approved):

PR #3988, opened 2023-04-13
OBS-1445 Check if order summary table exist before inserting

awalkowiak:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1266,6 +1266,17 @@ class OrderService {
         }
 
         if (statements) {
+            try {
+                // Check if table exists.
+                dataService.executeQuery("SELECT * FROM order_summary_mv LIMIT 1")
```
jmiranda, line 1271: Are there other exceptions that might get thrown here? I guess it doesn't matter since an exception signifies a problem. Just wondering.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1239,7 +1239,7 @@ class OrderService {
             // Drop mv temp table if somehow it still exists
             "DROP TABLE IF EXISTS order_summary_mv_temp;",
             // Create temp mv table from sql view (to shorten the time when MV is unavailable)
-            "CREATE TABLE order_summary_mv_temp AS SELECT * FROM order_summary;",
+            "CREATE TABLE order_summary_mv_temp AS SELECT DISTINCT * FROM order_summary;",
```
jmiranda, line 1242: This feels like an antipattern but it's probably the easiest way to solve the problem for now. We should also be looking for the ways in which we might encounter duplicate records and/or add a unique constraint across all columns.

But since that would likely cause downtime requiring manual intervention this feels like an ok approach for now.

jmiranda, line 1242: This feels like an antipattern but it's probably the easiest way to solve the problem for now. We should also be looking for the ways in which we might encounter duplicate records and/or add a unique constraint across all columns.

But since that would likely cause downtime requiring manual intervention this feels like an ok approach for now.

jmiranda, line 1242: This feels like an antipattern but it's probably the easiest way to solve the problem for now. We should also be looking for the ways in which we might encounter duplicate records and/or add a unique constraint across all columns.

But since that would likely cause downtime requiring manual intervention this feels like an ok approach for now.

jmiranda, line 1242: This feels like an antipattern but it's probably the easiest way to solve the problem for now. We should also be looking for the ways in which we might encounter duplicate records and/or add a unique constraint across all columns.

But since that would likely cause downtime requiring manual intervention this feels like an ok approach for now.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1266,6 +1266,17 @@ class OrderService {
...
 
         if (statements) {
+            try {
+                // Check if table exists.
+                dataService.executeQuery("SELECT * FROM order_summary_mv LIMIT 1")
+            } catch (Exception e) {
+                // UndeclaredThrowableException caused by MySQLSyntaxErrorException is thrown when
+                // the table 'order_summary_mv' doesn't exist. Then just simply refresh entire table
+                log.info "Refreshing order summary for ${orderIds} failed due to: ${e?.cause?.message}. Refreshing entire table now."
+                refreshOrderSummary()
```
jmiranda, line 1276: This scares me a ~bit~ lot. We just went from running a simple sub-second (when working properly) update to refreshing the entire order summary table whenever something goes wrong. You could have 10 different orders being updated at the same time and since there's no synchronization around those updates, this could cause 10 different threads to execute the refresh order summary process. 

I think we might need to pull this out into its own method for readability and wrap it execution (or use synchronized keyword) to prevent multiple threads from executing the code. 
```
def synchronized checkIfOrderSummaryExists() { ... }
```
I'm sure I'm guilty of doing this as well, so perhaps we need to think this through a bit more across the board.

jmiranda: One other way to deal with it would be to schedule the trigger through quartz and see if Quartz's @DisallowConcurrentExecution handles the concurrency issue. Ideally if we had 10 orders being updated simultaneously we could potentially have 10 triggers being scheduled but we'd only want one of them being executed.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
jmiranda: Sorry for the duplicate comments (since deleted). GitHub was not saving the comment, so I had to retry a few times.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
jmiranda: @awalkowiak The comment on this answer (https://stackoverflow.com/a/28880097/136597) makes me doubt that there's a solution to preventing the job from running more than once consecutively. 

> Still looking for a way that doesn't queue up the job, and instead just skips it. – 
[b15](https://stackoverflow.com/users/1736218/b15)
 [Jan 25, 2022 at 19:28](https://stackoverflow.com/questions/28879397/java-quartz-scheduled-job-disallow-concurrent-execution-of-job#comment125259452_28880097)

We can live with that for a bit but we'll need to solve this at some point in the future. Imagine a system with a high volume of purchase orders being created through an integration. In fact, we don't need to imagine. That's basically what PIH is/was using Rapidi to do. Thankfully, it is/was inserting those through the database so we don't touch the Hibernate/Grails layer so there are no triggers for this job.

jmiranda: I still want to confirm that this is, in fact, the behavior.

awalkowiak: @jmiranda Yeah, currently if multiple requests will fail because the table does not exist, then it will trigger the entire refresh multiple times. This makes me wonder if we could not move the "single order refresh" to the job which would also have the disabled concurrency, then maybe there is a chance the entire table would not be refreshed multiple times. I'll investigate this approach next. For now, I am pushing the refactor with `checkIfOrderSummaryExists()` method and triggering the job.

jmiranda (approved):
Looks good. Did this get tested thoroughly? If not we should make sure that it gets tested thoroughly during regression testing. This feels like one of those "small changes" that could require a hotfix in the first week after release.

PR #3994, opened 2023-04-17
OBPIH-5451 Decode g.message special characters used in script tag

drodzewicz:


web-app/js/decode.js
```diff
@@ -0,0 +1,5 @@
+// https://stackoverflow.com/a/34064434
+function htmlDecode(input) {
+  const doc = new DOMParser().parseFromString(input, 'text/html');
+  return doc.documentElement.textContent;
+}
```
drodzewicz, line 5: I wanted to put this utils function inside of the `custom.gsp` layout meta tag but it appears to be that this
```
<meta name="layout" content="custom" />
```
is being executed after all of the scripts which resulted in a case where I wanted to use `htmlDecode` and it threw _"not defined error"_ so I decided to just import this script directly before the script that I will be using this `htmlDecode(...)`

drodzewicz:
I was afraid that there will be more places where we have used `g.message(...)` inside of the `<script>` tag and I would have to hunt for each of them but luckily there were just a few, mainly on the Add items page of Purchase order workflow

alannadolny (approved):

PR #3995, opened 2023-04-17
OBPIH-5591 Add beta bar to outbound add items page with autosave (fixes after QA)

alannadolny:


src/js/reducers/showBarReducer.jsx
```diff
@@ -0,0 +1,24 @@
...
+  ADD_INFO_BAR,
+  CLOSE_INFO_BAR,
+  HIDE_INFO_BAR,
+  SHOW_INFO_BAR,
+} from 'actions/types';
+import { InfoBar } from 'consts/infoBar';
+
+const initialState = Object.keys(InfoBar).reduce((acc, bar) => ({ ...acc, [bar]: false }), {});
+
+export default function showBarReducer(state = initialState, action) {
```
awalkowiak, line 11: We have to sort out the reducers and action types namings and states on both reducers

awalkowiak (changes requested):

awalkowiak (approved):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -401,9 +400,6 @@ class AddItemsPage extends Component {
     // We want to hide the feature bar when unmounting the component
     // not to show it on any other page
     this.props.hideInfoBar(InfoBar.AUTOSAVE);
-    // This event acts like React's componentWillUnmount, but componentWillUnmount works only for
-    // react router redirects, not for "links" (<a href>)
-    window.removeEventListener('beforeunload', this.componentCleanup);
```
drodzewicz, line 406: Since we have split up `InfoBarReducer` into `InfoBarVisibilityReducer` which is responsible for the state of show/hide InfoBar component on every page and `InfoBarReducer` which is responsible for collection of InfoBars, modal state and permanent close state we don't need this work around using `beforeunload` event.

The problem that we previously encountered was because `InfoBarReducer` is persisted in the localStorage so when we use a standard page redirect it does not catch up with the state change and state inside of the localStorage stays unchaged. Since we are blakclisting `InfioBarVisibilityReducer` we don't have to worry about this thing anymore

PR #4003, opened 2023-04-25
OBPIH-5622 Fixed syntax error with SQL used to search products by pro…

jmiranda:
…duct type in inventory browser

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -761,12 +761,24 @@ class ProductAvailabilityService {
             // return products with product type with activity searchable no stock OR (searchable AND qoh > 0)
             else {
                 if (!command.showOutOfStockProducts) {
-                    add(Restrictions.disjunction().
-                            add(Restrictions.in("productType", searchableNoStockProductTypes)).
-                            add(Restrictions.conjunction().
-                                    add(Subqueries.lt(0, aggregatedQuantityQuery)).
-                                    add(Restrictions.in("productType", searchableProductTypes))
-                            ))
+
+                    // SUM(product availability.quantity_on_hand) > 0
+                    def quantityGreaterThanZero = Subqueries.lt(0, aggregatedQuantityQuery)
+                    // productType in (:searchableProductTypes)
+                    def inSearchableProductTypes = Restrictions.in("productType", searchableProductTypes)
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -761,12 +761,24 @@ class ProductAvailabilityService {
...
-                            add(Restrictions.conjunction().
-                                    add(Subqueries.lt(0, aggregatedQuantityQuery)).
-                                    add(Restrictions.in("productType", searchableProductTypes))
-                            ))
+
+                    // SUM(product availability.quantity_on_hand) > 0
+                    def quantityGreaterThanZero = Subqueries.lt(0, aggregatedQuantityQuery)
+                    // productType in (:searchableProductTypes)
+                    def inSearchableProductTypes = Restrictions.in("productType", searchableProductTypes)
+                    // productType in (:searchableProductTypes)
```
kchelstowski, line 769:
```suggestion
                    // productType in (:searchableNoStockProductTypes)
```

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -761,12 +761,24 @@ class ProductAvailabilityService {
...
+                    def inSearchableNoStockProductTypes = Restrictions.in("productType", searchableNoStockProductTypes)
+
+                    // Create a disjunction with none, one or both of the searchable product type restrictions
+                    def disjunction = Restrictions.disjunction()
+
+                    if (searchableNoStockProductTypes)
+                        disjunction.add(inSearchableNoStockProductTypes)
```
kchelstowski, line 776: I'm wondering how it works in Groovy if you do not add braces to `if` and you add some statements below? 
Does it only read up to one line below or?

kchelstowski (approved):
LGTM

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
drodzewicz: I believe it works just like in `Java` and `C`. If the `if` statement doesn't have braces it only includes the first line in the under the `if(...)`.

For example if we have such a statement
```
if (condition == true)
    println "first print"
    println "second print"
```
Even though based on code formatting it looks like the second print belongs to the `if` statement in reality it is outside of the block.

PR #4005, opened 2023-04-25
OBPIH-5616 Improve calculating product availability for picklist items

awalkowiak:


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -85,6 +85,31 @@ class ProductAvailabilityService {
         }
     }
 
+    def refreshProductsAvailability(String locationId, List<String> productIds, List<String> binLocationIds, Boolean forceRefresh) {
+        if (locationId && productIds && !binLocationIds) {
+            refreshProductsAvailability(locationId, productIds, forceRefresh)
+            return
+        } else if (!locationId || !productIds) {
```
alannadolny, line 92: If there is a `return` statement, there is no need to use `else`

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1733,6 +1739,42 @@ class InventoryService implements ApplicationContextAware {
...
+                }
+                // Or get all transactions for that bin location (with given product)
+                or {
+                    if (binLocations) {
+                        inList("binLocation", binLocations)
+                    }
+                    if (includeDefaultBinLocation) {
+                        isNull("binLocation")
+                    }
+                }
```
kchelstowski, line 1771: correct me if I'm wrong, but if you anyway do `or` here, to find by `isNull("binLocation")`, was it necessary above (lines 1743-1745) to filter `null` out?
If you hadn't done so, I think you could use here just:
```groovy
if (binLocations) {
  inList("binLocation", binLocations)
}
```
unless `inList` won't match `null`, then I understand everything.

alannadolny (changes requested):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
awalkowiak: @kchelstowski iirc `null` cases won't work with `inList` and `eq` in cases like that, and it requires `isNull`

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: changed

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1100,7 +1102,7 @@ class StockMovementService {
             // Save requisition item before PA refresh
             requisitionItem.save(flush: true)
 
-            productAvailabilityService.refreshProductsAvailability(requisitionItem?.requisition?.origin?.id, [requisitionItem?.product?.id], false)
+            productAvailabilityService.refreshProductsAvailability(requisitionItem?.requisition?.origin?.id, [requisitionItem?.product?.id], binLocations?.unique(), false)
```
drodzewicz, line 1105: I am wondering if the _binLocations_ List shouldn't be a Set instead. That way it would eliminate the need for performing a `binLocations?.unique()`, but I am unsure about the performance implications of the `Set` vs `List`. I see some people mentioning on [stack overflow](https://stackoverflow.com/a/10799483) that HashSets are not as well optimised while [other](https://www.baeldung.com/java-hashset-arraylist-contains-performance) say that it is preferable in this kind of a situation

alannadolny (approved):

awalkowiak:
(converted back to draft, because there will be one more improvement included)

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,33 +195,77 @@ class ProductAvailabilityService {
...
+                    LEFT JOIN inventory_item ii ON pli.inventory_item_id = ii.id
+                    LEFT JOIN location l ON pli.bin_location_id = l.id
+                    LEFT JOIN (
+                        SELECT
+                            GROUP_CONCAT(lsa_select.supported_activities_string) as activities,
+                            lsa_select.location_id as location_id
+                        FROM location_supported_activities lsa_select
+                        GROUP BY lsa_select.location_id
+                    ) lsa ON lsa.location_id = l.id
+                WHERE (r.origin_id = :locationId AND r.status IN (${RequisitionStatus.listPending().collect { "'$it'" }.join(',')}))
```
jmiranda, line 238: Why can't you pass this in as a query parameter?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,33 +195,77 @@ class ProductAvailabilityService {
...
-        """, [
-                location                    : location,
-                pendingRequisitionStatus    : RequisitionStatus.listPending(),
-                pendingOrderStatus          : OrderStatus.listPending(),
-                product                     : product?.id ?: ''
-        ], [readOnly: true])
-
-        return results.collect { [binLocation: it[0]?.id, inventoryItem: it[1]?.id, quantityAllocated: it[2]] }
+        def query = """
+            SELECT bin_location_id, inventory_item_id, SUM(quantity) FROM (
```
jmiranda, line 220: As I mentioned on the call the other day, we should include product and location in the select even though they are passed in as query parameters. There's a small chance that might lead to slower performance if those columns aren't in a covering index but I don't suspect that's the case. Also, the inner (raw query) should not include the sum / group by if we can help it. It's usually better to leave the group by until the end i.e. outer query. I could be wrong about this, but it feels wrong to apply aggregation inside a nested query if you're going to do the same on the outside. 

Also I see we're getting rid of the weird count + count / count(*). I think that's safe since we're simplifying the query by breaking into two separate queries and unioning the results. But we should try to understand and/or compare results from the previous query so we know we're getting the same results.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,33 +195,77 @@ class ProductAvailabilityService {
...
+                SELECT
+                    pli.bin_location_id as bin_location_id,
+                    pli.inventory_item_id as inventory_item_id,
+                    sum(pli.quantity) as quantity
+                FROM picklist_item pli
+                    INNER JOIN picklist p ON pli.picklist_id = p.id
+                    LEFT JOIN requisition_item ri ON pli.requisition_item_id = ri.id
```
jmiranda, line 227: In my opinion, we don't need to use LEFT JOIN any longer. That was one of the reasons why I wanted to do a UNION here. I think location (bin location) is the only table that needs to be LEFT JOIN'd.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,33 +195,77 @@ class ProductAvailabilityService {
...
+                    LEFT JOIN requisition r ON p.requisition_id = r.id
+                    LEFT JOIN inventory_item ii ON pli.inventory_item_id = ii.id
+                    LEFT JOIN location l ON pli.bin_location_id = l.id
+                    LEFT JOIN (
+                        SELECT
+                            GROUP_CONCAT(lsa_select.supported_activities_string) as activities,
+                            lsa_select.location_id as location_id
+                        FROM location_supported_activities lsa_select
```
jmiranda, line 235: Remember we want to check both location and location type. Location supported overrides location type supported activities. If that's too difficult, then we can save that for another time.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -155,33 +195,77 @@ class ProductAvailabilityService {
...
+                    LEFT JOIN inventory_item ii ON pli.inventory_item_id = ii.id
+                WHERE (o.origin_id = :locationId AND o.status IN (${OrderStatus.listPending().collect { "'$it'" }.join(',')}))
+                  AND (oi.id IS NOT NULL)
+                  AND (:productId = '' OR oi.product_id = :productId)
+                GROUP BY pli.bin_location_id, pli.inventory_item_id
+            ) as requisition_order_union
+            GROUP BY bin_location_id, inventory_item_id;
+        """
+
+        def results = dataService.executeQuery(
```
jmiranda, line 260: If you can't figure out the issue where, try using Hibernate to execute the query instead of the Groovy SQL client we're using. 

https://dzone.com/articles/grails-goodness-using

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: I think that's a fair assessment. The performance implications between set / list.unique will be negligible compared to the time spent doing anything with the database so it would probably be better to handle this properly with a set.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2117,13 +2120,14 @@ class StockMovementService {
             picklistItem.disableRefresh = Boolean.TRUE
             picklistItem.picklist?.removeFromPicklistItems(picklistItem)
             picklistItem.requisitionItem?.removeFromPicklistItems(picklistItem)
+            binLocations.add(picklistItems?.binLocation?.id)
```
jmiranda, line 2123: Should this be `picklistItem?.binLocation?.id`

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -74,12 +75,13 @@ class PicklistService {
                 it.disableRefresh = Boolean.TRUE
                 picklist.removeFromPicklistItems(it)
                 orderItem.removeFromPicklistItems(it)
+                binLocations.add(it.binLocation?.id)
```
jmiranda, line 78: Since we're not doing any filtering (like in cases above) we could probably just collect all of the bin locations before the findAll.
```
def binLocationIds = picklist?.picklistItems?.collect { it?.binLocation?.id }
```

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: This was causing the following stacktrace
```
2023-05-11 15:03:36,185 [http-8080-4] INFO  filters.AccessLogFilters  - stockMovementApi.reviseItems: [user:jmiranda, location:Distribution Center]
2023-05-11 15:03:36,193 [http-8080-4] INFO  filters.AccessLogFilters  - No rule for stockMovementApi:reviseItems -> allow anonymous
2023-05-11 15:03:36,198 [http-8080-4] INFO  core.UserService  - Is role ROLE_FINANCE in [ROLE_MANAGER, ROLE_AUTHENTICATED, ROLE_BROWSER, ROLE_ADMIN, ROLE_SUPERUSER, ROLE_ASSISTANT] = false
2023-05-11 15:03:36,198 [http-8080-4] INFO  core.UserService  - Is role ROLE_SUPERUSER in [ROLE_MANAGER, ROLE_AUTHENTICATED, ROLE_BROWSER, ROLE_ADMIN, ROLE_SUPERUSER, ROLE_ASSISTANT] = true
2023-05-11 15:03:36,266 [http-8080-4] INFO  inventory.StockMovementService  - Shipment workflow Stock movement shipment workflow
2023-05-11 15:03:36,385 [http-8080-4] INFO  api.StockMovementApiController  - binding lineItems: [[id:ff808181880c4df001880c53118d0003, reasonCode:LOW_STOCK, quantityRevised:1]]
2023-05-11 15:03:36,399 [http-8080-4] INFO  inventory.StockMovementService  - Removing previous changes, picklists and shipments, if present
2023-05-11 15:03:36,527 [http-8080-4] INFO  inventory.ProductAvailabilityService  - refreshProductsAvailability: 8a8a9e9665c4f59d0165c54ec6b10027, [10001], [[8a8a9e96687c94ce0168bab1c5a8285c]], false
2023-05-11 15:03:36,532 [pool-2-thread-1] INFO  inventory.RefreshProductAvailabilityEventService  - Application event been disabled by event publisher
2023-05-11 15:03:36,532 [pool-2-thread-1] INFO  inventory.RefreshProductAvailabilityEventService  - Application event [disableRefresh:true, forceRefresh:false, class:class org.pih.warehouse.inventory.RefreshProductAvailabilityEvent, locationId:8a8a9e9665c4f59d0165c54ec6b10027, productIds:[10001], synchronousRequired:false, timestamp:1683835416503, metaClass:org.codehaus.groovy.runtime.HandleMetaClass@4ba20442[groovy.lang.ExpandoMetaClass@4ba20442[class org.pih.warehouse.inventory.RefreshProductAvailabilityEvent]], source:org.pih.warehouse.picklist.PicklistItem : ff808181880c4df001880c59fde2000d]
2023-05-11 15:03:36,549 [http-8080-4] ERROR errors.GrailsExceptionResolver  - Exception occurred when processing request: [POST] /openboxes/api/stockMovements/ff808181880c4df001880c528d670001/reviseItems
Stacktrace follows:
org.hibernate.TypeMismatchException: Provided id of the wrong type for class org.pih.warehouse.core.Location. Expected: class java.lang.String, got class java.util.ArrayList
	at org.pih.warehouse.inventory.ProductAvailabilityService$_refreshProductsAvailability_closure3.doCall(ProductAvailabilityService.groovy:106)
	at org.pih.warehouse.inventory.ProductAvailabilityService.refreshProductsAvailability(ProductAvailabilityService.groovy:104)
	at org.pih.warehouse.inventory.ProductAvailabilityService$$FastClassByCGLIB$$7ee8fd31.invoke(<generated>)
	at net.sf.cglib.proxy.MethodProxy.invoke(MethodProxy.java:149)
	at org.pih.warehouse.inventory.ProductAvailabilityService$$EnhancerByCGLIB$$80269ed0.refreshProductsAvailability(<generated>)
	at org.pih.warehouse.inventory.ProductAvailabilityService$refreshProductsAvailability.call(Unknown Source)
	at org.pih.warehouse.inventory.StockMovementService.removeShipmentAndPicklistItemsForModifiedRequisitionItem(StockMovementService.groovy:2130)
	at org.pih.warehouse.inventory.StockMovementService$_reviseItems_closure65.doCall(StockMovementService.groovy:1944)
	at org.pih.warehouse.inventory.StockMovementService.reviseItems(StockMovementService.groovy:1934)
	at org.pih.warehouse.inventory.StockMovementService$$FastClassByCGLIB$$b65c0b56.invoke(<generated>)
	at net.sf.cglib.proxy.MethodProxy.invoke(MethodProxy.java:149)
	at org.pih.warehouse.inventory.StockMovementService$$EnhancerByCGLIB$$456057a7.reviseItems(<generated>)
	at org.pih.warehouse.inventory.StockMovementService$reviseItems.call(Unknown Source)
	at org.pih.warehouse.api.StockMovementApiController$_closure12.doCall(StockMovementApiController.groovy:265)
	at org.pih.warehouse.api.StockMovementApiController$_closure12.doCall(StockMovementApiController.groovy)
	at grails.plugin.springcache.web.GrailsFragmentCachingFilter.doFilter(GrailsFragmentCachingFilter.groovy:66)
	at net.sf.ehcache.constructs.web.filter.Filter.doFilter(Filter.java:86)
	at org.grails.plugin.resource.DevModeSanityFilter.doFilter(DevModeSanityFilter.groovy:44)
	at java.lang.Thread.run(Thread.java:748)
```

awalkowiak: @jmiranda yes this should be `picklistItem?.binLocation?.id` (singular item)

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: @jmiranda Because it was not handled (parsed to string) properly for sql to work as expected

jmiranda:
@awalkowiak Please review when you get a chance.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -235,37 +242,50 @@ class ProductAvailabilityService {
...
-        def results = dataService.executeQuery(
-            query,
+        SQLQuery sqlQuery = sessionFactory.currentSession.createSQLQuery(query)
+        List results = sqlQuery.addScalar("bin_location_id", Hibernate.STRING)
+                .addScalar("inventory_item_id", Hibernate.STRING)
+                .addScalar("quantity_allocated", Hibernate.BIG_DECIMAL)
+                .setString("locationId", location?.id)
+                .setString("productId", product?.id ?: '')
+                .setParameterList("pendingRequisitionStatuses", RequisitionStatus.listPending().collect { it.name() })
+                .setParameterList("pendingOrderStatuses", OrderStatus.listPending().collect { it.name() })
```
kchelstowski, line 277: LGTM, just wondering if it would work without `.collect` here, since `toString()` returns `name()`, if Hibernate would convert that properly

alannadolny (approved):

drodzewicz (approved):

awalkowiak:
@jmiranda the code looks good, but I still want to test it out locally. Did you omit the location type activity codes check on purpose or you forgot about this? (I am not 100% sure if this is required)

awalkowiak:
@jmiranda LGTM. I checked the "problematic" (refreshing product availability after generating picklist) part locally. It looks good. I cannot give the approval, since I created the PR, but I think it can be merged.

PR #4009, opened 2023-04-28
OBPIH-5635 Issue with select field on add items to stocklist

alannadolny:


web-app/css/openboxes.css
```diff
@@ -1191,7 +1191,9 @@ button.action-btn, button.previous, button.next {
 /* temporarily hide the nav bar until we figure out whether we want to keep it or not */
 .nav { display: none; }
 
-li.ui-menu-item { text-align: left; font-weight: bold; }
+ul.ui-autocomplete { width: auto !important; }
```
awalkowiak, line 1194: These two `importants` scare me a little bit. Maybe it would be better to add an additional class to it?

awalkowiak (changes requested):

drodzewicz (approved):

awalkowiak (approved):

PR #4010, opened 2023-04-28
OBPIH-5621 Fix circular reference for the location JSON marshaller

kchelstowski:
1. Circular reference for the location JSON marshaller - it was happening if an organization had a defaultLocation and this location's locationTypeCode was NOT: `DEPOT`, `INTERNAL`, `BIN_LOCATION`, `ZONE` - if it was e.g. `SUPPLIER`, we were falling back to `toJson()` method, which was referencing again the organization, and the organization was again referencing the location.

Organization.groovy
```groovy
Map toJson() { 
   return [
      ...
     defaultLocation: Location.toJson(defaultLocation),
     ...
  ]
}
```

`Location.groovy`:
```groovy
static Map toJson(Location location) {
  return location ? location.toJson(location?.locationType?.locationTypeCode) : null
}
```
...
```groovy
Map toJson(LocationTypeCode locationTypeCode) {
  switch(locationTypeCode) {
    ...
    default:
      return toJson()
  }
}
```
...
```groovy
Map toJson() {
        return [
                ...
                organization               : organization,
                ...
        ]
}
```
I also had to remove type checking from `toJson(LocationTypeCode locationTypeCode)` to `toJson(locationTypeCode)`, because if the `locationTypeCode` is `null`, it doesn't even evaluate this method, but immedietaly returns `null`, so it doesn't return the `defaultLocation`, but returns `defaultLocation: null`.
I also removed calling by default: `toJson()`, because it basically caused the circular reference.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/core/Location.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Location.groovy
@@ -335,6 +335,17 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return active ? LocationStatus.ENABLED : LocationStatus.DISABLED
     }
 
+    Map toBaseJson() {
+        return [
+                id: id,
+                name: name,
+                locationNumber: locationNumber,
+                active: active,
+                locationType: locationType,
+                locationTypeCode: locationType?.locationTypeCode?.name(),
+        ]
+    }
+
     Map toJson() {
         return [
                 id                         : id,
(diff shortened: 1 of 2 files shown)


grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -362,41 +362,25 @@ class Location implements Comparable<Location>, java.io.Serializable {
         ]
     }
 
-    Map toJson(LocationTypeCode locationTypeCode) {
+    Map toJson(locationTypeCode) {
+        def json = [
```
jmiranda, line 366: Let's make it a little more reuseable.
```
Map toBaseJson() { 
    return [
        id: id,
        name: name,
        locationNumber: locationNumber,
        active: active,
        locationType: locationType,
        locationTypeCode: locationType?.locationTypeCode?.name()
    ]
}
Map toJson(locationTypeCode) { 
    def json = toBaseJson() 
    switch (locationTypeCode) {
        ...
    }
    return json
}
```

jmiranda (changes requested):

jmiranda (approved):

PR #4013, opened 2023-05-02
OBPIH-5610 Significant performance issues with Export Orders

jmiranda:


kchelstowski:
Marked as "do not merge", since there are some differences visible in table vs csv

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -444,8 +444,8 @@ class InvoiceService {
                 
                     select distinct 
                         `order`.id as order_id,
-                        order_item.id,
                         `order`.order_number as order_number,
+                         order_item.id as order_item_id,
```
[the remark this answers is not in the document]
jmiranda: Thanks for fixing the column order issue.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -407,4 +409,67 @@ class InvoiceService {
...
+                ) as order_item_invoice_union
+                where order_id = :orderId
+                group by order_id, order_number, order_item_id, quantity_ordered
+                having invoiced = 1
+            ) as order_item_invoice_summary
+        """
+        SQLQuery sqlQuery = sessionFactory.currentSession.createSQLQuery(query)
+        final count = sqlQuery.with {
+            setString("orderId", order.id)
+            setString("prepaymentTypeCode", InvoiceTypeCode.PREPAYMENT_INVOICE.toString())
```
jmiranda, line 468: As I mentioned in the ticket, I'm not sure this is correct. If there are requirements for excluding prepayment invoices in the "invoiced" count, then we can move this into the query (we don't need to use query parameters).

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
```diff
@@ -155,7 +156,7 @@ class PurchaseOrderApiController {
                 order?.orderedOrderItems?.size() ?: 0,
                 order?.shippedOrderItems?.size() ?: 0,
                 order?.receivedOrderItems?.size() ?: 0,
-                order?.invoiceItems?.size() ?: 0,
+                invoiceService.countInvoicedOrderItems(order)?:0,
```
awalkowiak, line 159: `order?.invoiceItems?.size()` should not be used here at all in the beginning, because it's taking all invoice items instead of invoiced order items, plus I even left a comment above `order?.invoiceItems` some time ago that it should be used for one order only but not for a list of orders like it was used in this case.

Perhaps it would be better to add all these counts to the OrderSummary? We would have then everything in one place without the need to have additional queries. All `orderedOrderItems`, `shippedOrderItems`, `receivedOrderItems`, and `invoicedOrderItems` could be additionally (easily) calculated in the `helper views` and stored in the OrderSummary.

Refactoring or improving `orderItem?.invoiceItems` and `orderItem?.quantityInvoicedInStandardUom` should also be done definitely but it probably doesn't have to be done for PO export if were getting the item counts from view or the query that you created.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
kchelstowski: I explained in the ticket why it should be excluded and how it is excluded now in this method: https://github.com/openboxes/openboxes/blob/c6bc349696d2a5d05a56dff5def061d02bcaff6e/grails-app/domain/org/pih/warehouse/order/OrderItem.groovy#L344-L354

I've added it as query parameters to follow the convention you did with `orderId` - I didn't wanna mix things up, to have one parameter via query parameters, and one inside the query directly.

awalkowiak: https://github.com/openboxes/openboxes/blob/develop/grails-app/migrations/views/order-summary-helper-views.sql#LL87C43-L87C46 This one here is doing that for OrderSummary (I know it's not intuitive and requires at least a comment), Katarzyna also linked tickets regarding the functionality in the OBPIH-5610.

awalkowiak (commented):
I left two comments in the code. I think we should split these two cases and we should not mix those two:
1. Order.getInvoiceItems
2. Order.getInvoice**d**OrderItems 
because these are two different things. The first one should **not** be used for getting invoice**d** items for PO (but currently this one is used)

awalkowiak:
I am closing this one since it is fixed in another way in this PR https://github.com/openboxes/openboxes/pull/4027

PR #4017, opened 2023-05-05
OBPIH-5641 Invalid content of error message in outbound workflow

kchelstowski:
The reason why it was working before was that we were taking `defaultMessage` > `translatedMessage`, and as you can see, I changed the code of the message, because it was a copy paste from the outbound return validation message.
Why it is appearing just now, is because after my refactor of `ErrorsController`, we now take `translatedMessage` > `defaultMessage`, which is the correct way to handle messages.


alannadolny (approved):

awalkowiak (commented):
lgtm

PR #4018, opened 2023-05-05
OBPIH-5646 Improve the content of reason code required error message

kchelstowski:


drodzewicz (approved):
An ideal solution would be to avoid adding this additional **detailsArray** prop and just pass the stringified html message like we previously had, but since we are on a time constraint before the 
 release  0.8.22 this current solution should work fine.

alannadolny (approved):

awalkowiak (commented):
lgtm

PR #4019, opened 2023-05-05
OBPIH-5468 Removed javascript code that was causing a duplicate request

jmiranda:


grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -232,7 +232,7 @@
                 </div>
                 <div class="right">
                     <g:if test="${!order?.isPlaced()}">
-                        <g:link controller="order" action="placeOrder" id="${order?.id}" class="button validate" >
+                        <g:link controller="order" action="placeOrder" id="${order?.id}" class="button" >
```
awalkowiak, line 235: @jmiranda Could now the `$(".validate").click(...);` handler be removed too? I believe it is not required anymore.

PR #4020, opened 2023-05-05
OBPIH-5639 NullPointerException while updating product availability during product import 

drodzewicz:
Add guard condition for updateProductAvailability and updaeInventorySnapshots
prevent update query from executing if productCode on event.product object is null

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
@@ -410,7 +410,7 @@ class InventorySnapshotService {
     }
 
     void updateInventorySnapshots(Product product) {
-        if (!product?.id) {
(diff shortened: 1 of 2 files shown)


jmiranda:
@drodzewicz My hesitation with merging this PR was due to the fact that I knew @mdpearson had created a PR (https://github.com/openboxes/openboxes/pull/3449) that changed these lines a few months ago. I thought the change was never merged, but it looks like we did merge it.  

So for now, let's revert his change and go back to what it was before unless there's an argument against this. 
```
if (!product || !product.id || !product.productCode) {
```

jmiranda (changes requested):
See previous comment.

jmiranda (approved):

jmiranda:
Disregard the previous comment about reverting back the change. With the use of the elvis operator, @drodzewicz change is very close to what we had before.

PR #4027, opened 2023-05-10
OBPIH-5610 Improve PO export performance

awalkowiak:


jmiranda (approved):

kchelstowski (commented):
I will review it and test locally tomorrow

grails-app/domain/org/pih/warehouse/order/OrderItemSummary.groovy
```diff
@@ -20,6 +20,12 @@ class OrderItemSummary implements Serializable {
     Integer quantityReceived
     Integer quantityCanceled
     Integer quantityInvoiced
+
+    Boolean itemOrdered = Boolean.FALSE
+    Boolean itemShipped = Boolean.FALSE
+    Boolean itemReceived = Boolean.FALSE
+    Boolean itemInvoiced = Boolean.FALSE
```
kchelstowski, line 27: it's rather a statement, not a change request, if you don't feel it's worth of doing - but when first reading this domain's code yesterday I found it difficult to understand what this `itemOrdered` mean being a boolean - I was "expecting" an object here, so maybe would it be worth to name it as `isItemOrdered` etc? I don't know if we have any convention regarding booleans?

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
```diff
@@ -152,10 +152,10 @@ class PurchaseOrderApiController {
...
                 order?.paymentTerm?.name,
                 lineItemsSize,
-                order?.orderedOrderItems?.size() ?: 0,
-                order?.shippedOrderItems?.size() ?: 0,
-                order?.receivedOrderItems?.size() ?: 0,
-                order?.invoiceItems?.size() ?: 0,
+                it?.itemsOrdered ?: 0,
+                it?.itemsShipped ?: 0,
+                it?.itemsReceived ?: 0,
+                it?.itemsInvoiced ?: 0,
```
kchelstowski, line 158: Just a small suggestion even though it has been there for a long time, but maybe it'd be worth to change in line 117 from:
```groovy
def getOrdersCsv(List purchaseOrders)
```
to:
```groovy
def getOrdersCsv(List<OrderSummary> purchaseOrders)
```
so that the compiler understands what `itemsOrdered` etc is, instead of marking it as grey and saying: "_No candidates found for method call it?.itemsOrdered."_

kchelstowski (approved):
Tested locally - have not found any anomaly - the performance is huge now, good job!

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
awalkowiak: Yup, I want to do this plus stuff like changing `it` to named var, etc

grails-app/domain/org/pih/warehouse/order/OrderItemSummary.groovy
awalkowiak: Yes, I was thinking about changing the name for this and for those from OrderSummary, because `itemOrdered` and `itemsOrdered` are too close. Will do it.

awalkowiak:
@jmiranda @kchelstowski I pushed some minor improvements

kchelstowski (approved):

PR #4028, opened 2023-05-10
Release 0.8.22

awalkowiak:


kchelstowski (commented):
How does `Config` file look on production?
I basically ask for this part: https://github.com/openboxes/openboxes/blob/08d029c41de53e6ea9e8c15f15a1c710c263baa1/grails-app/conf/Config.groovy#L1873-L1952

If the `master` and production config itself takes those templates from `develop` branch, we should merge https://github.com/openboxes/openboxes/pull/4028/commits/293d731ae06bd176e4f2f03bc1ca2463c525481d to develop, because I have changed one of the templates.

awalkowiak:
@kchelstowski master is always merged into develop after release, hence it should be fine. The part that it is taking the file from develop by default could be changed to master by default, but it is not "crucial" at this point.

awalkowiak:
@jmiranda I am not sure why the unit failed, because "GitHub Actions has encountered an internal error when running your job.". Seems like it is GH issue, not the code issue (plus its on the new crowdin commit). 

Reminder: remember to not squash it ;)

kchelstowski:
@awalkowiak if so, then this LGTM

kchelstowski (approved):

awalkowiak:
@jmiranda 3rd rerun helped, it was github issue.

PR #4030, opened 2023-05-11
OBGM-398 Fix duplication of newly added items on save in inbound returns

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/StockTransferApiController.groovy
```diff
@@ -119,12 +119,8 @@ class StockTransferApiController {
         render([data: stockTransfer?.toJson()] as JSON)
     }
 
-    Boolean isNull(Object objectValue) {
-        return objectValue == JSONObject.NULL || objectValue == null || objectValue?.equals("")
-    }
-
```
drodzewicz, line 125: `isNull` is not being used anywhere so I decided to clean up and remove it.

grails-app/controllers/org/pih/warehouse/api/StockTransferApiController.groovy
```diff
@@ -187,8 +183,8 @@ class StockTransferApiController {
             }
 
             // For inbound returns
-            Date expirationDate = stockTransferItemMap.expirationDate && stockTransferItemMap.expirationDate != JSONObject.NULL ? Constants.EXPIRATION_DATE_FORMATTER.parse(stockTransferItemMap.expirationDate) : null
-            String lotNumber = stockTransferItemMap.lotNumber && stockTransferItemMap.lotNumber != JSONObject.NULL ? stockTransferItemMap.lotNumber : null
```
drodzewicz, line 191: the `JSONObject.NULL` appears to be not supported anymore by grails 3 so every time we encountered this code it threw an error

grails-app/controllers/org/pih/warehouse/api/StockTransferApiController.groovy
```diff
@@ -119,12 +119,8 @@ class StockTransferApiController {
         render([data: stockTransfer?.toJson()] as JSON)
     }
 
-    Boolean isNull(Object objectValue) {
-        return objectValue == JSONObject.NULL || objectValue == null || objectValue?.equals("")
-    }
-
     StockTransfer bindStockTransferData(StockTransfer stockTransfer, User currentUser, Location currentLocation, JSONObject jsonObject) {
-        bindData(stockTransfer, jsonObject)
+        bindData(stockTransfer, jsonObject, [exclude: ['stockTransferItems']])
```
drodzewicz, line 123: comparing both grails1 and grails3 versions I noticed that in grails1 when doing `bindData(stockTransfer, jsonObject)` it would not bind **stockTransferItems** so we had to do it manually.
In grails3 it does initally bind **stockTransferItems**, at least some of the values, but some other values we have to bind ourselves. 
I decided to mimic the grails1 behavior and just excluded **stockTransferItems** from being bound to **stockTransfer** object

kchelstowski (approved):

alannadolny (approved):

awalkowiak (approved):

PR #4038, opened 2023-05-16
OBPIH-5671 Add category parents filter to inventory reports and update download to show parent

alannadolny:


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -453,18 +454,34 @@ class ProductAvailabilityService {
         return quantityMap
     }
 
-    Map<Product, Map<Location, Integer>> getQuantityOnHandByProduct(Location[] locations) {
+    Map<Product, Map<Location, Integer>> getQuantityOnHandByProduct(Location[] locations, String[] categories, Boolean includeCategoryChildren) {
+        // Mysql throws an error when query with "category.id in (empty set)" is executed, but it works when NULL is included in this set
+        String queryWithCategories = categories ? "and category.id in (:categories)" : 'and category.id in (NULL)'
```
kchelstowski, line 459: wouldn't more reasonable be doing just a fallback to `null` if categories are empty? 
So in the `showInventoryByLocationReport` just doing:
```groovy
List categories = params.list("category").findAll { it } ?: null
```
The `findAll` is done to prevent something like this `[""]` being considered as truthy value (checkout `LocationController` 87 line): https://github.com/openboxes/openboxes/blob/63b53d3e7a9a60d4d30db09883fc0926b67cc77e/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy#L86-L87

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -453,18 +454,34 @@ class ProductAvailabilityService {
...
+
+        // Getting all subcategories of picked categories
+        def categoriesWithChildren = categories.collect { String categoryId ->
+            Category category = Category.load(categoryId)
+            category.getChildren() + category
```
kchelstowski, line 464:
```suggestion
            category?.children + category
```

just in case someone somehow applies non-existing id, e.g. via URL + you don't have to call the method, because it's a transient, so you can just call `children`

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -453,18 +454,34 @@ class ProductAvailabilityService {
...
+        }.flatten().collect { it.id }
+
+        // I am checking if "include all products in all subcategories" checkbox is checked to add children categories
+        def availableCategories = includeCategoryChildren ? categoriesWithChildren : categories
```
kchelstowski, line 468: if so, I'd do something to avoid calculating (fetching, collecting, etc) `categoriesWithChildren` from above, if you might not even use it if the `includeCategoryChildren` is not checked.

grails-app/views/report/showInventoryByLocationReport.gsp
```diff
@@ -35,8 +35,27 @@
...
                             </div>
+                            <label><warehouse:message code="category.label"/></label>
+                            <p>
+                                <g:selectCategory
+                                        id="category" multiple="true"
+                                        class="chzn-select-deselect filter"
+                                        data-placeholder="${g.message(code: 'category.selectCategory.label', default: 'Select a category')}"
+                                        name="category"
+                                        noSelection="['':'']"
+                                        value="${params?.category instanceof String ? params?.category : params?.category?.collect { it }}"
```
kchelstowski, line 46: could you elaborate on this? 
It seems like the fallback's collect does not have any effect:
```groovy
params?.category?.collect { it }
```
Isn't it equivalent to JS'
```javascript
array.map(x => x)
```
?

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -544,7 +545,10 @@ class ReportController {
     }
 
     def showInventoryByLocationReport = { MultiLocationInventoryReportCommand command ->
-        command.entries = productAvailabilityService.getQuantityOnHandByProduct(command.locations)
+        String[] categories = params.get('category') instanceof String ? [params.get('category')] :  params.get('category') ?: []
```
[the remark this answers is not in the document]
alannadolny: The reason for this logic is that I didn't know about the `params.list(...)` method 🙂

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -544,7 +545,10 @@ class ReportController {
...
+        // if params don't contain _includeCategoryChildrenKey, then we should set checkbox to true as a default
+        Boolean includeCategoryChildren = !params.containsKey('_includeCategoryChildren') ? true : !!params.includeCategoryChildren
```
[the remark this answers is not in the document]
alannadolny: `!params.containsKey('_includeCategoryChildren')` <- `_includeCategoryChildren` is not visible when you get on the report page for the first time, so it means that checkbox should be set to true (as a default value). `includeCategoryChildren` is not visible, when checkbox is set to false, so I can't rely on this value in this case.
`!!params.includeCategoryChildren` <- when `includeCategoryChildren` has value it means that checkbox should be set to true (it was set to true by user)

grails-app/views/report/showInventoryByLocationReport.gsp
alannadolny: It doesn't work without .collect { it }. Now, when I know about `params.list(...)` I changed this part of code to this solution and now It is simplified. It is hard to say, why I doesn't work. I saw { it.id } above and I decided to try something similar and it started working after this weird change.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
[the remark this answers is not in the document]
kchelstowski: what you could optionally do would be also:
```groovy
// When accessing the page for the first time, the flag should be set to true
Boolean includeCategoryChildren = true
// If we run the report manually and the checkbox is not checked
if (params.button == "run" && !params.includeCategoryChildren) {
 includeCategoryChildren = false
}
```

grails-app/views/report/showInventoryByLocationReport.gsp
```diff
@@ -35,8 +35,27 @@
...
+                                        class="chzn-select-deselect filter"
+                                        data-placeholder="${g.message(code: 'category.selectCategory.label', default: 'Select a category')}"
+                                        name="category"
+                                        noSelection="['':'']"
+                                        value="${params?.list('category')}"
+                                />
+                            </p>
+                            <p>
+                                <label>
+                                    <g:checkBox name="includeCategoryChildren" value="${params?.includeCategoryChildren}" checked="${includeCategoryChildren}"/>
```
kchelstowski, line 51: the initial value could be just set to `includeCategoryChildren` and you can get rid of `checked`, as it inherits from `value`:
```gsp
 <g:checkBox name="includeCategoryChildren" value="${includeCategoryChildren}" />
```

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -579,7 +587,7 @@ class ReportController {
                             sw.append('"' + (entry.key?.productCode ?: "").toString()?.replace('"', '""') + '"').append(",")
                             sw.append('"' + (entry.key?.displayNameWithLocaleCode ?: "").toString()?.replace('"', '""') + '"').append(",")
                             sw.append('"' + (entry.key?.productFamily?.name ?: "").toString()?.replace('"', '""') + '"').append(",")
-                            sw.append('"' + (entry.key?.category?.name ?: "").toString()?.replace('"', '""') + '"').append(",")
+                            sw.append('"' + (entry.key?.category?.getHierarchyAsString(" > ") ?: "").toString()?.replace('"', '""') + '"').append(",")
```
jmiranda, line 590: Was this requested? This could get expensive for lot of rows and will likely would add a lot of noise to the CSV.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -544,7 +545,14 @@ class ReportController {
...
     def showInventoryByLocationReport = { MultiLocationInventoryReportCommand command ->
-        command.entries = productAvailabilityService.getQuantityOnHandByProduct(command.locations)
+        String[] categories = params.list('category').findAll { it } ?: null
+        // When accessing the page for the first time, the flag should be set to true
+        Boolean includeCategoryChildren = true
+        // If we run the report manually and the checkbox is not checked
+        if (params.button == "run" && !params.includeCategoryChildren) {
+            includeCategoryChildren = false
+        }
+        command.entries = productAvailabilityService.getQuantityOnHandByProduct(command.locations, categories, includeCategoryChildren)
```
jmiranda, line 555: While I don't mind the logic being buried in getQuantityOnHandByProduct(), I think the categories + includeCategoryChildren logic could be handled with a call like this just to make it clear and reusable.
```
def categories = productService.getCategories(command.categories, includeCategoryChildren)
```
Also those two properties (categories, includeCategoryChildren) should be added to the command class.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -603,7 +611,7 @@ class ReportController {
             render(contentType: "text/csv", text: CSVUtils.prependBomToCsvString(sw.toString()), encoding: "UTF-8")
         }
 
-        render(view: 'showInventoryByLocationReport', model: [command: command])
+        render(view: 'showInventoryByLocationReport', model: [command: command, includeCategoryChildren: includeCategoryChildren])
```
jmiranda, line 614: As mentioned earlier, includeCategoryChildren should be included as a boolean property of the command class.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -453,18 +454,38 @@ class ProductAvailabilityService {
...
+        if (categories) {
+            categoriesQuery = "and category.id in (:categories)"
+            def availableCategories = includeCategoryChildren ? categories.collect { String categoryId ->
+                if (categoryId) {
+                    Category category = Category.load(categoryId)
+                    category?.children + category
+                }
+            }.flatten().collect { it?.id } : categories
+            queryArguments += [categories: availableCategories]
+        }
```
jmiranda, line 476: I need a little more time to review this code.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
alannadolny: Yes, it was written in ticket in this way: ```When I open the downloaded report, Then between “Family” and “Category” columns, I can see a column with “Parent Category” OR category appears as [Parent]>[Sub]>[Sub 2]. This option might be easier because there are sometimes several levels of category.```

src/groovy/org/pih/warehouse/report/MultiLocationInventoryReportCommand.groovy
```diff
@@ -10,16 +10,30 @@
...
 class MultiLocationInventoryReportCommand {
 
     Location[] locations
+    Category[] categories
+    Boolean includeSubcategories = Boolean.TRUE
+    String buttonAction
 
     Map<Product, Map<Location, Integer>> entries = [:]
 
+    static transients = ["isIncludeSubcategoriesEnabled"]
```
kchelstowski, line 25: seems like this has no effect?

kchelstowski (changes requested):

src/groovy/org/pih/warehouse/report/MultiLocationInventoryReportCommand.groovy
alannadolny: I forgot to delete this. Thanks.

grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
jmiranda: Ok, this could end up causing some performance issues but we'll leave it in for now.

jmiranda (approved):
I've made some changes to help demonstrate the correct way to use the command object (note: request parameters are bound to the command object automatically). 

@awalkowiak I'm good with the rest of the code so feel free to merge if this looks ok to you.

src/groovy/org/pih/warehouse/report/MultiLocationInventoryReportCommand.groovy
jmiranda: I've removed it.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1490,4 +1490,8 @@ class ProductService {
         }
         return synonym
     }
+
+    Category[] getCategories(String[] categoriesId) {
+        categoriesId.collect { Category.load(it) }
+    }
```
kchelstowski, line 1496: seems like it is not used anymore after @jmiranda's changes

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -453,18 +454,26 @@ class ProductAvailabilityService {
...
     }
 
-    Map<Product, Map<Location, Integer>> getQuantityOnHandByProduct(Location[] locations) {
+    Map<Product, Map<Location, Integer>> getQuantityOnHandByProduct(List<Location> locations, List<Category> categories) {
+        def categoriesQuery = "";
+        def queryArguments = [locations: locations]
+        // if we don't have categories and checkbox is checked, then we should get all of the products
+        if (categories) {
+            categoriesQuery = "and category.id in (:categories)"
+            queryArguments += [categories: categories.collect { it.id }]
```
kchelstowski, line 463: tidbit: Groovy allows you to do just `categories.id`

kchelstowski (approved):

PR #4044, opened 2023-05-18
OBPIH-5616 Improve calculating quantity allocated for product availability refresh

awalkowiak:


awalkowiak:
(recreating this one, because it was merged into hotfix1 branch, but we want to develop it on the develop branch)

PR #4045, opened 2023-05-18
OBPIH-5697 Filter out supplier locations for Return Order workflow

drodzewicz:


awalkowiak (approved):

jmiranda (approved):
LGTM. I still want to refactor some of these location service methods once we're done with the grails migration as the logic seems unnecessarily complicated and possibly in the wrong place (e.g. applyUserFilter, isReturnOrder).

PR #4048, opened 2023-05-19
OBPIH-5668 Save QOH on ward requests and show in edit phase

alannadolny:


grails-app/migrations/0.8.x/changelog-2023-05-17-1453-add-quantity-counted-column.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="AlanNadolny" id="678123853606-1">
```
kchelstowski, line 6: ideally the id should be the current date and time (the same that you gave to the file name)

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -821,7 +828,7 @@ class AddItemsPage extends Component {
         val => ({
           ...val,
           disabled: true,
-          quantityOnHand: this.state.isRequestFromWard ? '' : val.quantityOnHand,
+          quantityOnHand: val.quantityCounted,
```
kchelstowski, line 831: I'm not sure if it is correct? 
Do you have confirmed behaviors for:

- locations which are not ward 
- stocklist both types: pull, push?
- stock request without using a stocklist
?

src/js/components/stock-movement-wizard/request/EditPage.jsx
```diff
@@ -1033,6 +1033,7 @@ class EditItemsPage extends Component {
         quantityOnHand: val.quantityOnHand > 0 ? val.quantityOnHand : 0,
         quantityAvailable:
             val.quantityAvailable > 0 ? val.quantityAvailable : 0,
+        quantityOnHandRequesting: val.quantityCounted,
```
kchelstowski, line 1036: if you keep using `quantityCounted` everywhere, why here it is named as `quantityOnHandRequesting`?

kchelstowski (changes requested):

src/js/components/stock-movement-wizard/request/EditPage.jsx
kchelstowski: additional question - where is it even used? I can see you adding it here, but I don't see it being used anywhere (backend would anyway skip it, because you name it as `quantityCounted` everywhere)

alannadolny: I just followed the naming of the column. It was `quantityOnHandRequesting` in all of the fields in this component.

kchelstowski: after looking into the code, I get it now - I haven't known it's a read-only field - this is why I was confused on the naming.

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
alannadolny: I forgot about locations which are not ward. Thanks for catching that.

awalkowiak (approved):

jmiranda (commented):
I agree with @kchelstowski about the naming i.e. we should keep it consistent. But I'm not going to delay this getting merged since I don't understand the context well enough.

kchelstowski (approved):

PR #4052, opened 2023-05-22
OBPIH-5687 Use same product dropdown for both transfer stock and transfer out

drodzewicz:
- filter out locations that support MANAGE_INVENTORY and Supplier locations
- use same product dropdown for stock transfer in modal

### DIFF
--- a/grails-app/views/inventoryItem/_transferStock.gsp
+++ b/grails-app/views/inventoryItem/_transferStock.gsp
@@ -69,12 +69,9 @@
                     ${location.name}
                 </td>
                 <td>
-                    <g:selectLocation id="transferTo-${inventoryItem?.id}" name="otherLocation.id" class="chzn-select-deselect trigger-change"
-                                      noSelection="['null':'']"
-                                      data-placeholder="${g.message(code: 'inventoryItem.chooseStockTransfer.label', default: 'Choose where stock is being transferred to ...')}"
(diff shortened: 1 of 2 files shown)


awalkowiak (approved):

grails-app/views/inventoryItem/_transferStock.gsp
```diff
@@ -69,12 +69,9 @@
                     ${location.name}
                 </td>
                 <td>
-                    <g:selectLocation id="transferTo-${inventoryItem?.id}" name="otherLocation.id" class="chzn-select-deselect trigger-change"
-                                      noSelection="['null':'']"
-                                      data-placeholder="${g.message(code: 'inventoryItem.chooseStockTransfer.label', default: 'Choose where stock is being transferred to ...')}"
-                                      data-url="${request.contextPath}/inventoryItem/refreshBinLocation"
-                                      data-target="#transferTo-binLocation"
-                                      value=""/>
+                    <g:selectTransactionDestination name="otherLocation.id" class="chzn-select-deselect trigger-change"
```
jmiranda, line 72: What about refreshing the bin locations?

drodzewicz: thanks for noticing, I amended the commit and probably did a `git push` which failed, instead of a `git push --force`

PR #4053, opened 2023-05-22
OBPIH-5710 Add validation for quantity per UoM input field

drodzewicz:
validate quantity per UoM value to be not less than 1

### DIFF
--- a/grails-app/views/order/_orderItemFormDialog.gsp
+++ b/grails-app/views/order/_orderItemFormDialog.gsp
@@ -105,13 +105,12 @@
                     <label for="dlgQuantityPerUom"><warehouse:message code="orderItem.quantityPerUom.label"/></label>
                 </td>
                 <td valign="top" class="value">
-                    <input type="text"
-                           id="dlgQuantityPerUom"
-                           name="quantityPerUom"
-                           value="${g.formatNumber(number: orderItem?.quantityPerUom, maxFractionDigits: 0)?:1}"
-                           min="0"
-                           size="10"
-                           class="text" />
+                    <g:decimalNumberField decimal="0"
+                                          id="dlgQuantityPerUom"
+                                          name="quantityPerUom"
+                                          value="${g.formatNumber(number: orderItem?.quantityPerUom, maxFractionDigits: 0)?:1}"
+                                          size="10"
+                                          class="text" />
                 </td>
             </tr>
             <tr class="prop">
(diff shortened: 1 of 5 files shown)


grails-app/views/taglib/_decimalFormatField.gsp
```diff
@@ -16,7 +16,18 @@
...
+          // handle decimal points
+          if ("${attrs.decimal}") {
+            const decimalPoint = parseInt("${attrs.decimal}")
+            const [number, decimal] = parsedNumber?.split("${decimalSeparator}");
+            if (decimalPoint > 0) {
+              parsedNumber = number + "${decimalSeparator}" + decimal.substring(0, decimalPoint);
+            } else {
+              parsedNumber = number;
+            }
+          }
```
kchelstowski, line 30: I don't understand this part. Could you elaborate on that? The logic seems a bit complexed and is difficult to understand without proper examples?

grails-app/views/order/_orderItemFormDialog.gsp
```diff
@@ -362,6 +361,12 @@
     var budgetCode = $("#dlgBudgetCode").val();
     var quantity = parseInt($("#dlgQuantity").val());
     var quantityInShipments = parseInt($("#dlgQuantityInShipments").val())
+    var quantityPerUom = $("#dlgQuantityPerUom").val();
```
kchelstowski, line 364:
```suggestion
    const quantityPerUom = $("#dlgQuantityPerUom").val();
```
was there a scope-reason you used `var`?

grails-app/views/taglib/_decimalFormatField.gsp
drodzewicz: I was afraid that this might not be so straightforward at first glance but I tried making this logic as simple as I could.
I guess adding comment to each line would explain this process
```js
// only if attribute decimal is passed then apply below logic
if ("${attrs.decimal}") {
  // attrs.decimal is a string so we need to cast it to a number
  const decimalPoint = parseInt("${attrs.decimal}")
  // given we have a number "12,345"  with a decimal separator ","
  // we split the number string into two parts: number (12) and decimal (345)
  const [number, decimal] = parsedNumber?.split("${decimalSeparator}");
  if (decimalPoint > 0) {
    // if decimalPoint is > 0 (for example 2), then return  a number with fixed decimal points like: "12" + "," + "34" 
    parsedNumber = number + "${decimalSeparator}" + decimal.substring(0, decimalPoint);
  } else {
  // if decimal point is <= 0 the just return the number and ignore the decimal part
  parsedNumber = number;
}
```

grails-app/views/order/_orderItemFormDialog.gsp
drodzewicz: there was no scope reason, I was just following the previous examples of getting the value from a  selector.
I could change it to const but it doesn't;t really make any difference in the context of this function

grails-app/views/taglib/_decimalFormatField.gsp
kchelstowski: this is what was missing :laughing: 
Thanks, I've had a good grasp of what is going on, but wanted you to confirm

kchelstowski (approved):

alannadolny (approved):

awalkowiak (approved):

PR #4054, opened 2023-05-22
OBPIH-5690 Allow random product identifier generation for default pro…

awalkowiak:
…duct type

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
@@ -238,6 +238,15 @@ class IdentifierService {
     }
 
     def generateSequentialProductIdentifier(ProductType productType) {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -335,7 +386,7 @@ class IdentifierService {
         def products = Product.findAll("from Product as p where p.active = true and (p.productCode is null or p.productCode = '')")
         products.each { product ->
             try {
-                def productCode = generateProductIdentifier()
+                def productCode = product.productType ? generateProductIdentifier(product.productType) : generateProductIdentifier()
```
jmiranda, line 389: Oh shit, good catch. This would likely happen a bunch if we imported products without product code.

jmiranda (approved):

PR #4058, opened 2023-05-23
OBPIH-5717 Add Haitian to the default supported locale

awalkowiak:


grails-app/i18n/messages.properties
```diff
@@ -3119,6 +3119,7 @@ react.default.italian.label=Italiano
 react.default.spanish.label=Español
 react.default.portuguese.label=Português
 react.default.acoli.label=Acoli
+react.default.haitian.label=Haitian
```
jmiranda, line 3122: Haitian isn't a language and I want these names to be in the form that would be recognized by the native speaker, so this one should be Kreyòl.

jmiranda (changes requested):

grails-app/i18n/messages.properties
awalkowiak: @jmiranda I aligned it with what is returned by `Locale.getDisplayName()` here https://github.com/openboxes/openboxes/blob/develop/grails-app/views/common/_footer.gsp#LL28C9-L29C11. Otherwise we would get different labels on gsp and react footers. But if you want to turn it into `Kreyòl`, then we can change the `getDisplayName` in the `common/_footer.gsp` into `g:message`.

awalkowiak:
@jmiranda I made some improvements on that one. You can review it.

jmiranda:
Sorry I didn't get a chance to review this earlier today but I'm looking at it now on my phone. 

There's something with displaying the locale name that I'm not grasping and I think I need to be sitting in front of my laptop to fully understand. So give me until Tuesday before merging. 

The case I'm worried about is when a user configures more supported locales in openboxes-config.groovy.

jmiranda:
Never mind I see that you're handling that.

jmiranda (approved):

grails-app/i18n/messages.properties
```diff
@@ -1425,6 +1425,18 @@ localization.enable.label=Enable localization mode
 localization.disable.label=Disable localization mode
 localization.invalid.custom.message=To enable database storage for custom translations, please add <code>openboxes.locale.custom.enabled = true</code> to your <code>openboxes-config.properties</code> file.
 localizations.label=Localizations
+# Locales
+locale.ach.label=Acoli
```
jmiranda, line 1429: One minor change. I've been misspelling this for months. Could you change it to Acholi.

jmiranda (changes requested):

jmiranda:
Once that change is done feel free to dismiss my review and merge the PR.

PR #4060, opened 2023-05-24
OBPIH-5710 Add validation for per UoM field on import and fix validation message

drodzewicz:


alannadolny (approved):

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -138,8 +138,8 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         requestedBy(nullable: true)
         quantity(nullable: false, min: 1)
         quantityUom(nullable: true)
-        // Validate quantity per uom != 0 (value as boolean, to catch 0, 0.0 and '')
-        quantityPerUom(nullable: false, validator: { value -> value as boolean } )
+        // Validate quantity per uom > 0 (value as boolean, to catch 0, 0.0 and '')
+        quantityPerUom(nullable: false, validator: { value -> value as boolean && value > 0 } )
```
jmiranda, line 142: It probably doesn't hurt to keep it but do we still need the `value as boolean` part? I assume nullable:false handles empty string and null value but even if it doesn't I have a sneaking suspicion that it would be handled implicitly by the `value > 0` custom validation logic. I will check this using unit tests but I think we can get rid of this part of the logic.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -798,8 +798,8 @@ class OrderService {
                         UnitOfMeasure uom = uomParts.length > 1 ? UnitOfMeasure.findByCodeOrName(uomParts[0], uomParts[0]) : null
                         orderItem.quantityUom = uom
                         BigDecimal qtyPerUom = uomParts.length > 1 ? CSVUtils.parseNumber(uomParts[1], "unitOfMeasure"): null
-                        if (!qtyPerUom) {
-                            throw new IllegalArgumentException("Quantity per UoM cannot be 0 or empty")
+                        if (!qtyPerUom || qtyPerUom < 1) {
```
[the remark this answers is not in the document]
jmiranda: I agree but could you clarify how you think this could be done?

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
awalkowiak: @jmiranda Yup, you're right, the `as boolean` was for `non 0/empty/null`, since we now have `>= 1`, then it's not needed anymore.

awalkowiak (approved):

PR #4061, opened 2023-05-24
OBPIH-5669 Adjust confirm alert message after enabling adding items t…

kchelstowski:
…o a request after submission


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -686,24 +686,6 @@ class AddItemsPage extends Component {
...
-        {
-          label: this.props.translate('react.default.goBack.label', 'Go back'),
-        },
-        {
-          label: this.props.translate('react.default.submit.label', 'Submit'),
-          onClick: onConfirm,
-        },
-      ],
-    });
-  }
```
kchelstowski, line 706: I saw the same label being used in the outbound workflow, so I wanted to change it here, but it turned out that this `confirmSubmit` function is not even used in this component (probably a copy paste from the request was done here)

awalkowiak (approved):

PR #4063, opened 2023-05-25
OBPIH-5698 Inventory summary indicator below min qtys not accurate

alannadolny:


grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -370,15 +370,14 @@ class DashboardService {
...
         def lowStock = quantityMap.findAll { product, quantity ->
             def inventoryLevel = inventoryLevelMap[product]?.first()
-            def minQuantity = inventoryLevelMap[product]?.first()?.minQuantity
-            inventoryLevel?.status >= InventoryStatus.SUPPORTED && minQuantity && quantity > 0 && quantity <= minQuantity
+            def minQuantity = inventoryLevel?.minQuantity
+            inventoryLevel?.status >= InventoryStatus.SUPPORTED && minQuantity && quantity <= minQuantity
         }
 
         def reorderStock = quantityMap.findAll { product, quantity ->
             def inventoryLevel = inventoryLevelMap[product]?.first()
```
kchelstowski, line 378: At one point I thought, this might lead to some bugs, because locally I had some products with a few inventory levels:
Screenshot from 2023-05-25 11-49-46
so taking `first()` would not be fine, but those inventory levels were from 2018 and I can see that in 2021 we've added a validation for uniqueness:
Screenshot from 2023-05-25 12-51-44
so it should be fine.

kchelstowski (approved):
LGTM - tested locally and left a comment in the ticket.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
jmiranda: Thanks for bringing that up. It's definitely something to consider. No telling what data exists in some systems. Perhaps we can throw an explicit exception when we encounter more than one inventory level for a product and location. Can you add a comment to the ticket that links to this PR discussion so that Kelsey and kasia are aware.

jmiranda (approved):

PR #4064, opened 2023-05-25
OBPIH-5702 CSV consumption report does not contain data for certain l…

jmiranda:
…ocations


kchelstowski (approved):
LGTM, but still wondering why it would be the case - do you know the answer or it's a blind guess for now?

alannadolny (approved):

drodzewicz (approved):

PR #4066, opened 2023-05-26
OBPIH-5723 Make transfer stock a superuser action only

kchelstowski:


awalkowiak (changes requested):
I think you should also modify this one: https://github.com/openboxes/openboxes/blob/develop/grails-app/conf/RoleFilters.groovy#L20

awalkowiak (approved):

alannadolny (approved):

awalkowiak:
I forgot this was meant to be changed. I moved the ticket back to in progress to change it (I think there is no need to revert)

PR #4067, opened 2023-05-26
OBPIH-5672 Add new custom document - Requisition template

awalkowiak:


grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -192,4 +195,73 @@ class DocumentTemplateService {
...
+                report.process(context, outputStream)
+            }
+            log.info "Report dump: " + report.dump()
+        } finally {
+            PdfUtil.restoreBaseFonts()
+        }
+    }
+
+    private createRequisitionContext(IXDocReport report, Requisition requisitionInstance) {
+        def requisitionItems = requisitionInstance?.requisitionItems?.collect { RequisitionItem requisitionItem ->
```
awalkowiak, line 226: @jmiranda I had to go with this version either way, because a couple of properties used are transients, and this was easier. I tested it on 150-200 items, and the performance was very good.

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -192,4 +195,76 @@ class DocumentTemplateService {
...
+            def demand
+            def quantityOnHand
+            if (requisitionInstance?.destination?.isWard()) {
```
jmiranda, line 229: This logic makes me uneasy since there's no way to tell if demand is for a single location or it's aggregated. I think if there's we should only care about demand for the origin/destination pair and demand should be 0 if a demand record does not exist for that pair. 

Can you ask Kelsey about this?

jmiranda (changes requested):
LGTM other than the issue around demand.

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
kchelstowski: One thing worth mentioning - if it was meant to be working for locations without `MANAGE_INVENTORY` and with `SUBMIT_REQUEST`, it won't, because `isWard()` checks only if `locationTypeCode == LocationTypeCode.WARD`

awalkowiak: @jmiranda Kelsey wanted to have it working like on add items page. Where it is working like that - if a request is from "ward" then pull demand from the origin to the requestor, and if a request is not from the "ward", then pull demand outgoing from the requestor to all other locations.

@kchelstowski I specifically asked if this is fine on Friday's standup because we are using exactly this in the `stockMovementService.getAddPageItem` and it was fine.

awalkowiak:
I am merging to start testing and I'll confirm the above concerns today and improve if needed.

PR #4068, opened 2023-05-26
OBPIH-5714 Click Next on add items page should be enabled when some line items have qtt=0

alannadolny:


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -905,6 +900,13 @@ class AddItemsPage extends Component {
             // newly saved item to replace its equivalent in state
             const itemToChange = _.last(_.differenceBy(lineItemsBackendData, itemCandidatesToSave, 'id'));
             const lineItemsAfterSave = this.state.values.lineItems.map((item) => {
+              if (
+                parseInt(item.quantityRequested, 10) === 0 &&
+                item.rowSaveStatus === RowSaveStatus.SAVING &&
+                !_.includes(savedItemsIds, item.id)
+              ) {
+                return { ..._.omit(item, ['id', 'statusCode']), disabled: true, rowSaveStatus: RowSaveStatus.SAVED };
+              }
```
kchelstowski, line 909: what does it do? we get rid of `id` and disable the line, mark line as `SAVED`, but what was meant behind the scenes about this?

alannadolny: It means that the line has a quantity of 0 (I discussed it with Manon, and she said it should be marked as saved). So, after receiving a response with information that my line was deleted, I am removing `id` and `statusCode` properties. These two properties within the `lineItem` indicate that the line is saved, so I have to delete them, because changing the quantity of previously deleted line should treat this line as a new one instead of edited.

kchelstowski: but if the item is deleted, why should it be marked as saved and why should the line be **disabled** instead of deleted from the table form? I mean, I understand the logic with the omit, but I don't understand the `disabled` and SAVED in this case, if the item is DELETED, not saved.
Does it mean, even though the backend has already deleted the item, you have to delete it from table manually additionally?

alannadolny: All of the items with `qty = 0` will be deleted after going to next step or just saving this step manually. Users can change the quantity to 0 by accident, so why we should delete this line from the table? 

Screenshot from the ticket:
image

kchelstowski: ok, if so, then I understand now

kchelstowski (approved):

drodzewicz (approved):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -538,24 +537,18 @@ class AddItemsPage extends Component {
...
-    } else {
-      lineItemsData = _.map(data, val => ({ ...val, disabled: true }));
-    }
+    const lineItemsData = data.length ? _.map(data, val => ({ ...val, disabled: true })) :
+      new Array(1).fill({ sortOrder: 100, rowSaveStatus: RowSaveStatus.PENDING });
 
     const sortOrder = _.toInteger(_.last(lineItemsData).sortOrder) + 100;
     this.setState({
-      currentLineItems: this.props.isPaginated ?
+      currentLineItems: !setAllFetchedItems && this.props.isPaginated ?
```
awalkowiak, line 547: IIRC `startIndex == null` was indicating that it is not paginated. I am not sure if there is a need for `setAllFetchedItems`.

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -727,8 +720,8 @@ class AddItemsPage extends Component {
     return apiClient.get(url)
       .then((response) => {
         this.setState({
-          totalCount: response.data.data.length,
-        }, () => this.setLineItems(response, null));
+          totalCount: response.data.data.length || 1,
```
awalkowiak, line 723: Is this `1` ok if you remove all?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -926,7 +928,7 @@ class AddItemsPage extends Component {
 
             this.setState({
               values: { ...this.state.values, lineItems: lineItemsAfterSave },
-              currentLineItems: lineItemsAfterSave,
+              currentLineItems: lineItemsBackendData,
```
awalkowiak, line 931: Why `lineItems` are `lineItemsAfterSave`, and `currentLineItems` are `lineItemsBackendData`? Should these two be different in this state?

awalkowiak (commented):
@alannadolny overall LGTM, but I have a few questions

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: It is connected to the fact, that lines with 0 qty are not saved in database, but those lines should be visible in the table. Setting `currentLineItems` to `lineItemsBackendData`resolves the problem with duplicated items, because if I tried to change the quantity of items fastly to 0, some of the items were not deleted

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
awalkowiak: Is it possible to use for it `startIndex == null`?

awalkowiak (approved):

PR #4070, opened 2023-05-29
OBPIH-5724 Include tab separator to the CSVUtil and improve outbound …

kchelstowski:
…movement import template

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
@@ -22,6 +22,7 @@ import org.pih.warehouse.core.Document
 import org.pih.warehouse.core.DocumentCommand
 import org.pih.warehouse.core.DocumentType
 import org.pih.warehouse.core.Location
+import org.pih.warehouse.importer.CSVUtils
 import org.pih.warehouse.importer.ImportDataCommand
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.picklist.PicklistItem
@@ -360,8 +361,21 @@ class StockMovementController {
                 throw new IllegalArgumentException("File must be in CSV format")
             }
 
+            String[] columnKeys = [
+                    'requisitionItemId',
+                    'productCode',
+                    'productName',
+                    'packLevel1',
+                    'packLevel2',
+                    'lotNumber',
+                    'expirationDate',
+                    'quantityRequested',
+                    'receipientId'
+            ]
+
             String csv = new String(importFile.bytes)
-            def settings = [separatorChar: ',', skipLines: 1]
+            char separatorChar = CSVUtils.getSeparator(csv, columnKeys.length)
+            def settings = [separatorChar: separatorChar, skipLines: 1]
             Integer sortOrder = 0
             csv.toCsvReader(settings).eachLine { tokens ->
                 Boolean validateLotAndExpiry = stockMovement.getStockMovementDirection(currentLocation) != StockMovementDirection.OUTBOUND && !stockMovement.electronicType
--- a/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
@@ -361,20 +361,9 @@ class StockMovementController {
                 throw new IllegalArgumentException("File must be in CSV format")
             }
 
-            String[] columnKeys = [
-                    'requisitionItemId',
-                    'productCode',
-                    'productName',
-                    'packLevel1',
-                    'packLevel2',
-                    'lotNumber',
-                    'expirationDate',
-                    'quantityRequested',
-                    'receipientId'
-            ]
 
             String csv = new String(importFile.bytes)
-            char separatorChar = CSVUtils.getSeparator(csv, columnKeys.length)
+            char separatorChar = CSVUtils.getSeparator(csv, StockMovement.buildCsvRow()?.keySet()?.size())
             def settings = [separatorChar: separatorChar, skipLines: 1]
             Integer sortOrder = 0
             csv.toCsvReader(settings).eachLine { tokens ->
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -2741,17 +2741,7 @@ class StockMovementService {
         }
 
         def lineItems = stockMovement.lineItems.collect {
-            [
-                "Requisition item id"            : it?.id ?: "",
-                "Product code (required)"     : it?.product?.productCode ?: "",
-                "Product name"                  : it?.product?.displayNameWithLocaleCode ?: "",
-                "Pack level 1"                   : it?.palletName ?: "",
-                "Pack level 2"                      : it?.boxName ?: "",
-                "Lot number"                    : it?.lotNumber ?: "",
-                "Expiration date (MM/dd/yyyy)": it?.expirationDate ? it?.expirationDate?.format("MM/dd/yyyy") : "",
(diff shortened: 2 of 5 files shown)


alannadolny (approved):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -360,8 +361,21 @@ class StockMovementController {
                 throw new IllegalArgumentException("File must be in CSV format")
             }
 
+            String[] columnKeys = [
```
awalkowiak, line 364: This should be a const defined somewhere (either in consts, or StockMovement) and used in the exportCsv counterpart too

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -153,6 +153,12 @@ class CSVUtils {
         return new CSVPrinter(new StringBuilder(), CSVFormat.DEFAULT)
     }
 
+    static String findSeparator(String firstLine) {
+        List<String> allowedSeparators = [Constants.CUSTOM_COLUMN_SEPARATOR, Constants.TAB_COLUMN_SEPARATOR]
```
awalkowiak, line 157: According to this, `comma` is not allowed as a separator.

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -153,6 +153,12 @@ class CSVUtils {
...
+        String separator = allowedSeparators.find{ String it -> firstLine.contains(it) }
```
awalkowiak, line 158: If a file will be tab-separated and in the first column you will have a semicolon, then this won't work.

src/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -107,6 +107,7 @@ class Constants {
     // Default character used to separate columns and newlines in exports
     static final String DEFAULT_COLUMN_SEPARATOR = ","
     static final String CUSTOM_COLUMN_SEPARATOR = ";"
```
awalkowiak, line 109: Perhaps renaming it to `SEMICOLON_COLUMN_SEPARATOR` would be better since now `tab` is also a custom one.

awalkowiak (changes requested):

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
kchelstowski: I didn't add it here, because in previous logic it was falling back to comma even if the columns size != array size after spliting, but I guess it doesn't matter that much, so I'll add it here

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
kchelstowski: I know it won't work, but since the previous logic also didn't handle the case, where you'd have a semicolon in the first line and the separator was meant to be comma, it wouldn't work, so I thought we agreed then, that we didn't want to handle such an edge case.
But if you want to me now to handle this edge case, I'll do my best to come up with something

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: sure I can add it, I was just following what was done here: https://github.com/openboxes/openboxes/blob/33b93dcf20c46c073d13b6335350400b7778e4c8/grails-app/services/org/pih/warehouse/order/OrderService.groovy#L896-L913

As for exportCsv, it is defined as a Map like this:
https://github.com/openboxes/openboxes/blob/33b93dcf20c46c073d13b6335350400b7778e4c8/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy#L2757-L2769

so I'll have to think through it, if there is a reasonable way to make it reusable for both places. Would you say I should provide a change to `exportCsv` at this point as well?

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
drodzewicz: In that case you have to execute the whole algorithm for each column separator, not just check if the character is in the first line.
The reason why there was a `firstLine.contains(Constants.CUSTOM_COLUMN_SEPARATOR)` was like a first step:
- if there is `;` character in the first line there is  a chance that it might be a column separator
- then try to parse this first line with this `;` character 
- if the amount of parsed columns equals the expected column count then there is a high probability that the given character `;` is a column separator.
- if not then fallback on the default column separator which is a comma. 

If we want to handle many different characters as potential column separators, then we need to check all of these steps above for each of the characters in a potential character list
_pseudo code:_
```
List<String> allowedSeparators = [";", "\t"]
for (def separator in allowedSeparators) {
  if (firstLine.contains(separator)) { 
    def csvMapReader = new CSVMapReader(new StringReader(firstLine), [separatorChar: separator])

    if (csvMapReader.initFieldKeys().size() == columnCount) {
      return separator
      }
  }
}
// if none of them passed the test then fallback on comma
return ","
```

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
awalkowiak: Then name it `findCustomSepartor`

src/groovy/org/pih/warehouse/importer/CSVUtils.groovy
awalkowiak: @kchelstowski :point_up:

kchelstowski:
@awalkowiak @drodzewicz While you were giving review today, I was implementing the suggestions and have read yours just now - let me know if my version for finding the separator feels good to you or if I should change anything

src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -436,4 +436,18 @@ class StockMovement {
         return stockMovement
     }
 
+    static def buildCsvRow(StockMovementItem lineItem = null) {
```
kchelstowski, line 439: I don't feel confident about the name for this, since it sounds more reasonable for exporting part, and for import to get the column size I just do:
```groovy
StockMovement.buildCsvRow().keySet().size()
```
so let me know if it should stay or you have any other suggestion for the name of this

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2741,17 +2741,7 @@ class StockMovementService {
...
-                "Product code (required)"     : it?.product?.productCode ?: "",
-                "Product name"                  : it?.product?.displayNameWithLocaleCode ?: "",
-                "Pack level 1"                   : it?.palletName ?: "",
-                "Pack level 2"                      : it?.boxName ?: "",
-                "Lot number"                    : it?.lotNumber ?: "",
-                "Expiration date (MM/dd/yyyy)": it?.expirationDate ? it?.expirationDate?.format("MM/dd/yyyy") : "",
-                "Quantity (required)"        : it?.quantityRequested ?: "",
-                "Recipient id"                  : it?.recipient?.id ?: ""
-            ]
+            StockMovement.buildCsvRow(it)
```
kchelstowski, line 2744: Tested the export - works fine with that change

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: I've done it for both export/import, let me know what you think

drodzewicz (approved):
LGTM

alannadolny (approved):
👌

awalkowiak (approved):

PR #4075, opened 2023-05-30
OBPIH-5727 add category parents filter to in stock reports and update download to show parent

drodzewicz:


grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -531,9 +531,18 @@ class InventoryController {
...
+        // When accessing the page for the first time, the flag should be set to true
+        // Initially there not parameter button, only after running the report manually it is set
+        params.includeSubcategories = params.button ? params.includeSubcategories : true
+        if (params.includeSubcategories) {
+            categories = inventoryService.getExplodedCategories(categories)
+        }
+
+        def inventoryItems = dashboardService.getInStock(location, categories)
+
+        if (params.button == "download") {
```
kchelstowski, line 545: in this method and in methods below, a significant part seems to be the same for each of the methods (except `def inventoryItems`, but it could be also doable with some type of enum/flag I believe)  - maybe we could create a helper method or anything not to have to in the future refactor all of those three methods if ever necessary?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -475,15 +475,27 @@ class ProductAvailabilityService {
...
+        return getInventoryByProduct(location, [])
+    }
+
+    def getInventoryByProduct(Location location, List<Category> categories) {
+        def categoriesQuery = "";
+        def queryArguments = [location: location]
+
+        if (categories) {
+            categoriesQuery = "and pa.product.category.id in (:categories)"
+            queryArguments += [categories: categories.collect { it.id }]
```
kchelstowski, line 487: titbit: `categories.id` should be enough

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -531,9 +531,18 @@ class InventoryController {
...
-        def inventoryItems = dashboardService.getInStock(location)
+        List<Category> categories = Category.findAllByIdInList(params.list("categories"))
 
-        if (params.format == "csv") {
+        // When accessing the page for the first time, the flag should be set to true
+        // Initially there not parameter button, only after running the report manually it is set
+        params.includeSubcategories = params.button ? params.includeSubcategories : true
+        if (params.includeSubcategories) {
+            categories = inventoryService.getExplodedCategories(categories)
+        }
```
alannadolny, line 541: Maybe it is a good idea to create an additional function to handle this code, this part of code is duplicated multiple times. Additionally, if it is possible in Groovy to pass a function as an argument you can add calculating inventory items to the function (I am not sure if it is possible)

drodzewicz: Yes, you guys are right about code duplication and additionally, I noticed that I have missed more such lists that use the common `list.gsp`.
From what I can see it is possible to call the method based on the string value like [(docs)](http://www.groovy-lang.org/metaprogramming.html#_dynamic_method_names)
```groovy
dashboardService."somemethof"(args)

````
I think it would be best to do that but I am not sure where such a helper method should be. I don;t think we have ever done such a case for a controller. Any ideas?
@alannadolny @kchelstowski

Comment (kchelstowski) [reply grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy:541]: I would generally move most of this logic to the service, especially the fetching calls, but since we are tight on time, I think it could stay.
I'd recommend moving it to an InventoryService, but I don't have a quick idea of how to name the method

File: grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy [hunk 4]
```diff
@@ -638,52 +598,30 @@ class InventoryController {
...
+        def inventoryItems = dashboardService.getOutOfStock(location, params.abcClass, categories)
```
kchelstowski, line 622: Initially I wanted to write if you forgot to use the re-usable logic in this method as well, but I can see that it's a bit different than the others.
How about creating another method called, let's say `determineCategories` (ofc the name doesn't have to be like this) and doing there (pseudocode written by hand)
```groovy
List<Category> determineCategories(params) {
  List<Category> categories = Category.findAllByIdInList(params.list("categories"))
  params.includeSubcategories = params.button ? params.includeSubcategories : true
  if (params.includeSubcategories) {
    categories = inventoryService.getExplodedCategories(categories)
  }
}
```
and use it here and also change the `listStock` 488th line method to:
```groovy
...
List<Category> categories = determineCategories(params)
...
```

alannadolny (approved):

awalkowiak (approved):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -475,15 +475,27 @@ class ProductAvailabilityService {
...
+        return getInventoryByProduct(location, [])
+    }
+
+    def getInventoryByProduct(Location location, List<Category> categories) {
+        def categoriesQuery = "";
+        def queryArguments = [location: location]
+
+        if (categories) {
+            categoriesQuery = "and pa.product.category.id in (:categories)"
+            queryArguments += [categories: categories.id]
```
jmiranda, line 487: In the future, we should make this explicit just in case something changes. `categories.id` implies `categories*.id` which implies `categories.collect { it.id }`. I would prefer to be explicit.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -483,91 +483,56 @@ class InventoryController {
...
-
-
-    def listTotalStock = {
-        def location = Location.get(session.warehouse.id)
-        def inventoryItems = dashboardService.getTotalStock(location)
+        def inventoryItems = dashboardService."$methodName"(location, categories)
 
-        if (params.format == "csv") {
-            def filename = "Total stock - " + location.name + ".csv"
+        if (params.button == "download") {
```
jmiranda, line 504: Are we sure this isn't being used in other places?

PR #4076, opened 2023-05-31
OBPIH-5672 Show quantity issued only for approved items

awalkowiak:


grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -328,6 +328,10 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return !onHold
     }
 
+    Boolean hasNoManagedInventoryAndSupportsSubmittingRequests() {
```
jmiranda, line 331: My only comment is that this method name does not roll off the tongue. I'll try to think of a way to simplify it.

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -226,12 +226,13 @@ class DocumentTemplateService {
         def requisitionItems = requisitionInstance?.requisitionItems?.collect { RequisitionItem requisitionItem ->
             def demand
             def quantityOnHand
-            if (requisitionInstance?.destination?.isWard()) {
-                // if request FROM Ward then pull demand from origin to ward and use quantity counted as quantity on hand
+            if (requisitionInstance?.destination?.hasNoManagedInventoryAndSupportsSubmittingRequests()) {
+                // if request FROM "Ward" (location without managed inventory but supporting submitting requests),
+                // then pull demand from origin to that location
```
jmiranda, line 231: Let's no use "ward" because it could be all sorts of location types. Let's call it a downstream consumer. In fact, I think that's what I want the method to be called as well. Either of these sounds good to me.
```
Boolean isDownstreamConsumer() 
Boolean isDownstreamEndpoint()
```

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -243,7 +244,7 @@ class DocumentTemplateService {
                 requestorMonthlyDemand  : demand?.monthlyDemand ?: 0,
                 requestorQuantityOnHand : quantityOnHand,
                 quantityRequested       : requisitionItem?.quantity,
-                quantityIssued          : requisitionInstance?.isPending() ? "" : requisitionItem?.quantityIssued
+                quantityIssued          : requisitionItem.isApproved() && !requisitionInstance?.isPending() ? requisitionItem?.quantityIssued : ""
```
jmiranda, line 247: this could be encapsulated in a method like this ...
```
RequisitionItem.isQuantityAllocated()
```
or 
```
RequisitionItem.isQuantityIssued()
```
depending on which one is more accurate.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/core/Location.groovy
awalkowiak: Changed to `isDownstreamConsumer`

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
awalkowiak: Done

PR #4077, opened 2023-06-01
OBPIH-5714 Click Next on add items page should be enabled when some line items have qtt=0 (fixes after QA)

alannadolny:
I have to check `item.quantityRequested === undefined` instead of `!item.quantityRequested` because `quantityRequest` can be 0, so `!item.quantityRequested` will be true, but the main idea of checking that is just to validate the items which have empty quantity.

`if (item.id && newQty === oldQty && oldQty === 0) {
        lineItemsToBeUpdated.push(item);
        return;
      }` means that we want to send items with 0 qty which are added using stock list (before when we added items using stock list, they weren't sent to save, so items with 0 qty were visible on the next page / after save)

The issue with being unable to add new lines using arrows was caused by the asynchronous behavior of `setState`, so when I added `previousState` as an argument the issue was solved.

Additionally, I had to upgrade `react-confirm-alert` from 2.4.1 to 2.6.1, because, in the previous version, we weren't able to trigger two alerts in a row, so when we had an alert about items with 0 qty, the alert with duplicated items didn't appear, so we couldn't go forward in this case.

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -367,6 +367,7 @@ class AddItemsPage extends Component {
     this.saveAndTransitionToNextStep = this.saveAndTransitionToNextStep.bind(this);
     this.shouldShowAutosaveFeatureBar = this.shouldShowAutosaveFeatureBar.bind(this);
     this.shouldCreateAutosaveFeatureBar = this.shouldCreateAutosaveFeatureBar.bind(this);
+    this.didUserConfirmAlert = false;
     this.debouncedSave = _.debounce(() => {
       this.saveRequisitionItemsInCurrentStep(this.state.values.lineItems, false);
     }, 1000);
@@ -653,12 +654,19 @@ class AddItemsPage extends Component {
       buttons: [
         {
           label: this.props.translate('react.default.yes.label', 'Yes'),
-          onClick: onConfirm,
+          onClick: () => {
+            this.didUserConfirmAlert = true;
+          },
         },
         {
           label: this.props.translate('react.default.no.label', 'No'),
         },
       ],
+      afterClose: () => {
+        if (this.didUserConfirmAlert) {
+          onConfirm();
+        }
+      },
     });
   }
 
@@ -968,17 +976,17 @@ class AddItemsPage extends Component {
           return Promise.reject(new Error(this.props.translate('react.stockMovement.error.saveRequisitionItems.label', 'Could not save requisition items')));
         });
     }
-    this.setState({
+    this.setState(previousState => ({
       values: {
-        ...this.state.values,
-        lineItems: this.state.values.lineItems.map((item) => {
+        ...previousState.values,
+        lineItems: previousState.values.lineItems.map((item) => {
           if (parseInt(item.quantityRequested, 10) === 0) {
             return { ...item, disabled: true, rowSaveStatus: RowSaveStatus.SAVED };
           }
           return item;
         }),
       },
-    });
+    }));
 
     return Promise.resolve();
   }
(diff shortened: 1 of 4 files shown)


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -441,7 +442,7 @@ class AddItemsPage extends Component {
     const lineItemsToBeUpdated = [];
     _.forEach(lineItemsWithStatus, (item) => {
       // We wouldn't update items with quantity requested < 0
-      if (!item.quantityRequested || parseInt(item.quantityRequested, 10) < 0) {
+      if (!item.quantityRequested === undefined || parseInt(item.quantityRequested, 10) < 0) {
```
kchelstowski, line 445: I am not convinced about that - is `undefined` the only "empty" value `quantityRequested` can be? Can't it be `''` or `null`? 
I know it might turn into "spaghetti", but wouldn't it be more safe to do `(item.quantityRequested != 0 && !item.quantityRequested)`

If there is no way `quantityRequested` can be an empty string/null, then I'm fine with that

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -958,17 +976,17 @@ class AddItemsPage extends Component {
           return Promise.reject(new Error(this.props.translate('react.stockMovement.error.saveRequisitionItems.label', 'Could not save requisition items')));
         });
     }
-    this.setState({
+    this.setState(previousState => ({
       values: {
-        ...this.state.values,
-        lineItems: this.state.values.lineItems.map((item) => {
+        ...previousState.values,
```
kchelstowski, line 981: an often problem, nice catch

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -644,12 +654,19 @@ class AddItemsPage extends Component {
...
         },
         {
           label: this.props.translate('react.default.no.label', 'No'),
         },
       ],
+      afterClose: () => {
+        if (this.didUserConfirmAlert) {
+          onConfirm();
+        }
+      },
```
kchelstowski, line 669: nice workaround. Why was bumping up needed? Is it because `afterClose` is not available in lower versions?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
awalkowiak, line 445: Is the `!` at the beginning required?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -644,12 +654,19 @@ class AddItemsPage extends Component {
...
+          onClick: () => {
+            this.didUserConfirmAlert = true;
+          },
         },
         {
           label: this.props.translate('react.default.no.label', 'No'),
         },
       ],
+      afterClose: () => {
+        if (this.didUserConfirmAlert) {
```
awalkowiak, line 666: What will happen if the user confirms the first one, then not confirms the second one, and then retriggers again both modals and wants to close the first one, will it trigger the second one either way because `didUserConfirmAlert` is not reset to the initial `false` state?

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: Oh, good catch! Thanks

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: You're right, I didn't think about this case

kchelstowski (approved):

PR #4079, opened 2023-06-01
OBGM-406 Unable to access scaffolded pages

kchelstowski:
The issue was weird to me at first sight, because I was trying to find some weird parts in the controller/view, but then just read and compared scaffolding docs in Grails 3 vs Grails 1.3.9 and found the differences:
Grails 1.3.9:
Screenshot from 2023-06-01 18-25-30

Grails 3:
Screenshot from 2023-06-01 18-25-37

as you can see, in Grails 3 we don't have `list` action anymore and this is why the page was not working.

### DIFF
--- a/grails-app/conf/runtime.groovy
+++ b/grails-app/conf/runtime.groovy
@@ -375,7 +375,7 @@ openboxes {
                         [label: "partyRoles.label", defaultLabel: "Party roles", href: "/partyRole/list"],
                         [label: "partyTypes.label", defaultLabel: "Party types", href: "/partyType/list"],
                         [label: "person.list.label", defaultLabel: "People", href: "/person/list"],
-                        [label: "roles.label", defaultLabel: "Roles", href: "/role/list"],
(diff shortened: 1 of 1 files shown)


alannadolny (approved):

drodzewicz:
have you tried using just `/productComponent"` instead of `/productComponent/index"`.
As you know in HTML and Javascript keyword **index** can usually be omitted so I am wondering if it is the same for grails pages

grails-app/conf/runtime.groovy
drodzewicz, line 1: what about other pages like `/productSupplier/list` or `/paymentTerm/list` ?
Do you know if these paths don't require the same refactor?

kchelstowski:
@drodzewicz you might be right about getting rid of `/index`. Will test it out tomorrow. Thanks for suggestion

kchelstowski:
@drodzewicz you had a good grasp about removing `/index`. It is working without that, so I've removed it.

awalkowiak (approved):

PR #4080, opened 2023-06-02
OBPIH-5725 Improve initial load of stocklist list page

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -87,6 +87,11 @@ class LocationApiController extends BaseDomainApiController {
                 locations += locationService.getLocations(fields, params, isSuperuser, direction, currentLocation, currentUser, params.locationChooser ? true : false)
             }
         }
+
+        if (params.presentation == "toBaseJson") {
+            locations = locations?.collect { Location location -> location.toBaseJson()}
+        }
+
```
kchelstowski, line 94: if you do it like this, won't it anyway use the `toJson()` anyway while rendering when the Bootstrap will be invoked?

kchelstowski: so I assume it will invoke `toJson()`, so most of the properties which you got rid of in this line, will be `null` then (but the properties will be there)

grails-app/controllers/org/pih/warehouse/requisition/RequisitionTemplateController.groovy
```diff
@@ -460,27 +460,4 @@ class RequisitionTemplateController {
         redirect(action: "batch", id: params.id)
 
     }
-
-    private List<Location> getDepots() {
```
awalkowiak, line 464: @jmiranda the same two methods are on the `RequisitionController`, and seem to be not used too. Perhaps removing those would be useful too?

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
awalkowiak: @kchelstowski After doing this collect with `toBaseJson` we have now a list of maps instead of list of Locations, and it won't be processed by Location Marshaller

kchelstowski: thanks for the explanation

kchelstowski (approved):

jmiranda:
I still need some time for this one.

jmiranda (approved):
Approved as a temporary fix. 

For a long-term solution, here's the ticket I've mentioned a few times related to providing different representations of the same resource.

PR #4087, opened 2023-06-05
OBGM-461 Upgrade to npm 9

mdpearson:
This PR changes the version of `npm` we use to install javascript dependencies, from version 6 to version 9. Version 6 is no longer supported. The latest version, 9, still works with Node 14. Lucky us.

Note, the lock file format changes as part of this PR. The contents of the lock file are more or less the same.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -351,8 +351,16 @@ manifest {
 node {
     distBaseUrl = 'https://nodejs.org/dist'
     download = true
-    npmVersion = '6.14.18'
+    npmInstall.args = ['--loglevel', 'http']
+    npmInstallCommand = 'ci'
+    npmVersion = '9.6.7'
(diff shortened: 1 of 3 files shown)


jmiranda:
This one needs to be rebased. Also, this should come with some documentation on how to upgrade to NPM 9 in case it's not already there.

mdpearson:
@jmiranda The gradle node plugin will install the correct version of npm when you (or e.g. Bamboo) runs `grails prod war`; see the changes in the `node` block of `build.gradle`.

To upgrade `npm` on your local machine, there are any number of methods, see https://docs.npmjs.com/try-the-latest-stable-version-of-npm

awalkowiak:
Closing for now, we're going to comeback to it later.

PR #4092, opened 2023-06-06
OBGM-468 Transaction exception when creating a product type

kchelstowski:


grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
```diff
@@ -17,4 +20,19 @@ class ProductTypeService {
 
         return productType.sequenceNumber
     }
+
+    ProductType addProductType(ProductType productTypeInstance, Map params) {
+        productTypeInstance.productTypeCode = ProductTypeCode.GOOD
+        productTypeInstance.requiredFields = [ProductField.PRODUCT_CODE, ProductField.NAME, ProductField.CATEGORY, ProductField.GL_ACCOUNT]
```
jmiranda, line 26: I don't really like the hard-coded values here. If anything this should be set on the method when it's instantiated (similar to the create() method on the controller), not when we're saving it each time. In this case we're actually overriding whatever the user is setting in the UI which doesn't seem to make sense.

Also, now that we're in the service method we shouldn't be doing anything with the params. That should already be dealt with in the controller. 

I may need to pull down this PR and take a swipe at it because the existing code is confusing so I'm not seeing a path forward.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/product/ProductTypeService.groovy
kchelstowski: I also do not like, but I didn't want to refactor this kind of details, since it was existing code and I also didn't know the reason behind that. Let me know if I see this kind of spaghetti if I should go and refactor it deeply (and eventually let Katarzyna test it deeply if the behavior didn't change) or I should go the easiest way possible like here, so not touching the existing logic.
Thanks for the comment about params being passed to the service method. I left a comment in this PR: https://github.com/openboxes/openboxes/pull/4093 about my proposed approaches to that. Let me know your thoughts

jmiranda:
Resolved by configuration change in #4096

PR #4093, opened 2023-06-06
OBGM-413 Unable to create a new user

kchelstowski:


grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -118,24 +118,15 @@ class UserController {
...
         log.info "attempt to save the user; show form with validation errors on failure"
-        def userInstance = new User(params)
-
-        // Default value for active field on Person is set to True
-        // which is inherited by User
-        // but when creating a new user default value should be set to False
-        if (!params.hasProperty("active")) {
-            userInstance.active = false
+        User userInstance = new User(params)
+        User persistedUser = userService.addUser(userInstance, params)
```
jmiranda, line 122: Following our convention we should use save<DomainName> for the service.
```
saveUser(User user)
```
If there's a need to separate create and update then we can add two methods, but in this case we don't need that just yet.

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -460,4 +460,18 @@ class UserService {
         user.dashboardConfig = stringConfig
         return user.deserializeDashboardConfig()
     }
+
+    User addUser(User userInstance, Map params) {
```
jmiranda, line 464: I don't think we want to pass the params map to the service. Passing the params map is an anti-pattern because it couples the service to the web layer (params is actually an instance of GrailsParameterMap). 

Caveat: I know we do this a bunch in other places but since we're refactoring let's do it properly. 


According to the MVC design pattern (https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) 

> Controller 
> Accepts input and converts it to commands for the model or view.

That means that the controller is allowed (supposed) to do some of the work of processing the request and parameter processing is one of the things. Whether we're binding to a command object or to a domain instance, the controller should be allowed to do some basic manipulation of the model.

One thing I'd like to look into a bit more is how we're supposed to handle password hashing in Grails 3. In Grails 1, it went through a custom codec class (utils.PasswordCodec.groovy). In Grails 3 we should look into using bcrypt or spring security to handle this. 

But I think that should be reserved for another ticket since it comes with a lot more complexity (for example. Q: if we use a new method for hashing passwords, what happens to existing users A: they probably won't be able to log in until we hash those passwords as well).

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/core/UserService.groovy
kchelstowski: Ok, I get the point, I was also hesitant of doing that, but wanted to avoid any logic in the controller.
Let me know your thoughts of how I should handle that. My prepositions:

- do this:
```groovy
if (!params.hasProperty("active")) {
  userInstance.active = false
}
userInstance.password = params?.password?.encodeAsPassword()
userInstance.passwordConfirm = params?.passwordConfirm?.encodeAsPassword()
```
in the controller layer and pass only `userInstance` to the `addUser` method, so:
```groovy
User saveUser(User userInstance)
```

- extract the params that I'm dealing with then, so e.g. `params.password` like:
```groovy
User saveUser(User userInstance, String password, String passwordConfirm, Boolean hasActive)
```
and in the service deal with the logic above

- create some kind of command class, e.g. called `AddUserCommand`:
```groovy
class AddUserCommand {
  String password
  String passwordConfirm
  Boolean active
}
```
in the controller then bind the properties:
```groovy
AddUserCommand addUserCommand = new AddUserCommand(params)
```
and pass it to the service method like:
```groovy
User saveUser(User userInstance, AddUserCommand userCommand)
```
and deal with the logic from the first bullet in the service 

Let me know your thoughts guys. Happy to hear from you all @awalkowiak @alannadolny @drodzewicz @jmiranda

jmiranda:
This one is fixed by changes in #4096

grails-app/services/org/pih/warehouse/core/UserService.groovy
jmiranda: For now I think we'll use (1). In the future I think (3) might be best, but with a more generic UserCommand class so it can be reused in other places. Option (2) is also a good idea in some places but not with a generic saveUser method. It would be more appropriate for a UserService.updatePassword() method Although we may want something more semantically appropriate e.g. 

* resetPassword() would be appropriate if we were implementing 
* savePassword() might be better if we reused the same code when creating or updating user as well as resetting the password)   
 
And if we want to follow the grails convention it seems we'd pass parameters separately with encoding logic already applied.
```
securityService.doSomethingToPassword(params.id, params.password.encodeAsPassword(), ...)
```
or without if we wanted the 
```
securityService.doSomethingToPassword(params.id, params.password, ...)
```

Again we need to think outside of just the sliver of code we're looking at and figure out what we're trying to accomplish. In this case we're I guess we're creating a user. But what does it look like when we update that user, reset the password, etc. And then what does it look like when we're using spring security (or some other security plugin) OR an external service / API like okta, keycloak, etc.

PR #4095, opened 2023-06-07
OBGM-416 Unable to use product filter on add items to PO shipment

alannadolny:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1013,28 +1015,27 @@ class OrderService {
 
     def getProductsInOrders(String[] terms, Location destination, Location vendor) {
         return OrderItem.createCriteria().list {
+            createAlias('product', 'prod', JoinType.LEFT_OUTER_JOIN)
+            createAlias('product.synonyms', 'syn', JoinType.LEFT_OUTER_JOIN)
```
kchelstowski, line 1019: out of curosity it'd be worth also to check if here `prod.synonyms` would work

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1013,28 +1015,27 @@ class OrderService {
...
-                            ilike("productCode", term)
-                            ilike("description", "%" + term)
+            if (terms) {
+                terms.each { term ->
+                    term = term + "%"
+                    or {
+                        ilike("prod.name", "%" + term)
+                        and {
+                            ilike("syn.name", "%" + term)
+                                eq("syn.synonymTypeCode", SynonymTypeCode.DISPLAY_NAME)
```
kchelstowski, line 1035: could you fix the indent here, so it's not acting like it'd be nested inside `ilike`?

grails-app/services/org/pih/warehouse/order/OrderService.groovy
alannadolny: Yup, it works

kchelstowski (approved):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1013,28 +1015,27 @@ class OrderService {
 
     def getProductsInOrders(String[] terms, Location destination, Location vendor) {
         return OrderItem.createCriteria().list {
+            createAlias('product', 'prod', JoinType.LEFT_OUTER_JOIN)
```
awalkowiak, line 1018: Make these aliases a `product` and `synonym` instead of `prod` and `syn`

awalkowiak (changes requested):

awalkowiak (approved):

PR #4099, opened 2023-06-09
OBGM-410 Unable to receive inbound returns

alannadolny:
The reason for being unable to receive inbound returns was that, the `shipmentStatusCode` wasn't saved, so during clicking on the `receive` button our shipment still has `PENDING` status. I've added `.save(flush: true)` after the `publishEvent` and after that change, the status is changed to `SHIPPED` and it works fine. I tried it without `flush: true` but it doesn't work. I am a little bit curious if there is a better way for handling that save, maybe we can do something with saving within this published event?

### DIFF
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -1155,7 +1155,7 @@ class ShipmentService {
         }
 
         Holders.grailsApplication.mainContext.publishEvent(new ShipmentStatusTransitionEvent(shipmentInstance, ShipmentStatusCode.SHIPPED))
-
+        shipmentInstance.save(flush: true)
     }
 
 


kchelstowski (commented):
something doesn't feel right to me about this change - look, in line 1140, you have `save()` (without flush) and also we wanted to proceed the `save()` only if we DON'T have any errors (if statement in line 1128), but currently you've added an explicit save that would be proceeded even if the `shipmentInstance` HAS errors.
I don't know what causes this event not working properly in Grails 3, but we should definetely figure that out instead of throwing the additional save. If I find time, I will try to help you with this one today

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -1155,7 +1155,7 @@ class ShipmentService {
         }
 
         Holders.grailsApplication.mainContext.publishEvent(new ShipmentStatusTransitionEvent(shipmentInstance, ShipmentStatusCode.SHIPPED))
```
kchelstowski, line 1157: the only thing coming to my mind for now is that the event class is not Transactional, thus it doesn't commit the change (you call `this.shipmentStatusCode = shipmentStatusCode` and it should be persisted when the session flushes, but only if you have transaction open). 
You can try this:
- marking `ShipmentStatusTransitionEvent` as Transactional

jmiranda:
See #4118.

PR #4102, opened 2023-06-13
OBGM-506 Fix search bar on product list page

kchelstowski:


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -330,6 +330,8 @@ class ProductService {
                 }
             }
 
+            createAlias("synonyms", "synonym", JoinType.LEFT_OUTER_JOIN)
+
```
alannadolny, line 334: I am curious if this query returns only unique rows. @jmiranda described this issue here: [ticket](https://pihemr.atlassian.net/browse/OBGM-453?atlOrigin=eyJpIjoiNDAzMzZmMjc5YzU4NGRkMTgzY2RjNWU5MjA5MDNlYmMiLCJwIjoiaiJ9)

> To clarify my previous comment what I meant was that the searchProducts method (in line 13) returns duplicate products because there’s no listDistinct on the query and a list query with left outer joins usually produces duplicate results.

I can see that there is `setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)`, but it is applied only if `isCountQuery` is false. I am not sure if this issue occurs here and how the solution will look like in this case.

kchelstowski: + additionally the `listDistinct` never worked fine with the pagination, so I guess it wouldn't be achievable in an easy way

alannadolny (approved):

PR #4108, opened 2023-06-15
OBGM-486 Unable to upload documents on view shipment page

alannadolny:


awalkowiak (changes requested):
I did this regex search in grails-app folder: `log\.(info|debug|error)[ |(][^\"|^\']` and there are also two instances of `log.debug params` plus a couple of instances of logging maps. Please change the "debug params" and check if Map also triggers this issue.

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -48,7 +48,7 @@ class DashboardController {
 
     def showCacheStatistics() {
         def statistics = sessionFactory.statistics
-        log.info(statistics)
+        log.info("${statistics}")
         render statistics
     }
```
alannadolny, line 53: I am not sure if this method is used somewhere

kchelstowski (approved):
it LGTM, but I'm actually wondering if you have a clear answer why it is not working and has worked for Grails 1? What changed?

awalkowiak (approved):

PR #4109, opened 2023-06-15
OBGM-509 Hyperlinks to PO view page from PO shipment don't work 

alannadolny:


src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -91,7 +91,7 @@ const FIELDS = {
         getDynamicAttr: ({
           fieldValue,
         }) => ({
-          url: fieldValue && fieldValue.orderId ? `/order/show/${fieldValue.orderId}` : '',
+          url: fieldValue?.orderId ? stringUrlInterceptor(`/order/show/${fieldValue.orderId}`) : '',
```
kchelstowski, line 94: what about adding this url to our `urls.js` factory?
so to add:
```javascript
const CONTEXT_PATH = `/openboxes`
...
// ORDER
export const ORDER = `${CONTEXT_PATH}/order`
export const ORDER_SHOW = id => `${ORDER}/show/${id}`
```

awalkowiak: I am not sure if I prefer doing this all at once (replacing interceptor with url factory) or starting doing it right now while fixing bugs. But I guess if we start doing it now, we won't have to double-check if it works.

awalkowiak: But the context path cannot be hardcoded as it is in your example. It should be done as it is here: https://github.com/openboxes/openboxes/blob/feature/upgrade-to-grails-3.3.10/src/js/utils/apiClient.jsx#L111-L116

kchelstowski: are you sure? the `stringInterceptor` actually also adds the context path. I believe this url is not used for the `apiClient`, so it needs to have the context path manually added? Am I wrong?

kchelstowski: unless you meant the context path should not be hardcoded as `/openboxes`, but `window.CONTEXT_PATH`, then I can agree and sorry for the confusion in the comment above

kchelstowski (approved):

awalkowiak (approved):

PR #4111, opened 2023-06-16
OBGM-493 Unable to assign user's role for specific location manually

alannadolny:


grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -20,6 +20,7 @@ import org.pih.warehouse.core.User
 import javax.imageio.ImageIO as IIO
 import javax.swing.*
 import grails.gorm.transactions.Transactional
+import java.util.List
```
alannadolny, line 23: This error
```GroovyCastException: Cannot cast object '[Requestor]' with class 'java.util.ArrayList' to class 'java.awt.List' due to: groovy.lang.GroovyRuntimeException: Could not find matching constructor for: java.awt.List(org.pih.warehouse.core.Role)```
was caused by the fact that we used `List` from `java.awt` instead of `java.util`

**Review by kchelstowski — APPROVED** [review]

File: grails-app/controllers/org/pih/warehouse/user/UserController.groovy [hunk 2]
```diff
@@ -445,29 +446,8 @@ class UserController {
         User user = params.user.id ? User.get(params.user.id) : null
         Location location = params.location?.id ? Location.get(params.location.id) : null
         List<Role> roles = params.list("role.id").collect { roleId -> Role.get(roleId) }
-
-        // Update existing role
         LocationRole locationRole = LocationRole.get(params.id)
```
jmiranda, line 449: This is probably ok for now, but we should be careful about how we refactor these actions. I don't actually know what the original code is trying to do (I'm to blame as the original author) so adding a method with a bunch of arguments might not be the right solution. 

In order to refactor properly 

1. We need to be very thoughtful about what the interface / API is supposed to be doing here. For example, are we trying to do multiple actions? If so, maybe we need multiple methods.
1. Maybe think about passing each param separately and loading objects in the service that way we're doing all reading / writing in the same Hibernate session.

cc @awalkowiak @alannadolny @kchelstowski

jmiranda: We also need to consider the return value for new service methods. Again this is why I'd prefer to leave controllers alone unless we have very straightforward refactorings.

jmiranda (commented):
Should not have been merged.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
kchelstowski: @jmiranda I have not noticed this has not been removed. The reason I merged it alone, was that in the morning Katarzyna had problem with obdev3 not working and we thought a merge could help, before Artur is online (he starts work later than us). I wouldn't have merged it if it hasn't been the case. Sorry about that.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
alannadolny: This import was removed in the PR mentioned above.

PR #4112, opened 2023-06-16
OBGM-516 Unable to run inventory by location report

alannadolny:


src/main/groovy/org/pih/warehouse/report/MultiLocationInventoryReportCommand.groovy
```diff
@@ -35,11 +35,11 @@ class MultiLocationInventoryReportCommand implements Validateable {
     }
 
     Boolean getIsActionRun() {
-        return actionButton?.equalsIgnoreCase("run")
+        return actionButton?.equalsIgnoreCase("run") ?: false
     }
 
     Boolean getIsActionDownload() {
-        return actionButton?.equalsIgnoreCase("download")
+        return actionButton?.equalsIgnoreCase("download") ?: false
```
alannadolny, line 42: This part was throwing that `isActionRun` and `isActionDownload` cannot be null, so as a solution, I am returning false when `actionButton` is null.

src/main/groovy/org/pih/warehouse/report/MultiLocationInventoryReportCommand.groovy
```diff
@@ -35,11 +35,11 @@ class MultiLocationInventoryReportCommand implements Validateable {
     }
 
     Boolean getIsActionRun() {
-        return actionButton?.equalsIgnoreCase("run")
+        return actionButton?.equalsIgnoreCase("run") ?: false
```
kchelstowski, line 38: @alannadolny by default all Grails properties are `nullable: false`. To make it work "in a clean way", you need to add the transients to the constraints and mark them as `nullable: true`.
So to add:
```groovy
static constraints = {
...
isActionRun(nullable: true)
isActionDownload(nullable: true)
}

kchelstowski (approved):

awalkowiak (changes requested):

awalkowiak (approved):

PR #4113, opened 2023-06-19
OBGM-487 Unable to upload documents on product page 

alannadolny:


grails-app/conf/application.yml
```diff
@@ -172,6 +172,9 @@ grails:
         cache:
             maxsize: 1000
     controllers:
+        upload:
+            maxFileSize: 2097152
+            maxRequestSize: 2097152
```
alannadolny, line 177: It's a maximum of 2MB for uploading files (2 * 1024 * 1024 = 2097152)

Note about the right value for maxFileSize and maxRequestSize:
> It depends on the application type, but it's generally a bad idea to keep them too high. If you really need to upload big files, probably you should develop (or use) a dedicated (and separated) service.

awalkowiak (approved):

PR #4115, opened 2023-06-19
OBGM-510 Unable to add items to PO shipment through import

alannadolny:


kchelstowski (approved):
Why did it differ from the Grails 1? Was there any code change with that? Just out of curiosity

alannadolny:
The methods from `MultipartFile` are different. For now, it looks like that:
image

PR #4116, opened 2023-06-19
OBGM-386 Move recall logic to service layer

awalkowiak:
To fix issue with not having data from not yet finished session while refreshing product availability for all locations after recall or revert recall

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
@@ -10,7 +10,6 @@
 package org.pih.warehouse.inventory
 
 import grails.converters.JSON
-import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
 import groovy.time.TimeCategory
 import grails.plugins.csv.CSVWriter
@@ -32,7 +31,6 @@ import grails.core.GrailsApplication
 import java.text.DateFormat
 import java.text.SimpleDateFormat
 
-@Transactional
 class InventoryItemController {
 
     def dataService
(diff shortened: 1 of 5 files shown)


grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -32,7 +31,6 @@ import grails.core.GrailsApplication
 import java.text.DateFormat
 import java.text.SimpleDateFormat
 
-@Transactional
```
kchelstowski, line 35: On one hand there is probably no other way to make this work, but on the other hand I'm afraid of that this change considering the other methods, especially the saving ones, that do not use the flush, won't work.

With the setting `allow_update_outside_transaction` we will only auto-commit the change (which is also not good) if we use the flush (this is probably why you've added `flush: true` to the create method).

I would also be aware of the `InventoryItem` `publishPersistenceEvent` - without transaction and with auto-commit (`flush: true` on update) we might face some weird exceptions when the `afterUpdate` would be executed. Have you made sure those work? 

I think the best idea (and the "cleanest" one) would be just to live with the exceptions that might be thrown, create tickets and move the other methods to the service layer.

awalkowiak:
@kchelstowski @jmiranda this one can be re-reviewed

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -3425,4 +3425,20 @@ class InventoryService implements ApplicationContextAware {
             }
         }
     }
+
+    void recallInventoryItem(InventoryItem inventoryItem) {
```
jmiranda, line 3429: Given how massive the InventoryService class is we should be thinking longer term about how to refactor this into smaller classes. 

I think the obvious options are 
* Domain (Product, Inventory, InventoryItem) which we used and caused this mess
* Functional area (transfer vs movement vs recall) which seems way better but might have some overlap that we need to contend with 
* Other suggestions?

jmiranda (commented):
Approved. See comment for tech-huddle topic.

PR #4117, opened 2023-06-20
OBGM-471 Fix revising items on outbound stock movements

awalkowiak:
- bring back checking if there is need to undo previous changes
- fix requisition refresh
- minor NPE fix

### DIFF
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -687,7 +687,7 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
     }
 
     Set<RequisitionItem> getSubstitutionItems() {
-        return requisitionItems.findAll {
+        return requisitionItems?.findAll { RequisitionItem it ->
             it.requisitionItemType == RequisitionItemType.SUBSTITUTION
         }
     }
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1952,7 +1952,9 @@ class StockMovementService {
 
                 log.info 'Removing previous changes, picklists and shipments, if present'
                 removeShipmentAndPicklistItemsForModifiedRequisitionItem(requisitionItem)
-                requisitionItem.undoChanges()
+                if (requisitionItem.isChanged() || requisitionItem.isCanceled()) {
+                    requisitionItem.undoChanges()
+                }
 
                 log.info "Revising quantity for ${requisitionItem.id}"
                 requisitionItem.changeQuantity(
@@ -2364,7 +2366,7 @@ class StockMovementService {
     }
 
     void createMissingShipmentItems(StockMovement stockMovement) {
-        Requisition requisition = stockMovement.requisition.refresh()
+        Requisition requisition = Requisition.get(stockMovement.id)
 
         if (requisition) {
             Shipment shipment = Shipment.findByRequisition(requisition)


jmiranda:
I am looking forward to hearing more about this one.

grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
```diff
@@ -687,7 +687,7 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
     }
 
     Set<RequisitionItem> getSubstitutionItems() {
-        return requisitionItems.findAll {
+        return requisitionItems?.findAll { RequisitionItem it ->
```
jmiranda, line 690: I'm ok with these changes but can you explain why both of these were necessary?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1952,7 +1952,9 @@ class StockMovementService {
 
                 log.info 'Removing previous changes, picklists and shipments, if present'
                 removeShipmentAndPicklistItemsForModifiedRequisitionItem(requisitionItem)
-                requisitionItem.undoChanges()
+                if (requisitionItem.isChanged() || requisitionItem.isCanceled()) {
+                    requisitionItem.undoChanges()
```
jmiranda, line 1956: ?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2364,7 +2366,7 @@ class StockMovementService {
     }
 
     void createMissingShipmentItems(StockMovement stockMovement) {
-        Requisition requisition = stockMovement.requisition.refresh()
+        Requisition requisition = Requisition.get(stockMovement.id)
```
jmiranda, line 2369: I assume we're getting this requisition from the session so we should be getting the same object. Curious why we used refresh in the first place (https://pihemr.atlassian.net/browse/OBPIH-1257). 

As such, I wonder if this
```
Requisition requisition = stockMovement.requisition
```
would suffice. 

One worry I have is with stockMovement.id which could potentially represent another object like order, shipment, etc. Would that ever be the case? And if not, we should be explicit about handling these cases i.e. throw an exception.

jmiranda: But that could be done as a post-migration refactoring. 

I'm not a big fan of these methods, in general. For one, they are too generically named and should be part of a more specific to the work we're trying to accomplish i.e. 
```
void refreshPicklist(RequisitionItem requisitionItem)
```

jmiranda (commented):
Btw, I'm ok with these changes so feel free to merge if you get an approval.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski, line 2369: @jmiranda I assume the `refresh()` (now the fetch) is done to "undo" any changes introduced to requisition object, but it needs to be confirmed by @awalkowiak 
Since code LGTM, I will merge this one, so that it could be tested before the end of the sprint, and we can discuss your doubts later if it sounds OK.

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
awalkowiak: I checked locally the `stockMovement.requisition` and looks like I had weird glitch or whatever, because all versions:
```
stockmovement.requisition
stockmovement.requisition.refresh()
Requisition.get(stockmovement.id)
```
now all return the same state of requisitionItems. :thinking: 
I'll change it to the `stockmovement.requisition` and push within my next PR.

PR #4119, opened 2023-06-21
OBGM-505 Unable to add items and adjustments to PO

alannadolny:


grails-app/views/purchaseOrder/_enterOrderDetails.gsp
```diff
@@ -2,10 +2,10 @@
 <%@ page import="org.pih.warehouse.order.OrderType" %>
 <%@ page import="org.pih.warehouse.order.OrderTypeCode" %>
 
+<g:applyLayout name="custom">
 <html>
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
-<meta name="layout" content="custom" />
```
alannadolny, line 8: This meta tag doesn't apply the layout on those two files. During searching for possible solutions/problems to solve, I found this [stackOverflow discussion](https://stackoverflow.com/questions/9629155/using-3-layouts-on-same-gsp-page) which is not too much connected to this issue, but I found there `g:applyLayout` and it works as it should.

grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -509,7 +507,7 @@
                * TODO here, we may need to elsewhere -- should we escape all
                * TODO localized strings? Further discussion at OBGM-343.
                */
-              if ($("#validationCode").val() == 'WARN' && !confirm(_.escape(htmlDecode("${g.message(code: 'orderItem.warningSupplier.label')}")))) {
+              if ($("#validationCode").val() == 'WARN' && !confirm(htmlDecode("${g.message(code: 'orderItem.warningSupplier.label')}"))) {
```
alannadolny, line 510: Importing the lodash library was throwing error. After removing it I found there is an reference to global UMD variable. After checking how this message looks like I found that, it's broken:
image
There was a [ticket](https://pihemr.atlassian.net/browse/OBPIH-4561?atlOrigin=eyJpIjoiYjY5YzJlYjc4ZWQ4NDBhNWI5MzU5YzQzMjEwZmQzZmQiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) about the quotes.
After removing this method from lodash I found that this message looks correctly:
image
In the comment above this code is a comment about further discussion / investigation about it.
Here is the [ticket](https://pihemr.atlassian.net/browse/OBGM-343?atlOrigin=eyJpIjoiZDM0ZTJhZTJiNjJhNDc0N2E1NzViZmIwZDFjZDdlMGMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

awalkowiak (approved):
I am not sure if this is a correct usage of `applyLayout` in gsp templates, but since it works and unblocks work I am merging it

PR #4120, opened 2023-06-22
OBGM-410 Derived property needed to be set in beforeUpdate event handler

jmiranda:


kchelstowski (commented):
oh gosh, it could not have been an easier workaround than this :see_no_evil: 
LGTM, good job

PR #4121, opened 2023-06-22
OBGM-479 Remove unnecessary extend causing expirydate tag collision

awalkowiak:


kchelstowski (approved):

jmiranda (approved):
LGTM. I assume we have more work to do with OBGM-418 correct?

awalkowiak:
@jmiranda actually, IMHO none. Looks like everything other works properly.

PR #4123, opened 2023-06-23
OBGM-465 Unable to import product associations

alannadolny:


src/main/groovy/org/pih/warehouse/importer/OutboundStockMovementExcelImporter.groovy
```diff
@@ -58,7 +56,8 @@ class OutboundStockMovementExcelImporter extends AbstractExcelImporter {
     }
 
     List<Map> getData() {
-        return excelImportService.convertColumnMapConfigManyRows(workbook, columnMap, null, propertyMap)
+        Holders.grailsApplication.mainContext.getBean("excelImportService")
+                .convertColumnMapConfigManyRows(workbook, columnMap, null, null, propertyMap)
```
alannadolny, line 60: There is signature of `convertColumnMapConfigManyRows` method:
image
So, the second `null` was missing and I got error connected to being unable to find appropriate signature

awalkowiak (approved):

PR #4124, opened 2023-06-26
OBGM-469 Fix generating product code

awalkowiak:


grails-app/conf/application.yml
```diff
@@ -338,7 +338,7 @@ openboxes:
         productCode:
             generatorType: "SEQUENCE" # IdentifierGeneratorTypeCode.SEQUENCE
             delimiter: "-"
-            format: '$\{productTypeCode}$\{delimiter}$\{sequenceNumber}' # single quote for curly braces escape character
+            format: "${productTypeCode}${delimiter}${sequenceNumber}"
```
jmiranda, line 341: I was hoping it was going to be this easy.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -96,7 +95,7 @@ class ProductController {
         [products: product]
     }
 
-
+    @Transactional
```
jmiranda, line 98: Is this batchSave method being used somewhere?

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -48,21 +48,29 @@ import org.pih.warehouse.LocalizationUtil
  */
 class Product implements Comparable, Serializable {
 
-    def beforeInsert = {
+    def beforeInsert() {
         createdBy = AuthService.currentUser
```
jmiranda, line 52: In the Shipment domain I changed the convention to include updatedBy in the insert as well. I had assumed we were already doing this everywhere else but it looks like I was wrong.

Topics for tech-huddle maybe: 
- Do you think that should be applied everywhere? 
- If so do you think we can move this to a generic solution 
- Auditable abstract class or trait (assuming that's what traits are for) that adds these methods to domains that have these audit fields?

The first of these could be done during MVP1, but the other points should be MVP2/post-migration.

Now that we might have more consistency with the Grails 3 dirty properties API we could also try to make these persistence events more fine-grained. There are only a few properties on Product that should cause a refresh to InventorySnapshot. Same with other domains for ProductAvailability.

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -326,9 +326,9 @@ class IdentifierService {
      *          - empty productIdentifierFormat (will be generated basing on the openboxes.identifier.product.format)
      * */
     String generateProductIdentifier(ProductType productType) {
-        if (productType?.id == ConfigurationHolder.config.openboxes.identifier.defaultProductType.id) {
+        if (productType?.id == Holders.config.openboxes.identifier.defaultProductType.id) {
             // If product type is systems default then generate basing on the generator type
-            def generatorTypeConfig = ConfigurationHolder.config.openboxes.identifier.productCode.generatorType
+            def generatorTypeConfig = Holders.config.openboxes.identifier.productCode.generatorType
```
jmiranda, line 331: Do we have defaults for these in application.yml? 

We had an issue on the Dominica server a few weeks ago (I think I finally fixed it last week). It feels like we should discuss the expected vs actual behavior for the identifier config as part of a tech huddle.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: I just realized that you're just applying @Transactional to all methods so you can ignore this comment.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -185,13 +184,13 @@ class ProductController {
...
 
-        if (!productInstance.hasErrors() && productInstance.save(flush: true)) {
+        if (!productInstance.hasErrors() && productService.saveProduct(productInstance)) {
             log.info("saved product " + productInstance.errors)
             flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'product.label', default: 'Product').decodeHTML(), format.product(product: productInstance).decodeHTML()])}"
             sendProductCreatedNotification(productInstance)
         }
 
-        render(view: "edit", model: [productInstance: productInstance, rootCategory: productService.getRootCategory(), locationInstance: location])
+        redirect(action: "edit", params: [id: productInstance.id])
```
jmiranda, line 193: What about when errors occur?

jmiranda (commented):
Nothing specific to address other than maybe thinking about the render -> redirect change. Not sure if that's the behavior we want for the unhappy case.

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
awalkowiak: You're right... I'll restore this

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
awalkowiak: Yeah, it's SEQUENCE by default, needs to be changed according to needs. But I believe now everything works fine in that matter (in grails 1 instances)

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
awalkowiak: fixed

kchelstowski (approved):

jmiranda (approved):

PR #4125, opened 2023-06-27
OBGM-534 Substitutes are duplicated on pack and send page 

alannadolny:
Explanation: https://pihemr.atlassian.net/browse/OBGM-534?atlOrigin=eyJpIjoiMjQ5OTgzODA1MGQ3NDEyODk4NDQ2ODk5ODIxYzRjNzIiLCJwIjoiaiJ9

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -2387,6 +2387,7 @@ class StockMovementService {
                 ShipmentItem shipmentItem = createOrUpdateShipmentItem(requisitionItem)
                 shipmentItem.container = container
                 shipment.addToShipmentItems(shipmentItem)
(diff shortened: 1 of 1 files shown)


kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2395,6 +2395,7 @@ class StockMovementService {
                 shipmentItems.each { ShipmentItem shipmentItem ->
                     shipment.addToShipmentItems(shipmentItem)
                 }
+                shipment.save(flush: true)
```
awalkowiak, line 2398: And for the first case (`if`) it is not required?

alannadolny: Oh, you're right. I didn't see that before 🙉

PR #4126, opened 2023-06-27
OBGM-527,544 Fix creating and listing invoices

awalkowiak:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -157,6 +157,10 @@ class InvoiceService {
         return invoiceItemCandidates
     }
 
+    def saveInvoice(Invoice invoice) {
+        invoice.save(flush: true, failOnError: true)
```
kchelstowski, line 161: I know it was existing like that, but we've had a discussion not a long time ago, that we should avoid throwing `flush: true` everywhere. It necessary here?

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -157,6 +157,10 @@ class InvoiceService {
         return invoiceItemCandidates
     }
 
+    def saveInvoice(Invoice invoice) {
```
kchelstowski, line 160:
```suggestion
    Invoice saveInvoice(Invoice invoice) {
```

alannadolny (approved):

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
awalkowiak: Looks like it was not required.

PR #4129, opened 2023-06-29
OBGM-551 Unable to Place PO

alannadolny:
Roles passed from the application.yaml weren't properly compared with the values from `getEffectiveRoles(user)` (it always returned `false`, so we didn't have appropriate privileges to perform actions).

### DIFF
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -458,7 +458,7 @@ class OrderService {
     boolean canApproveOrder(Order order, User userInstance) {
         if (isApprovalRequired(order)) {
             List<RoleType> defaultRoleTypes = grailsApplication.config.openboxes.purchasing.approval.defaultRoleTypes
-                    .collect { RoleType."$it" }
+                    .collect { RoleType.valueOf(it) }
             return userService.hasAnyRoles(userInstance, defaultRoleTypes)
         }
         return Boolean.TRUE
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -458,6 +458,7 @@ class OrderService {
     boolean canApproveOrder(Order order, User userInstance) {
         if (isApprovalRequired(order)) {
             List<RoleType> defaultRoleTypes = grailsApplication.config.openboxes.purchasing.approval.defaultRoleTypes
+                    .collect { RoleType."$it" }
```
awalkowiak, line 461: Can you first try to define it as `defaultRoleTypes: [org.pih.warehouse.core.RoleType.ROLE_APPROVER]` instead of `"ROLE_APPROVER"` in the `application.yml`? If that will work this should be the change. Otherwise, if that won't work, then I think we should discuss it in the tech huddle. The change to string in the `applicaiton.yml` was a quick temporary hotfix.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
jmiranda: Also, if that does not work I recommend we change the enum lookup from
```
RoleType."$it"
```
to 
```
RoleType.valueOf(it)
```

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
jmiranda: Actually an even better way to do this would be to use the approach suggested here https://grails.org/blog/2016-08-31.html

Either inject the value into the class 
```
@Value('${openboxes.purchasing.approval.defaultRoleTypes:[RoleType.ROLE_APPROVER]}')
List<RoleType> defaultRoleTypes
```
or pull the property from the config object
```
List<RoleType> defaultRoleTypes = grailsApplication.config.getProperty("openboxes.purchasing.approval.defaultRoleTypes", List, [RoleType.ROLE_APPROVER])
```

Unfortunately neither of these approaches seem to coerce the object into the proper type List<RoleType>.

So let's just use the `collect { return RoleType.valueOf(it) }` approach.

awalkowiak (approved):

PR #4131, opened 2023-06-30
OBGM-552 Unable to add items to PO through import

alannadolny:


kchelstowski (approved):

jmiranda (approved):
LTGM, I guess. 

However, I don't really understand how the fix solves the problem because it's not obvious what is happening. 

@alannadolny could you please explain the issue in the ticket. What is different about YML that causes the issue. And also make sure this works if we were to override the configuration in openboxes-config.groovy.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4132, opened 2023-07-03
OBGM-530 Fix missing requisition sourceType on create request

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
drodzewicz, line 1: In short, the issue was that
```groovy
 stockMovement.sourceType = params.sourceType as RequisitionSourceType
```
was assigning a **null** value to the sourceType because `params.sourceType` is **null**, and also because it is a value from the requests body and not a param.

Although it looked like all I had to do was change `params.sourceType` to `request.JSON.sourceType` as we have done before, I noticed that when using the "inferred data binding" like 
```groovy
def create(StockMovement stockMovement) { ...
```
we lose access to the `requets.JSON` data.

I did an experiment and removed the `StockMovement stockMovemen` from the parameters and after that, I had access to `request.JSON` data.
Just wanted to point this out, not sure if it was already discussed/noticed in my absence and if someone knows what's up the let me know.

cc @awalkowiak @kchelstowski @alannadolny @jmiranda

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -114,8 +114,6 @@ class StockMovementApiController {
     def create(StockMovement stockMovement) {
         // Detect whether inbound or outbound stock movement
         def currentLocation = Location.get(session.warehouse.id)
-
-        stockMovement.sourceType = params.sourceType as RequisitionSourceType
```
kchelstowski, line 118: this is because `parseRequest: true` doesn't work anymore in Grails 3 (or even in 2), so as you said -> changing `params` to `request.JSON` should be done here, but since you are binding the data in the params, you "lose' the access to that and this is expected behavior, so LGTM

kchelstowski (approved):

alannadolny (approved):

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
jmiranda: @drodzewicz Good catch! Yes, the request.JSON object cannot be accessed twice so once the data is accessed in the command object binding (i.e. populating the StockMovement object), then we're done. The best solution might be to include RequisitionSourceType as a nullable property on StockMovement so it gets bound properly. 

To make it backward compatible, we'd need to figure out what was going on in Grails 1. I don't expect this to be true, but maybe it was adding all request.JSON to params so you could access it in either way.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
jmiranda: Boom goes the dynamite. Thanks for the reminder @kchelstowski.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
jmiranda: Nevermind, it looks like RequisitionSourceType is already on StockMovement, so just removing that line (as you did) should take care of the issue. Nice work, guys.

jmiranda (commented):
LGTM

PR #4133, opened 2023-07-03
OBGM-554 Unable to save adjustments when add them to PO 

alannadolny:


grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -132,8 +132,10 @@
                     </g:form>
                 </div>
                 <div id="add-adjustment" class="box">
-                    <g:set var="canManageAdjustments" value="${(order?.status >= OrderStatus.PLACED) && (isApprover
-                            || order?.status == OrderStatus.PENDING)}"/>
+                    <g:set
+                            var="canManageAdjustments"
+                            value="${(order?.status >= OrderStatus.PLACED && isApprover) || (order?.status == OrderStatus.PENDING)}"
```
kchelstowski, line 137: I don't know the context, but it feels like this `order?.status` only works "by a fortune". It seems to be comparing statuses by alphabetic order, not by the `sortOrder` as it was probably meant to be.
Could you double check that?
In my opinion it works by a fortune, but in general it should be written using:
```groovy
OrderStatus.compare(order?.status, OrderStatus.PLACED)
```
Be aware, that by using the `compare` method, a `NullPointerException` might be thrown, because we don't have null safe operators there:
https://github.com/openboxes/openboxes/blob/1e77b282da20c34b75ebec47b896e3528dae095e/src/groovy/org/pih/warehouse/order/OrderStatus.groovy#L33-L35

alannadolny: Yes, it was compared by alphabetical order (`>= placed` was returning correct values -> `placed`, `received`, `rejected`). I changed it as you suggested to avoid future problems connected to the order of those statuses.

awalkowiak (approved):

grails-app/views/purchaseOrder/_showOrderItems.gsp
jmiranda: > It seems to be comparing statuses by alphabetic order, not by the sortOrder as it was probably meant to be.

I don't think this is true. Or at least it was not true in Grails 1. 

The only reason it might be true is if within a GSP the order.status is somehow not being coerced into an OrderStatus enum and Grails thinks it's a String, forcing the OrderStatus.PLACED and OrderStatus.PENDING to be compared as Strings.

jmiranda (commented):
I'd like for someone to review this and confirm the "enums compared as strings" behavior. We have enum comparisons all over the place so it would be annoying to have to change this everywhere if it's not necessary.

jmiranda:
In Java 7, the following is true regarding enum compareTo.

> The natural order implemented by this method is the order in which the constants are declared.

https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#compareTo%28E%29

That means the order defined in the OrderStatus enum is the natural order. So this change was unnecessary since the order is correct here.

```
    PENDING(10, StatusType.WARNING),
    APPROVED(20, StatusType.PRIMARY),
    PLACED(30, StatusType.PRIMARY),
    PARTIALLY_RECEIVED(40, StatusType.PRIMARY),
    RECEIVED(50, StatusType.PRIMARY),
    COMPLETED(60, StatusType.SUCCESS),
    CANCELED(70, StatusType.DANGER),
    REJECTED(80, StatusType.DANGER)
```
Incidentally, this is also true for Java 8. 

https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#compareTo-E-

There might be something fishy going on with how the <g:set> tag is evaluating the expression, but comparing enums using relational operators (>, >=, ==, <=, <) is not the problem.

jmiranda:
@awalkowiak @alannadolny @kchelstowski @drodzewicz

kchelstowski:
@jmiranda that's very interesting. Thanks for pointing that out. I checked it in a console and it seems to be true what you've written. I remember we had a problem in some `createCriteria` where we compared the enum via `ge`/`lt` and it was giving some weird results (@awalkowiak should remember, as he caught it in one place).

Anyway to be always safe (e.g. accidentally we don't follow the `sortOrder` order while adding some new enum values or anything) I think we should override the `compareTo` methods to look at sort orders. What do you think guys?

PR #4134, opened 2023-07-03
OBGM-177 Dashboard > Encountered positional parameter near line 1, column 98 in HQL

alannadolny:
Those two places are the only ones that I found and they still had positional parameters. (It's not HQL, but I also changed those to follow the convention that we have)


awalkowiak (approved):

jmiranda (commented):
lgtm

PR #4136, opened 2023-07-04
OBGM-574 Disable allow_update_outside_transaction Hibernate setting

kchelstowski:
I know I could've deleted this setting at all, but I prefer to leave it with the `false` flag, because figuring out what Hibernate settings for an exact version are enabled/disabled by default, is a pain in the neck, so it will provide a better readability and understanding, what is enabled/disabled

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -209,7 +209,7 @@ slf4j.detectLoggerNameMismatch: true
 
 ---
 hibernate:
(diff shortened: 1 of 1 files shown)


kchelstowski:
I'm merging this one - the faster we find bugs related to that, the better.

jmiranda (commented):
Agree with @kchelstowski that we should be setting configuration explicitly (with very descriptive commit messages).

PR #4142, opened 2023-07-06
OBGM-299 Fix entering scaffolded create page for Role domain

awalkowiak:


drodzewicz (approved):

jmiranda (approved):
I'm ok with this, but why I would like to know why this fails in Grails 3 and not in Grails 1?

kchelstowski (approved):

PR #4144, opened 2023-07-06
OBGM-578,580,582 Move save and delete to service layer for BudgetCode, GlAccount and PaymentTerm

awalkowiak:


drodzewicz (approved):

jmiranda (commented):
As I mentioned in previous PRs, I don't want to create a service for every domain class. 
https://tedvinke.wordpress.com/2017/04/04/grails-anti-pattern-everything-is-a-service/

So I think we need to come up with a better strategy during a tech huddle. These are all things related to accounting, so maybe AccountingService makes the most sense. Or if we do decide we want to create a separate class for each domain then we should consider implementing the DAO or Repository pattern.

I just found an SO article on Domain Drive Development. I have no idea what it entails but we should add that to the discussion as well. 
https://stackoverflow.com/questions/2378069/ddd-with-grails

Anyway, we need to do some more thinking about this before proceeding. In the meantime, I would prefer using @Transactional on controller actions that are basically scaffolding. I know that's not ideal, but until we come up with a decent solution we shouldn't be refactoring code.

awalkowiak:
@jmiranda I am keeping this branch locally, and I created a new PR with simple `@Transactional` annotations here: https://github.com/openboxes/openboxes/pull/4148 (I also added it for 4 other tickets)

PR #4145, opened 2023-07-07
OBGM-560 Fix association between transaction and shipment while rollbacking and fix refresh product availability when deleting outbound return

kchelstowski:
As this ticket drove me nuts, a little description (as always :smile: ):

**1. rollback problem** - we have a bidirectional relationship between `Transaction` and `Shipment`. The problem there was that we were not assigning `outgoingShipment` to the `Transaction` instance, we were only doing it for one direction, so for shipment like:
```groovy
shipmentInstance.addToOutgoingTransactions(debitTransaction)
```
For correct bidirectional relationships, we have to add it for both sides.
Grails is that clever, that when doing `addToOutgoingTransactions`, it was automatically assigning the `outgoingShipment` for the `debitTransaction` also, but it was not persisted. In grails 1 it was working due to OSIV, that was catching those "dirty" changes and here, it doesn't work, because we saved the `debitTransaction` a few lines above.

What also fixed the issue was removing this:
```groovy
 if (!debitTransaction.save()) {
       log.info "debit transaction errors " + debitTransaction.errors
       throw new ValidationException("An error occurred while saving ${debitTransaction?.transactionType?.transactionCode} transaction", debitTransaction.errors)
}
```
because then, the `debitTransaction` was saved at the end of the session and the `outgoingShipment` set by the:
```groovy
shipmentInstance.addToOutgoingTransactions(debitTransaction)
```
was caught.

We could either change it to `validate()` and let Grails do the work with associations, but I think the path I went with assigning it explicitly is a more proper solution.

**2. quantity available not updating when deleting the shipment:**

This drove me nuts.... since we were refreshing product availability asynchronously and the session that was deleting picklist items was not yet flushed and the `productAvailabilityService` was creating its own session and transaction, it didn't know "what is going on" in the first session, that is deleting the entities.... so it was thinking that picklist item is still in the DB (because the deletion was still in batch in first session)....
After long debugging it turned out, that we have missing properties for product availability jobs, which was the delay.....
Adding the delay, fixed the issue, because the first session "managed" to finish and to be flushed before the second session (from the job) initialized.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -21,31 +21,37 @@ import org.pih.warehouse.shipping.ShipmentStatusCode
 
 class Order implements Serializable {
 
-    def beforeInsert = {
+    def beforeInsert() {
         createdBy = AuthService.currentUser
     }
 
-    def beforeUpdate = {
+    def beforeUpdate() {
         updatedBy = AuthService.currentUser
     }
 
-    def publishRefreshEvent = {
+    def publishRefreshEvent() {
         if (isPurchaseOrder) {
             Holders.grailsApplication.mainContext.publishEvent(new RefreshOrderSummaryEvent(this))
         }
     }
 
-    def publishRefreshEvenBeforeDelete = {
+    def publishRefreshEvenBeforeDelete() {
         if (isPurchaseOrder) {
             Holders.grailsApplication.mainContext.publishEvent(new RefreshOrderSummaryEvent(this, true))
         }
     }
 
-    def afterInsert = publishRefreshEvent
+    def afterInsert() {
+        publishRefreshEvent()
+    }
 
-    def afterUpdate = publishRefreshEvent
+    def afterUpdate() {
+        publishRefreshEvent()
+    }
 
-    def beforeDelete = publishRefreshEvenBeforeDelete
+    def beforeDelete() {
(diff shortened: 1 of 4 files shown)


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -21,31 +21,37 @@ import org.pih.warehouse.shipping.ShipmentStatusCode
 
 class Order implements Serializable {
 
-    def beforeInsert = {
+    def beforeInsert() {
```
kchelstowski, line 24: closures for events work also in Grails 3, but just the code is "gray", so since I was "touching" this code while debugging, I decided to change it to the methods, to follow the Grails 3 convention

drodzewicz: If I understand correctly this change to closures doesn't affect the functionality of the event and you changed it because it is a preferable way to do it in grails 3 and not because it fixes something, right?

kchelstowski: @drodzewicz yes, exactly. It is working, but if we take a look at grails 3 docs, they (event) are specified as methods:
https://docs.grails.org/3.0.x/guide/GORM.html#eventsAutoTimestamping

drodzewicz (approved):

alannadolny (approved):

awalkowiak (approved):

PR #4148, opened 2023-07-07
OBGM-578:583 Add Transactional on controller actions that require it

awalkowiak:


drodzewicz (approved):
Since Justin advised against creating a separate service for each domain adding @Transactional should be good enough for this moment.
Fingers crossed that this does not cause problems in the future :crossed_fingers:

kchelstowski (approved):
we all know it's not the best approach, but I guess it could wait until MVP instead of spending time figuring out what services we should create or we shouldn't create for those micro classes

PR #4153, opened 2023-07-10
OBGM-558 Fix accessing nested map values

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/PartialReceivingApiController.groovy
```diff
@@ -171,7 +171,7 @@ class PartialReceivingApiController {
 
             // Bind the container
             PartialReceiptContainer partialReceiptContainer =
-                    partialReceipt.findPartialReceiptContainer(containerMap["container.id"])
+                    partialReceipt.findPartialReceiptContainer(containerMap.container?.id)
```
drodzewicz, line 174: I have come across this issue already several times where we are trying to access nested map values using strings like
`map["key1.key2.key3"]` or `map.get("key1.key2.key3")`.
I guess the newest version of Groovy has removed this feature so the alternative way is to use the key chaining with null safety.
I was thinking of maybe creating a spike ticket to investigate and fix places where this exact problem can be found, but I am not sure if it is a good idea. I guess we'll talk about it in the tech-hiddle

awalkowiak (approved):

PR #4154, opened 2023-07-10
OBGM-408 Unable to cancel requests

alannadolny:
Explanation: https://pihemr.atlassian.net/browse/OBGM-408?atlOrigin=eyJpIjoiYTUxMjQxZGZjYzcxNDQ5NGFiOTZlZDcxNzhmNDdmMWEiLCJwIjoiaiJ9

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -193,7 +193,9 @@ class StockMovementApiController {
             if (stockMovement.isDeleteOrRollbackAuthorized(currentLocation)) {
                 if (stockMovement?.isPending() || !stockMovement?.shipment?.currentStatus) {
                     try {
-                        stockMovementService.deleteStockMovement(params.id)
+                        stockMovementService.deleteStockMovement(stockMovement)
+                        stockMovement?.shipment = null
+                        stockMovement?.requisition = null
                     } catch (Exception e) {
                         def message = "${g.message(code: 'stockMovement.delete.error.message', default: 'The Stock Movement could not be deleted')}"
                         response.status = 400
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -194,8 +194,6 @@ class StockMovementApiController {
                 if (stockMovement?.isPending() || !stockMovement?.shipment?.currentStatus) {
                     try {
                         stockMovementService.deleteStockMovement(stockMovement)
-                        stockMovement?.shipment = null
-                        stockMovement?.requisition = null
                     } catch (Exception e) {
                         def message = "${g.message(code: 'stockMovement.delete.error.message', default: 'The Stock Movement could not be deleted')}"
                         response.status = 400
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -342,7 +342,8 @@ class StockMovementService {
         deleteStockMovement(stockMovement)
     }
 
-    void deleteStockMovement(StockMovement stockMovement) {
+    // stockMovement could be an instance of StockMovement or OutboundStockMovement
+    void deleteStockMovement(def stockMovement) {
         if (stockMovement?.requisition) {
             def shipments = stockMovement?.requisition?.shipments
             shipments.toArray().each { Shipment shipment ->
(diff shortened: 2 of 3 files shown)


grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -504,6 +506,13 @@ class RequisitionService {
         return requisition
     }
 
+    void deleteRequisition(StockMovement stockMovement, Requisition requisition) {
+        OutboundStockMovement associatedStockMovement = OutboundStockMovement.get(stockMovement.id)
```
kchelstowski, line 510: since you fetch it once in the `StockMovementService`, can't you just pass it instead of re-fetching it here?

drodzewicz (approved):
After reading the article you have linked in the issue comments I now understand the problem  and well... it is a headache :smiling_face_with_tear: 
It's weird to me that that's how Grails works, good job on the investigation.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -352,7 +352,10 @@ class StockMovementService {
                 }
                 shipmentService.deleteShipment(shipment)
             }
-            requisitionService.deleteRequisition(stockMovement?.requisition)
+            OutboundStockMovement associatedStockMovement = OutboundStockMovement.get(stockMovement?.id)
```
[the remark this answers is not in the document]
alannadolny: We have two domains: StockMovement and OutboundStockMovement. Both of them are looking the same, have the same id, etc. But two objects (one is created based on StockMovement, and the second is based on OutboundStockMovement) can exist independently. At the begging, I tried to delete only the shipment and requisition from the association, but I still got the error that both of them are associated.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -352,7 +352,10 @@ class StockMovementService {
...
+            stockMovement?.shipment = null
+            associatedStockMovement.shipment = null
+            requisitionService.deleteRequisition(stockMovement, associatedStockMovement, stockMovement.requisition)
```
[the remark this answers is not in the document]
alannadolny: If I have to set an association from outboundStockMovement and from stockMovement to null, I have to pass it

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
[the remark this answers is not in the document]
awalkowiak: @alannadolny as I showed on the call, the issue is that we first pull OutboundStockMovement in the controller and then pass `params.id` to `deleteStockMovement` and there is again refetch of StockMovement. When you change the type of parameter to `def` to allow passing OutboundStockMovement or StockMovement to that method and pass the first object to it and do all the required cleanup, then it works properly.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -193,7 +193,9 @@ class StockMovementApiController {
             if (stockMovement.isDeleteOrRollbackAuthorized(currentLocation)) {
                 if (stockMovement?.isPending() || !stockMovement?.shipment?.currentStatus) {
                     try {
-                        stockMovementService.deleteStockMovement(params.id)
+                        stockMovementService.deleteStockMovement(stockMovement)
+                        stockMovement?.shipment = null
+                        stockMovement?.requisition = null
```
kchelstowski, line 198: let's do it in the service layer, before any `delete()` (GORM's delete) happens (in case anyone forced to add flush there, because it'd break then)

kchelstowski: what I mean is to add:
```groovy
 stockMovement?.shipment = null
```
 to line 354 (above `shipmentService.deleteShipment`)
and 
```groovy
stockMovement?.requisition = null
```
to line 356 (above `requisitionService.deleteRequisition(stockMovement?.requisition)`)

kchelstowski: or actually it sucks, because you then pass `stockMovement.requisition` which would be `null` to the `deleteRequisition`, so you should consider creating "a copy" of that object as:
```groovy
Requisition requisition = stockMovement?.requisition
```
use it like that:
```groovy
requisitionService.deleteRequisition(requisition)
```
but above that do:
```groovy
stockMovement?.requisition = null
```

drodzewicz (approved):

kchelstowski (approved):

awalkowiak (approved):

PR #4155, opened 2023-07-10
OBGM-565 Fix accessing nested map values

drodzewicz:


alannadolny (approved):

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/api/ReplenishmentApiController.groovy
```diff
@@ -57,7 +57,7 @@ class ReplenishmentApiController {
 
         bindReplenishmentData(replenishment, currentUser, currentLocation, jsonObject)
         Order order = replenishmentService.createOrUpdateOrderFromReplenishment(replenishment)
-        if (order.hasErrors() || !order.save(flush: true)) {
+        if (order.hasErrors()) {
```
jmiranda, line 60: We should not be doing that check here in the controller. I think I saw that this was moved to the service in another PR and I just wanted to mention that I think is the proper approach.

grails-app/controllers/org/pih/warehouse/api/ReplenishmentApiController.groovy
```diff
@@ -99,7 +99,7 @@ class ReplenishmentApiController {
     }
 
     Replenishment bindReplenishmentData(Replenishment replenishment, User currentUser, Location currentLocation, JSONObject jsonObject) {
-        bindData(replenishment, jsonObject)
+        bindData(replenishment, jsonObject, [exclude: ['replenishmentItems']])
```
jmiranda, line 102: A comment would be helpful here so we know why the exclude is being used.

jmiranda (commented):
Nice work, @drodzewicz

PR #4156, opened 2023-07-11
OBGM-468 Fix no transaction for product type update and delete methods

kchelstowski:
Katarzyna has asked me to fix also update and delete methods in this ticket, so those are the changes to include them.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy
@@ -10,13 +10,11 @@
 package org.pih.warehouse.product
 
 import grails.util.Holders
-import org.pih.warehouse.api.GenericApiService
 import org.springframework.dao.DataIntegrityViolationException
 
 class ProductTypeController {
 
     ProductTypeService productTypeService
-    GenericApiService genericApiService
 
     static allowedMethods = [save: "POST", update: "POST", delete: "POST"]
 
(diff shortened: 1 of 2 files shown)


alannadolny (approved):

drodzewicz (changes requested):
Looking at the changes you have made reminded me of what Justin mentioned not too long ago about not wanting to create separate domain-based services.
Although it is not the exact case here, I think we can avoid implementing simple `get()` `delete()` etc... actions in services and instead use a generic approach like `GenericApiService`.
Both **delete** and **get** are already implemented there

drodzewicz:
Alternatively, there is also this approach of creating an abstract class with generic actions and extending some other services with this abstract class like it is shown in the examples of [this little article](https://ultimate.systems/web/blog/generic-controllers-and-services-in-spring-boot-java). 
Both approaches are in my opinion good and it is just a matter of preference, but either way, it definitely reduces the amount of redundant code that we have to implement and keeps the codebase a little bit more DRY.

kchelstowski:
@drodzewicz as for `get` I can agree, but I don't like the delete method of `genericApiService`. It does an unnecessary additional fetch to get the object and then to delete it, so I'd have N+1 queries there.

drodzewicz (approved):

awalkowiak (approved):

jmiranda (approved):
LGTM. 

By the way, this is one of those scaffolded CRUD controllers that I'd prefer to be @Transactional until we figure out where these methods should go i.e. probably ProductService. But since this one snuck in, we can grandfather it in and refactor later if we choose.

PR #4157, opened 2023-07-11
OBGM-575 Unable to create product source 

alannadolny:


kchelstowski (approved):

drodzewicz (commented):
I believe you will get some pushback from Justin [on creating new services](https://github.com/openboxes/openboxes/pull/4144#pullrequestreview-1517324473) just for this save action, in my opinion, a more generic approach would be better like using `GenericApiService`

I started a [conversation thread](https://github.com/openboxes/openboxes/pull/4156#pullrequestreview-1523856770) about this topic under Kacpers PR, so let's resolve it there and then we can proceed with the changes

awalkowiak (approved):

awalkowiak:
As @alannadolny suggests there is more logic that could be moved to this service from that controller (for example entire get list logic) I am merging this as is.

jmiranda (commented):
ugh

PR #4158, opened 2023-07-11
OBGM-465 Fix product supplier attribute import

kchelstowski:


alannadolny (approved):

drodzewicz (approved):

jmiranda (commented):
This seems like an ok blanket approach, but just wanted to note that validate probably should be read-only.

PR #4159, opened 2023-07-11
OBGM-517 Exclude putawayItems from implicit data binding

drodzewicz:


kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
```diff
@@ -92,7 +92,7 @@ class PutawayApiController {
 
     private Putaway bindPutawayData(Putaway putaway, User currentUser, Location currentLocation, JSONObject jsonObject) {
         // Bind the putaway
-        bindData(putaway, jsonObject)
+        bindData(putaway, jsonObject, [exclude: ['putawayItems']])
```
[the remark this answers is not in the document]
drodzewicz: this is not the first time I came across this issue.
Before in grails 1 it didn't bind deeply nested items, which is why later in the code we manually iterate through `jsonObject.putawayItems` and bind them ourselves.
https://github.com/openboxes/openboxes/blob/e523fa3a9964532f1340b41080608c7f3402ae79/grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy#L111-L123

Grails 3 apparently does handle properly binding nested objects, hence why there appears a problem of duplicated items: first set is from implicit binding and the other is from manual binding.
So we have to choose one or the other.

I chose to exclude the implicit binding and stick to the manual to avoid unnecessary regression.

grails-app/services/org/pih/warehouse/putaway/PutawayService.groovy
```diff
@@ -220,7 +220,7 @@ class PutawayService {
             }
         }
 
-        order.save(failOnError: true)
+        order.save(failOnError: true, flush: true)
         return order
```
kchelstowski, line 224: can you try just to return the line that saves? so to do:
```groovy
return order.save(failOnError: true)
```
calling save and then below returning the `order` (the not persisted one) feels weird to me
Try to do it without flush and let me know if it is working, as this flush should be unnecessary here

drodzewicz: Unfortunately, it does not solve the problem.
the newly added split item is still not saved at that point and so when we return the putaway after updating, it does not contain the newly added split item.

kchelstowski: What should help is removing @Transactional from the controller. Opening the transaction at controller level makes the transaction being commited not at the service level (so your changes made in service are not passed to the controller), but it is commited after you call `render` (so you pass not yet flushed data to the frontend). 
What should be also visible is that you should see your changes "with a delay" - what I mean by that, is that when you try to do adding the `splitItems` twice, at the second attempt, you would suddenly see the changes from the first attempt.

grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
```diff
@@ -107,23 +106,7 @@ class PutawayApiController {
...
-            // Bind the split items
-            putawayItemMap.splitItems.each { splitItemMap ->
-                PutawayItem splitItem = new PutawayItem()
-                bindData(splitItem, splitItemMap)
-                putawayItem.splitItems.add(splitItem)
-            }
-
-            putaway.putawayItems.add(putawayItem)
-        }
-
```
kchelstowski, line 124: if you had made sure the binder handles that 2nd level nested (splitItemMap) cases, LGTM

kchelstowski (approved):

alannadolny (approved):

awalkowiak (approved):

PR #4160, opened 2023-07-11
OBGM-584 Fix issue with missing session while executing jobs

awalkowiak:


grails-app/jobs/org/pih/warehouse/jobs/UpdateExchangeRatesJob.groovy
```diff
@@ -6,6 +6,11 @@ class UpdateExchangeRatesJob {
 
     static concurrent = false
 
+    // By default this is true on QuartzDisplayJob, which invokes execute()
+    // and if sessionRequired is true, then QuartzDisplayJob tries to do session flush
+    // even if there is no session
+    static sessionRequired = false
+
```
kchelstowski, line 13: I haven't worked with jobs, hence I don't the flow, but I am wondering if when disabling this setting, we shouldn't still make sure that services that we are calling from `execute()` methods shouldn't be marked as @Transactional? 
If so: `currencyService`, `gparsService` are not @Transactional for example.

kchelstowski (commented):
Just a few questions out of curiosity, to get to know the flow. But overall good job @awalkowiak, it was not an easy task!

grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
```diff
@@ -6,6 +6,11 @@ class AssignIdentifierJob {
 
     static concurrent = false
 
+    // By default this is true on QuartzDisplayJob, which invokes execute()
+    // and if sessionRequired is true, then QuartzDisplayJob tries to do session flush
+    // even if there is no session
+    static sessionRequired = false
```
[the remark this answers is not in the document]
awalkowiak: As I wrote in the ticket, the data is persisted without the session. And for the majority of jobs, we don't have to have the session. But in case we might require it then we can do:
```groovy
void execute() {
  Domain.withNewSession {
    ...
  }
}
```
Currently, I am trying to do a `SessionAwareJob`, so we don't have to always remember to do this wrapping if we'd need it.

grails-app/jobs/org/pih/warehouse/jobs/UpdateExchangeRatesJob.groovy
awalkowiak: Yeah, Transactional is still required (due to `allow_update_outside_transaction: false`). I'll add it in places that are missing it.

awalkowiak:
I added the missing `Transactionals` and made minor cleanup. I also created a new ticket for creating a `SessionAwareJob` with proof of concept of it.

kchelstowski (approved):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -379,7 +380,7 @@ class ReportService implements ApplicationContextAware {
             SELECT 0, transaction_type.transaction_code, substring_index(transaction_type.name, '|', 1), transaction_type.id
             FROM transaction_type
         """
-        executeStatements([insertStatement])
+        dataService.executeStatements([insertStatement])
```
jmiranda, line 383: What the ...?

grails-app/jobs/org/pih/warehouse/jobs/SessionlessJob.groovy
```diff
@@ -0,0 +1,8 @@
+package org.pih.warehouse.jobs
+
+class SessionlessJob {
+    // By default this is true on QuartzDisplayJob, which invokes execute()
+    // and if sessionRequired is true, then QuartzDisplayJob tries to do session flush
+    // even if there is no session
+    static sessionRequired = false
```
jmiranda, line 7: I don't think this needs to be static.

FYI, here are some examples where the property is set in the class.
https://github.com/search?q=sessionRequired+language%3AGroovy+path%3Agrails-app%2Fjobs&type=code

jmiranda (commented):
This looks fine to me. However, I will say that I would prefer the sessionRequired property to be explicitly added to each job (similar to concurrent property) because it's just a property and there's no functionality associated with it. This would also provide a less opaque solution and would let us control the sessionRequired property for each job in the rare case we needed to deal allow sessions for some of the jobs. 

The parent class would have been if we needed to do something like this. But just adding a property seems like overkill. 
```
void execute() {
    doSomethingBefore()
    execute()
    doSomethingAfter()
}
```

With that said, it seems like a fine solution.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
jmiranda: How did this happen? Maybe a bad rebase?

grails-app/jobs/org/pih/warehouse/jobs/SessionlessJob.groovy
awalkowiak: No, you're right it does not need to be static

grails-app/services/org/pih/warehouse/report/ReportService.groovy
awalkowiak: I am not sure if it's rebase fault, or what, since it's here from the first migration commit.

awalkowiak:
@jmiranda I moved the `sessionRequired` property back to each job, plus if there was a transaction with gorm save in the execute I wrapped its content with `withNewSession`.

PR #4161, opened 2023-07-12
OBGM-512 Unable to rollback shipped PO shipment 

alannadolny:


drodzewicz (approved):

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -226,7 +226,7 @@ class StockMovementApiController {
 
     def deleteStatus() {
         stockMovementService.rollbackStockMovement(params.id)
-        forward(action: "read")
+        redirect(action: "read", params: params)
```
jmiranda, line 229: Wondering if maybe we return a 200 here and require the client to do the read.

jmiranda (commented):
See comment. We can discuss during a tech huddle and make create tickets for any subsequent refactoring work.

PR #4162, opened 2023-07-12
OBGM-528 Unable to print delivery note

kchelstowski:
The issue was that the `<g:javascript>` implementation changed and it seems not to be invoking jquery anymore.

Grails 1:
Screenshot from 2023-07-12 12-15-19

Grails 3:
Screenshot from 2023-07-12 12-15-56

The jquery in those places was not working at all, an error like this could be visible on such pages:
Screenshot from 2023-07-12 12-17-11

It is because those `print` pages don't use the layout of `custom.gsp` like: `<g:applyLayout name="custom">`
and because `custom.gsp` invokes jquery like this:
```html
<script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
```
the jquery is still working on most of pages (because they apply the custom layout, which invokes jquery).

To fix the issue I just invoke the jquery manually here, but not to repeat the same code, I've created a custom tag lib `<g:jquery>` for that, so that if we face this issue in more places, we can reuse it.


grails-app/taglib/org/pih/warehouse/ScriptTagLib.groovy
```diff
@@ -0,0 +1,13 @@
+package org.pih.warehouse
+
+class ScriptTagLib {
+
+    Closure jquery = { Map attrs, Closure body ->
+        StringBuilder script = new StringBuilder()
+        script.append("<script src=\"https://code.jquery.com/jquery-3.3.1.slim.min.js\"")
+        script.append(" integrity=\"sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo\"")
+        script.append(" crossorigin=\"anonymous\">")
+        script.append("</script>")
```
kchelstowski, line 10: I decided to use `StringBuilder` for a better visibility, instead of either using `+` or to paste whole script in one line

grails-app/views/deliveryNote/print.gsp
```diff
@@ -9,7 +9,7 @@
     <g:set var="entityName" value="${warehouse.message(code: 'requisition.label', default: 'Requisition')}"/>
     <title><warehouse:message code="default.show.label" args="[entityName]"/></title>
     <link rel="stylesheet" href="${resource(dir:'css',file:'buttons.css')}" type="text/css" media="all" />
-    <g:javascript library="jquery" plugin="jquery" />
+    <g:jquery />
```
jmiranda, line 12: We can just explicitly include the jquery library we want to use into the print.gsp layout (grails-app/views/layouts/print.gsp), assuming the use of jquery-3.3.1 doesn't break anything.

So replace this 
```
<g:javascript library="jquery" plugin="jquery" />
```
with this
```
    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
```

jmiranda (changes requested):

grails-app/views/deliveryNote/print.gsp
kchelstowski: @jmiranda could you elaborate on this, because I’m confused? 
I’m exactly doing this, but assigned it to a „variable” (a custom tag), so that if we ever need to bump up the version, we’ll just have to change it in the taglib, not in every file

alannadolny (approved):

PR #4165, opened 2023-07-13
OBGM-575 Unable to create product source (fix for update and delete after QA)

alannadolny:


awalkowiak (approved):

kchelstowski (approved):

drodzewicz (approved):

grails-app/services/org/pih/warehouse/product/ProductSupplierService.groovy
```diff
@@ -6,6 +6,13 @@ import grails.gorm.transactions.Transactional
...
 
     ProductSupplier saveProductSupplier(ProductSupplier productSupplier) {
-        return productSupplier.save()
+        if (!productSupplier.hasErrors()) {
+            return productSupplier.save()
+        }
+        return null
+    }
+
+    void deleteProductSupplier(ProductSupplier productSupplier) {
```
jmiranda, line 15: I know for this case we do the lookup in the controller, but I think our service layer should allow a delete by ID. For now, we can add that method when we need it. I just wanted to make sure that I set expectations.

```
void deleteProductSupplier(String id) {
    ProductSupplier productSupplier = ProductSupplier.get(id)
    productSupplier.delete()
}
```

jmiranda (commented):
LGTM

PR #4170, opened 2023-07-14
OBGM-567 / OBGM-566 Unable to reload autopick after editing qty of pick item to 0 and styling of qty atp validation not visible

alannadolny:
OBGM-566:
The previous handling of our messages in "red boxes" were deleted, so I reverted it, but I did it in a more reusable way - I created a new `apiClient` (we cannot override previously set `interceptor` behavior, adding new one results in showing popup error and the 'red box'). Within `apiClient.js` I override the behavior while getting a response but let us use custom behavior for a response. Additionally, I made the `handleValidationErrors` curried function to pass the `setState` for each component that needs the content of the message to be set.
OBGM-567:
The first thing was that the interceptor behavior wasn't overridden, so we didn't get the error messages in the 'red boxes'. The second thing was a change in Exception, we didn't have errors and fullMessage in the same place in the exception object as we had before. This change should fix other problems with messages caused by the appropriate handling ValidationException.

I also unwrapped ValidationException from Exception, because the content of the fullMessage and errors were in a different properties because of that. (errors and fullMessage weren't in the same place as without wrapping it).

And I found that fix made before:
`// Throw checked exception with ValidationException cause to not rollback deleted picklist while throwing unchecked exception`
won't work because of the change in handling exceptions in transactions. **Previously checked exception not caused a rollback, but unchecked caused. Withing the GORM 6+ version this behavior was changed and all of the exceptions causing a rollback for now.**
It can be found in the documentation (http://docs.grails.org/latest/guide/services.html#declarativeTransactions)
> Version Grails 3.2.0 was the first version to use GORM 6 by default. Checked exceptions did not roll back transactions before GORM 6. Only a method that threw a runtime exception (i.e. one that extends RuntimeException) rollbacked a transaction. 

Referring to that, @kchelstowski was right in saying that rollbacks didn't work in Grails 1 - but it was caused only because of the handling of different types of exceptions (checked / unchecked).
He recorded a video showing the rollback didn't work (grails 1, as you can see the item is saved in the DB):
https://www.loom.com/share/b4c53546dada42d19b74b39d282cabf4?sid=879e5441-575b-4a9c-9a7a-5a9b9c4f53ea
And there is a video showing that it is not saved in the DB (grails 3):
https://www.loom.com/share/09884a401f2349d4b174426175a2a25e?sid=16caaa84-59e5-4991-8dfe-977d1a82004d

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
@@ -97,12 +97,12 @@ class ErrorsController {
     def handleValidationErrors() {
         if (RequestUtil.isAjax(request)) {
             response.status = 400
-            BeanPropertyBindingResult errors = request?.exception?.cause?.errors
+            BeanPropertyBindingResult errors = request?.exception?.cause?.target?.errors ?: request?.exception?.cause?.errors
             List<String> errorMessages = errors.allErrors.collect {
                 return g.message(error: it, locale: localizationService.currentLocale)
             }
             render([errorCode: 400,
-                    errorMessage: "Validation error. " + request?.exception?.cause?.fullMessage,
+                    errorMessage: "Validation error. " + request?.exception?.cause?.target?.fullMessage ?: request?.exception?.cause?.fullMessage,
                     errorMessages: errorMessages
             ] as JSON)
             return
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1237,7 +1237,7 @@ class StockMovementService {
                 // FIXME: Refactor this part to get rid of need to throw wrapped exception plus consider not allowing zeroing out picked stock
                 // Throw checked exception with ValidationException cause to not rollback deleted picklist while throwing unchecked exception
                 // See: https://pihemr.atlassian.net/browse/OBPIH-5318
-                throw new Exception(errorMessage, new ValidationException(errorMessage, requisitionItem.errors))
+                throw new ValidationException(errorMessage, requisitionItem.errors)
             }
         }
     }
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1234,9 +1234,7 @@ class StockMovementService {
                         requisitionItem?.product?.productCode,
                 ].toArray(), errorMessage)
 
-                // FIXME: Refactor this part to get rid of need to throw wrapped exception plus consider not allowing zeroing out picked stock
-                // Throw checked exception with ValidationException cause to not rollback deleted picklist while throwing unchecked exception
-                // See: https://pihemr.atlassian.net/browse/OBPIH-5318
+                // FIXME: consider not allowing zeroing out picked stock
                 throw new ValidationException(errorMessage, requisitionItem.errors)
             }
         }
--- a/src/js/components/stock-movement-wizard/outbound/PackingPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/PackingPage.jsx
@@ -18,8 +18,11 @@ import SelectField from 'components/form-elements/SelectField';
 import TextField from 'components/form-elements/TextField';
 import PackingSplitLineModal from 'components/stock-movement-wizard/modals/PackingSplitLineModal';
 import AlertMessage from 'utils/AlertMessage';
-import apiClient, {
+import {
+  apiClientCustomResponseHandler as apiClient,
   flattenRequest,
+  handleSuccess,
+  handleValidationErrors,
   stringUrlInterceptor,
 } from 'utils/apiClient';
 import { renderFormField } from 'utils/form-utils';
@@ -211,11 +214,14 @@ class PackingPage extends Component {
     this.saveSplitLines = this.saveSplitLines.bind(this);
     this.isRowLoaded = this.isRowLoaded.bind(this);
     this.loadMoreRows = this.loadMoreRows.bind(this);
+    this.setState = this.setState.bind(this);
 
     this.debouncedPeopleFetch =
       debouncePeopleFetch(this.props.debounceTime, this.props.minSearchLength);
 
     this.props.showSpinner();
+
+    apiClient.interceptors.response.use(handleSuccess, handleValidationErrors(this.setState));
   }
 
   componentDidMount() {
--- a/src/js/components/stock-movement-wizard/outbound/PickPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/PickPage.jsx
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 
-import axios from 'axios';
 import arrayMutators from 'final-form-arrays';
 import update from 'immutability-helper';
 import fileDownload from 'js-file-download';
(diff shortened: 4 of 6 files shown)


kchelstowski (approved):

jmiranda:
Marked as do not merge because I want to review this carefully.

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -97,12 +97,12 @@ class ErrorsController {
     def handleValidationErrors() {
         if (RequestUtil.isAjax(request)) {
             response.status = 400
-            BeanPropertyBindingResult errors = request?.exception?.cause?.errors
+            BeanPropertyBindingResult errors = request?.exception?.cause?.target?.errors ?: request?.exception?.cause?.errors
```
jmiranda, line 100: I'm worried about this change. I would prefer to go backwards and figure out if we can rely on request.exception instead of digging deeper into these exceptions. I think a better approach might be to figure out what kind of exception we have and deciding what to do based on that. 

In other words, if we can be sure that handleValidationErrors() will only be invoked when a ValidationException has been thrown then we should be able to do this
```
request.exception.errors 
```
I don't remember why we decided to jump into the cause. 

If it is necessary for certain cases  then it should probably be done in way that cascades. But I really don't like this approach. 
```
request?.exception?.cause?.target?.errors ?: 
    request?.exception?.cause?.errors ?: 
        request?.exception?.errors
```

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -97,12 +97,12 @@ class ErrorsController {
...
             List<String> errorMessages = errors.allErrors.collect {
                 return g.message(error: it, locale: localizationService.currentLocale)
             }
             render([errorCode: 400,
-                    errorMessage: "Validation error. " + request?.exception?.cause?.fullMessage,
+                    errorMessage: "Validation error. " + request?.exception?.cause?.target?.fullMessage ?: request?.exception?.cause?.fullMessage,
```
jmiranda, line 105: Same with this one.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1234,10 +1234,8 @@ class StockMovementService {
                         requisitionItem?.product?.productCode,
                 ].toArray(), errorMessage)
 
-                // FIXME: Refactor this part to get rid of need to throw wrapped exception plus consider not allowing zeroing out picked stock
-                // Throw checked exception with ValidationException cause to not rollback deleted picklist while throwing unchecked exception
-                // See: https://pihemr.atlassian.net/browse/OBPIH-5318
-                throw new Exception(errorMessage, new ValidationException(errorMessage, requisitionItem.errors))
+                // FIXME: consider not allowing zeroing out picked stock
+                throw new ValidationException(errorMessage, requisitionItem.errors)
```
jmiranda, line 1238: I'm all for this change. I just don't understand the rollback issue well enough yet.

jmiranda: Leaving this for @awalkowiak to review, but I think this might require more scope or a spike to investigate and resolve properly.

jmiranda (commented):
This one needs a bit more work, but I need @awalkowiak to comment on the rollback issue.

jmiranda:
@alannadolny I must have missed that in Hibernate settings report, but (I think) you're absolutely correct here. 

> Referring to that, @kchelstowski was right in saying that rollbacks didn't work in Grails 1 - but it was caused only because of the handling of different types of exceptions (checked / unchecked).

The reason why those comments were saved to the database in Grails 1 was that the code was throwing a checked exception (Exception) instead of an unchecked exception (RuntimeException). 

@kchelstowski Let me know if you have a chance to test that out and record another video showing the behavior. This is not urgent, but it would be good to close the loop on that.

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
alannadolny: As I can see `request?.exception` doesn't have either errors or fullMessage property. So, I think that was the reason for jumping into the cause. I tried `request.exception.errors` on validation errors within locations and stock movements and the same situation is here.
image

alannadolny: It looks like we are using `cause` since introducing this kind of handling validation error.
image

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
awalkowiak: Ok, so here is the deal. Looks like this is working, but **might be** broken right now. If the user will have a picklist item that won't be available anymore, then if he tries to reload the picklist he will be getting the error about it, but the picklist won't be deleted. I am ok with merging this one, but this: https://pihemr.atlassian.net/browse/OBPIH-5318 needs to be thoroughly retested (plus similar scenarios regarding the validation). 
If the rollback on the deleted picklist while throwing this exception will cause troubles on validation, then we might have to find another way of returning these messages to the API client.

awalkowiak (commented):
Other than one minor request, I am ok with this one. I feel like we have to live with this one for now (if validation on deprecated picklists works properly), and revisit entire error/exception handling in MVP2 or further (we could split `400` and `409` into separate categories, plus figure out proper message handling)

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
jmiranda: So in Grails 1 if an exception goes uncaught the GrailsExceptionResolver will be invoked to figure out what to do with it. I believe it uses the UrlMappings with error code "500" to determine where to forward/redirect the request. For ValidationExceptions, we send the request to the handleValidationErrors action of the ErrorsController. Pretty basic. 
```
"500"(controller: "errors", action: "handleException")
"500"(controller: "errors", action: "handleNotFound", exception: ObjectNotFoundException)
"500"(controller: "errors", action: "handleValidationErrors", exception: ValidationException)
"500"(controller: "errors", action: "handleUnauthorized", exception: AuthenticationException)
```
The GrailsExceptionResolver will wrap a runtime exception with a GrailsWrappedRuntimeException which seems to happen in the setStatus method below. 

https://github.com/grails/grails-core/blob/a8336f1f68355e42ada68e3d25f1c5ed23b62b94/grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java#L87

The resolveException method then forwards to the ErrorsController.handleValidationErrors() method as described above.

https://github.com/grails/grails-core/blob/a8336f1f68355e42ada68e3d25f1c5ed23b62b94/grails-web-mvc/src/main/groovy/org/grails/web/errors/GrailsExceptionResolver.java#L93C18-L93C38

When the error is handled by the ErrorsController it is now a GrailsWrappedRuntimeException, which is why we need to get the "cause". 
```
request.exception.cause
```
I assume this is just calling ExceptionUtils.getRootCause() but haven't had a chance to verify that yet. In any case, cause returns the ValidationException. 

So there should be no reason to dig deeper to get `target` unless we're doing something wrong somewhere. So I guess my confusion is around the where target comes from.

jmiranda: My assumption is that `target` is related to the InvocationTargetException which I still don't understand completely. 
```
2023-07-21 01:22:09,780 ERROR [080-exec-3] StackTrace                              : Full Stack Trace:
java.lang.reflect.InvocationTargetException: null
```
I see it in the logs from time to time, but I feel like this is only an issue in development environments. With that said, I don't know for sure.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: Ok. Just take a look at my comments above before merging.

jmiranda (commented):
Looks good, just needs a little more TLC around the `target` issue.

awalkowiak:
@jmiranda is it still in the `do not marge` state?

PR #4173, opened 2023-07-17
OBGM-531 Error when creating location via locations configurations wizard

alannadolny:


src/js/components/locations-configuration/LocationDetails.jsx
```diff
@@ -356,17 +356,8 @@ class LocationDetails extends Component {
...
-        locationNumber: values.locationNumber,
-        'organization.id': values.organization.id,
-        'locationGroup.id': _.get(values.locationGroup, 'id') || '',
-        'manager.id': _.get(values.manager, 'id') || '',
-        'locationType.id': _.get(values.locationType, 'id') || '',
+        ...values,
         supportedActivities: _.map(values.supportedActivities, val => val.value),
-        logo: values.logo,
-        bgColor: values.bgColor || '',
-        fgColor: values.fgColor || '',
```
kchelstowski, line 369: have you made sure that it doesn't matter if you pass `''` or `undefined` for those?

alannadolny: It doesn't matter. It's just null.
image

alannadolny: I am not sending those properties, so they're not being assigned to the Location object. In the domain we have:
```
    String fgColor = "000000"
    String bgColor = "FFFFFF"
```
So we're not overriding it, because of not sending nulls. And we didn't fail our validation, because we are using those default values.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
drodzewicz, line 1: I don't really like that we are returning `null` on save when there are errors on the organization object.
In my opinion, we should throw a `ValidationException` if there are errors.
From what I see we are already throwing a Validation error but instead in the controller

drodzewicz (changes requested):

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -67,6 +67,13 @@ class OrganizationService {
         return organization
     }
 
+    Organization saveOrganization(Organization organization) {
+        if (organization.validate() && !organization.hasErrors()) {
+            return organization.save()
+        }
+        throw new ValidationException("Invalid organization", organization.errors)
+    }
+
```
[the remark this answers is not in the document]
jmiranda: Here's an example from Peter Ledbrook, a prominent Grails developer. This might be from Grails 2, but seems to be a decent way of handling the save in the service.

https://github.com/GrailsInAction/graina2/blob/fd1e5ac24fc2708c69b44392e8674796fdb22da4/ch11/hubbub/grails-app/services/com/grailsinaction/PostService.groovy#L18

Here's an example from our codebase if we don't want to return anything from the save.

https://github.com/openboxes/openboxes/blob/1e77b282da20c34b75ebec47b896e3528dae095e/grails-app/services/org/pih/warehouse/data/LocationDataService.groovy#L121

And another one

https://github.com/streamr-dev/core-api/blob/9db3ae26621f504daa060c88d6a50571d8d3310d/grails-app/services/com/streamr/core/service/EthereumUserService.groovy#L28

And spring security 

https://github.com/grails/grails-spring-security-ui/blob/e71f47abe9648314bc304c15d76b487980d6209f/plugin/grails-app/services/grails/plugin/springsecurity/ui/SpringSecurityUiService.groovy#L352

And one that basically acts as a thin wrapper around GORM in order to add transactions.

https://github.com/grails/grails-spring-security-ui/blob/e71f47abe9648314bc304c15d76b487980d6209f/plugin/grails-app/services/grails/plugin/springsecurity/ui/SpringSecurityUiService.groovy#L352

So I think maybe something like this

```
Domain saveDomain(Domain object) { 
    if (object.validate() && object.save()) {
        return object
    } else {
        throw new ValidationException("Invalid object", object.errors)
    }
}
```
or just something more straightforward if we want the client to perform error handling
```
Domain saveDomain(Domain object) { 
    return object.save()
}
```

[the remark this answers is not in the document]
jmiranda: Note we'll probably move this service to @Service so perhaps this doesn't matter. But I just wanted to go down this path because I want a decent solution in case we choose not to go down the @Service route.

grails-app/controllers/org/pih/warehouse/api/OrganizationApiController.groovy
```diff
@@ -41,9 +41,7 @@ class OrganizationApiController extends BaseDomainApiController {
         Organization organization = new Organization()
         bindOrganizationData(organization, jsonObject)
 
-        if (organization.hasErrors() || !organization.save(flush: true)) {
-            throw new ValidationException("Invalid organization", organization.errors)
-        }
+        organizationService.saveOrganization(organization)
```
jmiranda, line 44: Let's see if we can remove the first three line 
```
JSONObject jsonObject = request.JSON
Organization organization = new Organization()
bindOrganizationData(organization, jsonObject)
```
and replace the method with this 
```
def create(Organization organization) { 
    organizationService.saveOrganization(organization)
    render([data: [id: organization.id]] as JSON)
}
```
If that works in the happy case, does it actually need to be 
```
def create(Organization organization) { 
    organization = organizationService.saveOrganization(organization)
    render([data: [id: organization.id]] as JSON)
}
```
or something completely different?

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -67,6 +67,13 @@ class OrganizationService {
         return organization
     }
 
+    Organization saveOrganization(Organization organization) {
+        if (organization.validate() && !organization.hasErrors()) {
```
kchelstowski, line 71: @alannadolny the order should be reverted, so:
```groovy
!organization.hasErrors() && organization.validate()
```
@jmiranda @alannadolny in this case the `hasErrors()` might have errors even before the `validate()` is executed.
It is because when you are **binding** data, e.g. `new Object(params)` or like it is done here in the controller using the `bindData`, then a validation of **data types** is proceeded - so the `organization.errors` might not be empty at this point if e.g. we provided a `name` that would be an `Integer` - it would already contain that error.
`validate()` on the other hand proceeds with the **constraints** validation, so to check for example if the name is unique, etc.

So I'd say it might be a good pattern to first check `hasErrors()`, not to invoke the `validate()` (which is "deeper" than when binding data, that only checks the data type), if there are already errors from data types.
On the other hand it might be good to let the user know all the errors immediately, so that he/she can correct him/herself once.

I'm fine with both, just wanted to point out, that a different validation is proceeded when binding data and when calling `validate()`.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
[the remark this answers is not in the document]
kchelstowski: thanks for the resources @jmiranda , will take a look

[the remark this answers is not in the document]
drodzewicz: I might be wrong but I feel like even this code
```groovy
Domain saveDomain(Domain object) { 
    if (object.validate() && object.save()) {
        return object
    } else {
        throw new ValidationException("Invalid object", object.errors)
    }
}
```

could be shortended to just
```groovy
Domain saveDomain(Domain object) { 
   object.save(failOnError: true)
   return object
}
```

citing the grails [documentation](https://docs.grails.org/latest/ref/Domain%20Classes/save.html):

> When set to true the save method with throw a grails.validation.ValidationException if [validation](https://docs.grails.org/latest/guide/validation.html) fails.

grails-app/controllers/org/pih/warehouse/api/OrganizationApiController.groovy
alannadolny: `bindOrganizationData` generates `organization.code` and `organization.partyType`
```
        if (!organization.code) {
            organization.code =
                    identifierService.generateOrganizationIdentifier(organization.name)
        }

        if (!organization.partyType) {
            organization.partyType = PartyType.findByCode(Constants.DEFAULT_ORGANIZATION_CODE)
        }
```
so it's necessary to be working properly (without this bind validation doesn't pass).

awalkowiak: @alannadolny I think this part could be moved out of the bind into create action with the Organization param, I think this will look cleaner. Other than that lgtm.

jmiranda: Yeah, this is borderline "business logic". I would either move it into the controller like @awalkowiak said OR add a createOrganization() method in the service to handle this logic. Create would call saveOrganization().

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
jmiranda: My preference would be to return the domain object with the errors from the service. Let the caller decide how they want to handle the validation errors.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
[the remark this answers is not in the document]
kchelstowski: @jmiranda just wanted to point out, that the `save()` that is created by data services, throws `ValidationException` by default, so it doesn't return `null` in case, something was not saved. We can't manipulate it.

> If any of the parameters don’t match up to a property on the domain class then a compilation error will occur.

> If a validation error occurs then a ValidationException will be thrown from the service.

so it feels like Grails recommends this approach if they make it as default and unchangeable for data services.

awalkowiak (approved):

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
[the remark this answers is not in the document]
jmiranda: @kchelstowski Ooh, interesting. Thanks for that piece of information. 

image

PR #4176, opened 2023-07-18
OBGM-518 Change generate pdf for putaway request url

drodzewicz:


src/js/components/put-away/PutAwaySecondPage.jsx
```diff
@@ -504,10 +504,10 @@ class PutAwaySecondPage extends Component {
    */
   generatePutAwayList() {
     this.props.showSpinner();
-    const url = '/putAway/generatePdf';
+    const url = `/putAway/generatePdf/${this.state.putAway.id}`;
```
kchelstowski, line 507: As discussed at the beginning of the migration - whenever we touch the urls, we should add them to our urls factory.
Would you mind adding it there?

drodzewicz: sure, will do

PR #4178, opened 2023-07-18
OBGM-562 Unable to Generate Stock Transfer document

alannadolny:


grails-app/views/layouts/print.gsp
```diff
@@ -19,7 +19,9 @@
 	<link rel="stylesheet" href="${resource(dir:'css',file:'wizard.css')}" type="text/css" media="print, screen, projection" />
 	<link rel="stylesheet" href="${resource(dir:'js/jquery.ui/css/smoothness', file:'jquery-ui.css')}" type="text/css" media="print, screen, projection" />
 	<link rel="stylesheet" href="${resource(dir:'css',file:'custom.css')}" type="text/css" media="print, screen, projection" />
-	<g:javascript library="jquery" plugin="jquery" />
+	<script src="https://code.jquery.com/jquery-3.3.1.slim.min.js"
+			integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous">
+	</script>
```
alannadolny, line 24: @kchelstowski explained here why the `<g:javascript ... />` doesn't work. I just applied the script here to use the jquery plugin in all of the print pages that use it. Kacpers explanation: https://github.com/openboxes/openboxes/pull/4162

grails-app/views/replenishment/print.gsp
```diff
@@ -27,7 +27,7 @@
     <hr/>
 </div>
 
-<g:set var="enablePageBreak" value="${params.pageBreak != 'Disable page break'}"/>
+<g:set var="enablePageBreak" value="${params.pageBreak == 'Enable page break'}"/>
```
alannadolny, line 30: I changed it like that, because Katarzyna said the page break should be disabled by default. For now we can select:
- "" (this one is selected be default)
- "Enable page break"
- "Disable page break"
So previously, when it was `params.pageBreak != 'Disable page break'` the `enablePageBreak` was set to `true` by default.

grails-app/views/layouts/print.gsp
jmiranda: Just make sure that jquery 3.3.1 can be used in all cases here i.e. whenever the print layout is used.
```
<meta name="layout" content="print" />
```
If you look at the [print picklist](https://obdev1.pih-emr.org/openboxes/picklist/print/ff8080818715c634018752961f715d05) feature on OBDEV1, you'll notice that the `<g:javascript library="jquery" plugin="jquery"/>` tag includes jquery 1.7.2 by default.
```
<script type="text/javascript" src="/openboxes/plugins/jquery-1.7.2/js/jquery/jquery-1.7.2.min.js">
```
I would obviously prefer to use JQuery 3.3.1, but just want to make sure we don't break anything.

grails-app/views/replenishment/print.gsp
jmiranda: :+1:

jmiranda (commented):
This looks good to me. I just want to be sure that none of the pages that use this layout break due to the jquery version upgrade.

Search for `content="print"` to see all pages that reference this layout.

kchelstowski (approved):

drodzewicz (approved):

alannadolny:
I checked all of those pages, and as I can see there are no issues with bumping the JQuery to 3.3.1

awalkowiak (approved):

PR #4180, opened 2023-07-19
OBGM-615 Create product supplier gorm data service for CRUD operations

kchelstowski:


kchelstowski:
I've tried to do a generic interface for `save()`, `delete()`, `get()` methods, like:
```groovy
interface CrudDataService<T, ID> {
  T save(T object);
  void delete(ID id);
  T get(ID id);
}
```
and that our interfaces for each domain would just extend this one like:
```groovy
@Service(value = ProductSupplier, name = "productSupplierGormService")
interface ProductSupplierDataService extends CrudDataService<ProductSupplier, String> {
  <some custom methods>
}
```
but unfortunately Grails doesn't support it :cry:
https://stackoverflow.com/questions/56201213/grails-gorm-extend-generic-interface

jmiranda (approved):
LGTM. I'm still bothered by the naming `productSupplierGormService` but we should create a tech debt ticket to figure out the naming for the other ProductSupplierDataService. And as part of that ticket, we should rename this bean productSupplierDataService.

Bravo @kchelstowski. Thanks for figuring out the @Service part.

 (there we go)

grails-app/services/org/pih/warehouse/product/ProductSupplierDataService.groovy
```diff
@@ -0,0 +1,12 @@
+package org.pih.warehouse.product
+
+import grails.gorm.services.Service
+
+@Service(value = ProductSupplier, name = "productSupplierGormService")
```
awalkowiak, line 5: @kchelstowski Could it be done without the name? As a:
```groovy
@Service(ProductSupplier)
interface ProductSupplierDataService {
    //...
}
```
 Is this additional name required here?

awalkowiak: Ok nvm, we discussed it on slack that it won't work. You can disregard this comment.

alannadolny (approved):

PR #4184, opened 2023-07-20
OBGM-536 Pagination not visible on inventory browser 

alannadolny:
Pagination was not visible because `commandInstance?.searchResults instanceof grails.orm.PagedResultList` was false - PagedResultList is an instance of `org.pih.warehouse.PagedResultList`.

After fixing the issue described above I noticed that there is an error after changing the page with filters applied. In the URL we had two similar params: `location.id=...` and `location=[id: ...]`. This second one wasn't visible after changing the page with filters. I saw that the command instance assigned location to its property when there was just `location.id` passed. After passing the `location=[id: ...]` command used this one parameter to get the location property and after that error with an invalid constructor of the Location class was thrown. I removed the `.id` from the location param, so `location=[id: ...]` got overridden and the command assign an appropriate location to its property.

### DIFF
--- a/grails-app/views/inventory/_filters.gsp
+++ b/grails-app/views/inventory/_filters.gsp
@@ -1,7 +1,6 @@
 <div class="filters" >
 	<g:form method="GET" controller="inventory" action="browse">
 		<g:hiddenField name="max" value="${params.max?:10 }"/>
-		<g:hiddenField name="location" value="${session.warehouse.id}"/>
 		<div class="box">
 			<h2><warehouse:message code="inventory.filterByProduct.label"/></h2>
 			<div class="filter-list-item">
--- a/grails-app/views/inventory/_filters.gsp
+++ b/grails-app/views/inventory/_filters.gsp
@@ -1,7 +1,7 @@
 <div class="filters" >
 	<g:form method="GET" controller="inventory" action="browse">
 		<g:hiddenField name="max" value="${params.max?:10 }"/>
-		<g:hiddenField name="location.id" value="${session.warehouse.id}"/>
+		<g:hiddenField name="location" value="${session.warehouse.id}"/>
 		<div class="box">
 			<h2><warehouse:message code="inventory.filterByProduct.label"/></h2>
 			<div class="filter-list-item">
--- a/grails-app/views/inventory/browse.gsp
+++ b/grails-app/views/inventory/browse.gsp
@@ -1,4 +1,5 @@
-<%@ page import="org.pih.warehouse.PagedResultList; org.pih.warehouse.core.Location" %>
+<%@ page import="org.pih.warehouse.PagedResultList" %>
+<%@ page import="org.pih.warehouse.core.Location" %>
 <%@ page import="org.pih.warehouse.product.Product" %>
 <html>
     <head>
--- a/grails-app/views/inventory/browse.gsp
+++ b/grails-app/views/inventory/browse.gsp
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.core.Location" %>
+<%@ page import="org.pih.warehouse.PagedResultList; org.pih.warehouse.core.Location" %>
 <%@ page import="org.pih.warehouse.product.Product" %>
 <html>
     <head>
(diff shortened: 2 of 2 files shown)


kchelstowski (approved):

drodzewicz (approved):

grails-app/views/inventory/browse.gsp
```diff
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.core.Location" %>
+<%@ page import="org.pih.warehouse.PagedResultList; org.pih.warehouse.core.Location" %>
```
awalkowiak, line 1: Please split into separate lines

grails-app/views/inventory/browse.gsp
```diff
@@ -151,7 +151,7 @@
 
 								</form>
 							</div>
-							<g:if test="${commandInstance?.searchResults instanceof grails.orm.PagedResultList}">
+							<g:if test="${commandInstance?.searchResults instanceof PagedResultList}">
```
awalkowiak, line 154: Could you try using `grails.gorm.PagedResultList` (**g**orm)? (replace the `org.pih.warehouse.PagedResultList` with it in the `searchProducts` method)

awalkowiak (changes requested):
see comments above

grails-app/views/inventory/browse.gsp
jmiranda: I would agree with this ... if it was easy. Unfortunately, I remember the PagedResultList class (there seem to be two in Grails 3: grails.orm.PagedResultList and grails.gorm.PagedResultList) does not have a constructor that supports our use case. So if there's not an easy way to do it, I would leave it alone and create a ticket to refactor this by fixing the query itself so we're returning the correct paginated data from the query.

grails-app/views/inventory/browse.gsp
jmiranda: :+1:

grails-app/views/inventory/_filters.gsp
```diff
@@ -1,7 +1,7 @@
 <div class="filters" >
 	<g:form method="GET" controller="inventory" action="browse">
 		<g:hiddenField name="max" value="${params.max?:10 }"/>
-		<g:hiddenField name="location.id" value="${session.warehouse.id}"/>
+		<g:hiddenField name="location" value="${session.warehouse.id}"/>
```
jmiranda, line 4: I don't like this solution because it suggests we don't really know enough to do it properly. I want to make sure we completely understand the problem as well as the ramifications of this type of change. 

In other words, once we start doing it this way we'll need to do it everywhere and that's going to be a lot of work and a lot of re-testing. If the Grails 3 data binding is not working the way it was in Grails 1, then we need to know what has changed and how we should support it.

Because of that, I think we should remove this change and add a new ticket to explore this issue.

grails-app/views/inventory/browse.gsp
alannadolny: Yup, you're right @jmiranda. There is no constructor that supports our use case: 
`Could not find matching constructor for: grails.orm.PagedResultList(java.util.ArrayList, java.lang.Integer)`

awalkowiak (approved):

grails-app/views/inventory/_filters.gsp
```diff
@@ -1,7 +1,6 @@
 <div class="filters" >
 	<g:form method="GET" controller="inventory" action="browse">
 		<g:hiddenField name="max" value="${params.max?:10 }"/>
-		<g:hiddenField name="location.id" value="${session.warehouse.id}"/>
```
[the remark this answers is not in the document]
jmiranda: With that said, we can leave it as-is since the browse action does a look up on the same location ID if one is not passed in the params.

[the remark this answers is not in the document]
awalkowiak: From what I understood from @alannadolny, if it would be as a `location.id` then it would fail to bind and would throw an error.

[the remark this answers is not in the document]
jmiranda: @awalkowiak Only when clicking one of the paginate links, not on the POST request when performing a search. See the discussion in Slack.

It's related to this ticket
https://pihemr.atlassian.net/browse/OBGM-291

PR #4187, opened 2023-07-24
OBGM-599 Unable to import shipment template with exp date

alannadolny:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/shipping/CombinedShipmentService.groovy
```diff
@@ -130,7 +132,9 @@ class CombinedShipmentService {
                     line.errors << "Unable to parse expiry date: ${line.expiry}"
                     valid = false
                 }
-                Date date = grailsApplication.config.openboxes.expirationDate.minValue
+                Date date = Constants.EXPIRATION_DATE_FORMATTER.parse(
+                    Holders.grailsApplication.config.openboxes.expirationDate.minValue
+                )
```
jmiranda, line 137: My only comment here would be to make this more DRY somehow. If we had done that from the start we would have only need to update the implementation and not the 

Something like this would be ok
```
ConfigUtil.getMinimumExpirationDate()
```
Or better yet, allow Grails to try different date formats by telling it what type you want to get
```
Holders.config.getProperty("openboxes.expirationDate.minValue", Date.class)
```
and remove the Constants.EXPIRATION_DATE_FORMATTER and Constants.EXPIRATION_DATE_FORMAT. making sure the format is in application.yml
```
grails:    
    databinding:
        dateFormats: 
            - MM/dd/yyyy
```

jmiranda (commented):
LGTM; see comment

grails-app/services/org/pih/warehouse/shipping/CombinedShipmentService.groovy
jmiranda: You could still put this into a ConfigUtil (or I guess we called it ConfigHelper), but because grails supports multiple date formats, we no longer need to create our own date formatters. 

There's no need to do this now, but we should add a ticket to revisit date formatting and config properties in MVP2 or post-migration.

PR #4188, opened 2023-07-24
OBGM-592:594 Add InvoiceDataService for Invoice domain

awalkowiak:
- move save and delete methods to InvoiceDataService
- use new data service methods in InvoiceController and InvoiceApiController


grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -24,6 +24,7 @@ import org.pih.warehouse.invoice.InvoiceStatus
 class InvoiceApiController {
 
     def identifierService
+    def invoiceDataService
```
alannadolny, line 27:
```suggestion
    InvoiceDataService invoiceDataService
```

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -53,7 +54,7 @@ class InvoiceApiController {
         Invoice invoice = new Invoice()
         bindInvoiceData(invoice, currentLocation, jsonObject)
 
-        if (invoice.hasErrors() || !invoiceService.saveInvoice(invoice)) {
+        if (invoice.hasErrors() || !invoiceDataService.save(invoice)) {
```
alannadolny, line 57: I think this checking errors is not needed here, because this save will throw validationException by itself.
Correct me @kchelstowski if I am wrong.

jmiranda: Yeah, I don't have a strong opinion, but @kchelstowski brings up a good point. If we do want to 

Ideally the action would look like this

```
// Data binding should happen automatically 
def create(Invoice invoice) { 

    // FIXME this needs its own spike but this should probably needs to be more DRY 
    // i.e. move it to a central place either as a utility method (maybe added to the thread 
    // local implementation) or as an AOP annotation (e.g. @LocationRequired) that can 
    // be added to a controller action. We could check to see what Grails does with the 
    // tenant when using multi-tenancy since that should use a similar pattern.
    Location currentLocation = Location.get(session.warehouse.id)
    if (!currentLocation) {
        throw new IllegalArgumentException("User must be logged into a location to create invoice")
    }

    // If the save fails, the data service should throw a ValidationException that'll be handled by 
    // the Grails exception handler. If we need to catch the validation exception and render it in 
    // a non-default way then we can do that here. 
    invoiceDataService.save(invoice)
    
    render render([data: invoice?.toJson()] as JSON)
}
```

jmiranda (approved):
lgtm

PR #4191, opened 2023-07-25
OBGM-595 Fix deleting a user role for specific location

kchelstowski:


alannadolny (approved):

kchelstowski:
I decided to try the `abstract class` approach, that GORM Data Services also use. You can compare two commits and let me know, which version fits better for you. 
If it were up to me, I'd go with the 2nd version (abstract class).

drodzewicz (approved):
I agree with you, the 2nd version is much more appealing to me

awalkowiak (approved):

grails-app/services/org/pih/warehouse/core/ILocationRoleDataService.groovy
```diff
@@ -0,0 +1,9 @@
+package org.pih.warehouse.core
+
+interface ILocationRoleDataService {
```
jmiranda, line 3: Let's leave the interface as LocationRoleDataService and use something different for the implementation class (BaseLocationRoleDataService, AbstractLocationRoleDataService, etc).

Using the I-interface-name convention feels very Microsoft / C# to me. I have nothing against Hungarians but the Hungarian notation should have been burned with fire long ago.

Some light reading 
https://developer.okta.com/blog/2019/06/25/iinterface-considered-harmful
https://stackoverflow.com/questions/2814805/java-interfaces-implementation-naming-convention
https://stackoverflow.com/questions/541912/interface-naming-in-java
https://softwareengineering.stackexchange.com/questions/117348/should-interface-names-begin-with-an-i-prefix

jmiranda: Actually, you don't need the interface at all. Just throw all of the interface methods into the class as abstract methods like this
https://github.com/MTUHIDE/CCLC-Queue/blob/535d339ebce367440f129e69efd460c96836e8da/grails-app/services/edu/mtu/cs/queue/QueueAnswerService.groovy#L5

jmiranda: If you did have an interface and an implementation class, you could also put them in the same file which feels like a better convention to me.

https://github.com/rundeck/rundeck/blob/ee6fedc65d917ebb910ab0787c85e95a95ff8b6a/rundeckapp/grails-app/services/rundeck/services/data/ScheduledExecutionDataService.groovy#L19

jmiranda (changes requested):
Changes to the interface naming convention.

grails-app/services/org/pih/warehouse/core/ILocationRoleDataService.groovy
kchelstowski: @jmiranda to be honest... I also hate the `ILocationDataService` naming, but the reason I went that path since in GORM data services docs they tend to use it.

kchelstowski: @jmiranda I've moved the methods to the abstract class as you suggested

jmiranda (approved):

PR #4192, opened 2023-07-26
OBGM-549 Unpacked packing unit not visible on Packing list on SM view page 

alannadolny:
In Grails 1 `<g:set var="previousContainer"/>` set value of `previousContainer` to empty string:
image
In Grails 3 value of `previousContainer` is null.
Untitled
So, `previousContainer != shipmentItem?.container` was evaluated to false at the beginning, because it was `null != null`. It should be true to display `unpacked` (in Grails 1 it was `'' != null`)


drodzewicz (approved):

kchelstowski (approved):
Nice catch :clap:

awalkowiak (approved):

jmiranda (approved):

PR #4193, opened 2023-07-27
OBGM-539 No longer able to download certificate of donation from send page of outbound and requests

alannadolny:
Validation of merging regions in xls has changed. In the new version number of merging cells has to be higher than 1:
Untitled
Meanwhile, in Grails 1 the only validation that is being checked is:
image
So, I removed the `initialRowIndex` and `finalRowIndex` variables. In the begging, both of them have 0 as a value. Later in each iteration, we assigned `row.getRowNum()` to both of them, so they have the same value every time.
When the number of cells was greater than 16, this part of the code was triggered:
`sheet.addMergedRegion(CellRangeAddress.valueOf("A${initialRowIndex + 1}:A${finalRowIndex + 1}"))`
So, because of the fact, that `initialRowIndex` and `finalRowIndex` are equal, we are doing sth like this:
`sheet.addMergedRegion(CellRangeAddress.valueOf("A17:A17"))`
`sheet.addMergedRegion(CellRangeAddress.valueOf("A18:A18"))`
`sheet.addMergedRegion(CellRangeAddress.valueOf("A19:A19"))`
and so on.
So we were always trying to merge only one cell and the validation didn't pass.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
@@ -1150,7 +1150,7 @@ class DocumentService {
             row.getCell(CELL_INDEX++).setCellStyle(tableHeaderLeftStyle)
 
             def totalPrice = 0
-            def previousContainer = "", initialRowIndex = 0, finalRowIndex = 0
+            def previousContainer = ""
             shipmentInstance.sortShipmentItemsBySortOrder().each { itemInstance ->
 
                 CELL_INDEX = 0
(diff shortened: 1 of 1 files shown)


kchelstowski (approved):

awalkowiak (approved):

jmiranda (approved):
I have no idea what's happening here, so LGTM

jmiranda:
To be clear, I understand that the POI API has changed. I just don't understand why I used addMergedRegion() for most of these fields when there were only really a handful that needed it.

If I could do it all over again, I would have come up with our own XLS config / templating system so we didn't have to do all of this annoying hard-coding all over the place.

PR #4194, opened 2023-07-27
OBGM-610 Fix creating an internal location

kchelstowski:
The `selectZoneLocationByLocation` is defined like this:
```groovy
def selectZoneLocationByLocation = { attrs, body ->
        def location = Location.get(attrs.id)

        if (location) {
            attrs.from = locationService.getZones(location)
        }

        attrs["class"] = "chzn-select-deselect"
        attrs["noSelection"] = ["null": ""]
        attrs.optionKey = 'id'
        attrs.optionValue = 'name'

        out << g.select(attrs)
    }
```
so we were only assigning `from` param if `location` is not falsy.

From the Grails 3 docs we can see that `from` param is required for `g:select`: 
Screenshot from 2023-07-27 15-35-38
and that the implementation of the select has this if statement (2nd from the top):

Screenshot from 2023-07-27 15-36-11

which I couldn't find in the implementation in Grails 1. It's weird, because Grails 1 docs also say, that `from` param is required.
Screenshot from 2023-07-27 15-38-41

### DIFF
--- a/grails-app/views/location/edit.gsp
+++ b/grails-app/views/location/edit.gsp
@@ -138,7 +138,6 @@
                                                 id="${locationInstance?.parentLocation?.id}"
                                                 value="${locationInstance?.zone?.id}"
                                                 noSelection="['null': '']"
-                                                from="[]"
                                                 class="chzn-select-deselect"/>
                                     </td>
                                 </tr>
--- a/grails-app/views/location/edit.gsp
+++ b/grails-app/views/location/edit.gsp
@@ -138,6 +138,7 @@
                                                 id="${locationInstance?.parentLocation?.id}"
                                                 value="${locationInstance?.zone?.id}"
                                                 noSelection="['null': '']"
+                                                from="[]"
                                                 class="chzn-select-deselect"/>
                                     </td>
                                 </tr>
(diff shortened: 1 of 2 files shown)


alannadolny (approved):
> It's weird, because Grails 1 docs also say, that from param is required.

It's weird, maybe they copy-pasted creating the docs? :see_no_evil: :hear_no_evil:

drodzewicz (approved):

grails-app/views/location/edit.gsp
```diff
@@ -138,6 +138,7 @@
                                                 id="${locationInstance?.parentLocation?.id}"
                                                 value="${locationInstance?.zone?.id}"
                                                 noSelection="['null': '']"
+                                                from="[]"
```
awalkowiak, line 141: Add this as a default in the tag instead of here (when no location for a given id).

awalkowiak (changes requested):

grails-app/views/location/edit.gsp
jmiranda: :+1:

jmiranda (commented):
See @awalkowiak suggest.

grails-app/views/location/edit.gsp
kchelstowski: fixed

awalkowiak (approved):

PR #4195, opened 2023-07-27
OBGM-601 Add a product package unique constraint on ProductPackage domain

drodzewicz:
### A quick rundown for those who are not in the loop
Currently, on the ProductPackage table, we have a database constraint that requires to have a unique set of ProductPackages based on _product, productSupplier, uom, quantity_.
This constraint is only set using the luquibase change log
https://github.com/openboxes/openboxes/blob/f63862292ba27bae0316e076f7bf3b40844619ef/grails-app/migrations/install/changelog-create-schema.groovy#L3441-L3443  
The issue that ticket mentions is that when a user triggers this validation constraint it doesn't display the validation error in a user-friendly way but just redirects to the common error page with a whole stack trace, meaning that we are not properly handling this error 
image

### My current solution
So after looking through [grails documentation](https://docs.grails.org/latest/ref/Constraints/unique.html), I found out that we can recreate this constraint on the ProductPackage domain using a **unique** keyword and passing an array of columns that we want to validate against, basically exactly what I was looking for.
That way we don't need to handle any exceptions and catch the error before we try to persist the data.
One thing that kind of concerns me with this solution is that now we will have two exactly same constraints _(one on domain class and the other on DB table)_, so if we proceed with the current solution, do you think I should remove the **product_package_uniq_idx** constraint?

### Alternative solution
If we don't want to go the _"add unique constraint on ProductPackage"_ route, we can just handle this issue manually by catching the exception in the controller 
```groovy
  if (!productInstance.hasErrors()) {
    try {
        productService.saveProduct(productInstance)
        flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'package.label', default: 'Product'), packageInstance.name])}"
    } catch (org.springframework.dao.DataIntegrityViolationException e) {
        flash.message = "Some error message"
    }
    redirect(action: "edit", id: productInstance?.id)
}
```

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -393,7 +393,7 @@ class ProductController {
             }
         }
 
-        if (!productInstance.hasErrors() && productInstance.save(flush: true)) {
+        if (!productInstance.hasErrors() && packageInstance.validate() && productInstance.save(flush: true)) {
             flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'package.label', default: 'Product'), packageInstance.name])}"
             redirect(action: "edit", id: productInstance?.id)
         } else {
--- a/grails-app/domain/org/pih/warehouse/product/ProductPackage.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductPackage.groovy
@@ -54,7 +54,7 @@ class ProductPackage implements Comparable<ProductPackage>, Serializable {
         name(nullable: true)
         description(nullable: true)
         gtin(nullable: true)
-        uom(nullable: true)
+        product(nullable: true, unique: ['uom', 'quantity', 'productSupplier'])
         productPrice(nullable: true)
         quantity(nullable: false)
         createdBy(nullable: true)
--- a/grails-app/domain/org/pih/warehouse/product/ProductPackage.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductPackage.groovy
@@ -54,6 +54,7 @@ class ProductPackage implements Comparable<ProductPackage>, Serializable {
         name(nullable: true)
         description(nullable: true)
         gtin(nullable: true)
+        uom(nullable: true)
         product(nullable: true, unique: ['uom', 'quantity', 'productSupplier'])
         productPrice(nullable: true)
         quantity(nullable: false)
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -2019,6 +2019,7 @@ productCatalogItems.label=Product Catalog Items
 productPackage.label=Product Package
 productPackages.label=Product Packages
 productPackage.quantity.label=Pack Quantity
+productPackage.product.unique=Similar product package already exists on this product
 # Product suppliers
 productSupplier.label=Product Source
 productSupplier.code.label=Source Code
--- a/grails-app/views/product/edit.gsp
+++ b/grails-app/views/product/edit.gsp
@@ -37,6 +37,11 @@
 	                <g:renderErrors bean="${productInstance}" as="list" />
 	            </div>
             </g:hasErrors>
+            <g:hasErrors bean="${packageInstance}">
+                <div class="errors">
+                    <g:renderErrors bean="${packageInstance}" as="list" />
+                </div>
+            </g:hasErrors>
 
    			<g:if test="${productInstance?.id }">
 				<g:render template="summary" model="[productInstance:productInstance]"/>


kchelstowski (approved):
We definitely don't wanna go 2nd route (catching the exceptions manually). 
You can check out my Hibernate settings docs to get to know why it would be bad.

alannadolny (approved):

grails-app/domain/org/pih/warehouse/product/ProductPackage.groovy
```diff
@@ -54,7 +54,7 @@ class ProductPackage implements Comparable<ProductPackage>, Serializable {
         name(nullable: true)
         description(nullable: true)
         gtin(nullable: true)
-        uom(nullable: true)
```
awalkowiak, line 57: Was this expected to be removed? Is uom nullable or not?

drodzewicz: sorry, my bad.
I'll bring it back

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -393,7 +393,7 @@ class ProductController {
             }
         }
 
-        if (!productInstance.hasErrors() && productInstance.save(flush: true)) {
+        if (!productInstance.hasErrors() && packageInstance.validate() && productInstance.save(flush: true)) {
```
jmiranda, line 396: This seems unnecessary or at least overly cautious. We should be able to rely on just the save to do what we want here. Or find a common pattern used by other Grails 3 projects.

jmiranda (commented):
I'm going to respond in the conversation tab.

jmiranda:
> One thing that kind of concerns me with this solution is that now we will have two exactly same constraints (one on domain class and the other on DB table), so if we proceed with the current solution, do you think I should remove the product_package_uniq_idx constraint?

This is a good instinct but I actually have the opposite feeling. At the moment, the fact that there isn't a unique constraint at both the logical (GORM) and physical (database) layer kind of bothers me. 

In other words, let's say the opposite situation was true and we only had a unique constraint in the domain class. I should not get a constraint validation exception executing this code
```
new ProductPackage(product:product, productSupplier: productSupplier, uom: uom, quantity: quantity).save()
```
and then be able to turn around and execute the following SQL without any 
```
sql.execute("insert into product_package values (:product, :productSupplier, :uom, :quantity)
```

Using ORM tools (like Hibernate, JPA, or GORM) to model our domains and map them to the database is just for our convenience and is the first step in the modeling process. Generating the database level objects i.e. tables, columns, constraints, from that design is the next step. And having all structures, validation, logic in both places seems rather important to me.  

In our current case, there's not really a problem because the application still encounters the constraint validation error but you could probably imagine a case where the reverse is true (logic happens in app layer, not db) and we get ourselves in trouble.

jmiranda:
Also take a look at this article. I remember reading it a while back and thinking "hmm, that could be bad". It might not be concerning for individual inserts/updates, but bulk inserting product packages might take a big performance hit. 

https://medium.com/wizpanda/change-one-line-to-improve-the-performance-of-your-grails-app-drastically-861dbfc77291

NOTE: I have not found documentation to support the use of "unique" in the mapping, so that should be confirmed. 

Lastly. I'm slightly concerned about the "unique" constraint loophole that has gotcha'd us a couple of times in the past. That is the one where a null in any of those fields causes the constraint to short-circuit and allow duplicate records to be inserted. 

But the following comment from this Grails issue https://github.com/grails/grails-data-mapping/issues/1363#issuecomment-1237227746 seems to explain the behavior. 

> If you consider null to be unknown instead of a single shared value, I think the current behavior makes sense.

jmiranda:
To get back to the issue at hand, I honestly don't know the right answer. My gut says we shouldn't have to handle these kinds of validation exceptions ourselves. But if you go back and look at the auto-generated delete actions in Grails 1 controllers, there's definitely explicit handling of database integrity issues like this. 

So I think we still have some work to do:

a) see how other projects handle these issues (search for some of these exceptions in github search to see if people are explicitly handling unique exceptions
b) continue the discussion in a tech huddle and try to come to a consensus 
c) register a generic exception handler mapped in UrlMappings

jmiranda:
Based on what I've seen with the deletes, it seems reasonable to explicitly catch DataIntegrityViolationExceptions in the controller, especially if we need to display a specific error message. 

https://github.com/search?q=catch+%28DataIntegrityViolationException+language%3AGroovy+path%3Agrails-app%2Fcontrollers&type=code

Here's someone who has the same gut feeling as @kkaczmarczyk re: explicitly catching these exceptions in the controller.

https://stackoverflow.com/questions/36498327/catch-dataintegrityviolationexception-in-transactional-service-method

And this one is not exactly advocating for one way or the other, but discusses the losing of information due to the fact that exceptions are wrapped.

https://stackoverflow.com/a/2110836/136597

jmiranda:
Lastly, one other avenue to explore would be to add a controller-level exception handler for some exceptions. 

https://www.tothenew.com/blog/handling-exceptions-in-grails/

I think there's a way to make this more generic (define the handler in a base class and each controller that needs to handle it would extend that base class).

https://www.tothenew.com/blog/handling-exceptions-in-grails/

I don't know if that helps us and more than the UrlMapping / ErrorsController, but worth looking into.

kchelstowski:
To extend what @jmiranda said about:

> One thing that kind of concerns me with this solution is that now we will have two exactly same constraints (one on domain class and the other on DB table), so if we proceed with the current solution, do you think I should remove the product_package_uniq_idx constraint?

Those are two separate things - if you have a constraint on domain, it does not "add" a constraint to the DB table. If we used the auto generation (ddl) of tables, then it would add this to the table creation, but since we have to do the migrations manually, we have to kinda keep the constraints the same on domain and on the table, but manually. The same case is with the relations - if you make some OneToMany or ManyToMany and not using the ddl, you are the one, who has to take care of creating a foreign key/a separate table to hold foreign keys.

kchelstowski:
@jmiranda @drodzewicz 

> Based on what I've seen with the deletes, it seems reasonable to explicitly catch DataIntegrityViolationExceptions in the controller, especially if we need to display a specific error message.

Remember, that if we decide to catch the exception manually, we'd have to proceed the eventual rollback manually. I know this would not be a big case for the `save` (`delete` would be worse), but still, we have to keep that in mind.

kchelstowski:
So as @jmiranda probably suggested - it'd be better to eventually add some kind of logic to the `ErrorsController` for this type of exception.

jmiranda (approved):

PR #4197, opened 2023-07-31
OBGM-624 Change CRUD operations performed in the organization controller to data service methods

drodzewicz:


kchelstowski:
@drodzewicz the problem with the `AssertionFailure` you've raised on Slack, was due to the roles orphanDelete.
Most probably the newer Hibernate version we are using is strict with the owners of those, and since we are inheriting from `Party` and `Party` is the owner of this orphanDelete, the error has been thrown.
After moving this part to the `Organization` domain, everything seems to be working fine.

PS. The Data Services are not the problem here, because the same is thrown when trying to delete via `organizationInstance.delete()`.

drodzewicz:
Thanks @kchelstowski for your help.
I mistakenly assumed that while **extending** _Organization_ with _Party_ it would inherit the mappings that are defined in the _Party class_.

Side note,  I was a bit skeptical if we should completely remove this cascade mapping from  _Party_ class so I tested it by bringing it back and having it both defined on Party and Organization and after that when trying to delete Organization it would still throw this Exception `Unable to determine collection owner identifier for orphan-delete processing` so I guess you were right to completely move it from Party to Organization.

grails-app/services/org/pih/warehouse/core/OrganizationDataService.groovy
```diff
@@ -0,0 +1,11 @@
+package org.pih.warehouse.core
+import grails.gorm.services.Service
+
+@Service(Organization)
+interface OrganizationDataService {
+    void delete(String id)
+
+    Organization save(Organization productGroup)
```
kchelstowski, line 8:
```suggestion
    Organization save(Organization organization)
```

drodzewicz: :hear_no_evil:

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/core/OrganizationController.groovy
```diff
@@ -102,9 +101,10 @@ class OrganizationController {
                         identifierService.generateOrganizationIdentifier(organizationInstance.name)
             }
 
-            if (!organizationInstance.hasErrors() && organizationInstance.save(flush: true)) {
-                flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'organization.label', default: 'Organization'), organizationInstance.id])}"
-                redirect(action: "edit", id: organizationInstance.id)
+            if (organizationInstance.validate()) {
+                Organization persistedOrganization = organizationDataService.save(organizationInstance)
```
alannadolny, line 105: I think this variable is not necessary, you can do only `organizationDataService.save(organizationInstance)` and leave this redirect below as it was before. You are passing reference to the organizationInstance object, so it should contain the id from the database after saving.

alannadolny (approved):

jmiranda (commented):
See my comments in the ticket. I think we can merge this one but please create a ticket to keep investigating. Something isn't right about this and I want to get to the bottom of it.

PR #4198, opened 2023-07-31
OBGM-603 Replace GORM event handler closures with methods

alannadolny:
Closures worked fine, but I also changed them to methods to make the code more consistent (I had to check all of the places during looking for createdBy / updatedBy, so there was no additional work because of that)

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -23,21 +23,16 @@ class Order implements Serializable {
 
     def beforeInsert() {
         createdBy = AuthService.currentUser
+        updatedBy = AuthService.currentUser
     }
 
     def beforeUpdate() {
         updatedBy = AuthService.currentUser
     }
 
-    def publishRefreshEvent() {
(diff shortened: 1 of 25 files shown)


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -23,21 +23,16 @@ class Order implements Serializable {
...
     }
 
-    def publishRefreshEvent() {
-        if (isPurchaseOrder) {
-            Holders.grailsApplication.mainContext.publishEvent(new RefreshOrderSummaryEvent(this))
-        }
-    }
-
-    def publishRefreshEvenBeforeDelete() {
+    def publishRefreshEvent(Boolean forceRefresh = false) {
```
kchelstowski, line 33: it should be named `isDelete` - the same, as the second parameter of the `RefreshOrderSummaryEvent`

jmiranda (approved):
lgtm. and thank you. :pray:

kchelstowski (approved):

drodzewicz (approved):

awalkowiak (approved):

PR #4200, opened 2023-08-01
OBGM-642 Error when try to delete lots on manage lot page

alannadolny:


grails-app/services/org/pih/warehouse/inventory/InventoryItemDataService.groovy
```diff
@@ -0,0 +1,15 @@
...
+import grails.gorm.services.Join
+import grails.gorm.services.Service
+
+@Service(InventoryItem)
+interface InventoryItemDataService {
+
+    InventoryItem save(InventoryItem inventoryItem);
+
+    @Join("product")
+    InventoryItem get(String id);
```
alannadolny, line 12: That's because `lazyInitializationException` was thrown during getting a product from inventoryItem: `inventoryItem?.product?.id`

awalkowiak: Where you are getting this product?

kchelstowski, line 12: I don't like this, because `get` is somewhat "basic" operation and we don't always want to fetch the product. It is a similar case if you made the `product` as fetch type eager by default.

Don't get me wrong - it's a good solution, but the method should be named `getWithProduct`/`findWithProduct` or something like this (check what is working): a resource that might help: https://gorm.grails.org/latest/hibernate/manual/index.html#dataServices

But I would leave `get` without any join fetches.

grails-app/services/org/pih/warehouse/inventory/InventoryItemDataService.groovy
```diff
@@ -1,15 +1,15 @@
...
 import grails.gorm.services.Service
 
 @Service(InventoryItem)
 interface InventoryItemDataService {
 
     InventoryItem save(InventoryItem inventoryItem);
 
-    @Join("product")
-    InventoryItem get(String id);
+    @Query("select item from InventoryItem item join fetch item.product where item.id = $id")
```
kchelstowski, line 11: couldn't you make it work with `@Join` annotation? 
I was hoping it'd work, because in the resource I've sent, there was an example of `findAnother` and was wondering what might hide under the `Another`. 
Did you check the same, but not with `get`, but with `find`? 
You could compare the classes that are generated with each of them.
I would be grateful for the summary of your tries.

kchelstowski: @alannadolny could you try this?
```groovy
@Join("product")
InventoryItem findWithProduct(String id)
```

alannadolny: I tried this, but I didn't restart the application and it didn't work without it 🙈 🙉

kchelstowski: @alannadolny cool, sorry for nitpickness, but if `getWithProduct` would also work, I would prefer this version instead of `find`.

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryItemDataService.groovy
```diff
@@ -0,0 +1,15 @@
...
+import grails.gorm.services.Join
+import grails.gorm.services.Service
+
+@Service(InventoryItem)
+interface InventoryItemDataService {
+
+    InventoryItem save(InventoryItem inventoryItem);
+
+    @Join("product")
+    InventoryItem getWithProduct(String id);
```
jmiranda, line 12: I don't want methods like getWithX() unless that's an agreed upon convention in the world of data services.

I'm ok with this just being called get() and we always join on Product with get(). It probably makes more sense to add the join to the domain mapping at that point though so we get that join with any fetch from the database.

Or if that's not suitable we should look into other ways to name methods that eager fetch associations.

jmiranda (changes requested):
See comment

drodzewicz (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryItemDataService.groovy
alannadolny: I found that getWithX() is being used in a lot of articles. I think we shouldn't get Product with each InventoryItem - the main idea of using lazy loading is to avoid fetching unused entities. Moreover, it's not the best idea from a performance perspective.

Articles with this kind of naming:

https://vladmihalcea.com/join-fetch-pagination-spring/?fbclid=IwAR3U_-QucAZoqOXSeA4N4wa3EnTNI-sU7_vNgW32jclD898TaWLwmebUFc4
image


(This one I found in polish language, but it also contains this naming)
https://www.sztukakodu.pl/lazy-initialization-exception-jak-sobie-z-nim-radzic-spring-hibernate/
image

(And two threads from StackOverflow)
https://stackoverflow.com/questions/34193716/spring-jpa-service-calls-vs-jpa-query
image

https://stackoverflow.com/questions/55727390/jpql-query-for-page-of-children-and-join-fetch-of-grandchildren-throws-exception
image

drodzewicz: I believe it is a valid pattern that is used in grails project @service layer, so to me this look good.
I am very skeptical to the idea of adding eager fetching to associations on domains. Eager fetching is very costly on performance and we should always be aware of this so in my opinion whenever we want to eager load we should always explicitly specify it in the method.

kchelstowski: thanks for those resources @alannadolny .
I will only confirm my initial requests - I don't feel comfortable with basic `get` having any eager fetches. Changing the fetch type to EAGER on domain is also an anti-pattern and might cause performance issues (unless we ALWAYS need the product to be fetched eagerly). 
You can just imagine how costly it would be if we had a loop and were doing those gets.
As Dariusz mentioned above - if we want to eager fetch anything, we should just create a proper join fetch query for that not to kill the lazy-loading flow.

I think this is a common pattern at least in Spring to name those methods like this according to resources Alan has sent.

awalkowiak: If @jmiranda you don't like `getWithX`, then another option would be to name this as `getEagerly` or `findEagerly` as it is done here: https://guides.grails.org/grails-gorm-data-services/guide/index.html#joinQuery

cc @alannadolny

awalkowiak (commented):
Please remove semicolons.

grails-app/services/org/pih/warehouse/inventory/InventoryItemDataService.groovy
kchelstowski: I'd have mixed feelings about `getEagerly`, because in my opinion this was only an example name that they provided.
In my opinion such a method would sound like it fetches every association eagerly, not only one or two.

Assuming we wanted to use `getEagerly` when joining only `Product`, then how would we name a method that would eagerly fetch `Product` and e.g. `comments`? 

I guess this is an important decision we have to make, because a lot of such methods are gonna be created soon.

awalkowiak (approved):

PR #4202, opened 2023-08-01
OBGM-586 Fix user reference that caused header of stocklist not editable

kchelstowski:
The problem was difficult to identify, because random errors were thrown at the end of the session - sometimes regarding the template, sometimes regarding the `isUserAdmin` etc functions, so that led me to the point, that something must be wrong with the user.

The error I was getting the most times:
Screenshot from 2023-08-01 14-26-19

The problem is that when we bind the properties to `requisition`, we also bind users instances here like `createdBy`, `requestedBy`, `updatedBy`.

When assigning the `requisition.updatedBy = session.user`, we were trying to assign a User object of identifier that is already known for this session (read the error message).
So in fact we were having two "different" (the same) objects with the same ID in the same session, which Hibernate doesn't allow.
Look at the numbers of the User object:

1. The `createdBy`, `updatedBy`, `requestedBy` after binding (23141):
Screenshot from 2023-08-01 14-22-32
2. `session.user` (22004):
Screenshot from 2023-08-01 14-23-27

3. `AuthService.currentUser` (23141):
Screenshot from 2023-08-01 14-24-13

So you can see that `AuthService` is "compatible" with the session boundary, but the `session.*` itself, seems to be "deprecated" and it "lives" in its own session boundary.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/requisition/RequisitionTemplateController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/requisition/RequisitionTemplateController.groovy
@@ -14,6 +14,7 @@ import org.apache.commons.lang.StringEscapeUtils
 import grails.plugins.csv.CSVWriter
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.product.Product
+import org.pih.warehouse.auth.AuthService
 
 @Transactional
 class RequisitionTemplateController {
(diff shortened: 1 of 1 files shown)


jmiranda (approved):
looks good to me. review the comment in the ticket for more information about this bug.

PR #4203, opened 2023-08-02
OBGM-619 Include out of stock checkbox not working on inventory browser 

alannadolny:
The `showOutOfStockProducts` was not bound correctly, it is explained here: https://stackoverflow.com/questions/60990362/grails-command-object-not-binding-to-def-type

Part of the documentation linked in question above:
> Properties which are not bindable by default are those related to transient fields, dynamically typed properties and static properties.

### DIFF
--- a/src/main/groovy/org/pih/warehouse/inventory/InventoryCommand.groovy
+++ b/src/main/groovy/org/pih/warehouse/inventory/InventoryCommand.groovy
@@ -31,11 +31,11 @@ class InventoryCommand implements Validateable {
     List searchResults = []
 
     // indicates whether to display hidden products
-    def showHiddenProducts = Boolean.FALSE
+    Boolean showHiddenProducts = Boolean.FALSE
     // indicates whether unsupported products for the warehouse should be included
-    def showUnsupportedProducts = Boolean.FALSE
+    Boolean showUnsupportedProducts = Boolean.FALSE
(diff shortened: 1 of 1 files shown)


kchelstowski (approved):

kchelstowski:
I am wondering if you shouldn't also change the booleans above. Won't they also cause problems in the future?

alannadolny:
As I remember, we discussed on the tech huddle that we shouldn't do more work than it's described in the ticket to track the changes better. Moreover, it needs some checks about how it is used - is it set in the searching function, or should it be just bound from the params.

src/main/groovy/org/pih/warehouse/inventory/InventoryCommand.groovy
```diff
@@ -37,7 +37,7 @@ class InventoryCommand implements Validateable {
     // indicates whether non-inventory products for the warehouse should be included
     def showNonInventoryProducts = Boolean.FALSE
     // indicates whether out of stock products for the warehouse should be included
-    def showOutOfStockProducts = Boolean.FALSE
+    Boolean showOutOfStockProducts = Boolean.FALSE
```
awalkowiak, line 40: I think this should be also done for the booleans above since it will be causing the same issue.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4208, opened 2023-08-04
OBGM-639 Error while trying to import category excel file

alannadolny:


drodzewicz (approved):
LGTM.
I assume that this is a case where grails1 and grails 3 have different paths to the contentType property.
From searching the project on keyword `fileItem.contentType` I was able to find few more places where it is being used.
image
Please create a new ticket if it doesn't exist yet to investigate those other cases

kchelstowski (approved):

awalkowiak (approved):

PR #4211, opened 2023-08-07
OBGM-608 Handle errors while generating product identifier during create, update and import

drodzewicz:
There were several issues connected to this ticket so let me explain them quickly

#### How to reproduce
To reproduce this issue, you would need to throw an exception while creating/updating or importing a product, that way a transaction would end, and the issue of lazyInitializationException would be thrown for many fields.
One of the ways to reproduce this in our app was to choose a ProductType which had a **product Identifier Format** without a **0** in it, thus when generating the product code it would throw an exception.
https://github.com/openboxes/openboxes/blob/92d83295b4ffd52f9dfe9c2a364657d264c21a34/grails-app/services/org/pih/warehouse/core/IdentifierService.groovy#L248-L250

 #### Ignored exceptions while generating a product identifier
Even though exceptions were thrown during the process of generating product identifiers, they were not handled whatsoever, instead, they were caught and ignored. 
https://github.com/openboxes/openboxes/blob/92d83295b4ffd52f9dfe9c2a364657d264c21a34/grails-app/services/org/pih/warehouse/product/ProductService.groovy#L1129-L1142
To give some kind of feedback to the user I decided that it would be better to rethrow this exception and show the message to the user.
I am not too sure why are we not handling this validation on productType create and prevent users from creating invalid formats.

#### LazyInitializationException problem after the exception was thrown
A second issue was that since an exception was thrown during the transaction it rolled-back and ended the transaction, so later when we are trying to access some of the fields like `requiredFields` on productType or `supportedActivities` on location, we are not able to because these values are lazy loaded and we have no access to the.
So to fix this issue I needed to eagerly fetch all of the necessary properties to avoid the `LazyInitializationException`.

I was stuck on this part for a while consulting with @kchelstowski and researching on the internet, and this seems like an approach that most of the developers working with Hibernate take in their projects.


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -266,8 +275,12 @@ class ProductController {
...
                 // when the session is closed.
                 if (productInstance.validate()) {
-                    if (!productInstance.productCode) {
-                        productInstance.productCode = productService.generateProductIdentifier(productInstance.productType)
+                    try {
+                        if (!productInstance.productCode) {
+                            productInstance.productCode = productService.generateProductIdentifier(productInstance.productType)
+                        }
+                    } catch (Exception e) {
+                        productInstance.errors.rejectValue("productCode", e.message)
```
kchelstowski, line 283: I'd for now let the errors handler do the job, instead of having to write a lot of `try/catches`. If the users complain about the "generic" error message for this one (I assume they wouldn't, because this error should be rare), we can refactor it in MVP1/2.
As refactor I'd rather mean refactor the errors handler.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -814,12 +814,16 @@ class ProductService {
...
 
             if (!product?.id || product.validate()) {
-                if (!product.productCode) {
-                    product.productCode = generateProductIdentifier(product.productType)
+                try {
+                    if (!product.productCode) {
+                        product.productCode = generateProductIdentifier(product.productType)
+                    }
+                } catch (Exception e) {
+                    product.errors.rejectValue("productCode", e.message)
```
kchelstowski, line 822: same as above

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1137,6 +1141,7 @@ class ProductService {
 
         } catch (Exception e) {
             log.warn("Error generating unique product code " + e.message, e)
+            throw e
```
kchelstowski, line 1144: I'd remove this `try/catch` at all, since it does only `log.warn` and it prevents our errors handler from working if you don't do `try/catch` also in controller as you did above.

grails-app/services/org/pih/warehouse/product/ProductDataService.groovy
```diff
@@ -0,0 +1,23 @@
...
+
+    Product get(String id)
+
+    @Query("""Select p from Product p
+              left join fetch p.documents
+              left join fetch p.synonyms
+              left join fetch p.productCatalogItems pci
+              left join fetch pci.productCatalog
+              where p.id=$id""")
+    Product getWithDocumentsAndSynonymsAndProductCatalog(String id)
```
kchelstowski, line 21: for me it looks good, since it is a known convention, but I know there will be a discussion around that.
`getEagerly`/`getWithAssociations` or anything similar seems not appropiate to me, since we MIGHT have some other similar methods and this kind of name would be confusing, because as I mentioned somewhere else - such names would fit if we fetched ALL possible associations

grails-app/taglib/org/pih/warehouse/InventoryTagLib.groovy
```diff
@@ -20,11 +20,10 @@ class InventoryTagLib {
     def inventoryService
 
     def productStatus = { attrs, body ->
-        def product = Product.get(attrs.product)
```
kchelstowski, line 23: :+1:

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
drodzewicz: Sure I see your point and it makes sense so, for now, I'll just leave the error handling to the generic controller event handler but I am not too sure how are we going to handle these kinds of exceptions for template views generically, I still think we will have to manually try/catch but I guess we'll see on MVP2

drodzewicz:
Based on all of the points we discussed on yesterday's tech-huddle i reevaluated this issue and came to a much more comfortable solution than before _(at least I think...)_

I was not aware of the fact that there were certain (invalid) productTypes in the database that we can't do anything about, which is why we were catching the exception in `generateProductIdentifier()` and not propagating it to the user, but instead, we return an empty string as a **productCode** and relying on the `assignProductIdentifier` job to assign it a product code, buuuut... even if it was expected behavior I don't think it ever worked like that since by assigning an empty string to a product code it wouldn't even let us create the product since there is a `nullable: false` constraint on `prodocutCode`.

For my solution, firstly I added a constraint validator to the `productIdentifierFormat` on ProductType to prevent these _invalid_ productType from being created.
Secondly,  during the `generateProductIdentfier` I first check if the productType is valid, if yes then create a code based on this productType, if not then ignore it and as a fallback create the product code based on identifier format specified in the config. 

Now, since this issue does not throw an exception, there is no need to eagerly fetch all of this data and let the ORM do it's job.

I'd like to continue this discussion at the next tech-huddle, but I hope I am going in the right direction. 
cc @jmiranda

kchelstowski:
@drodzewicz I have an interesting discovery about this certain thing after reading your last comment. Will show it live on a tech huddle. Basically after throwing an exception we clear the persistence context and that's why afterwards we were getting those lazy inititalization exceptions e.g. in the domains.

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/product/ProductType.groovy
```diff
@@ -34,7 +34,9 @@ class ProductType {
         name(blank: false, unique: true)
         code(nullable: true, unique: true)
         productTypeCode(nullable: false)
-        productIdentifierFormat(nullable: true)
+        productIdentifierFormat(nullable: true, validator: { val ->
+            return val.contains("0") ? true : ["productType.identifierFormatNotContainSequentialPart.message"]
```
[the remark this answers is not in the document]
awalkowiak: Plus it needs to be adjusted in the `generateSequentialProductIdentifier` to not throw the `IllegalArgumentException` if there is no sequential part in the `PFI`

jmiranda:
> even if it was expected behavior I don't think it ever worked like that since by assigning an empty string to a product code it wouldn't even let us create the product since there is a `nullable: false` constraint on `prodocutCode`.

Can you show me where you are seeing productCode nullable: false? 

This is what I see on the Product domain.
```
productCode(nullable: true, maxSize: 255, unique: true)
```

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -340,9 +338,11 @@ class IdentifierService {
...
         }
 
-        // If product type is not deafult and has sequential identifier format or code,
-        // then generate sequentional, otherwise random
-        if (productType?.productIdentifierFormat?.contains("0") || productType?.code) {
+        // If product type is not default
+        // and has sequential identifier format containing "0" character which indicates the sequential part
+        // or identifier format is not specified when generatorTypeCode is SEQUENCE but has a code
+        // then generate sequential, otherwise random
+        if (productType?.productIdentifierFormat?.contains("0") || (!productType?.productIdentifierFormat && productType?.code && generatorTypeCode == IdentifierGeneratorTypeCode.SEQUENCE)) {
```
drodzewicz, line 345: this long if statement bothers me a bit and probably would be better to put it in a named variable, but I am unsure of the naming for it.
Maybe (probably not) `Boolean isSequentialWithCodeAndNoPIF`
If you have any suggestions let me know or we can leave it as is.

jmiranda: I agree but can't think of a good name either. Maybe it's less about the specific logic and more about what you plan to do if it's true.
```
Boolean requireSequentialIdentifier = ...
```

awalkowiak: @alannadolny since Dariusz is out of the office today, could you make this small improvement?

awalkowiak: @alannadolny since Dariusz is out of the office today, could you make this small improvement?

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
```diff
@@ -340,9 +338,14 @@ class IdentifierService {
...
         }
 
-        // If product type is not deafult and has sequential identifier format or code,
-        // then generate sequentional, otherwise random
-        if (productType?.productIdentifierFormat?.contains("0") || productType?.code) {
+        // If product type is not default
+        // and has sequential identifier format containing "0" character which indicates the sequential part
+        // or identifier format is not specified when generatorTypeCode is SEQUENCE but has a code
+        // then generate sequential, otherwise random
+        Boolean shouldGenerateSequential = !productType?.productIdentifierFormat &&
```
awalkowiak, line 345: @alannadolny `!productType?.productIdentifierFormat &&` this part is not required here, since this does not indicate if it needs to be sequential (plus it don't have to be checked at all at this level)

awalkowiak (approved):

PR #4212, opened 2023-08-07
OBGM-627 Unable to go forward/backward on Address page in location configuration 

alannadolny:
In this ticket, I found two issues that needed fixing. First was connected to `flattenRequest` function, which we used to flatten our request before being sent (In Grails 1). Apart from flattening, this function removed properties that were null.
I logged results for empty properties in the address with `flattenRequest` and without it. The results were as follows:
image
So, removing empty properties solved saving addresses and moving forward and backward without addresses.
After fixing it, I was thinking about the case when we want to delete the address and going forward and backward - removing inputted address and going backward didn't send the changes we made, because we removed it in the solution described above. So, I had to check again what's the difference between Grails 1 and Grails 3 ... And I noticed that when we didn't input the address we sent an empty address object, but when we added an address and then removed it, we sent an object like this:
[Screenshot from Grails 1]
image
So I did the same payload as we had in Grails 1 and noticed that the validation is also thrown after removing the address. After searching for this behavior I found a difference in config. Grails 3 didn't allow us to bind empty strings, because he converts empty strings to null, and then the validation fails on `nullable: false` in `Address` domains constraints. 
image
And below that, I found a setting that leaves the empty string as an empty string, not a null.
image
In Grails 1, it wasn't converted to null, so this change can potentially fix some similar issues. In GitHub, I found that someone noticed this change between Grails 3.1.9 and 3.3.1.

There is some information about it:
[GitHub issue with this change]
https://github.com/grails/grails-core/issues/10846
[Similar problem in stackOverflow]
https://stackoverflow.com/questions/24207822/why-are-empty-strings-converted-to-null-passing-to-the-constructor-of-the-domain
[Docs with this information about nullable]
https://grails.github.io/legacy-grails-doc/3.1.10/ref/Constraints/nullable.html

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -198,6 +198,7 @@ grails:
     databinding:
         dateFormats:
             - 'MM/dd/yyyy'
+        convertEmptyStringsToNull: false
 endpoints:
     jmx:
         unique-names: true
--- a/src/js/api/urls.js
+++ b/src/js/api/urls.js
@@ -54,6 +54,7 @@ export const USERS_OPTIONS = `${API}/users`;
 export const LOCATION_API = `${API}/locations`;
 export const LOCATION_TYPES = `${LOCATION_API}/locationTypes`;
 export const LOCATION_TEMPLATE = `${CONTEXT_PATH}${LOCATION_API}/template`;
+export const LOCATION = id => `${LOCATION_API}/${id}`;
 
 // ORDER
 export const ORDER = `${CONTEXT_PATH}/order`;
--- a/src/js/components/locations-configuration/LocationAddress.jsx
+++ b/src/js/components/locations-configuration/LocationAddress.jsx
@@ -9,8 +9,9 @@ import { withRouter } from 'react-router-dom';
 import Alert from 'react-s-alert';
 
 import { hideSpinner, showSpinner } from 'actions';
+import { LOCATION } from 'api/urls';
 import TextField from 'components/form-elements/TextField';
-import apiClient, { flattenRequest } from 'utils/apiClient';
+import apiClient, { mapToEmptyString } from 'utils/apiClient';
 import { renderFormField } from 'utils/form-utils';
 import Translate, { translateWithDefaultMessage } from 'utils/Translate';
 
@@ -99,14 +100,14 @@ class LocationAddress extends Component {
   }
 
   saveAddressOfLocation(values, callback) {
-    const valuesAsAddressObject = {
-      address: {
-        ...values,
-      },
-    };
     this.props.showSpinner();
-    const locationUrl = `/api/locations/${this.state.locationId}`;
-    apiClient.post(locationUrl, flattenRequest(valuesAsAddressObject))
+    const payload = mapToEmptyString(values, [null]);
+    // If field was edited it is undefined and we need to send empty string
+    // to remove it from location. If it is null, there is no need to send it, because
+    // it will be null be default
+    // We can't send empty address, because address.address is not nullable field
+    const addressPayload = !_.isEmpty(payload) ? { address: payload } : {};
+    apiClient.post(LOCATION(this.state.locationId), addressPayload)
       .then(() => {
         this.props.hideSpinner();
         callback({
--- a/src/js/utils/apiClient.jsx
+++ b/src/js/utils/apiClient.jsx
@@ -8,6 +8,7 @@ import { confirmAlert } from 'react-confirm-alert';
 import notification from 'components/Layout/notifications/notification';
 import LoginModal from 'components/LoginModal';
 import NotificationType from 'consts/notificationTypes';
+import { object } from 'prop-types';
 
 export const justRejectRequestError = (error) => Promise.reject(error);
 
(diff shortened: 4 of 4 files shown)


src/js/api/urls.js
```diff
@@ -53,6 +53,7 @@ export const USERS_OPTIONS = `${API}/users`;
 // LOCATIONS
 export const LOCATION_API = `${API}/locations`;
 export const LOCATION_TYPES = `${LOCATION_API}/locationTypes`;
+export const LOCATION_POST = id => `${LOCATION_API}/${id}`;
```
kchelstowski, line 56: this should be renamed, because could be used for get one, delete, edit

src/js/components/locations-configuration/LocationAddress.jsx
```diff
@@ -99,14 +100,12 @@ class LocationAddress extends Component {
...
-      },
-    };
     this.props.showSpinner();
-    const locationUrl = `/api/locations/${this.state.locationId}`;
-    apiClient.post(locationUrl, flattenRequest(valuesAsAddressObject))
+    const payload = Object.keys(values).reduce((acc, curr) => {
+      if (values[curr]) return { address: { ...acc.address, [curr]: values[curr] } };
+      return { address: { ...acc.address, [curr]: '' } };
+    }, { address: {} });
+    apiClient.post(LOCATION_POST(this.state.locationId), payload)
```
kchelstowski, line 108: This doesn't feel right if we don't want to send address at all - right now you always send address with empty strings.

If I am not right, at least add the brackets after `if`, and format the JSON not to have whole JSON in one line please.
```javascript
if (values[curr]) {
  ...
}
```

kchelstowski: also consider adding a comment of what is going on

kchelstowski (approved):

grails-app/conf/application.yml
```diff
@@ -198,6 +198,7 @@ grails:
     databinding:
         dateFormats:
             - 'MM/dd/yyyy'
+        convertEmptyStringsToNull: false
```
awalkowiak, line 201: I think this would be worth putting in a separate commit with a proper commit message that would not be squashed.

src/js/components/locations-configuration/LocationAddress.jsx
```diff
@@ -99,14 +100,32 @@ class LocationAddress extends Component {
...
-      },
-    };
     this.props.showSpinner();
-    const locationUrl = `/api/locations/${this.state.locationId}`;
-    apiClient.post(locationUrl, flattenRequest(valuesAsAddressObject))
+    const payload = Object.keys(values).reduce((acc, curr) => {
+      // If field was edited it is undefined and we need to send empty string
+      // to remove it from location. If it is null, there is no need to send it, because
+      // it will be null be default
+      if (values[curr] === null) {
```
awalkowiak, line 108: Perhaps this is something worth moving to a util API client function? Like with have `parseResponse` or `flattenRequest`.

awalkowiak (changes requested):

src/js/utils/apiClient.jsx
```diff
@@ -137,6 +137,23 @@ export const handleValidationErrors = (setState) => (error) => {
   return handleError(error);
 };
 
+export const mapToEmptyString = (values, valuesToSkip = []) => Object.keys(values)
```
awalkowiak, line 140: @alannadolny Will this support nested objects?

grails-app/conf/application.yml
jmiranda: I don't really know enough about what's happening here so I'm definitely going to advocate for a separate spike for this one. What impact does it have on the API in general? I would assume that the correct behavior would be "user doesn't enter any address info, client sends a NULL address or does not include the address object". If we are including empty string for address properties does that mean we'd get a validation error since streetAddress is nulable:false? Again, my assumption would be that the address should be ignored if there are no properties set. 

If we do a spike, we should also consider using trimString. 
```
databinding.trimStrings = true
```

jmiranda (commented):
I need to leave a review to get this out of my pending reviews list so here's my review:
```
Awaiting more information
```

If you guys make progress and feel good about merging, please do.

src/js/utils/apiClient.jsx
alannadolny: After my current commit - yes 😀

grails-app/conf/application.yml
alannadolny: This setting basically restores the Grails 1 behavior. The conversion of blank strings to null was introduced in Grails 2.3. So I think it can help keep the API consistent with this from Grails 1.

Getting back to the address binding:
In Grails 1
When we didn't enter an address we just send an empty object - I did the same in Grails 3
When we entered an address, we sent the data that we entered - I did the same
When we removed the address after saving it, we sent an object with empty strings as properties - I restored the same behavior in Grails 3

> If we are including empty string for address properties does that mean we'd get a validation error since streetAddress is nullable:false?
Yes, I got an error, because of that. Before we sent empty strings, so it works (Grails 1 data binding didn't convert empty strings to null)

awalkowiak (approved):

PR #4213, opened 2023-08-07
OBGM-638 Make LoadDataService Transactional

drodzewicz:


drodzewicz:
The issue was that some of the data relies on previously imported data like fro example **Catalogs** and **productCatalogItems**.
When importing productCatalogItems it tries to query the catalog which is not yet available since the transaction has not yet ended.

kchelstowski (approved):

alannadolny (approved):

PR #4217, opened 2023-08-09
OBGM-650 Unable to import dates in product source preference import 

alannadolny:
This issue was caused by importing dates as Excel dates, not as a string. It leads to different dates in the table and in the message above the table. It's visible on the screenshot that Katarzyna pasted to the ticket:
image
This number is visible instead of the normal date, because of the method that Excel uses to store his dates - it is a number of days counted from 1st January of the 1900 year.

Here is an explanation of how it works:
https://community.fabric.microsoft.com/t5/Desktop/Convert-a-date-value-like-42917-to-date/m-p/559598

So, I created a function `getDateFromExcel` which converts days to normal dates. In this function, I had to subtract two days from the result, because of the Excel oddity. 
The first day to subtract is because of that, Excel counts 01/01/1900 as a first day, but we are counting this day as a day 0. 
The second day is a result of the "Lap year problem": https://en.wikipedia.org/wiki/Leap_year_problem

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
@@ -30,16 +30,10 @@ class ProductSupplierPreferenceDataService {
             def supplierCode = params.supplierCode
             def organizationCode = params.organizationCode
             def preferenceTypeName = params.preferenceTypeName
-            Date validityStartDate, validityEndDate
-            if (params.validityStartDate.isNumber()) {
-                validityStartDate = CSVUtils.parseDateFromExcel(params.validityStartDate as Integer)
-                validityEndDate = CSVUtils.parseDateFromExcel(params.validityEndDate as Integer)
-                params.validityStartDate = validityStartDate.format("MM/dd/yyyy")
-                params.validityEndDate = validityEndDate.format("MM/dd/yyyy")
-            } else {
-                validityStartDate = Date.from(params.validityStartDate)
-                validityEndDate = Date.from(params.validityEndDate)
-            }
+            params.validityStartDate = params.validityStartDate?.isNumber() ?
+                    CSVUtils.getDateFromExcel(params.validityStartDate as Integer) : params.validityStartDate
+            params.validityEndDate = params.validityEndDate?.isNumber() ?
+                    CSVUtils.getDateFromExcel(params.validityEndDate as Integer) : params.validityEndDate
 
             if (!supplierCode) {
                 command.errors.reject("Row ${index + 1}: Product source code is required")
@@ -59,19 +53,20 @@ class ProductSupplierPreferenceDataService {
                 command.errors.reject("Row ${index + 1}: Preference Type with name: ${preferenceTypeName} does not exist")
             }
 
-            if (validityStartDate) {
+            def dateFormat = new SimpleDateFormat("MM/dd/yyyy")
+            if (params.validityStartDate) {
                 try {
-                    validityStartDate.format("MM/dd/yyyy")
+                    dateFormat.parse(params.validityStartDate)
                 } catch (Exception e) {
-                    command.errors.reject("Row ${index + 1}: Validity start date ${validityStartDate} is invalid")
+                    command.errors.reject("Row ${index + 1}: Validity start date ${params.validityStartDate} is invalid")
                 }
             }
 
-            if (validityEndDate) {
+            if (params.validityEndDate) {
                 try {
-                    validityEndDate.format("MM/dd/yyyy")
+                    dateFormat.parse(params.validityEndDate)
                 } catch (Exception e) {
-                    command.errors.reject("Row ${index + 1}: Validity end date ${validityEndDate} is invalid")
+                    command.errors.reject("Row ${index + 1}: Validity end date ${params.validityEndDate} is invalid")
                 }
             }
         }
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
@@ -12,7 +12,6 @@ package org.pih.warehouse.data
 import grails.gorm.transactions.Transactional
 import org.pih.warehouse.core.Organization
 import org.pih.warehouse.core.PreferenceType
-import org.pih.warehouse.importer.CSVUtils
 import org.pih.warehouse.importer.ImportDataCommand
 import org.pih.warehouse.product.ProductSupplier
 import org.pih.warehouse.product.ProductSupplierPreference
@@ -27,13 +26,11 @@ class ProductSupplierPreferenceDataService {
         log.info "Validate data " + command.filename
         command.data.eachWithIndex { params, index ->
 
-            def supplierCode = params.supplierCode
-            def organizationCode = params.organizationCode
-            def preferenceTypeName = params.preferenceTypeName
-            params.validityStartDate = params.validityStartDate?.isNumber() ?
-                    CSVUtils.parseExcelDate(params.validityStartDate as Integer) : params.validityStartDate
-            params.validityEndDate = params.validityEndDate?.isNumber() ?
-                    CSVUtils.parseExcelDate(params.validityEndDate as Integer) : params.validityEndDate
+            String supplierCode = params.supplierCode
+            String organizationCode = params.organizationCode
+            String preferenceTypeName = params.preferenceTypeName
+            Date validityStartDate = params.validityStartDate
+            Date validityEndDate = params.validityEndDate
 
             if (!supplierCode) {
                 command.errors.reject("Row ${index + 1}: Product source code is required")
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -200,4 +200,11 @@ class CSVUtils {
         detector.dataEnd();
         return detector.getDetectedCharset() ?: 'MacRoman';
     }
+
+    static String getDateFromExcel(Integer daysNumber) {
```
drodzewicz, line 204: `getDateFromExcel()` sound fine to me but maybe a better naming would be something like `parseExcelDate` instead?

drodzewicz (approved):
overall I think i looks good

kchelstowski (approved):
good catch

jmiranda:
I just did a quick check. DataType gets you a date and DateJavaType gets you a datetime. 

image

There's still a validation error but that's probably because the code is trying to parse a date. We should be able to leave this code
```
Date validityStartDate = params.validityStartDate
Date validityEndDate = params.validityEndDate
```
and get rid of the two code blocks used to parse the start and end dates. 
```
def dateFormat = new SimpleDateFormat("MM/dd/yyyy")
if (validityStartDate) {
    try {
        dateFormat.parse(validityStartDate)
    } catch (Exception e) {
        command.errors.reject("Row ${index + 1}: Validity start date ${validityStartDate} is invalid")
    }
}
```

jmiranda (changes requested):
See comments

alannadolny:
I didn't remove those two blocks to parse dates, just changed them a little bit, to show dates in the format that we had previously in the table.
@jmiranda Thanks a lot for the information that I can change the types in the `propertyMap`!

src/main/groovy/org/pih/warehouse/importer/ProductSupplierPreferenceImporter.groovy
```diff
@@ -34,8 +34,8 @@ class ProductSupplierPreferenceImporter extends AbstractExcelImporter {
             organizationCode     : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
             organizationName     : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
             preferenceTypeName   : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
-            validityStartDate    : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
-            validityEndDate      : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            validityStartDate    : ([expectedType: ExpectedPropertyType.DateJavaType, defaultValue: null]),
```
jmiranda, line 37: This should probably just be DateType since we don't need the timestamp component.

grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
```diff
@@ -50,20 +50,20 @@ class ProductSupplierPreferenceDataService {
                 command.errors.reject("Row ${index + 1}: Preference Type with name: ${preferenceTypeName} does not exist")
             }
 
-            def dateFormat = new SimpleDateFormat("MM/dd/yyyy")
+            String dateFormat = "MM/dd/yyyy"
             if (validityStartDate) {
                 try {
-                    dateFormat.parse(validityStartDate)
+                    params.validityStartDate = validityStartDate.format(dateFormat)
```
jmiranda, line 56: Why are we going backwards here?

src/main/groovy/org/pih/warehouse/importer/ProductSupplierPreferenceImporter.groovy
jmiranda: Unless you can explain why we need the DateJavaType.

jmiranda (changes requested):

src/main/groovy/org/pih/warehouse/importer/ProductSupplierPreferenceImporter.groovy
alannadolny: DataType is LocalDate, so I got it formatted other than 'MM/dd/yyyy', so that's the reason that I have to assign it to params.validityStartDate with format method. Because of that assignment, I got a date displayed in the table with the proper format and also properly imported. I changed it to DateType, but I am still using the format method.
By the "DataType is LocalDate" I mean the fact that my imported Date looks like this (at the left the date is using the format method, at the right is just DateType):
image

Because of using DateType / DateJavaType in our import, wrong data is not visible in the table to import, so it can be imported - it can lead to entering wrong data, with some missing dates.
By this, I mean:
- My Excel looks like this:
image
- Dates to import look like this (I didn't see any validation)
image
Let me know what you think about this behavior of importing with `ExpectedPropertyType.DateType` - maybe it can be the one argument for using our own date parser

For now, I am pushing the change with DateJavaType -> DateType and with getting rid of:
def dateFormat = new SimpleDateFormat("MM/dd/yyyy")
if (validityStartDate) {
    try {
        dateFormat.parse(validityStartDate)
    } catch (Exception e) {
        command.errors.reject("Row ${index + 1}: Validity start date ${validityStartDate} is invalid")
    }
}

grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
```diff
@@ -24,13 +24,14 @@ class ProductSupplierPreferenceDataService {
...
 
-            def supplierCode = params.supplierCode
-            def organizationCode = params.organizationCode
-            def preferenceTypeName = params.preferenceTypeName
-            def validityStartDate = params.validityStartDate
-            def validityEndDate = params.validityEndDate
+            String supplierCode = params.supplierCode
+            String organizationCode = params.organizationCode
+            String preferenceTypeName = params.preferenceTypeName
+            params.validityStartDate = params?.validityStartDate?.format(dateFormat)
```
jmiranda, line 33: I'm confused why you keep trying to do this. You were right about using DateJavaType as it's difficult to convert from LocalDate to Date. But once you define those columns as DataJavaType you don't need to do anything else with the date parameters (no parsing, no formatting). Just use the `validatityStartDate` and `validatityEndDate` as Date objects.

So the PR should look like this.
https://github.com/openboxes/openboxes/pull/4228/files

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/data/ProductSupplierPreferenceDataService.groovy
alannadolny: @jmiranda I pushed the code that you proposed. I tried to parse the date because with the current version of the code it looks like that:
image
I think it's not the date format that we would like to have.

src/main/groovy/org/pih/warehouse/importer/ProductSupplierPreferenceImporter.groovy
jmiranda: The PR looks good. However, you're right ... the silent failure on invalid dates doesn't feel right. I'll let @awalkowiak make a final decision on whether we're ok merging as-is or whether we want to do deal with that in this PR. 

The only way I can think of doing that is turning the data type back to String and doing the parsing / validation ourselves again. Otherwise, if the importer plugin just returns null for bad data, then we won't have a reference to the bad data and we'll never know if a value was entered into the column.

One other way that might not yet be explored is if there's some way to configure the importer to warn us when it gets invalid data. That would be the best solution.

jmiranda: Alright, I did a little digging and found that there's a way to do this through the plugin infrastructure. 
```
    static cellReporter = new DefaultImportCellCollector()

    List<Map> getData() {
        return Holders.grailsApplication.mainContext.getBean("excelImportService")
                .convertColumnMapConfigManyRows(workbook, columnMap, cellReporter, null, propertyMap)
    }
```
I'm not sure if static is correct here but that's what I saw with the excel import plugin test code.
https://github.com/gpc/grails-excel-import/blob/master/src/test/groovy/sample/BookImporter.groovy

But that might have been done as a convenience for testing purposes.

then we can get error messages using the cell reporter (after the data has been uploaded)
```
dataImporter.cellReporter.messagesBySeverityLevel
```

Unfortunately, there's an error when the DefaultImportCellCollector insteance tries to report an error for a cell with JavaDateType property type. The default cell reporter doesn't seem to know how to handle this data type.
```
2023-08-20 09:29:03,527 ERROR [080-exec-7] o.g.p.e.DefaultImportCellCollector      : Exception while trying to report potential problem
groovy.lang.MissingMethodException: No signature of method: org.grails.plugins.excelimport.ImportSeverityLevelEnum.severityLevel() is applicable for argument types: (java.lang.Integer, org.grails.plugins.excelimport.ExpectedPropertyType) values: [1, DateJavaType]
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:71)
        at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:53)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
        at org.grails.plugins.excelimport.ImportSeverityMappingEnum$severityLevel.call(Unknown Source)
        at 
```
So I think we just need to override the default cell reporter or create our own CellReporter class to log these issues and prevent the import from proceeding if there are errors.

My two cents: This is a refactoring and should be a separate ticket. 

And instead of putting the decision on @awalkowiak to merge or not merge, it'll probably be best to ask Kelsey and Manon if they are ok with the silent fail behavior here. If not then we'll need to implement this custom error reporter mechanism, so this ticket becomes at least a 5.

jmiranda: And for now, if Kelsey and Manon *do* want to prevent bad date data, then we can go back to handling the date as a StringType or DoubleType.

https://stackoverflow.com/a/19028309/136597

```
import org.apache.poi.ss.usermodel.DateUtil

Date validityStartDate
if (params.validityStartDate) {
    validityStartDate = DateUtil.getJavaDate(params.validityStartDate)
}
```

PR #4220, opened 2023-08-10
OBGM-654 Add additional sort fallback condition for receiptItems sorting

drodzewicz:
The issue was that there was a case where it was impossible to sort this list since all of the values specified in the comparator conditions had the same values so it resulted in a random order list each time it was generated.

 To fix this I added an additional last fallback comparator to sort by product name.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
+++ b/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
@@ -76,5 +76,16 @@ class Receipt implements Serializable, Comparable<Receipt> {
                         id <=> otherReceipt?.id
     }
 
+    List<ReceiptItem> sortReceiptItemsBySortOrder() {
+        def receiptItemsComparator = { a, b ->
+            return a.shipmentItem?.requisitionItem?.orderIndex <=> b.shipmentItem?.requisitionItem?.orderIndex ?:
+                    a.shipmentItem?.sortOrder <=> b.shipmentItem?.sortOrder ?:
+                            a?.sortOrder <=> b?.sortOrder ?:
+                                    a?.inventoryItem?.product?.name <=>  b?.inventoryItem?.product?.name
+        }
+
+        return receiptItems?.sort(receiptItemsComparator)
+    }
+
 
 }
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -1057,7 +1057,8 @@ class StockMovementService {
         List<ReceiptItem> receiptItems = shipments*.receipts*.receiptItems?.flatten()?.sort { a, b ->
             a.shipmentItem?.requisitionItem?.orderIndex <=> b.shipmentItem?.requisitionItem?.orderIndex ?:
                     a.shipmentItem?.sortOrder <=> b.shipmentItem?.sortOrder ?:
-                            a?.sortOrder <=> b?.sortOrder
+                            a?.sortOrder <=> b?.sortOrder ?:
+                                    a?.inventoryItem?.product?.name <=>  b?.inventoryItem?.product?.name
         }
         return receiptItems
     }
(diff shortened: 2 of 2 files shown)


drodzewicz:
A question to developers. @jmiranda @awalkowiak @kchelstowski @alannadolny 
I have noticed that there are many instances in the code where a "pretty long" sorting function is defined in the service or a controller, some of them are even duplicated across the codebase.

Do you think we should define these kinds of sorting methods on the domain class like I did in the PR? I saw that we have implemented this kind of method on some domains but not everywhere, or maybe instead create a separate service method for that specific sort.

I wasn't able to find any "best practices" for this kind of case but I think it all comes down to making the code more readable and maintainable.

kchelstowski (approved):
@drodzewicz in my opinion those should be in the domain. A default comparator `compareTo` is also defined in the domains.

jmiranda:
@drodzewicz Yes, this is a really good candidate for a https://github.com/openboxes/openboxes/labels/tech%20debt ticket.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1054,21 +1054,13 @@ class StockMovementService {
 
     List<ReceiptItem> getRequisitionBasedStockMovementReceiptItems(def stockMovement) {
         def shipments = Shipment.findAllByRequisition(stockMovement.requisition)
-        List<ReceiptItem> receiptItems = shipments*.receipts*.receiptItems?.flatten()?.sort { a, b ->
-            a.shipmentItem?.requisitionItem?.orderIndex <=> b.shipmentItem?.requisitionItem?.orderIndex ?:
-                    a.shipmentItem?.sortOrder <=> b.shipmentItem?.sortOrder ?:
-                            a?.sortOrder <=> b?.sortOrder
-        }
+        List<ReceiptItem> receiptItems = shipments*.receipts?.flatten()*.sortReceiptItemsBySortOrder()?.flatten()
```
awalkowiak, line 1057: Is the flatten at the end required?

jmiranda: I think the one in the middle is unnecessary, but the one on the end is necessary.  I'm starting to think this just needs to be a query in a data service getSortedReceiptItems(). But that can be a refactoring for another day.

grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
```diff
@@ -76,5 +76,16 @@ class Receipt implements Serializable, Comparable<Receipt> {
                         id <=> otherReceipt?.id
     }
 
+    List<ReceiptItem> sortReceiptItemsBySortOrder() {
```
jmiranda, line 79: I think this is fine, but I think a getSortedReceiptItems() - especially if there's only one way to sort - would be best.

Therefore you can just say 
```
shipments*.receipts*.sortedReceiptItems
```
We should also maybe think about pushing the comparator down to the child's compareTo() method and then it becomes
```
shipments*.receipts*.receiptItems.flatten().sort()
```
or in the case of multiple comparators
```
def receiptItems = Collections.sort(shipments*.receipts*.receiptItems.flatten(), ReceiptItem.customComparator1())
```
The one thing that bothers me is the fact that we're going through multiple objects to get these receipt items. Instead, we should have a named query and/or data service method that retrieves and sorts receipt items. But again, that refactoring can wait until later.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
drodzewicz: both are necessary.
Notice that for `getShipmentBasedStockMovementReceiptItems()` it has `shipment` and for `getRequisitionBasedStockMovementReceiptItems()` it is `shipments`

awalkowiak:
I am merging as is, and I am adding the topics covered here to my list of tech debt/ tickets to create

PR #4221, opened 2023-08-10
OBGM-622 Previous and Next Refresh not filled with value on transaction report

alannadolny:
The `triggers` variable didn't contain any triggers after calling the `getTriggersOfJob` method. As I checked `JobKey` created a key with the `DEFAULT` group:
image
In the `quartzScheduler` I found that `RefreshTransactionFactJob` belongs to the `GRAILS_JOBS` group. After adding this second parameter, the triggers were correctly found.
image

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
@@ -279,7 +279,7 @@ class ReportController {
         command.location = Location.get(session.warehouse.id)
         command.rootCategory = productService.getRootCategory()
 
-        def triggers = quartzScheduler.getTriggersOfJob(new JobKey("org.pih.warehouse.jobs.RefreshTransactionFactJob"))
+        def triggers = quartzScheduler.getTriggersOfJob(new JobKey("org.pih.warehouse.jobs.RefreshTransactionFactJob", "GRAILS_JOBS"))
         def previousFireTime = triggers*.previousFireTime.max()
         def nextFireTime = triggers*.nextFireTime.max()
         def locationKey = LocationDimension.findByLocationId(command?.location?.id)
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -279,7 +279,7 @@ class ReportController {
         command.location = Location.get(session.warehouse.id)
         command.rootCategory = productService.getRootCategory()
 
-        def triggers = quartzScheduler.getTriggersOfJob(new JobKey("org.pih.warehouse.jobs.RefreshTransactionFactJob"))
+        def triggers = quartzScheduler.getTriggersOfJob(new JobKey("org.pih.warehouse.jobs.RefreshTransactionFactJob", "GRAILS_JOBS"))
```
jmiranda, line 282: Where does this magic string come from?

alannadolny: From quartzScheduler/jobStore/jobsByKey in the debugger 😀
image

jmiranda: Sorry I meant where is it defined? In our code or the plugin?

Quick search on GitHub ... 
https://github.com/search?q=GRAILS_JOBS+repo%3Agrails-plugins%2Fgrails-quartz&type=code

reveals that it's defined as a constant in the Quartz plugin
```
GrailsJobClassConstants.DEFAULT_GROUP = "GRAILS_JOBS"
```
So we should use that. 

Or better yet, see if the getTriggers() method (that is dynamically added to the Job class by the plugin) returns the triggers.

    def triggers = RefreshTransactionFactJob.getTriggers()

jmiranda: Looks like this 
```
def triggers = RefreshTransactionFactJob.getTriggers()
```
returns the closure from the job class so that's not what we want. 

So just clean up the magic string and call it a day.
```
def triggers = quartzScheduler.getTriggersOfJob(new JobKey("org.pih.warehouse.jobs.RefreshTransactionFactJob", GrailsJobClassConstants.DEFAULT_GROUP))
```

jmiranda (changes requested):
Apply the magic string fix and we're good to go.

awalkowiak (approved):

PR #4222, opened 2023-08-10
OBGM-659 Change origin and destination validators on Order

drodzewicz:
location organzation should not be required on wards, dispensary and pharmacy type locations


jmiranda (approved):
This seems like a reasonable change, so I'm fine with moving it forward. However, it feels like the "ward" logic itself was flawed.

The organization requirement is only necessary for purchase orders. So we probably should have checked the order type rather than the location type of origin/destination. 

If the location type is important, can we think of a way to handle this through supported activities instead of location types?

 @awalkowiak thoughts?

alannadolny (approved):

kchelstowski (approved):

awalkowiak:
@jmiranda so maybe this should simply check if the location supports `PLACE_ORDER`?

PR #4223, opened 2023-08-10
OBGM-620 UI issue in Fr locale: where tabs on PO add items are displayed on the same page

alannadolny:
This issue was caused by the quotation marks used in languages other than English. In the confirmation dialog in English, we used single quotation marks
image
But In the French language, we were using both - single quotation marks and standard quotation marks:
image
So changing it to backticks fixed this issue

### DIFF
--- a/grails-app/views/purchaseOrder/_showOrderItems.gsp
+++ b/grails-app/views/purchaseOrder/_showOrderItems.gsp
@@ -509,7 +509,7 @@
                * TODO here, we may need to elsewhere -- should we escape all
                * TODO localized strings? Further discussion at OBGM-343.
                */
-              if ($("#validationCode").val() == 'WARN' && !confirm(htmlDecode("${g.message(code: 'orderItem.warningSupplier.label')}"))) {
+              if ($("#validationCode").val() == 'WARN' && !confirm(htmlDecode(`${g.message(code: 'orderItem.warningSupplier.label')}`))) {
                 return false
               } else {
                 $.ajax({


grails-app/views/purchaseOrder/_showOrderItems.gsp
```diff
@@ -509,7 +509,7 @@
                * TODO here, we may need to elsewhere -- should we escape all
                * TODO localized strings? Further discussion at OBGM-343.
                */
-              if ($("#validationCode").val() == 'WARN' && !confirm(htmlDecode("${g.message(code: 'orderItem.warningSupplier.label')}"))) {
+              if ($("#validationCode").val() == 'WARN' && !confirm(htmlDecode(`${g.message(code: 'orderItem.warningSupplier.label')}`))) {
```
kchelstowski, line 512: what caused the differences in quotations, that in english it was `'low quality'` and in French it was `"low quality"`?
Is it `g.message` implementation that was parsing it? Or is it `htmlDecode`? 
What's the behavior in Grails 1 for this?

kchelstowski (approved):

grails-app/views/purchaseOrder/_showOrderItems.gsp
alannadolny: Yep, as you can see on this screenshot the string ended before "low" in French
image

alannadolny: In Grails 1 we are replacing those strings
image
In the browser it looks like that:
image

kchelstowski: Ok, I just thought it was working "magically" on Grails 1 and was wondering what made it not working here. But if it's `replaceAll` missing, then I'm fine. thanks

awalkowiak (approved):

grails-app/views/purchaseOrder/_showOrderItems.gsp
jmiranda: So what's the proper approach there? Use the replaceAll or backticks?

jmiranda: Reminder: Sometimes messages include HTML tags and attributes that use double quotes so we need a solution that will work for all of these cases. Meaning, we can't just encode/decode our way out of this, as far as I can tell.

1. Use backticks like Alan did above (not sure of the pros / cons)

1. Use replaceAll to remove double quotes and other characters (this might break HTML tags)

1. Remove double quotes and HTML tags from our messages.properties. (might be difficult since there are times where they might be necessary)

1. Make sure we're not using GSP in javascript like this (not sure if this would solve the problem, but this would be my recommended approach going forward).
```
javascriptFunctionCall("${g.message()"}
```

Aside: Honestly, I hate mixing javascript and GSP as it's much less readable and maintainable. We could even set up HTML elements in the document to handle the separation.
```
<input type="hidden" id="messages.supplierWarning" value="${g.message(code: 'orderItem.supplierWarning.label')}"
<input type="hidden" id="messages.unauthorizedError" value="${g.message(code: 'errors.unauthorized.label')}"
```
and then in the javascript
```
!confirm(htmlDecode($("#messages.supplierWarning").val())
```
Or have the page load the messages via API during page load and have javascript pull the errors from that dictionary.

--- 

NOTE: I don't have a strong opinion about the correct approach. We should discuss this here or in a tech huddle. 

For now, I'd like to have a reasonably straightforward solution (if not, a best practice) to complete this PR and then we should create an MVP2/post-migration ticket to refactor some of these cases based on

jmiranda (commented):
See comments

grails-app/views/purchaseOrder/_showOrderItems.gsp
alannadolny: As I can see using backticks is a popular approach, some examples:
 - [https://bobbyhadz.com/blog/javascript-escape-quotes-in-string]
image
In this article is also written that:
> The String.replaceAll() method returns a new string with the matches of the pattern replaced. The method doesn't change the original string. Strings are immutable in JavaScript.

So, I think `replaceAll` is a little bit worse than the backticks. Moreover, I think the backticks look much cleaner (just a personal opinion) and we don't have to change a lot as you showed in the example with hidden inputs. Using backticks we can leave our code as it is, we just have to change quotation marks to backticks in some places.

- [https://www.digitalocean.com/community/tutorials/how-to-work-with-strings-in-javascript]
image

PR #4225, opened 2023-08-11
OBGM-636 Fix ability to adjust inventory to a negative qty value

kchelstowski:
Creating a draft PR before I'm out of office.
Basically the `jquery-validation` library was not added, but even after adding it, it still doesn't work, but I hope it's a good starting point for anyone who takes this one on.

Resource that might help: https://github.com/limcheekin/grails-jquery-validation/

The changes I've made are basically copying the Grails 1 version - there was a missing package, a missing config (the config in the `application.yml` I've added), and the import of resources in the `custom.gsp`.

What I've tried and is not in the commit:

- changing `implementation` to `runtime`, but no success either.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -432,6 +432,12 @@ dependencies {
      */
     runtime "com.bertramlabs.plugins:asset-pipeline-grails:${assetPipelineVersion}"
 
+    // https://mvnrepository.com/artifact/org.grails.plugins/jquery-validation-ui
+    implementation 'org.grails.plugins:jquery-validation-ui:1.4.7'
+
+    // https://mvnrepository.com/artifact/org.grails.plugins/jquery-validation
+    implementation 'org.grails.plugins:jquery-validation:1.9'
+
     implementation "com.github.librepdf:openpdf:${openPdfVersion}"
 
     // security patch
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -648,6 +648,65 @@ uiperformance:
     enabled: false
 
 # TODO OBGM 1: Add jqueryValidation plugin config if will be needed
+
+jqueryValidation:
+    packed: true
+    cdn: false  # false or "microsoft"
+    additionalMethods: false
+
+jqueryValidationUi:
+    errorClass: 'error'
+    validClass: 'valid'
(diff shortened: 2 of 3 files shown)


jmiranda:
Nice work, even if it's not working. I think if this becomes a big headache, we'll need to assess whether it's worth the effort. That means checking to see where it's being used and if it's not critical we should remove it or if possible, replace it with server-side validation in all of those cases. I believe the server-side validation is already there so removal is also an option.

alannadolny:
I spent some time working on this, searching for possible solutions - but I am still stuck in the same place. I checked how many places have this kind of validation in the case of removing jquery-validation-ui - there are 11 files using it.

This validation mentioned in the ticket can be replaced by adding `min: 0` to the property newQuantity of the `AdjustStockCommand`. In case of using server-side validation should we create another ticket for covering other places or just replace them within this ticket?
cc @jmiranda

jmiranda:
> This validation mentioned in the ticket can be replaced by adding min: 0 to the property newQuantity of the AdjustStockCommand. 

Ooh, I like that as a quick fix. And then we can discuss the ideal solution for MVP2.

> In case of using server-side validation should we create another ticket for covering other places or just replace them within this ticket?

There should be server-side validation already. The problem is that we don't (at least I don't think we do) have a user-friendly way to handle the validation errors in the modal. The modal closes and the error messages (if we did it correctly) show on the underlying page. 

So we have a few options
 
1. [js] Either we send the form as an async request and add in our own ability to validate the request and create an error response using our API. We then just need to render the errors in the modal similar to the jqval plugin.  
2. [gsp] We treat the form submission as a regular request and render the error on the underlying page (rather than in the modal - which may be confusing to the user). This should be happening on some features already.
3. [gsp] Move the modal to its own GSP page and use a normal form submission so that the error handling will happen on the new page.
4. Move the jqval plug into our codebase as an inline plugin and modify it to work for our use case - with the idea that we'll be removing / replacing it at some point in the future as we don't necessarily want to maintain this plugin for the long-term.
 
Let me know if you can think of others.

I prefer option 1 as it's the least disruptive to the user (and developer) and might be the least amount of work. But it also might come with some complexity that we haven't considered.

PR #4226, opened 2023-08-14
OBGM-648 Unable to edit Package size from other to each in product source import 

alannadolny:
When we were updating package size through import and we had previously added package and product price we were just updating our default product package like that:
`defaultProductPackage.productPrice.price = price`
So, the save action was performed only on `ProductPrice` object, because the reference didn't change. Our default product package is this one with the newest `lastUpdated`. So after this kind of update, the `lastUpdated` wasn't triggered and after update, we saw the same package size as we had before import.
As a solution, I decided to update the `lastUpdated` property manually.

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
@@ -202,6 +202,7 @@ class ProductSupplierDataService {
                 defaultProductPackage.productPrice = productPrice
             } else if (price && defaultProductPackage.productPrice) {
                 defaultProductPackage.productPrice.price = price
+                defaultProductPackage.lastUpdated = new Date()
             }
         }
 


grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy
```diff
@@ -202,6 +202,7 @@ class ProductSupplierDataService {
                 defaultProductPackage.productPrice = productPrice
             } else if (price && defaultProductPackage.productPrice) {
                 defaultProductPackage.productPrice.price = price
+                defaultProductPackage.lastUpdated = new Date()
```
jmiranda, line 205: Hmm, that doesn't seem right. I'm going to take a look at this when I get a chance.

jmiranda: Nevermind, now I understand. 

At first glance, I thought this was a cheap trick to get Grails to do something it should do on its own. Like when we update transaction entries and expect the transaction last updated to be updated as well. 

But we actually have logic where the default package is defined by the fact that it's the most recently updated package. 
```    
ProductPackage getDefaultProductPackage() {
    return productPackages ? productPackages.sort { it.lastUpdated }.last() : null
}
```
So while you are right, I think we need to be more explicit about what is happening here. Either we should add an explicit comment here to make it clear what's going on. Or better yet, add a method to the product supplier to make it obvious what we're doing. 
```    
void setDefaultProductPackage(ProductPackage productPackage) {
    productPackage.lastUpdated = new Date()
}
```
If using a setter on a transient property causes problems we can either remove "defaultProductPackage" from transients or add a custom method like this
```
void markAsDefaultProductPackage(ProductPackage productPackage) {
    productPackage.lastUpdated = new Date()
}
```
Using this approach leads to self-documented code and also gives us the flexibility to change the default behavior in the future. I could imagine adding an explicit defaultProductPackage property in the future, so we'd only really need to add a changeset to add the property (database column) and remove the getter/setter from ProductSupplier. The rest of the code would be the same.

jmiranda: For now, I'll merge this. But we should add a comment to the code or a tech debt ticket to handle the minor refactoring.

alannadolny: @jmiranda I created a ticket for that - If you would like to add something, there is a link for this task: https://pihemr.atlassian.net/browse/OBGM-680?atlOrigin=eyJpIjoiMDU0YWI3ZDZkNDgyNGExMjkzMWQ1ZTE0OGE5ZWIwMTIiLCJwIjoiaiJ9

kchelstowski (commented):
1. it's weird that it was working in Grails 1
2. I think a possible solution could be also to add an `beforeUpdate`/`beforeInsert` where we would set the `lastUpdated` like we do in many places.
3. I'm fine with that being merged, but it would be good to understand why it worked in Grails 1. Did you research that @alannadolny ?

kchelstowski:
actually I've checked the code and it seems like the `ProductPackage` domain already has those events I mentioned, but the `lastUpdated` is missing there.

awalkowiak:
@kchelstowski I think that the issue here is that what is updated here is `ProductPrice` not `ProductPackage` ->
`defaultProductPackage.productPrice.price = price`, and that's why the last updated is not changed on it tight now, and perhaps previously it was flushed :thinking: (?)

alannadolny:
@awalkowiak Yup, you're right

PR #4227, opened 2023-08-15
OBGM-626 Generate temporary changelog for schema and data for brand n…

jmiranda:
…ew installs


jmiranda:
I'm going to merge this one since it should only impact new installations with an empty database, of which we have none. This one still needs work as I haven't reviewed (nor am I confident) that the imported data is :100:. But I was able to create the database without the order_type_id bug and was able to log into the newly installed system.

PR #4229, opened 2023-08-17
OBGM-636 Able to adjust inventory to negative qty value 

alannadolny:
In the ticket, I included all of the places that used the jquery validation: https://pihemr.atlassian.net/browse/OBGM-636?atlOrigin=eyJpIjoiZmMwMmUzNDM3OTAwNDFlMjgyYjVmMTUzZjQ3MzMzN2QiLCJwIjoiaiJ9

I just added the validation on min qty to the command instance, so after inputting qty lower than 0, we will see the error message (All of the error messages are visible in this way in adjusting inventory):
image
With the jquery plugin, the validation looked like that:
image

The second error in adjusting inventory (reproduction steps 2 from the ticket) was caused by data binding. When we provided the transaction id using a hidden field the transaction wasn't bound correctly. In the params, it looks like that:
image
After removing this hidden field it looks like that:
image
So now, Grails is using nested values, instead of flattened.
Before removing this hidden field, we didn't have a transaction instance, and the .hasErrors method threw the error
image
After fixing, we have the transaction method and the .hasErrors method is accessible.
image

### DIFF
--- a/grails-app/views/inventory/_inventoryAdjustment.gsp
+++ b/grails-app/views/inventory/_inventoryAdjustment.gsp
@@ -2,7 +2,6 @@
 <g:form action="saveAdjustmentTransaction">
     <div class="box">
         <h2><g:message code="inventory.adjustStock.label"/></h2>
-        <g:hiddenField name="transactionInstance.id" value="${command?.transactionInstance?.id}"/>
         <g:hiddenField name="transactionInstance.inventory.id" value="${command?.warehouseInstance?.inventory?.id}"/>
         <g:hiddenField name="transactionInstance.transactionType.id" value="${command?.transactionInstance?.transactionType?.id }"/>
         <table>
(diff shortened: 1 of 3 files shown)


jmiranda (changes requested):
The min seems fine to add on the AdjustStockCommand, but there might be times where we want to allow negative quantities for other use cases. Although I think they might always be positive when we save to the transaction entry table since the transaction type code is what tells us whether to debit or credit.

Regarding the second issue, please revert the change and create a new ticket. In this case (adjust stock), it's likely that the transaction ID is not needed, but we can't just remove request parameters like this without understanding what's going on.

alannadolny:
@jmiranda I think there is no need to revert the second fix. When passing `transaction.id` we are trying to bind an object with this id, but it is empty (it is trying to be fetched), so this object doesn't exist yet - it's still not saved. So, I am passing the whole object to bind it correctly, without fetching by empty id. Moreover, we are passing the id in the object itself, so I didn't know what was the reason that we are passing it separately by the hidden field.
What do you think @awalkowiak and @kchelstowski?

awalkowiak (approved):
I am approving and merging this one as is (for the id parameter @alannadolny will create a ticket for further investigation, and regarding the adjusting to the negative quantity we can do some brainstorming on the tech huddle if we want to address it somehow because currently, transaction uses `Adjustment` type which has transaction code `CREDIT`)

alannadolny:
There is a new ticket for the investigation https://pihemr.atlassian.net/browse/OBGM-697?atlOrigin=eyJpIjoiYWNhZDM5MzA1NzUzNDVlODk0OTdmNTI4MGFmMTRhZDYiLCJwIjoiaiJ9

PR #4230, opened 2023-08-17
OBGM-633 Data chooser issues on record stock page 

alannadolny:


jmiranda (commented):
I don't like this syntax (`[Calendar.YEAR]`) but it seems like it was added a long time ago so I won't block this from being.
```
${ConfigHelper.getMinimumExpirationDate()[Calendar.YEAR]}
```
Either of the following seem more intuitive and correct.
```
${ConfigHelper.getMinimumExpirationDate().year}
```
or
```
${ConfigHelper.getMinimumExpirationDate().get(Calendar.YEAR)}
```

kchelstowski (approved):

drodzewicz (changes requested):
having something like 
`${ConfigHelper.minimumExpirationDate.year}`
would be the best and nicer solution but I feel like it is going to be harder to implements.

Correct me if I am wrong, but I believe when you have a getter class method specified like `def getSomething` you can access it's value by calling the property without get and a method call like `obj.something`.
 
So I think for now just use

```groovy
ConfigHelper.minimumExpirationDate[Calendar.YEAR]
```

drodzewicz (approved):

alannadolny:
@jmiranda `.year` is marked as deprecated and the `.get` method throws an exception. So, I am finally changing this to the version that Dariusz proposed.

awalkowiak (approved):

PR #4231, opened 2023-08-18
OBGM-662 Endless attempt to save (spinning) instead of not saved indication (red) when quantity input while offline

alannadolny:


jmiranda (commented):
I don't understand the problem well enough to review, but I don't like these kinds of changes because it feels like we're going to need to do this in a bunch of places. If it's the right thing to do then we need to make a ticket that's more generic that addresses. If there's a better way to handle it without changing `item['product.id']` then let's do that. In either case, let's have a tech huddle discussion.

kchelstowski (approved):

kchelstowski:
@jmiranda @alannadolny if I remember correctly @drodzewicz was working on spike which was supposed to target those. I don't remember if we have any outcome about this specific thing though, but I believe we discussed it many times.

drodzewicz:
> @jmiranda @alannadolny if I remember correctly @drodzewicz was working on spike which was supposed to target those. I don't remember if we have any outcome about this specific thing though, but I believe we discussed it many times.

You are right, I was working on a similar spike for which I was looking through the code and trying to find and fix these instances of flattened properties, I must have missed that one.

awalkowiak (approved):

PR #4232, opened 2023-08-18
OBGM-671 Unable to import locations through new import page

alannadolny:


grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -18,17 +18,16 @@ import org.apache.poi.hssf.usermodel.HSSFWorkbook
...
 import grails.plugins.csv.CSVMapReader
+import org.pih.warehouse.data.LocationDataService
 import org.pih.warehouse.importer.ImportDataCommand
 import util.ConfigHelper
 
 @Transactional
 class LocationService {
 
     GrailsApplication grailsApplication
-    def dataService
```
alannadolny, line 28: dataService was not used, so I removed it

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -271,7 +271,7 @@ class LocationApiController extends BaseDomainApiController {
             throw new IllegalArgumentException("File cannot be empty")
         }
 
-        if (importFile.fileItem.contentType != "text/csv") {
+        if (importFile.contentType != "text/csv") {
```
jmiranda, line 274: Is this a verified API change for uploaded files? If so, let's include a link to that documentation in the PR and/or ticket.

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -18,17 +18,16 @@ import org.apache.poi.hssf.usermodel.HSSFWorkbook
...
-    def productTypeService
-    def userService
-
+    UserService userService
+    LocationDataService locationDataService
```
jmiranda, line 30: Do we actually use this? If not, we can remove this too.

jmiranda (commented):
See comments

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -18,17 +18,16 @@ import org.apache.poi.hssf.usermodel.HSSFWorkbook
...
 import util.ConfigHelper
 
 @Transactional
 class LocationService {
 
     GrailsApplication grailsApplication
-    def dataService
-    def productTypeService
-    def userService
-
```
kchelstowski, line 31: how about those? are they used? You mentioned only `dataService` that is not used

drodzewicz (approved):

grails-app/services/org/pih/warehouse/core/LocationService.groovy
alannadolny: @kchelstowski @jmiranda 
userService:
image
locationDataService:
image

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
alannadolny: We changed it many times, there is a ticket for investigating the rest of those places with `fileItem`: [ticket](https://pihemr.atlassian.net/browse/OBGM-669?atlOrigin=eyJpIjoiMWY2MWNiMzgwMjFjNDMzMzg3ZjNjNTY0ZGQzYjgwMDgiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

ImportFile is a type of MultipartFile - on this instance, we can call getContentType, getSize, getBytes, etc.
Docs: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/multipart/MultipartFile.html

awalkowiak (approved):

PR #4234, opened 2023-08-18
OBGM-679 Error message when try to edit transaction

alannadolny:


jmiranda (commented):
LGTM; but please note that we did this on purpose. I'm hoping that Grails 3 has fixed whatever issue required this. I don't recall exactly why this was necessary, but I believe Grails would try to prepend the template with "inventory" even if we specified the path with a leading forward slash.

So in the former version of template would correctly resolve to 
```
inventory/../transaction/details.gsp
```
and the latter version would be incorrect
```
/inventory/transaction/details.gsp
```

kchelstowski (approved):
lgtm, I remember fixing those type of errors like that even before the code freeze, so I assume this is an expected solution

drodzewicz (approved):

awalkowiak (approved):

PR #4236, opened 2023-08-21
OBGM-608 Add validation to ignore gl accounts if current location does not support required accounting activity

drodzewicz:
After QA Katarzyna pointed out that it should be possible to create a product without selecting the GL Account if you are logged in to a location without **required accounting** activity enabled.
After looking into this issue I only found this validation specified on the frontend  on form validation handler

https://github.com/openboxes/openboxes/blob/c5747dfb99214a680dac20c768b6d03c5f3afaf0/grails-app/views/product/edit.gsp#L114-L123

We should probably also do the validation on API as well, so I modified the `validateRequiredFields` method to account for this case.

The reason why this issue came out now is that in my previous pull request I removed the `?.save()` which was called on productInstance during `updateTags()` method call which cased a bit of a problem 

https://github.com/openboxes/openboxes/pull/4211/files#diff-c3f63f71bf15c619a5adb3baf3aa146ab7bd663d0668fa644745680ae57ae409L1049
image


grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -596,15 +596,19 @@ class Product implements Comparable, Serializable {
         }
     }
 
-    def validateRequiredFields() {
+    Boolean validateRequiredFields() {
+        return this.validateRequiredFields([])
+    }
+
+    Boolean validateRequiredFields(List<ProductField> ignoreProductFields) {
```
alannadolny, line 603: Did you try to assign an empty array as a default value instead of creating a second method?
I mean:
```
Boolean validateRequiredFields(List<ProductField> ignoreProductFields = []) {
```
And now when you call this method without a passed argument, the empty array will be used.

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -596,15 +596,19 @@ class Product implements Comparable, Serializable {
         }
     }
 
-    def validateRequiredFields() {
+    Boolean validateRequiredFields() {
+        return this.validateRequiredFields([])
```
kchelstowski, line 600: `this` is redundant

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -271,7 +276,11 @@ class ProductController {
                     }
                 }
 
-                productInstance.validateRequiredFields()
+                if (location.isAccountingRequired()) {
+                    productInstance.validateRequiredFields()
+                } else {
+                    productInstance.validateRequiredFields([ProductField.GL_ACCOUNT])
```
alannadolny, line 282:
```suggestion
List<ProductField> ignoreProductFields = location.isAccountingRequired() ? [] : [ProductField.GL_ACCOUNT]
productInstance.validateRequiredFields(ignoreProductFields)
```

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -182,7 +182,11 @@ class ProductController {
             }
         }
 
-        productInstance.validateRequiredFields()
+        if (location.isAccountingRequired()) {
+            productInstance.validateRequiredFields()
+        } else {
+            productInstance.validateRequiredFields([ProductField.GL_ACCOUNT])
+        }
```
alannadolny, line 189: The same as below

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -596,15 +596,19 @@ class Product implements Comparable, Serializable {
         }
     }
 
-    def validateRequiredFields() {
+    Boolean validateRequiredFields() {
```
kchelstowski, line 599: you could use primitive `boolean` for that, since it never returns `null`

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -596,15 +596,19 @@ class Product implements Comparable, Serializable {
...
+        return this.validateRequiredFields([])
+    }
+
+    Boolean validateRequiredFields(List<ProductField> ignoreProductFields) {
         if (!productType || !productType?.requiredFields || productType?.requiredFields?.isEmpty()) {
             return true
         }
 
-        def isValid = true
+        Boolean isValid = true
```
kchelstowski, line 608: same as above
```suggestion
        boolean isValid = true
```

grails-app/domain/org/pih/warehouse/product/Product.groovy
kchelstowski: btw I think this method is unnecessary, if you just specify a default argument for that in the line `603` as `[]`, so:
```groovy
 Boolean validateRequiredFields(List<ProductField> ignoreProductFields = [])
```

grails-app/domain/org/pih/warehouse/product/Product.groovy
drodzewicz: makes sense, will do

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
drodzewicz: this was my first version :D (look at the second commit) but I changed it because I thought it might look a bit better but I guess you guys disagree, I'll roll it back

grails-app/domain/org/pih/warehouse/product/Product.groovy
drodzewicz: I did not try this, I just followed the conversation that we usually do for these types of cases in the project

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -182,7 +182,8 @@ class ProductController {
             }
         }
 
-        productInstance.validateRequiredFields()
+        List<ProductField> ignoreFields = location.isAccountingRequired() ? [] : [ProductField.GL_ACCOUNT]
+        productInstance.validateRequiredFields(ignoreFields)
```
awalkowiak, line 186: You could also do here a location-based validation: `validateRequiredFieldsInLocation(location)`
```groovy
    Boolean validateRequiredFieldsInLocation(Location location) {
        if (location.isAccountingRequired()) {
            return validateRequiredFields([])
        }

        return validateRequiredFields([ProductField.GL_ACCOUNT])
    }
```
Which could be potentially expanded by other supported activities if needed.

drodzewicz: looks good, I'll do that

PR #4238, opened 2023-08-21
OBGM-666 Unable to filter by Blank Payment Terms on PO List

alannadolny:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -64,7 +64,6 @@ class OrderService {
     InventoryService inventoryService
     ProductSupplierDataService productSupplierDataService
     PersonDataService personDataService
-    def stockMovementService
```
alannadolny, line 67: I removed it because it was not used

drodzewicz (approved):

awalkowiak (approved):

PR #4239, opened 2023-08-23
OBGM-670 Unable to assign to user "No access" role

alannadolny:
It was interesting that in the debugger roles were bound correctly, but during saving the error with mismatching type was thrown. When I added "no access" to the roles and evaluated the user's role after binding everything looked good. When I let the debugger go line after the save method I saw that one error was visible.

👇 This part of the code tried to bind the roles with appropriate mapping to the Role object, but the "no access" had a "null" value and it could not be fetched, so I bound the roles manually. Grails 1 looks like he didn't have a problem with just skipping the "null"
```
userInstance.properties = params
```
I also moved the `updatedRoles` higher in the function to avoid fetching the roles twice.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -34,10 +34,11 @@ class UserService {
 
     def updateUser(String userId, String currentUserId, Map params) {
 
-        def userInstance = User.get(userId)
+        User userInstance = User.get(userId)
+        List<Role> updatedRoles = params.roles ? Role.findAllByIdInList(params.list("roles")) : []
 
         if (params.roles) {
-            userInstance.roles = Role.findAllByIdInList(params.list("roles"))
+            userInstance.roles = updatedRoles
             params.remove("roles")
         }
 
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -34,18 +34,23 @@ class UserService {
 
     def updateUser(String userId, String currentUserId, Map params) {
 
-        def userInstance = User.get(userId)
+        User userInstance = User.get(userId)
+        List<Role> updatedRoles = params.roles ? Role.findAllByIdInList(params.list("roles")) : []
```
kchelstowski, line 38: It actually doesn't matter, because when binding it has to fetch it anyway (but maybe only a proxy (?)), but you can check if just:
```groovy
userInstance.roles = params.list("roles")
```
would be enough

drodzewicz: this looks fine to me.
At first I thought that we might want to use `bindData(userInstance, params)` that way it will properly deal with binding and fetching all of the appropriate data for us, but apparently we can not use bindData outside of the controller _(at least it looks that way after tested it localy)_ so I am wondering if we maybe should move the binding part to controller and then pass the userInstance as a parameter and not a `user.id`
```groovy
bindData(userInstance, params)
 userInstance = userService.updateUser(userInstance, session.user.id, params)

```

alannadolny: I also thought about `bindData`, but you're right that we can't use it outside the controller. For me, it looks weird that we are passing a user instance as an argument and then we are assigning the result to the user instance

awalkowiak (approved):

PR #4240, opened 2023-08-23
OBGM-657 Handle validation exceptions generically on product groups

drodzewicz:
For the moment I am pushing and documenting everything that I was able to investigate about handling errors more generically for forms on GSP views, but unfortunately, I don't feel strongly about following this conversation, but let me explain everything in order.

## Investigation
At first, when I picked up this ticket I started browsing through many different public repositories in order to try to find any examples of other projects implementing at least a similar "Generic error handling" system for GSP views, but I wasn't successful in my search and noticed that most of the project handle errors separately or the most common approach is to render that generic error view like we already have
```groovy
render(view: "/error")
``` 
One thing that put me on the right path is the grails [documentation](https://guides.grails.org/grails3/command-objects-and-forms/guide/index.html) which describes how to properly handle form data. 
There was an example similar to many of our cases like:
```groovy

def save(Player player) {
    if (player == null) {
        render status: HttpStatus.NOT_FOUND
        return
    }

    if (player.hasErrors()) {
        respond player.errors, view: 'create'
        return
    }
}
```
If there are validation errors we return a  _response_ which takes in a **bean** and a **view** attribute _(besides many other attribuets)_ and renders the view, and to get errors from this bean we can render them using the grails tags
```gsp

<g:hasErrors bean="${player}">
    <div class="errors">
        <g:renderErrors bean="${player} as="list" />
    </div>
</g:hasErrors>
```
As you can see, in the GPS we are assigning bean as `player` even though we have not given it such an alias on the response 
`respond player.errors, view: 'create'`.
Grails is smart enough to find all of the beans that are provided in the view and to access it we need to call the object by its domain/class name _(in this case the Domain name is Player so we assign the beans as player)_.
If we want to include errors from all of the beans, we can omit the bean and just render
```gsp

<g:hasErrors>
    <div class="errors">
        <g:renderErrors as="list" />
    </div>
</g:hasErrors>
```
That way it will access all of the available beans and render all of the errors.
### My implementation
So after finding out that grails will take care of all of the errors in a more or less generic way, without me specifying the bean I thought that maybe I should move this logic to the global `ErrorsContorler` `handleValidationErrors()`.
The one challenge was to figure out how to render the appropriate GSP template view for each error, so for the time being I specified an action/error mapping variable that defines the relation to which view to render for each action which throws an exception 
https://github.com/openboxes/openboxes/blob/ddb403d7bbbe1c0c4795ad6f843eb6f0eb9d1069/grails-app/conf/runtime.groovy#L1076-L1080

So given we have an exception thrown on `ProductGroupController` `save,` in the error handler we extract this information from the request forwardURI
```groovy
def path = request.forwardURI - request.contextPath
def engine = groovyPagesTemplateEngine
```
which returns a `/productGroup/save`, which then we match with the errorMapper 
```groovy
Map templateMapping = grailsApplication.config.openboxes. error handler.templateMapping
def viewPath = templateMapping[path] ?: path
```
which returns a path to the view which we want to render `/productGroup/create`

then we check if such a view exists in the project 
```groovy
if (engine.getResourceForUri("${viewPath}.gsp").exists() || engine.getResourceForUri("/grails-app/views/${viewPath}.gsp").exists()) {
    respond errors.target, view: viewPath
    return
}
```
If it does exist then we render that view otherwise we render the generic error page
```groovy
render(view: "/error")
```

## The problem with this approach
One of the problems that I noticed was, in the case of the productGroup edit, when an exception was thrown and it was handled by this implementation of the generic error handler, it failed on lazyInitialization since inside of the `edit.gsp` we are accessing some associations _(product and siblings)_ lazily.
Since the exception was thrown before, it ended the session and we can't access them. 
A solution to that would be to eagerly fetch those _products_ and _siblings_ but we already discussed this in previous PR's and I think that it would be a little annoying to do everywhere and this is something that should be handled by the ORM.

Another problem is that sometimes we have a bit more complicated models that need to be passed to the view. In the case of the `productGroup/create` we only need the productGroup instance, but then there are sometimes more complicated models that are being passed down like
```groovy
render(template: 'products', model: [productGroup: productGroup, products: isProductFamily ? productGroup?.siblings : productGroup?.products])
```
And it would be pretty hard to handle this generically since `response errors.taget, view: viewPath` will have access only to the bean in which the validation was thrown.

In conclusion, I will leave this PR on `Draft` and we will probably discuss this on the tech huddle, maybe after discussing this with other developers something will click and I'll be able to finish it, but at the moment I will leave this PR and will go with a solution which is using the `try/catch` to catch the error and return the appropriate view and errors manually

### DIFF
--- a/grails-app/conf/runtime.groovy
+++ b/grails-app/conf/runtime.groovy
@@ -1073,6 +1073,12 @@ openboxes.supportLinks = [
     knowledgeBase: 'https://openboxes.helpscoutdocs.com/',
 ]
 
+openboxes.errorHandler.templateMapping = [
+        "/productGroup/save": "/productGroup/create",
+        "/productGroup/addToProductGroup": "/productGroup/create",
+        "/productGroup/update": "/productGroup/edit",
+]
+
 // Reset an instance
 
 openboxes.resettingInstance.command = "wget https://raw.githubusercontent.com/openboxes/openboxes/develop/reset-database.sh | sh"
--- a/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
@@ -12,6 +12,7 @@ package org.pih.warehouse.core
 import grails.converters.JSON
 import grails.core.GrailsApplication
 import org.grails.exceptions.ExceptionUtils
+import org.grails.gsp.GroovyPagesTemplateEngine
 import org.pih.warehouse.RequestUtil
 import org.springframework.validation.BeanPropertyBindingResult
 import util.ConfigHelper
@@ -24,6 +25,7 @@ class ErrorsController {
     GrailsApplication grailsApplication
     def userAgentIdentService
     def localizationService
+    GroovyPagesTemplateEngine groovyPagesTemplateEngine
 
     def handleException() {
         if (RequestUtil.isAjax(request)) {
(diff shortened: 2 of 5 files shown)


jmiranda:
Apologies. I need more time on this one.

PR #4242, opened 2023-08-23
OBGM-681 Unable to create location through Add Destination modal in Outbounds

alannadolny:


src/js/components/stock-movement-wizard/modals/AddDestinationModal.jsx
```diff
@@ -121,7 +121,7 @@ class AddDestinationModal extends Component {
     const url = '/api/locations?useDefaultActivities=true';
 
     const { name, locationType, ...address } = values;
-    const payload = { name, 'locationType.id': _.get(locationType, 'id') || '' };
+    const payload = { name, locationType: _.get(locationType, 'id') || '' };
```
kchelstowski, line 124: shouldn't it be: 
```javascript
const payload = { name, locationType: { id: locationType?.id ?? '' }};
```
previously you were passing `locationType.id: <id>`, and now you have `locationType: <id>`, but I assume if we are passing `locationType`, it should be `locationType: <entity>`.

Unless Grails magic interprets it correctly if we pass the id.

kchelstowski (approved):

drodzewicz (approved):

PR #4243, opened 2023-08-24
OBGM-629 Fix binding supportedActivities and displayedFields parameters when no selection is provided

drodzewicz:


grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy
```diff
@@ -94,16 +94,10 @@ class ProductTypeController {
...
-            if (params.requiredFields) {
-                params.requiredFields = params.list("requiredFields") as ProductField[]
-            }
-            if (params.displayedFields) {
-                params.displayedFields = params.list("displayedFields") as ProductField[]
-            }
-            productTypeInstance.properties = params
+            params.supportedActivities = params.list("supportedActivities")
+            params.displayedFields = params.list("displayedFields")
+            bindData(productTypeInstance, params)
```
kchelstowski, line 99: what was causing the issue?

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/product/ProductTypeController.groovy
drodzewicz: The `<g:select />` tag did not send any data when no options were selected causing some issues when trying to clear all of the options

PR #4244, opened 2023-08-24
OBGM-637 Error message when try to create attribute without input Code

alannadolny:
The first change:
```
render(view: "edit", model: [attributeInstance: attributeInstance])
```
is precisely the same as Justin wrote under the ticket - there is only an edit view, and the create view doesn't exist. Even the create action renders the edit view page. The issue connected to this was visible when creating a new attribute with existing code other than an empty string.

The second change connected to validation is because of the situation - when we didn't input code, it is an empty string and it didn't fail on `unique: true` (I mean the situation when one product attribute with empty code already exists, so it should fail). It was throwing an exception
image
Katarzyna asked me to make this error more meaningful, so I have created our "unique" validation which takes empty strings into account. By comparing ids I am checking if we are creating an attribute or we are editing an already existing one.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
@@ -53,7 +53,7 @@ class Attribute {
     }
 
     static constraints = {
-        code(nullable: true, unique: true, validator: { code, attribute ->
+        code(nullable: true, validator: { code, attribute ->
             Attribute existingAttribute = Attribute.findByCode(code)
             if(existingAttribute && existingAttribute.id != attribute.id) {
                 ['default.not.unique.message']
--- a/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
@@ -53,7 +53,12 @@ class Attribute {
     }
 
     static constraints = {
-        code(nullable: false, unique: true)
+        code(nullable: true, unique: true, validator: { code, attribute ->
+            Attribute existingAttribute = Attribute.findByCode(code)
+            if(existingAttribute && existingAttribute.id != attribute.id) {
+                ['default.not.unique.message']
+            }
+        })
         name(nullable: false, maxSize: 255)
         description(nullable: true)
         defaultValue(nullable: true)
--- a/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/Attribute.groovy
@@ -53,9 +53,9 @@ class Attribute {
     }
 
     static constraints = {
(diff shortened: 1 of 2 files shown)


grails-app/domain/org/pih/warehouse/product/Attribute.groovy
```diff
@@ -53,7 +53,12 @@ class Attribute {
     }
 
     static constraints = {
-        code(nullable: false, unique: true)
+        code(nullable: true, validator: { code, attribute ->
```
kchelstowski, line 56: is it intended that you made it nullable?

kchelstowski (commented):
+ from the description I still don't quite understand the problem that you had to refactor this unique part. Would this mean we have to change it for this kind of version wherever we have `unique: true`? 
From what I remember this works well in many places with correct messages like "Name must be unique".
I would rather take a look at those things (instead of redoing validator):

1. if when calling `attributeInstance.save(flush: true)`, the `attributeInstance` does already have some errors?
2. if answer for 1 is "No", I would consider calling `.validate()` before the save and see the behavior then.
3. to be announced when you check my first two :laughing:

grails-app/domain/org/pih/warehouse/product/Attribute.groovy
alannadolny: yh, no 🙈

kchelstowski: then you could remove it at all, since `nullable` is `false` by default:
https://docs.grails.org/latest/ref/Constraints/nullable.html

alannadolny: 1. No, `unique: true` doesn't catch this error if we pass an empty string, so I got this error during save:
image
2. I tried `validate`, but it also doesn't work.

grails-app/domain/org/pih/warehouse/product/Attribute.groovy
```diff
@@ -53,7 +53,12 @@ class Attribute {
     }
 
     static constraints = {
-        code(nullable: false, unique: true)
+        code(validator: { code, attribute ->
+            Attribute existingAttribute = Attribute.findByCode(code)
+            if(existingAttribute && existingAttribute.id != attribute.id) {
```
kchelstowski, line 58:
```suggestion
            if (existingAttribute && existingAttribute.id != attribute.id) {
```

kchelstowski (approved):

grails-app/domain/org/pih/warehouse/product/Attribute.groovy
alannadolny: In Grails 1 we got an error connected to a missing view page when proceeding with not unique value (screenshot from the ticket):
image

awalkowiak (approved):

PR #4245, opened 2023-08-24
OBGM-686 Unable to display transaction entries

alannadolny:


drodzewicz (approved):
Looks good to me.
Did something change between versions 1 and 3 that it was missing this scaffold?

From the documentation, I see that the only thing that changed was the definition of this property

> Note: The old alternative of defining scaffold property:
> ```groovy
> class BookController {
>     static scaffold = true
> }
> ```
> is no longer supported above Grails 3.0.

But here I see that it was completely missing :thinking:

kchelstowski (approved):
@drodzewicz nothing changed, it was just removed in one of the PRs. The reason would be unknown for now. https://github.com/openboxes/openboxes/commit/b98250eb2318a51acbac10035aecc669dfa4ede5

PR #4246, opened 2023-08-25
OBGM-683 Set flush on shipment delete method

drodzewicz:
I hesitated I bit before adding this `flush: true` property on shipment delete but at this moment I don't see an alternative solution.

For deleting a shipment we are handling any errors that come up using the try/catch block. Since we are not flushing the delete, it only throws an exception at the end of the transaction and renders the default error page, so adding the flush allows us to catch it in the middle of the transaction.


kchelstowski (approved):
Those are small functions that could be easily moved to the service layer without a harm, but I'm fine with leaving it for now.

alannadolny (approved):
I have the same opinion as @kchelstowski, but LGTM

awalkowiak (approved):

PR #4247, opened 2023-08-25
OBGM-690 Unable to clear values on Forecasting tab on edit location

alannadolny:
In one of my previous PRs, I introduced the `convertEmptyStringsToNull` setting in our config because, in the resources I found and pasted in https://github.com/openboxes/openboxes/pull/4212, it looked like it was introduced between Grails 3.1.9 and 3.3.1 versions of Grails and it looked like help for us to get closer with the Grails 1 behavior. This setting resolved actually some tickets, so I think it is a good change, but in this ticket, I found that we weren't able to assign empty strings to a few properties, because those empty strings were used in the BigDecimal constructor - it normally throws an error, so in this case Grails doesn't bind those values. Those values had to be null to be bounded correctly and they were nulls in Grails 1. 🤔 So it's a little bit different that in this case it is a different behavior than those I noticed before - and was noticed in some of the resolved tickets.

So, for those situations that need to convert empty strings to nulls, I created a generic data binding helper. It can be used by @BindUsing with this converter class as a parameter. It will be applied to the whole class on which it is applied. It can be also applied only for some of the properties, but I don't like how it will look in this case:
```
   // Lead time in days (safety stock is lead time days x daily forecast quantity)
    @BindUsing({ obj, source, propertyName -> source["expectedLeadTimeDays"] == "" ? null : source["expectedLeadTimeDays"]})
    BigDecimal expectedLeadTimeDays

    // Replenishment period in days
    @BindUsing({ obj, source -> source["replenishmentPeriodDays"] == "" ? null : source["replenishmentPeriodDays"]})
    BigDecimal replenishmentPeriodDays

    // Demand time period in days
    @BindUsing({ obj, source -> source["demandTimePeriodDays"] == "" ? null : source["demandTimePeriodDays"]})
    BigDecimal demandTimePeriodDays
```
If you guys think this approach with applying it only to some of the properties is better than what I pushed to the PR I am willing to change it.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
@@ -9,10 +9,13 @@
  **/
 package org.pih.warehouse.inventory
 
+import grails.databinding.BindUsing
+import org.pih.warehouse.NullBinderHelper
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.product.Product
 import util.InventoryUtil
 
+@BindUsing(NullBinderHelper)
 class InventoryLevel {
 
     String id
--- a/grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
@@ -15,7 +15,6 @@ import org.pih.warehouse.core.Location
 import org.pih.warehouse.product.Product
 import util.InventoryUtil
 
-@BindUsing(EmptyStringsToNullBinder)
 class InventoryLevel {
 
     String id
@@ -48,12 +47,15 @@ class InventoryLevel {
     BigDecimal forecastPeriodDays = 30
 
     // Lead time in days (safety stock is lead time days x daily forecast quantity)
+    @BindUsing({ obj, source -> EmptyStringsToNullBinder.bindEmptyStringToNull(source, "expectedLeadTimeDays")})
     BigDecimal expectedLeadTimeDays
 
     // Replenishment period in days
+    @BindUsing({ obj, source -> EmptyStringsToNullBinder.bindEmptyStringToNull(source, "replenishmentPeriodDays")})
     BigDecimal replenishmentPeriodDays
 
     // Demand time period in days
+    @BindUsing({ obj, source -> EmptyStringsToNullBinder.bindEmptyStringToNull(source, "demandTimePeriodDays")})
     BigDecimal demandTimePeriodDays
 
     // Preferred bin location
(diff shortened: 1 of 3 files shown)


alannadolny:
There is a documentation for `BindUsing`: https://grails.github.io/legacy-grails-doc/3.3.10/api/grails/databinding/BindUsing.html

kchelstowski (commented):
Even though this looks good to me, I have a few questions;

- I remember you saying that in Grails 1 the empty strings are NOT converted to nulls, so why in that case you say they were converted? Would it mean, that the type of the property also matters in this case? If it's `BigDecimal` that IS converted to `null`, maybe only for `String` type Grails 1 wasn't converting it to `null`? Would that be the truth?
- we should answer the 1st bullet, because unfortunately I think there might be places where we were relying on this binding, so I think a safer option would be to add this annotation to fields that really need it? Are you sure that whole `InventoryLevel` class (all properties) are fine with that annotation?

alannadolny:
@kchelstowski As I found empty strings were not converted and it was introduced later - as I saw it fixed some of our tickets (for example, this one: https://pihemr.atlassian.net/browse/OBGM-646?focusedCommentId=145182&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-145182), so this looks right I think, but I don't know why in this one case it looks different - it can be because of the `BigDecimal` but I couldn't find anything that would make me convinced about this.

I checked editing the rest of the fields and it looks good, but if you think it would be better to add this code only to those 3 properties I can add it, of course.

grails-app/utils/org/pih/warehouse/NullBinderHelper.groovy
```diff
@@ -0,0 +1,12 @@
+package org.pih.warehouse
+
+import grails.databinding.BindingHelper
+import grails.databinding.DataBindingSource
+
+class NullBinderHelper<T> implements BindingHelper<T> {
```
drodzewicz, line 6: I think we need to be more explicit with the naming of the binder like for example `EmptyStringsToNullBinder` so that we know exactly what is it doing

drodzewicz (changes requested):
Other than the naming change I requested, I think it's a good approach for now.
I know that the changes in grails 3 brought some unwanted side effect and it is too bad that we can't fix this issue with a config change _(at least that we know of at this moment)_, but I think it's a good approach given the problem.

drodzewicz (approved):

grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
```diff
@@ -9,10 +9,13 @@
  **/
 package org.pih.warehouse.inventory
 
+import grails.databinding.BindUsing
+import org.pih.warehouse.EmptyStringsToNullBinder
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.product.Product
 import util.InventoryUtil
 
+@BindUsing(EmptyStringsToNullBinder)
```
awalkowiak, line 18: @alannadolny do you want to use this binder on every property or on a specific one?

awalkowiak (commented):
Ok I am merging this one as is but I think this still needs discussion (@jmiranda)

PR #4248, opened 2023-08-25
OBGM-682 Remove locations withour organization from destination dropdown on PO create page

drodzewicz:


grails-app/views/purchaseOrder/_enterOrderDetails.gsp
```diff
@@ -88,6 +88,7 @@
                                         <g:selectLocation class="chzn-select-deselect filter"
                                                           id="destination.id"
                                                           name="destination.id"
+                                                          hasOrganization="${true}"
```
kchelstowski, line 91: I think `withOrganization` would sound better? :thinking:  but since I'm not master of naming, treat it as a loose suggestion

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
```diff
@@ -486,6 +486,10 @@ class SelectTagLib {
             ActivityCode activityCode = attrs.activityCode ?: null
             attrs.from = locationService.getAllLocations().sort { it?.name?.toLowerCase() }
 
+            if (attrs.hasOrganization) {
+                attrs.from = attrs.from.findAll { !!it.organization }
```
kchelstowski, line 490: I think the double negation is not needed here, `it.organization` should be enough.

kchelstowski:
And a clarifying question - is it something that was also not working in Grails 1 and that we decided to fix now? Because I kinda don't see any reason why it would work in Grails 1?

drodzewicz:
> And a clarifying question - is it something that was also not working in Grails 1 and that we decided to fix now? Because I kinda don't see any reason why it would work in Grails 1?

sorry for not clarifying this in the PR, I left a [comment under the ticket ](https://pihemr.atlassian.net/browse/OBGM-682?focusedCommentId=145217)

Basically, there was a change I did not too long ago, adding an additional condition to **origin** and **destination** constraint validators which in a way caused this issue. Additionally, I would need to add some new condition that would throw this error for Orders from PO but I feel this validator will become too big and unreadable, so after talking about it with Katarzyna we decided to just hide these locations from the dropdown

grails-app/taglib/org/pih/warehouse/SelectTagLib.groovy
awalkowiak, line 490: What's the point of filtering it here if you can already filter it at the "fetching" level

awalkowiak (changes requested):

awalkowiak (approved):

PR #4249, opened 2023-08-25
OBGM-686 Unable to display transaction entries (fix after QA)

alannadolny:
As I saw the data wasn't visible in any of our pages, so I fixed it for all of the pages connected to transaction entries. I don't know why IntelliJ added those spaces after saving those files, so I have to leave them.


kchelstowski (approved):
this is weird for me, because the docs for every version clearly say that:

> Also, the standard scaffold views expect model variables of the form <propertyName>InstanceList for collections and <propertyName>Instance for single instances. It’s tempting to use properties like 'books' and 'book', but those won’t work.

but we already know that those docs like lying

drodzewicz (approved):

awalkowiak (commented):
I am merging that one, but I am also baffled about these changes a bit.

awalkowiak:
ok I found answer https://github.com/grails/scaffolding/issues/32

awalkowiak:
ok I found answer https://github.com/grails/scaffolding/issues/32

PR #4250, opened 2023-08-28
OBGM-691 Unable to catch exception inside of the controller when importing non existing location zone

drodzewicz:
- refactor LocationController from transaction to regular and use services to handle transaction actions
- create LocationDataService for location CRUD operations

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
@@ -18,13 +18,14 @@ import org.pih.warehouse.requisition.Requisition
 import org.pih.warehouse.shipping.Shipment
 import org.springframework.web.multipart.MultipartFile
 
-@Transactional
 class LocationController {
 
     def inventoryService
     def locationService
     def dataService
     def organizationService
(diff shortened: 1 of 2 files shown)


kchelstowski (commented):
Could you describe/change the PR's description to understand what the problem was?

kchelstowski:
At the moment it's difficult to understand what the problem was, since I can see there was a transaction around those actions (transactional on the controller). Was it wrong error message or?

drodzewicz:
The issue was that when importing a none-existing location zone it would throw and Exception which we had to catch to display an appropriate error message to the user but since the exception was thrown in the transactional service and the method which was calling this service was also transaction it instead threw the LazyLocalization Exception.

Removing the `@Transaction` from the controller and replacing the Gorm domain crud methods with service methods fixed the issue.

There are few controller methods that I left with the `@Transactional` because I am pretty sure that they are not used anywhere but I am afraid of removing them completely at the moment.
I think we will have to dedicate a whole investigatory ticket to cleaning up some "dead code", especially in the non-API controllers

awalkowiak (approved):

PR #4253, opened 2023-08-28
OBGM-698 Fix product merge functionality

awalkowiak:


grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -34,6 +35,7 @@ class ProductMergeService {
      * Preforms product swapping for a product pairs passed as params
      * TODO: OBPIH-5484 Refactor this part and move each relation swapping to the separate function
      * */
+    @Transactional
```
kchelstowski, line 38: I'd consider adding it for whole service to be consistent throughout the services and also to avoid some future problems connected to the fact, that a method would not be `@Transactional`.
I know, that probably for now, all the methods below are called by the `mergeProduct` method, so we won't experience any at the moment.

awalkowiak: Exactly, as you pointed out everything is basically called from the `mergeProduct`, hence I don't see any need to wrap the entire service. Plus if someone will change or add anything new then he will have to do it either way.

PR #4257, opened 2023-08-30
OBGM-129 Add some logging in JobUtils

awalkowiak:


jmiranda (commented):
I'm not sure we want more logging on every execution. Can you provide a justification?

awalkowiak:
@jmiranda no bigger reason besides to see if something is disabled or not

PR #4265, opened 2023-09-05
OBGM-631 Add missing Left outer join in criteria queries which associations possbile null values

drodzewicz:


grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -1603,7 +1604,7 @@ class DocumentService {
                     documents {
```
kchelstowski, line 1604: are we sure this is not nullable?

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -113,12 +113,12 @@ class OrderService {
                     if (params.createdBy) {
                         eq("createdBy.id", params.createdBy)
                     }
-                    if (params.paymentTerm) {
-                        or {
-                            if (params.list("paymentTerm").contains("null")) {
-                                isNull("paymentTerm")
-                            }
-                            paymentTerm(JoinType.LEFT_OUTER_JOIN.joinTypeValue) {
+                    paymentTerm(JoinType.LEFT_OUTER_JOIN.joinTypeValue) {
```
kchelstowski, line 116: why has it been pulled out of the `if`?
I would say it is not necessary to join this table if we don't specify `params.paymentTerm`? Or am I missing something here?

grails-app/controllers/org/pih/warehouse/data/MigrationController.groovy
```diff
@@ -265,6 +265,7 @@ class MigrationController {
         }
         else {
             data = data.collect {
+                def href = g.createLink(controller: "stockMovement", action: "show", id: it?.id)
```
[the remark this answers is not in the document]
drodzewicz: it was missing and the dialog was not rendering because of it

grails-app/services/org/pih/warehouse/order/OrderService.groovy
drodzewicz: if you scroll down few llines and look at the **order(...)**  logic, you will see that we also support ordering by **productTerm**
```groovy
paymentTerm {
    order("name", params.order ?: 'asc')
}
```
if you dislike my approach of pulling out the association with a join then I will have to specify the joint type on the paymentTerm for ordering too like:
```groovy
paymentTerm(JoinType.LEFT_OUTER_JOIN.joinTypeValue) {
    order("name", params.order ?: 'asc')
}
```
what do you think? Should we have the join type defined in both places?

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
drodzewicz: I believe that in this situation INNER JOIN actually makes sense because we are querying **Order** table for documents and if documents does not exists on Order then we want to skip it. Do you disagree?

grails-app/services/org/pih/warehouse/order/OrderService.groovy
kchelstowski: This is kind of explanation I expected to get :laughing: 
I have not seen that closure at the bottom. If that's the case, then I'm completely fine with the change.

kchelstowski (approved):

awalkowiak (approved):

PR #4267, opened 2023-09-06
OBGM-673 Remove unnecessary main.css file from assets/stylesheets

drodzewicz:
assets/stylesheets/main.css was causing issue with loading a webapp/css/main.css

After investigating a styling issue with certain print pages I found the root of the problem which was a wrong `main.css` file resource link.

Currently on our OBGM branch we have two versions of `main.css`, one is in `webapp/css/main.css` and the other is in `assests/stylesheets/main.css`.
After looking the generated resource links for Layout/print 
```gsp
<link rel="stylesheet" href="${resource(dir:'css',file:'main.css')}" type="text/css" media="print, screen, projection" />
<link rel="stylesheet" href="${resource(dir:'css',file:'menu.css')}" type="text/css" media="print, screen, projection" />	
```
image
I noticed that the generated resource link for main.css is not the one that I was expecting.

After looking into the problem I found a [post on stack](https://stackoverflow.com/questions/36680396/adding-static-resources-css-js-htm-to-grails-3-application)  which mentioned
> Since Grails 2.4, static assets are managed by the [Asset Pipeline plugin](http://bertramdev.github.io/grails-asset-pipeline/).

based on the documentation of assets-pipeline which mentioned that all of the static files should be in **assets** folder I assume what is happening is that first it checks the assets folder for matching files (main.css) and if it doesn't find find anything it moves on to look for it in other places. Since we already had the `main.css` file in assets, it matched it and loaded that file.

The reason why I am removing the main.css from assets folder is, after inspection and a sanity check I think this file was moved here by mistake during the[ initial migration proccess](https://github.com/openboxes/openboxes/commit/0858c694ab1a3a838d9f06d1a156fea3c083405f). After looking for the origin file on `develop` branch I was only able to find it in the build and plugins directories
image

I will also create a separate Spike ticket where we will have to move all of the static files to the assets folder as the grails framework and assets-pipeline suggests


drodzewicz:
I have created a ticket for moving the static files to assets folder https://pihemr.atlassian.net/browse/OBPIH-5852

awalkowiak (approved):
I think this one was generated by grails, but is not required

PR #4269, opened 2023-09-07
OBGM-532 Add UI for switching root category on category tree page

drodzewicz:


drodzewicz:
As was discussed on the tech huddle, I am adding a UI change so a user has the ability to choose the root category to display its full tree
image
That way a user will be able to view all of the available category trees and modify/remove them if it is necessary.

grails-app/controllers/org/pih/warehouse/product/CategoryController.groovy
```diff
@@ -53,24 +56,29 @@ class CategoryController {
             categoryInstance.properties = params
         }
 
-        if (!categoryInstance.hasErrors() && categoryInstance.save()) {
+        try {
+            categoryGormService.save(categoryInstance)
             flash.message = "${warehouse.message(code: 'category.saved.message', arg: [format.category(category: categoryInstance).decodeHTML()])}"
             redirect(action: "tree", model: [rootCategory: productService.getRootCategory()])
-        } else {
+        } catch (Exception e) {
```
kchelstowski, line 63: was there a reason for this `try/catch` block, which we agreed we should avoid?
Can't we just refactor the if, not to use `categoryInstance.save()`, but `categoryGormService.save(categoryInstance)`?

Moreover - see the comment below

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -457,6 +457,10 @@ class ProductService {
         return category;
     }
 
+    List<Category> getCategoriesWithoutParent() {
+        return Category.findAllByParentCategoryIsNull().sort {a, b -> a.sortOrder <=> b.sortOrder }
```
kchelstowski, line 461:
```suggestion
        return Category.findAllByParentCategoryIsNull().sort { a, b -> a.sortOrder <=> b.sortOrder }
```
:see_no_evil:

grails-app/views/category/tree.gsp
```diff
@@ -11,16 +11,33 @@
             <g:if test="${flash.message}">
             	<div class="message">${flash.message}</div>
             </g:if>
-			<g:hasErrors bean="${categoryInstance}">
-				<div class="errors"><g:renderErrors bean="${categoryInstance}" as="list" /></div>
+            <g:if test="${flash.error}">
+                <div class="errors">${flash.error}</div>
+            </g:if>
+			<g:hasErrors>
```
kchelstowski, line 17: the indent feels weird here, doesn't it?

grails-app/controllers/org/pih/warehouse/product/CategoryController.groovy
```diff
@@ -93,11 +101,11 @@ class CategoryController {
...
     def save() {
         def categoryInstance = new Category(params)
-
-        if (categoryInstance.save(flush: true)) {
+        try {
+            categoryGormService.save(categoryInstance)
             flash.message = "${warehouse.message(code: 'default.created.message', args: [warehouse.message(code: 'category.label', default: 'Category'), categoryInstance.id])}"
             redirect(action: "tree", id: categoryInstance.id)
-        } else {
+        } catch(Exception e) {
```
kchelstowski, line 108: this is not correct. If you get any validation exceptions while saving, you will be redirected to the `tree` anyway and will see the success message, because `save()` in Grails doesn't throw a `ValidationException` when there are any, but returns `null`.

kchelstowski (changes requested):

grails-app/migrations/0.9.x/changelog-2023-09-07-1140-add-missing-is_root-value-to-root-catalog.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <changeSet author="drodzewicz" id="070920231140-0">
+    <preConditions onFail="MARK_RAN">
+        <sqlCheck expectedResult="1">
```
awalkowiak, line 8: I think it would be better if this checked for `WHERE id = 'ROOT' and is_root is NULL` and the expected result being 0.

awalkowiak (changes requested):
One minor thing in migration, rest LGTM (other than what Kacper already requested)

grails-app/views/category/tree.gsp
drodzewicz: It does, but I don't know what's wrong :man_shrugging: 
I used tabs in both cases and it looks good in the edditor

grails-app/controllers/org/pih/warehouse/product/CategoryController.groovy
drodzewicz: Not exactly, using save method provided by `@Service` will throw an exception if it encounters a validation error unlike using `Domain.save()` which will indeed return a null on validation error.

As for using try/catch I believe we talked about letting the generic error handler handle the exception but in the case of  rendering page templates we don't have a generic error handler that can do it and we probably won't have.
I did an investigation into implementing this handler #4240 and unfortunately, I wasn't able to find a proper way to do it.

grails-app/migrations/0.9.x/changelog-2023-09-07-1140-add-missing-is_root-value-to-root-catalog.xml
drodzewicz: I added an additional WHERE condition.
As for the expected result 0 I disagree.
The precondition is set to `onFail="MARK_RAN"` which will skip this changelog if category id = ROOT and is_root = NULL won't be found, which is what I expect.

awalkowiak (approved):

PR #4270, opened 2023-09-08
OBGM-501: Improve performance of products search API

kchelstowski:
In the ticket you will find a table with timings with/without improvement. Basically a thing that improved the performance the most, was not to return whole product objects from `searchProducts` method, but only the ids of products, and refactor the `where` clause respectively from:
```groovy
and pa.product in (:products)
```
to 
```groovy
and pa.product.id in (:products)
```
I wanted to make sure that I don't cause any regression, so I didn't decide to deep refactor the current methods, but to add some additional, helper flags, like `returnsIds`, `excludeNegativeQuantity`

Speaking about `excludeNegativeQuantity`, it was second part of improvement - instead of filtering the list like this:
```groovy
availableItems = availableItems.findAll { it.quantityOnHand > 0 }
```
I filter those at the query level when it's needed.

If we wanted to improve the performance even more, we could consider (if we don't need it) removing the `sort` from synonyms mapping in the product domain: https://github.com/openboxes/openboxes/blob/a18cecb090c97a07a10cc1b0f600730b79a9ac8f/grails-app/domain/org/pih/warehouse/product/Product.groovy#L283
I explained in the ticket, why the sorting causes an error while trying to return distinct.

If we did so, I could return the **distinct** product ids from `searchProducts` method, so to redo it from:
```groovy
 if (returnIds) {
  projections {
    property("id")
  }
 }
```
to 
```groovy
 if (returnIds) {
  projections {
    distinct("id")
  }
 }
```
for now I assumed that it was added there on purpose, so I didn't decide to change it, but I'll include the timings with that change as well.


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -650,13 +650,14 @@ class ProductAvailabilityService {
         return data
     }
 
-    List getAvailableItems(Location location, List<Product> products) {
+    List getAvailableItems(Location location, List products, boolean excludeNegativeQuantity = false) {
```
kchelstowski, line 653: I tried to do a polymorphic method like:
```groovy
List getAvailableItems(Location location, List<String> products)
```
and leave the one:
```groovy
List getAvailableItems(Location location, List<Product> products)
```
but unfortunately it was shouting that there is already such a method, so it doesn't understand that `List<Product>` and `List<String>` is not the same.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -650,13 +650,14 @@ class ProductAvailabilityService {
...
+        // Flag to check if products is a list of products objects or list of products ids
+        boolean containsProductsIds = ListUtil.isTypeOfString(products)
```
kchelstowski, line 655: I tried to do a polymorphic method like:
```groovy
List getAvailableItems(Location location, List<String> products)
```
and leave the one:
```groovy
List getAvailableItems(Location location, List<Product> products)
```
but unfortunately it was shouting that there is already such a method, so it doesn't understand that `List<Product>` and `List<String>` is not the same.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -665,22 +666,22 @@ class ProductAvailabilityService {
 						left outer join pa.inventoryItem ii
 						left outer join pa.binLocation bl
 						where pa.location = :location
-						and pa.product in (:products)
-						""", [location: location, products: products])
-            def status = { quantity -> quantity > 0 ? "inStock" : "outOfStock" }
```
kchelstowski, line 670: I removed this at all, because it has not been used anywhere: we were doing this:
```groovy
 [
    status                      : status(quantity),
    product                     : it[0],
    inventoryItem               : inventoryItem,
    binLocation                 : binLocation,
    quantityOnHand              : quantity,
    quantityAvailableToPromise  : it[4]
]
and then immedietaly mapping this to `new AvailableItem(...)`
So I also removed the `product` property, because of the same reason

**Review by awalkowiak — COMMENTED** [review]
The idea looks good and promising, but I am not sure if this is a good implementation of it (this hacky way of overloading). @jmiranda could you also take a look at this one (and my comment)?

File: grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy [hunk 4]
```diff
@@ -939,13 +941,15 @@ class StockMovementService {
...
-        def productsMap = products.inject([:]) { map, item -> map << [(item.id): item] }
+        List<Product> products = Product.findAllByIdInList(data.collect { it.product_id })
+        Map productsMap = products.inject([:]) { map, item -> map << [(item.id): item] }
 
         Requisition requisition = Requisition.get(data.first()?.requisition_id)
         def picklistItemsMap = requisition?.picklist?.pickablePicklistItemsByProductId
 
-        def availableItemsMap = productAvailabilityService.getAllAvailableBinLocations(requisition.origin, products).groupBy { it?.inventoryItem?.product?.id }
+        Map<String, List<AvailableItem>> availableItemsMap = productAvailabilityService
+                .getAllAvailableBinLocations(requisition.origin, productsMap.keySet().stream().collect(Collectors.toList()))
```
awalkowiak, line 951: This was my initial suggestion. Later I suggested moving `data.collect { it.product_id }` from line 944 to separate variable and use it in `findAll...` and here.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1490,12 +1494,12 @@ class StockMovementService {
                     def picklistItems = requisitionItem.substitutionItems.findAll { it.product == associatedProduct }
                             .collect { it.picklistItems }?.flatten()
 
-                    availableItems = productAvailabilityService.getAllAvailableBinLocations(location, associatedProduct)
+                    availableItems = productAvailabilityService.getAllAvailableBinLocations(location, associatedProduct?.id)
                     availableItems = availableItems.findAll { it.quantityOnHand > 0 }
```
awalkowiak, line 1498: @kchelstowski one more thing. I think you can also use here the same logic with `excludeNegativeQuantity` parm and pass it down to the `getAvailableItems` (since it is false by default)

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -650,13 +648,12 @@ class ProductAvailabilityService {
         return data
     }
 
-    List getAvailableItems(Location location, List<Product> products) {
-        log.info("getQuantityOnHandByBinLocation: location=${location} product=${products}")
-        def data = []
+    List getAvailableItems(Location location, List<String> productsIds, boolean excludeNegativeQuantity = false) {
```
awalkowiak, line 651: @kchelstowski should it be `true` by defualt? Looks like we are very often doing `availableItems.findAll { it.quantityOnHand > 0 }`

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: What should be enough here is to just change the method from `getAllAvailableBinLocations` to `getAvailableBinLocations`, since the only different between them is that flag, that makes this filtering:
```groovy
List<AvailableItem> getAvailableBinLocations(Location location, List<String> productsIds) {
        List<AvailableItem> availableItems = getAvailableItems(location, productsIds, true)

        availableItems = sortAvailableItems(availableItems)

        return availableItems
}

// Include also bin locations with negative qty (needed for edit page items)
List<AvailableItem> getAllAvailableBinLocations(Location location, List products) {
  List<AvailableItem> availableBinLocations = getAvailableItems(location, products)
  return availableBinLocations
}
```
so I'd say we should not have used the `getAll` method here if we filter the results below.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
kchelstowski: The only two places where this method is used are: `getAllAvailableBinLocations` and `getAvailableBinLocations` mentioned above. 
So I'd say we wouldn't achieve anything by changing that to `true`.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: fixed

awalkowiak (approved):

PR #4271, opened 2023-09-11
OBGM-687 price and package issue in product source import

drodzewicz:
This is a rather ugly bug and I think we should investigate it a little more, but first, let me explain what is going on, because there are many connected things that cause this issue. 
_Sorry in advance for my complicated explanation_

First I need to give some context into domain structure and create product source logic, so Let me highlight the relationship and constraint on certain fields in this hierarchy tree.

On **ProductSupplier** we have a 1:many relationship with a **ProductPackage**, and **ProductPackage** has a `nullable: true` **ProductPrice** which has a field **price** that is `nullable: false`.

When creating a product source using import data, without specifying the package price in the column, we are expecting it to throw a validation exception, because we are assigning `null` to **ProductPrice** which has **price** as a required field.

https://github.com/openboxes/openboxes/blob/c283d367017dc6886d1ba30fa54fd8fdf7217aa3/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy#L195-L198

Calling a `productPrice.validate()` does return `false` and `productPrice.errors` does contain the validation error that we expect, but when we are at the point of saving the `ProductSupplier`
https://github.com/openboxes/openboxes/blob/c283d367017dc6886d1ba30fa54fd8fdf7217aa3/grails-app/services/org/pih/warehouse/data/ProductSupplierDataService.groovy#L146-L149
an exception is not being thrown, instead It creates the ProductSupllier with ProductPackage that has assigned **product_price_id** for non existing **ProductPrice**, because during save, **ProductPrice** was not saved due to validation error.
So it creates a problem where ProductPackage has an orphan product_price_id which does not correspond to any record in ProductPrice table since it doesn't exist.

So when GORM was trying to fetch `ProductPackage.productPrice` it caused an error since it can't populate this field with a given id, because no record with such an id exists. 

To prevent this issue from happening I changed the default `null` to `0` on ProductPrice during import which is probably what should have been there in the first place, since the same logic is applied for ProductPackage creation on `ProductController > savePackage`
https://github.com/openboxes/openboxes/blob/2d0fdc948f936357d26476b10d6fec47e6f8a0bf/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy#L379-L383 

and I also made a DB migration to remove the orphan foreign keys on ProductPackage table.

One of the problems seems to be when saving nested data, it does not "bubble up the exceptions" which should prevent the record from being created. I have noticed this previously but did not give it much attention since previously all I had to do was call the`.validate()` on that nested object.
I tried to investigate it, and looked into some configuration that we might have missed but was unsuccessful in finding a solution.
My first thought was to add `save(deepValidate: true)`  which in theory should be true by default, but it did not solve the issue.

### DIFF
--- a/grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml
+++ b/grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <changeSet author="drodzewicz" id="080920231340-0">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <sqlCheck expectedResult="0">
+          SELECT COUNT(*) FROM product_package ppkg
+          LEFT JOIN product_price pp
+          ON ppkg.product_price_id=pp.id
+          WHERE ppkg.product_price_id IS NOT NULL AND pp.id IS NULL;
+        </sqlCheck>
+      </not>
+    </preConditions>
+    <sql>
+      UPDATE product_package pkg
+      SET pkg.product_price_id = NULL
+      WHERE pkg.id IN (SELECT * FROM (
+      SELECT ppkg.id FROM product_package ppkg
+      LEFT JOIN product_price pp
+      ON ppkg.product_price_id=pp.id
+      WHERE ppkg.product_price_id IS NOT NULL AND pp.id IS NULL
+      ) as ppkg_ids);
+    </sql>
+  </changeSet>
+</databaseChangeLog>
--- a/grails-app/migrations/0.9.x/changelog.xml
+++ b/grails-app/migrations/0.9.x/changelog.xml
@@ -5,4 +5,5 @@
 
     <include file="0.9.x/changelog-2023-02-21-1340-alter-table-order-change-date-ordered-column-type.xml" />
     <include file="0.9.x/changelog-2023-09-07-1140-add-missing-is_root-value-to-root-catalog.xml" />
+    <include file="0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml" />
 </databaseChangeLog>
--- a/grails-app/migrations/0.9.x/changelog.xml
+++ b/grails-app/migrations/0.9.x/changelog.xml
@@ -5,5 +5,5 @@
 
     <include file="0.9.x/changelog-2023-02-21-1340-alter-table-order-change-date-ordered-column-type.xml" />
     <include file="0.9.x/changelog-2023-09-07-1140-add-missing-is_root-value-to-root-catalog.xml" />
-    <include file="0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml" />
+    <include file="0.9.x/changelog-2023-09-08-1340-remove-product-price-fk-on-product-package-pointing-to-non-existing-record.xml" />
 </databaseChangeLog>
--- a/grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-product-price-fk-on-product-package-pointing-to-non-existing-record.xml
+++ b/grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-product-price-fk-on-product-package-pointing-to-non-existing-record.xml
@@ -14,6 +14,13 @@
         </sqlCheck>
       </not>
     </preConditions>
+    <comment>
+      This migration is meant to fix a bug found on OBGM-687
+      which caused an issue when importing product sources.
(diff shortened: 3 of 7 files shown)


kchelstowski (commented):
A reason why the deepValidate did not work for you could be because of this `cascadeValidate` setting:
```groovy
grails.gorm.default.mapping = {
    '*'(cascadeValidate: 'none')
    id generator: 'uuid'
}
```

drodzewicz:
You are right, thanks @kchelstowski, I missed this file `application.groovy` :see_no_evil: 

I am not sure as for the reason why this config was added, there seems to be a whole ticket dedicated to this but no explanation. https://pihemr.atlassian.net/browse/OBGM-216
Maybe we should turn it back ON globally and do `deepValidate: false` for places where we don't want it?
However I am not sure if it will work since calling `deepValidate: true` when this config was defined did not override it.

P.S.
I wonder if we can move config from `application.groovy` to `application.yml` so the configuration is not scattered through different files

awalkowiak:
@drodzewicz if it is build config, then it should be in the application.yml, if it is runtime or depends on application classes it should be in the runtime.groovy

grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml
```diff
@@ -0,0 +1,28 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <changeSet author="drodzewicz" id="080920231340-0">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <sqlCheck expectedResult="0">
+          SELECT COUNT(*) FROM product_package ppkg
+          LEFT JOIN product_price pp
+          ON ppkg.product_price_id=pp.id
```
awalkowiak, line 12: These two lines make no sense to me `ON ppkg.product_price_id=pp.id` and `WHERE ppkg.product_price_id IS NOT NULL AND pp.id IS NULL` it looks like they are mutually excluding

grails-app/migrations/0.9.x/changelog.xml
```diff
@@ -5,4 +5,5 @@
 
     <include file="0.9.x/changelog-2023-02-21-1340-alter-table-order-change-date-ordered-column-type.xml" />
     <include file="0.9.x/changelog-2023-09-07-1140-add-missing-is_root-value-to-root-catalog.xml" />
+    <include file="0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml" />
```
awalkowiak, line 8: The name of this file is a bit misleading.

grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml
```diff
@@ -0,0 +1,28 @@
...
+          WHERE ppkg.product_price_id IS NOT NULL AND pp.id IS NULL;
+        </sqlCheck>
+      </not>
+    </preConditions>
+    <sql>
+      UPDATE product_package pkg
```
awalkowiak, line 18: Could you add a comment here (in code) on what issue is this migration addressing?

awalkowiak (changes requested):
I think it should not create a ProductPrice if the price is null instead of having 0 as a default.

drodzewicz:
> I think it should not create a ProductPrice if the price is null instead of having 0 as a default.

@awalkowiak 
Sure, I can fix that.
I was just following the solution that we have implemented on create product package
https://github.com/openboxes/openboxes/blob/2d0fdc948f936357d26476b10d6fec47e6f8a0bf/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy#L379-L383

kchelstowski:
@drodzewicz 

> However I am not sure if it will work since calling deepValidate: true when this config was defined did not override it.

I believe if you disable this setting globally like we do, then you can't "override" it by saying `deepValidate: true`, so either we remove this setting globally and as you said - mark `deepValidate: false` wherever we don't want it, or I don't know :man_shrugging: 

But I'd ask - why wouldn't we want to `deepValidate` everything? :thinking:

grails-app/migrations/0.9.x/changelog-2023-09-08-1340-remove-missing-price-fk-on-product-packages.xml
drodzewicz: Yes, I understand your confusion, but this is exactly the bug the lack of deep validation caused.
Let me explain it with an example

When creating a ProductSupplier, even though there was a validation error on ProductPrice, it still created ProductSupplier and ProductPackage. It also created a foreign key on ProductPackage pointing to ProductPrice, but did not persist that ProductPrice to the database.
So the FK was pointing to nothing, to a non existing record.

so executing this SQL statement
```SQL
SELECT ppkg.id, ppkg.product_price_id, pp.id FROM product_package ppkg
LEFT JOIN product_price pp
ON ppkg.product_price_id=pp.id
WHERE ppkg.product_price_id IS NOT NULL AND pp.id IS NULL;
```

will return such a result
image

grails-app/migrations/0.9.x/changelog.xml
drodzewicz: what is your suggestion?
Does `remove-product-price-fk-on-product-package-pointing-to-non-existing-record` sound better?

awalkowiak (approved):

PR #4272, opened 2023-09-11
OBGM-695 Improve content of error message when try to delete bin with inventory

alannadolny:


grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -242,7 +242,12 @@ class LocationApiController extends BaseDomainApiController {
         if (existingLocation.isZoneLocation() && Location.findAllByZone(existingLocation)) {
             throw new IllegalArgumentException("You cannot delete zone that is assigned to a bin location ${params.id}")
         }
-        locationService.deleteLocation(existingLocation)
+
+        try {
+            locationService.deleteLocation(existingLocation)
+        } catch (Exception e) {
```
drodzewicz, line 248: It would be best to avoid catching exceptions in API controllers and let generic error handlers deal with the exception.

My first suggestion was to maybe find the code for this exception and add it to `messages.properties` file, but I think this only works for validation errors and this is a database constraint violation which probably doesn't have an error code.

Could you look into how we can handle these kinds of errors generically in the ErrorsControler if it is possible, and if not I think try/catch here should be fine, but please change the Exception you are catching to `org.springframework.dao.DataIntegrityViolationException e` so that we let through other exceptions and let the the ErrorsController handle them, unless you dissagree.

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
```diff
@@ -242,7 +242,12 @@ class LocationApiController extends BaseDomainApiController {
         if (existingLocation.isZoneLocation() && Location.findAllByZone(existingLocation)) {
             throw new IllegalArgumentException("You cannot delete zone that is assigned to a bin location ${params.id}")
         }
-        locationService.deleteLocation(existingLocation)
+
+        try {
+            locationService.deleteLocation(existingLocation)
+        } catch (org.springframework.dao.DataIntegrityViolationException e) {
+            throw new Exception("${warehouse.message(code: 'default.not.deleted.with.reason.message', args: [warehouse.message(code: 'location.label', default: 'Location'), existingLocation.id, e.message])}")
+        }
```
kchelstowski, line 250: I know the reason you've done it, but I also don't feel comfortable with that change - we should in my opinion come up with better error handling (in the `ErrorsController`), because just throwing everywhere `try/catch` feels to me like a "lazy solution", that doesn't solve the real problem.

Could you please debug the error properties inside the `ErrorsController` and paste a screenshot here, so that maybe we could come up with some solution?
Dariusz's suggestion with getting to know the code of the exception would be perfect, but I don't remember if it's possible.

Maybe a good expirement would be to use the data services' delete method in the `deleteLocation` method instead of the domain one? How would the error look like then? (Both in the UI and caught in the `ErrorsController`)?

kchelstowski (commented):
If we know the name of the constraint that is throwing this exception, I know we could somehow build a handler for this exception. We'd have to define it in the `messages.properties` like:
```
unq_product_name=Product name must be unique
```
and then in handler create maybe some kind of a Map, that would contain those like:
```groovy

{ 
  "unq_product_name": ERROR_MESSAGE_KEY
  ...
}
```
and then in the handler, catch the constraint and look if a message for that exists in our Map.

I know this might sound complicated and most probably not for this ticket, but I'm just throwing maybe a possible solution for the future.

grails-app/controllers/org/pih/warehouse/api/LocationApiController.groovy
alannadolny: > Could you please debug the error properties inside the ErrorsController and paste a screenshot here, so that maybe we could come up with some solution?

For me it doesn't make sense, I remember a situation in which I based on error properties (handling validation exceptions) and it worked only locally, it broke handling those errors on obdev3. But I can paste a screenshot for you 😀

image


> Maybe a good expirement would be to use the data services' delete method in the deleteLocation method instead of the domain one? How would the error look like then? (Both in the UI and caught in the ErrorsController)?

The main reason for my change was to make the error more meaningful for users, even if it will be an error from data service it still wouldn't be something as I have at this moment:
 `Location ff8081818a83f2ac018a83f6e2bf0003 could not be deleted - could not execute statement; SQL [n/a]; constraint [null]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement`

awalkowiak (approved):

PR #4273, opened 2023-09-11
OBGM-467 Fix invoking crowdin script on react pages

kchelstowski:
In the ticket you will find a link to a video where I show what the problem is.


webpack.config.js
```diff
@@ -69,7 +69,15 @@ module.exports = {
           contextPath: '\${util.ConfigHelper.contextPath}',
           jsSource: `\${resource(dir: '${path.basename(WEBPACK_OUTPUT)}', file: 'bundle.${compilation.hash}.js')}`,
           cssSource: `\${resource(dir: '${path.basename(WEBPACK_OUTPUT)}', file: 'bundle.${compilation.hash}.css')}`,
-          receivingIfStatement: '',
+          receivingIfStatement:
```
awalkowiak, line 72: This should not be injected like that. If you'd like to inject it through `templateParameters` it should have its own parameter instead of using the of related to receiving. However, you correctly found the reason behind it, but missed one key part in your video. The issue is here:

Develop's `grails-template.html`::
https://github.com/openboxes/openboxes/blob/develop/src/assets/grails-template.html#L4

OBGM feature branch's `grails-template.html`::
https://github.com/openboxes/openboxes/blob/feature/upgrade-to-grails-3.3.10/src/assets/grails-template.html#L4

Develop's `main.gsp`:
https://github.com/openboxes/openboxes/blob/develop/grails-app/views/layouts/main.gsp

Feature branch's `main.gsp`, and `react.gsp`:
https://github.com/openboxes/openboxes/blob/feature/upgrade-to-grails-3.3.10/grails-app/views/layouts/main.gsp
https://github.com/openboxes/openboxes/blob/feature/upgrade-to-grails-3.3.10/grails-app/views/layouts/react.gsp

So it's better to fix the layout used in the `grails-template.html` than additionally injecting it by webpack.

awalkowiak (changes requested):

webpack.config.js
awalkowiak: @kchelstowski looks like react pages don't load layout properly

awalkowiak: @kchelstowski if you did not get any progress with layout rendering, add a new template param, that will inject into header entire content of the header from `main.gsp` and please create a ticket to investigate the layouts rendering on react.

grails-app/controllers/org/pih/warehouse/putaway/PutAwayController.groovy
```diff
@@ -23,7 +23,7 @@ class PutAwayController {
     // This template is generated by webpack during application start
     def create() {
         def contextPath = grailsApplication.config.server.contextPath
-        render(template: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
+        render(view: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
```
awalkowiak, line 26: I think you can safely remove the `model` part here since the context part is taken from util inside the template.

grails-app/controllers/org/pih/warehouse/stocklist/StocklistManagementController.groovy
```diff
@@ -4,6 +4,6 @@ class StocklistManagementController {
     def index() {
         // This template is generated by webpack during application start
         def contextPath = grailsApplication.config.server.contextPath
-        render(template: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
+        render(view: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
```
awalkowiak, line 7: Same as above

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -117,12 +117,12 @@ class DashboardController {
             return
         }
         def contextPath = grailsApplication.config.server.contextPath
-        render(template: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
+        render(view: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
```
awalkowiak, line 120: and here

grails-app/controllers/org/pih/warehouse/user/DashboardController.groovy
```diff
@@ -117,12 +117,12 @@ class DashboardController {
...
     }
 
     def supplier() {
         def contextPath = grailsApplication.config.server.contextPath
-        render(template: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
+        render(view: "/common/react", model: [contextPath: contextPath != '/' ? contextPath : ''])
```
awalkowiak, line 125: and here

grails-app/views/layouts/react.gsp
```diff
@@ -3,6 +3,13 @@
 <head>
     <title><g:layoutTitle default="OpenBoxes" /></title>
     <link rel="stylesheet" href="//netdna.bootstrapcdn.com/font-awesome/3.2.1/css/font-awesome.min.css">
+    <g:if test="${session.useDebugLocale}">
```
awalkowiak, line 6: Please either use `main.gsp` or add **all** header scripts. It is missing `hotjar`, `googleSiteTag` and `NewRelic`

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/putaway/PutAwayController.groovy
kchelstowski: removed. Working fine

grails-app/views/layouts/react.gsp
kchelstowski: I've decided to migrate those from `main.gsp` to `react.gsp` layout, since our `main.gsp` is different in Grails 3 and in Grails 1, so if I wanted to use the `main.gsp`, the page looked like this:
Screenshot from 2023-09-12 17-23-29

PR #4274, opened 2023-09-12
OBGM-454 Unable to preview all background jobs 

alannadolny:


drodzewicz (approved):
I remember there was something similar you investigated a while ago in #4221 
So this looks like something that has changed in grails 3
LGMT

awalkowiak (approved):

PR #4275, opened 2023-09-12
OBGM-569 Replace PagedResultList temprorary utils with PaginatedList

drodzewicz:
Because the implementation of `PagedResultList` class has changed since Grails version 2.x^, we were unable to use this class in our modified data lists which is why the implementation was copied from Grails 1 and manually moved to Grails 3.

Instead of completely removing this utility, which is quite handy in some situations I decided to implement a much simpler List Class with the help of `@Delegate`.

An alternative solution without this util class would be to return all of the data for example in a `Map` like 
```return [data: stockMovements, totalCount: shipments.totalCount]``` and so I think having this utils is a much cleaner solution.

**PagedResultList** in grails 1 was basically a `List` with an additional `totalCount` property, so with the help of `@Delegate` which basically implements all of the methods of the annotated property for us (in our case it is **List**), making the implementation file much smaller and more maintainable I created an alternative util class `PaginatedList`.

### CODE REVIEW

**Review by awalkowiak — APPROVED** [review]
LGTM, but I'd like to see others opinion

**Comment by drodzewicz** [comment]
I have two other potential solutions

##  Map
This was my initial solution and was the most basic one.
Since in the end, we are still returning a Map of `[data, totalCount]` in the controller, we can simply return the same map from the service like `return [data.stockMovements, totalCount: shipments.totalCount]`
https://github.com/openboxes/openboxes/blob/5e09b22ff0d03b3f0821f921d0e5d00875c16102/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy#L78-L87
But then we have to be aware of explicitly passing `stockMovements.data` in places where we want to do something with that data like generate a csv or a pdf.

## Tuple and destructing
Instead of returning a Map and accessing this data by property name, we can instead return a 
```groovy
return new Tuple(stockMovements, shipments.totalCount)
```
and in the controller, we can destruct this variable into two separate variables like
```groovy
def (stockMovements, totalCount) = stockMovementService.getStockMovements(stockMovement, params) 
```

## Util List Class
In the end, I liked the util class a bit more since we didn't have to deal with "preparing" the list of data like destructing it or accessing it by property name, instead, we can implicitly treat this data structure as a `List` and if we need a **totalCount** we can access it by property name.

@awalkowiak

alannadolny (approved):

grails-app/utils/org/pih/warehouse/PaginatedList.groovy
```diff
@@ -0,0 +1,17 @@
+package org.pih.warehouse
+
+class PaginatedList {
```
kchelstowski, line 3: overall this looks good to me, however I'm wondering if we can make this "type safe", so to do it like:
```groovy
class PaginatedList<T> {
   @Delegate
   List<T> list;
   
   int totalCount;
}
```
and use it like:
```groovy
return new PaginatedList<StockMovement>(stockMovements, requisitions.totalCount)
```

if it's possible, then I think it'd be a nice addition, if not, then LGTM

drodzewicz: sure, sounds good

PR #4277, opened 2023-09-13
OBPIH-5768 Add request approval supported activity to the Location do…

kchelstowski:
…main

Either can be merged and for now use ROLE_APPOVER for user to check if it works fine, or do not merge and wait for the OBPIH-5799 to replace the role with REQUEST_APPROVER or whatever it will be.


grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -339,6 +346,13 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return active ? LocationStatus.ENABLED : LocationStatus.DISABLED
     }
 
+    List<User> getUsersWithRole(RoleType roleType) {
+        withNewSession {
+            List<Role> roles = Role.findAllByRoleType(roleType)
```
kchelstowski, line 351: decided to use `findAll` to be "safe", since hypothetically a user can create a few roles with the roleType of REQUEST_APPROVER

grails-app/i18n/messages.properties
```diff
@@ -1456,6 +1457,7 @@ location.bgColor.label=Background color
 location.fgColor.label=Foreground color
 location.bgColor.invalid.matchingcolor=Colors cannot be the same
 location.supportedActivities.invalid.supportedActivities=You cannot set supported activity NONE in combination with other supported activities
+location.supportedActivities.noRequestApprovers=You must add one or several Approvers to the location to support the requests approval workflow. Go to users list to add an Approver.
```
kchelstowski, line 1460: unfortunately it was not possible to make the link to users list work fine here, since we render the errors on location edit page like that:
```gsp
<g:hasErrors bean="${locationInstance}">
        <div class="errors">
            <g:renderErrors bean="${locationInstance}" as="list"/>
        </div>
</g:hasErrors>
```
When added a href tag to the message, it was reading it as a string:
Screenshot from 2023-09-13 09-34-56

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -98,6 +98,13 @@ class Location implements Comparable<Location>, java.io.Serializable {
             if (activities?.contains(ActivityCode.NONE.id) && activities?.size() > 1) {
                 return ['invalid.supportedActivities']
             }
+            // Don't allow to set REQUEST_APPROVAL activity if there are not any users that
+            // have Request Approver location role for this location
+            // TODO: Replace ROLE_APPROVER with the ROLE_REQUEST_APPROVER when OBPIH-5799 is done
+            if (activities?.contains(ActivityCode.REQUEST_APPROVAL.id)
```
awalkowiak, line 104: Was it meant to be a validator? I thought it was a message to inform the user that he has to do this **after** adding this supported activity to the location

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -64,7 +64,9 @@ enum ActivityCode {
 
     AUTOSAVE('AUTOSAVE'),
 
-    NONE('NONE')
+    NONE('NONE'),
```
awalkowiak, line 67: Can you put it at the end?

grails-app/domain/org/pih/warehouse/core/Location.groovy
kchelstowski: I apologize, I was confused because of many different versions that we were speaking about, but I've consulted with Katarzyna and she confirmed this could be just a message, not a validation.
I fixed that part.

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
kchelstowski: moved

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -112,7 +112,11 @@ class LocationController {
                     if (locationInstance?.id == session?.warehouse?.id) {
                         session.warehouse = locationInstance
                     }
-                    flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id])}"
+                    flash.message = []
+                    flash.message.add([code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id]])
```
alannadolny, line 116: Can it be done without initializing empty array? I mean merging those two lines into one:
```
flash.message = [[code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id]]]
```

grails-app/views/location/edit.gsp
```diff
@@ -17,10 +17,7 @@
 
 <body>
 <div class="body">
-
-    <g:if test="${flash.message}">
-        <div class="message">${flash.message}</div>
-    </g:if>
+    <g:renderMessage />
```
alannadolny, line 20: I like this solution since rendering messages with HTML tags doesn't work well in Grails 1. The first situation that came to my mind was the message after logging in with incorrect credentials. (This message works on Grails 3 - the same code, without any changes) 👍👍

alannadolny (approved):

grails-app/taglib/org/pih/warehouse/RenderMessageTagLib.groovy
```diff
@@ -0,0 +1,35 @@
...
+package org.pih.warehouse
+
+/**
+ * available attributes:
+ * * style - style of a message container ex: message, error (default: message)
+ * * message - a single or a list of messages to be rendered in the template
+ * by default takes the value stored in flash.message.
+ * accepts data like: String, List<String>, Map<code: string, args: obj>, List<Map<code: string, args: obj>>
+ */
+class RenderMessageTagLib {
```
awalkowiak, line 19: @drodzewicz I am not sure if this is a good place for this tag lib (if there is need for a new tag lib). I wonder if this rather should be in the `MessageTagLib`.

grails-app/views/location/edit.gsp
```diff
@@ -17,10 +17,7 @@
 
 <body>
 <div class="body">
-
-    <g:if test="${flash.message}">
-        <div class="message">${flash.message}</div>
-    </g:if>
+    <warehouse:renderMessage />
```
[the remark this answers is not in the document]
jmiranda: Using a join(".") in the controller might allow us to avoid this.

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -64,6 +64,8 @@ enum ActivityCode {
 
     AUTOSAVE('AUTOSAVE'),
 
+    REQUEST_APPROVAL('REQUEST_APPROVAL'),
```
jmiranda, line 67: Why can't we use APPROVE_REQUEST above?

jmiranda (changes requested):

grails-app/views/location/edit.gsp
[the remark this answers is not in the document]
kchelstowski: > Can you explain the reason for this?

The reason for that was to be able to include a link inside the message. This has nothing to do with the dot.

src/groovy/org/pih/warehouse/core/ActivityCode.groovy
kchelstowski: Honestly I have not seen this before. I checked it and it seems to be existing, but not having any effect.
It's also displayed as `Approve requisition` - I saw your comments in @alannadolny 's ticket and I am a bit confused. In tickets it was supposed to be named with "Request" and you suggest that it should be "Requisition" instead of "Request".
So would it mean this activity code should be also renamed from `APPROVE_REQUEST` to `APPROVE_REQUISITION`?

I think we should not mix "Requisition" and "Request", so the question is, what is the path we want to follow?

awalkowiak (approved):

awalkowiak (changes requested):
@drodzewicz There are still some pending requests (If everything is fixed, @kchelstowski you have green light to merge this one)

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -112,7 +112,12 @@ class LocationController {
                     if (locationInstance?.id == session?.warehouse?.id) {
                         session.warehouse = locationInstance
                     }
-                    flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id])}"
+                    flash.message = [
+                        [code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id]]
+                    ]
+                    if (locationInstance.supportedActivities.contains(ActivityCode.REQUEST_APPROVAL.id)) {
+                        flash.message.add([code: 'location.supportedActivities.noRequestApprovers', args: [g.createLink(controller: "user", action: "list")]])
+                    }
```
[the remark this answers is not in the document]
drodzewicz: The comma isn't necessary anymore since the message is rendered in the new line like
image

Previously @kchelstowski must have added the comma to separate both messages since they were rendered in the same line.

grails-app/views/location/edit.gsp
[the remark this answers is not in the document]
drodzewicz: The reason behind creating this taglib util function for rendering messages is because when translating labels in the controller and passing them using `flash.messages` it would either escape the HTML or render the HTML as a string.

So as a solution we want to pass to the `flash.message` all of the arguments needed for rendering the translated label and let the `g.message` handle the rendering.

I implemented this `<warehouse:renderMessage />` taglib function in a way so it would be possible to render either a single message or a list of messages.
We can also both pass a normal string or as I mentioned before all of the arguments for rendering message.properties labels like **code**, **args**, **default**

PR #4282, opened 2023-09-15
OBPIH-5820 Email alert to approver(s) - Approvals email alerts

alannadolny:
I am not calling this function in this PR, because in the draft created by Justin during TH, we decided to call it on an event. So it should be placed like that (code from notes from TH):
```
class RequisitionStatusTransitionEventService implements EventListener<RequisitionStatusTransitionEvent>
    def onApplicationEvent (RequisitionStatusTransitionEvent event) {
        triggerRequisitionTransitions(event)
        publishWebhookEvent(event)
        publishMessageQueueEvent(event) // FHIR event for a specific thingy thing
        publishDefaultEmailNotifications(event)
    }
}
```
I created here `RequisitionStatusTransitionEventService` just for the `publishDefaultEmailNotifications` method. The rest of the code (including event) is a part of another ticket.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
@@ -21,9 +21,9 @@ class RequisitionStatusTransitionEventService {
 
 
     void publishDefaultEmailNotifications(Requisition requisition, Collection receivers) {
-        String redirectToRequestsList = "/openboxes/stockMovement/list?direction=OUTBOUND&sourceType=ELECTRONIC"
-        String subject = "${requisition.requestNumber} ${requisition.name}"
         def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+        String subject = "${requisition.requestNumber} ${requisition.name}"
+        String redirectToRequestsList = "${g.createLink(uri: "/stockMovement/list?direction=OUTBOUND&sourceType=ELECTRONIC", absolute: true)}"
         GString body = "${g.render(template: "/email/approvalsAlert", model: [requisition: requisition, redirectUrl: redirectToRequestsList])}"
 
         mailService.sendHtmlMail(subject, body, receivers)
--- a/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
@@ -11,6 +11,7 @@ package org.pih.warehouse.inventory
 
 import org.codehaus.groovy.grails.commons.GrailsApplication
 import org.pih.warehouse.core.MailService
+import org.pih.warehouse.core.Person
 import org.pih.warehouse.requisition.Requisition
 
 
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -0,0 +1,31 @@
...
+    GrailsApplication grailsApplication
+
+
+    void publishDefaultEmailNotifications(Requisition requisition, Collection receivers) {
+        def g = grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+        String subject = "${requisition.requestNumber} ${requisition.name}"
+        String redirectToRequestsList = "${g.createLink(uri: "/stockMovement/list?direction=OUTBOUND&sourceType=ELECTRONIC", absolute: true)}"
+        GString body = "${g.render(template: "/email/approvalsAlert", model: [requisition: requisition, redirectUrl: redirectToRequestsList])}"
+
+        mailService.sendHtmlMail(subject, body, receivers)
```
kchelstowski, line 29: I don't know if this method should not be placed in the NotificationService (it is done like that e.g. in `ShipmentStatusTransitionEventService`).

If you and the others agree that it should, then consider renaming the method - "default" word feels weird to me in this case.

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -0,0 +1,31 @@
...
+**/
+package org.pih.warehouse.inventory
+
+import org.codehaus.groovy.grails.commons.GrailsApplication
+import org.pih.warehouse.core.MailService
+import org.pih.warehouse.requisition.Requisition
+
+
+class RequisitionStatusTransitionEventService {
+
```
kchelstowski, line 18: please add
```groovy
static transactional = true
```
so that we remember to mark this service as Transactional when pulling it into Grails 3 branch

grails-app/views/email/_approvalsAlert.gsp
```diff
@@ -0,0 +1,53 @@
+<%@ page contentType="text/html"%>
+<style>
+label {
```
kchelstowski, line 3: add indent

grails-app/views/email/_approvalsAlert.gsp
```diff
@@ -0,0 +1,53 @@
...
+    font-weight: bold;
+}
+</style>
+<div>
+    <div class="header">
+        <g:render template="/email/header"/>
+    </div>
+    <div>
+        <g:message code='email.requestReceived.message' args='[requisition.destination, requisition.requestedBy, redirectUrl]' />
```
kchelstowski, line 12: nitpicky: let's use one type of `""` `''` - here you use a single one, and for others you use double one.

grails-app/views/email/_approvalsAlert.gsp
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+    &nbsp;
+</div>
+<div>
+    <table>
+        <thead style="text-align: left;">
```
kchelstowski, line 18: wouldn't `class="text-left"` work? It should be available in bootstrap

grails-app/views/email/_approvalsAlert.gsp
```diff
@@ -0,0 +1,53 @@
...
+                    <td>
+                        ${item?.product?.productCode}
+                    </td>
+                    <td>
+                        ${item?.product?.name}
+                    </td>
+                    <td>
+                        ${item?.product?.unitOfMeasure}
+                    </td>
+                    <td style="text-align: center">
```
kchelstowski, line 46: same as above, but with `class="text-center"`

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
alannadolny: I placed it here because of the draft that Justin created. I placed the "default" word in the name because of the same reason.

awalkowiak, line 29: I wonder if this should loop over receivers and send an email to each separately instead of one email to all. Additionally, the next related ticket is to make a link point to that approver's list of requests, not to all requests (which indicates it will have to have different link for each one of recipients)

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
alannadolny: Is there a reason that it should be transactional in the current form in Grails 3? 🧐

grails-app/views/email/_approvalsAlert.gsp
alannadolny: it doesn't work here

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
alannadolny: Good point, fixed

awalkowiak (approved):
I am merging this as is, and this one can be adjusted or improved if needed

Issue #4235, opened 2023-08-20
Volunteer requesting to help clean up API docs - some points


micshar92:
Hi, OpenBoxes team. I'm a technical writer with a background in supply chain, and I want to contribute to your documentation because you're in that field and I can stand behind your product ethically. Fair warning: I have to yet to be hired, for pay, as a technical writer— I just have the "book learning" right now, and a few practice experiences. ([here's my portfolio](https://micshar92.github.io/), and [here's my last open-source contribution](https://github.com/json-schema-org/understanding-json-schema/pull/205), for proof.)

I want to contribute to your API documentation to help me get my foot in the door for my first salaried job. So with that in mind, I've read a lot of best practices, API design, how they work, etc. docs over these past few months. I suggest the following changes to yours. (All up for debate; tell me if you think each is a good idea.)

1. The [overview](https://docs.openboxes.com/en/develop/api-guide/) page I think needs to be merged with the [getting started](https://docs.openboxes.com/en/develop/api-guide/getting-started/) page. Maybe rename the "Generic API" title to "Using the Generic API".
2. It's generally best practice to have a "one-pager" because, at least from what I've read developers say, they like to use ``Ctrl+F`` to easily find what they're looking for instead of searching through multiple pages with different headings. This doesn't mean ONLY a one-pager— the other conceptual/task pages are important, too. 
3. "Authentication and authorization" is a typical page for API docs— you currently have the info on a page that falls under a different header. I think it needs its own page.
4. Maybe a quick reference would be useful? [example](https://s3.us-west-1.wasabisys.com/idbwmedia.com/images/api/sample_api_diagram.pdf)
5. Maybe a glossary would be useful? [example](https://developer.lyft.com/docs/glossary)
6. General proofreading of the Swagger API. There are some typographical errors.
7. I might be overstepping my authority here, but I think the very dark blue text that denotes a link is too hard to distinguish from the regular text. The links are not underlined, either, so it's hard to tell (absent context clues) what is a link in the text. If my contributions include links, I think it would be better to make them more visibly distinct than the regular text. I also think that your code samples on all pages would greatly benefit from being color-coded (I can easily set this in the ``mkdocs.yml``). Maybe even change the code block background to black, with color text? When I scroll through the pages, the drab grayness makes the code harder to read and blurs together. But I'd also understand if you want to stick to your brand colors. Let me know what you think.
8. A few small things that don't impact much (for instance, a broken list on the core identifiers page).

[...]

PR #4285, opened 2023-09-16
Requesting JavaScript code review for new glossary popovers feature

micshar92:
Hi, everyone. I'm a junior technical writer volunteering here to beef up my portfolio. My background is in supply chain and data, and I want to prove my ability in APIs and code. That said, while I understand Python, SQL, and markup languages, JavaScript is proving existentially frustrating.

I'm requesting a JS code review for my glossary popovers feature. The idea is to have an `<a>` that when hovered over shows the term and definition (bootstrap popover title and content respectively) fetched from the _glossary.json_ file. Wikipedia and the GitHub docs have something similar. Mine, however, show up like this:

Screenshot 2023-09-15 150825

Where `bin_lTerm` is supposed to be "bin location" and `bin_lDef` is supposed to be its definition, as I declare earlier in the JS. Here's an example of what I'm trying to do:

Screenshot 2023-09-16 113530

I might be on the right track, I just can't implement it correctly. Forgive me; I don't think I have a developer's brain.

My idea is:

1. fetch() from a glossary.json file
2. parse() the "title" and "definition", stringify(), store as variable
3. call variable names in title and content of popover
4. concatenate the "html" json string with "See also `<a href="LINK">ARTICLE</a>`" if the term links to a tutorial

But here are the problems I know of:

1. Basically everything under the first block in _POPOVER_TEST_PAGE.md_ is wrong.
2. On the same document, why doesn't `trigger="focus hover"` work? It only works on click.
3. Is my JSON structure right in _glossary.json_? I feel it's weird.
4. In _glossary.md_, there are more probably-wrong `fetch()` functions. I'm also not sure about the concatenation for definitions that include an HTML link after it.

Please review / help. Thanks!

(*Note: _glossary.md_ is only populated with two sample terms as of now. I didn't want to write every term if the code for them was wrong.*)

(*Note 2: Once this is all correct, I'll get to work on finding terms across the site's guide pages and making them into the popover class. I _think_ that's right.*)


jmiranda:
@micshar92 Apologies for the delay. I probably won't have a chance to review this week but have requested a review by other developers. I will take a look next week.

awalkowiak (commented):
@micshar92 the pages that you want to add are `markdown`, and you are trying to squeeze in html and jquery. I am not sure if this makes sense. Here is how you can achieve tooltips in markdown: https://github.com/BoostIO/BoostNote-Legacy/issues/3579#issuecomment-855449583 (not a very good-looking version of tooltips, but works).

awalkowiak:
@micshar92 here is an example of "tooltip" for bin location:
markdown:
```markdown 
Lorem ipsum [bin location](## "Bin location
A location that designates a particular shelf, pallet position, or bin within a given zone in a given depot. For example, within Zone A, you may have a bin location for pharmacy items.")  something something
```
And how it looks like:
Lorem ipsum [bin location](## "Bin location
A location that designates a particular shelf, pallet position, or bin within a given zone in a given depot. For example, within Zone A, you may have a bin location for pharmacy items.") something something

micshar92:
@awalkowiak I briefly considered the HTML `<abbr>` tag, which is basically the same thing with a little styling, but I thought it was subpar compared to the popover. Furthermore, if a definition changes or is removed, any page with that term would have to be updated, and that would quickly spiral into busy work and neglect ('cause ain't nobody got time for that). I know glossaries are not as important as regular docs, but the popover would in some cases have an internal link to another article for SEO benefits and sometimes reducing user frustration. Not so with a tooltip.

Still, I think it's best to populate the terms from an external source file. Markdown isn't built for huge amounts of code inside `<script>` tags, I know. I could make the _glossary.md_ -> _glossary.html_, but I don't think it would change much. And a glossary page is very low traffic if it doesn't appear to be linked to other pages. It's a little like a sprig of parsley on a high-class dish.

Maybe I'm trying to do something too complex. It seems simple in theory to me, just hard to actually write the code. Is it actually impossible?

awalkowiak:
@micshar92 my example is still a link, but I have not used the link part in it, so it can be added. I am still not sure about using HTML and jQuery when the convention (other docs here in the project) is using a markdown.

@jmiranda what are your thoughts about this?

micshar92:
@awalkowiak Yeah, you're right. Scrap the HTML idea.

As for the tooltip you showed... I can compromise on it. To me it's close-to-perfect, where the popover would be perfect. BUT maybe that's just me wanting my idea. It seems to carry the same functions, and it's drastically easier.

Not all of the terms have links, though. I'm thinking a simple style change: make the underline dotted for no-click terms, make all terms a regardless of type cadet blue. If my original idea is impossible, we now have a Plan B.

micshar92:
@awalkowiak @jmiranda How about this as a compromise: since I have little desire to learn JavaScript, and because I don't think a definition change will take that long to fix or be that hard (ctrl-F), let's drop the JSON/JavaScript idea. Instead, I'll go purely with HTML and SCSS.

I saw this screenshot in the recent release notes. I'll just copy the style you use here, and add `<a>` links when necessary. Simple. Easy.

ca574934750ebe59ea9c4ef8214f389a8b796a40

Tell me how you made these, and I'll use the same style for the terms, if it's possible.

awalkowiak:
@micshar92 these pages are made in React, and for these particular popups, we used this package: https://www.npmjs.com/package/react-tippy.

I have a question though - was the HTML + CSS requested here to be used instead of a simple markdown?

micshar92:
@awalkowiak Markdown lets you use `<html>` and `<style>` tags for added flexibility: that's the HTML and SCSS I mean. I just think the terms need to be styled a little differently than a regular hyperlink, and the html is largely `<dl>`, `<dt>`, and `<dd>`: definition list, definition term, definition definition (SEO stuff). Terms are styled by just a few added lines to the main.scss file. The glossary page will still be Markdown.

micshar92:
(React doesn't appear to work with Markdown, but...)
@awalkowiak Hey! So I wrote this tooltip entirely in CSS. Here it is in action before hover:

Screenshot 2023-10-04 135234

... and after hover.

Untitled


...but then I realized there's no opportunity for me to use this. Why did I make this? Outside of the PiH pages or the wiki-- neither repo of which I can edit-- there are no terms on any page to use a tooltip with.

Anyway, I did still make the glossary page, but it's like the garnish on a fancy dish and I don't expect it to get much traffic. Oh well. If you choose to accept this super simple thing, here you go.

Looking forward to something more substantial, like what @jmiranda and I talked about in #4235 with Postman. OpenBoxes is directly in my writing niche, so I'll stick around.

micshar92:
(React doesn't appear to work with Markdown, but...)
@awalkowiak Hey! So I wrote this tooltip entirely in CSS. Here it is in action before hover:

Screenshot 2023-10-04 135234

... and after hover.

Untitled


...but then I realized there's no opportunity for me to use this. Why did I make this? Outside of the PiH pages or the wiki-- neither repo of which I can edit-- there are no terms to use a tooltip with on any page.

Anyway, I did still make the glossary page, but it's like the garnish on a fancy dish and I don't expect it to get much traffic. Oh well. If you choose to accept this super simple thing, here you go.

Looking forward to something more substantial, like what @jmiranda and I talked about in #4235 with Postman. OpenBoxes is directly in my writing niche, so I'll stick around.

jmiranda:
Looks good @micshar92. 

FWIW, the following mkdocs plugin (https://github.com/AngryMane/mkdocs-glossary-plugin) might be useful. It doesn't solve the problem but gets us close.

When a page is loaded, the mkdocs glossary plugin will iterate over the glossary terms defined in the glossary directory

image

and attempt to find those glossary terms within the page. When it finds one, it annotates (wraps) the term with a link to the glossary term.

image

And that link leads us to an individual page with the glossary term definition.

image

This isn't exactly what we want but it gets us part of the way by allowing us to 
1. Define glossary terms 
2. Find and replace glossary terms within the DOM

So we can either extend the plugin to allow for a link vs tooltip option. This might be more difficult than I'm imagining but I thought this was cool. 

The other bummer about this plugin is that we need to maintain a separate file per glossary term instead of a JSON file with everything defined. But it solves the one problem that I expected to be difficult i.e. automating the application of glossary terms in the DOM. So perhaps it buys us enough to work with.

jmiranda:
The mkdocs Annotations plugin (https://squidfunk.github.io/mkdocs-material/reference/annotations/) also looks promising but requires the annotations to be included in the source markdown content.

image

Not sure if mkdocs has a pre-process/post-process hooks, but if so we could potentially rewrite the markdown content during the deployment process OR perhaps use the two plugins together.

jmiranda:
Here's another potential solution.
https://github.com/squidfunk/mkdocs-material/issues/1919

docs/support/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+    text-decoration: underline;
+  }
+</style>
+
+<html>
+<dl>
+  <dt>ABC Analysis class</dt>
+  <dd>This class system categorizes products based on their impact on cost. Products in Class A contribute to the majority of costs because you use and buy them most often. You buy Class B products more infrequently. Class C items are not high value and do not need many follow-ups.</dd>
+  
+  <br>
```
jmiranda, line 22: You can remove the <br/> elements as we can add space using CSS.

jmiranda (changes requested):
LGTM just remove the <br/> tags and we can merge this.

docs/support/glossary.md
micshar92: Done. You're right: in my typography notes, I wrote down long ago to never use carriage returns. Looks like I forgot this time. Good lesson to learn before making that mistake for a big company.

micshar92: @jmiranda Whoops didn't @ you. Anyway, it's done.

PR #4286, opened 2023-09-18
OBPIH-5802 select approver in create request approvals end to end workflow

drodzewicz:


src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
```diff
@@ -198,6 +231,14 @@ class CreateStockMovement extends Component {
     if (this.state.values.origin && this.state.values.destination) {
       this.fetchStockLists(this.state.values.origin, this.state.values.destination);
     }
+    if (this.state.values.origin) {
+      if (this.state.values.origin?.supportedActivities?.includes(ActivityCode.APPROVE_REQUEST)) {
+        this.setSupportsApprover(true);
+        this.fetchAvailableApprovers();
+      } else {
+        this.setSupportsApprover(false);
+      }
```
alannadolny, line 240: What is the default value of `supportsApprover`? Maybe if it is false it could be shortened to
```
      if (this.state.values.origin?.supportedActivities?.includes(ActivityCode.APPROVE_REQUEST)) {
        this.setSupportsApprover(true);
        this.fetchAvailableApprovers();
      }
```
(if it works I can see 2 more similar places to change)

src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
```diff
@@ -198,6 +231,14 @@ class CreateStockMovement extends Component {
     if (this.state.values.origin && this.state.values.destination) {
       this.fetchStockLists(this.state.values.origin, this.state.values.destination);
     }
+    if (this.state.values.origin) {
+      if (this.state.values.origin?.supportedActivities?.includes(ActivityCode.APPROVE_REQUEST)) {
```
kchelstowski, line 235: what about destination's supportedActivities? Has it been clarified what we should look at exactly?

src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
kchelstowski: @alannadolny you can change the origin many times, so you could potentially set an origin that doesn't support the request approval and the flag would not change from `true` to `false`

src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
```diff
@@ -253,6 +301,20 @@ class CreateStockMovement extends Component {
...
+      params: { roleTypes: RoleType.ROLE_REQUISITION_APPROVER },
+    })
+      .then((response) => {
+        this.setState({
+          availableApprovers: response.data.data,
+        }, () => this.props.hideSpinner());
+
+        return response.data.data;
+      })
+      .catch(() => this.props.hideSpinner());
```
kchelstowski, line 315: let's maybe add `.finally` here that would hide the spinner instead of hiding in both in `.catch` and in `.then`?

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -69,7 +69,13 @@ class SelectOptionsApiController {
     }
 
     def usersOptions = {
-        List<User> users = userService.findUsers(params)
+        params.roleTypes = params.list("roleTypes")
+        if (params.roleTypes && !params.location) {
+            params.location = session.warehouse?.id
+        }
+        List<User> users = userService.findUsers(params).collect {
+            [id: it.id, label: it.name, value: it.id ]
```
kchelstowski, line 77: maybe not for now, but maybe we should think about creating a dto called `SelectOptionDto` for those select options?
In all places we return the same object which is object with `id`, `label`, `value`.
It'd be easier to manage those select options if we ever wanted e.g. to add a new field to be returned etc.
But as I said - it's probably not for now, but it was a good moment to raise it.

src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
drodzewicz: the default value is `false`, but this `this.setSupportsApprover(false);` is needed in cases when we are changing the selected fulfilling location from  **supports APPROVE_REQUEST** to  **not supports APPROVE_REQUEST**.

If my explanation is a bit confusing, let me give you an example.
I chose a location that supports APPROVE_REQUEST  -> the select should appear
then I change the location which does not support APPROVE_REQUEST  -> the select should disappear

src/js/components/stock-movement-wizard/request/CreateStockMovement.jsx
drodzewicz: good idea, I was just copy/pasting the previous fetch methods in the component and followed the "convention"

grails-app/conf/UrlMappings.groovy
```diff
@@ -82,7 +82,7 @@ class UrlMappings {
             action = [GET: "paymentTermOptions"]
         }
 
-        "/api/users" {
+        "/api/usersOptions" {
```
awalkowiak, line 85: Why this change? You should not change an existing endpoint like that.

awalkowiak, line 85: Why this change? You should not change an existing endpoint like that.

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
awalkowiak: Did the structure of the response change? If yes then it should be a separate endpoint.

awalkowiak: Did the structure of the response change? If yes then it should be a separate endpoint.

awalkowiak (changes requested):

grails-app/conf/UrlMappings.groovy
drodzewicz: In my opinion, this should have been done like that in the first place.
This is an endpoint from `selectOptionsApiController` and looking at previous "...Options" endpoints, this is the approach we went with for these kinds of endpoints.

Also, this endpoint is only used in debounce fetch for user-select which data was transformed on the frontend.

awalkowiak:
@drodzewicz but you still changed the url of this endpoint

awalkowiak (approved):

PR #4287, opened 2023-09-19
OBPIH-5802 Fetch users based on role in fulfilling location

drodzewicz:
Forgot to add the location parameter on the frontend when fetching approvers based on location roles.

~Additionally fixed the multi-select tooltip, which was showing up because it's value is an `Array` instead of simple `[]`.
When casting an `Array(0)` to a `boolean` it returns `true` unlike casting a `[]` which returns `false`~

Additionally fixed the multi-select tooltip, which was showing up because it's value is an array which is a truthy even when it is empty.


kchelstowski (approved):
> unlike casting a [] which returns false 

`[]` is truthy in JS, so is it a typo or I don't understand something?

drodzewicz:
@kchelstowski disregard the previous comment I was thinking about a different thing, but yes you are right, array is a truth value.

awalkowiak (approved):

PR #4288, opened 2023-09-20
OBPIH-5803 Add submit for approval functionality to create request workflow

kchelstowski:


grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -33,4 +39,13 @@ class RequisitionStatusTransitionEventService {
             }
         }
     }
+
+    void onApplicationEvent(RequisitionStatusTransitionEvent event) {
+        requisitionService.triggerRequisitionStatusTransition(event.requisition, event.newStatus)
+
+    }
```
alannadolny, line 46: Please, can you call `publishDefaultEmailNotifications` as it was mentioned in the tech huddle notes? I will be grateful.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -150,7 +151,8 @@ class Requisition implements Comparable<Requisition>, Serializable {
         statusSortOrder formula: RequisitionStatus.getStatusSortOrderFormula()
         monthRequested formula: "date_format(date_requested, '%M %Y')"
         comments joinTable: [name: "requisition_comment", key: "requisition_id"]
-        events joinTable: [name: "requisition_event", key: "requisition_id"]
+        // Fetch join in the events should be removed in Grails 3 and replaced with proper join query using data services
+        events joinTable: [name: "requisition_event", key: "requisition_id"], fetch: 'join'
```
kchelstowski, line 155: I was getting a `LazyInitializationException` when accessing the events in the `RequisitionStatusTransitionEvent` - it is because the event is done in a different session (after the session where we fetch the Requisition ends).

Current solution actually adds an eager fetching of the events, but using the `left join` clause, not calling a second query.
Normally I would not do it and would not recommend doing an eager fetch "globally", but I decided that it was not worth to spend time to create a custom query for that, because in Grails 1 we don't have **data services.** 
This is why I left a comment to refactor this ASAP in Grails 3, to use a proper joining query using **data services**

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -230,6 +231,7 @@ class StockMovementService {
     void updateRequisitionStatus(String id, RequisitionStatus status) {
 
         log.info "Update status ${id} " + status
+        // TODO: In Grails the get below should be replaced by the data service get that joins the Events
```
kchelstowski, line 234: here it sohuld be replaced by the data service query, something like `getWithEvents` that would `@Join('events')`
I know that @drodzewicz has run into the same issue but with approvers, so most probably we'll have to write the query using `@Query` and join those two tables (if I forget - remember to take care of **left** join there)

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -812,4 +816,30 @@ class RequisitionService {
...
+        if (status == RequisitionStatus.PENDING_APPROVAL) {
+            if (requisition.origin.approvalRequired) {
+                Event event = createEvent(EventCode.PENDING_APPROVAL, requisition.origin, new Date())
+                requisition.addToEvents(event)
+                requisition.status = RequisitionStatus.PENDING_APPROVAL
+                requisition.approvalRequired = true
+                requisition.save()
+                return
+            }
+            Event event = createEvent(EventCode.SUBMITTED, requisition.origin, new Date())
```
kchelstowski, line 838: I am not sure if `SUBMITTED` is a correct code here, so ready for discussion/suggestions.
(It is if we submit a request "in an old way", so if we don't support the request approval workflow)

grails-app/domain/org/pih/warehouse/core/Event.groovy
```diff
@@ -20,12 +21,18 @@ package org.pih.warehouse.core
  *{eventDate: 5/5/2010, eventLocation: Customs, eventType: ARRIVED}*/
 class Event implements Comparable, Serializable {
 
+    def beforeInsert = {
```
alannadolny: I remember that we were changing this from closures to methods. I am not sure if the methods work in Grails 1, but if they work, I think it would be better to change them to avoid inconsistency in Grails 3.

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -339,6 +339,10 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return active ? LocationStatus.ENABLED : LocationStatus.DISABLED
     }
 
+    Boolean isApprovalRequired() {
+        return supportedActivities?.contains(ActivityCode.APPROVE_REQUEST.id)
+    }
+
```
alannadolny, line 345: I can see that there is a method called `supports` maybe it would be better to use it?
I mean:
```
Boolean isApprovalRequired() {
        return supports(ActivityCode.APPROVE_REQUEST.id)
}
```

grails-app/migrations/0.8.x/changelog-2023-09-21-1330-alter-table-requisition-add-rejectedby-field.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="kchelstowski" id="210920231330-1">
```
alannadolny, line 6: Just a question, why are you numbering change sets from 1 instead of 0? 🤔

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -13,12 +13,18 @@ import org.codehaus.groovy.grails.commons.GrailsApplication
...
 import org.pih.warehouse.requisition.Requisition
+import org.pih.warehouse.requisition.RequisitionService
+import org.pih.warehouse.requisition.RequisitionStatusTransitionEvent
+import org.springframework.context.ApplicationEvent
+import org.springframework.context.ApplicationListener
 
+class RequisitionStatusTransitionEventService implements ApplicationListener<RequisitionStatusTransitionEvent> {
 
-class RequisitionStatusTransitionEventService {
+    static transactional = true
```
alannadolny, line 23: Is this transactional really necessary? (I know that you marked it to remember adding a transaction in Grails 3, but I am not sure if there is a need for the transaction)

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -121,6 +121,7 @@ class Requisition implements Comparable<Requisition>, Serializable {
 
     // Request approval fields
     Person approvedBy
+    Person rejectedBy
```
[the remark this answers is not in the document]
alannadolny: I know that @drodzewicz also added it within the ticket that he is working on, so we have to remember to avoid duplicating them

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
kchelstowski: @drodzewicz this would mean we’d have to write a pure SQL in order to fetch that. 
I just thought it’s waste of time for now, since we’ll migrate in a moment and shouldn’t experience any performance issues during that time

grails-app/domain/org/pih/warehouse/core/Event.groovy
kchelstowski: This is outdated - I had to remove this approach, since triggering an event creates its own session, hence the currentUser was null

grails-app/domain/org/pih/warehouse/core/Location.groovy
kchelstowski: Yeah, thanks for pointing that out, I forgot to ask about it as well - if we should use contains like I did, or supports, because supports also checks for locationType’s supportedActivities and I was not sure whether the request approval is supposed to work if we add this activity to a locationType

grails-app/migrations/0.8.x/changelog-2023-09-21-1330-alter-table-requisition-add-rejectedby-field.xml
kchelstowski: I don’t know 😄

grails-app/domain/org/pih/warehouse/core/Location.groovy
awalkowiak: I think it should use `supports`

grails-app/domain/org/pih/warehouse/core/Event.groovy
```diff
@@ -9,6 +9,7 @@
  **/
 package org.pih.warehouse.core
 
+import org.pih.warehouse.auth.AuthService
```
awalkowiak, line 12: Is this required?

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
```diff
@@ -0,0 +1,66 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="kchelstowski" id="210920231400-1">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="event_type" />
+        </preConditions>
+        <insert tableName="event_type">
+            <column name="id" value="abb5b3d2-da3e-4a4c-9e3c-de139147f822"/>
+            <column name="version" valueNumeric="0"/>
+            <column name="date_created" valueDate="2023-09-21T00:00:00.0"/>
```
awalkowiak, line 13: Perhaps it's better to use `<property name="now" value="now()"/>` for these dates.

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
```diff
@@ -0,0 +1,66 @@
...
+            <column name="description" value="Request has been sent for approval"/>
+            <column name="event_code" value="PENDING_APPROVAL"/>
+            <column name="last_updated" valueDate="2023-09-21T00:00:00.0"/>
+            <column name="name" value="Waiting for approval"/>
+            <column name="sort_order" valueNumeric="1"/>
```
awalkowiak, line 18: I am not sure if these sort orders make sense, but I am not sure if this is currently used for anything at all.

grails-app/domain/org/pih/warehouse/core/Location.groovy
jmiranda: Agreed

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
jmiranda: I would probably try a withNewSession in the event handler. 
```
Requisition.withNewSession { 
    requisition = Requisition.read() // or Requisition.get() if you need to set something
}
```
Or if that doesn't work, just do a Requisition.read(id) or get(id) in the event handler. 

Lastly, you could potentially use refresh the event handler but that one always scares me.
```
requisition.refresh()
```
Not sure if you need the withSession or withNewSession wrapper so I'll leave that as an exercise for the reader.

grails-app/migrations/0.8.x/changelog-2023-09-21-1330-alter-table-requisition-add-rejectedby-field.xml
jmiranda: Blasphemous behavior. :smiling_imp:

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
jmiranda: Interesting. I hadn't really thought about the use of now() until you mentioned it here. I'm not sure I agree, but I'm fine with either approach. There's something to be said for having a static value for static data to ensure that everyone has the exact same data, down to the timestamps. 

There's also an argument for knowing the exact time the data was created in case you are troubleshooting a database migration issue during deployment (i.e. a failed precondition). 

And just from an aesthetic point of view, sometimes when I do a `select * from event_type` or some other static data table, there's a nice fuzzy feeling of seeing 2012 in the date_created. That is usually followed quickly with dread and despair of knowing I'm old af. 

Ok then to prevent existential crises ... let's use now().

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
jmiranda: It's a good question. If we ever do 
```
event1 > event2
```
then we probably need to deal with it. 

But I assume we're using the ordinal value on the EventCode to do the sorting, so probably not an issue.

I'm ok leaving this for now OR changing sort order to 0.

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
jmiranda: If we're just doing reads then we don't need a transaction. If you're writing to the database then we need a transaction.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: You *could* (but don't have to) add a getWithEvents() to the Requisition class now.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -244,6 +246,7 @@ class StockMovementService {
         } else {
             requisition.status = status
             requisition.save(flush: true)
+            publishEvent(new RequisitionStatusTransitionEvent(requisition, requisition.status))
```
jmiranda, line 249: You could pass the requisition.id here in order to force a read or get on the event service. We could think about doing that across all of our events.

jmiranda: Is there any need to pass both the from and to status here?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -812,4 +816,31 @@ class RequisitionService {
             WHERE r.origin = :location AND r.dateCreated >= :startDate AND r.dateCreated <= :endDate AND r.isTemplate = false 
             """, ['location': location, 'startDate': startDate, 'endDate': endDate])[0] ?: 0
     }
+
+    Event createEvent(EventCode eventCode, Location eventLocation, Date eventDate, User currentUser) {
+        EventType eventType = EventType.findByEventCode(eventCode)
+        if (!eventType) {
+            throw new NotFoundException("No event type with code ${eventCode} has been found")
```
jmiranda, line 823: Good call on the exception.

jmiranda (commented):
I can't remember if I requested changes, so just leaving this as a comment. LGTM.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
kchelstowski: of course `get`/`read` will work, but I wanted to avoid additional query to refetch it, but if you are fine with it, I can just fetch it again in the event service.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: yeah, but since those data service annotations are not available in Grails 1, I would anyway have to write this HQL/SQL query by myself using `.executeQuery`, wouldn't I?

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
kchelstowski: In general I would not agree, because not having a transaction boundary also causes LazyInitializationExceptions while reading (assuming you don't use osiv). Anyway as I mentioned, it was actually added for debugging purposes and finding a solution for the Lazy that was thrown.
I will remove it.

grails-app/domain/org/pih/warehouse/core/Event.groovy
kchelstowski: Not anymore, forgot to delete it in the latest commit.

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
kchelstowski: Definitely we should avoid existential crises :laughing: so I'll change that

grails-app/migrations/0.8.x/changelog-2023-09-21-1400-add-eventtypes-for-request-approval-workflow.xml
kchelstowski: it seems not to be used anywhere at the moment, but since the default event types also have the sortOrder, I decided to add it.
Not to cause confusions I can either remove it (since it is nullable) or change it to 0. Let me know.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: Could you elaborate on what you mean by "from and to" status?

awalkowiak (approved):
I think it LGTM

PR #4289, opened 2023-09-20
OBPIH-5821 Move send notification logic to notificationService and add individual links for approvers to request list

drodzewicz:
@alannadolny When I called the `publishDefaultEmailNotifications()` in the `onApplicationEvent` there was a problem with `g.render` and `g.crealink` which threw an exception

```
Method threw 'java.lang.IllegalStateException' exception.

No thread-bound request found: Are you referring to request attributes outside of an actual web request, or processing a request outside of the originally receiving thread? If you are actually operating within a web request and still receive this message, your code is probably running outside of DispatcherServlet/DispatcherPortlet: In this case, use RequestContextListener or RequestContextFilter to expose the current request.
```

looking at other methods connected to sending mail located in `notificationService` I noticed that they use `renderTemplate()` helper function to render these templates. Most likely it is a workaround for the same problem which I mentioned before.

Additionally, I decided to move this method to **notificationService** since most of our "sendMail" functions live there, unless you disagree with this decision then I'll move it back.

--------------------------------------------------------------------------------------------------------------------------------------------
For now, I selected a base branch `OBPIH-5803` because I needed the functionality of triggering the event on submit to send an email.
When that branch is going to be merged to develop I will rebase and will change the base branch to develop.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
@@ -23,7 +23,7 @@ class RequisitionStatusTransitionEventService implements ApplicationListener<Req
 
     void onApplicationEvent(RequisitionStatusTransitionEvent event) {
         Requisition requisition = event.requisition
-        notificationService.publishRequisitionStatusTransitionNotifications(requisition, event.newStatus)
-        requisitionService.triggerRequisitionStatusTransition(requisition, event.newStatus, event.createdBy)
+        requisitionService.triggerRequisitionStatusTransition(requisition, event.createdBy)
+        notificationService.publishRequisitionStatusTransitionNotifications(requisition)
     }
 }
--- a/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
@@ -10,6 +10,8 @@
 package org.pih.warehouse.inventory
 
 import org.pih.warehouse.core.Person
+import org.pih.warehouse.core.UserService
+import org.pih.warehouse.core.RoleType
 import org.pih.warehouse.report.NotificationService
 import org.pih.warehouse.requisition.Requisition
 import org.pih.warehouse.requisition.RequisitionService
@@ -22,14 +24,21 @@ class RequisitionStatusTransitionEventService implements ApplicationListener<Req
 
     RequisitionService requisitionService
     NotificationService notificationService
-
+    UserService userService
 
     void onApplicationEvent(RequisitionStatusTransitionEvent event) {
         if (event.newStatus == RequisitionStatus.PENDING_APPROVAL) {
             Requisition req = event.requisition
-            List<Person> approvers = Person.findAllByIdInList(event.requisition.approvers.collect {it.id})
+
+            List<Person> approvers
+            if (req.approvers?.size()) {
+                approvers = Person.findAllByIdInList(req.approvers.collect { it.id })
+            } else {
+                // if there are not assigned users as approvers then notify all approvers
+                approvers = userService.findUsersByRoleTypes(req.origin, [RoleType.ROLE_REQUISITION_APPROVER])
+            }
             notificationService.sendRequestPendingForApprovalNotification(req, approvers)
         }
-        requisitionService.triggerRequisitionStatusTransition(event.requisition, event.newStatus, event.createdBy)
+        requisitionService.triggerRequisitionStatusTransition(req, event.newStatus, event.createdBy)
     }
 }
--- a/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
@@ -9,36 +9,27 @@
 **/
 package org.pih.warehouse.inventory
 
-import org.codehaus.groovy.grails.commons.GrailsApplication
-import org.pih.warehouse.core.MailService
 import org.pih.warehouse.core.Person
+import org.pih.warehouse.report.NotificationService
 import org.pih.warehouse.requisition.Requisition
 import org.pih.warehouse.requisition.RequisitionService
 import org.pih.warehouse.requisition.RequisitionStatusTransitionEvent
+import org.pih.warehouse.requisition.RequisitionStatus
 import org.springframework.context.ApplicationEvent
 import org.springframework.context.ApplicationListener
 
 class RequisitionStatusTransitionEventService implements ApplicationListener<RequisitionStatusTransitionEvent> {
 
-    MailService mailService
-    GrailsApplication grailsApplication
     RequisitionService requisitionService
+    NotificationService notificationService
 
-
-    void publishDefaultEmailNotifications(Requisition requisition, List<Person> receivers) {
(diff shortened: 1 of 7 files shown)


grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -9,40 +9,30 @@
...
-            if (receiver.email) {
-                mailService.sendHtmlMail(subject, body, receiver.email)
-            }
-        }
-    }
-
     void onApplicationEvent(RequisitionStatusTransitionEvent event) {
+        if (event.newStatus == RequisitionStatus.PENDING_APPROVAL) {
+            Requisition req = event.requisition
+            List<Person> approvers = Person.findAllByIdInList(event.requisition.approvers.collect {it.id})
```
drodzewicz, line 32: When looping through users when sending the notification I was getting a LazyInitialization exception, which is why I needed to fetch the approvers list separately.

This looks a bit weird to me but I didn't find any other approach to fetch approvers eagerly, if you have a suggestion then please let me know.

kchelstowski: You can follow what I've done for the same problem with requisition events in my draft PR #4288  - you can add `fetchMode: 'join'` in Requisition mapping in this place:
https://github.com/openboxes/openboxes/blob/98806dcf18313fa6b2d27fb78641361a69bd0d0a/grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy#L154

It will `left join` approvers when fetching a Requisition and won't cause an additional query to fetch the approvers, as you've currently done.

Normally, I would not recommend doing this, since it kinda adds fetch mode eager for the Requisition, but since we'll migrate to Grails 3 in a moment, I've just left a comment above mapping, that it should be removed and replaced by a proper data service get method with joining those.

Data services are not available in Grails 1, so I think it is not worth to "fight" with that Lazy in a "pattern" way, since we'll refactor that in a moment.
Remember to add a comment to refactor this in Grails 3 - the same, as I've added for events.

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -9,36 +9,36 @@
...
-            if (receiver.email) {
-                mailService.sendHtmlMail(subject, body, receiver.email)
+            List<Person> approvers
+            if (req.approvers?.size()) {
+                approvers = Person.findAllByIdInList(req.approvers.collect { it.id })
+            } else {
+                // if there are not assigned users as approvers then notify all approvers
+                approvers = userService.findUsersByRoleTypes(req.origin, [RoleType.ROLE_REQUISITION_APPROVER])
             }
+            notificationService.sendRequestPendingForApprovalNotification(req, approvers)
```
jmiranda, line 40: Why can't we do all of the above in a method in the NotficationService? 

In fact let's create a method to resolve notification recipients (given a requisition). That way it's DRY and we we can extend it for other notifications. In addition, this makes it more testable. 
```
List<Person> resolveNotificationRecipients(Requisition requisition) { 
    List<Person> recipients = []
        if (requisition.approvers?.size()) {
            recipients = Person.findAllByIdInList(req.approvers.collect { it.id })
        } else {
            // if there are not assigned users as approvers then notify all approvers
            recipients = userService.findUsersByRoleTypes(req.origin, [RoleType.ROLE_REQUISITION_APPROVER])
        }
    }
    return recipients
}
```

jmiranda: If it wasn't for the userService call I would have recommended moving the "recipients" logic to the Requisition domain since it has all of the information needed to query for that.

jmiranda: Lastly, we should consider a slight renaming of the method as `sendRequestPendingForApprovalNotification(req, approvers)` is a little awkward.

I think `publishDefaultEmailNotifications` is closer to what I/we want although I'm glad you removed the "email" part since I'd like the notification service to be transport agnostic.

If we want the logic for what emails get sent in the event listener, then I guess
`publishRequisitionPendingApprovalNotifications` would be best.

If we want to allow the notification service to determine the what / when / who, then the following would be enough
```
publishNotifications(Requisition requisition)
```

jmiranda: So in summary 
```
        if (event.newStatus == RequisitionStatus.PENDING_APPROVAL) {
            notificationService.publishRequisitionPendingApprovalNotifications(event.requisition)
        } 
}
```

```
void publishRequisitionPendingApprovalNotifications(Requisition requisition) {
    List<Recipient> recipients = resolveNotificationRecipients(requisition)
    ...
}
```

Or 
```
        notificationService.publishRequisitionStatusTransitionNotifications(event.requisition)
        requisitionService.triggerRequisitionStatusTransition(event.requisition, event.newStatus, event.createdBy)
}
```
with NotificationService looking like this 
```
void publishRequisitionStatusTransitionNotifications(Requisition requisition) {
    List<Recipient> recipients = resolveNotificationRecipients(requisition)
        notificationService.publishRequisitionPendingApprovalNotifications(event.requisition)
    } 
}

Comment (jmiranda) [reply grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy:40]: I was in the middle of this comment when i had to stop so we can discuss this in more details on Monday's tech huddle.

**Review by jmiranda — COMMENTED** [review]
Let's tech huddle this.

**Review by alannadolny — APPROVED** [review]

File: grails-app/services/org/pih/warehouse/report/NotificationService.groovy [hunk 3]
```diff
@@ -226,4 +228,41 @@ class NotificationService {
...
+                    // if there are not assigned users as approvers then notify all approvers
+                    recipients = userService.findUsersByRoleTypes(requisition.origin, [RoleType.ROLE_REQUISITION_APPROVER])
+                }
+            }
+        }
+        return recipients
+    }
+
+    void publishRequisitionStatusTransitionNotifications(Requisition requisition, RequisitionStatus newStatus) {
+        List<Person> approvers = resolveNotificationRecipients(requisition)
```
jmiranda, line 248: Call this variable recipients

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -226,4 +228,41 @@ class NotificationService {
...
+
+        if (newStatus == RequisitionStatus.PENDING_APPROVAL) {
+            publishRequisitionPendingApprovalNotifications(requisition, approvers)
+        }
+    }
+
+    void publishRequisitionPendingApprovalNotifications(Requisition requisition, List<Person> receivers) {
```
jmiranda, line 255: recipients

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -226,4 +228,41 @@ class NotificationService {
...
+        String subject = "${requisition.requestNumber} ${requisition.name}"
+        String template = "/email/approvalsAlert"
+
+        receivers.each { receiver ->
+            if (receiver.email) {
+                String redirectToRequestsList = "/stockMovement/list?direction=OUTBOUND&sourceType=ELECTRONIC&approver=${receiver.id}"
+                String body = renderTemplate(template, [requisition: requisition, redirectUrl: redirectToRequestsList])
+                mailService.sendHtmlMail(subject, body, receiver.email)
```
jmiranda, line 263: recipient

PR #4290, opened 2023-09-20
OBPIH-5854 New approval statuses in SM list pages - Approvals lists pages

alannadolny:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,35 @@ class StockMovementApiController {
...
+        render([data: options, availableStatuses: availableStatuses] as JSON)
+    }
+
+    List<RequisitionStatus> getRequisitionStatusCodesList(boolean hasRequestApprovalActivity, boolean isRequestsList) {
+        if (isRequestsList) {
+            return hasRequestApprovalActivity ? RequisitionStatus.listRequisitionOptions() : RequisitionStatus.listOutboundOptions()
+        }
+
+        return hasRequestApprovalActivity ? RequisitionStatus.listOutboundOptionsWhenSupportingRequestApproval() : RequisitionStatus.listOutboundOptions()
+    }
```
alannadolny, line 831: The idea of this function is to use an appropriate list based on the list page that we are currently working on:
1. If we are on the requests list:
   a) If our location supports request approval activity - add waiting for approval and approved
   b) If our location doesn't support request approval - the options should be the same as we had before
2. If we are on the outbound list:
   a) If our location supports request approval activity - add the approved status
   b) If our location doesn't support request approval - the options should be the same as we had before

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,35 @@ class StockMovementApiController {
...
+        if (params.location) {
+            Location location = Location.get(params.location)
+            hasRequestApprovalActivity = location.supports(ActivityCode.APPROVE_REQUEST)
         }
-        render([data: options] as JSON)
+
+        List<RequisitionStatus> statuses = getRequisitionStatusCodesList(hasRequestApprovalActivity, isRequestsList)
+
+        List options = statuses?.collect(mapStatusCodes)
+        List availableStatuses = RequisitionStatus.listAll().collect(mapStatusCodes)
```
alannadolny, line 820: I had to split options on the dropdown from statuses that we can have on the list page. In the frontend part, we used fetched statuses from the dropdown to display their names on the list, but when we removed the verified status we still had at least one stock movement with this status, and the error was thrown. So I split those two cases (We were trying to show the status that we didn't fetch). Regarding the validation with visible requests in "approved"/"waiting for approval" status - we have a ticket for this and I think this should be improved/resolved here

src/js/hooks/list-pages/outbound/useOutboundFilters.jsx
```diff
@@ -35,10 +35,12 @@ const useOutboundFilters = (isRequestsList) => {
   useEffect(() => {
     // TODO: When having full React, if once fetched, fetch only if a current language differs
     // TODO: from the language, that we were fetching this for
-    dispatch(fetchRequisitionStatusCodes());
     dispatch(fetchShipmentTypes());
   }, [currentLocale]);
 
+  useEffect(() => {
+    dispatch(fetchRequisitionStatusCodes(currentLocation?.id, isRequestsList));
+  }, [currentLocale, currentLocation, isRequestsList]);
```
alannadolny, line 43: When we have different statuses on the dropdown, we should fetch them when:
- We are changing location (second location can not support requests approval, so this options should not be visible)
- We are changing locale (different translations)
- Changing list pages

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -74,6 +74,14 @@ enum RequisitionStatus {
         [CREATED, EDITING, VERIFYING, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
+    static listOutboundOptionsWhenSupportingRequestApproval() {
+        [CREATED, EDITING, APPROVED, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
+    }
+
+    static listRequisitionOptions() {
```
kchelstowski, line 81: I don't know if it's a good naming here - what do you mean by "Requisition" here?

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
kchelstowski: if `options` are still based on what `getRequisitionStatusCodesList` returns, can't we still run into this problem, since this method can filter out some of the statuses?

src/js/actions/index.js
```diff
@@ -531,13 +531,13 @@ export function fetchShipmentStatusCodes() {
 }
 
 
-export function fetchRequisitionStatusCodes() {
+export function fetchRequisitionStatusCodes(locationId = null, isRequestsList = false) {
```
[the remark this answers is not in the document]
kchelstowski: never mind - I see that it is also used in the `useOutboundFilters` hook, but anyway, I'll leave this comment here

kchelstowski (commented):
A general comment is that we probably need some comments for the RequisitionStatus methods.
In one place you use `listRequisitionOptions`, in other place `listOutboundOptions`, in other place `listOutboundOptionsWhenSupportingRequestApproval`, in other place `listAll()` - it's easy to get a headache while reading that and we should make sure to at least add comments above them to know, what they are supposed to exclude etc. 
I agree the naming of them might say that, but for example - I don't know what we mean by `listRequisitionOptions` here, and what's the difference between this and `listOutboundOptions` since outbound is a requisition based SM :man_shrugging:

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
alannadolny: I changed the naming here, but I am still open to better propositions

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -70,15 +70,18 @@ enum RequisitionStatus {
...
         [CREATED, EDITING, VERIFYING, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
+    // Options for outbounds when we are supporting request approval (Added approved)
     static listOutboundOptionsWhenSupportingRequestApproval() {
         [CREATED, EDITING, APPROVED, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
-    static listRequisitionOptions() {
+    // Options for request list when we are supporting request approval (Added approved and waiting for approval)
```
kchelstowski, line 83: "we" => current location

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -70,15 +70,18 @@ enum RequisitionStatus {
...
+    static listRequestOptionsWhenSupportingRequestApproval() {
         [CREATED, EDITING, APPROVED, WAITING_FOR_APPROVAL, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
```
kchelstowski, line 85: looking at this stuff, I've now started to think if we should not refactor those statuses methods to:

- create a common list method that would contain the intersection of all the methods
- for particular methods just concat it with the statuses that we are adding, e.g: `listCommon().concat([APPROVED])`
- or filter it if we want to exclude something

don't know if it should be part of this ticket, but I think it would improve the readability of those methods.

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -70,15 +70,18 @@ enum RequisitionStatus {
...
     }
 
+    // Options for outbounds when we are supporting request approval (Added approved)
     static listOutboundOptionsWhenSupportingRequestApproval() {
         [CREATED, EDITING, APPROVED, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
-    static listRequisitionOptions() {
+    // Options for request list when we are supporting request approval (Added approved and waiting for approval)
+    static listRequestOptionsWhenSupportingRequestApproval() {
```
kchelstowski, line 84: I'm stlil not sure about the naming for this one, but I don't have any suggestion so I'll leave it for others maybe

src/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
```diff
@@ -18,8 +18,10 @@ enum StockMovementStatusCode {
     CREATED(0),
     REQUESTING(1, PENDING),
     REQUESTED(2, PENDING),
+    WAITING_FOR_APPROVAL(2),
```
drodzewicz, line 21: Shouldn't this be `PENDING_APPROVAL`?
I saw Kacper has implemented these enums in #4288 so lets stick to the same name

src/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
```diff
@@ -18,8 +18,10 @@ enum StockMovementStatusCode {
...
     VALIDATING(2, PENDING),
     VALIDATED(3, PENDING),
+    APPROVED(3),
```
drodzewicz, line 24: `REJECTED` is probably missing right?

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
drodzewicz: Good point, I think we can even implement some kind of a `RequisitionStatusFactory` to generate a list of the statuses based on some given parameters.
But for now, unless anyone disagrees I think we can proceed with the individual lists

src/js/actions/index.js
```diff
@@ -531,13 +531,13 @@ export function fetchShipmentStatusCodes() {
 }
 
 
-export function fetchRequisitionStatusCodes() {
+export function fetchRequisitionStatusCodes(isRequestsList = false) {
   return (dispatch) => {
-    apiClient.get('/openboxes/api/stockMovements/requisitionsStatusCodes')
+    apiClient.get(`/openboxes/api/stockMovements/requisitionsStatusCodes?isRequestsList=${isRequestsList}`)
```
drodzewicz, line 536: let's use `params` attribute to pass parameters instead of template literals
```suggestion
    apiClient.get('/openboxes/api/stockMovements/requisitionsStatusCodes', {
        params: { isRequestsList },
    })
```

drodzewicz (changes requested):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
drodzewicz: I think I understand the problem but not fully. To me, this approach doesn't seem right but maybe I am missing some context.
Anyway, we will probably need to discuss this part further.

src/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
alannadolny: I don't need this status here

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -128,6 +139,8 @@ enum RequisitionStatus {
         return "(case status " +
             "when '${CREATED}' then ${CREATED.sortOrder} " +
             "when '${EDITING}' then ${EDITING.sortOrder} " +
+            "when '${PENDING_APPROVAL}' then ${PENDING_APPROVAL.sortOrder} " +
```
awalkowiak, line 142: Kacper's PR might add the same thing, watch out for conflicts.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,31 @@ class StockMovementApiController {
...
-            ]
+        Location currentLocation = Location.get(session.warehouse.id)
+        Boolean isRequestsList = params.getBoolean("isRequestsList")
+
+        List<RequisitionStatus> statuses = getRequisitionStatusCodesList(currentLocation.isApprovalRequired(), isRequestsList)
+
+        List options = statuses?.collect(mapStatusCodes)
+        List allStatuses = RequisitionStatus.listAll().collect(mapStatusCodes)
+
+        render([data: options, allStatuses: allStatuses] as JSON)
```
awalkowiak, line 818: Why you can't just return `[data: options]` based on the case?

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
awalkowiak: Yeah this is something that we could think about a bit

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,28 @@ class StockMovementApiController {
...
+    }
+
+    // Function for getting appropriate filter options based on current list and supporting requests approval
+    List<RequisitionStatus> getRequisitionStatusCodesList(Boolean hasRequestApprovalActivity, Boolean isRequestsList) {
+        if (isRequestsList) {
+            return hasRequestApprovalActivity ? RequisitionStatus.listRequestOptionsWhenSupportingRequestApproval() : RequisitionStatus.listOutboundOptions()
         }
-        render([data: options] as JSON)
+
+        return hasRequestApprovalActivity ? RequisitionStatus.listOutboundOptionsWhenSupportingRequestApproval() : RequisitionStatus.listOutboundOptions()
```
kchelstowski, line 824: it'd be cleaner if you reverted the statement, so:
```groovy
// If a location doesn't support the request approval, return listOutboundOptions no matter what list is displayed
if (!hasRequestApprovalActivity) {
 return RequisitionStatus.listOutboundOptions()
}
// If supports request approval, check what type of list page it is and return appropriate statuses
if (isRequestList) {
 return RequisitionStatus.listRequestOptionsWhenSupportingRequestApproval()
}
return RequisitionStatus.listOutboundOptionsWhenSupportingRequestApproval()
```

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
kchelstowski: I assume we don't want to spend time thinking about naming this one? @awalkowiak ?

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
```diff
@@ -43,6 +43,8 @@ const StockMovementOutboundTable = ({
     deleteConfirmAlert,
   } = useOutboundListTableData(filterParams);
 
+  useTranslation('requisitionStatus');
```
kchelstowski, line 46: nitpicky: I think this call should be placed in the `useOutboundListTableData` hook, but no need to change.

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
```diff
@@ -129,19 +131,20 @@ const StockMovementOutboundTable = ({
...
-      },
+      Cell: row => (
+        <TableCell
+          {...row}
+          tooltip
+          tooltipLabel={getStatusTooltip(row.value)}
+        >
+          <StatusIndicator
+            variant={row?.original?.statusVariant}
+            status={translate(`react.requisitionStatus.enum.${row?.original?.status}`, row?.original?.status)}
```
kchelstowski, line 144: please make sure that the label will change when you change the language - I remember having difficult times with `translate` behavior when changing the language on list pages - most of the times I was just deciding to translate what I needed on the backend.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,31 @@ class StockMovementApiController {
...
+
+        List<RequisitionStatus> statuses = getRequisitionStatusCodesList(currentLocation.isApprovalRequired(), isRequestsList)
+
+        List options = statuses?.collect(mapStatusCodes)
+        List allStatuses = RequisitionStatus.listAll().collect(mapStatusCodes)
+
+        render([data: options, allStatuses: allStatuses] as JSON)
+    }
+
+    List<RequisitionStatus> getRequisitionStatusCodesList(Boolean hasRequestApprovalActivity, Boolean isRequestsList) {
```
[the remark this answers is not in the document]
jmiranda: I might think about adding another enum `RequisitionStatusWorkflow` (??) to represent the different sub-workflows: RequestWorkflow, ApprovalWorkflow, PickPackWorkflow. Each status could have a workflow associated with it 
```
    CREATED(1, null, StatusType.SUCCESS, Workflow.REQUEST),
    EDITING(2, PENDING, StatusType.PRIMARY, Workflow.REQUEST),
    PENDING_APPROVAL(3, null, StatusType.WARNING, Workflow.APPROVAL),
    APPROVED(4, null, StatusType.SUCCESS, Workflow.APPROVAL),
    REJECTED(5, null, StatusType.DANGER, Workflow.APPROVAL),
    VERIFYING(6, PENDING, StatusType.WARNING, Workflow.REQUEST),
```
and you could pass a single workflow or a list of workflows to get the appropriate list back.
```
RequisitionStatus.listOptions(Workflow.ALL)
RequisitionStatus.listOptions(Workflow.REQUEST)
RequisitionStatus.listOptions([Workflow.REQUEST, Workflow.REQUEST_APPROVAL])
```
NOTE: I don't understand the need for listOutboundOptions. Could someone explain that to me?


Aside: This is one of the reasons I dislike the Stock Movement workflow so much. It's not just one workflow, it's anywhere from 3-5 different workflows, each with its own complex set of steps, that are being squashed into a single, barely usable flow.

grails-app/i18n/messages.properties
```diff
@@ -3481,6 +3484,22 @@ react.stockMovement.enum.AvailableItemStatus.PICKED=Picked
 react.stockMovement.enum.AvailableItemStatus.RECALLED=Recalled
 react.stockMovement.enum.AvailableItemStatus.HOLD=Hold
 react.stockMovement.enum.AvailableItemStatus.NOT_AVAILABLE=Not Available
+react.requisitionStatus.enum.CREATED=Created
```
jmiranda, line 3487: Can we stop duplicating these? I don't understand why the react app can just use the existing key
```
enum.RequisitionStatus.STATUS
```

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
alannadolny: I prefer this version + changes in naming that @jmiranda suggested, than moving to the function higher + using if/else. So, I am leaving this part of the code in the version that you suggested, unless there is a strong opinion against it.

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
alannadolny: Eh, you're right, it doesn't work

grails-app/i18n/messages.properties
alannadolny: As I know we were using only those labels starting with "react" on the react side. For now, when I use our function for fetching translations, I have to pass "requisitionStatus" as an argument to fetch all of the translations that start with "react.requistionStatus...".

I am moving the translation to the backend because those translations don't work as @kchelstowski suggested in a comment. So, I am removing those duplications.

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
```diff
@@ -108,6 +108,7 @@ class OutboundStockMovementListItem implements Serializable {
                 name                : name,
                 description         : description,
                 statusCode          : statusCode?.toString(),
+                statusVariant       : status?.variant?.name,
```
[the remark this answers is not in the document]
alannadolny: The variant here is an indicator that can be as an example: "warning", "success", etc. It is represented as a colorful dot on our list pages:
image

I had to move it here, because of the implementation that we had. This implementation doesn't allow us to have different options on the filtering dropdown, than the statuses on the stock movement list. In this ticket I have to remove "Verified" from the filtering when the location supports "request approval", but on the current list we can have "verified" status

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -807,14 +807,28 @@ class StockMovementApiController {
...
+        if (isRequestsList) {
+            return hasRequestApprovalActivity ? RequisitionStatus.listRequestOptionsWhenSupportingRequestApproval() : RequisitionStatus.listOutboundOptions()
         }
-        render([data: options] as JSON)
+
+        return hasRequestApprovalActivity ? RequisitionStatus.listOutboundOptionsWhenSupportingRequestApproval() : RequisitionStatus.listOutboundOptions()
+    }
+
+    Closure mapStatusCodes = { RequisitionStatus status ->
```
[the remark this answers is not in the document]
alannadolny: moved

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -157,6 +158,15 @@ enum RequisitionStatus {
             "end)"
     }
 
+    static Closure mapStatusCodes = { RequisitionStatus status ->
```
drodzewicz, line 161: I am wondering if we should specify what we are mapping to exactly.
Since we are mapping it to options maybe this should be called something like `mapToOption` or `toOption` and we can omit the "StatusCode" in the name since we are going to be calling it on RequisitionStatus so saying that we are mapping it to StatusCodes seems redundant, but that just my take.

To me this felts similar to what we are doing when mapping to JSON `.toJson()`

alannadolny: Good idea, thanks

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
```diff
@@ -103,12 +104,14 @@ class OutboundStockMovementListItem implements Serializable {
...
 
     Map toJson() {
+        def g = ApplicationHolder.application.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
         return [
                 id                  : id,
                 name                : name,
                 description         : description,
                 statusCode          : statusCode?.toString(),
                 statusVariant       : status?.variant?.name,
+                statusLabel         : "${g.message(code: 'enum.RequisitionStatus.' + status.name())}",
```
drodzewicz, line 114: Since `status` on `OutboundStockMovementListItem` is a **RequisitionStatus**, I feel like we can use the `mapStatusCodes()` _or whatever the name for it will be_, to extract the label from it.
I am not sure if it is going to work but lets try doing something like

```suggestion
                statusLabel         : status.mapStatusCodes()?.label,
```

drodzewicz (changes requested):

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
alannadolny: Yup, it works

drodzewicz (approved):

src/js/hooks/list-pages/outbound/useOutboundFilters.jsx
```diff
@@ -35,10 +35,12 @@ const useOutboundFilters = (isRequestsList) => {
   useEffect(() => {
     // TODO: When having full React, if once fetched, fetch only if a current language differs
     // TODO: from the language, that we were fetching this for
-    dispatch(fetchRequisitionStatusCodes());
     dispatch(fetchShipmentTypes());
   }, [currentLocale]);
 
+  useEffect(() => {
+    dispatch(fetchRequisitionStatusCodes(isRequestsList));
+  }, [currentLocale, currentLocation, isRequestsList]);
```
jmiranda, line 43: Should pass source type

jmiranda (approved):
Minor changes if necessary otherwise LGTM

awalkowiak (changes requested):

awalkowiak (approved):

PR #4291, opened 2023-09-21
OBPIH-5818 Approvers in request view page - Approvals view page

alannadolny:


drodzewicz (approved):

grails-app/views/stockMovement/show.gsp
```diff
@@ -206,6 +206,14 @@
                                 ${stockMovement?.stocklist?.name?:warehouse.message(code:"default.none.label")}
                             </td>
                         </tr>
+                        <tr class="prop">
+                            <td class="name">
+                                <warehouse:message code="stockMovement.approvers.label"/>
+                            </td>
+                            <td class="value">
+                                ${stockMovement?.requisition?.approvers?.join(", ") ?: warehouse.message(code:"default.none.label")}
```
kchelstowski, line 214:
```suggestion
                                ${stockMovement?.requisition?.approvers?.join(", ") ?: g.message(code:"default.none.label")}
```
let's use g.message, since we will want to get rid of warehouse.message

alannadolny:
@awalkowiak After working on a similar ticket to this one, I started thinking about hiding this row if the location doesn't support "request approval". Should it be hidden in this case or it should stay with just "none" as it is done right now?

grails-app/views/stockMovement/show.gsp
```diff
@@ -206,6 +206,14 @@
                                 ${stockMovement?.stocklist?.name?:warehouse.message(code:"default.none.label")}
                             </td>
                         </tr>
+                        <tr class="prop">
```
awalkowiak, line 209: Perhaps it would be good to hide this part if there is no approvers at all (this indicates it is not supported)

grails-app/views/stockMovement/show.gsp
```diff
@@ -206,6 +206,14 @@
...
+                            <td class="name">
+                                <warehouse:message code="stockMovement.approvers.label"/>
+                            </td>
+                            <td class="value">
+                                ${stockMovement?.requisition?.approvers?.join(", ") ?: g.message(code:"default.none.label")}
```
awalkowiak, line 214: Do you want to display all approvers? Perhaps it would be better to use "ellipsis" and display all only on hover, or `+x` for those that don't fit in space

awalkowiak (changes requested):

awalkowiak (approved):

PR #4293, opened 2023-09-22
OBPIH-5806 approval and reject action approvals end to end workflow

drodzewicz:


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -244,6 +245,49 @@ class StockMovementController {
         redirect(action: "list", params: params << ["flash": flash as JSON])
     }
 
+    def updateStatus = {
+        StockMovement stockMovement = stockMovementService.getStockMovement(params.id)
+
+        try {
```
kchelstowski, line 251: I really don't like throwing that try/catch everywhere - lately we've been doing that very much. Just wanted to mention that a potential rollback might not take place in Grails 1 when we catch exceptions.
Is it in requirements that after an exception we want to show the same view page?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -244,6 +245,49 @@ class StockMovementController {
...
+                default:
+                    break
+            }
+        } catch(Exception e) {
+            flash.message = e.message
+            redirect(action: "show", params: [id: stockMovement.id])
+            return
+        }
+
+        redirect(action: "list", params: ["flash": flash as JSON, sourceType: "ELECTRONIC", direction: "OUTBOUND"])
```
kchelstowski, line 278:
```suggestion
        redirect(action: "list", params: ["flash": flash as JSON, sourceType: RequisitionSourceType.ELECTRONIC, direction: StockMovementDirection.OUTBOUND.toUpperCase()])
```

grails-app/domain/org/pih/warehouse/core/User.groovy
```diff
@@ -96,6 +96,11 @@ class User extends Person {
         return roles.roleType.find { RoleType.listPrimaryRoleTypes().contains(it) }
     }
 
+    boolean hasRoles(Location location, List<RoleType> roleTypes) {
```
kchelstowski, line 99: maybe a comment above would be valuable?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -236,6 +236,13 @@ class UserService {
         return isSuperuser(currentUser) || (currentUser.getHighestRole(location) >= otherUser.getHighestRole(location))
     }
 
+    Boolean userHasRoles(String userId, Collection roleTypes, String locationId) {
```
kchelstowski, line 239: I see an inconsistency here - you return the method you've created in the domain, but the return types are different - here it is `Boolean` and in the domain it is `boolean` - I think it should be `boolean` here as well.

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -407,7 +414,11 @@ class UserService {
 
     public def getEffectiveRoles(User user) {
         def currentLocation = AuthService.currentLocation?.get()
-        return user.getEffectiveRoles(currentLocation)
+        return getEffectiveRoles(user, currentLocation)
+    }
+
+    public def getEffectiveRoles(User user, Location location) {
+        return user.getEffectiveRoles(location)
```
kchelstowski, line 421: I don't mind that, but was it necessary? I think the method above was clean enough?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -842,6 +842,24 @@ class RequisitionService {
...
+                return
+            } else if (status == RequisitionStatus.REJECTED) {
+                requisition.dateRejected = new Date()
+                requisition.rejectedBy = currentUser
+                requisition.status = RequisitionStatus.REJECTED
+
+                Event event = createEvent(EventCode.REJECTED, requisition.origin, new Date(), currentUser)
+                requisition.addToEvents(event)
+                requisition.save()
+                return
```
kchelstowski, line 862: I think this could be refactored. We have almost the same content in all of the statements, hence we could create a helper method to handle that, something like:
```groovy
Requisition transitionRequisitionStatus(Requisition requisition, RequisitionStatus requisitionStatus, EventCode eventCode, User currentUser) {
   requisition.dateApproved = new Date()
   requisition.approvedBy = currentUser
   requisition.status = requisitionStatus

   Event event = createEvent(eventCode, requisition.origin, new Date(), currentUser)
   requisition.addToEvents(event)
   requisition.save()
}
```
and here:
```groovy
if (status == RequisitionStatus.PENDING_APPROVAL) {
    if (requisition.origin.approvalRequired) {
        Event event = createEvent(EventCode.PENDING_APPROVAL, requisition.origin, new Date(), currentUser)
        requisition.addToEvents(event)
        requisition.status = RequisitionStatus.PENDING_APPROVAL
        requisition.approvalRequired = true
        requisition.save()
        return
    }
    Event event = createEvent(EventCode.SUBMITTED, requisition.origin, new Date(), currentUser)
    requisition.addToEvents(event)
    requisition.status = RequisitionStatus.VERIFYING
    requisition.approvalRequired = false
    requisition.save()
    return
} 
if (status == RequisitionStatus.APPROVED) {
   transitionRequisitionStatus(requisition, RequisitionStatus.APPROVED, EventCode.APPROVED, currentUser)
   return
}
if (status == RequisitionStatus.REJECTED) {
   transitionRequisitionStatus(requisition, RequisitionStatus.REJECTED, EventCode.REJECTED, currentUser)
   return
}
```
we could go even further and you could refactor also the statements I've made with one minor change:
```groovy
if (status == RequisitionStatus.PENDING_APPROVAL) {
    if (requisition.origin.approvalRequired) {
       Requisition updatedRequisition = transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
       updatedRequisition.approvalRequired = true
       // requisition.save() <--- notice I've commented that, it shouldn't be necessary anymore, because the event which has not been yet persisted has been already saved in the transition method, hence another save should not be necessary and the approvalRequired flag should be persisted at the end of the session, but please, double check that.
       return
    }
    Requisition updatedRequisition = transitionRequisitionStatus(requisition, RequisitionStatus.VERIFYING, EventCode.SUBMITTED, currentUser)
    requisition.approvalRequired = false
    return
} 
if (status == RequisitionStatus.APPROVED) {
   transitionRequisitionStatus(requisition, RequisitionStatus.APPROVED, EventCode.APPROVED, currentUser)
   return
}
if (status == RequisitionStatus.REJECTED) {
   transitionRequisitionStatus(requisition, RequisitionStatus.REJECTED, EventCode.REJECTED, currentUser)
   return
}
```

grails-app/views/stockMovement/show.gsp
```diff
@@ -1,6 +1,8 @@
 <%@ page import="org.pih.warehouse.requisition.RequisitionStatus; org.pih.warehouse.shipping.ShipmentStatusCode" %>
 <%@ page import="org.pih.warehouse.core.RoleType" %>
 <%@ page import="org.pih.warehouse.requisition.RequisitionSourceType" %>
+<%@ page import="org.pih.warehouse.inventory.StockMovementStatusCode" %>
+<%@page import="org.pih.warehouse.core.ActivityCode"%>
```
kchelstowski, line 5: nitpicky: 
```suggestion
<%@ page import="org.pih.warehouse.core.ActivityCode"%>
```

grails-app/views/stockMovement/show.gsp
```diff
@@ -75,78 +77,119 @@
...
-                <img src="${resource(dir: 'images/icons/', file: 'handtruck.png')}" />&nbsp;
-                <warehouse:message code="default.button.receive.label" />
-            </g:link>
-            <g:isUserAdmin>
-                <g:if test="${stockMovement?.hasBeenReceived() || stockMovement?.hasBeenPartiallyReceived()}">
-                    <g:link controller="partialReceiving" action="rollbackLastReceipt" id="${stockMovement?.shipment?.id}" class="button">
-                        <img src="${resource(dir: 'images/icons/silk', file: 'arrow_rotate_anticlockwise.png')}" />&nbsp;
-                        <warehouse:message code="stockMovement.rollbackLastReceipt.label" />
+            <g:set var="userHasRequestApproverRole" value="${false}"/>
+            <g:userHasRoles location="${stockMovement?.origin?.id}" roles="${[RoleType.ROLE_REQUISITION_APPROVER]}">
```
kchelstowski, line 81: I know @jmiranda is not a big fan of that, but I don't know if I see any other possible solution at the moment
But I like the fact, that you've actually checked it once, and assigned it to a variable, so we wouldn't have to call this `<g:userHasRoles>` in many places, like we do for `<g:isSuperuser>` in many places

grails-app/views/stockMovement/show.gsp
```diff
@@ -75,78 +77,119 @@
...
-                            <g:link class="button" controller="stockTransfer" action="remove" id="${stockMovement?.id}" params="[orderId: stockMovement?.order?.id]"
-                                    onclick="return confirm('${warehouse.message(code: 'default.button.delete.confirm.message', default: 'Are you sure?')}');">
-                                <img src="${resource(dir: 'images/icons/silk', file: 'delete.png')}" />
-                                &nbsp;<warehouse:message code="default.button.delete.label" />
+                </g:else>
+                <g:link controller="partialReceiving" action="create" id="${stockMovement?.shipment?.id}" class="button">
+                    <img src="${resource(dir: 'images/icons/', file: 'handtruck.png')}" />&nbsp;
+                    <warehouse:message code="default.button.receive.label" />
+                </g:link>
+                <g:isUserAdmin>
```
kchelstowski, line 101: I built you up in the comment above for assigning it to a variable, but here I can see that you call the `<g:isUserAdmin>` a few times - maybe it'd be good from the performance POV to do it also in this place?

btw... I don't know if it'd be a good direction in the future, but why can't we just return user roles in the `model` while rendering the view?... it's a "loud thinking", but I'll leave this comment here in case anyone wanted to comment on that.

kchelstowski (changes requested):
overall good job, those are only "beautify" suggestions

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: It was not mentioned if it should be the same page so it was my own interpretation.

It was mentioned that reject and approve buttons should be disabled if a request is already approved/rejected.
This try catch was more like a failsafe I added so I think I can remove it and let the generic handler do it's job and render generic error page.

grails-app/services/org/pih/warehouse/core/UserService.groovy
drodzewicz: yes, it should be `boolean`

grails-app/services/org/pih/warehouse/core/UserService.groovy
drodzewicz: If you are wondering why I split the `getEffectiveRoles()` methods, it is because previously it defaulted to current location and I needed roles for a user in origin location so I added an additional method which  receives a location parameter.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
drodzewicz: It would be great to move some common logic into some other method but the reason why I didn't do it at first was that there a slight differences in each of the status changes like in example of a APPROVED we assign `approvedBy` but in case of the REJECETD we must assign `rejectedBy`.
These are small differences so a common method should work but I'll think about if there is a better implementation for all of this

grails-app/services/org/pih/warehouse/core/UserService.groovy
kchelstowski: ok, I did not notice that there was a currentLocation :+1:

grails-app/views/stockMovement/show.gsp
drodzewicz: this is old code whcih github highlighted as a new change since I wrapped a  `<g:if`> and `<g:userHasRoles>` around it

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
kchelstowski: then you can just follow the path I've shown with my statements with the `approvalRequired` - so just from this method return the common stuff, and then just operate on it like:
```groovy
Requisition updatedRequisition = transitionRequisitionStatus(...)
```

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -244,6 +245,49 @@ class StockMovementController {
...
+                case RequisitionStatus.REJECTED:
+                    flash.message = g.message(
+                            code: "request.rejected.message.label",
+                            default: "You have successfully rejected the request {0}",
+                            args: [stockMovement.id]
+                    )
+                    break
+                default:
+                    break
+            }
```
alannadolny, line 271: Maybe we can shorten it to, but if you don't like this solution, let's leave it as is
```
if (stockMovement.requisition?.status in [RequisitionStatus.APPROVED, RequisitionStatus.REJECTED]) {
   String status = stockMovement.requisition?.status?.toLowerCase()   
   flash.message = g.message(
                            code: "request.${status}.message.label",
                            default: "You have successfully ${status} the request {0}",
                            args: [stockMovement.id]
                    )
}
```

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -825,23 +825,34 @@ class RequisitionService {
...
-                requisition.status = RequisitionStatus.VERIFYING
+                transitionRequisitionStatus(requisition, RequisitionStatus.VERIFYING, EventCode.SUBMITTED, currentUser)
                 requisition.approvalRequired = false
-                requisition.save()
+            } else if (requisition.status == RequisitionStatus.APPROVED) {
+                transitionRequisitionStatus(requisition, RequisitionStatus.APPROVED, EventCode.APPROVED, currentUser)
+                requisition.dateApproved = new Date()
+                requisition.approvedBy = currentUser
+                return
+            } else if (requisition.status == RequisitionStatus.REJECTED) {
```
kchelstowski, line 851: nitpicky: `else` is not required if you call `return` above (take a look at my suggestion below)

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -236,6 +236,13 @@ class UserService {
         return isSuperuser(currentUser) || (currentUser.getHighestRole(location) >= otherUser.getHighestRole(location))
     }
 
+    boolean userHasRoles(String userId, Collection roleTypes, String locationId) {
```
jmiranda, line 239: Use `hasRoles` as the name here too. Although, I'd also be fine with 
* doesUserHaveRoles 
* isUserInRoles
* isUserInAllRoles
* isUserInAnyRole

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -825,23 +825,34 @@ class RequisitionService {
...
+        requisition.status = requisitionStatus
+        Event event = createEvent(eventCode, requisition.origin, new Date(), currentUser)
+        requisition.addToEvents(event)
+        requisition.save()
+    }
+
     void triggerRequisitionStatusTransition(Requisition requisitionInstance, User currentUser) {
         Requisition.withSession {
             Requisition requisition = Requisition.get(requisitionInstance.id)
             if (requisition.status == RequisitionStatus.PENDING_APPROVAL) {
```
jmiranda, line 838: What happens if we call triggerRequisitionStatusTransition() multiple times with PENDING_APPROVAL, APPROVED, REJECTED? Is there any protection against that?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
jmiranda: I'd rather remove the return than the else.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
jmiranda: I would also be happy with a switch statement for the top level status check 

```
switch (status) { 
    case PENDING_APPROVAL: 
        doPendingApprovalStuff()
        return
    case APPROVED: 
        transitionRequisitionStatus(requisition, RequisitionStatus.APPROVED, EventCode.APPROVED, currentUser)
        return
    ...  
}
```

jmiranda: > requisition.save() <--- notice I've commented that, it shouldn't be necessary anymore, because the event which has not been yet persisted has been already saved in the transition method, hence another save should not be necessary and the approvalRequired flag should be persisted at the end of the session, but please, double check that.

Not sure if that'll be true on Grails 3.

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -88,6 +88,12 @@ class AuthTagLib {
             out << body()
     }
 
+    def userHasRoles = {attrs, body ->
```
jmiranda, line 91: Again, use a boolean verb here.

jmiranda: Actually this should probably be ifUserHasRoles since that's the behavior we seem to be implementing with this tag, right?

grails-app/views/stockMovement/show.gsp
jmiranda: Yeah, I hate this paradigm, but I think I'm the one who made it popular so I don't blame @drodzewicz.

What is interesting is that the <g:set /> actually gets invoked with that. I guess that makes sense, but I assumed g:if did something special. I guess it's just that GSP is rendered/invoked recursively so whether the next tag is rendering or setting doesn't really matter ... it just gets invoked. But it would be great  if someone investigated the Grails tab library where "g:if" is defined to see what they do in this case.

And as I said above, I would prefer `ifUserHasRoles` over `userHasRoles`.

jmiranda: It probably needs to be done as a tech debt refactoring but we probably should have thought about converging our permission/supports taglibs with something closer to Spring Security (like using `authorize` with logic expressions).

grails-app/views/stockMovement/show.gsp
jmiranda: Can someone add to the tech debt backlog an investigation into the right way to do these types of tag libs? It should be under Security and Performance. We invoke things like "isInRole" all the time throughout the code and I know it degrades performance. But what I don't understand is how Spring Security gets away with it. Do all of the security invocations go through the thread local?

grails-app/views/stockMovement/show.gsp
```diff
@@ -75,78 +77,119 @@
...
+                    <g:if test="${stockMovement?.hasBeenReceived() || stockMovement?.hasBeenPartiallyReceived()}">
+                        <g:link controller="partialReceiving" action="rollbackLastReceipt" id="${stockMovement?.shipment?.id}" class="button">
+                            <img src="${resource(dir: 'images/icons/silk', file: 'arrow_rotate_anticlockwise.png')}" />&nbsp;
+                            <warehouse:message code="stockMovement.rollbackLastReceipt.label" />
+                        </g:link>
+                    </g:if>
+                    <g:elseif test="${stockMovement?.hasBeenIssued() || ((stockMovement?.hasBeenShipped() ||
+                            stockMovement?.hasBeenPartiallyReceived()) && stockMovement?.isFromOrder)}">
+                        <g:link controller="stockMovement" action="rollback" id="${stockMovement.id}" class="button">
```
jmiranda, line 110: @drodzewicz Are you saying that all of the green code is old? If not we definitely need to add another tech debt item for cleaning up our logic because this is getting out of hand.

At the very least we need to push more of this logic into the stock movement.

grails-app/views/stockMovement/show.gsp
```diff
@@ -75,78 +77,119 @@
...
+                            controller="stockRequest"
+                            action="updateStatus"
+                            id="${stockMovement.id}"
+                            params="[status: StockMovementStatusCode.APPROVED]"
+                            disabled="${!stockMovement.pendingApproval}"
+                            disabledMessage="Request is not pending approval"
                     >
-                        <img src="${resource(dir: 'images/icons/silk', file: 'delete.png')}" />&nbsp;
-                        <warehouse:message code="default.button.delete.label" />
+                        <img src="${resource(dir: 'images/icons/silk', file: 'accept.png')}" />&nbsp;
```
jmiranda, line 169: Is this the only place we use this icon on this page? If so accept.png (approved) and decline.png (rejected) should work perfectly. Otherwise, we might want to consider using a different pair (like tick.png, cross.png).

grails-app/views/stockMovement/show.gsp
```diff
@@ -75,78 +77,119 @@
...
+                        <g:message code="request.approval.approve.label"  default="approve" />
```
jmiranda, line 170: Why lowecase?

jmiranda (commented):
LGTM. I can't remember if I requested any changes, so I'll leave this as a comment for now.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -244,6 +245,37 @@ class StockMovementController {
...
+            String status = stockMovement.requisition?.status
+            flash.message = g.message(
+                    code: "request.${status}.message.label",
+                    default: "You have successfully ${status} the request {0}",
+                    args: [stockMovement.id]
+            )
+        }
+
+        redirect(action: "list", params: [
+                "flash"     : flash as JSON,
```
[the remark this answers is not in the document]
drodzewicz: honestly, I am not sure my self why I did this that way.
I just need to pass `message` string as a parameter, I don;t need to assign i t o `flash.message` and then cast it as JSON.

I probably looked at the example above for `def remove` which does this exact thing and just copy & pasted it for `updateStatus`

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
drodzewicz: yes

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
drodzewicz: I like the switch case a lot more

grails-app/views/stockMovement/show.gsp
drodzewicz: by pushing more of this logic to stock Movement do you mean pushing authorization logic there, like eg. `stockMovement.approverAuhtorized(user)` `stockMovement.canEddit(user)`?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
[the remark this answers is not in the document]
awalkowiak: @jmiranda This is our workaround to get messages from "grails" pages to "react" pages (when it is done through a redirect). This won't be needed if, in some bright future, we separate frontend from backend (read: we finally get rid of gsp templates and server-side rendering).

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -825,23 +825,38 @@ class RequisitionService {
...
-                    requisition.approvalRequired = true
-                    requisition.save()
-                    return
-                }
-                Event event = createEvent(EventCode.SUBMITTED, requisition.origin, new Date(), currentUser)
-                requisition.addToEvents(event)
-                requisition.status = RequisitionStatus.VERIFYING
-                requisition.approvalRequired = false
-                requisition.save()
+            switch(requisition.status) {
```
awalkowiak, line 838: @drodzewicz Perhaps a comment here would be handy? Please add a note here about why we have various options for status transition here, with a link to request approvals epic ticket.

grails-app/taglib/org/pih/warehouse/AuthTagLib.groovy
```diff
@@ -88,6 +88,12 @@ class AuthTagLib {
             out << body()
     }
 
+    def ifUserHasRoles = {attrs, body ->
```
awalkowiak, line 91: Please for now follow here the convention of `isUserInRole` and make this one `isUserInAllRoles` instead `ifUserHasRoles`

awalkowiak (changes requested):
@drodzewicz two small requests from my side, other than that I think it LGTM

awalkowiak (approved):

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
kchelstowski: @jmiranda 

> Not sure if that'll be true on Grails 3.

yes, it will, because it's Hibernate's mechanism, not Grails'

PR #4299, opened 2023-09-26
OBPIH-5822 email alert to requestor approvals email alerts

drodzewicz:


grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -209,6 +210,12 @@ class Requisition implements Comparable<Requisition>, Serializable {
         approvalRequired(nullable: true)
     }
 
+    Comment getRecentComment() {
+        return comments?.sort({ a, b ->
+            b.dateCreated <=> a.dateCreated
+        }).find()
```
kchelstowski, line 216: wouldn't `.first()` fit better here?
I believe `find()` should take some closure as an argument? I'm surprised that it works without anything

grails-app/i18n/messages.properties
```diff
@@ -691,6 +691,8 @@ email.contactAdmin.message=Please contact your local OpenBoxes administrator wit
 email.contactAdminReceiving.message=Please contact the receiver listed above or your local OpenBoxes administrator with any questions you have about the status of this item. If you see a quantity in the cancelled column, this means that the warehouse received fewer items than expected, possibly as a result of a data error or a loss during shipment. The receiver can explain the situation. If you have an OpenBoxes account, you can view the shipment via
 email.thisLink.label=this link
 email.requestReceived.message=You have received a new request from: {0}, created by: {1}. Click <a href="{2}">here</a> to review your requests pending your approval.
+email.statusChange.message=Your request status has been updated to: {0}. Click <a href="{1}">here</a> to review your requests.
+email.withComment.message=With comment:
```
kchelstowski, line 695: I know @jmiranda was not a big fan or adding some `:` or commas in the message itself, but I'm fine with that.

grails-app/views/email/_approvalsStatusChanged.gsp
```diff
@@ -0,0 +1,60 @@
...
+<div>
+    <div class="header">
+        <g:render template="/email/header"/>
+    </div>
+    <div>
+        <g:message code="email.statusChange.message" args="[format.metadata(obj:requisition.status), g.createLink(uri: '/', absolute: true)]" />
+    </div>
+    <div>
+        <g:message code="email.withComment.message" />
+        <g:set var="comment" value="${requisition.recentComment}" />
```
kchelstowski, line 16: i assume we can't run into a `NullPointerException` here, because it would not be rendered if we don't have requisitoin?

grails-app/i18n/messages.properties
drodzewicz: You have a point, this should probably not be in the message label

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -245,10 +245,11 @@ class NotificationService {
     }
 
     void publishRequisitionStatusTransitionNotifications(Requisition requisition) {
-        List<Person> recipients = resolveNotificationRecipients(requisition)
-
         if (requisition.status == RequisitionStatus.PENDING_APPROVAL) {
+            List<Person> recipients = resolveNotificationRecipients(requisition)
             publishRequisitionPendingApprovalNotifications(requisition, recipients)
+        } else if (requisition.status in [RequisitionStatus.APPROVED, RequisitionStatus.REJECTED]) {
+            publishRequisitionStatusUpdateNotifications(requisition, requisition.requestedBy)
```
alannadolny, line 252: I am not a big fan of using else, maybe this approach would be cleaner:
```groovy
void publishRequisitionStatusTransitionNotifications(Requisition requisition) {
            List<Person> recipients = resolveNotificationRecipients(requisition)
            publishRequisitionPendingApprovalNotifications(requisition, recipients)
            return
        }

        if (requisition.status in [RequisitionStatus.APPROVED, RequisitionStatus.REJECTED]) {
            publishRequisitionStatusUpdateNotifications(requisition, requisition.requestedBy)
        }
    }
```

drodzewicz: instead I went with an approach of switch/case

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
awalkowiak: @drodzewicz I think you still need to have here a null-safety operator with `find()` too.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4302, opened 2023-09-28
OBPIH-5805 Approver cannot edit requests - Approvals end to end workflow

alannadolny:
I asked Manon about the designated behavior of the edit options:
1. When we are on the view page - after clicking edit we are redirected to the add items view page of the selected request, but this view is read-only. The "edit items" button is not visible.
2. When we are on the requests list - the edit option on the dropdown for the selected requests is not visible.

The edit option is not visible when:
1. Our location supports requests approval - request can be edited when it is not in approved/rejected status. It can be edited only by the requestor.
2. Our location doesn't support request approval - it should stay as it was before, every person can edit every request (restrictions are only based on locations)

### DIFF
--- a/src/js/components/stock-movement-wizard/request/EditPage.jsx
+++ b/src/js/components/stock-movement-wizard/request/EditPage.jsx
@@ -23,6 +23,7 @@ import TableRowWithSubfields from 'components/form-elements/TableRowWithSubfield
 import TextField from 'components/form-elements/TextField';
 import DetailsModal from 'components/stock-movement-wizard/modals/DetailsModal';
 import SubstitutionsModal from 'components/stock-movement-wizard/modals/SubstitutionsModal';
+import { canEditRequest } from 'components/stock-transfer/utils';
 import RequisitionStatus from 'consts/requisitionStatus';
 import apiClient from 'utils/apiClient';
 import { renderFormField } from 'utils/form-utils';
@@ -1512,14 +1513,23 @@ class EditItemsPage extends Component {
    */
   isUserAllowedToAddItems() {
     const { requestedBy, origin, destination } = this.state.values;
-    const { currentUserId, currentLocationId } = this.props;
-    if (currentLocationId === origin?.id) {
+    const { currentUser, currentLocation } = this.props;
+    if (currentLocation?.id === origin?.id) {
       return true;
     }
-    return currentLocationId === destination?.id && requestedBy?.id === currentUserId;
+    return currentLocation?.id === destination?.id && requestedBy?.id === currentUser?.id;
   }
 
   isAddingItemsAllowed() {
+    const { currentUser, currentLocation } = this.props;
+
+    if (
+      this.state.values?.isElectronicType &&
+      !canEditRequest(currentUser, this.state.values, currentLocation)
+    ) {
+      return false;
+    }
+
     const allowedStatuses = [
       RequisitionStatus.CREATED,
       RequisitionStatus.EDITING,
@@ -1676,8 +1686,8 @@ const mapStateToProps = state => ({
   pageSize: state.session.pageSize,
   supportedActivities: state.session.supportedActivities,
   currentLocale: state.session.activeLanguage,
-  currentUserId: state.session.user?.id,
-  currentLocationId: state.session.currentLocation?.id,
+  currentUser: state.session.user,
+  currentLocation: state.session.currentLocation,
 });
 
 export default withRouter(connect(mapStateToProps, {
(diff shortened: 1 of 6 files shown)


src/js/components/stock-movement-wizard/request/EditPage.jsx
```diff
@@ -1512,14 +1513,23 @@ class EditItemsPage extends Component {
...
     }
-    return currentLocationId === destination?.id && requestedBy?.id === currentUserId;
+    return currentLocation?.id === destination?.id && requestedBy?.id === currentUser?.id;
   }
 
   isAddingItemsAllowed() {
+    const { currentUser, currentLocation } = this.props;
+
+    if (
+      this.state.values?.isElectronicType &&
```
kchelstowski, line 1527: let's do:
```javascript
const { values } = this.state;
```
and use it like:
```javascript
values?.isElectronicType && !canEditRequest(currentUser, values, currentLocation)
```

src/js/components/stock-transfer/utils.jsx
```diff
@@ -94,3 +98,22 @@ export function prepareRequest(stockTransfer, status) {
...
+  if (!isRequestApprovalSupported) {
+    return true;
+  }
+
+  // If the location supports request approval, only the requestor is able to edit
+  // and the request couldn't be approved / rejected when editing
+  return isUserRequestor &&
+    request.statusCode !== RequisitionStatus.APPROVED &&
+    request.statusCode !== RequisitionStatus.REJECTED;
+};
```
kchelstowski, line 119: why is this util in `stock-transfer` utils? It doesn't fit there

src/js/components/stock-transfer/utils.jsx
```diff
@@ -94,3 +98,22 @@ export function prepareRequest(stockTransfer, status) {
...
+  const isRequestApprovalSupported = supports(
+    location?.supportedActivities,
+    ActivityCode.APPROVE_REQUEST,
+  );
+  const isUserRequestor = currentUser?.id === request?.requestedBy?.id;
+
+  // If request approval is not supported by the location we are able to edit every request
+  if (!isRequestApprovalSupported) {
+    return true;
+  }
```
kchelstowski, line 112: nitpicky, but since we don't need `isUserRequestor` at this point yet, we should move either this `if` above the `isUserRequestor` or the `isUserRequestor` below this `if`.

src/js/components/stock-transfer/utils.jsx
```diff
@@ -94,3 +98,22 @@ export function prepareRequest(stockTransfer, status) {
...
+
+  // If the location supports request approval, only the requestor is able to edit
+  // and the request couldn't be approved / rejected when editing
```
kchelstowski, line 115:
```suggestion
  // and the request can't be approved / rejected when editing
```

src/js/components/stock-movement-wizard/request/EditPage.jsx
```diff
@@ -1512,14 +1513,23 @@ class EditItemsPage extends Component {
...
   isUserAllowedToAddItems() {
     const { requestedBy, origin, destination } = this.state.values;
-    const { currentUserId, currentLocationId } = this.props;
-    if (currentLocationId === origin?.id) {
+    const { currentUser, currentLocation } = this.props;
+    if (currentLocation?.id === origin?.id) {
       return true;
     }
-    return currentLocationId === destination?.id && requestedBy?.id === currentUserId;
+    return currentLocation?.id === destination?.id && requestedBy?.id === currentUser?.id;
```
kchelstowski, line 1520: I think we should either move this part to the util you've created or move the content of your util here.
Your util also checks for the user's permissions and so does this util, so someone might be confused about the differences of those. Can we somehow make it more clear?
For example both here and in your util you check for the `requestedBy === currentUser` - this is another thing that could be maybe moved to a separate method or refactor it in a way, where this is checked only once.
```javascript
return currentLocation?.id === destination?.id && requestedBy?.id === currentUser?.id;
```

kchelstowski: I don't think it's even necessary to check the electronicType here, because this component is used clearly for the request workflow, so it must be electronicType - you won't edit the outbounds using this component. 
If you can, please spend a bit time on figuring out whether it is possible to somehow merge those methods (that are currently determining whether the "Edit request items" button should be visible or not).

It's nitpicky, but for me it feels weird that we have methods: `isAddingItemsAllowed` and `canEditRequest` separately. If we ever have to change anything in the future, we might be confused what the differences between them are.
If you don't figure out anything, we can try together.

awalkowiak (commented):
The majority looks good, but something feels weird with the utils, I'll come back to it tomorrow.

src/js/utils/permissionUtils.js
```diff
@@ -0,0 +1,45 @@
+import ActivityCode from 'consts/activityCode';
+import RequisitionStatus from 'consts/requisitionStatus';
+import { supports } from 'utils/supportedActivitiesUtils';
+
+const canEditRequestWithoutRequiredApproval = (location, origin, destination, isUserRequestor) => {
```
awalkowiak, line 5: What is the advantage of putting this in a separate function? I feel like this one could be directly in the `canEditRequest`. Plus looks like you are also using `isLocationOrigin` and `isLocationDestination` in both functions, so these consts could be moved to the top of the `canEditRequest` util and used when needed.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4305, opened 2023-09-29
OBPIH-5814 approvers column and filter in requests list page approvals lists pages

drodzewicz:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -61,6 +61,7 @@ class StockMovementApiController {
         stockMovement.requisitionStatusCodes = params.requisitionStatusCode ? params?.list("requisitionStatusCode") as RequisitionStatus[] : null
         stockMovement.requestType = params.requestType ? params.requestType as RequisitionType : null
         stockMovement.sourceType = params.sourceType ? params.sourceType as RequisitionSourceType : null
+        stockMovement.approvers = params.approver ? User.getAll(params.list("approver")) : null
```
kchelstowski, line 64: are we aware of the fact, that if somehow `User` of given id won't exist, it will fill the List with the `null` value? I assume Hibernate will ignore it when `createCriteria` is triggered, but just thought that if it doesn't, we can always just `findAll` by the identity (`it`), and it should filter those out

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -100,6 +100,13 @@ class OutboundStockMovementService {
                     }
                 }
             }
+            if (stockMovement.approvers) {
+                requisition {
+                    approvers {
+                        'in'("id", stockMovement.approvers.collect { it.id })
```
kchelstowski, line 106: I didn't see it when I was writing my comment above, but anyway, to be sure nothing unexpected will happen, please do it as `it?.id`, because when having `null` in your List created from `getAll`, you would get an NPE here.

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -143,6 +150,6 @@ class OutboundStockMovementService {
             }
         }
 
-        return stockMovements
+        return stockMovements.unique()
```
kchelstowski, line 153: this is not good in my opinion. Pagination might break at this point.

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
```diff
@@ -207,6 +207,19 @@ const StockMovementOutboundTable = ({
...
+      Header: <Translate id="react.stockMovement.request.approvers.label" defaultMessage="Approvers" />,
+      accessor: 'approvers',
+      minWidth: 250,
+      sortable: false,
+      Cell: row =>
+        (<TableCell
+          {...row}
+          tooltip
+          defaultValue="None"
+          value={row.value.map(catalog => catalog.name).join(', ')}
```
kchelstowski, line 220: I know it's probably copied :laughing: but please change it from `catalog` to `user`/`approver`

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
```diff
@@ -75,6 +82,7 @@ const useOutboundListTableData = (filterParams) => {
     if (!isRequisitionStatusesFetched || !requisitionStatuses.length) {
       dispatch(fetchRequisitionStatusCodes());
     }
+    dispatch(fetchAvailableApprovers());
```
kchelstowski, line 85: don't we want to refetch them if a location changes?

src/js/reducers/approversReducer.jsx
```diff
@@ -0,0 +1,25 @@
...
+  fetched: false,
+};
+
+export default function (state = initialState, action) {
+  switch (action.type) {
+    case FETCH_APPROVERS:
+      if (action.payload !== undefined) {
+        const users = _.map(action.payload, user => (
+          {
+            value: user.id, id: user.id, label: user.name, name: user.name,
```
kchelstowski, line 16: please let's do it as one property per line

src/js/reducers/approversReducer.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+const initialState = {
+  data: [],
+  fetched: false,
+};
+
+export default function (state = initialState, action) {
+  switch (action.type) {
+    case FETCH_APPROVERS:
+      if (action.payload !== undefined) {
```
kchelstowski, line 13: why are we strict to `=== undefined` instead of checking the truthiness?

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
drodzewicz: you're right

src/js/reducers/approversReducer.jsx
drodzewicz: This is something I also noticed in other reducers, but I am not sure what is it trying to solve.
When removing the `!== undefined` everything seems to work fine but maybe I am not aware of some case which this `if()` fixes.
It should be safe to remove it I guess

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
drodzewicz: you are right, this needs a different approach, unfortunately using `.distinctList()` does not work with paginated arguments such as `max` and `offset` so probably I'll need to fetch the `totalCount` separately

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
```diff
@@ -207,6 +207,19 @@ const StockMovementOutboundTable = ({
       sortable: false,
       Cell: row => (<TableCell {...row} defaultValue="None" />),
     },
+    {
+      Header: <Translate id="react.stockMovement.request.approvers.label" defaultMessage="Approvers" />,
+      accessor: 'approvers',
```
alannadolny, line 212: I think we shouldn't mix single quotation marks with double

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
alannadolny: Additionally, maybe we can merge those two methods into one? Each of them makes one iteration over values, it would be nice if we can reduce them to one iteration. I mean something like:
```
row.value?.reduce((acc, curr) => acc + catalog.name , "")
```
(wrote by hand, not sure if works)

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
drodzewicz: This is the eslint rule we have configured, we use single quotes for string values of an object and double quotes for attributes string values

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
drodzewicz: we need commas in between the items so your code would need to look at least like
```js
row.value?.reduce((acc, curr) => acc + `${catalog.name}, ` , "")
```
Additionally we are joining elements with a comma `, ` and in the code above we would always have commas at the end of the last element, and in case of a single element array it will also appear unless we do some kind of a `if(index is not lastElement)` logic.

My opinion is that we should not sacrifice code readability with trying to achieve peak performance, so I think it shouldn't be a big deal

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
kchelstowski: it's a pain in the neck, but I believe you can follow what I've suggested to @alannadolny once a long time ago, but I don't remember where it was. Probably something connected to products? Where we've built two createCriterias but one of them was just returning a `totalCount`, and one of them was returning the List.
@alannadolny do you remember where it was, so that @drodzewicz could follow the approach?

src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
alannadolny: I forgot about the last element with comma, so it looks good to me

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
alannadolny: Do you mean this pr? https://github.com/openboxes/openboxes/pull/3859

alannadolny (approved):

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -143,6 +159,18 @@ class OutboundStockMovementService {
             }
         }
 
+        def stockMovements = OutboundStockMovementListItem.createCriteria().list(max: max, offset: offset) {
+            query.delegate = delegate
+            query(false)
+        }
+
+        // Get result count
+        def stockMovementsCount = OutboundStockMovementListItem.createCriteria().get() {
```
awalkowiak, line 168: Why does this have to be split to get the paginated list and get the total count?

awalkowiak (commented):
Front LGTM, but I have a question about one thing on the backend

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
kchelstowski: @awalkowiak https://github.com/openboxes/openboxes/pull/4305#discussion_r1342349050

kchelstowski: pagination does not work well with distinct - if @drodzewicz used `stockMovements.unique()`, he would have less records than the `totalCount` would return. Another case is that there could be for example 7 records per page and stuff like that.
It is because approvers is `OneToMany`, so if a stockMovement has 3 approvers, it'd be returned three times from the query.

src/js/hooks/list-pages/outbound/useOutboundFilters.jsx
```diff
@@ -39,8 +41,13 @@ const useOutboundFilters = (sourceType) => {
   }, [currentLocale]);
 
   useEffect(() => {
+    dispatch(fetchAvailableApprovers());
```
kchelstowski, line 44: Won't it be triggered twice now when the location changes?
You have `currentLocation.id` as dependency in two `useEffects`.
I think `currentLocale` should not matter for the users to be refetched? So maybe remove it from here and leave it only in the 48-49 line?

drodzewicz: `currentLocale` dependency was there before and is probably needed for `fetchRequisitionStatusCodes` and `fetchAvailableApprovers` is not dependent on the currentLocale which is why I put it in a separate useEffect.

The reason why I used `currentLocation.id` instead of `currentLocation` was because it would trigger the useEffect more than one time. I did not go into detail about which exact part of the `currentLocation` object causes the useEffect to trigger but `currentLocation.id` should be the only reason that indicates that the location has changed.

drodzewicz: I did not notice the `dispatch(fetchAvailableApprovers());` above the comments, so you are right, I need to remove this.

kchelstowski (approved):

PR #4309, opened 2023-10-02
OBPIH-5811 Requestor dashboard statuses - Approvals statuses

alannadolny:
I am unsure if this is a correct approach - I created a function for checking if the status is in a list of newly created statuses from the approval workflow. Based on that I am getting an appropriate status, if the status is in the list, I am taking status from requisition, otherwise, I am taking shipment status.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -11,7 +11,6 @@ package org.pih.warehouse.inventory
 
 import grails.orm.PagedResultList
 import grails.validation.ValidationException
-import org.codehaus.groovy.grails.commons.ApplicationHolder
 import org.codehaus.groovy.grails.web.json.JSONObject
 import org.hibernate.ObjectNotFoundException
 import org.pih.warehouse.api.AvailableItem
@@ -57,7 +56,6 @@ import org.pih.warehouse.requisition.RequisitionType
 import org.pih.warehouse.shipping.Container
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentItem
-import org.pih.warehouse.shipping.ShipmentStatus
 import org.pih.warehouse.shipping.ShipmentStatusCode
 import org.pih.warehouse.shipping.ShipmentType
 import org.pih.warehouse.shipping.ShipmentWorkflow
@@ -2815,12 +2813,5 @@ class StockMovementService {
             validateQuantityRequested(stockMovement)
         }
         return true
-    }
-
-    static String getRequestStatusLabel(RequisitionStatus requisitionStatus, ShipmentStatus shipmentStatus) {
-        def g = ApplicationHolder.application.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
-        String status = RequisitionStatus.isApprovalStatus(requisitionStatus) ? requisitionStatus.name() : shipmentStatus.name
-        String enumName = RequisitionStatus.isApprovalStatus(requisitionStatus) ? "RequisitionStatus" : "ShipmentStatusCode"
-        return "${g.message(code: "enum.${enumName}." + status)}"
     }
 }
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -11,6 +11,7 @@ package org.pih.warehouse.inventory
 
 import grails.orm.PagedResultList
 import grails.validation.ValidationException
+import org.codehaus.groovy.grails.commons.ApplicationHolder
 import org.codehaus.groovy.grails.web.json.JSONObject
 import org.hibernate.ObjectNotFoundException
 import org.pih.warehouse.api.AvailableItem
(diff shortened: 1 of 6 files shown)


src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -167,6 +172,14 @@ enum RequisitionStatus {
         ]
     }
 
+    static String getRequestStatusLabel(RequisitionStatus requisitionStatus, ShipmentStatus shipmentStatus) {
```
awalkowiak, line 175: I think you should not mix together RequisitionStatus and ShipmentStatus here in the RequisitionStatus enum. I feel like it should rather be in StockMovementService instead of here. Plus I think you should do it a bit similarly to the way it's done in the FormatTagLib - long story short - take the `getClass().getSimpleName()`, and then pass it to `g.message`. Just decide which status to use, and return:
 `"${g.message(code: 'enum.' + status.getClass().getSimpleName() + '.' + status.name())}"`

awalkowiak (changes requested):

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
jmiranda: Yeah, I agree with @awalkowiak. If nothing else the logic should probably be in the StockMovement command. You probably need to add more states to the StockMovementStatusCode enum and map them in the fromStockMovementStatus and toStockMovementStatus methods.

src/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -86,6 +87,10 @@ enum RequisitionStatus {
         [CREATED, EDITING, APPROVED, PENDING_APPROVAL, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
+    static isApprovalStatus(RequisitionStatus status) {
+        return status in [APPROVED, REJECTED, PENDING_APPROVAL]
+    }
```
jmiranda, line 92: I would make this more readable and call it something like `isInApprovalWorkflow` or something like that.
```
static isInApprovalWorkflow(RequisitionStatus status) { 
    return status in RequsitionStatus.listApproval()
}

static listApproval() { 
    [APPROVED, REJECTED, PENDING_APPROVAL]
}
```

jmiranda: isInApprovalWorkflow doesn't sound right to me either, so if someone comes up with something better let me know.

src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -129,6 +129,7 @@ class StockMovement {
             name                : name,
             description         : description,
             statusCode          : statusCode,
+            statusLabel         : RequisitionStatus.getRequestStatusLabel(requisition?.status, shipment?.status),
```
jmiranda, line 132: If we end up keeping this, let's call this displayStatus (we use this in other places like Order and Shipment). 

FWIW, we already have displayStatusCode in StockMovementStatusCode enum as well. But I think this is used to "mask" the actual status from certain users. For example, whether a request is in REQUESTED or PICKING doesn't matter to the requestor ... the request is in a PENDING state. If we need to display the status to the user, then it seems we just need to add all of the approval workflow states to StockMovementStatusCode along with the appropriate displayStatusCode which is just the same status.
```
PENDING_APPROVAL(2, PENDING_APPROVAL),
APPROVED(2, APPROVED),
REJECTED(2, REJECTED),
```

jmiranda (commented):
See comments.

src/groovy/org/pih/warehouse/api/StockMovement.groovy
jmiranda: Lastly, in order to uncomplicate things (i.e. having requisition status and shipment status), I don't think we need to show SHIPPED as a status. For a requisition, it's just ISSUED.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2814,4 +2815,10 @@ class StockMovementService {
         }
         return true
     }
+
+    static String getRequestStatusLabel(Shipment shipment, Requisition requisition) {
```
awalkowiak, line 2819: This should be named `getDisplayStatus` or `getStockMovementDisplayStatus`

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2814,4 +2815,23 @@ class StockMovementService {
         }
         return true
     }
+
+    static String getDisplayStatus(Shipment shipment, Requisition requisition) {
```
awalkowiak, line 2819: @alannadolny Actually now when I see this method, I think it should be directly in the StockMovement instead of here (if you only use it in the `StockMovement` - initially I thought it might end up being used in the `OutboundStockMovement` too), and you won't have to pass shipment and requisition in params

awalkowiak (approved):

PR #4310, opened 2023-10-03
OBPIH-5824 add a comment approvals comment

drodzewicz:


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -33,6 +35,8 @@ import org.pih.warehouse.requisition.RequisitionStatus
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentStatusCode
```
alannadolny, line 37: Is this line break done on purpose?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+        }
+        Comment comment = Comment.get(params?.id)
+        if (!comment) {
+            flash.message = "${g.message(code: 'default.not.found.message', args: [g.message(code: 'comment.label', default: 'Comment'), params.id])}"
+            redirect(action: "show", id: stockMovement?.id)
+            return
+        }
+
+        if (comment.sender.id != session.user.id) {
+            throw new AccessDeniedException("You are not authorized to perform this action")
```
awalkowiak, line 337: I think we usually used 
`throw new UnsupportedOperationException("${warehouse.message(code: 'errors.noPermissions.label')}")`
in such cases

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+        }
+
+        stockMovement.requisition?.removeFromComments(comment)
+        if (!stockMovement.requisition.hasErrors() && stockMovement.requisition?.save()) {
+            flash.message = "${g.message(code: 'default.comments.deleted.label')}"
```
awalkowiak, line 342: Well I think you technically not deleted the comment yet (you only removed it from requisition so far)

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+            redirect(action: "show", id: stockMovement.id)
+            return
+        }
+        render(view: "show", model: [stockMovement: stockMovement])
+    }
+
+    def saveComment = {
+        def stockMovement = getStockMovement(params['stockMovement.id'])
+        if (!stockMovement) {
```
awalkowiak, line 351: I feel like this check makes no sense here. You don't have to validate the requisition, to make changes on the comment.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+            redirect(action: "list")
+            return
+        }
+        Comment comment = Comment.get(params?.id)
+        if (comment) {
+            if (comment.sender.id != session.user.id) {
+                throw new AccessDeniedException("You are not authorized to perform this action")
+            }
+            comment.properties = params
+            if (!stockMovement.requisition.hasErrors() && stockMovement.requisition.save()) {
```
awalkowiak, line 362: Why you are not saving existing comments directly but you are doing it through requisition (without flush)?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+                flash.message = "${g.message(code: 'default.comments.updated.label')}"
+                redirect(action: "show", id: stockMovement.id)
+                return
+            }
+        } else {
+            comment = new Comment(params)
+            stockMovement.requisition.addToComments(comment)
+            if (!stockMovement.requisition.hasErrors() && stockMovement.requisition.save()) {
```
awalkowiak, line 370: Should it have a `flush`? I think this controller is not transacitonal

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+
+    def addComment = {
+        def stockMovement = getStockMovement(params.id)
+        List<User> recipients = User.list()
+        return [stockMovement: stockMovement, comment: new Comment(), recipients: recipients]
+    }
+
+    def editComment = {
+        def stockMovement = getStockMovement(params['stockMovement.id'])
+        if (!stockMovement) {
```
awalkowiak, line 303: Is this check required?

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+            redirect(action: "list")
+            return
+        }
+        Comment comment = Comment.get(params?.id)
+        if (!comment) {
+            flash.message = "${g.message(code: 'default.not.found.message', args: [g.message(code: 'comment.label', default: 'Comment'), comment.id])}"
+            redirect(action: "show", id: stockMovement?.id)
+            return
+        }
+        if (comment.sender.id != session.user.id) {
```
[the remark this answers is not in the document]
kchelstowski: theoretically a sender is nullable property, so it might be `null`, but should not.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: I agree, and I think this should be handled by the `all-delete-orphan` in the Requisition domain.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: I'd say both saves are not necessary if a Comment already exists. Calling save explicilty is only necessary if an entity is not yet persisted, but I don't mind validating it here.
I agree this is weird that it is done through the Requisition. I don't remember if we got rid of disabling the cascade validation (`deepValidation` problem), but if we had not, this would not validate the comment anyway if you did it through the requisitoin.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: in my opinion most of this method should be a service method, especially the save calls.
Having this code, it would:

- break in Grails 3 when using flush immedietaly in this line (no transaction in progress error)
- break "silently" (or at the end of the session) in Grails 3 when not using flush (the error would happen "in the background")
- it will work fine in Grails 1 due to the OSIV

so it should be refactored in my opinion and at least moved to the service.

Moreover I think we used to use `update` method when we update an entity, and when we create new, we used to use `create` method, don't we?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
     }
 
+    def comments = {
+        def stockMovement = getStockMovement(params.id)
+        render(template: "comments", model: [stockMovement: stockMovement])
+    }
+
+    def addComment = {
+        def stockMovement = getStockMovement(params.id)
+        List<User> recipients = User.list()
```
kchelstowski, line 297: do we return all users on purpose?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+        return [stockMovement: stockMovement, comment: new Comment(), recipients: recipients]
```
kchelstowski, line 298: does it work fine with `return`? Most of the times for gsp templates we tend to either do `render` or `redirect`. I don't remind myself of seeing `return`.
Unless it is used in some of the methods below and I just don't see the context, then I'm fine with that.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: see the comment below - this should be at least moved to the service layer not to experience any problems in Grails 3.

grails-app/views/stockMovement/show.gsp
```diff
@@ -548,6 +554,13 @@
                             <warehouse:message code="documents.label" default="Documents"/>
                         </a>
                     </li>
+                    <g:if test="${stockMovement?.requisition}">
+                        <li>
+                            <a href="${request.contextPath}/stockMovement/comments/${stockMovement?.id}">
```
kchelstowski, line 559: can't we just use `<g:link>` with proper params here?

kchelstowski (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+        if (comment.sender.id != session.user.id) {
+            flash.message = "${g.message(code: 'auth.notAuthorized.message')}"
+            redirect(action: "show", id: stockMovement?.id)
+            return
+        }
+        render(view: "addComment", model: [stockMovement: stockMovement, comment: comment])
+    }
+
+    def deleteComment = {
+        def stockMovement = getStockMovement(params['stockMovement.id'])
```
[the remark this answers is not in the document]
jmiranda: why aren't we calling stockMovementService.getStockMovement()

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: I don't mind the check, but think the closure should render an error to the page / modal that is being rendered. Same with when the comment is not found.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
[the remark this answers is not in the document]
jmiranda: Same with this one. Render the error in the modal. Even if that's all that gets rendered.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: AccessDeniedException is actually a file io exception so this is no bueno anyway.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: @kchelstowski To clarify, you mean we should only use `comment.delete()` and let the all-delete-orphan handle removing the comments from the association, correct?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: For updating a comment, we should be fine with 
```
comment.save()
```
For adding a comment, we should only need to add the comment to the comments association on Requisition and save the requisition.
```
requisition.save()
```

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: Agree wholeheartedly with the "move to service layer" suggestion.

jmiranda (commented):
See comments.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: I'd say if we update a comment it should not be even necessary to call the `comment.save()` (but I don't mind doing so), because it's already persisted and hibernate would save it at the end of the session. 
As for second statement, I agree, it would be enough, but if we use the cascade.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: @jmiranda I meant that doing `stockMovement.requisition.removeFromComments(comment)` should be enough if we use the orphan removal.
Setting orphan on the comment would mean, that a comment's existence "does not have sense" if it doesn't have requisition associated, so it's gonna be deleted whenever we unassign a comment from a requisition.

tl;dr - calling `comment.delete()` is not neccesary if we use the `all-delete-orphan`. `requisition.removeFromComments` should do the job.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -33,6 +35,8 @@ import org.pih.warehouse.requisition.RequisitionStatus
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentStatusCode
 
+import java.nio.file.AccessDeniedException
```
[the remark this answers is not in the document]
drodzewicz: understood

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: As an example, I was following the comments on Purchase Orders which gives you a list of Users as comment receivers.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: in all cases an error is being rendered.
- if stockMovement doesn't;t exist we redirect to the list page and display an error
- if comment doesn't exist we redirect to stockMovement view page and display an error
- and if user is not a sender then redirect to view page and display an error 

in all cases an error is rendered to a user.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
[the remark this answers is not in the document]
drodzewicz: @alannadolny I used `def` because when getting a stockMovement using `getStockMovement(...)`
we can receive either `StockMovement` or `OutboundStockMovement` types because we have such logic implemented in the function
```groovy
  if (order) {
      stockMovement = outboundStockMovementService.getStockMovement(stockMovementId)
  } else {
      stockMovement = stockMovementService.getStockMovement(stockMovementId)
  }
```
but since we have comments only on requests I think I can get away with getting the stockMovement using `stockMovementService.getStockMovement()`

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: you are right, `Requisiton` is missing `cascade: "all-delete-orphan"`

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: @kchelstowski the cascade validate problem we encountered was on grails 3. DeepValidate should work on G1

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,95 @@ class StockMovementController {
...
+        }
+        Comment comment = Comment.get(params?.id)
+        if (comment) {
+            if (comment.sender.id != session.user.id) {
+                throw new AccessDeniedException("You are not authorized to perform this action")
+            }
+            comment.properties = params
+            if (!stockMovement.requisition.hasErrors() && stockMovement.requisition.save()) {
+                flash.message = "${g.message(code: 'default.comments.updated.label')}"
+                redirect(action: "show", id: stockMovement.id)
```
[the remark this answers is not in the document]
drodzewicz: shoulnd't be necessary since I am checking if the given StockMovement exists.
If it doesn't, we shouldn't even be at this point in the code.

grails-app/views/stockMovement/show.gsp
drodzewicz: I did not give it much attention and copy pasted the lines from above which use anchor tags.
I can refactor that.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2814,4 +2814,18 @@ class StockMovementService {
...
+    void deleteComment(Comment comment, Requisition requisition) {
+        requisition.removeFromComments(comment)
+    }
+
+    void saveComment(Comment comment) {
+        comment.save(failOnError: true)
+    }
+
+    void addCommentToRequisition(Comment comment, Requisition requisition) {
+        comment.save(failOnError: true)
```
kchelstowski, line 2827: I think this is unnecessary - the `requisition.save()` should handle this by cascade - maybe instead of `failOnError` let's do a `validate()` and then eventually return `comment.errors` to the view instead of throwing a `ValidationException`?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2814,4 +2814,18 @@ class StockMovementService {
         }
         return true
     }
+
+    void deleteComment(Comment comment, Requisition requisition) {
+        requisition.removeFromComments(comment)
+    }
+
+    void saveComment(Comment comment) {
```
kchelstowski, line 2822: I'm wondering if it should not return the `Comment`, so to be `Comment saveComment(Comment comment)` - usually persist methods return an entity, not void.
I know that anyway your reference to the comment in the controller will handle the eventual "persistence changes', but just saying.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
awalkowiak: The best case would be to use async select for this one, or I think at least you could use the `selectUser` tag (which takes only active users, I feel like there is no point in allowing selecting inactive users)

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,101 @@ class StockMovementController {
         render(template: "documents", model: [stockMovement: stockMovement])
     }
 
+    def comments = {
+        def stockMovement = getStockMovement(params.id)
```
kchelstowski, line 290: what's the difference in using that and using the one from the `stockMovementService`? Is there any case that you use this one instead of the one from the service?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,101 @@ class StockMovementController {
...
+
+    def saveComment = {
+        StockMovement stockMovement = stockMovementService.getStockMovement(params['stockMovement.id'])
+        if (!stockMovement) {
+            flash.message = "${g.message(code: 'default.not.found.message', args: [g.message(code: 'stockMovement.label', default: 'StockMovement'), params['stockMovement.id']])}"
+            redirect(action: "list")
+            return
+        }
+        Comment comment = new Comment(params)
+        stockMovementService.saveComment(comment)
```
kchelstowski, line 350: it is not needed, because the comment would be save by cascade in the method below `addCommentToRequisition`.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,101 @@ class StockMovementController {
...
+        stockMovementService.addCommentToRequisition(comment, stockMovement.requisition)
+         if (comment?.hasErrors()) {
```
kchelstowski, line 352: shouldn't we revert the statement and do something like:
```groovy
if (comment.validate()) {
  stockMovementService.addCommentToRequisition(comment, stockMovement.requisition)
  flash.message = "${g.message(code: 'default.comments.created.label')}"
  redirect(action: "show", id: stockMovement.id)
  return
}
stockMovement = stockMovementService.getStockMovement(params['stockMovement.id'])
render(view: "addComment", model: [stockMovement: stockMovement, comment: comment])
```

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,101 @@ class StockMovementController {
...
+            flash.message = "${g.message(code: 'default.not.found.message', args: [g.message(code: 'stockMovement.label', default: 'StockMovement'), params['stockMovement.id']])}"
+            redirect(action: "list")
+            return
+        }
+        Comment comment = Comment.get(params?.id)
+        if (comment.sender.id != session.user.id) {
+            throw new UnsupportedOperationException("${g.message(code: 'errors.noPermissions.label')}")
+        }
+        comment.properties = params
+        stockMovementService.saveComment(comment)
```
kchelstowski, line 373: since comment is a managed entity at this point, it is not necessary to save() it explicitly - we could (like I proposed above), just do:
```groovy
if (comment.validate()) {
  flash.message = "${g.message(code: 'default.comments.updated.label')}"
  redirect(action: "show", id: stockMovement.id)
  return
}
stockMovement = stockMovementService.getStockMovement(params['stockMovement.id'])
render(view: "addComment", model: [stockMovement: stockMovement, comment: comment])
```
but in this place, don't treat is as a "must change request".

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: There was one previous comment which addressed the difference and I had to change them to use `stockMovementService.getStockMovement`, must have missed this one because the rest are using `stockMovementService.getStockMovement`

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: I need to get the validation errors, so it's either I save it and get the error bean or do a validate on a comment inside addCommentToRequisition()
cascade save requisition did not return any errors fro me.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: sure

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
awalkowiak: @kchelstowski @drodzewicz Some time ago I added this utility method in this controller (while I was fixing packing list, documents and receipts tab for returns), but I do not remember why exactly
 ```groovy
// Used by SM show page 'tabs' actions - packing list, documents and receipts
def getStockMovement(String stockMovementId) {
    ...
}
```

Comment (awalkowiak) [reply grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy:290]: Looks like it was some sort of performance improvement (see https://pihemr.atlassian.net/browse/OBPIH-4544)

File: grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy (hunk 18 above)
Comment (kchelstowski) [reply grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy:350]: I knew that. This is why I suggested the `comment.validate()` in the comment below

File: grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy [hunk 21]
```diff
@@ -283,6 +286,101 @@ class StockMovementController {
         render(template: "documents", model: [stockMovement: stockMovement])
     }
 
+    def comments = {
+        def stockMovement = getStockMovement(params.id)
+        render(template: "comments", model: [stockMovement: stockMovement])
+    }
+
+    def addComment = {
+        StockMovement stockMovement = stockMovementService.getStockMovement(params.id)
```
awalkowiak, line 295: @drodzewicz now I started to think if this is a good idea performance-wise. If you should not use requisition here directly. For huge stock requests, this might end up being a bit slow (here and in every one of these actions)

awalkowiak (changes requested):

kchelstowski (commented):
I have one more nitpicky comment/suggestion - maybe it would be better to move those methods `saveComment`, `deleteComment` to `RequisitionService`, since it feels a bit weird to me that we call those methods with `stockMovement.requisition` argument.
What do you think guys?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
drodzewicz: Kacper made a good point about moving these comment methods to RequisitionService since comments are only attached to the requisition, so I'll probably do that and resolve this issue too.

drodzewicz: the problem is that I still need `stockMovement` to render the summary header 
```gsp
<g:render template="summary" model="[stockMovement:stockMovement]" />
```
unless it is not necessary then I think I can replace all of the stockMovements I use with requisition

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +287,98 @@ class StockMovementController {
...
+            throw new UnsupportedOperationException("${g.message(code: 'errors.noPermissions.label')}")
+        }
+        comment.properties = params
+        if (comment.validate()) {
+            requisitionService.saveComment(comment)
+            render(view: "addComment", model: [stockMovement: stockMovement, comment: comment])
+            return
+        }
+        flash.message = "${g.message(code: 'default.comments.updated.label')}"
+        redirect(action: "show", id: stockMovement.id)
```
jmiranda, line 378: I don't understand this flow. If the comment is valid, save it. Cool, but there's no handling of if the comment is not saved properly. And then if the comment is not valid, we still say it has been updated?

jmiranda: Also we need default namespace here unless we use it without the "comments" part.

```
flash.message = "${g.message(code: 'default.updated.message', args: [g.message(code: 'comment.label', default: 'Comment'), comment.id])}"
```

grails-app/i18n/messages.properties
```diff
@@ -353,6 +353,9 @@ default.color.label=Color
 default.comment.label=Comment
 default.comment.message=Enter a comment ...
 default.comments.label=Comments
+default.comments.created.label=Comment has been added
+default.comments.updated.label=Comment has been updated
+default.comments.deleted.label=Comment has been deleted
```
jmiranda, line 358: Again, these messages should just use the default messages i.e. default.updated.message

The messages above (i.e. default.comment.label should follow conventions and look like this
```
comment.label=Comment
comments.label=Comments
comment.comment.placeholder=Enter a comment...
```

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,87 @@ class StockMovementController {
...
+        redirect(action: "show", id: requisition.id)
+    }
+
+    def saveComment = {
+        Requisition requisition = Requisition.get(params.requisition)
+        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
+
+        Comment comment = new Comment(params)
+         if (comment.validate()) {
+             requisitionService.saveComment(comment)
```
drodzewicz, line 342: I need this save to get the comment id

kchelstowski: this save is needed anyway, because you create a **brand new** comment which is not yet managed, so nothing wrong with that :)

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -356,13 +357,13 @@ class StockMovementController {
             throw new UnsupportedOperationException("${g.message(code: 'errors.noPermissions.label')}")
         }
         comment.properties = params
-        if (!comment.validate()) {
+        if (comment.validate()) {
             requisitionService.saveComment(comment)
```
kchelstowski, line 361: previously I meant that this save here would not be needed, because it is already a managed entity and calling `save()` explicitly is not needed, because it'd be saved at the end of the transaction.

It's not a "must change request", but just wanted to point that out, because in some cases (not this one, but I want to explain a potential problem somewhere), calling those `save` methods when they are not needed, might cause performance issues.

Assuming you have a managed entity that stores hundreds of items or something like that and for each change in the transaction you would call `entity.save()`, this would e.g. go through those items (and validate them for example) that many times, that you've called this save + 1 (because the save would be triggered at the end of the transaction, because it's the managed entity).

Not to be groundless - lately, thanks to removing that one, additional unnecessary `save` call, I've improved the performance of one request from **28 seconds to 3 seconds**.

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
awalkowiak: However comments don't have hundreds of items, so this should not be a problem here.

grails-app/views/stockMovement/addComment.gsp
```diff
@@ -0,0 +1,73 @@
...
+
+	<div class="body">
+		<g:if test="${flash.message}">
+			<div class="message">
+				${flash.message}
+			</div>
+		</g:if>
+		<g:hasErrors bean="${stockMovement}">
+			<div class="errors">
+				<g:renderErrors bean="${stockMovement}" as="list" />
```
awalkowiak, line 19: Is this populated?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,87 @@ class StockMovementController {
...
     }
 
+    def comments = {
+        Requisition requisition = Requisition.get(params.id)
+        render(template: "comments", model: [requisition: requisition])
+    }
+
+    def addComment = {
+        Requisition requisition = Requisition.get(params.id)
+        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
```
awalkowiak, line 296: I think this is not needed because show action will do it on its own again and I think addComment don't have to use `stockMovement` (correct me if I am wrong)

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,87 @@ class StockMovementController {
...
+        [stockMovement: stockMovement, comment: new Comment()]
+    }
+
+    def editComment = {
+        Requisition requisition = Requisition.get(params.requisition)
+        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
```
awalkowiak, line 302: Same question/comment as in the `addComment` action

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,87 @@ class StockMovementController {
...
+            throw new UnsupportedOperationException("${g.message(code: 'errors.noPermissions.label')}")
+        }
+        requisitionService.deleteComment(comment, requisition)
+        flash.message = "${g.message(code: 'default.deleted.message', args: [g.message(code: 'comment.label', default: 'Comment'), comment.id])}"
+        redirect(action: "show", id: requisition.id)
+    }
+
+    def saveComment = {
+        Requisition requisition = Requisition.get(params.requisition)
+        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
```
awalkowiak, line 338: Again, same question/comment as in the `addComment` action.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -283,6 +286,87 @@ class StockMovementController {
...
+             flash.message = "${g.message(code: 'default.created.message', args: [g.message(code: 'comment.label', default: 'Comment'), comment.id])}"
+             redirect(action: "show", id: stockMovement.id)
+             return
+         }
+        render(view: "addComment", model: [stockMovement: stockMovement, comment: comment])
+    }
+
+    def updateComment = {
+        Requisition requisition = Requisition.get(params.requisition)
+        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
```
awalkowiak, line 353: same

awalkowiak (commented):
@drodzewicz Overall LGTM, I have one question for a couple of actions (see comments)

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: this is why I gave an approval and wrote this:

> because in some cases (not this one, but I want to explain a potential problem somewhere),

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
awalkowiak: Oh ok, this makes sense to me now.

awalkowiak (approved):

PR #4311, opened 2023-10-03
OBPIH-5854 New approval statuses in SM list pages - Approvals lists pages (fix after QA)

alannadolny:


grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -39,6 +39,7 @@ class OutboundStockMovementService {
         Date createdAfter = params.createdAfter ? Date.parse("MM/dd/yyyy", params.createdAfter) : null
         Date createdBefore = params.createdBefore ? Date.parse("MM/dd/yyyy", params.createdBefore) : null
         List<ShipmentType> shipmentTypes = params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
+        Boolean isApprovalRequired = stockMovement?.origin?.isApprovalRequired()
```
kchelstowski, line 42: not a change request: since it is a transient, you can access it through the property like: `origin.approvalRequired`

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -98,6 +99,20 @@ class OutboundStockMovementService {
...
+                    }
+                }
+            } else {
+                not {
+                    'in'("status", [RequisitionStatus.PENDING_APPROVAL, RequisitionStatus.REJECTED])
+                }
+                if (!isApprovalRequired) {
+                    not {
+                        'in'("status", [RequisitionStatus.APPROVED])
+                    }
```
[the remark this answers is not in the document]
alannadolny: When we are on the stock movements list we don't pass any params with sourceType - it's just null

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -98,6 +99,20 @@ class OutboundStockMovementService {
                     not {
```
awalkowiak, line 99: Should this `not` be together with the next one or it should be one of these depending on the case if it it approval required or not?

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -98,6 +99,20 @@ class OutboundStockMovementService {
...
                         'in'("status", [RequisitionStatus.CREATED, RequisitionStatus.ISSUED, RequisitionStatus.CANCELED])
                     }
+                    if (!isApprovalRequired) {
+                        not {
+                            'in'("status", RequisitionStatus.listApproval())
+                        }
+                    }
+                }
+            } else {
+                not {
```
awalkowiak, line 109: same question as above

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
alannadolny: Those `not`s are separated by purpose:
For example, it can be read also as
- if approval is required: check if status is in [CREATED, ISSUED, CANCELED]
- if approval is not required: check if the status is in [CREATED, ISSUED, CANCELED, APPROVED, PENDING_APPROVAL, REJECTED]

I just moved the common part higher, and because of that I got rid of one additional else and longer lists of statuses

awalkowiak: I think it might be easier to follow the code if it was (with some commentary around it):
```
if required {
    not in the list of some statuses + list of other statuses
} else {
    not in the list of some statuses
}
```

alannadolny: Are you sure about that? This redundancy of statuses looks odd to me, so maybe it would be enough when I add just comments around the current code?

alannadolny: I added some comments for now, if you think it's not enough, let me know

awalkowiak: Ok, I think this makes sense for now

PR #4315, opened 2023-10-04
OBPIH-5873 Investigate empty date approved after approval action

alannadolny:


grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -856,7 +856,8 @@ class RequisitionService {
                 requisition.rejectedBy = currentUser
                 break
             default:
-                break
+                requisition.status = newStatus
         }
+        requisition.save(flush: true)
```
kchelstowski, line 861: I think we should get rid of one of the saves - either this one or the one in the `transitionRequisitionStatus`.
If removing the `flush` won't cause any problems with the event that is published after this method is called, we could remove this save in my opinion.

src/groovy/org/pih/warehouse/requisition/RequisitionStatusTransitionEvent.groovy
```diff
@@ -1,16 +1,15 @@
 package org.pih.warehouse.requisition
 
-import org.pih.warehouse.core.User
 import org.springframework.context.ApplicationEvent
 
 class RequisitionStatusTransitionEvent extends ApplicationEvent {
     Requisition requisition
 
-    User createdBy
+    RequisitionStatus status
```
kchelstowski, line 8: is it necessary? Can't you get it from the requisition?

awalkowiak: Yeah, I thought we were going to do this synchronously to have a valid status on the requisition...

alannadolny: So here is one more problem I noticed - setting status synchronously fixed the problem with "row was updated by another transaction".  But here appears the second issue in sending emails asynchronously. We are getting requisition in `RequisitionStatusTransitionEventService.groovy` in this way:
```
Requisition.withSession {
            Requisition requisition = Requisition.get(event.requisition.id)
            notificationService.publishRequisitionStatusTransitionNotifications(requisition, event.status)
        }
```
it was a workaround for the LazyInitializationException. This session created for sending emails differs from this one in which we are setting the status for the requisition.  This requisition that we are getting sometimes differs from the one that we have in the function for changing status, so that's the reason why I passed the status directly.
I think there are three possible ways to solve this issue:
1. Pass the status directly to the function for sending emails (it's the actual version of the code in this PR). We don't change other values that are used in the emails.
2. Fetch status in the requisition eagerly - I remember that @kchelstowski tried to do it in one of his PRs previously. In this version:
- the `withSession` is no longer needed
- we don't have to pass status 
- we don't have to fetch requisition "by hand"
- it will be just adding the fetch:
```
events joinTable: [name: "requisition_event", key: "requisition_id"], fetch: 'join'
```
3. Create a DTO, with all of the fields that will be needed in this event and pass it to the event directly.
4. [PROBABLY JUST IN GRAILS 3+] Get similar behavior to:
`@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)` which is available in Spring. Due to that, we will trigger our event after committing to the first session.
Something similar in Grails: https://async.grails.org/latest/api/grails/events/annotation/Publisher.html

What do you think about these approaches / current versions? @awalkowiak @jmiranda

awalkowiak (approved):

grails-app/services/org/pih/warehouse/inventory/RequisitionStatusTransitionEventService.groovy
```diff
@@ -10,21 +10,14 @@
...
-    RequisitionService requisitionService
     NotificationService notificationService
 
     void onApplicationEvent(RequisitionStatusTransitionEvent event) {
-        Requisition.withSession {
-            Requisition requisition = Requisition.get(event.requisition.id)
-            requisitionService.triggerRequisitionStatusTransition(requisition, event.createdBy)
-            notificationService.publishRequisitionStatusTransitionNotifications(requisition)
-        }
+            notificationService.publishRequisitionStatusTransitionNotifications(event.requisition)
```
kchelstowski, line 21: the indent feels too big, doesn't it?

kchelstowski (approved):

PR #4317, opened 2023-10-05
OBPIH-5806 Update with PENDING_APPROVAL status on approval submit and REQUESTED on reagular outbound

drodzewicz:
lets discuss...

The issue came up when trying to create an outbound with a location which has the supported activity for request approval.
The expected behaviour would be that on Edit step an outbound should have status `VERIFYING` but instead it was `WAITING FOR APPROVAL`.

The reason for that was that in both cases (Request and Outbound) we update status with `{ status: 'REQUESTED' }` which is mapped to `PENDING_APPROVAL` (change which was introduced in [here](https://github.com/openboxes/openboxes/pull/4288/files#diff-64dcf115d2530bcb242b9a4d135db1e44c97f2ca945b602816d3b78f33eedd58R115))

Was there a particular reason why we didn't separate VERIFYING and PEWNDING_APPROVAL as individual cases?

### DIFF
--- a/grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
+++ b/grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
@@ -836,13 +836,15 @@ class RequisitionService {
         // OBPIH-5134 Request approval feature implements additional status transitions for a request
         // If fulfilling location does not require approval we omit all other status transitions and set it to VERIFYING
         switch(newStatus) {
+            case RequisitionStatus.PENDING_APPROVAL:
+                if (requisition.origin.approvalRequired) {
+                    transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
+                    requisition.approvalRequired = true
+                    break
+                }
             case RequisitionStatus.VERIFYING:
                 transitionRequisitionStatus(requisition, RequisitionStatus.VERIFYING, EventCode.SUBMITTED, currentUser)
                 requisition.approvalRequired = false
-                break
-            case RequisitionStatus.PENDING_APPROVAL:
-                transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
-                requisition.approvalRequired = true
                 break
             case RequisitionStatus.APPROVED:
                 transitionRequisitionStatus(requisition, RequisitionStatus.APPROVED, EventCode.APPROVED, currentUser)
--- a/grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
+++ b/grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
@@ -834,24 +834,30 @@ class RequisitionService {
 
     void triggerRequisitionStatusTransition(Requisition requisition, User currentUser, RequisitionStatus newStatus) {
         // OBPIH-5134 Request approval feature implements additional status transitions for a request
-        // If fulfilling location does not require approval we omit all other status transitions and set it to VERIFYING
         switch(newStatus) {
-            case RequisitionStatus.PENDING_APPROVAL:
-                if (requisition.origin.approvalRequired) {
-                    transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
-                    requisition.approvalRequired = true
-                    break
-                }
             case RequisitionStatus.VERIFYING:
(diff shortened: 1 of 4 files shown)


grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -834,16 +834,14 @@ class RequisitionService {
...
+                requisition.approvalRequired = false
+                break
             case RequisitionStatus.PENDING_APPROVAL:
-                if (requisition.origin.approvalRequired) {
-                    transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
-                    requisition.approvalRequired = true
-                } else {
-                    transitionRequisitionStatus(requisition, RequisitionStatus.VERIFYING, EventCode.SUBMITTED, currentUser)
-                    requisition.approvalRequired = false
-                }
```
kchelstowski, line 846: why have you got rid of those lines?
If you changed on frontend to send `PENDING_APPROVAL` no matter of location supported activities, if a location doesn't support the approvals it would set the status for `PENDING_APPROVAL` not for `VERIFYING`, wouldn't it?

src/js/components/stock-movement-wizard/request/AddItemsPage.jsx
```diff
@@ -1229,7 +1229,7 @@ class AddItemsPage extends Component {
   submitRequest(lineItems) {
     const nonEmptyLineItems = _.filter(lineItems, val => !_.isEmpty(val) && val.product);
     this.saveRequisitionItems(nonEmptyLineItems)
-      .then(() => this.transitionToNextStep('REQUESTED'));
+      .then(() => this.transitionToNextStep('PENDING_APPROVAL'));
```
kchelstowski, line 1232: see the comment above - either you bring back those lines in the `PENDING_APPROVAL` case, or we should should check (or actually we might do it in both places) here, if a location supports approval workflow - if it does, send `PENDING_APPROVAL`, otherwise after your refactor, I guess we should send `VERIFYING` directly, instead of `REQUESTED`.

kchelstowski (commented):
answering your question - it was probably missed because we forgot that those methods are also used for outbounds, and we focused on conditions (if a location does/does not support approvals).
It is probably my fault that I have caused this regression, so I'm sorry that you had to fix that :see_no_evil:

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
drodzewicz: yes, good point., this is what I wanted to hear, because I had a feeling I was missing something.
I'll bring back those lines and leave the VERYFING case too.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -834,16 +834,16 @@ class RequisitionService {
...
-        // If fulfilling location does not require approval we omit all other status transitions and set it to VERIFYING
         switch(requisition.status) {
             case RequisitionStatus.PENDING_APPROVAL:
                 if (requisition.origin.approvalRequired) {
                     transitionRequisitionStatus(requisition, RequisitionStatus.PENDING_APPROVAL, EventCode.PENDING_APPROVAL, currentUser)
                     requisition.approvalRequired = true
-                } else {
-                    transitionRequisitionStatus(requisition, RequisitionStatus.VERIFYING, EventCode.SUBMITTED, currentUser)
-                    requisition.approvalRequired = false
+                    break
```
kchelstowski, line 842: now - don't we want to send an event at all if we set it for `VERIFYING`? (when we send a request when origin doesn't support approvals)?

drodzewicz: what do you mean?
the `break` is inside of an `if()` and if `requisition.origin.approvalRequired` is `false` it will go to next case which is `case RequisitionStatus.VERIFYING:`.

I wanted to avoid duplicate code with this approach but if you feel it is too confusing then let me know, I'll either add a comment or duplicate it for the sake of clarity.

kchelstowski: yes, it will go there, but from the frontend you still always send `PENDING_APPROVAL` and if a location doesn't support the approvals, it won't run the `transitionRequisitionStatus` method at all. Or am I missing something?

drodzewicz: it will run `triggerRequisitionStatusTransition` and set the status to `VERIFYING`.
Anyways, these state transitions are getting a bit too confusing, I think I need to rethink my approach to which status to send and how to react to it.

kchelstowski (changes requested):

kchelstowski (approved):

awalkowiak (approved):

PR #4318, opened 2023-10-05
OBPIH-5809 Approval rollback - Approvals end to end workflow

alannadolny:


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -267,7 +268,30 @@ class StockMovementController {
...
+        String stockMovementId = params.get("id")
+        if (!stockMovementService.canRollbackApproval(AuthService.currentUser.get(), stockMovementId)) {
+            flash.message = g.message(
+                    code: "request.errors.approval.rollback.message",
+                    default: "Unable to rollback approval: You do not have permissions to perform this action",
+                    args: [g.message(code: "errors.noPermissions.label")]
+            )
+        }
+
+        try {
```
kchelstowski, line 280: I wish we were full React, not to have those ugly try/catches to display an error message.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -393,6 +394,12 @@ class Requisition implements Comparable<Requisition>, Serializable {
         }?.sum() ?: 0
     }
 
+    Event getMostRecentEvent() {
+        if (events?.size() > 0) {
+            return events.iterator().next()
```
kchelstowski, line 399: are we sure the recent event would be the first in the list?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -859,4 +860,31 @@ class RequisitionService {
...
+        Event event = requisition.mostRecentEvent
+        if (!event) {
+            throw new RuntimeException(g.message(
+                    code: "requisition.error.rollback.noRecentEvents",
+                    default: "Cannot rollback requisition because there are no recent events"
+            ))
+        }
+        try {
+            deleteEvent(requisition, event)
+        } catch (Exception e) {
```
kchelstowski, line 881: I think this catch won't do anything, since the eventual commit of the transaction (and the error that might happen then) from `deleteEvent` would happen at the end of the session?

kchelstowski (changes requested):

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
kchelstowski: I think you've probably followed what is done for `Shipment`, but take a look, that there is `SortedSet` used there. 
Since `compareTo` method in `Event` domain: https://github.com/openboxes/openboxes/blob/c6d16a41c38d8b0812a98af23de80be64673ab3d/grails-app/domain/org/pih/warehouse/core/Event.groovy#L45-L54
compares by date, we are fine for the `SortedSet`, but in Requisition events are just a `Set`, so it probably won't work the way you did.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -267,7 +268,30 @@ class StockMovementController {
...
+            stockMovementService.rollbackApproval(stockMovementId)
+            flash.message = g.message(
+                    code: "request.successfulApproval.rollback.message",
+                    default: "Successfully rolled back approval",
+            )
+        } catch (Exception e) {
```
kchelstowski, line 286: I also feel hesitant about this `catch`, because it might prevent potential rollback from working (I don't remember if it was Grails 1 or Grails 3 that was not rolling back when checked exception happens).
Since you do a few operations, like - deleting an event, updating a requisition (setting some properties to `null`), I think something weird might happen, but I would have to debug it myself to confirm that. Just wanted to point that out.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2814,4 +2816,28 @@ class StockMovementService {
...
+            requisition.dateRejected = null
+        }
+    }
+
+    Boolean canRollbackApproval(User user, String stockMovementId) {
+        StockMovement stockMovement = getStockMovement(stockMovementId)
+
+        // When a user is an approver or requestor or is a manager or superuser, then can rollback approval
+        Boolean isApproverOrRequestor = stockMovement?.approvers?.contains(user) || stockMovement?.requestedBy == user
+        Boolean isManagerOrSuperuser = userService.isSuperuser(user) || userService.isUserAdmin(user)
```
jmiranda, line 2839: Manager and Admin are different roles. Do the requirements state manager or admin?

I would just interrogate the user object to see if has a role greater than manager. In fact, we could throw these methods into the domain / command classes.

Do we need to take the location into account when dealing with the user roles?
```
Boolean canUserRollbackApproval(String id, User user) {
    StockMovement stockMovement = getStockMovement(stockMovementId)     
    return stockMovement.canUserRollback(user) || userService.isUserManager()
}
```
Do we need to take the origin or destination location into account when considering roles?

One other thing we'll want to check here is whether the stock movement is in an approval state. We aren't checking whether the user is able to rollback other events, so we shouldn't allow it. Something like 

StockMovement
```
Boolean isInApprovalWorkflow() { 
    status in RequisitionStatus.listApprovalStatuses()
}
```

jmiranda (changes requested):
take a look at the comments

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: @jmiranda I know it is. I meant that weird things might happen if the **transaction rollback** would happen.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
alannadolny: The main purpose of this catch is to show a more meaningful error, with: "Error rolling back most recent event".

kchelstowski: I know what's the purpose of that, but I meant that you won't catch any exception here, because any error that would occur from `deleteEvent` would happen at the end of the session - take a look at it is done in the same transaction - it is not the same case if you had `try/catch` in the controller and persistence actions in the service - here, you have try/catch in the service.

alannadolny: ok, now it's clear

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
alannadolny: > Manager and Admin are different roles. Do the requirements state manager or admin?

Uh, at the beginning, I wrote "Manager" by accident, and I forgot to change the variable name later 😕

src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -482,4 +482,13 @@ class StockMovement {
...
         return "${g.message(code: 'enum.' + status?.getClass()?.getSimpleName() + '.' + status)}"
     }
+
+    Boolean canUserRollbackApproval(User user) {
+        // Approval can be rolled back by user who is approver or requestor
+        return approvers?.contains(user) || requestedBy == user
+    }
+
+    Boolean isInApprovalState() {
+        return requisition?.status in [RequisitionStatus.APPROVED, RequisitionStatus.REJECTED]
```
kchelstowski, line 492: how about `PENDING_APPROVAL`?

src/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -482,4 +482,13 @@ class StockMovement {
         def g = ApplicationHolder.application.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
         return "${g.message(code: 'enum.' + status?.getClass()?.getSimpleName() + '.' + status)}"
     }
+
+    Boolean canUserRollbackApproval(User user) {
+        // Approval can be rolled back by user who is approver or requestor
+        return approvers?.contains(user) || requestedBy == user
```
kchelstowski, line 488: I don't remember how groovy will handles that type of comparison, but maybe you've checked that - I am not sure if `requestedBy == user` will evaluate to `true` if we have different references? 
If it doesn't, then compare by id. Or maybe we have the same reference at this point?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2813,4 +2815,33 @@ class StockMovementService {
         }
         return true
     }
+
+    void rollbackApproval(String stockMovementId) {
+        StockMovement stockMovement = getStockMovement(stockMovementId)
+        if (!canRollbackApproval(AuthService.currentUser.get(), stockMovement)) {
```
kchelstowski, line 2821: we NEED to remember to change it in Grails 3 - in Grails 3 we should use `AuthService.currentUser`, since adding `get()` makes it returning `null`.

kchelstowski: please add an appropiate TODO note or check if in Grails 1 it would work without `get()`

src/groovy/org/pih/warehouse/api/StockMovement.groovy
kchelstowski: I guess never mind, after going through the code again, I probably understand now.

alannadolny: We can rollback only in APPROVED or REJECTED status

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
alannadolny: the current user is wrapped in `ThreadLocal`
image
and this type has method get to return the stored value

kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: @alannadolny yes, this is the correct way to retrieve the User in Grails 1 because current user is a ThreadLocal. In Grails 3, the user object is still a threadlocal<user> but I think we added a way to retrieve the user object from the AuthService without the additional get(). In other words I think we renamed the property and added a method getCurrentUser() so that we could just invoke AuthService.currentUser to get access to the User. So @kchelstowski just wants a reminder TODO to make sure don't forget to change it when the code is ported.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2813,4 +2815,34 @@ class StockMovementService {
         }
         return true
     }
+
+    void rollbackApproval(String stockMovementId) {
+        StockMovement stockMovement = getStockMovement(stockMovementId)
+        // TODO: remove .get() from current user in Grails 3 (with .get it will return null)
+        if (!canRollbackApproval(AuthService.currentUser.get(), stockMovement)) {
+            String errorMessage = applicationTagLib.message(
+                    code: "request.errors.approval.rollback.message",
```
jmiranda, line 2824: Now that I've seen this, I'm convince the messages should follow the convention
```
request.rollbackApproval...
```
and this particular message should be 
```
request.rollbackApproval.insufficientPermissions.message
```

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2813,4 +2815,34 @@ class StockMovementService {
...
+        }
+
+        Requisition requisition = stockMovement?.requisition
+        if (requisition.status in [RequisitionStatus.APPROVED, RequisitionStatus.REJECTED]) {
+            requisitionService.rollbackLastEvent(requisition)
+            requisition.status = RequisitionStatus.PENDING_APPROVAL
+            requisition.approvedBy = null
+            requisition.rejectedBy = null
+            requisition.dateApproved = null
+            requisition.dateRejected = null
```
jmiranda, line 2837: I don't know if we want to remove the rejection, do we? If a user has made a request and it has been rejected and then approved, don't we want the rollback to only undo the approval? Or does approval also remove the rejected?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -267,7 +268,22 @@ class StockMovementController {
     }
 
     def rollbackApproval = {
-        throw new NotImplementedException("Action ${params.actionName} not implemented")
+        String stockMovementId = params.get("id")
+        try {
+            stockMovementService.rollbackApproval(stockMovementId)
+            flash.message = g.message(
+                    code: "request.successfulApproval.rollback.message",
```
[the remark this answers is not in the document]
jmiranda: As you'll read below, I think the second one is best. And we should follow that convention going forward.
```
object.action.success.message
object.action.error.message
```
I see a lot of places where this pattern is not used so we should add a todo to the "consolidate message.properties" tech debt ticket (assuming it still exists).

jmiranda (changes requested):
LGTM just make a few minor changes and this can be merged.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
alannadolny: AuthService.currentUser is pointing to that line:
```
    static ThreadLocal<User> currentUser = new ThreadLocal<User>()
```
I added a comment with reminder to change that

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
alannadolny: A request can't be rejected and then approved. It has to be rolled back before approval so it won't have approved and rejected data.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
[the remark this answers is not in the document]
alannadolny: Yup, I missed those requested changes, thanks for pointing that out.

awalkowiak (approved):

PR #4319, opened 2023-10-06
OBPIH-5816 approval and reject actions on list page approvals lists pages

drodzewicz:


src/js/components/stock-movement/outbound/StockMovementOutboundTable.jsx
```diff
@@ -42,20 +48,79 @@ const StockMovementOutboundTable = ({
...
+      if (statusCode === RequisitionStatus.PENDING_APPROVAL) {
+        const approveAction = {
+          defaultLabel: 'Approve',
+          label: 'react.stockMovement.action.approve.label',
+          leftIcon: <RiCheckFill />,
+          onClick: id => approveRequest(id, identifier),
+        };
+        actions.push(approveAction);
+      }
+      if (statusCode === RequisitionStatus.PENDING_APPROVAL) {
```
kchelstowski, line 78: you can move the content of that if into the if above, since those are the same "ifs"

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
```diff
@@ -132,6 +133,34 @@ const useOutboundListTableData = (filterParams) => {
...
+      const successMessage = `You have successfully Rejected the request ${identifier}`;
+      Alert.success(successMessage);
+      fireFetchData();
+    } finally {
+      dispatch(hideSpinner());
+    }
+  };
+
+  const rollbackRequest = () => {
+    Alert.error('Not implemented');
```
kchelstowski, line 161: will it be handled in another ticket or are you waiting for it to be merged?

src/js/reducers/sessionReducer.jsx
```diff
@@ -90,6 +91,7 @@ export default function (state = initialState, action) {
         isSuperuser: _.get(action, 'payload.data.data.isSuperuser'),
         isUserAdmin: _.get(action, 'payload.data.data.isUserAdmin'),
         isUserApprover: _.get(action, 'payload.data.data.isUserApprover', false),
+        isUserRequestApprover: _.get(action, 'payload.data.data.isUserRequestApprover', false),
```
kchelstowski, line 94: whenever I see this ugly `_.get` I wish we refactored it in the future :pray: 
to something like:
```js
const { isUserRequestApprover, ... } = payload.data.data
...
isUserApprover: isUserApprover ?? false
```

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
```diff
@@ -132,6 +133,34 @@ const useOutboundListTableData = (filterParams) => {
...
+      dispatch(hideSpinner());
+    }
+  };
+
+  const rejectRequest = async (id, identifier) => {
+    dispatch(showSpinner());
+    try {
+      await stockMovementApi.updateStatus(id, RequisitionStatus.REJECTED);
+      const successMessage = `You have successfully Rejected the request ${identifier}`;
+      Alert.success(successMessage);
```
kchelstowski, line 153: and a question why don't you use our new notifications?
```js
notification(NotificationType.SUCCESS)({
  message: <message>
})
```

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
drodzewicz: I am waiting for the changes to be merged.
I can either wait for it or add it in the following PR.

src/js/hooks/list-pages/outbound/useOutboundListTableData.jsx
drodzewicz: I simply used the same Alert as we have in this file

kchelstowski: because `notification` was implemented later this is why in this file `Alert` was used. I think we should use `notification`, since it lets us do more and this should become a convention. (I know in this case it would evaluate to the same, but just not to mix `Alert` vs `notification`.

kchelstowski (approved):
lgtm

jmiranda (approved):
LGTM

awalkowiak (approved):

PR #4320, opened 2023-10-06
OBPIH-5805 Approver cannot edit requests - Approvals end to end workflow (fix after QA)

alannadolny:


src/js/utils/permissionUtils.js
```diff
@@ -11,12 +11,17 @@ const canEditRequest = (currentUser, request, location) => {
     ActivityCode.APPROVE_REQUEST,
   );
 
-  // If the location supports request approval, only the requestor is able to edit
-  // and the request can't be approved / rejected when editing
   if (isApprovalRequired) {
+    // If the request is approved, approver cannot edit request,
+    // but normal warehouse user can edit it
```
kchelstowski, line 16: what do we mean by "normal warehouse user can edit it"?
Below I can see that you check if `isUserRequestor`, so that would mean that if request is approved, only requestor can edit that, is that true?

src/js/utils/permissionUtils.js
```diff
@@ -11,12 +11,17 @@ const canEditRequest = (currentUser, request, location) => {
...
+    const isUserApprover = request?.approvers?.find(user => user?.id === currentUser?.id);
+    if (request.statusCode === RequisitionStatus.APPROVED && isUserApprover) {
+      return false;
+    }
+    // If the location supports request approval, only the requestor is able to edit
```
kchelstowski, line 21: here you answered my question from above, but I'm confused about this "normal warehouse user can edit it".

src/js/utils/permissionUtils.js
alannadolny: I forgot that I did all of the requests from one account - so I was the only requestor here 🫣

kchelstowski (approved):
something still feels weird to me here, but I feel like I don't understand something from "domain level", so I hope someone else will also take a look at it.

src/js/utils/permissionUtils.js
```diff
@@ -11,13 +11,18 @@ const canEditRequest = (currentUser, request, location) => {
...
-      request.statusCode !== RequisitionStatus.REJECTED;
+    // If the location supports request approval, only the requestor is able to edit
+    // If the request is rejected, then it cannot be edited
+    if (request.statusCode !== RequisitionStatus.APPROVED) {
+      return isUserRequestor &&
+        (isLocationDestination || isLocationOrigin) &&
+        request.statusCode !== RequisitionStatus.REJECTED;
+    }
+    // If the request is approved, approver cannot edit request,
+    // but fulfilling user can edit it
```
awalkowiak, line 23: This makes no sense to me (I understand that it was requested to work that way), but this still makes no sense to me. I see the use case where I have one person doing approvals and fulfilling.

awalkowiak (approved):

PR #4322, opened 2023-10-09
Basic implementation of webhooks for moriana integration

jmiranda:


grails-app/conf/Config.groovy
```diff
@@ -1091,6 +1091,18 @@ openboxes.browser.connection.status.timeout = 8000
 //google.analytics.enabled = false
 //google.analytics.webPropertyID = "UA-xxxxxx-x"
 
+// Default webhook configuration
+openboxes {
+    webhook {
+        enabled = false
+        endpoint {
+            url = 'https://webhook.site/9969a2d6-9780-483d-beb4-50667d66497d'
```
awalkowiak, line 1099: Do you want to have these values here? Or it should be empty string as a default?

grails-app/conf/Config.groovy
```diff
@@ -1091,6 +1091,18 @@ openboxes.browser.connection.status.timeout = 8000
...
+            headers  = [
+                header1: "value1"
```
awalkowiak, line 1101: Do you want to have these values here? Or it should be empty string as a default?

grails-app/services/org/pih/warehouse/core/ApiClientService.groovy
```diff
@@ -34,8 +34,8 @@ class ApiClientService {
         return execute(new HttpGet(url))
     }
 
-    JSONObject post(String url, Map requestData) {
-        return execute(new HttpPost(url), requestData)
+    JSONObject post(String url, Map payload, Map headers) {
```
awalkowiak, line 37: Looks like this is used in the `RecaptchaService`. I think you should either add a default option for headers here or add an empty map there.

grails-app/services/org/pih/warehouse/core/ApiClientService.groovy
```diff
@@ -46,14 +46,22 @@ class ApiClientService {
         return execute(new HttpPut(url), requestData)
     }
 
-    JSONObject execute(HttpEntityEnclosingRequestBase request, Map requestData) {
-        if (requestData) {
-            JSONObject jsonObject = new JSONObject(requestData)
+    JSONObject execute(HttpEntityEnclosingRequestBase request, Map payload, Map headers) {
```
awalkowiak, line 49: Again, a default empty map or null would be good to have here.

grails-app/services/org/pih/warehouse/core/WebhookPublisherService.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+import org.codehaus.groovy.grails.commons.ConfigurationHolder
+import org.pih.warehouse.shipping.Shipment
+import org.pih.warehouse.auth.AuthService
+
+
+class WebhookPublisherService {
+
+    def apiClientService
+    //def grailsApplication
```
awalkowiak, line 20: If not required, then please remove

grails-app/services/org/pih/warehouse/core/WebhookPublisherService.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+    def publishShippedEvent(Shipment shipment) {
+
+        boolean webhooksEnabled = shipment.origin.supports(ActivityCode.ENABLE_WEBHOOKS)
+        if (!webhooksEnabled) {
+            log.info "Location ${shipment.origin} does not support activity code ${ActivityCode.ENABLE_WEBHOOKS}"
+            return
+        }
+
+        Map payload = [
```
awalkowiak, line 32: Perhaps it's worth moving into `Shipment`, as `toWebhookJson()` or something like that (?)

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -63,6 +63,7 @@ class ShipmentService {
     def documentService
     def personDataService
     def productAvailabilityService
+    def webhookPublisherService
```
awalkowiak, line 66: Is it used here?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/core/WebhookPublisherService.groovy
awalkowiak: Or this should be related to a specific webhook? What I imagine is there might be a couple of webhooks configured at the same time and each could have its own "mapping". So perhaps there would be a better place for webhook payload models instead of having this in the shipment?

grails-app/conf/Config.groovy
jmiranda: I was ok with leaving it as an example since the endpoint is disabled by default.

grails-app/conf/Config.groovy
jmiranda: Same here. If there's a better way to provide documentation here I'm all ears, but given that this is just an example I don't mind it being in the code.

grails-app/services/org/pih/warehouse/core/ApiClientService.groovy
jmiranda: Good point. I thought I had left the previous method, but I guess not. I like using a default option instead of chaining overriding methods so I'll go with that approach.

grails-app/services/org/pih/warehouse/core/WebhookPublisherService.groovy
jmiranda: I was letting the implementation of this marinate in my head a bit. I agree each domain should know what it's payload is, but this isn't just a webhook thing. Instead it needs to support generic notifications about each domain and state change. We are planning to deal with this in the improved webhook feature in OBGM, so I think we can implement it there. You can review Chetan's PR #4300 and provide feedback there 

But essentially, I want the application to emit application-level and database-level events for everything that happens in the system (i.e. ShipmentStatusUpdateEvent, HibernateUpdateEvent, ProductCreatedEvent, UserLoginEvent, etc). 

A system administrator will be able to define different notification mechanisms: webhook, email, SMS, and push notifications. The application will allow the user to subscribe to any of the events that implement the NotifiableEvent interface (or extend it if we decide it should be an abstract class). For now, these subscriptions will be configured in the application configuration, but eventually, we'll have a table where we can manage the event notification subscriptions. 

When an application event is fired (we need to manually add these in the codebase for the app-level events, but we'll have a Hibernate persistence listener to publish the low-level database events) those events will be sent to the event bus. We'll have a listener listening for instances of NotifiableEvent or whatever we call them. These events need to provide enough information for us to send a notification, so we will convert them, I think, to a NotificationEvent which has a context (based on the source object) and configuration (type = webhook, SMS, email, templates, etc). The NotificationEvent is published to the same event bus but now the listeners are going to be implementations of a NotificationEventService which know how to publish to SMTP server, SMS endpoint, or webhook. The notification event service takes the notification context and the config to create the components needed for each type of notification i.e. email requires an string subject, HTML body, email recipients; webhook requires a URL, payload, headers. 

Anyway, that's the idea behind the improved version (#4300) but I still need to review the code to make sure we implemented it that way.

I think the latest discussion of the webhooks happened in the #moriana Slack channel, but it might be buried in a document or ticket somewhere as well.
https://openboxes.slack.com/archives/C05SDV26YKC/p1695918514682159

Here's the original ticket
https://pihemr.atlassian.net/browse/OBS-1515

cc @awalkowiak @drodzewicz @alannadolny @kkaczmarczyk @chetanmaharishi

PR #4323, opened 2023-10-10
OBPIH-5814 Add NONE option to approvers filter

drodzewicz:
Unfortunately, the approach with the subqueries that we discussed on the tech-huddle wasn't ideal, the issue that came up was with filtering by requests with **empty**  associations, so I decided to go with the simplest solution of querying the request ids and then querying the requests by the list of the these ids.

the approach with subqueries looks something like
```groovy
  if (params.list("approver")) {
      or {
          createAlias("requisition", "requisition")
          if (params.list("approver").contains("null")) {
              isEmpty("requisition.approvers")
          }
          if (stockMovement.approvers) {
              createAlias("requisition.approvers", "approver")
              add Subqueries.exists(
                      DetachedCriteria.forClass(Person.class)
                              .add(Restrictions.in("approver.id", stockMovement.approvers.id))
                              .setProjection(Projections.property("id"))
              )
          }
      }
  }
```
The issue I mentioned came up when trying to filter by **None** and some _other users_. It did not return any requests with empty approvers because `createAlias("requisition.approvers", "approver")` by default sets the association to INNER_JOIN.
When changing the jointype on **Approvers** to LEFT_JOIN, we face the same problem as in the beginning, of approver column missing the rest of the data (other than specified in filter approvers)

This issue started quite the discussion of the functionality/logic and UX of filtering by empty fields values, so I think we will have to come back to this discussing and come up with a final verdict.

For now I think we should follow the approach of how it is done on **Purchase orders** with **paymentTerms** filter.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
@@ -42,15 +42,12 @@ class OutboundStockMovementService {
         List<ShipmentType> shipmentTypes = params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
         Boolean isApprovalRequired = stockMovement?.origin?.isApprovalRequired()
 
-        def query = { isCountQuery ->
-            if (isCountQuery) {
-                projections {
-                    countDistinct "id"
-                }
-            } else {
-                setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
-            }
 
+        def stockMovementsIds = OutboundStockMovementListItem.createCriteria().list() {
+            projections {
+                property "id"
+            }
+            setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
 
             if (stockMovement?.receiptStatusCodes) {
                 'in'("shipmentStatus", stockMovement.receiptStatusCodes)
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -42,15 +42,12 @@ class OutboundStockMovementService {
...
-        def query = { isCountQuery ->
-            if (isCountQuery) {
-                projections {
-                    countDistinct "id"
-                }
-            } else {
-                setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
-            }
 
+        def stockMovementsIds = OutboundStockMovementListItem.createCriteria().list() {
```
kchelstowski, line 46: I'd suggest adding a comment above with the name of the ticket:
"OBPIH-5814: Because stock movement approvers could be filtered out from the requisition by given approvers in params, here we only return ids of stock movements and below we had to make another query to get stock movements with given id's, so that second query doesn't anymore cut any approvers"

or something similar.
I won't lie that even though I knew what you've done here, I had to think a bit when looking at the code of what is going on and how it is working.

grails-app/services/org/pih/warehouse/inventory/OutboundStockMovementService.groovy
```diff
@@ -148,6 +152,11 @@ class OutboundStockMovementService {
                     'in'("shipmentType", shipmentTypes)
                 }
             }
+        }
+
+        // Get result count
```
kchelstowski, line 157: "Get result count" is not accurate I'd say, because we also return stock movements from there, not only "total count" (which I think you meant here).

Here I'd add a short note: "This query is about to return data with filters applied from the query above, but this will not filter out any approvers from the result. See the comment above".

kchelstowski:
besides adding a comment/comments, lgtm.
I wish I had more time to help, maybe it'd be possible to figure it for all cases mentioned by you, using `Restrictions/Subqueries` approach.

kchelstowski (approved):

awalkowiak (approved):

PR #4324, opened 2023-10-10
OBPIH-5838 Option to disable approval notifications

awalkowiak:


src/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -43,6 +43,10 @@ enum ActivityCode {
 
     // Notifications
     ENABLE_NOTIFICATIONS('ENABLE_NOTIFICATIONS'),
+    // For now this is meant for requests between two depots (checked only in that case).
+    // Perhaps we could also distinguish it for different type of downstream consumers and/or "upstream suppliers"
+    // (depot vs non managed locally ward)?
+    DISABLE_APPROVAL_NOTIFICATIONS('DISABLE_APPROVAL_NOTIFICATIONS'),
```
awalkowiak, line 49: @jmiranda Actually now if I think about it, enabling by default should be as easy as disabling by default if we want to rely on users to enable it. This would not need any migration.

The only problematic case would be distinguishing when it is "Depot" requestor vs "Ward" requestor and then deciding who should be getting the notification (if it is enabled) based on the status (?)

jmiranda: > Actually now if I think about it, enabling by default should be as easy as disabling by default if we want to rely on users to enable it. This would not need any migration.

In general, I think the default should be "enable approval notifications" at all location types that fulfill stock or request stock. But we can leave the migration for later if it becomes necessary. 

> The only problematic case would be distinguishing when it is "Depot" requestor vs "Ward" requestor and then deciding who should be getting the notification (if it is enabled) based on the status (?)

Yeah that's what I was trying to unravel in my head during the tech huddle. There's a matrix here that I just can't wrap my head around yet, both on the configuration side as well as on the evaluation side. I think we need to work out the evaluation side first and then we can make our way back to the configuration. 

We just need to figure out the dimensions

* Requisition Status
* Requesting Location Type
* Fulfilling Location Type

Is that enough? Do we need to take user roles into account as well?

* User Roles

For example, one scenario in the matrix might look like this
```
Prevent the system from sending notifications to Approvers and Requestors when a request from a depot has been submitted, approved, or rejected.

When:
  Requisition Status: Approved
  Inbound Location Type: Depot
  Inbound User Role: Requestor

Then:
  Enable Approval Notifications: False
```

I just can't figure out how to set this as a supported activity on locations. Maybe this?
```
ENABLE_APPROVAL_NOTIFICATIONS: Sends all approval notifications
ENABLE_CONSUMER_APPROVAL_NOTIFICATIONS: Send notifications when requesting location = CONSUMER
ENABLE_DEPOT_APPROVAL_NOTIFICATIONS: Send notifications when requesting location has MANAGE_INVENTORY?
```

Does that configuration even cover all of the different scenarios we need to handle? The naming here is also going to be atrocious.

awalkowiak: @jmiranda ok here how it looks like I think:
<p data-pm-slice="0 0 []">Assumptions:</p><ol class="ak-ol"><li><p>All notifications are enabled by default,</p></li><li><p>Option to disable on location types (all notifications, only requestor notifications or only approver notifications),</p></li><li><p>Expected notification matrix (by PIH):</p></li></ol>

Requestor | Fulfilling | Request Submitted notification (status= PENDING_APROVAL, receives approver) | Request Approved/rejected notification (status=APPROVED or REJECTED, receives requestor)
-- | -- | -- | --
Depot | Depot | No | No
Ward | Depot | Yes | Yes

<p>which would translate into:</p><ul class="ak-ul"><li><p>If request from <strong>“Depot”</strong>, then if status <span class="code" spellcheck="false">PENDING_APPROVAL</span>, then send a notification to <strong>approvers</strong> if the <span class="code" spellcheck="false">origin</span> <strong>does not </strong>support <span class="code" spellcheck="false">DISABLE_DEPOT_APPROVAL_NOTIFICATIONS</span> and <span class="code" spellcheck="false">DISABLE_APPROVAL_NOTIFICATIONS</span> (to disable set the “fulfiller” notification activity code or “global” approval notification code),</p></li><li><p>If request from <strong>"Depot"</strong>, then if status <span class="code" spellcheck="false">APPROVED</span> or <span class="code" spellcheck="false">REJECTED</span>, then send a notification to <strong>requestors</strong> if the <span class="code" spellcheck="false">destination</span> <strong>does not </strong>support <span class="code" spellcheck="false">DISABLE_CONSUMER_APPROVAL_NOTIFICATIONS</span> and <span class="code" spellcheck="false">DISABLE_APPROVAL_NOTIFICATIONS</span>(to disable set the consumer notification activity code or “global” approval notification code),</p></li><li><p>If request from <strong>"Ward"</strong>,  then if status <span class="code" spellcheck="false">PENDING_APPROVAL</span>, then send a notification to <strong>approvers</strong> if the <span class="code" spellcheck="false">origin</span> <strong>does not </strong>support <span class="code" spellcheck="false">DISABLE_APPROVAL_NOTIFICATIONS</span> (to disable set the “global” approval notification activity code),</p></li><li><p>If request from <strong>"Ward"</strong>,  then if status <span class="code" spellcheck="false">APPROVED</span> or <span class="code" spellcheck="false">REJECTED</span>, then send a notification to <strong>requestors</strong> if the <span class="code" spellcheck="false">destination</span> <strong>does not </strong>support <span class="code" spellcheck="false">DISABLE_APPROVAL_NOTIFICATIONS</span> (to disable set the “global” approval notification activity code),</p></li></ul><p></p>

Please take a look at the `shouldSendApprovalNotification` code. I was wondering how `DISABLE_DEPOT_APPROVAL_NOTIFICATIONS` should be named (I went with your option for now) and I wondered if `DISABLE_SUPPLIER_APPROVAL_NOTIFICATIONS` would be more suitable, as an opposite to the `DISABLE_CONSUMER_APPROVAL_NOTIFICATIONS`.

awalkowiak: Tech huddle:
1. `..._FULFILLER_APPROVAL_NOTIFICATIONS` and `..._REQUESTOR_APPROVAL_NOTIFICATIONS` instead of `..._DEPOT_APPROVAL_NOTIFICATIONS` and `..._CONSUMER_APPROVAL_NOTIFICATIONS` 
2. Let's make it enabled by default (with migration) and disable it by unassigning the activity code (on a location type level)
3. Let's stick to two activity codes (if possible)

awalkowiak: @jmiranda I did it with only two activity codes. To disable this for any type of location it needs to be disabled (removed) on the **requesting** location (`ENABLE_FULFILLER_APPROVAL_NOTIFICATIONS` **to not send** notification to approvers and `ENABLE_REQUESTOR_APPROVAL_NOTIFICATIONS` **to not get** notification about approval or rejection)

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -401,6 +402,22 @@ class Requisition implements Comparable<Requisition>, Serializable {
...
+            // and should send notification to approvers
+            return destination.supports(ActivityCode.ENABLE_FULFILLER_APPROVAL_NOTIFICATIONS)
+        } else if ([RequisitionStatus.APPROVED, RequisitionStatus.REJECTED].contains(status)) {
+            // if approved or rejected, then check if destination (requesting location)
+            // has enabled notifications and should get notification  about approval or rejection
+            return destination.supports(ActivityCode.ENABLE_REQUESTOR_APPROVAL_NOTIFICATIONS)
+        }
+
+        // by default always send approval workflow notifications (if not handled above)
+        return true
```
jmiranda, line 418: Given that we're handling all of the states above it doesn't seem possible. Can you identify a scenario where this could happen? 

Also I feel like I'm missing something here. It seems like the "shouldSendApprovalNotification" method should be location-dependent? In other words, we should enable fulfiller notifications on the origin and requestor approval notifications on the requestor? So asking whether to send the approval notifications should be again the stock movement or the 

Pending Approval: send Approval notification to approvers if origin has approval notifications enabled
Approved: send Approved notification to requestors if destination has approval notifications enabled
Rejected: send Rejected notification to requestors if destination has approval notifications enabled

That way we can turn each notification type on / off for each side of the stock movement. 

Can you explain why it's better for the downstream endpoint to control the notification? I'm thinking of a scenario where, as an implementer in a distribution center, I want to disable notifications for Pending Approval because I've set up my system to allow users to manage their workflow through the dashboard. I don't want one of my requestors to then re-enable notifications because they have a different workflow.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -268,7 +268,9 @@ class StockMovementService {
             log.warn("Transition from ${requisition.status.name()} to ${status.name()} is not allowed - use rollback instead")
         } else {
             requisitionService.triggerRequisitionStatusTransition(requisition, AuthService.currentUser.get(), status, comment)
-            publishEvent(new RequisitionStatusTransitionEvent(requisition))
+            if (requisition.shouldSendApprovalNotification()) {
+                publishEvent(new RequisitionStatusTransitionEvent(requisition))
```
jmiranda, line 272: This should happen no matter what. 

Whether the notification is sent should be up to the event listener.

jmiranda (changes requested):

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
awalkowiak: Actually, this is an obsolete version. You are right that it should be handled above. I am gonna change it to "false", just in case someone triggers this for the wrong status that is not handled above.

grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -401,6 +402,38 @@ class Requisition implements Comparable<Requisition>, Serializable {
         return null
     }
 
+    boolean shouldSendApprovalNotification() {
+        if (destination.isDownstreamConsumer()) {
```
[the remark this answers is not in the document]
awalkowiak: I am assuming this is part of the older review, because this `if` is not there anymore

PR #4328, opened 2023-10-12
OBPIH-5868 mandatory comment after reject action gsp view page approvals end to end workflow

drodzewicz:


grails-app/views/stockMovement/addComment.gsp
```diff
@@ -69,5 +92,13 @@
 			</div>
 		</div>
 	</div>
+<script type="text/javascript">
+	$(document).ready(function() {
+		const commentInput = $("form [name='comment']")
+		if(${isRequestRejected}) {
```
alannadolny, line 98:
```suggestion
		if (${isRequestRejected}) {
```
😅

alannadolny (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -114,10 +114,11 @@ class StockMovementService {
         Boolean statusOnly =
                 jsonObject.containsKey("statusOnly") ? jsonObject.getBoolean("statusOnly") : false
 
+        Comment comment = null
```
kchelstowski, line 117: I don't mind this, but I think it would be maybe better to make it as a default value of the method?
I leave the decision for you

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -860,9 +859,15 @@ class RequisitionService {
...
-                transitionRequisitionStatus(requisition, RequisitionStatus.REJECTED, EventCode.REJECTED, currentUser)
+                requisition.status = RequisitionStatus.REJECTED
                 requisition.dateRejected = new Date()
                 requisition.rejectedBy = currentUser
+
+                Event event = createEvent(EventCode.REJECTED, requisition.origin, new Date(), currentUser)
+                event.comment = comment
+                requisition.addToEvents(event)
+
+                requisition.addToComments(comment)
```
kchelstowski, line 870: Instead of getting rid of calling the `transitionRequisitionStatus` here, I'd suggest refactoring the `transitionRequisitionStatus` method to include this part, so to add there something like:
```groovy
if (comment) {
  event.comment = comment
  requisition.addToEvents(event)
}
```
By the way it feels a bit dangerous to me, that we use the same comment instance adding it to a Requisition and to an Event - do we really need in both places? 
Assuming we'd want to delete a comment, we'd have to remember to check if a Comment is not also associated with an Event, because otherwise it would not allow us to delete it (some of the comments would, some not, that's why it makes me confused about this approach).
I will come back to it on Monday, because I feel like I miss some context.

grails-app/views/stockMovement/addComment.gsp
```diff
@@ -29,9 +30,25 @@
...
 				<h2><warehouse:message code="default.add.label" args="[entityName]" /></h2>
-				<g:form action="${comment?.id ? 'updateComment' : 'saveComment'}">
-					<g:hiddenField name="id" value="${comment?.id}" />
+
+				<g:set var="formAction"  value="saveComment" />
+				<g:set var="isRequestRejected" value="${approvalStatus == StockMovementStatusCode.REJECTED}" />
+				<g:if test="${comment?.id}">
+					<g:set var="formAction"  value="updateComment" />
+				</g:if>
+				<g:elseif test="${isRequestRejected}">
```
kchelstowski, line 39: since you don't have more statements below, `g:else` should be enough here.

grails-app/views/stockMovement/addComment.gsp
```diff
@@ -29,9 +30,25 @@
...
+					<g:set var="formAction" value="updateStatus" />
+				</g:elseif>
+				<g:form action="${formAction}">
+					<g:if test="${isRequestRejected}">
+						<g:hiddenField name="status" value="${StockMovementStatusCode.REJECTED}" />
+						<g:hiddenField name="id" value="${stockMovement?.id}" />
+					</g:if>
+					<g:if test="${comment?.id}">
+						<g:hiddenField name="id" value="${comment?.id}" />
+					</g:if>
```
kchelstowski, line 49: I kinda don't understand what is going on here, could you be that kind and explain those cases?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
kchelstowski: The reason I also ask about it, is that I don't see you "extracting" a Comment from an Event anywhere (in the view, or anywhere on the backend side) - I see you only adding it, so I don't know what's the purpose of making a relationship between an Event and a Comment.

awalkowiak: I agree this could be in the `transitionRequisitionStatus`. It could have a null by default comment parameter.

grails-app/migrations/0.8.x/changelog-2023-10-11-1000-alter-table-event-add-comment-field.xml
```diff
@@ -0,0 +1,31 @@
...
+            </not>
+        </preConditions>
+        <addColumn tableName="event">
+            <column name="comment_id" type="CHAR(38)"/>
+        </addColumn>
+    </changeSet>
+    <changeSet author="drodzewicz" id="111020231000-1">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <foreignKeyConstraintExists foreignKeyName="fk_comment"/>
```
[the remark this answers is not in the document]
awalkowiak: +1

grails-app/views/stockMovement/addComment.gsp
awalkowiak: What if Dariusz wants to catch if request is rejected here, but not approved without comment?

awalkowiak: btw I feel like `isRejected` should be enough

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -313,6 +318,14 @@ class StockMovementController {
         [stockMovement: stockMovement, comment: new Comment()]
     }
 
+    def rejectRequestWithComment = {
```
awalkowiak, line 321: Do you have to be that explicit here? I think `reject` or `rejectRequest` would be fine if the comment is mandatory.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: Yes please rename as reject.

grails-app/views/stockMovement/addComment.gsp
jmiranda: Is there a better way to do this without using GSP inside javascript? For example, using a hidden input field? Both are hacky but I love mixing languages less than hidden input fields.

jmiranda (changes requested):

grails-app/migrations/0.8.x/changelog-2023-10-11-1000-alter-table-event-add-comment-field.xml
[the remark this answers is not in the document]
drodzewicz: thanks for pointing that out

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
drodzewicz: I could add default value to a comment but it wouldn't change much.
It might be difficult to see on github view why I am defining a null comment at the begging but in short code looks something like this

```
Comment comment = null

switch(status){
  case CREATED:
    break
  case REQUESTED:
    break
  ....
  case REJECTED:
    comment = new Comment(parameters)
    break
   ...
}

updateRequisitionStatus(stockMovement.id, requisitionStatus, comment)

```

`updateRequisitionStatus` is called at the end for all status transitions, and I only need to pass the comment on REJECT.
So even if I instantiate the comment on REJECT I still need to access it in the outer scope so that's why I defined it at the top.

grails-app/views/stockMovement/addComment.gsp
drodzewicz: Because I am using this same template for both `updateComment` and `updateStatus` which both expect a different id parameter, for updateComment `id = commented`  and for updateStatus `id = stockMovementId` I needed to wrap them inside the if statements to avoid stacking of values since calling this code
```
<g:hiddenField name="id" value="${stockMovement?.id}" />
<g:hiddenField name="id" value="${comment?.id}" />
```

will result in `prams.id` to have a value of an Array, it doesn't redefine the value but appends to the existing one of both are called.

grails-app/views/stockMovement/addComment.gsp
drodzewicz: I completely removed this now and set **required** attribute on the comment field since I believe that it should be required either way, if it is a regular comment or a mandatory comment when rejecting.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: As I said, I don't mind it, it was only a suggestion to let you decide. I'm fine with it as it is currently.

grails-app/views/stockMovement/addComment.gsp
kchelstowski: yes, you both are right, I have not realized that you have some statement inside the elseif.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -891,6 +895,10 @@ class RequisitionService {
 
     void deleteComment(Comment comment, Requisition requisition) {
         requisition.removeFromComments(comment)
+        Event event = Event.findByComment(comment)
+        if (event) {
+            event.comment = null
```
kchelstowski, line 900: Will the comment be deleted at this point? I assume it won't, because you don't have cascade/orphan delete on `comment` on the `Event` side, do you?

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -827,14 +827,18 @@ class RequisitionService {
...
 
-    Requisition transitionRequisitionStatus(Requisition requisition, RequisitionStatus requisitionStatus, EventCode eventCode, User currentUser) {
+    Requisition transitionRequisitionStatus(Requisition requisition, RequisitionStatus requisitionStatus, EventCode eventCode, User currentUser, Comment comment = null) {
         requisition.status = requisitionStatus
         Event event = createEvent(eventCode, requisition.origin, new Date(), currentUser)
         requisition.addToEvents(event)
-        requisition.save()
+
+        if (comment) {
+            event.comment = comment
```
kchelstowski, line 836: I still kinda don't get why the same comment has to be in both event and requisition - I don't see it (`event.comment` being used anywhere), but I won't delay it any longer, but I hope to get some explanation anyway to understand the context.

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
drodzewicz: but I do on Requisition so `requisition.removeFromComments(comment)` should handle the delete.

awalkowiak (approved):

PR #4332, opened 2023-10-13
OBPIH-5870 Mandatory comment after reject action (React list page) - Approvals end to end workflow

alannadolny:


src/js/api/services/StockMovementApi.js
```diff
@@ -4,11 +4,19 @@ import {
...
   STOCK_MOVEMENT_UPDATE_STATUS,
 } from 'api/urls';
+import RequisitionStatus from 'consts/requisitionStatus';
 import apiClient from 'utils/apiClient';
 
 export default {
   getStockMovements: config => apiClient.get(STOCK_MOVEMENT_API, config),
   deleteStockMovement: id => apiClient.delete(STOCK_MOVEMENT_DELETE(id)),
   updateStatus: (id, status) => apiClient.post(STOCK_MOVEMENT_UPDATE_STATUS(id), { status }),
+  rejectRequest: (id, sender, recipient, comment) =>
```
kchelstowski, line 14: let's do it using `({ id, sender, recipient, comment })` convention, so that you don't have to remember about the order of those arguments which might be annoying.

src/js/components/stock-movement/modals/RejectRequestModalButtons.jsx
```diff
@@ -0,0 +1,29 @@
...
+      onClick={closeRejectionModal}
+    >
+      <Translate id="react.rejectRequestModal.cancel.label" defaultMessage="Cancel" />
+    </button>
+    <button
+      type="submit"
+      className="btn btn-primary align-self-end"
+    >
+      <Translate id="react.rejectRequestModal.confirmReject.label" defaultMessage="Confirm reject" />
+    </button>
```
kchelstowski, line 21: I don't know how those buttons look like (it'd be good if you attached a screenshot of the modal), but I'm wondering if it would be possible maybe to use our reusable `Button` component for that using a proper `variant` prop?

src/js/components/stock-movement/modals/RejectRequestModalContent.jsx
```diff
@@ -0,0 +1,121 @@
...
+  currentUser,
+  requestor,
+  debounceTime,
+  minSearchLength,
+  closeRejectionModal,
+  rejectRequest,
+}) => {
+  const initialValues = {
+    sender: { id: currentUser?.id, label: currentUser?.name },
+    recipient: { id: requestor?.id, label: requestor?.name },
```
kchelstowski, line 61: I don't mind it, but personally I don't like having all properties in one line, so if it were up to me, I'd do it like:
```js
sender: {
  id: currentUser?.id,
  label: currentUser?.name,
}
```
etc.

We should set this rule on the ESLint I believe, not to mix the conventions.

src/js/components/stock-movement/modals/RejectRequestModalHeader.jsx
```diff
@@ -0,0 +1,22 @@
...
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+const RejectRequestModalHeader = ({ id }) => (
+  <div className="header">
+    <h4 className="title">
+      <Translate
+        id="react.autosaveFeatureModal.title.label"
+        defaultMessage={`Please provide a reason for rejecting request: ${id}`}
```
kchelstowski, line 12: At first I wanted to write that from the user perspective in my opinion it should display a code of a request, not id, but then I realized you do it, so maybe not to cause confusions either rename the prop to `identifier` (as it is returned from the backend) or even `requestCode`, but I think we should follow the first approach - so to keep the names the same backend <=> frontend - in this case -> `identifier`

src/js/components/stock-movement/modals/RejectRequestModalHeader.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+const RejectRequestModalHeader = ({ id }) => (
+  <div className="header">
+    <h4 className="title">
+      <Translate
+        id="react.autosaveFeatureModal.title.label"
```
awalkowiak, line 11: `react.autosaveFeatureModal.title.label`? You sure about that one?

awalkowiak (commented):
Other than what Kacper suggested and one minor label issue I think it lgtm

src/js/components/stock-movement/modals/RejectRequestModalButtons.jsx
alannadolny: That's the same button, so I am changing it.
image

src/js/components/stock-movement/modals/RejectRequestModalHeader.jsx
alannadolny: 🫣

src/js/components/stock-movement/modals/RejectRequestModal.scss
```diff
@@ -0,0 +1,119 @@
+.reject-request-modal-content {
+  position: absolute;
+  top: 20%;
+  left: 23%;
+  background-color: #ffffff;
+  width: 55vw;
+  padding: 20px;
+  display: flex;
+  flex-direction: column;
+}
```
drodzewicz, line 10: this also looks like a copy of `configuration-modal-content` class so maybe we should define the default styling for modals somewhere globaly.

src/js/components/stock-movement/modals/RejectRequestModal.scss
```diff
@@ -0,0 +1,119 @@
...
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(0, 0, 0, 0.45);
+}
+
+// Classic form
+.classic-form {
```
drodzewicz, line 22: as much as I would love to use some bootstrap styling or global classes here I understand that it is hard so this can stay.
I was hoping that we can use `.panel-body .classic-form` classes which are defined on stockMovement.scss but it is written specifically for that page, unfortunately...
If you can do anything to shorten this CSS and maybe use bootstrap classes in some places then it would be great, if not then I won't force you to, unless you want to define some global styles for form input yourself which we will be able to sue in the future.

drodzewicz (changes requested):

src/js/components/stock-movement/modals/RejectRequestModalHeader.jsx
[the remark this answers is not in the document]
kchelstowski: I like the approach of splitting those parts to separate components, because it'd be easier to debug in the future if we knew, that some error/unexpected behavior happens in the header.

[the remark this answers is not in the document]
drodzewicz: I don't mind the separate header component since it creates something like a semantic separation between header and content components, but my point was, how much benefit do we get from creating a new component for rendering 3 simple tags/components.
We need to be mindful of that and not create an endless nesting of new components.
In my opinion, a new component should be created if:
1. It holds a separate logic that can be moved out of the main component (following the principle of separation of concerns)
2. It is a reusable component that can be used in other places.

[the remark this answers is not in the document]
kchelstowski: I wouldn't agree with that, because it would mean that if a component has 3000 lines of code, but doesn't create any separate logic and is not reusable, it should all be done in file.
In general I agree, but readability is also important, hence I don't mind it being done the way Alan did.

drodzewicz (approved):

awalkowiak (approved):

PR #4333, opened 2023-10-16
OBPIH-5881 Fix consumption reports ignoring filter values set on first run

drodzewicz:


grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
```diff
@@ -60,7 +60,7 @@ class ConsumptionController {
         command.transactionTypes = command.defaultTransactionTypes
 
         // If any parameters have changed we need to reset filters
-        if (command.hasParameterChanged()) {
+        if (command.parametersHash && command.hasParameterChanged()) {
```
kchelstowski, line 63: lgtm, but I'm curious to know, what is that `parametersHash` and why adding this `command.parametersHash &&` fixes the issue?

awalkowiak (approved):

PR #4334, opened 2023-10-16
OBPIH-5847 Improve items ordering and derived status fetching on show PO page

awalkowiak:


grails-app/views/order/_orderAdjustments.gsp
```diff
@@ -17,7 +17,7 @@
             </tr>
             </thead>
             <tbody>
-            <g:each var="orderAdjustment" in="${orderInstance.orderAdjustments}" status="status">
+            <g:each var="orderAdjustment" in="${orderInstance.orderAdjustments?.sort { a,b -> a.dateCreated <=> b.dateCreated }}" status="status">
```
kchelstowski, line 20: nitpicky: I think `.sort { it.dateCreated }` should be enough (the same as you've done for `shipmentItem` in 21th line in `_orderShipments`.

kchelstowski (approved):

grails-app/views/order/_orderAdjustments.gsp
awalkowiak: @kchelstowski true, initially I added one more sort condition, but then I realized that for adjustment `dateCreated` is enough, and I forgot to make it back to `.sort { it.dateCreated }`

drodzewicz (approved):

PR #4336, opened 2023-10-17
OBPIH-5884 Add missing inventoryItem data on getStockMovementItem

drodzewicz:
When getting a list of stockMovementItems on inbound create page, inventoryItem variable was missing data which caused some validation issues.

I noticed that we are not assigning the inventoryItem on stockMovementItem for a case where SM is from requisition.
This is the logic how it;s done when SM is from shipment.
```groovy
 StockMovementItem stockMovementItem = StockMovementItem.createFromShipmentItem(shipmentItem)
                if (stockMovementItem.inventoryItem) {
                    def quantity = productAvailabilityService.getQuantityOnHand(stockMovementItem.inventoryItem)
                    stockMovementItem.inventoryItem.quantity = quantity
                }
                stockMovementItems.add(stockMovementItem)
```

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -749,6 +749,11 @@ class StockMovementService {
             }
             requisitionItems.each { requisitionItem ->
                 StockMovementItem stockMovementItem = StockMovementItem.createFromRequisitionItem(requisitionItem)
+                InventoryItem inventoryItem = inventoryService.findInventoryItemByProductAndLotNumber(stockMovementItem.product, stockMovementItem.lotNumber)
+                if (inventoryItem) {
+                    inventoryItem.quantity = productAvailabilityService.getQuantityOnHand(inventoryItem)
+                    stockMovementItem.inventoryItem = inventoryItem
+                }
                 stockMovementItems.add(stockMovementItem)
             }
         } else {


alannadolny (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -749,6 +749,15 @@ class StockMovementService {
             }
             requisitionItems.each { requisitionItem ->
                 StockMovementItem stockMovementItem = StockMovementItem.createFromRequisitionItem(requisitionItem)
+                InventoryItem inventoryItem = inventoryService.findOrCreateInventoryItem(
```
jmiranda, line 752: This does not feel safe in a method used to GET data. Can we just findInventoryItem(), and if we cannot find one, we either a) ignore and leave inventory item NULL OR b) blow things up with a validation error if that's more appropriate? I assume an inventory item is not always populated at certain points in the stock movement so maybe (a) is fine. 

I think we need to discuss these validation issues again in a tech huddle. I'm confused by what we're trying to accomplish. It seems the issue is on the React side because there's not enough data available which you're trying to solve by creating new data.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -749,6 +749,15 @@ class StockMovementService {
...
+                        stockMovementItem.product,
+                        stockMovementItem.lotNumber,
+                        stockMovementItem.expirationDate,
+                )
+                if (inventoryItem) {
+                    inventoryItem.quantity = productAvailabilityService.getQuantityOnHand(inventoryItem)
```
jmiranda, line 758: This is adding an N+1 query to all already dog slow method.

jmiranda (commented):
Let's discuss on a tech huddle.

jmiranda (approved):
As discussed in the ticket and Slack, the change here does not seem to add significantly to the response time of the stock movement items request, so it's ok to merge this. However, we should discuss the change with PIH and get their approval before doing so.

PR #4338, opened 2023-10-18
OBPIH-5882 Expiration date validation appears when entering new lot in ship PO

alannadolny:


grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -3432,4 +3436,22 @@ class InventoryService implements ApplicationContextAware {
...
+    Boolean areLotsValid(List<Object> items) {
+        return items?.every {
+            InventoryItem inventoryItem = InventoryItem.findByLotNumberAndProduct(it?.lotNumber, Product.findByProductCode(it?.productCode))
+
+            if (!inventoryItem) {
+                return true
+            }
+
+            Integer availableQuantity = productAvailabilityService.getQuantityOnHand(inventoryItem)
+            Date expirationDate = it?.expirationDate != JSONObject.NULL ? Date.parse("MM/dd/yyyy", it?.expirationDate) : null
```
alannadolny, line 3449: I think that I saw this `JSONObject.NULL` causing some issues, but I am curious how we should handle this in Grails 1 in this case?

cc @kchelstowski @drodzewicz @awalkowiak

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -3432,4 +3436,22 @@ class InventoryService implements ApplicationContextAware {
             }
         }
     }
+
+    Boolean areLotsValid(List<Object> items) {
```
jmiranda, line 3440: validateInventoryItems

grails-app/controllers/org/pih/warehouse/api/InventoryItemApiController.groovy
```diff
@@ -0,0 +1,30 @@
...
+import org.codehaus.groovy.grails.web.json.JSONObject
+import org.pih.warehouse.inventory.InventoryService
+
+class InventoryItemApiController {
+
+    InventoryService inventoryService
+
+    def validateLot = {
+        JSONObject jsonObject = request.JSON
+        Boolean areLotsValid = inventoryService.areLotsValid(jsonObject?.items)
```
jmiranda, line 21: Let's just make this an action "validateInventoryItems" or "validateProductLots" instead of create a new logic operator. Also let's call product availability service since we can implement the logic there without needing to add the circular dependency between inventory service and product availability service.
```
Boolean isValid = productAvailabilityService.validateInventoryItems(jsonObject.inventoryItems)
```

grails-app/controllers/org/pih/warehouse/api/InventoryItemApiController.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+        if (!areLotsValid) {
+            render(status: 400)
```
jmiranda, line 24: Should we have an error response? This is going to be part of our API so it should be well-designed and work intuitively for all use cases, not just this very specific feature.

grails-app/conf/UrlMappings.groovy
```diff
@@ -776,6 +776,12 @@ class UrlMappings {
             action = [GET: "listOfDemoData"]
         }
 
+        // Inventory Item
+        "/api/inventoryItem/validateLot" {
```
jmiranda, line 780: Lots to dig into here ...

The first issue is the API should be plural
```
/api/inventoryItems/validateLot
```
Secondly, as @awalkowiak knows, I really don't like remote procedure calls (RPC) masquerading as REST API requests. If this is a REST API, the resource should always be a noun. 

So the next question is ... do we even need the extra verb (validateLot)? Why can't we use a GET to retrieve the inventory item associated with the product and lot number.
```
/api/products/:productId/inventoryItems/:lotNumber
```
The API returns the details, including the quantity on hand from product availability and allow the client to decide whether it's valid or not. 

If we really want to encapsulate validation rules on the API, we probably want a resource to represent the validation 
https://stackoverflow.com/questions/11676550/how-to-expose-a-validation-api-in-a-restful-way

or we can use the given resource and a different verb e.g. HEAD or OPTIONS
https://softwareengineering.stackexchange.com/questions/347148/restful-service-for-checking-if-email-is-unique

jmiranda: One last thing ... if we add the InventoryItemApiController, it should extend BaseDomainApiController so that we get the benefit of the rest of the verbs.

grails-app/controllers/org/pih/warehouse/api/InventoryItemApiController.groovy
jmiranda: Why are we passing multiple items?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -3432,4 +3436,22 @@ class InventoryService implements ApplicationContextAware {
...
+        return items?.every {
+            InventoryItem inventoryItem = InventoryItem.findByLotNumberAndProduct(it?.lotNumber, Product.findByProductCode(it?.productCode))
+
+            if (!inventoryItem) {
+                return true
+            }
+
+            Integer availableQuantity = productAvailabilityService.getQuantityOnHand(inventoryItem)
```
jmiranda, line 3448: This is confusing. Quantity on Hand != Quantity Available. Variable name should reflect the data we are working with.

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -583,18 +584,25 @@ class AddItemsPage extends Component {
...
+    // we want to validate the newly inputted date
+    const itemsForValidation = lineItems.filter(item =>
+      (item?.expirationDate !== item?.inventoryItem?.expirationDate ||
+        item?.lotNumber !== item?.inventoryItem?.lotNumber) &&
+      item?.inventoryItem?.quantity !== '0');
+
+    if (itemsForValidation.length) {
+      try {
+        // If the lot is invalid the response status will be 400, so it will be caught
+        await InventoryItemApi.validateLot({ items: itemsForValidation });
```
jmiranda, line 598: Why aren't we just validating the requisition items in the next step?

jmiranda (changes requested):
I don't know the right way to do this because this is not what I thought we were trying to do here. I thought the validation would happen as the user enters the data, not when they click the Next button. We probably want to discuss this a bit more unless you think you have a way out of this.

PR #4343, opened 2023-10-19
OBPIH-5882 Expiration date validation appears when entering new lot in ship PO (fetching inventory item by entered lot number)

alannadolny:
It's the second version of the PR with a validation expiration date. In this PR I am fetching inventory items by the currently entered lot number and by the product code. Based on that I am doing the validation on the frontend part. Inventory item is fetched after losing focus on the lot number field or after 1 second (debounced fetch). I do not fix the requested changes in the previous PR, because I don't know if this version looks better than the previous one. If the previous one is better let me know, and I will fix those requested changes.

Previous PR: https://github.com/openboxes/openboxes/pull/4338


src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -584,17 +602,57 @@ class AddItemsPage extends Component {
...
+        ...previousState.values,
+        lineItems: mappedLineItems,
+      },
+    }));
+  }
+
+  fetchMissingInventoryItems(values) {
+    values.lineItems.forEach((item, index) => {
+      if (!item.fetchedInventoryItem) {
+        this.fetchInventoryItem(values, index);
```
kchelstowski, line 653: this would mean, the amount of sent requests is equal to lineItems. 
Can't we make a batch request to fetch all missing at once?

kchelstowski: second question - what will the "missing inventory items" be? when will they be "missed"?
I assume that when you add a new one, it can't be missed, because you trigger the fetch immediately, don't you?
Unless by "missing inventory items" we mean items, that have been added there "some time ago" and we come back to the shipment to add some more?

kchelstowski (commented):
if no one will look into that, I will come back to it tomorrow, since I got lost in the middle of trying to understand what is going on there and what we want to achieve.

kchelstowski:
+ a loose thought - I think validating it only on the frontend side might be dangerous - assuming you will create a shipment a few minutes, the quantity on hand of for example first item might've already changed and we wouldn't catch that.

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -584,17 +602,57 @@ class AddItemsPage extends Component {
...
+      item?.fetchedInventoryItem?.quantityOnHand > 0)) {
+      this.confirmInventoryItemExpirationDateUpdate(() =>
+        this.saveRequisitionItemsAndTransitionToNextStep(formValues, lineItems));
+      return;
+    }
+    this.saveRequisitionItemsAndTransitionToNextStep(formValues, lineItems);
+  }
+
+  async fetchInventoryItem(values, rowIndex, isDebounced = false) {
+    if (isDebounced) {
```
awalkowiak, line 617: Why you can't just call `fetchInventoryItem` or `debouncedInventoryItem` when needed? I feel like this nesting one into another will make a headache in the future.

awalkowiak (changes requested):

awalkowiak (approved):

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -584,17 +602,52 @@ class AddItemsPage extends Component {
...
-        this.confirmInventoryItemExpirationDateUpdate(() =>
-          this.saveRequisitionItemsAndTransitionToNextStep(formValues, lineItems));
-      } else {
-        this.saveRequisitionItemsAndTransitionToNextStep(formValues, lineItems);
-      }
-    } else {
-      this.saveRequisitionItemsAndTransitionToNextStep(formValues, lineItems);
+    const expirationDateOrLotChanged = _.some(lineItems, (item) => {
+      const inventoryItem = item?.fetchedInventoryItem?.inventoryItem || item?.inventoryItem;
+      const quantity = item?.fetchedInventoryItem?.quantity || item?.inventoryItem?.quantity;
```
awalkowiak, line 607: @alannadolny please make sure that this `item?.inventoryItem?.quantity` is always populated with the proper quantity on hand from the product availability (and where it is populated). If that isn't correct qoh, then I think there will be a need to fetch inventory item data again when the user changes the expiry date for some row after **re-entering** the page (**not** after adding new row)

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -274,4 +275,13 @@ class ProductApiController extends BaseDomainApiController {
         render([monthlyDemand: demand.monthlyDemand, quantityOnHand: quantityOnHand] as JSON)
     }
 
+    def getInventoryItem = {
+        InventoryItem inventoryItem = InventoryItem.findByLotNumberAndProduct(params.lotNumber, Product.findByProductCode(params.productCode))
```
jmiranda, line 279: Nitpick: Change the order of the parameters.
```
findByProductAndLotNumber(Product.findByProductCode(params.productCode), params.lotNumber)
```

jmiranda: With that said, sending the product ID in the API request would probably be better so we don't have to look up the product and then look up the inventory item.
```
InventoryItem.findByLotNumberAndProduct(Product.load(params.productId), params.lotNumber)
```

grails-app/conf/UrlMappings.groovy
```diff
@@ -101,6 +101,11 @@ class UrlMappings {
             controller = { "productApi" }
         }
 
+        "/api/products/$productCode/inventoryItems/$lotNumber"(parseRequest: true) {
```
jmiranda, line 104: Sending the ID instead of productCode would save us avoid an extra query on the database.

grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
```diff
@@ -274,4 +275,13 @@ class ProductApiController extends BaseDomainApiController {
...
+        if (inventoryItem) {
+            Integer quantityOnHand = productAvailabilityService.getQuantityOnHand(inventoryItem)
+            render([inventoryItem: inventoryItem, quantityOnHand: quantityOnHand] as JSON)
```
jmiranda, line 282: This is fine for now, but ...

I'd like to return a single inventory item JSON object.
```
inventoryItem.quantityOnHand = productAvailabilityService.getQuantityOnHand(inventoryItem)
render([inventoryItem:inventoryItem])
```
This would require a small refactoring that I don't think I want to do right now. 
InventoryItem.toJson() already includes the quantities, but Bootstrap.groovy defines its own marshaller for InventoryItem, so we'd need to fix the quantities returned by InventoryItem.toJson() method and also register the improved method as the default object marshaller. 

```
JSON.registerObjectMarshaller(InventoryItem) { InventoryItem inventoryItem ->
    return inventoryItem.toJson()
}
```

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
jmiranda: Yeah, we should always distinguish between quantityAvailable and quantityOnHand.

jmiranda (changes requested):
Looks good, just some small changes.

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
alannadolny: @awalkowiak 
image
The quantity that we are using on fetching line items is from `StockMovementService` line 777. So, it's quantity on hand. When the inventory items are new, the quantity is null instead of 0, that's the reason it didn't work when I was showing that during the last tech huddle.

alannadolny:
I added the improvements that were discussed during the tech huddle and I fixed the requested changes. I recorded a video showing how this validation looks right now: https://jam.dev/c/3c83e7a8-d9da-47b0-a9c8-8f1389c2699d

I pasted the link under the ticket also.

jmiranda (approved):
The API side LGTM. 

@alannadolny Please add the toJson() refactoring I mentioned to the Tech Debt backlog. Just link directly to that comment.

@awalkowiak I just realized a way to deal with tech debt. If developers don't fix tech debt issues in the given PR then they are responsible for creating and owning the tech debt until it's complete.

alannadolny:
Created tech debt ticket: https://pihemr.atlassian.net/browse/OBPIH-5924?atlOrigin=eyJpIjoiYTVhMDU2M2QyNDJmNDZkN2E4YmIzNGZiMmM2NjgxNzAiLCJwIjoiaiJ9

alannadolny:
@awalkowiak yup, you're right it should be targeting the grails 3 feature branch.

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -505,7 +544,7 @@ class AddItemsPage extends Component {
    * @public
    */
   fetchLineItems() {
-    const url = `/api/stockMovements/${this.state.values.stockMovementId}/stockMovementItems?stepNumber=2`;
+    const url = `/openboxes/api/stockMovements/${this.state.values.stockMovementId}/stockMovementItems?stepNumber=2`;
```
drodzewicz, line 547: I see there are a lot of URL that you have updated (added context _/openboxes_).
Since there is no time pressure to release this fix, maybe you can move these URLs to `api/url.js`.

It could also be done in some separate spike ticket, but if you can then please do it.

drodzewicz (approved):
Overall LGTM.
Consider the comment I made about moving URLs

awalkowiak (changes requested):
@alannadolny please remove the hardcoded `/openboxes` context path from urls and this one is good to go.

kchelstowski (approved):

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
jmiranda: @drodzewicz @alannadolny If this didn't get done we should create a tech debt ticket for any remaining issues. Or we should keep the "replace context path" ticket open until we're ready for testing.

alannadolny: @jmiranda I did it in this PR

PR #4344, opened 2023-10-19
OBPIH-5910 Add grouped items by product and UoM to PO template

awalkowiak:


grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -94,11 +94,11 @@ class DocumentTemplateService {
...
     private createOrderContext(IXDocReport report, Order orderInstance) {
-
-        // Add data to the context
-        def orderItems = orderInstance?.orderItems?.findAll {
+        def activeOrderItems = orderInstance?.orderItems?.findAll {
             it.orderItemStatusCode != OrderItemStatusCode.CANCELED
-        }?.collect { OrderItem orderItem ->
+        }
+        // Add data to the context
+        def orderItems = activeOrderItems?.collect { OrderItem orderItem ->
```
jmiranda, line 101: Why was this change needed? I think the code was more readable with the chaining.

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -130,11 +130,28 @@ class DocumentTemplateService {
...
+                quantity: v?.sum { it.quantity },
+                totalPrice: v?.sum { it.totalPrice() }
+            ]
+        }
 
         def order = orderInstance.toJson()
         order.dateOrdered = orderInstance.dateOrdered
         order.dateApproved = orderInstance.dateApproved
         order.dateCompleted = orderInstance.dateCompleted
+        order.currentDate = new Date()
```
jmiranda, line 154: :confused: 

This should be added to the context on its own. Call it today or now.

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -184,15 +201,26 @@ class DocumentTemplateService {
         metadata.addFieldAsList("orderItems.totalPrice")
         metadata.addFieldAsList("orderItems.expectedShippingDate")
 
+        metadata.addFieldAsList("groupedOrderItems.code")
```
jmiranda, line 204: The naming is fine for now. 

But just want to mention that if we end up supporting different grouping methods then we'll need to come up with a convention to distinguish between these different representations. Coming up with a convention now would prevent us from having to refactor a bunch of document templates in the future.

orderItemsByCategory
orderItemsByProductCode
orderItemsByProductCodeAndUom

jmiranda: I don't love variable names like `orderItemsBy...` but just wanted to point out that we're painting ourselves into a corner with `groupedOrderItems`

jmiranda (changes requested):
LGTM, just a few minor changes. Feel free to ignore or make the name change for groupedOrderItems if you think of anything clever.

PR #4348, opened 2023-10-23
OBPIH-5909 Fix list table pagination reset on filter change

drodzewicz:
The solution which I implemented in my previous PR
```js
tableRef.current.onPageChange(0);
tableRef.current.fireFetchData();
```

introduced I bug which caused the table to double fetch the data when resetting the pagination to the first page.
The issue seems to be that, `onPageChange(0)` triggers the `fireFetchData()` which would result in two `fireFetchData()` API calls to race (one with offset 0 and another with previous offset value).

At first, I thought that replacing `tableRef.current.fireFetchData();` with  `tableRef.current.onPageChange(0);` would solve the problem, but `onPageChange(0)` does not trigger the fireFetch on the initial load, basically it does not trigger the fireFetchData when we are trying to changePage to the same value as the currentPage.

I hope the comment I left in the code explains clearly enough the reasoning for this approach.

### DIFF
--- a/src/js/hooks/list-pages/useTableData.jsx
+++ b/src/js/hooks/list-pages/useTableData.jsx
@@ -43,8 +43,14 @@ const useTableData = ({
     // Each time we fetch, we want to 'reset' the token/signal
     sourceRef.current = CancelToken.source();
     // reset pagination on each search execution
-    tableRef.current.onPageChange(0);
-    tableRef.current.fireFetchData();
+    if (tableRef.current?.state?.page > 0) {
+      // onPageChange(pageIndex) triggers fireFetchData() when pageIndex !== currenPage
+      // which is why we are calling onPageChange(0) and fireFetchData() separately
+      // by doing that we are trying to avoid double fetching
+      tableRef.current.onPageChange(0);
+    } else {
+      tableRef.current.fireFetchData();
+    }
   };
 
   // If filterParams change, refetch the data with applied filters


src/js/hooks/list-pages/useTableData.jsx
```diff
@@ -43,8 +43,14 @@ const useTableData = ({
...
-    tableRef.current.onPageChange(0);
-    tableRef.current.fireFetchData();
+    if (tableRef.current?.state?.page > 0) {
+      // onPageChange(pageIndex) triggers fireFetchData() when pageIndex !== currenPage
+      // which is why we are calling onPageChange(0) and fireFetchData() separately
+      // by doing that we are trying to avoid double fetching
+      tableRef.current.onPageChange(0);
+    } else {
+      tableRef.current.fireFetchData();
+    }
```
kchelstowski, line 53: a question (shout at me if I just don't understand the context), but if `onPageChange(0)` anyway triggers the `fireFetchData`, can't we just remove the `tableRef.current.fireFetchData()` from line 47th, and that would be it?

drodzewicz: It is being triggered by useEffect.
The useEffect is calling `fireFetchData()` on the initial load and when filters change.
https://github.com/openboxes/openboxes/blob/a4a07bf1ee4fe01e6699d41b884d8ed4dcbcd66c/src/js/hooks/list-pages/useTableData.jsx#L57

awalkowiak (approved):

PR #4350, opened 2023-10-23
OBPIH-5771 Fix ordering by year on requisition by year indicator

drodzewicz:


grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -451,17 +451,13 @@ class IndicatorDataService {
             } else {
                 results[yearLabel] = it[0]
             }
-
-            // Check if the list with labels contains yearLabel
-            if (!listLabel.contains(yearLabel)) {
-                listLabel << yearLabel
-            }
         }
+        results = results.sort()
```
kchelstowski, line 455: out of curiosity - what is it sorted by? I ask, because I don't see what type is hidden under the `results`

kchelstowski (approved):

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
drodzewicz: `results` is of type `LinkedHashMap`, apparently by default it sorts its values by keys

awalkowiak (approved):

PR #4353, opened 2023-10-24
OBGM-264 Fix additional columns for CSV in consumption report

kchelstowski:
### Issues:

1. `selectedProperties` not being binded in `ShowConsumptionCommand` command in `show` method
2. getting rid of deprecated in Grails 3: `new DefaultGrailsDomainClass(Product.class)` and replacing it with `Holders.grailsApplication.mappingContext.getPersistentEntity(Product.class.name).persistentPropertyNames`

### Solutions:

1. using TYPE instead of `def`, because `def` is not binded for command since Grails 2: [Properties which are not bindable by default are those related to transient fields, dynamically typed properties and static properties.](https://grails.github.io/grails2-doc/2.5.6/ref/Constraints/bindable.html)
2. The problem was that 1. we were using persistentProperty**Names**, not persistent**Properties**, but even with that, we didn't have access to fields such as `typePropertyName` and `naturalName` which we had in Grails 1.
  In order to achieve the same behavior, I've just created static methods which implement the same behavior from Grails 1, by just checking with "CTRL" what is the implementation of `typePropertyName` and `naturalName` in Grails 1:

Screenshot from 2023-10-24 12-22-13
Screenshot from 2023-10-24 12-23-02

### General conclusions:
Do not use `def` when it is not needed. It drove me nuts to find why the `selectedProperties` was not binded for `show` method. 
I would never assume that because of the type, it won't work. 
We have typed language, and we should in my opinion use `def` only when it is necessary (when a type of variable can change, e.g. stockMovement stuff - either it can be `OutboundStockMovement` or `StockMovement`, not to cause such unexpected behaviors like this one.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
@@ -10,19 +10,12 @@
 package org.pih.warehouse.reporting
 
 import grails.converters.JSON
-import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
-import grails.util.GrailsNameUtils
-import grails.util.Holders
 import grails.validation.Validateable
 import groovy.time.TimeCategory
 import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.collections.FactoryUtils
 import org.apache.commons.collections.list.LazyList
-import org.apache.commons.lang.ClassUtils
-import org.grails.core.DefaultGrailsDomainClass
-import org.grails.datastore.mapping.model.PersistentEntity
-import org.grails.datastore.mapping.model.PersistentProperty
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Tag
--- a/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
@@ -512,23 +512,6 @@ class ShowConsumptionCommand implements Validateable {
         parametersHash(nullable: true)
     }
 
-    static String getTypePropertyName(Class type) {
-        String shortTypeName = ClassUtils.getShortClassName(type)
-        return shortTypeName.substring(0, 1).toLowerCase(Locale.ENGLISH) + shortTypeName.substring(1)
-    }
-
-    static List<ConsumptionReportProductProperty> getAvailableProperties() {
-        List<PersistentProperty> properties =
-            Holders.grailsApplication.mappingContext.getPersistentEntity(Product.class.name).persistentProperties
-        return properties.collect { PersistentProperty it ->
-            new ConsumptionReportProductProperty([
-                name: it.name,
-                naturalName: GrailsNameUtils.getNaturalName(it.name),
-                typePropertyName: getTypePropertyName(it.type)
-            ])
-        }
-    }
-
     Boolean hasParameterChanged() {
         String newParametersHash = generateParametersHash()
         return !parametersHash.equals(newParametersHash)
--- a/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
@@ -531,12 +531,6 @@ class ShowConsumptionCommand implements Validateable {
     }
 }
 
-class ConsumptionReportProductProperty {
-    String name
-    String naturalName
-    String typePropertyName
-}
-
 class ShowConsumptionRowCommand implements Validateable {
 
     Product product
--- a/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
@@ -12,14 +12,17 @@ package org.pih.warehouse.reporting
 import grails.converters.JSON
 import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
+import grails.util.GrailsNameUtils
 import grails.util.Holders
 import grails.validation.Validateable
 import groovy.time.TimeCategory
 import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.collections.FactoryUtils
 import org.apache.commons.collections.list.LazyList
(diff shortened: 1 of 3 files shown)


grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
```diff
@@ -509,8 +512,21 @@ class ShowConsumptionCommand implements Validateable {
         parametersHash(nullable: true)
     }
 
-    def getAvailableProperties() {
-        Holders.grailsApplication.mappingContext.getPersistentEntity(Product.class.name).persistentPropertyNames
+    static String getTypePropertyName(Class type) {
```
drodzewicz, line 515: if we are going with this approach I think we need to add a comment above pointing out that these methods are copied from Grails 1 due to the fact that they are not present in Grails 3.

I would also suggest to move/create a utils class for this since this looks like a generic method that all of the `Command` classes are missing.

grails-app/views/consumption/_filters.gsp
```diff
@@ -131,12 +131,12 @@
                         <g:hasRoleFinance>
                             <g:set var="hasRoleFinance" value="${true}"/>
                         </g:hasRoleFinance>
-                        <g:each var="propertyName" in="${command.availableProperties}">
-                            <g:set var="disabled" value="${'pricePerUnit'.equals(propertyName) && !hasRoleFinance}"/>
-                            <option value="${propertyName}"
-                                ${command.selectedProperties?.toList()?.contains(propertyName)?'selected':''}
+                        <g:each var="property" in="${command.availableProperties}">
```
drodzewicz, line 134: I am also wondering if we could go with a different approach and instead of bringing these methods from Grails 1, get a list of `persistentProperties`  and map them to a `messeges.properties` label so we have these option labels translatable.

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
kchelstowski: I don't know if there is a sense to create a util class for that, because consumption report is the only place where it is used. Nowhere else this is needed.

jmiranda (changes requested):
Good detective work, @kchelstowski.

jmiranda:
@kchelstowski One minor bug is that the selected properties are not bound on the first request

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
```diff
@@ -509,8 +512,21 @@ class ShowConsumptionCommand implements Validateable {
...
+        String shortTypeName = ClassUtils.getShortClassName(type)
+        return shortTypeName.substring(0, 1).toLowerCase(Locale.ENGLISH) + shortTypeName.substring(1)
+    }
+
+    static List<ConsumptionReportProductProperty> getAvailableProperties() {
+        List<PersistentProperty> properties =
+            Holders.grailsApplication.mappingContext.getPersistentEntity(Product.class.name).persistentProperties
+        return properties.collect { PersistentProperty it ->
+            new ConsumptionReportProductProperty([
```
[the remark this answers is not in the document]
kchelstowski: The reason I've created a class for that one is only for "IDE suggestion purposes". If I returned a generic Map here, I'd have to return `List<Map>`, so that the compiler would not know what properties I might have here, so I wouldn't have suggestions in the `_filters.gsp` about `name`, `naturalName` etc.
Let me know if I should change it to the generic Map, or I can proceed with the class if figured out a better name,

kchelstowski:
@jmiranda I've watched the video, but I can't see you clicking the submit (Run report), but probably refreshing the page :thinking: ?
Moreover - I believe it should work only for `CSV` format (as it is mentioned)  and you have `HTML` chosen. 
By the way - for me even after refreshing the page, they are not cleared, so could you please give me some more details if there really is a bug?

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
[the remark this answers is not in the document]
awalkowiak: I feel like we have a ton of other places with this "issue". And trying to address it in some places and leaving it in others will cause some chaos. Let's use a generic map for now, and we can discuss this topic on the TechHuddle as a tech debt idea.

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
awalkowiak: +1 for the util

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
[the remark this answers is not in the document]
awalkowiak: (unless you already have something better in mind than `ConsumptionReportProductProperty` and `DomainClassProperty` that can be used as a util class, then you can use it)

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
kchelstowski: Moved to a util and made generic so that `getAvailableProperties` can accept a domain class.

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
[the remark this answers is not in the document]
kchelstowski: Went with generic Map for now. Will create a tech debt ticket to handle those.

grails-app/controllers/org/pih/warehouse/reporting/ConsumptionController.groovy
```diff
@@ -539,6 +531,12 @@ class ShowConsumptionCommand implements Validateable {
     }
 }
 
+class ConsumptionReportProductProperty {
```
awalkowiak, line 534: Is this still required?

kchelstowski: No, it is not, forgot to remove it 🙈

kchelstowski: removed

awalkowiak (approved):

PR #4354, opened 2023-10-25
OBGM-711 Empty active field read as inactive in product sources import

alannadolny:
The default value for this property was set in the domain:
image

In Grails 1 when we don't provide value for the active column, we don't have this property in the `params` object, so in this line:
`productSupplier = new ProductSupplier(params)`
we used the default value from the domain.

In Grails 3 when we don't provide the value, we have `active: null` in the params object, so the `true` value from the domain is overridden by the null value and this is treated as a false.

Besides the solution that I pushed, I am thinking also about:
- Just write:
```
productSupplier.active = params["active"] ? params["active"] : true
```
but I think the current solution is much cleaner.
- Use the `@BindUsing` annotation and check if we provide the value inside it.
```
    @BindUsing({ obj, source -> source['active] ? source['active'] : true })
    Boolean active = Boolean.TRUE
```
Let me know what you think about those approaches.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
@@ -65,7 +65,7 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
     Date dateCreated
     Date lastUpdated
 
-    @BindUsing({ obj, source -> source["active"] != null ? source["active"] : true })
+    @BindUsing({ productSupplier, source -> source["active"] != null ? source["active"] : true })
     Boolean active = Boolean.TRUE
 
     static transients = ["defaultProductPackage", "globalProductSupplierPreference", "attributes"]
--- a/grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
@@ -9,6 +9,7 @@
  **/
 package org.pih.warehouse.product
 
+import grails.databinding.BindUsing
 import org.pih.warehouse.core.Organization
 import org.pih.warehouse.core.ProductPrice
 import org.pih.warehouse.core.RatingTypeCode
@@ -64,6 +65,7 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
     Date dateCreated
     Date lastUpdated
 
+    @BindUsing({ obj, source -> source["active"] != null ? source["active"] : true })
     Boolean active = Boolean.TRUE
 
     static transients = ["defaultProductPackage", "globalProductSupplierPreference", "attributes"]
--- a/src/main/groovy/org/pih/warehouse/importer/ProductSupplierExcelImporter.groovy
+++ b/src/main/groovy/org/pih/warehouse/importer/ProductSupplierExcelImporter.groovy
@@ -48,7 +48,7 @@ class ProductSupplierExcelImporter extends AbstractExcelImporter {
     ]
 
     static Map propertyMap = [
-            active                                : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            active                                : ([expectedType: ExpectedPropertyType.StringType, defaultValue: true]),
             id                                    : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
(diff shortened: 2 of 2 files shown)


kchelstowski (approved):
I'm fine with all approaches mentioned by you, but `BindUsing` would probably make us safer for other options of adding a product source - current solution only aims for import.
I'm wondering what happens when creating a product source manually - is it also sent as `active: null` or?

jmiranda:
@alannadolny Can you explain what changed in our code for these two statements to be true. In other words, why is the active param not included in Grails 1, but it is in Grails 3.

> In Grails 1 when we don't provide value for the active column, we don't have this property in the params object, so in this line:

> In Grails 3 when we don't provide the value, we have active: null in the params object, so the true value from the domain is overridden by the null value and this is treated as a false.

src/main/groovy/org/pih/warehouse/importer/ProductSupplierExcelImporter.groovy
```diff
@@ -48,7 +48,7 @@ class ProductSupplierExcelImporter extends AbstractExcelImporter {
     ]
 
     static Map propertyMap = [
-            active                                : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            active                                : ([expectedType: ExpectedPropertyType.StringType, defaultValue: true]),
```
jmiranda, line 51: My issue with this solution is that we've defined a default value in two places. I would prefer to define the default in one place, so I think using @BindUsing gets us closer to that outcome. However, I'm still trying to understand why this is even an issue, so please respond to my other question when you get a chance.

jmiranda (changes requested):

src/main/groovy/org/pih/warehouse/importer/ProductSupplierExcelImporter.groovy
alannadolny: @jmiranda This change comes from `ExcelImportService`.
We are calling the `convertColumnMapConfigManyRows` function from this service and this function calls `convertColumnMapOneRow` to get the value for a single row. This function iterates over the column map for that row and for each cell checks:
```
if (value == null || ''.equals(value)) { //cheking for null, because otherwise 0 value will fail here
        log.debug "Value for column $columnName row $rowIndex is null or empty String.  Was trying to find property $propertyName.  Skipping setting its value in param map"
        returnParams[propertyName] = propertyConfigurationMap?.get(propertyName)?.defaultValue
} else {
        anyNonNullValues = true
        returnParams[propertyName] = value
}
```
and because of the line following line, it is included in our params map.
```
returnParams[propertyName] = propertyConfigurationMap?.get(propertyName)?.defaultValue
```
In Grails 1 the same part of the code looks like this:
```
if (value==null || ''.equals(value)) { //cheking for null, because otherwise 0 value will fail here
    log.info "Value for column $columnName row $rowIndex is null or empty String.  Was trying to set $propertyName.  Skipping setting its value in param map"
} else {
    returnParams[propertyName] = value
}
```

drodzewicz (approved):

src/main/groovy/org/pih/warehouse/importer/ProductSupplierExcelImporter.groovy
jmiranda: Thank you @alannadolny. That makes sense.

jmiranda (changes requested):

kchelstowski (approved):

jmiranda:
@kchelstowski why did this get merged? I requested changes?

alannadolny:
@jmiranda If you mean changing the name from `obj` to `productSupplier`, I made this change. Or is there something more that needs to be fixed?

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
```diff
@@ -64,6 +65,7 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
     Date dateCreated
     Date lastUpdated
 
+    @BindUsing({ obj, source -> source["active"] != null ? source["active"] : true })
```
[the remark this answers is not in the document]
jmiranda: @alannadolny This comment ^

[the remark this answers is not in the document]
jmiranda: Using the elvis operator to make it more readable.

PR #4355, opened 2023-10-25
OBGM-705 Fix refresh of product availability after merging a product

kchelstowski:
In the ticket you will find details about the fix, especially the one in the `InventoryService`, because magic happens there.

### DIFF
--- a/grails-app/domain/org/pih/warehouse/product/ProductAvailability.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductAvailability.groovy
@@ -60,4 +60,88 @@ class ProductAvailability {
     Boolean isPickable() {
         return (inventoryItem ? inventoryItem.pickable : true) && (binLocation ? binLocation.pickable : true)
     }
+
+    @Override
+    boolean equals(o) {
+        if (this.is(o)) {
+            return true
+        }
+        if (getClass() != o.class) {
+            return false
+        }
+
+        ProductAvailability that = (ProductAvailability) o
+
+        if (binLocation != that.binLocation) {
+            return false
+        }
+        if (binLocationName != that.binLocationName) {
+            return false
+        }
+        if (dateCreated != that.dateCreated) {
+            return false
+        }
+        if (id != that.id) {
+            return false
+        }
+        if (inventoryItem != that.inventoryItem) {
+            return false
+        }
+        if (lastUpdated != that.lastUpdated) {
+            return false
+        }
+        if (location != that.location) {
+            return false
+        }
+        if (lotNumber != that.lotNumber) {
+            return false
+        }
+        if (product != that.product) {
+            return false
+        }
+        if (productCode != that.productCode) {
+            return false
+        }
+        if (quantityAllocated != that.quantityAllocated) {
+            return false
+        }
+        if (quantityAvailableToPromise != that.quantityAvailableToPromise) {
+            return false
+        }
+        if (quantityNotPicked != that.quantityNotPicked) {
+            return false
+        }
+        if (quantityOnHand != that.quantityOnHand) {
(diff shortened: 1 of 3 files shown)


chetanmaharishi (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1231,7 +1231,10 @@ class InventoryService implements ApplicationContextAware {
                     def quantityOnHand = quantityBinLocationMap[product][inventoryItem][binLocation]
 
                     def quantityAvailableMap = quantityAvailableInventoryItemMap[product][inventoryItem]
-                    def quantityAvailable = quantityAvailableMap ? quantityAvailableMap[binLocation] : 0
+                    def quantityAvailable =
+                        quantityAvailableMap
+                            ? quantityAvailableMap.find { key, value -> key?.id == binLocation?.id }?.value
```
jmiranda, line 1236: Given that this is probably the most important (and worst performing) algorithm in the system (my bad), I'm not sure we want to change the lookup from constant (1) to linear (N) here. 

Is there another way to deal with this?

grails-app/domain/org/pih/warehouse/product/ProductAvailability.groovy
```diff
@@ -60,4 +60,88 @@ class ProductAvailability {
...
+            return true
+        }
+        if (getClass() != o.class) {
+            return false
+        }
+
+        ProductAvailability that = (ProductAvailability) o
+
+        if (binLocation != that.binLocation) {
+            return false
```
jmiranda, line 76: The order of these should probably matters. I imagine we want id to be first and only, then we compare the objects that are used to define uniqueness (product, inventory item, location, bin location, etc). The quantity, dates, and indexable fields seem like overkill so let me know the justification for including them. Could we ever be in a situation where the unique fields aren't enough? 

We'll need to look at the data but the equals implementation might be simpler if we use those indexable fields (productCode, lotNumber, binLocationName, etc) instead of the objects (product, inventoryItem, binLocation). But I worry about consistency with default values around lotNumber (where we use empty string, NULL, etc) so again we probably need to look at the data there. We don't want two inventory items with different IDs (one with NULL and one with empty string) being unequal when comparing the IDs but equal when comparing the lot number (i.e. assuming there's a case where empty string and NULL are equal).

jmiranda (commented):
I'm not sure if I'm requesting a change, but it would be great to utilize one of the many code assist libraries we have at our disposal (groovy AST, Lombok). I moved OBGM-314 to MVP2 Could because I didn't want to implement these for ALL of our domain classes (due to the ramifications for testing). But I think it would make sense to get the ball rolling for specific classes where we know we are encountering an issue with equals and hashCode.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
awalkowiak: I think taking `binLocation?.id` (or `null`) as a key might do the work here. Depending on how much you have to refactor.

kchelstowski: Fortunately after diving into the "source" of that Map it turned out that I could've safely changed the key to be the `id`, not the whole object.

grails-app/domain/org/pih/warehouse/product/ProductAvailability.groovy
kchelstowski: For now I've reverted the `equals`/`hashCode` implementation and I rely on checking `.id`.

awalkowiak (approved):

PR #4356, opened 2023-10-26
OBPIH-5884 Fix duplicating items and expiry date not saving on expirydate validation

drodzewicz:
- use saveRequisitionItemsInCurrentStep instead of saveRequisitionItems when clicking next button which updates lineItems with data from response
- add additional condition for lineItemsToBeUpdated when expiryDate on item and inventoryItem do not match

I decided to remove `saveRequisitionItems` method since the only place where it was used was in `saveAndTransitionToNextStep` which I updated to use `saveRequisitionItemsInCurrentStep` instead.

`saveRequisitionItemsInCurrentStep` and `saveRequisitionItems` are nearly identical, the difference is that `saveRequisitionItemsInCurrentStep` update the lineItems state with the response that it gets form the API

### DIFF
--- a/src/js/components/stock-movement-wizard/inbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/inbound/AddItemsPage.jsx
@@ -276,6 +276,9 @@ class AddItemsPage extends Component {
         lineItemsToBeUpdated.push(item);
       } else if (newQty !== oldQty || newRecipient !== oldRecipient) {
         lineItemsToBeUpdated.push(item);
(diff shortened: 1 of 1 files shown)


drodzewicz:
cherry picked commits to a new branch which will go to OBGM feature branch https://github.com/openboxes/openboxes/pull/4359
It was decided that this will not go into release and instead will be part of 0.9.0 version thus I am closing this pull request.

drodzewicz:
cherry picked commits to a new branch which will go to OBGM feature branch https://github.com/openboxes/openboxes/pull/4359
It was decided that this will not go into release and instead will be part of 0.9.0 version thus I am closing this pull request.

PR #4358, opened 2023-10-27
OBGM-699 SPIKE investigate other places with hardcoded context path

drodzewicz:
This Pull Request has quite a lot of file changes but all of them are quite simple.

In the scope of this PR I created a centralized URL management system where we can define all of the URLs in `applicationUrls.js` file and use these variables throughout our applications instead of hardcoded links.
This way, we have a better and easier control of the links that we use in our application.

on `BrowserRouter` I have set a **basename** of `CONTEXT_PATH` which prefixes all of our routes when using `history` or `react-router-dom` **Link** components.
For cases where we redirect to `GSP` page we must wrap the link with `stringUrlsInterceptor()` util.

As an improvement in cases where we redirect to React page I refactored usage of HTML anchor tags with either `history.push` or `<Link />` component.


src/js/api/urls.js
```diff
@@ -69,3 +69,10 @@ export const PUTAWAY_GENERATE_PDF = id => `/putAway/generatePdf/${id}`;
 
 // SUPPORT LINKS
 export const SUPPORT_LINKS = `${CONTEXT_PATH}${API}/supportLinks`;
+
+export const HELPSCOUT_CONFIGURATION = `${CONTEXT_PATH}${API}/helpscout/configuration/`;
```
alannadolny, line 73: why at the end of this URL is `/`? As I can see we are not using this at the end of other links.

src/js/components/Layout/Footer.jsx
```diff
@@ -59,7 +60,7 @@ const Footer = ({
               <a
                 className={`${locale === language.code ? 'selected' : ''}`}
                 key={language.code}
-                href={`/openboxes/user/disableLocalizationMode?locale=${language.code}`}
+                href={`${DISABLE_LOCALIZATION}?locale=${language.code}`}
```
alannadolny, line 63: Maybe we can move this `locale=${language.code}` to `DISABLE_LOCALIZATION`? I mean adding sth like:
```
export const DISABLE_LOCALIZATION = (languageCode) => `${CONTEXT_PATH}/user/disableLocalizationMode${languageCode ? `?locale=${languageCode}` : ''}`;
```
What do you think?
Or if there are more than one param, we can create a util for mapping those params

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -340,7 +341,7 @@ class AddItemsPage extends Component {
   saveAndExit(formValues) {
     this.saveInvoiceItems(formValues)
       .then(() => {
-        window.location = stringUrlInterceptor(`/invoice/show/${formValues.id}`);
+        window.location = stringUrlInterceptor(INVOICE_URL.show(formValues.id));
```
alannadolny, line 344: You said in the description that we have to use `stringUrlInterceptor` for redirecting to list pages. Maybe in the redirect URLs, we can include the host name in the URL in the same way as we are passing `CONTEXT_PATH`?

Or maybe use the interceptor in the URL generator instead of using it on each page?

alannadolny (commented):
Great job!

src/js/api/urls.js
jmiranda: This is an external link so it doesn't really matter here as long as the URL works. Some web frameworks (Django for example) prefer (require?) the forward slash at the end. I think it's dumb but there seems to be a semantic reason for it.

https://devpress.csdn.net/python/630453e9c67703293080b33d.html

https://learndjango.com/tutorials/trailing-url-slashes-django

jmiranda: One small caveat is that if the external app is expecting a trailing slash and we don't provide one then the app sends a redirect to the browser with the trailing slash. So I don't actually mean "it doesn't matter" as there is overhead but for all intents and purposes it doesn't really matter.

jmiranda:
I think since this PR covers a lot of tech debt I'd like to bring it up on the next tech huddle to make sure we outline the conventions / expectations going forward.

src/js/components/invoice/AddItemsPage.jsx
jmiranda: I think I like the second suggestion best.

src/js/api/urls.js
kchelstowski: @jmiranda I think Alan's point was that we potentially chain a few consts like this:
```js
${API}/${SOMETHING}
```
if `${API}` ended with `/`, we'd have link like: `/api//something` (with two slashes), so it would be better to follow the convention instead of mixing stuff like:
```js
${API}${SOMETHING}
```
and 
```js
${API}/${SOMETHING}
```

src/js/components/purchaseOrder/PurchaseOrderListHeader.jsx
```diff
@@ -16,10 +18,10 @@ const PurchaseOrderListHeader = ({ history, supportedActivities }) => (
       <Button
         defaultLabel="Create Shipment from PO"
         label="react.purchaseOrder.createShipmentFromPo.label"
-        onClick={() => history.push('/openboxes/stockMovement/createCombinedShipments?direction=INBOUND')}
+        onClick={() => history.push({ pathname: STOCK_MOVEMENT_URL.createCombinedShipments(), search: 'direction=INBOUND' })}
```
kchelstowski, line 21: wouldn't it be possible for link to accept eventual search params instead? I guess it would be cleaner, but on the other hand I don't mind this version.

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -62,45 +64,45 @@ const PurchaseOrderListTable = ({
       label: 'react.purchaseOrder.viewOrderDetails.label',
       defaultLabel: 'View order details',
       leftIcon: <RiInformationLine />,
-      href: '/openboxes/order/show',
+      href: ORDER_URL.show,
```
kchelstowski, line 67: I think this might not work due to the same reason as for invoice actions I described in details above.

src/js/utils/ActionDots.jsx
```diff
@@ -38,6 +39,21 @@ const ActionDots = ({
...
+  const buildLink = (action) => {
+    if (typeof action.href === 'string') {
+      return action.href + (action.appendId === false ? '' : `/${id}`);
+    }
+    if (typeof action.href === 'function') {
+      return action.href(id);
+    }
+    return null;
+  };
+
```
kchelstowski, line 56: while giving the review from the top, I have not seen there.
I won't delete my comments, but they don't make sense anymore seeing this.
Anyway I have mixed feelings if we should some kind of parsing here, or make the links clear for everyone that we don't have to remember: "oh, here I have to use interceptor, and here I don't have to".
If it were up to me, I'd add the context path to the links and forget about `stringUrlInteceptor` and then eventually get rid of the context path if a page becomes a React one.

kchelstowski: the function itself also is difficult to read for me honestly. By doing it like `action.appendId === false`, wouldn't we have to make sure that in every action defined, we'd have to add `appendId: false`? 
I assume in most places we might not define it at all, so it would be `undefined`, and `undefined !== false` so we would append the id.

kchelstowski:
@jmiranda answering the:

> I know this probably has something to do with how the URLs are used (i.e. within a React component vs browser redirect), but it's not entirely clear why some URLs have context path and others don't.

It's because our `apiClient` that uses `axios`, uses an interceptor that adds the context path to the api calls:
```js
apiClient.interceptors.request.use(urlInterceptor, justRejectRequestError);
```

src/js/components/invoice/InvoiceListTable.jsx
```diff
@@ -34,19 +36,19 @@ const InvoiceListTable = ({
...
       leftIcon: <RiFileLine />,
-      href: '/openboxes/invoice/addDocument',
+      href: INVOICE_URL.addDocument,
     },
     {
       label: 'react.invoice.edit.label',
       defaultLabel: 'Edit Invoice',
       leftIcon: <RiPencilLine />,
-      href: '/openboxes/invoice/create',
+      href: INVOICE_URL.edit,
```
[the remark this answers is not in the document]
jmiranda: One thing I worry about with URLs is that we're creating a layer make our code more DRY and configurable, but at what cost? In this particular case we're losing readability by close to 100%, obfuscating what is a pretty standard and obvious thing (/openboxes/invoice/create) and making the reader/developer think "wtf?!?". 

We need to not do that.

src/js/components/invoice/InvoiceListTable.jsx
```diff
@@ -92,7 +94,7 @@ const InvoiceListTable = ({
       Header: <Translate id="react.invoice.column.invoiceNumber.label" defaultMessage="Invoice Number" />,
       accessor: 'invoiceNumber',
       sortable: false,
-      Cell: row => <TableCell {...row} link={`/openboxes/invoice/show/${row.original.id}`} />,
+      Cell: row => <TableCell {...row} link={stringUrlInterceptor(INVOICE_URL.show(row.original.id))} />,
```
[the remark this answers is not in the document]
jmiranda: Previous comment x 1000.

src/js/utils/ActionDots.jsx
jmiranda: I think I might have been the one to steer Artur in a direction toward the stringUrlInterceptor back in 2018. If that is so, I'll take the blame. 

I think we can all probably agree on which option below is more readable. Option 1 is more readable, but all probably not the right approach. 
```
href={window.CONTEXT_PATH + '/location/list')}
```
OR
```
href={stringUrlInterceptor(LOCATION_URL.list())}
```
We need to find a way to create URLs that are easy to read without obfuscating the code into oblivion.

What are other projects doing? Has there never been a project that dealt with this issue before? And if the answer is no, then why? If the answer to that is because we are colocating React code with server code in the same repository, then I'm going to vote for moving the React application into its own repository so we don't have to deal with these kinds of issues.

drodzewicz:
There are quite a lot of comments under this pull request, so instead of replying to every single one of them and creating multiple threads I'll try to respond to everything here.
After this PR is going to be resolved, I will move the most import points to the ticket comments.

### stringUrlInterceptor is an implementation detail hidden from the developer
I totally agree that this wrapper util function should be hidden from the developer, and the developer shouldn't waste time trying to figure out if he should include it or not when redirecting to a page.
Just as suggested in the comments, I will add **CONTEXT_PATH** prefix to the URL factory methods for links that redirect to GSP pages.
The important thing to point out, of why we need this prefix for GSP pages is that in case of redirects inside of the react app we should rely on the `Browser router` to handle the navigation (using` <Link />` and `history.push(...)`), since we have defined the context on the router, it preppends the URL with given `prefix`, but for non-react redirects we must use the javascript `windows.location`, `<a href />` which require a full URL.

### we don't have ${API} and ${CONTEXT_PATH} in every URL definition
For this Pull Request, I focused on tackling the problem with hardcoded page redirect URLs, I tried to catch all of the places where we had hardcoded strings and moved them into `consts/applciationURLs.js` for better maintainability in the future.
As for the other file, the `api/url.js`, it holds the endpoint URLs which we use for API calls. Cleaning it up should be done in another ticket which I will create.
In the end, all of the URLs should be defined without the context `/openboxes ` and `/api` prefixes, these prefixes should be set on the main instances handling the navigation or API calls.

### we have a simple API we can use to generate a URL with or without parameters
My goal was to create a URL factory function which would look something like `[namespace].action(parameters)` where 
 - **namespace**: StockMovement, Order, Invoice etc.. 
 - **action** : index, show, edit, create, edit
 - **parameters**: usually ids

I also wanted to keep the URLs as pure as possible, meaning that they should return only a pathname without any `search` parameters. I think that trying to handle search queries in these factory methods will result in the unnecessary complexity of these methods, but that's just my opinion.

### Needs to consider API versioning (at some point)
Just as I mentioned before, my focus here was tackling the hardcoded context in the navigation URLs and for the API calls we should leave this topic for another ticket.

### Completely removing the strinUrlInterceptor
There are still a few places that require this wrapper function, mainly for links that we receive from the backend (mostly dashboard card links) that do not include the context prefix. Until we decide to return links with with context included, the `stringUrlInterceptor` should stay in those components.

###
Additionally, I have removed the logic from `ActionDots` with `appendId: bool`. This prop was added when we were implementing list pages and by default, it appended the `id` of the item to the provided link (string). There were cases when we needed a different id than the one provided by default (eg. instead of provided SM id we needed Order id ) which is why this attribute was added.
Now since I implemented that `href` can receive a string or a function with **id** prop, we can use the function approach in these places instead of the `appendId = false` @kchelstowski

src/js/utils/ActionDots.jsx
awalkowiak: @jmiranda I still like having this centralized url factory, and using using "defined" URLs in href. Imho this is a fairly clean and way better approach than the `stringUrlInterceptor`.

src/js/utils/ActionDots.jsx
```diff
@@ -104,7 +119,6 @@ ActionDots.propTypes = {
     defaultLabel: PropTypes.string.isRequired,
     href: PropTypes.string,
     reactLink: PropTypes.bool,
-    appendId: PropTypes.bool,
```
kchelstowski, line 107: to clean everything out I think we could also remove the `appendId` from the places where you were sending it
e.g:
https://github.com/openboxes/openboxes/blob/9d783dcc5d668d05f244dd9036f6591241ce21f8/src/js/components/stock-movement/inbound/StockMovementInboundTable.jsx#L77

src/js/components/stock-movement-wizard/request/EditPage.jsx
```diff
@@ -1597,9 +1597,11 @@ class EditItemsPage extends Component {
               <button
                 type="button"
                 onClick={() => {
-                  window.location = !this.props.supportedActivities.includes('MANAGE_INVENTORY') && this.props.supportedActivities.includes('SUBMIT_REQUEST')
-                    ? stringUrlInterceptor('/dashboard')
-                    : stringUrlInterceptor('/stockMovement/list?direction=INBOUND');
+                  if (!this.props.supportedActivities.includes('MANAGE_INVENTORY') && this.props.supportedActivities.includes('SUBMIT_REQUEST')) {
+                    this.props.history.push(DASHBOARD_URL.base);
+                  } else {
+                    this.props.history.push({ pathname: STOCK_MOVEMENT_URL.list(), search: 'direction=INBOUND' });
```
kchelstowski, line 1603: I get your explanation why you did it this way, but what I've figured out now is that we could maybe just add another function like `.listInbound()` which would look like this?
```javascript
 list: `${STOCK_MOVEMENT_URL.base}/list`,
 inboundList: () => `${STOCK_MOVEMENT_URL.list}/?direction=${StockMovementDirection.INBOUND}`,
 outboundList: () => `${STOCK_MOVEMENT_URL.list}/?direction=${StockMovementDirection.OUTBOUND}`,
 requestList: () => `${STOCK_MOVEMENT_URL.list}/?direction=${StockMovementDirection.OUTBOUND}&sourceType=${RequisitionSourceType.ELECTRONIC}`,
```
later it could be even improved by creating a method to invoke the search params into url instead of hardcoding them like I did, but for now I think it'd be fine.

What do you think guys about this version instead of doing it like this:
```js
this.props.history.push({ pathname: STOCK_MOVEMENT_URL.list(), search: 'direction=INBOUND' });
```
do it like this:
```js
this.props.history.push(STOCK_MOVEMENT_URL.inboundList());
```

drodzewicz: yes, cases like **inbound**, **outbound**, and **request** list makes sense since we treat them as separate pages.
I will implement it just like you showed in the example

awalkowiak (changes requested):
@drodzewicz I have one issue with this one. While it is steering us into a good direction, we are heading there in "The Ministry of Silly Walks" way. The problem here is that we have a context path handling scattered now in four (?) different places on the front-end:
1. ApiClient
2. api/urls.js
3. consts/applicationUrls
4. BrowserRouter basename
This is getting hard to understand what, where, and why. And maintainability is getting harder and harder.

Initially, it was done in two places: request interceptor in `apiClient` and `stringUrlInterceptor`. The first one was a simple way for handling all API calls and the second one was for handling cases like href links, `window.location` changes or history push (cases where it did not come through centralized handling).

When @kchelstowski introduced `api/urls` I was happy, because we could easily get rid of **both** request interceptor and `stringUrlInterceptor`. I imagined this all to be in one place, as a simple: 

 ```javascript
 // pseudo example

const { CONTEXT_PATH } = window;

// whatever api endpoints
const API = `${CONTEXT_PATH}/api`;
const GENERIC_API = `${API}/generic`;

export const SOME_ENDPOINT_API_URL = `${API}/someEndpoint`;
export const SOME_ENDPOINT_DELETE = id => `${SOME_ENDPOINT_API_URL}/${id}`;
export const SOME_ENDPOINT_SHOW = id => `${SOME_ENDPOINT_API_URL}/${id}/show`;

// whatever non-api endpoint links
export const SOME_NON_API_URL = `${CONTEXT_PATH}/someNonApiUrl`;
export const ANOTHER_LINK_URL = `${CONTEXT_PATH}/anotherLinkUrl`;
 ```

Points to change or discuss:
1. Currently using `BrowserRouter basename` is not simplifying anything but rather complicating. I would get rid of it until we have a full front-end covered with React.
2. I am ok with leaving separate api urls and application urls, but I think I'd prefer having these two file in one place (i guess `consts` dir is more suitable).
3. Each base link in the applicationsUrls should be using context path (at least for now, while we cannot properly use use `basename`)
4. api urls don't have to be handled/refactored yet, because the request interceptor is handling it (it is sanitazing unnecesary hardcoded context path and adding a proper one).

cc @jmiranda @kchelstowski @alannadolny

src/js/consts/applicationUrls.js
```diff
@@ -35,6 +35,9 @@ const LOCATION_URL = {
 const STOCK_MOVEMENT_URL = {
   base: '/stockMovement',
   list: () => `${STOCK_MOVEMENT_URL.base}/list`,
+  listInbound: () => `${STOCK_MOVEMENT_URL.base}/list?direction=INBOUND`,
+  listOutbound: () => `${STOCK_MOVEMENT_URL.base}/list?direction=OUTBOUND`,
+  listRequest: () => `${STOCK_MOVEMENT_URL.base}/list?direction=OUTBOUND&sourceType=ELECTRONIC`,
```
kchelstowski, line 40: why can't it be like I've shown? I mean, using the `${STOCK_MOVEMENT_URL.list}/?direction=StockMovementDirection.INBOUND` instead of hardcoding `/list` in every link?

kchelstowski: `STOCK_MOVEMENT_URL.list` at this point feels unnecessary if we don't use it for `listInbound` and the rest

src/js/components/dashboard/GraphCard.jsx
```diff
@@ -250,17 +251,18 @@ const GraphCard = SortableElement(({
   return (
     <div className={`graph-card ${size === 'big' ? 'big-size' : ''} ${cardType === 'error' ? 'error-card' : ''}`}>
       <div className="header-card">
-        {cardLink ?
-          <a target="_blank" rel="noopener noreferrer" href={stringUrlInterceptor(cardLink.code)} className="title-link">
+        {cardLink ? (
+          <Link target="_blank" rel="noopener noreferrer" to={cardLink} className="title-link">
```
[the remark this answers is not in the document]
drodzewicz: `cardLink` doesn't have the property `code` and the link was not working before

kchelstowski (approved):

awalkowiak (approved):

src/js/components/dashboard/GraphCard.jsx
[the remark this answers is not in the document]
awalkowiak: makes sense, thanks :)

PR #4359, opened 2023-10-27
OBPIH-5884 when re enter inbound does not validate expiry change

drodzewicz:


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -740,6 +740,11 @@ class StockMovementService {
             }
             requisitionItems.each { requisitionItem ->
                 StockMovementItem stockMovementItem = StockMovementItem.createFromRequisitionItem(requisitionItem)
+                InventoryItem inventoryItem = inventoryService.findInventoryItemByProductAndLotNumber(stockMovementItem.product, stockMovementItem.lotNumber)
+                if (inventoryItem) {
+                    inventoryItem.quantity = productAvailabilityService.getQuantityOnHand(inventoryItem)
+                    stockMovementItem.inventoryItem = inventoryItem
+                }
```
kchelstowski, line 747: @jmiranda do you feel confident in pushing that one through, since this method can become costly? 
We have a few DB calls (including PA one) in a loop.
I wanted to make sure I have a green light before merging this one.

kchelstowski: @jmiranda the change seem to be very similar to the one @alannadolny did in this PR: https://github.com/openboxes/openboxes/pull/4343
the difference is that in Alan's PR we do it for one item, and here, we might do it for many of them.

jmiranda: @kchelstowski Did you miss the tech huddles on this? 

We discussed this extensively last week (see discussions in the ticket https://pihemr.atlassian.net/browse/OBPIH-5884) and decided that while it wasn't the ideal approach, it didn't seem to add a huge amount of overhead to each API request. It will almost certainly create an N+1 query and cause a performance delay but @drodzewicz tested on a 300 line order and the added time was negligible. I was mostly worried about the fact that this API is used in places other than inbound and that we'd see some bigger performance issues there. But I've been told

With that said, we should monitor performance of this end point and this should definitely be added to th top of our tech debt backlog. The stock movement API, in general, needs to be rearchitected and the stock movement items sub-resource, in particular, is one of the first APIs I'd like to see refactored.

src/js/components/stock-movement-wizard/inbound/AddItemsPage.jsx
```diff
@@ -597,7 +600,7 @@ class AddItemsPage extends Component {
   saveAndTransitionToNextStep(formValues, lineItems) {
     this.props.showSpinner();
 
-    this.saveRequisitionItems(lineItems)
+    this.saveRequisitionItemsInCurrentStep(lineItems)
```
jmiranda, line 603: Why did we need to change this method?

jmiranda: Never mind, I see there are two methods. I would prefer keeping the name saveRequisitionItems rather than saveRequisitionItemsInCurrentStep().

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
kchelstowski: @jmiranda that must’ve been the tech huddles where I was not present due to involvement in another project, so thanks for the explanation. That sounds good to me now.

kchelstowski (approved):

PR #4360, opened 2023-10-27
OBPIH-5936 Add tests for REST APIs to avoid regression

chetanmaharishi:
First commit of Rest assured testing for ProductApi list

### DIFF
--- a/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
@@ -5,43 +5,113 @@ import io.restassured.specification.RequestSpecification
 import org.testng.Assert
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test
+import spock.lang.Shared
 import spock.lang.Specification;
 
 import static io.restassured.RestAssured.*;
 import io.restassured.response.Response;
 
-class ProductApiTest  extends Specification {
+class ProductApiTest extends Specification {
 
-    Map headers = [:]
+    @Shared
     String baseURI = "http://localhost:8080/openboxes";
 
+    @Shared
+    Map headers = [:]
+
+    @Shared
+    boolean setupInitialized = false
+
+    @Shared
+    String productId
+
+    @Shared
+    String locationId = 1
 
     void setup() {
-        println "inside setup"
-        RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
-        Response response = httpRequest.post(baseURI + "/auth/handleLogin");
-        System.out.println(response.statusCode());
-        System.out.println(response.getBody().asString());
-        response.headers().each {
-            if (it.name == "Set-Cookie") {
-                headers["Cookie"] = it.value.substring(0, (it.value.indexOf(";") + 1))
+        if (!setupInitialized) {
+            RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
+            Response response = httpRequest.post(baseURI + "/auth/handleLogin");
+            response.headers().each {
+                if (it.name == "Set-Cookie") {
+                    headers["Cookie"] = it.value.substring(0, (it.value.indexOf(";")))
+                }
             }
+            httpRequest = given().header("Cookie", "${headers.Cookie}");
+            response = httpRequest.get(baseURI + "/dashboard/chooseLocation/${locationId}");
+            setupInitialized = true
         }
+
     }
 
     void "read product list"() {
         when: "/api/products end point called"
-            RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
-            Response response = httpRequest.get(baseURI + "/api/products");
-            // Get JSON Representation from Response Body
-            JsonPath jsonPathEvaluator = response.jsonPath();
-            // Get specific element from JSON document
+        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
+        Response response = httpRequest.get(baseURI + "/api/products");
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath();
+        // Get specific element from JSON document
         then: "List of products should return"
-            List data = jsonPathEvaluator.getList("data");
-            int statusCode = response.getStatusCode();
-        and: "response status should 200"
-            Assert.assertEquals(statusCode, 200);
-        and: "data list size should be 1"
-            Assert.assertEquals(data.size(), 1);
+        List data = jsonPathEvaluator.getList("data");
+        int statusCode = response.getStatusCode();
+        and:
+        Assert.assertEquals(statusCode, 200, "response status should 200");
+        and:
+        Assert.assertEquals(data.size(), 1, "data list size should be 1");
+        when:
+        def product = data.get(0)
+        productId = product.id
+        then:
+        Assert.assertNotNull(productId, "Product id should not be null")
     }
+
+    void "get demand of product"() {
+        when:
+        "api/products/${productId}/demand called"
+        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
+        Response response = httpRequest.get(baseURI + "/api/products/${productId}/demand");
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath();
+        then:
+        def data = jsonPathEvaluator.get("data")
+        Assert.assertNotNull(data)
+        and:
+        Assert.assertNotNull(data.product, "Response should have product object")
+        and:
+        Assert.assertEquals(data.product.id, productId, "Product id should match with requested id:${productId}")
+        and:
+        Assert.assertNotNull(data.demand, "Demand Object should not be null")
+    }
+
+    void "get demand summary of product"() {
+        when:
+        "api/products/${productId}/demandSummary called"
+        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
+        Response response = httpRequest.get(baseURI + "/api/products/${productId}/demandSummary");
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath();
+        then:
+        def data = jsonPathEvaluator.getList("data")
+        Assert.assertNotNull(data)
+        and:
+        Assert.assertEquals(data.size(), 0, "Demand summary should be empty")
+    }
+
+    void "get summary of product"() {
+        when:
+        "api/products/${productId}/productSummary called"
+        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
+        Response response = httpRequest.get(baseURI + "/api/products/${productId}/productSummary");
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath();
+        then:
+        def data = jsonPathEvaluator.get("data")
+        Assert.assertEquals(data.product.id, productId, "Product id should be ${productId}")
+        and:
+        Assert.assertEquals(data.location.id, locationId, "Location id should be ${locationId}")
+        and:
+        Integer quantityOnHand = data.quantityOnHand ?: 0
+        Assert.assertEquals(quantityOnHand, 0, "QuantityOnHand should be 0")
+    }
+
 }
--- a/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
@@ -4,41 +4,44 @@ import io.restassured.path.json.JsonPath
 import io.restassured.specification.RequestSpecification
 import org.testng.Assert
 import org.testng.annotations.BeforeClass;
-import org.testng.annotations.Test;
+import org.testng.annotations.Test
+import spock.lang.Specification;
 
 import static io.restassured.RestAssured.*;
 import io.restassured.response.Response;
-class ProductApiTest{
+
+class ProductApiTest  extends Specification {
 
     Map headers = [:]
     String baseURI = "http://localhost:8080/openboxes";
 
 
-    @BeforeClass
-    public void setup() {
+    void setup() {
         println "inside setup"
         RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
-        Response response = httpRequest.post(baseURI+"/auth/handleLogin");
+        Response response = httpRequest.post(baseURI + "/auth/handleLogin");
         System.out.println(response.statusCode());
         System.out.println(response.getBody().asString());
         response.headers().each {
-            if(it.name == "Set-Cookie"){
-                headers["Cookie"] = it.value.substring(0, (it.value.indexOf(";")+1))
+            if (it.name == "Set-Cookie") {
+                headers["Cookie"] = it.value.substring(0, (it.value.indexOf(";") + 1))
             }
         }
-
     }
 
-    @Test
-    public void test() {
-        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
-        Response response = httpRequest.get(baseURI+"/api/products");
-        // Get JSON Representation from Response Body
-        JsonPath jsonPathEvaluator = response.jsonPath();
-        // Get specific element from JSON document
-        List data = jsonPathEvaluator.getList("data");
-        int statusCode = response.getStatusCode();
-        Assert.assertEquals(statusCode, 200);
-        Assert.assertEquals(data, []);
+    void "read product list"() {
+        when: "/api/products end point called"
+            RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
+            Response response = httpRequest.get(baseURI + "/api/products");
+            // Get JSON Representation from Response Body
+            JsonPath jsonPathEvaluator = response.jsonPath();
+            // Get specific element from JSON document
+        then: "List of products should return"
+            List data = jsonPathEvaluator.getList("data");
+            int statusCode = response.getStatusCode();
+        and: "response status should 200"
+            Assert.assertEquals(statusCode, 200);
+        and: "data list size should be 1"
+            Assert.assertEquals(data.size(), 1);
     }
 }
--- a/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
@@ -1,6 +1,10 @@
 package unit.org.pih.warehouse.api
 
+import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.SerializationConfig
 import groovy.transform.builder.Builder
 import io.restassured.RestAssured
 import io.restassured.builder.RequestSpecBuilder
@@ -12,11 +16,17 @@ import lombok.AllArgsConstructor
 import lombok.Data
 import lombok.NoArgsConstructor
 import org.grails.web.json.JSONObject
+import org.junit.Test
+import org.pih.warehouse.product.ProductType
+import org.pih.warehouse.product.ProductTypeCode
 import spock.lang.Shared
 import spock.lang.Specification
+import util.StringUtil
+
 import static io.restassured.RestAssured.*
 import io.restassured.response.Response
 import static org.hamcrest.MatcherAssert.assertThat
+import static org.hamcrest.Matchers.equalTo
 import static org.hamcrest.Matchers.is
 import static org.hamcrest.Matchers.notNullValue
 import static org.hamcrest.Matchers.lessThanOrEqualTo
@@ -43,6 +53,12 @@ class ProductApiTest extends Specification {
 
     @Shared
     RequestSpecification requestSpecification
+
+    @Shared
+    String categoryId
+
+    @Shared
+    String productTypeId
 
     // Executed before the first spock test is executed
     void setupSpec() {
@@ -77,6 +93,31 @@ class ProductApiTest extends Specification {
                 setAccept(ContentType.JSON).build()
     }
 
+    void "should have at-least one product type"(){
+        given: "an authenticated user"
+        RequestSpecification request = given().spec(requestSpecification).log().all()
+
+        when: "we get products from the API"
+        // FIXME figure out better way to set up baseURI using RestAssured API
+        Response response = request.when().get("/api/generic/productType?max=10")
+        println "response.body():${response.body().toString()}"
+
+        // FIXME Should be replaced by Product domain class once we
+        List<ProductTypeResponse> productTypes = response.then().log().all().extract().body().jsonPath().
+                getList("data", ProductTypeResponse.class)
+
+        then: "expected a 200 response with at most 10 products"
+        response.then().assertThat().statusCode(200)
+
+        // Get a single product to pass to the next test
+        when: "we attempt to share a product with the next test"
+        ProductTypeResponse productType = response.then().body().extract().jsonPath().getObject("data[0]", ProductTypeResponse)
+        productTypeId = productType.id
+        then: "productType should not be null"
+        assertThat(productTypeId, is(notNullValue()))
+    }
+
+    @Test
     void "should return a list of products"() {
         given: "an authenticated user"
         RequestSpecification request = given().spec(requestSpecification).log().all()
@@ -98,6 +139,8 @@ class ProductApiTest extends Specification {
         when: "we attempt to share a product with the next test"
         ProductResponse product = response.then().body().extract().jsonPath().getObject("data[0]", ProductResponse)
         productId = product.id
+        categoryId = product.categoryId
+        println "categoryId:${categoryId}, productId:${productId}"
 
         then: "product should not be null"
         assertThat(productId, is(notNullValue()))
@@ -117,14 +160,6 @@ class ProductApiTest extends Specification {
         then:
         response.then().assertThat().statusCode(200)
 
-//        then: "data should not be null"
-//        Assert.assertNotNull(data)
-//        and: "product should not be null"
-//        Assert.assertNotNull(data.product, "Response should have product object")
-//        and: "product ID should not be null"
-//        Assert.assertEquals(data.product.id, productId, "Product id should match with requested id:${productId}")
-//        and: "demand should not be null"
-//        Assert.assertNotNull(data.demand, "Demand Object should not be null")
     }
 
     void "get demand summary of product"() {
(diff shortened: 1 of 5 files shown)


awalkowiak (commented):
@chetanmaharishi please don't use `;` at the end of lines and add spaces around operators, before opening `{`, and after the `if` keyword.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,44 @@
+package unit.org.pih.warehouse.api
+
+import io.restassured.path.json.JsonPath
+import io.restassured.specification.RequestSpecification
+import org.testng.Assert
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static io.restassured.RestAssured.*;
+import io.restassured.response.Response;
```
kchelstowski, line 10: 1. remove `;` (in all lines above also)
2. make an indent between `import` and `class`

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,44 @@
+package unit.org.pih.warehouse.api
+
+import io.restassured.path.json.JsonPath
+import io.restassured.specification.RequestSpecification
+import org.testng.Assert
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static io.restassured.RestAssured.*;
```
kchelstowski, line 9: I would say importing everything from a package is not a good practice. We should import only what is needed. (Remove `*` and replace it with imports that are needed)

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+import io.restassured.response.Response;
+class ProductApiTest{
+
+    Map headers = [:]
+    String baseURI = "http://localhost:8080/openboxes";
```
kchelstowski, line 14: I don't think if hardcoding this url is a good idea, but since I am not a good test-writer, I don't know if we can get an access to context path at this point.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+
+    @BeforeClass
+    public void setup() {
+        println "inside setup"
+        RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
+        Response response = httpRequest.post(baseURI+"/auth/handleLogin");
+        System.out.println(response.statusCode());
+        System.out.println(response.getBody().asString());
```
kchelstowski, line 23: do we want the printlns to stay?

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+        Response response = httpRequest.get(baseURI+"/api/products");
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath();
+        // Get specific element from JSON document
+        List data = jsonPathEvaluator.getList("data");
+        int statusCode = response.getStatusCode();
+        Assert.assertEquals(statusCode, 200);
+        Assert.assertEquals(data, []);
+    }
+}
```
kchelstowski, line 44: add an empty line at the end of the file

kchelstowski (changes requested):
and as @awalkowiak said, add indents between `{`, `(`, `)`, `+` and remove `;` as it is not necessary to write them in Groovy.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Thanks @kchelstowski I was just going to mention this.

jmiranda: Note that the Rest Assured tests are executed in an isolated execution context (either as a separate process or on a separate machine) so it's sort of like running postman tests. Therefore, the API instance will be up and running before these tests are run which means we have the ability to change this URL when we execute the tests. We could even move these tests to a separate github repository to make it more clear that the tests are somewhat independent from the API. However, I don't want to do that because I want them to evolve together so we can create new tests in feature branches that live alongside the code.

I'm currently reading Testing Web APIs to see if there are any best practices to keep in mind when running these Rest Assured tests https://www.manning.com/books/testing-web-apis but I suspect that we should at least provide a default API URL in configuration that can be overridden by a system environment variable or external configuration file.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Nope. They should be removed.

jmiranda (commented):
Added some preliminary comments for now.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+    @Shared
+    String productId
+
+    @Shared
+    String locationId = 1
+
+    void setup() {
+        if (!setupInitialized) {
+            RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
+            Response response = httpRequest.post(baseURI + "/auth/handleLogin");
```
jmiranda, line 34: Use the /api/login so you can pass the location ID there.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+            httpRequest = given().header("Cookie", "${headers.Cookie}");
+            response = httpRequest.get(baseURI + "/dashboard/chooseLocation/${locationId}");
+            setupInitialized = true
+        }
+
+    }
+
+    void "read product list"() {
+        when: "/api/products end point called"
+        RequestSpecification httpRequest = given().header("Cookie", "${headers.Cookie}");
```
jmiranda, line 49: Make sure you pass an Accept header to ensure you get a JSON response.
```
RequestSpecification httpRequest = given().header("Accept", "application/json")
    .header("Cookie", "${headers.Cookie}")
```
Otherwise we get this when running tests
```
io.restassured.path.json.exception.JsonPathException: Failed to parse the JSON document
	at io.restassured.path.json.JsonPath$ExceptionCatcher.invoke(JsonPath.java:930)
	at io.restassured.path.json.JsonPath$4.doParseWith(JsonPath.java:895)
	at io.restassured.path.json.JsonPath$JsonParser.parseWith(JsonPath.java:975)
	at io.restassured.path.json.JsonPath.get(JsonPath.java:201)
	at unit.org.pih.warehouse.api.ProductApiTest.get demand of product(ProductApiTest.groovy:76)
Caused by: groovy.json.JsonException: Lexing failed on line: 1, column: 1, while reading '<', no possible valid JSON value or punctuation could be recognized.
	at groovy.json.JsonLexer.nextToken(JsonLexer.java:87)
	at io.restassured.internal.path.json.ConfigurableJsonSlurper.parse(ConfigurableJsonSlurper.groovy:97)
	at io.restassured.internal.path.json.ConfigurableJsonSlurper.parseText(ConfigurableJsonSlurper.groovy:83)
	at io.restassured.path.json.JsonPath$4$1.method(JsonPath.java:893)
	at io.restassured.path.json.JsonPath$ExceptionCatcher.invoke(JsonPath.java:928)
	... 4 more
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+    @Shared
+    String productId
+
+    @Shared
+    String locationId = 1
+
+    void setup() {
+        if (!setupInitialized) {
+            RequestSpecification httpRequest = given().formParam("username", "admin").formParam("password", "password");
```
jmiranda, line 33: Username and password should be @Shared and should probably be configurable like the URL.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
chetanmaharishi: ok i will make this changes

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
chetanmaharishi: println statements are removed

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
chetanmaharishi: ok

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
chetanmaharishi: ok

jmiranda:
Can you guys review this PR again? And if possible, help us figure out what to do about the failing unit tests. Chetan thinks these REST API tests should be executed as unit tests because they don't require the application to be started i.e. we can tests another OpenBoxes instance somewhere else. Moving the tests to the integration test phase will create a lot more overhead, but I can't see another way of segmenting these REST API tests from the unit tests.

cc @awalkowiak @kchelstowski @alannadolny @drodzewicz

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+import io.restassured.response.Response
+import static org.hamcrest.MatcherAssert.assertThat
+import static org.hamcrest.Matchers.is
+import static org.hamcrest.Matchers.notNullValue
+import static org.hamcrest.Matchers.lessThanOrEqualTo
+
+class ProductApiTest extends Specification {
+
+    @Shared
+    String baseURI = System.getProperty("server.host")?:"http://localhost"
```
alannadolny, line 27: Please add whitespace after and before the Elvis operators
```suggestion
    String baseURI = System.getProperty("server.host") ?: "http://localhost"
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+    @Shared
+    String basePath = System.getProperty("server.path")?:"/openboxes"
+
+    @Shared
+    int port = System.getProperty("server.port")?Integer.valueOf(System.getProperty("server.port")):8080
```
alannadolny, line 33: Add the whitespace after and before the ternary operators
```suggestion
    int port = System.getProperty("server.port") ? Integer.valueOf(System.getProperty("server.port")) : 8080
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+    void setupSpec() {
+
+        RestAssured.basePath = basePath
+        RestAssured.baseURI = baseURI
+        RestAssured.port = port
+
+        JSONObject jsonObject = new JSONObject().
+                put("username", username).
+                put("password", password).
+                put("location", 1)
```
alannadolny, line 57: Those dots shouldn't be at the beginning of the next lines?
```suggestion
        JSONObject jsonObject = new JSONObject()
                .put("username", username)
                .put("password", password)
                .put("location", 1)
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+                getList("data", ProductResponse.class)
+
+        then: "expected a 200 response with at most 10 products"
+        response.then().assertThat().statusCode(200)
+        assertThat(products, is(notNullValue()))
+        assertThat(products.size(), is(lessThanOrEqualTo(10)))
+
+        // Get a single product to pass to the next test
+        when: "we attempt to share a product with the next test"
+        ProductResponse product = response.then().body().extract().jsonPath().getObject("data[0]", ProductResponse)
```
alannadolny, line 99: Shouldn't we add new lines when we are calling multiple methods at once? I think we should try to avoid long lines like that (for example, on the frontend we have configured eslint with a maximum length of 100). It is harder to read for example in code review, when the line is too long.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+        response.then().assertThat().statusCode(200)
+
+//        then: "data should not be null"
+//        Assert.assertNotNull(data)
+//        and: "product should not be null"
+//        Assert.assertNotNull(data.product, "Response should have product object")
+//        and: "product ID should not be null"
+//        Assert.assertEquals(data.product.id, productId, "Product id should match with requested id:${productId}")
+//        and: "demand should not be null"
+//        Assert.assertNotNull(data.demand, "Demand Object should not be null")
```
alannadolny, line 127: Should it be commented?

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+class ProductResponse {
+    String id
```
alannadolny, line 175: Add new line
```suggestion
class ProductResponse {

    String id
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+import lombok.NoArgsConstructor
+import org.grails.web.json.JSONObject
+import spock.lang.Shared
+import spock.lang.Specification
+import static io.restassured.RestAssured.*
+import io.restassured.response.Response
+import static org.hamcrest.MatcherAssert.assertThat
+import static org.hamcrest.Matchers.is
+import static org.hamcrest.Matchers.notNullValue
+import static org.hamcrest.Matchers.lessThanOrEqualTo
```
alannadolny, line 22: I think `import static` should be placed below the "normal" imports. This is also the default behavior when we are adding new imports through Intellij. And also we shouldn't import things using `*`, because there can be a lot of unnecessary imported things (I think this change request was mentioned before, but I am not sure)
```suggestion
import io.restassured.response.Response

import static io.restassured.RestAssured.*
import static org.hamcrest.MatcherAssert.assertThat
import static org.hamcrest.Matchers.is
import static org.hamcrest.Matchers.notNullValue
import static org.hamcrest.Matchers.lessThanOrEqualTo
```

alannadolny (commented):
I will get back to this later to try to check the failing unit tests. (Unless someone else comes up with the solution  😄)

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+import io.restassured.http.Cookie
+import io.restassured.path.json.JsonPath
+import io.restassured.specification.RequestSpecification
+import lombok.AllArgsConstructor
+import lombok.Data
+import lombok.NoArgsConstructor
+import org.grails.web.json.JSONObject
+import spock.lang.Shared
+import spock.lang.Specification
+import static io.restassured.RestAssured.*
```
kchelstowski, line 17: the `*` imports have not been fixed (I've mentioned that in my initial review)

kchelstowski: I wish we had something like pmd checkstyle (I don't know if it is available in Groovy/Grails), but it would catch those things.
I also think that the imports should be grouped differently - the static imports should be at the beginning and be separated with one empty line, but since we don't have any strict rules regarding that, you can skip that grouping - just wanted to mention that.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
kchelstowski: remember about the other two below also

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        JSONObject jsonObject = new JSONObject().
+                put("username", username).
+                put("password", password).
+                put("location", 1)
+
+        Response response = given().log().all().
+                contentType(ContentType.JSON).
+                body(jsonObject.toString()).
+                post("/api/login")
```
kchelstowski, line 62: here dots also should be at the beginning of the line

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        // Make sure we were successfully authenticated
+        response.then().assertThat().statusCode(200)
```
kchelstowski, line 65: it would be good to do it by one line -> one call:
```groovy
response
  .then()
  .assertThat()
  .statusCode(200)
```

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+        Cookie cookie = response.getDetailedCookie("JSESSIONID")
+
+        // FIXME For some reason ReqestSpecBuilder doesn't respect the config settings above
+        // so we need to set the baseURI, basePath, port again
+        requestSpecification = new RequestSpecBuilder().
+                setBaseUri(baseURI).
+                setBasePath(basePath).
+                setPort(port).
+                addCookie(cookie).
+                setAccept(ContentType.JSON).build()
```
kchelstowski, line 77: dots

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    void "should return a list of products"() {
+        given: "an authenticated user"
+        RequestSpecification request = given().spec(requestSpecification).log().all()
```
kchelstowski, line 82: chain

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        when: "we get products from the API"
+        // FIXME figure out better way to set up baseURI using RestAssured API
+        Response response = request.when().get("/api/products?max=10")
```
kchelstowski, line 86: chain

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        // FIXME Should be replaced by Product domain class once we
+        List<ProductResponse> products = response.then().log().all().extract().body().jsonPath().
```
kchelstowski, line 89: chain

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+        Response response = httpRequest.get("/api/products/{productId}/demandSummary")
+        // Get JSON Representation from Response Body
+        JsonPath jsonPathEvaluator = response.jsonPath()
+        def data = jsonPathEvaluator.getList("data")
+
+        then:
+        response.then().assertThat().statusCode(200)
+//        Assert.that(is(notNullValue(data)))
+//        and:
+//        Assert.that(is(data.size(), equalTo(0)))
```
kchelstowski, line 145: should it removed/uncommented?

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
```diff
@@ -0,0 +1,178 @@
...
+        def data = jsonPathEvaluator.get("data")
+
+        then:
+        response.then().assertThat().statusCode(200)
+//        Assert.assertEquals(data.product.id, productId, "Product id should be ${productId}")
+//        and:
+//        Assert.assertEquals(data.location.id, locationId, "Location id should be ${locationId}")
+//        and:
+//        Integer quantityOnHand = data.quantityOnHand ?: 0
+//        Assert.assertEquals(quantityOnHand, 0, "QuantityOnHand should be 0")
```
kchelstowski, line 165: should it be removed/uncommented?

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
kchelstowski: I think there is no need to add an empty line here.
Lately I've been trying to do a research whether the properties should be even separated by an empty line if they don't contain any annotations above, but can't find any good resource about that.
I meant:
```groovy
class ProductResponse {

  String id

  String productCode
 
  String name
}
```
vs
```groovy
class ProductResponse {
  String id
  String productCode
  String name
}
```

kchelstowski (changes requested):

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
alannadolny: I am a fan of adding a new line before each property (the first version), but if u think there is no need, so it can stay

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: I actually hate the extra whitespace, but fine.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Hmm, I've never seen that syntax. Do you have a reference for that somewhere in Java / Groovy docs?

The rest assured docs provide a bunch of examples that use the syntax I used above.
https://github.com/rest-assured/rest-assured/wiki/Usage

And it looks like IntelliJ sometimes breaks when using the period at the beginning of the line (this is for a package name, but still makes me wary)
https://youtrack.jetbrains.com/issue/IDEA-242633/Groovy-code-formatting-incorrectly-wraps-qualified-new-expression-which-leads-to-compiler-error

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: I don't mind breaking it up if there's a logical break (maybe after body and then again after jsonPath) but I also don't think it's necessary unless we go over the 100-120 character limit of the line. Which I'm guessing we are. 

But I don't have a strong feeling. In fact, I don't mind the one line per method (example below). I just don't want these tests to be 1000s of lines unnecessarily.
https://www.baeldung.com/rest-assured-response

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Chetan is supposed to turn those into rest assured asserts.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Can we spend some more time on the substance? This class is going to be removed once we deal with the inconsistences within our JSON responses. We just needed a temporary solution to write these reference implementation tests.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Check style seems to suggest the opposite. 
https://checkstyle.sourceforge.io/checks/imports/importorder.html#:~:text=group%20of%20static%20imports%20is,and%20aren't%20separated%20internally

So let's find a definitive source  or just ignore it.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Agreed, it would be nice.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: Again, where have you seen this syntax? Is that a javascript thing?

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: @chetanmaharishi can you turn these into valid rest-assured assert thingees like the first example I did.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: @chetanmaharishi Same.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
alannadolny: Hm, I can see similar imports to this which I suggested in our code
obraz
If the static imports should be placed before normal imports (as I can see on the page you linked), then I am sorry for the confusion, I mean to keep those imports cleaner.
obraz

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
alannadolny: I've never seen that syntax 😅
I tried to search for breaking chain calls with the dot at the end, and I found a bunch of pages with this.
- https://unibeautify.com/docs/option-break-chained-methods
- https://github.com/prettier/prettier/issues/1565
- https://www.baeldung.com/java-config-spring-security
- I can see it also in the spring security official docs https://spring.io/guides/gs/securing-web/

jmiranda: Ok, so there's some evidence for leading over trailing, so we should have a discussion about this. I guess I've mostly stayed to one line when chaining but I've only ever remembered seeing a trailing dot whenever I've seen chaining. I honestly don't recall ever seeing a leading dot. 

Interesting discussion in the Ruby https://github.com/standardrb/standard/issues/75 with mixed feelings.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: @alannadolny I honestly don't know what's best. When I saw the recommendation for putting static imports up top I was surprised. Just adding this to "we don't yet know the definitive answer" so we'll need to research it. I just don't want to spend too much time researching some of these syntax issues when we could be spending that time actually making tests work.

src/test/groovy/unit/org/pih/warehouse/api/ProductApiTest.groovy
jmiranda: @kchelstowski @alannadolny Ok, after giving this a few days to marinate, I think I'm starting to warm up to the leading dot. I'm not 100% sold yet and I don't necessarily want to apply the line per method convention (like in the following example). But the leading dot does make it more readable.

awalkowiak:
We are temporarily archiving / ice-boxing this until we make a decision on where these tests go

PR #4361, opened 2023-10-30
OBGM-712 Fix import and usage of AuthService and publishEvent in Grai…

kchelstowski:
…ls 3

There was import of `AuthService` missing after rebase and also `.get()` from `AuthService.currentUser` was not removed (todo was left above with an explanation of why it is needed in G3), which would cause the user to be `null`, so I've fixed that as well to avoid some other minor bugs being created.
As for `publishEvent`, in G3 we publish event via `Holders.grailsApplication.mainContext.publishEvent` (in Grails 1 it was enough to write `publishEvent`). Since in the sm service we have `grailsApplication` injected, we don't have to use `Holders`.


alannadolny (approved):

awalkowiak:
LGTM

PR #4362, opened 2023-10-30
OBPIH-5865 Apostrophe in product search gives error

alannadolny:
According to the MySQL documentation: https://dev.mysql.com/doc/refman/8.0/en/string-literals.html#character-escape-sequences

I added a `replaceAll` for all passed single quotes to execute within an SQL query. There is a table with special characters escape sequences:
image

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ProductApiController.groovy
@@ -118,7 +118,7 @@ class ProductApiController extends BaseDomainApiController {
             return
         }
 
-        String[] terms = params?.name?.split(",| ")?.findAll { it }
+        String[] terms = params?.name?.replaceAll("'", "\\\\'")?.split(",| ")?.findAll { it }
         def products, availableItems = []
         if(params.availableItems) {
             Location location = Location.get(session.warehouse.id)


kchelstowski (approved):
I know this ticket relates only to apostrophe problem, hence I give an approve, but I assume the problem might come back to us, if someone uses double quotes or anything weird in the future.
Nice job with investigation.

alannadolny:
@kchelstowski I tried other special characters, so:
- double quotes work
- characters like "tab" cannot be entered, because it automatically changes the focus to the next field/button
- Entering `\` leads to a "bad request" error
- Entering `%` leads to
image

jmiranda:
The reason single quotes break the query is because Hibernate uses single quotes to wrap values in the WHERE clause. 
```
WHERE column like 'value%'
```
Double quotes aren't a problem because it's just another character. It would be a problem if Hibernate used double quotes to wrap the value.
```
WHERE column like "value%"
```
I don't understand the other two (forward slash and percentage) so we need to make sure those get mentioned in the ticket and/or as a new tech debt ticket. 

We need to make sure PR conversations / conclusions make their way back to the ticket.

jmiranda (approved):
Nice work @alannadolny

PR #4363, opened 2023-10-30
OBGM-713 Fix rollback approvals

kchelstowski:
The issue was that in Grails 3 we just have a different package name for this taglib.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -13,6 +13,7 @@ import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
 import grails.orm.PagedResultList
 import grails.validation.ValidationException
+import org.grails.plugins.web.taglib.ApplicationTagLib
 import org.grails.web.json.JSONObject
 import org.hibernate.ObjectNotFoundException
 import org.hibernate.sql.JoinType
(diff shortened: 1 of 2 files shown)


alannadolny (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -2913,7 +2914,7 @@ class StockMovementService {
                 stockMovement.isInApprovalState()
     }
 
-    def getApplicationTagLib() {
-        return grailsApplication.mainContext.getBean('org.codehaus.groovy.grails.plugins.web.taglib.ApplicationTagLib')
+    ApplicationTagLib getApplicationTagLib() {
+        return grailsApplication.mainContext.getBean(ApplicationTagLib)
```
jmiranda, line 2918: Unless you are sure that the getBean method takes a class as an argument, I would highly recommend passing a String argument. As you mentioned in the first comment on this PR, this was an issue with the package.

jmiranda (changes requested):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: Nevermind, it looks like getBean(Class) does exist.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/BeanFactory.html#getBean(java.lang.Class)

kchelstowski: yes, exactly. I've obviously checked if it works and it does. 
We also have this already existing in e.g. Product domain:
https://github.com/openboxes/openboxes/blob/742e5f54a8d2ba5a6cf75f2a99bf0c10d4fa6b0b/grails-app/domain/org/pih/warehouse/product/Product.groovy#L682-L684
Screenshot from 2023-10-30 22-20-31

If you are fine with that one, please approve/merge it today if it's possible.

jmiranda (approved):
LGTM

PR #4357, opened 2023-10-26
Move 0.8.23 into 0.9.0 feature branch

awalkowiak:
I cherry-picked all the commits from the 0.8.23 to the feature 0.9.0 branch. 

I also did some required fixes for the new version:
- Moved new Config.groovy configurations to the runtime.groovy
- Moved some files to proper directory
- Moved RequisitionStatusTransitionEventService to proper package
- Replaced transactional = true with [at]Transactional
- Replaced old ConfigurationHolder and ApplicationHolder with grails.util.Holders

**DO NOT SQUASH THIS PR**

PR #4364, opened 2023-10-31
OBGM-714 Approve request label broken + no flash message when adding APPROVE_REQUEST activity

kchelstowski:
…… (#4277)

This one seems to have been skipped while cherry-picking request approval PRs into Grails 3.
It's #4277 that I've cherry-picked now and has been skipped here: https://github.com/openboxes/openboxes/pull/4357


alannadolny (approved):
Nice catch!

jmiranda (changes requested):
I didn't really like this solution when it was implemented so I think we should try to fix it in this PR. The solution is to use the raw codec https://blog.mrhaki.com/2013/11/grails-goodness-generating-raw-output.html
```
<g:if test="${flash.message}">
    <div class="message">${raw(flash.message)}</div>
</g:if>
```

jmiranda:
Here's a screenshot from the latest version
image

PR #4366, opened 2023-11-03
OBPIH-5793 Reload button issue in bin replenishment

alannadolny:
This issue was hard to spot like any other issue caused by the OSIV. After clicking the "refresh" button on the frontend side, this happened on the backend:
clear picklist -> refresh product availability -> create picklist
When we are creating a picklist we are getting available and suggested items. Suggested items are based on available items and additionally, they have to meet the following conditions:
```
availableItems?.findAll { it.quantityAvailable > 0 && it.pickable }
```
but each item had a quantity available equal to 0.
Available items are calculated using this query:
```
ProductAvailability.executeQuery("""
					select 
				        ii,
					pa.binLocation,
					pa.quantityOnHand,
				        pa.quantityAvailableToPromise
				from ProductAvailability pa
				where pa.location = :location
			""" +
                 "${excludeNegativeQuantity ? "and pa.quantityOnHand > 0" : ""}" +
                 "and pa.product.id in (:products)", [location: location, products: productsIds])
```
and here is the main point - when this query is executed, items that were picked to the picklist are still "picked" and their quantity is equal to 0. When we are executing this query the quantity should be greater than 0 for each item that we previously had in the picklist (even though we can change the quantity). Those quantities are 0 because we didn't commit those changes to the database, and creating and clearing picklist methods are separately called transactions that don't know about each other.

This issue could be solved by using `flush: true` when saving the picklist, but I remember that we tried to avoid this kind of change and I decided to spend more time, working on other solution, and I decided to move the refresh product availability method call to the controller (I left the same behavior that was previously in the `clearPicklist` method, so it should be working in other similar cases.)

### DIFF
--- a/grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
+++ b/grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
@@ -80,7 +80,7 @@ class PicklistService {
                 binLocations.add(it.binLocation?.id)
                 it.delete()
             }
-            picklist.save()
+            picklist.save(flush: true)
         }
 
         productAvailabilityService.refreshProductsAvailability(orderItem?.order?.origin?.id, [orderItem?.product?.id], binLocations?.unique(), false)


kchelstowski (approved):
To extend the topic, since we've spent some time together of figuring that one out. 
The problem is that this call:
https://github.com/openboxes/openboxes/blob/9d783dcc5d668d05f244dd9036f6591241ce21f8/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy#L292-L331

the sql query that uses `new Sql(dataSource)` "doesn't know" about the transaction where the picklist items were deleted here:
https://github.com/openboxes/openboxes/blob/9d783dcc5d668d05f244dd9036f6591241ce21f8/grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy#L67-L85

The `dataSource` here just creates its own connection/pool and doesn't follow any changes that might be made throughout the request (but in different transactions).
So the only way to do it is either like Alan did or call the flush on delete of picklist item.

kchelstowski:
@jmiranda yes, OSIV is **disabled** in Grails 3. Alan meant that in Grails 1 it was working thanks/due to OSIV and in Grails 3 those "osiv issues" are the most difficult to diagnose, as it goes deeply into transactions/sessions lifecycle.

jmiranda (changes requested):
Let's go with the flush approach for now unless a better option presents itself.

drodzewicz (approved):

awalkowiak (approved):

PR #4367, opened 2023-11-06
OBGM-338 Remove shipping report

alannadolny:
As I know it was decided to remove this report because it is not used anymore. While removing the unused code I noticed that there are some methods which printing reports that I hadn't seen before and maybe we could also remove them.
Those reports:
- PaginatedPackingListReport
- PickListReport
- ShippingReport (It is basically the pdf / HTML which we could generate on `/report/showShippingReport`, it is visible on the old shipment view page - "packing list")
image
- PaginatedPackingListReport
Each of them uses the `reportService.generateShippingReport()` method, so I can't delete it for now. If someone is sure that I can also remove those methods I can do that in this PR.
cc @jmiranda @awalkowiak


grails-app/controllers/org/pih/warehouse/reporting/ReportController.groovy
```diff
@@ -303,14 +303,6 @@ class ReportController {
         render(template: "dataTableDialog", model: [url: url])
     }
 
-    def showShippingReport(ChecklistReportCommand command) {
-        command.rootCategory = productService.getRootCategory()
-        if (!command?.hasErrors()) {
-            reportService.generateShippingReport(command)
```
awalkowiak, line 309: Is this still used? Are these other reports that use this service method still used or perhaps we could also remove them?

awalkowiak: @alannadolny now i see the description :see_no_evil: if it is is used then lets leave it as is

awalkowiak (approved):

PR #4368, opened 2023-11-06
OBGM-112 create database migration to set a name for all roles without a name under 0 8 x

drodzewicz:
Added a migration for all of the roles that are missing the name value.
All of the roles with missing `name` can be found in the installation changelog `changelog-insert-data.xml`
https://github.com/openboxes/openboxes/blob/4a9048f5b99975c34e1eaf1e5484c67f112951bb/grails-app/migrations/install/changelog-insert-data.xml#L1202-L1278

The roles are added but name columns is not set to any value (at least first 11 roles).

I have also added a migration to add the `nullable false` constraint on the column to be consistent with the constraint specified in the `Role.groovy`


alannadolny (approved):

grails-app/migrations/0.9.x/changelog-2023-11-06-1330-drop-nullable-constraint-on-name-column-for-role-table.xml
```diff
@@ -0,0 +1,19 @@
...
+        SELECT IS_NULLABLE
+        FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE table_name="role"
+        AND column_name="name"
+        AND TABLE_SCHEMA = DATABASE();
+      </sqlCheck>
+    </preConditions>
+    <addNotNullConstraint tableName="role" columnName="name" defaultNullValue="" columnDataType="VARCHAR(255)" />
+  </changeSet>
+</databaseChangeLog>
```
kchelstowski, line 19: why is the file named "drop nullable constraint" if you add it here?
```xml
 <addNotNullConstraint tableName="role" columnName="name" defaultNullValue="" columnDataType="VARCHAR(255)" />
```

kchelstowski:
except that one question LGTM

grails-app/migrations/0.9.x/changelog-2023-11-06-1330-drop-nullable-constraint-on-name-column-for-role-table.xml
drodzewicz: When creating the changelog file I thought that `nullable: false` was a separate constraint that I would have to drop, so in a way it does makes sense but I do agree that it would be better to rename it to something like 
`changelog-2023-11-06-1330-add-not-nullable-constraint-on-name-column-for-role-table.xml`

thanks for pointing that out :+1:

kchelstowski (approved):

awalkowiak (approved):

grails-app/migrations/0.9.x/changelog-2023-11-06-1330-add-not-nullable-constraint-on-name-column-for-role-table.xml
```diff
@@ -0,0 +1,19 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+
+  <changeSet author="drodzewicz" id="061120231330-0">
+    <preConditions onFail="MARK_RAN" onError="CONTINUE">
+      <sqlCheck expectedResult="YES">
+        SELECT IS_NULLABLE
+        FROM INFORMATION_SCHEMA.COLUMNS
```
jmiranda, line 11: The openboxes user may not have access to system tables like this. I think that might be why you are using onError="CONTINUE", but we should find a different way of doing this if we can.

grails-app/migrations/0.9.x/changelog-2023-11-06-1230-add-missing-role-names.xml
```diff
@@ -0,0 +1,185 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+  <changeSet author="drodzewicz" id="061120231230-0">
+    <preConditions onFail="MARK_RAN">
+      <sqlCheck expectedResult="1">
+        SELECT COUNT(*) FROM role WHERE role_type = "ROLE_ADMIN" AND (name IS NULL OR name = "");
```
jmiranda, line 9: Use single quotes if possible. I encountered an issue with DigitalOcean's database service in which it uses ANSI quotes SQL Mode setting that causes issues when using double quotes. It might be common for this setting to be configured with some of the cloud providers, so using single quotes will help us avoid unnecessary headaches.

https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes

jmiranda (changes requested):

grails-app/migrations/0.9.x/changelog-2023-11-06-1330-add-not-nullable-constraint-on-name-column-for-role-table.xml
jmiranda: The changeSet needs to be marked as only executing on MySQL.

jmiranda: Do we want to use a check on information schema or just allow the addNotNullConstraint to fail. It feels icky to let it fail if it is something we want to ensure since it'll allow data integrity issues on some systems, so I think the info schema path is correct fi that's what Liquibase recommends.

awalkowiak (approved):

PR #4369, opened 2023-11-06
OBGM-359 /location/show/id won't load due to GORM error

alannadolny:
I deleted the `location/show` view (it was empty) and changed redirects from `show` to `edit`, to avoid one unnecessary redirect (the `show` action also redirects to `edit`)


awalkowiak (approved):

kchelstowski:
@alannadolny if so, can't method `def show` be also deleted from the controller?

PR #4370, opened 2023-11-07
OBGM-717 Fix issue causing Invoice to not be deleted

drodzewicz:
use invoiceService.deleteInvoice method which alongside deleteing an Invoice removes InvoiceItems which previosuly caused IntegrityVioliationExzception on invoice delete

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
@@ -68,14 +68,14 @@ class InvoiceController {
         def invoiceInstance = Invoice.get(params.id)
         if (invoiceInstance) {
             try {
-                invoiceDataService.delete(invoiceInstance.id)
+                invoiceService.deleteInvoice(invoiceInstance)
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
```diff
@@ -68,14 +68,14 @@ class InvoiceController {
         def invoiceInstance = Invoice.get(params.id)
         if (invoiceInstance) {
             try {
-                invoiceDataService.delete(invoiceInstance.id)
+                invoiceService.deleteInvoice(invoiceInstance)
```
kchelstowski, line 71: why was that the issue?

kchelstowski (approved):
I thought this might be doable by just changing the data service of Invoice to be an abstract class and to define the delete method there, but the time to catch that was probably when those methods were implemented, so I'm fine with the fix.

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
awalkowiak: I guess as a short-term solution it is an okay solution but would be good if there was some sort of annotation handling that case (like join for gets) or perhaps g3 has better working cascade deletes for this than it was on g1.

awalkowiak (approved):

PR #4371, opened 2023-11-13
OBGM-47 refactor excel importers

drodzewicz:
For the refactoring of the `[Domain]ExcelImporter` classes I followed such a structure

First I created an interface of `DataImporter` which should be implemented by every **ExcelDataImporter** because it reassures us with three core methods that we expect the DataImporter to implement:
- **getData()** - which returns the list of imported data
- **validateData(ImportDataCommand command)** - which handles data validation
- **importData(ImportDataCommand command)** - which handles the data importing

When discussing this refactor I suggested to move the `importData` and `validateData` logic to the ExcelImporter class but after think about it, it was probably not the best approach since there are places in the code where we use the same importData and validateData logic when importing the `CSV` files.
Which is why I decided to create a separate `[Domain]ImportDataService` which implements `ImportDataService` interface so we have to implement the validate and import methods, so we have a more separated and maintainable structure.
In the future when we decide to support both excel and csv imports for certain Domains we can us the same Service for both.

When defining the `[Domain]ImportDataService` in the `[Domain]ExcelImporter` I annotated the ImportDataService with `@Delegate` which saves us from writing few unnecessary lines of code like
```groovy
    importDatatService.importData(command)
}

void validateData(ImportDataCommand command) {
    importDatatService.validateData(command)
}
```  

Additionally, I have followed the example showed in [this article](https://medium.com/@aeronz/grails-3-upload-excel-file-23a85e5289d4) for implementing DataImporters using `excel-import` plugin for grails 3 and added `static cellReporter = new DefaultImportCellCollector()` for each importer which hopefully we will be able to use to validate data more easily.

And lastly, when going through the `[Domain]DataService` classes I tried to move the validate and import methods to the new `[Domain]ImportDataService` and other utility methods like`createSomething()` `findSomething()` I moved to the `[Domain]Service` so currently the only `[Domain]DataService` is the one which is annotated with `@Service`, eliminating the class naming collision which we had before.


grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+        if (!location) {
+            location = new Location()
+        }
+
+        location.name = params.name
+        location.active = Boolean.valueOf(params.active)
+        location.locationNumber = params.locationNumber
+        location.locationGroup = params.locationGroup ? LocationGroup.findByName(params.locationGroup) : null
+        location.parentLocation = params.parentLocation ? Location.findByNameOrLocationNumber(params.parentLocation, params.parentLocation) : null
+        location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
```
kchelstowski, line 746: couldn't we do such binding through command/dto object? So to create some kind of dto that would accept:
```groovy
class LocationCreateDto {
  String locationNumber
  LocationGroup locationGroup
  ...
}
```
and to think about a way to map the dto to entity (Location)?

Maybe this is not good time to do this (I assume you've copied the existing code), but wanted to raise that.
I just think this method has a bunch of unnecessary "boilerplate code"

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+        location.address = createOrUpdateAddress(params, location?.address?.id)
+        if (!location.inventory && !location.parentLocation) {
+            location.inventory = new Inventory(['warehouse': location])
+        }
+        def locationType = null
```
kchelstowski, line 751: let's use types whenever this is possibile:
```suggestion
        LocationType locationType = null
```

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+        if (params.locationType) {
+            String locationTypeName = LocalizationUtil.getDefaultString(params.locationType as String)
+            // TODO: Replace with a single GORM .find with Closure when in Grails 3 (available since Grails 2.0)
+            LocationType matchedLocationType = LocationType
+                    .findAllByNameLike(locationTypeName + "%")
+                    .find{ LocalizationUtil.getDefaultString(it.name) == locationTypeName}
+            locationType = matchedLocationType
+        }
+
+        def currentLocationType = location.locationType
```
kchelstowski, line 761:
```suggestion
        LocationType currentLocationType = location.locationType
```

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+        location.parentLocation = params.parentLocation ? Location.findByNameOrLocationNumber(params.parentLocation, params.parentLocation) : null
+        location.organization = params.organization ? Organization.findByCodeOrName(params.organization, params.organization) : null
+        location.address = createOrUpdateAddress(params, location?.address?.id)
+        if (!location.inventory && !location.parentLocation) {
+            location.inventory = new Inventory(['warehouse': location])
+        }
+        def locationType = null
+        if (params.locationType) {
+            String locationTypeName = LocalizationUtil.getDefaultString(params.locationType as String)
+            // TODO: Replace with a single GORM .find with Closure when in Grails 3 (available since Grails 2.0)
```
kchelstowski, line 754: as far as I remember I was the one to add this TODO - could you try to proceed it? It should be as simple as removing the `.findAllByNameLike` here - as this find would be then the GORM one, not the one related to List interface.

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+            address = new Address()
+        }
+
+        address.address = params.streetAddress ?: ''
+        address.address2 = params.streetAddress2 ?: ''
+        address.city = params.city ?: ''
+        address.stateOrProvince = params.stateOrProvince ?: ''
+        address.postalCode = params.postalCode ?: ''
+        address.country = params.country ?: ''
+        address.description = params.description ?: ''
```
kchelstowski, line 798: This is another part that I think could be easily achieved with the command/dto object.

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+        if (!(location.locationType?.isInternalLocation() || location.locationType?.isZone()) && !location.organization) {
+            def locationCode = identifierService.generateOrganizationIdentifier(params.organization)
+            Organization organization = (location.locationType?.locationTypeCode == LocationTypeCode.SUPPLIER) ?
+                    organizationService.findOrCreateSupplierOrganization(params?.organization, locationCode) :
+                    organizationService.findOrCreateOrganization(params?.organization, locationCode)
+
+            location.organization = organization
+        }
+
+        return location
```
kchelstowski, line 779: I can see some inconsistency - as for address, we persist it in the `createOrUpdateAddress` method (which is fine in my opinion), but we don't persist the location here, but we return it unpersisted. Shouldn't we standardize it?

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+
+        address.address = params.streetAddress ?: ''
+        address.address2 = params.streetAddress2 ?: ''
+        address.city = params.city ?: ''
+        address.stateOrProvince = params.stateOrProvince ?: ''
+        address.postalCode = params.postalCode ?: ''
+        address.country = params.country ?: ''
+        address.description = params.description ?: ''
+
+        if (address.validate() && !address.hasErrors()) {
```
kchelstowski, line 800: I think `address.validate()` should be enough, since if it doesn't pass the validation, it'd return `false`, hence it wouldn't proceed with the `address.save()`

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +724,83 @@ class LocationService {
...
+                    organizationService.findOrCreateSupplierOrganization(params?.organization, locationCode) :
+                    organizationService.findOrCreateOrganization(params?.organization, locationCode)
+
+            location.organization = organization
+        }
+
+        return location
+    }
+
+    Address createOrUpdateAddress(Map params, @Nullable String addressId) {
```
kchelstowski, line 782: weren't you supposed to remove the `@Nullable` annotation here?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -520,4 +520,30 @@ class UserService {
         }
         user.save(failOnError: true)
     }
+
+    User createOrUpdateUser(Map params) {
+        User user = User.findByUsername(params.username)
+        if (!user) {
+            user = new User(params)
+            user.active = true
+            user.password = "password"
```
kchelstowski, line 529: do we really set the password like that?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -520,4 +520,30 @@ class UserService {
...
+        } else {
+            user.properties = params
+        }
+        return user
```
kchelstowski, line 533: same as for location - we return unpersisted user here.

grails-app/services/org/pih/warehouse/data/CategoryService.groovy
```diff
@@ -0,0 +1,28 @@
...
+    Category createOrUpdateCategory(Map params) {
+        Category category = Category.findByIdOrName(params.id, params.name)
+        if (!category) {
+            category = new Category()
+        }
+        category.name = params.name
+        if (params.parentCategoryId && category?.parentCategory?.id != params.parentCategoryId) {
+            category.parentCategory = Category.get(params.parentCategoryId)
+        }
+        return category
```
kchelstowski, line 26: same as for location and user (returning an unpersisted category)

grails-app/services/org/pih/warehouse/data/ProductPackageService.groovy
```diff
@@ -0,0 +1,52 @@
...
+        productPackage.properties = params
+        if (!productPackage.name) {
+            productPackage.name = "${unitOfMeasure.code}/${params.quantity}"
+        }
+        if (!productPackage.description) {
+            productPackage.description = "${unitOfMeasure.name} of ${params.quantity}"
+        }
+        productPackage.product = product
+        productPackage.productSupplier = productSupplier
+        productPackage.uom = unitOfMeasure
```
kchelstowski, line 37: same as above - we can consider avoiding this boilerplate binding by using dto/command approach.

grails-app/services/org/pih/warehouse/data/ProductPackageService.groovy
```diff
@@ -0,0 +1,52 @@
...
+        return productPackage
```
kchelstowski, line 38: same as above - returning an unpersisted entity

grails-app/services/org/pih/warehouse/data/ProductPackageService.groovy
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    def findOrCreate(Map params) {
```
kchelstowski, line 41:
```suggestion
    ProductPackage findOrCreate(Map params) {
```

grails-app/services/org/pih/warehouse/data/ProductSupplierAttributeService.groovy
```diff
@@ -0,0 +1,49 @@
...
+            productAttribute.unitOfMeasure = uom
+        } else {
+            productAttribute = new ProductAttribute()
+            productAttribute.attribute = attribute
+            productAttribute.productSupplier = productSupplier
+            productAttribute.value = attributeValue
+            productAttribute.unitOfMeasure = uom
+            product.addToAttributes(productAttribute)
+        }
+        return productAttribute
```
kchelstowski, line 47: returning an unpersisted entity + boilerplate binding

grails-app/services/org/pih/warehouse/data/ProductSupplierAttributeService.groovy
```diff
@@ -0,0 +1,49 @@
...
+import org.pih.warehouse.core.UnitOfMeasure
+import org.pih.warehouse.importer.ImportDataCommand
+import org.pih.warehouse.product.Attribute
+import org.pih.warehouse.product.Product
+import org.pih.warehouse.product.ProductAttribute
+import org.pih.warehouse.product.ProductSupplier
+
+@Transactional
+class ProductSupplierAttributeService {
+    def createOrUpdate(Map params) {
```
kchelstowski, line 23:
```suggestion
    ProductAttribute createOrUpdate(Map params) {
```

grails-app/services/org/pih/warehouse/data/TagService.groovy
```diff
@@ -0,0 +1,26 @@
...
+@Transactional
+class TagService {
+    Tag createOrUpdateTag(Map params) {
+        Tag tag = Tag.findByIdOrTag(params.id, params.tag)
+        if (!tag) {
+            tag = new Tag()
+        }
+        tag.id = params.id
+        tag.tag = params.tag
+        return tag
```
kchelstowski, line 24: returning an unpersisted entity

grails-app/services/org/pih/warehouse/importer/ImportDataService.groovy
```diff
@@ -0,0 +1,16 @@
...
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.importer
+
+interface ImportDataService {
+    void validateData(ImportDataCommand command)
+
+    void importData(ImportDataCommand command)
+}
```
kchelstowski, line 16: I appreciate the interface

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -0,0 +1,199 @@
...
+                    command.warnings[index] << "Product name [${row.product}] does not appear to be the same as in the database [${product.name}] (Levenshtein distance: ${levenshteinDistance})"
+                }
+
+            }
+
+            if (row.quantity && (row.quantity as int) < 0) {
+                command.errors.reject("error.quantity.negative", "Row ${rowIndex}: Product '${row.productCode}' must have positive quantity")
+            }
+        }
+    }
```
kchelstowski, line 126: the complicated logic here won with me :see_no_evil: 
I'll skip that for later
what I can say for now is that at least we should add types where it's possible, instead of spamming def.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -0,0 +1,34 @@
...
+package org.pih.warehouse.importer
+
+import grails.gorm.transactions.Transactional
+import org.pih.warehouse.data.DataService
+
+@Transactional
+class InventoryLevelImportDataService implements ImportDataService {
+    DataService dataService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 19: let's add the `@Override` annotations if you implement methods from an interface. I know it's not necessary and the compiler understands it, but it's easier to read the code having those.

grails-app/services/org/pih/warehouse/importer/LocationImportDataService.groovy
```diff
@@ -7,32 +7,22 @@
...
-import org.pih.warehouse.LocalizationUtil
-
-import javax.annotation.Nullable
 
 @Transactional
-class LocationDataService {
-
-    def organizationService
-    def identifierService
+class LocationImportDataService implements ImportDataService {
```
kchelstowski, line 22: I know there was a reason it had not been named as `ImportDataService` - since it's also been used for exports and the Import name did not fit well.

grails-app/services/org/pih/warehouse/importer/OutboundStockMovementImportDataService.groovy
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
+        log.info "Import data " + command.filename
+        command.data.eachWithIndex {params, index ->
+            RequisitionItem requisitionItem = requisitionService.buildRequisitionItem(params)
+            if(requisitionItem.validate()){
```
kchelstowski, line 46:
```suggestion
            if (requisitionItem.validate()) {
```

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
```diff
@@ -0,0 +1,40 @@
...
+                    command.errors.reject("Row ${index + 1} name = ${person.name}: ${error.getFieldError()}")
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
+        command.data.eachWithIndex { params, index ->
+            Person person = personService.createOrUpdatePerson(params)
+            if (person.validate()) {
```
kchelstowski, line 35: this is what I meant when saying that `create` method should also proceed with the save - I think it should be `create` method's responsibility to persist the data.
Additional note - to follow "good practices" we should call the `.save()` only if we create a brand new object (or some associations inside it is new), because calling `.save()` for a persistent entity is unnecessary, since Hibernate will persist it at the transaction commit.

grails-app/services/org/pih/warehouse/importer/ProductAssociationImportDataService.groovy
```diff
@@ -48,18 +45,18 @@ class ProductAssociationDataService {
                 if (params.hasMutualAssociation || it.hasMutualAssociation) {
                     return  (
                             params['product'] == it['product'] &&
-                            params['associatedProduct'] == it['associatedProduct'] &&
-                            params['code'] == it['code']
-                        ) || (
+                                    params['associatedProduct'] == it['associatedProduct'] &&
```
kchelstowski, line 48: is it only github going crazy, or some additional indents came in? 
+ a nitpicky comment that we should add into our coding conventions and just reminded myself about it  - the `&&`/`||` should be placed at the beginning of a line, not at the end.

grails-app/services/org/pih/warehouse/importer/ProductAssociationImportDataService.groovy
```diff
@@ -77,22 +74,22 @@ class ProductAssociationDataService {
...
             if (!productAssociationInstance.product) {
                 command.errors.reject(
-                    "Row ${index + 1}: Product with code '${params['product.productCode']}' does not exist"
+                        "Row ${index + 1}: Product with code '${params['product.productCode']}' does not exist"
                 )
             }
             if (!productAssociationInstance.associatedProduct) {
                 command.errors.reject(
-                    "Row ${index + 1}: Product with code '${params['associatedProduct.productCode']}' does not exist"
+                        "Row ${index + 1}: Product with code '${params['associatedProduct.productCode']}' does not exist"
```
kchelstowski, line 82: I can also see some additional indents here

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
```diff
@@ -0,0 +1,40 @@
...
+import grails.gorm.transactions.Transactional
+import org.pih.warehouse.core.Person
+import org.pih.warehouse.data.PersonService
+import org.springframework.validation.BeanPropertyBindingResult
+
+@Transactional
+class PersonImportDataService implements ImportDataService {
+    PersonService personService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 21:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
```diff
@@ -0,0 +1,40 @@
...
+            Person person = personService.createOrUpdatePerson(params)
+            if (!person.validate()) {
+                person.errors.each { BeanPropertyBindingResult error ->
+                    command.errors.reject("Row ${index + 1} name = ${person.name}: ${error.getFieldError()}")
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 32:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductAttributeImportDataService.groovy
```diff
@@ -0,0 +1,57 @@
...
+import org.pih.warehouse.product.Attribute
+import org.pih.warehouse.product.Product
+import org.pih.warehouse.product.ProductAttribute
+import org.springframework.validation.BeanPropertyBindingResult
+
+
+@Transactional
+class ProductAttributeImportDataService implements ImportDataService {
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 23:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductAttributeImportDataService.groovy
```diff
@@ -0,0 +1,57 @@
...
+            ProductAttribute productAttribute = createOrUpdateProductAttribute(params)
+            if (!productAttribute.validate()) {
+                productAttribute.errors.each { BeanPropertyBindingResult error ->
+                    command.errors.reject("Row ${index + 1}: Product attribute ${productAttribute} is invalid: ${error.getFieldError()}")
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 34:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductAttributeImportDataService.groovy
```diff
@@ -0,0 +1,57 @@
...
+        UnitOfMeasure unitOfMeasure = UnitOfMeasure.findByCode(params.unitOfMeasureCode)
+        ProductAttribute productAttribute = ProductAttribute.findByProductAndAttribute(product, attribute)
+        if (!productAttribute) {
+            productAttribute = new ProductAttribute()
+            productAttribute.attribute = attribute
+            product.addToAttributes(productAttribute)
+        }
+        productAttribute.value = params.attributeValue
+        productAttribute.unitOfMeasure = unitOfMeasure?:attribute?.unitOfMeasureClass?.baseUom
+        return productAttribute
```
kchelstowski, line 55: same comment as for previous cases - eventual persistent of a new entity should happen in my opinion here, not somewhere else

grails-app/services/org/pih/warehouse/importer/ProductCatalogImportDataService.groovy
```diff
@@ -0,0 +1,43 @@
...
+import org.pih.warehouse.product.Category
+import org.pih.warehouse.product.ProductCatalog
+import org.pih.warehouse.product.ProductCatalogService
+import org.springframework.validation.BeanPropertyBindingResult
+
+@Transactional
+class ProductCatalogImportDataService implements ImportDataService {
+    ProductCatalogService productCatalogService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 23:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductCatalogImportDataService.groovy
```diff
@@ -0,0 +1,43 @@
...
+            if (!productCatalog.validate()) {
+                productCatalog.errors.each { BeanPropertyBindingResult error ->
+                    command.errors.reject("Row ${index + 1}: Product catalog ${productCatalog.name} is invalid: ${error.getFieldError()}")
+                }
+            }
+            productCatalog.discard()
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 35:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductCatalogItemImportDataService.groovy
```diff
@@ -0,0 +1,43 @@
...
+import org.pih.warehouse.product.ProductCatalog
+import org.pih.warehouse.product.ProductCatalogItem
+import org.pih.warehouse.product.ProductCatalogService
+import org.springframework.validation.BeanPropertyBindingResult
+
+@Transactional
+class ProductCatalogItemImportDataService implements ImportDataService {
+    ProductCatalogService productCatalogService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 24:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductCatalogItemImportDataService.groovy
```diff
@@ -0,0 +1,43 @@
...
+            ProductCatalogItem productCatalogItem = productCatalogService.createOrUpdateProductCatalogItem(params)
+            if (!productCatalogItem.validate()) {
+                productCatalogItem.errors.each { BeanPropertyBindingResult error ->
+                    command.errors.reject("Row ${index + 1}: Product catalog item is invalid: ${error.getFieldError()}")
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 35:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductImportDataService.groovy
```diff
@@ -0,0 +1,41 @@
...
+ **/
+package org.pih.warehouse.importer
+
+import grails.gorm.transactions.Transactional
+import org.pih.warehouse.product.Product
+
+@Transactional
+class ProductImportDataService implements ImportDataService {
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 18:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductImportDataService.groovy
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    void importData(ImportDataCommand command) {
+        log.info "import data"
+
+        try {
+            // Iterate over each row
+            command?.data?.each { Map params ->
+                log.info "import data " + params
```
kchelstowski, line 34: why nothing else except logging happens here?

grails-app/services/org/pih/warehouse/importer/ProductPackageImportDataService.groovy
```diff
@@ -7,24 +7,21 @@
...
 
 @Transactional
-class ProductPackageDataService {
+class ProductPackageImportDataService implements ImportDataService {
+    ProductPackageService productPackageService
 
-    def uomService
-
-    Boolean validate(ImportDataCommand command) {
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 24:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductPackageImportDataService.groovy
```diff
@@ -63,7 +60,7 @@ class ProductPackageDataService {
             }
 
 
-            def productPackage = createOrUpdate(params)
+            def productPackage = productPackageService.createOrUpdate(params)
```
kchelstowski, line 63:
```suggestion
            ProductPackage productPackage = productPackageService.createOrUpdate(params)
```

grails-app/services/org/pih/warehouse/importer/ProductPackageImportDataService.groovy
```diff
@@ -75,46 +72,15 @@ class ProductPackageDataService {
         }
     }
 
-    void process(ImportDataCommand command) {
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 75:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSupplierAttributeImportDataService.groovy
```diff
@@ -7,21 +7,22 @@
...
 import org.pih.warehouse.product.ProductAttribute
 import org.pih.warehouse.product.ProductSupplier
 
 @Transactional
-class ProductSupplierAttributeDataService {
+class ProductSupplierAttributeImportDataService implements ImportDataService {
+    ProductSupplierAttributeService productSupplierAttributeService
 
-    Boolean validate(ImportDataCommand command) {
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 25:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSupplierAttributeImportDataService.groovy
```diff
@@ -92,46 +93,19 @@ class ProductSupplierAttributeDataService {
             UnitOfMeasure uom = unitOfMeasure ? UnitOfMeasure.findByCode(unitOfMeasure) : null
             if (attribute.unitOfMeasureClass && attribute.unitOfMeasureClass?.type != uom?.uomClass?.type) {
                 command.errors.reject("Row ${index + 1}: Unit of Measure ${unitOfMeasure} " +
-                    "data type does not match the unit of measure data type for given attribute")
+                        "data type does not match the unit of measure data type for given attribute")
```
kchelstowski, line 96: indent? :eyes:

grails-app/services/org/pih/warehouse/importer/ProductSupplierAttributeImportDataService.groovy
```diff
@@ -92,46 +93,19 @@ class ProductSupplierAttributeDataService {
...
             }
         }
     }
 
-    void process(ImportDataCommand command) {
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 101:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSupplierImportDataService.groovy
```diff
@@ -0,0 +1,142 @@
...
+import org.pih.warehouse.product.ProductSupplier
+import util.ConfigHelper
+
+import java.text.SimpleDateFormat
+
+@Transactional
+class ProductSupplierImportDataService implements ImportDataService {
+    ProductSupplierService productSupplierService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 28:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSupplierImportDataService.groovy
```diff
@@ -0,0 +1,142 @@
...
+            def active = params.active
+            def productCode = params.productCode
+            def supplierName = params.supplierName
+            def manufacturerName = params.manufacturerName
+            def preferenceType = params.globalPreferenceTypeName
+            def uomCode = params.defaultProductPackageUomCode
+            def packageQuantity = params.defaultProductPackageQuantity
+            def validityStartDate = params.globalPreferenceTypeValidityStartDate
+            def validityEndDate = params.globalPreferenceTypeValidityEndDate
+            def contractPriceValidUntil = params.contractPriceValidUntil
```
kchelstowski, line 41: types + boilerplate binding

grails-app/services/org/pih/warehouse/importer/ProductSupplierImportDataService.groovy
```diff
@@ -0,0 +1,142 @@
...
+                        command.errors.reject("Row ${index + 1}: Contract Price Valid Until date ${contractPriceValidUntil} is invalid. Please enter a date after ${minDate.getYear()+1900}.")
+                    }
+                } catch (Exception e) {
+                    command.errors.reject("Row ${index + 1}: Contract Price Valid Until date ${contractPriceValidUntil} is invalid. " + e.message)
+                }
+            }
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 134:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSupplierPreferenceImportDataService.groovy
```diff
@@ -0,0 +1,63 @@
...
+import org.pih.warehouse.product.ProductSupplier
+import org.pih.warehouse.product.ProductSupplierPreference
+
+@Transactional
+class ProductSupplierPreferenceImportDataService implements ImportDataService {
+    ProductSupplierPreferenceService productSupplierPreferenceService
+
+    String DEFAULT = "DEFAULT"
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 25:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/ProductSynonymImportDataService.groovy
```diff
@@ -7,28 +7,29 @@
...
-class ProductSynonymDataService {
-
-    def productService
+@Transactional
+class ProductSynonymImportDataService implements ImportDataService {
+    ProductService productService
     GrailsApplication grailsApplication
 
-    Boolean validateData(ImportDataCommand command) {
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 24:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/TagImportDataService.groovy
```diff
@@ -0,0 +1,41 @@
...
+import grails.gorm.transactions.Transactional
+import org.pih.warehouse.core.Tag
+import org.pih.warehouse.data.TagService
+import org.springframework.validation.BeanPropertyBindingResult
+
+@Transactional
+class TagImportDataService implements ImportDataService {
+    TagService tagService
+
+    void validateData(ImportDataCommand command) {
```
kchelstowski, line 21:
```suggestion
    @Override
    void validateData(ImportDataCommand command) {
```

grails-app/services/org/pih/warehouse/importer/TagImportDataService.groovy
```diff
@@ -0,0 +1,41 @@
...
+            if (!tag.validate()) {
+                tag.errors.each { BeanPropertyBindingResult error ->
+                    command.errors.reject("Row ${index + 1}: Tag ${tag.tag} is invalid: ${error.getFieldError()}")
+                }
+            }
+            tag.discard()
+        }
+    }
+
+    void importData(ImportDataCommand command) {
```
kchelstowski, line 33:
```suggestion
    @Override
    void importData(ImportDataCommand command) {
```
and so on... I won't add them to all of the files ;)

grails-app/services/org/pih/warehouse/importer/TagImportDataService.groovy
```diff
@@ -0,0 +1,41 @@
...
+        command.data.eachWithIndex { params, index ->
+            Tag tag = tagService.createOrUpdateTag(params)
+            if (tag.validate()) {
+                tag.save(failOnError: true)
```
kchelstowski, line 37: as mentioned above - I think this should be `createOrUpdate` responsibility, but to be discussed probably.
We should at least standardize it everywhere

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -907,4 +908,73 @@ class RequisitionService {
         requisition.addToComments(comment)
         requisition.save()
     }
+
+    RequisitionItem buildRequisitionItem(Map params) {
+        String productCode = params.productCode
+        Product product = Product.findByProductCode(productCode)
+        if(!product) {
```
kchelstowski, line 915:
```suggestion
        if (!product) {
```

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -907,4 +908,73 @@ class RequisitionService {
...
+        }
+
+        def deliveryDate = params.requestedDeliveryDate
+        if (!isDateOneWeekFromNow(deliveryDate)) {
+            throw new IllegalArgumentException("Delivery date must be after seven days from now")
+        }
+
+        def comments = params.destination
+        def requestNumber = params.requestNumber
+        def requisition = Requisition.findByRequestNumber(requestNumber)
```
kchelstowski, line 931: types

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -907,4 +908,73 @@ class RequisitionService {
...
+            requisition.origin = origin
+
+            Location destination = Location.findByLocationNumber(params.destination)
+            if (!destination) {
+                throw new IllegalArgumentException("Location not found for destination ${params.destination}")
+            }
+            requisition.destination = destination
+            requisition.requestedDeliveryDate = deliveryDate.toDate()
+            requisition.requestedBy = authService.currentUser
+            requisition.save(failOnError: true)
```
kchelstowski, line 952: also a bunch of boilerplate binding is being done here

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -907,4 +908,73 @@ class RequisitionService {
...
+        }
+
+        def requisitionItem = RequisitionItem.createCriteria().get {
```
kchelstowski, line 955:
```suggestion
        RequisitionItem requisitionItem = RequisitionItem.createCriteria().get {
```

src/main/groovy/org/pih/warehouse/importer/CategoryExcelImporter.groovy
```diff
@@ -11,15 +11,17 @@ package org.pih.warehouse.importer
...
 
-class CategoryExcelImporter extends AbstractExcelImporter {
+class CategoryExcelImporter extends AbstractExcelImporter implements DataImporter {
 
-    ExcelImportService excelImportService
-    CategoryDataService categoryDataService
+    static cellReporter = new DefaultImportCellCollector()
 
+    ExcelImportService excelImportService
+    @Delegate
```
kchelstowski, line 23: indent before @Delegate and ExcelImportService

src/main/groovy/org/pih/warehouse/importer/DataImporter.groovy
```diff
@@ -0,0 +1,19 @@
...
+package org.pih.warehouse.importer
+
+interface DataImporter {
+
+    List<Map> getData()
+
+    void validateData(ImportDataCommand command)
+
+    void importData(ImportDataCommand command)
+}
```
kchelstowski, line 19: since the `ImportDataService` import is almost the same as this one, maybe let's just extend this one and add the `getData()` method here?
```groovy
interface DataImporter extends ImportDataService {
  List<Map> getData()
}
```

kchelstowski (changes requested):
General comment:

- @drodzewicz I know 90% of this code was only copied by you, but I let myself comment on it anyway, because I don't know if we are eventually supposed to "remove" some tech debt at this point.
- in my opinion most of this boilerplate binding could be avoided if we used dto/command approach for binding, so I raised that topic (however I'm not sure if the current implementation of `importData` in BatchController would allow that...)
- as long as I am concerned there had been a reason why those were not named `*ImportService` before
- we should standardize the create methods for those services - for some of them, we proceed with the save in the create methods, and for some, we only return an eventual unpersisted entity - reminder: we should avoid calling `save()` for persistent entities (those, that already exist and are not brand new, because Hibernate would save them at the transaction commit), so potentially we could also distinguish create and update methods and come up with some common part of them.
- I appreciate the interfaces so much, but we should add `@Override` annotations for the methods we implement for a better readability: https://rules.sonarsource.com/java/RSPEC-1161/

drodzewicz:
Thank for your feedback @kchelstowski 
I tried to address most of the issues that you have pointed out. Some that I don't feel to confident with I left out and feel like they should be worked on separately.
That is, refactoring/optimizing some validation logic (some places have missing methods which do nothing) but at this moment I would treat this PR as a little cleanup and setting a specific structure to how we should handle imports.

I also noticed that contrary to Excels, we don't have the same system for CSV imports, instead they are scattered around the Services, so we should probably look into it in the future.

There are many improvements that come to my mind around data imports but I'll be sure to mention them at our next tech huddle if we have time.

As for the `DataImportService` classes I get where you are coming from and I am in favor for renaming it if we agree on some kind of a name.
Me being a lazy developer I asked ChatGPT for suggestions and got some results:

- **DataProcessor**: This name can still work if the class is responsible for processing data both during import and export.
- **DataIOHandler**: This suggests that the class is a general handler for input and output operations.
- **DataExchangeManager**: If the class manages the exchange of data, encompassing both import and export operations.
- **ImportExportHandler**: A straightforward name that explicitly states the class's dual role in handling both imports and exports.
- **DataInOutProcessor**: Combining "in" for import and "out" for export, this name indicates a versatile class for data handling.
- **DataTransferManager**: If the class is responsible for transferring data in and out of the system.

jmiranda (changes requested):
I love the overall feel of the refactoring. Well done. 

I am marking this as "Request Changes" here because I don't want solutions discussed in the PR, they belong in the ticket. 

I'll come back to review when that is done.

cc @awalkowiak @kchelstowski @alannadolny 

@awalkowiak Remind me to discuss this on the tech huddle. I've repeate

grails-app/services/org/pih/warehouse/core/LocationService.groovy
jmiranda: Going back to @kchelstowski's point about standardizing, perhaps we just need a method name that reflect what it's doing.
```
Address address = buildAddress(params)
Address address = populateAddress(params)
```
Or actually, it's possible doing more of a findOrCreate
```
Address address = findOrCreateAddress(params)
```

grails-app/services/org/pih/warehouse/core/UserService.groovy
jmiranda: For imported users, yes. I think the reason was due to the fact that we added this feature as part of an implementation so we didn't need to create users manually. The user that was doing the importing had a process to follow with respect to communicating with users and having them reset the password. 

This should have been accompanied with a feature that required the user to reset their password if it was set to this default value. However, that part never happened. 

I just commented about this tech debt in a ticket https://pihemr.atlassian.net/browse/OBPIH-5531?focusedCommentId=147513

grails-app/services/org/pih/warehouse/data/ProductPackageService.groovy
jmiranda: Again, the issue is with the data binding mechanism. In controllers we're always passing in IDs so the data binding layer is able to convert the ID to a domain object. When importing data via CSV/Excel, we're passing more natural identifiers like product codes and location numbers. These values would not be handled by the data binding mechanism, making the command object less useful in that you'd still need to do these lookups somewhere.

grails-app/services/org/pih/warehouse/importer/ImportDataService.groovy
jmiranda: :100:

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
jmiranda: This is the first place where I think a command object with validation rules makes sense.

jmiranda: Also this crap might be a good candidate for cell reporters. 
```
command.warnings[index]
```

grails-app/services/org/pih/warehouse/importer/LocationImportDataService.groovy
jmiranda: Yeah let's talk about that on the tech huddle. ImportDataService is a little awkward.

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
jmiranda: No no no. 

createOrUpdate* should not persist. Ever. They are used for populating an object that can be validated and persisted by caller. This is because sometimes we call createOrUpdate in the validateData() method becuase we need to construct the data object before it is validated and we don't want the validateData method to persist the object. The validateData and importData are completely separate steps in the import process. 

Also, we don't want createOrUpdate* to persist for performance reasons. Read this https://www.naleid.com/2009/10/01/batch-import-performance-with-grails-and-mysql.html

So if createOrUpdate is confusing, then we need to rename to something that makes it clear that we are constructing the object, not persisting the object.

jmiranda: The Person handler is a simple encapsulation of why we shouldn't persist in the createOrUpdate method 
```
    Boolean validateData(ImportDataCommand command) {
            Person person = createOrUpdatePerson(params)
            if (!person.validate()) {
                person.errors.each { BeanPropertyBindingResult error ->
                    command.errors.reject("Row ${index + 1} name = ${person.name}: ${error.getFieldError()}")
                }
            }
        }
    }

            Person person = createOrUpdatePerson(params)
            if (person.validate()) {
                person.save(failOnError: true)
            }
        }
    }
```
and why we don't need a separate command class.
```
    Person createOrUpdatePerson(Map params) {
        Person person = Person.findByFirstNameAndLastName(params.firstName, params.lastName)
        if (!person) {
            person = new Person(params)
        } else {
            person.properties = params
        }
        return person
    }
```

grails-app/services/org/pih/warehouse/importer/ProductImportDataService.groovy
jmiranda: This was code moved from the ProductService. I don't think we ever exposed a "product" import type on the Data Import form (we use import products CSV instead), so this probably won't ever be invoked. I would like to fix that eventually so feel free to add a FIXME or TODO in the code.

src/main/groovy/org/pih/warehouse/importer/DataImporter.groovy
jmiranda: This seems awkward but it's too late for me to think clearly about a solution so let's discuss during tomorrow's tech huddle.

jmiranda (changes requested):
It's been awhile since i started this review, but I assume there were some changes requested along the way.

grails-app/services/org/pih/warehouse/core/LocationService.groovy
kchelstowski: @jmiranda to be clear, I meant even something like:
```groovy
Location createOrUpdateLocation(LocationCreateDto locationDto) {
  ...
}
```
but this would have to happen at the controller level which as I wrote in on of the comments when I realized it - it might be difficult to achieve with the current "generic" implementation of the DataService methods. 

I agree that the `findByName` cases done for `locationGroup` would be difficult to achieve (unless if we maybe overwrote the setter for the particular dto), but my point was to just maybe try to bind the "common" part, and the rest, that requires some logic, like `locationGroup`, `parentLocation` to be still done manually.
It was just a "loose thought" that I wish we followed this dto/command approach in the future in the places where it is possible.

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
kchelstowski: @jmiranda as for performance side, I quite don't understand what current code does in order to improve the performance? 
At the moment we still call the save inside the loop:
```groovy

            Person person = createOrUpdatePerson(params)
            if (person.validate()) {
                person.save(failOnError: true)
            }
        }
    }
```
so wouldn't calling the save inside the `createOrUpdatePerson` actuallly be just moving the `if (person.validate()) {...}` part there? which in fact doesn't change anything?

I know that in Spring there is `saveAll()` method which might improve performance for some batch operations. As I can see, in Grails we also have it, but from the domain level, e.g.`User.saveAll()` - maybe it could be considered somewhere for an experiment :man_shrugging:

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -718,4 +723,83 @@ class LocationService {
         }
         existingLocation.delete(flush: true)
     }
+
+    Location createOrUpdateLocation(Map params) {
```
jmiranda, line 727: @drodzewicz commented that this are just doing data binding so bindLocation is probably more appropriate. And we should move them into the *DataImportService because 1) that's where they are primarily used 2) we can repurpose DataImportService to be generically useful for any type of import / export activities.

jmiranda: One concern is that we also do some binding like this on APIs so we might want to make sure that the DataImportService is named appropriately to support other use cases like CRUD operations on the API.

Thoughts:
DataMarshaller, DataMapper - take a look at jackson to see how it handles the API for patterns related to moving data back and forth between string and object.
The other part of this that is complicated related to the conversion from string to object of an primary key, code/identifier, name.

grails-app/services/org/pih/warehouse/importer/PersonImportDataService.groovy
jmiranda: Tech Huddle: Add tech debt to look into performance difference between calling save on individual objects or saveAll on a collection of objects.

awalkowiak:
@drodzewicz there is a minor conflict, please fix it. Did you resolve every change request here?

awalkowiak (approved):
Ok, I **think** it LGTM. Anyone wants to proof-review it too? Or can I merge it?

grails-app/services/org/pih/warehouse/importer/LocationImportDataService.groovy
jmiranda: Nothing to do now, but after more thought I'm thinking we're probably going to call these classes *DataImportService or *DataMigrationService. If we combine data import and data export into a single implementation class then it'll probably be the latter with the potential for using interfaces for the two types of data migration activities.
```
class LocationDataMigrationService implements DataImportService, DataExportService {
    void validateData(ImportDataCommand command) { ... } 
    void importData(ImportDataCommand command) { ... }
    List exportData(ExportDataCommand command) { ... }
}
```
Not entirely sure yet if the command classes will follow suit or if I am ok with command classes named by an action (import data) rather than a noun (data import).

jmiranda (approved):
I reviewed as much as I could, but there are too many changes. We're going to have to let QA determine whether there are any issues.

src/main/groovy/org/pih/warehouse/importer/ProductAssociationExcelImporter.groovy
```diff
@@ -12,10 +12,18 @@ package org.pih.warehouse.importer
...
 
 @Transactional
-class ProductAssociationExcelImporter extends AbstractExcelImporter {
+class ProductAssociationExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
+
+    ExcelImportService excelImportService
+    @Delegate
+    ProductAssociationImportDataService productAssociationImportDataService
```
kchelstowski, line 26: I believe there is an indent missing between `ExcelImportService` and the `@Delegate` annotation.
Would be good to fix at least that.

src/main/groovy/org/pih/warehouse/importer/ProductCatalogExcelImporter.groovy
```diff
@@ -12,15 +12,18 @@ package org.pih.warehouse.importer
...
 
 @Transactional
-class ProductCatalogExcelImporter extends AbstractExcelImporter {
+class ProductCatalogExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
 
     ExcelImportService excelImportService
+    @Delegate
+    ProductCatalogImportDataService productCatalogImportDataService
```
kchelstowski, line 26: same as above

src/main/groovy/org/pih/warehouse/importer/ProductCatalogItemExcelImporter.groovy
```diff
@@ -20,9 +21,13 @@ import org.pih.warehouse.product.ProductCatalogItem
...
 
 @Transactional
-class ProductCatalogItemExcelImporter extends AbstractExcelImporter {
+class ProductCatalogItemExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
 
     ExcelImportService excelImportService
+    @Delegate
+    ProductCatalogItemImportDataService productCatalogItemImportDataService
```
kchelstowski, line 30: same as above

src/main/groovy/org/pih/warehouse/importer/ProductExcelImporter.groovy
```diff
@@ -11,12 +11,17 @@ package org.pih.warehouse.importer
...
-class ProductExcelImporter extends AbstractExcelImporter {
+class ProductExcelImporter extends AbstractExcelImporter implements DataImporter {
 
-    def productService
-    def excelImportService
+    static cellReporter = new DefaultImportCellCollector()
+
+    ExcelImportService excelImportService
+    @Delegate
+    ProductImportDataService productImportDataService
```
kchelstowski, line 24: same as above

src/main/groovy/org/pih/warehouse/importer/ProductPackageExcelImporter.groovy
```diff
@@ -11,12 +11,17 @@ package org.pih.warehouse.importer
...
 import org.grails.plugins.excelimport.ExpectedPropertyType
 
-class ProductPackageExcelImporter extends AbstractExcelImporter {
+class ProductPackageExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
 
     ExcelImportService excelImportService
+    @Delegate
+    ProductPackageImportDataService productPackageImportDataService
```
kchelstowski, line 24: same as above

src/main/groovy/org/pih/warehouse/importer/ProductSupplierAttributeExcelImporter.groovy
```diff
@@ -11,9 +11,17 @@ package org.pih.warehouse.importer
...
 import org.grails.plugins.excelimport.ExpectedPropertyType
 
-class ProductSupplierAttributeImporter extends AbstractExcelImporter {
+class ProductSupplierAttributeExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
+
+    ExcelImportService excelImportService
+    @Delegate
+    ProductSupplierAttributeImportDataService productSupplierAttributeImportDataService
```
kchelstowski, line 24: same as above

src/main/groovy/org/pih/warehouse/importer/ProductSupplierPreferenceExcelImporter.groovy
```diff
@@ -11,9 +11,24 @@ package org.pih.warehouse.importer
...
 import org.grails.plugins.excelimport.ExpectedPropertyType
 
-class ProductSupplierPreferenceImporter extends AbstractExcelImporter {
+class ProductSupplierPreferenceExcelImporter extends AbstractExcelImporter implements DataImporter {
+
+    static cellReporter = new DefaultImportCellCollector()
+
+    ExcelImportService excelImportService
+    @Delegate
+    ProductSupplierPreferenceImportDataService productSupplierPreferenceImportDataService
```
kchelstowski, line 24: same as above (just check all the files, so I won't repeat the same comment throughout all the files)

kchelstowski (commented):
I can see that most of my suggestions have been included, some have not been (e.g. adding types in some places), but I don't wanna be nitpicky anymore, since it has not been Dariusz's code and the scope of his ticket.
Anyway I hope to cover them some time in the future.

alannadolny (commented):
the amount of this code won with me...

src/main/groovy/org/pih/warehouse/importer/ProductAssociationExcelImporter.groovy
drodzewicz: At first I did not understand what do you mean by indent and thought that maybe github has formatted the code for you in some weird way because I don't see any wrong indentation (tabs from the left).
But now I understand that you meant spcaing between service bean declaration.

Sure I can fix that :D

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -31,21 +32,21 @@ import java.text.SimpleDateFormat
 @Transactional
 class LoadDataService {
 
-    def locationDataService
+    def locationImportDataService
+    def productSupplierImportDataService
```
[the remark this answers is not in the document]
drodzewicz: I must have missed this one.
In general, I agree with the good practices of defining types for variables, but in cases of defining services I don't think it makes that much difference since we are not passing it as a variable or doing anything with it where the type would be important. 
And for intelisense, I believe the IDE handles in perfectly find.
But I agree, when possible we should define types for eveyting.

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -0,0 +1,201 @@
...
+                command.warnings[index] = []
+            }
+
+            def product = Product.findByProductCode(row.productCode)
+            if (!product) {
+                command.errors.reject("error.product.notExists", "Row ${rowIndex}: Product '${row.productCode}' does not exist")
+                command.warnings[index] << "Product '${row.productCode}' does not exist"
+            } else {
+                def lotNumber = row.lotNumber
+                if (lotNumber instanceof Double) {
```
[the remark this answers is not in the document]
drodzewicz: I did not go that deep into the logic behind validation or import, all I did was move these methods into appropriate files.

I feel like there are generally much bigger problems with some of this logic which needs to be refactored that we need to address individually.

I specifically avoided making any logical changes so I don't introduce any potential bugs that weren't there before.

alannadolny (approved):

PR #4372, opened 2023-11-13
OBPIH-5966 Fix issue causing user without purchase approver role to not be able to import purchase items

drodzewicz:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1039,7 +1039,7 @@ class OrderService {
     }
 
     def canOrderItemBeEdited(OrderItem orderItem, User user) {
-        def isPending = orderItem?.order?.status == OrderStatus.PENDING
+        def isPending = orderItem?.order?.status < OrderStatus.PENDING
```
kchelstowski, line 1042: actually why didn't it work before? There seems not to be any "lower" status than pending? :thinking:

awalkowiak: This one baffles me a little bit. This does not make sense to me. Could you try using `order.isPending` instead?

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -709,6 +709,9 @@ class OrderService {
             log.info "Order line items " + orderItems
 
             Order order = Order.get(orderId)
+            if (!isOrderEditable(order, user)) {
+                throw new UnsupportedOperationException("You do not have permissions to perform this action")
```
kchelstowski, line 713: How about translating it? We throw this exception in many places using the translated version 
```suggestion
                throw new UnsupportedOperationException("${g.message(code: 'errors.noPermissions.label')}")
```

kchelstowski (approved):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -1174,8 +1175,10 @@ class OrderService {
 
     def deleteAdjustment(OrderAdjustment orderAdjustment, User user) {
         Order order = orderAdjustment.order;
+        def g = grailsApplication.mainContext.getBean('org.grails.plugins.web.taglib.ApplicationTagLib')
+
         if (!canManageAdjustments(order, user) || orderAdjustment.hasInvoices){
-            throw new UnsupportedOperationException("You do not have permissions to perform this action")
+            throw new UnsupportedOperationException(g.message(code: "errors.noPermissions.label", "You do not have permissions to perform this action"))
```
kchelstowski, line 1181: does it work without `default: `? I mean the fallback part

PR #4373, opened 2023-11-13
OBGM-699 Fix omited refactor of dashboard redirects

drodzewicz:
- added missing CONTEXT_PATH to /dashboard link
- refactored omited hardcoded dshboard redirects
- fix create new location redirect

### DIFF
--- a/src/js/components/locations-configuration/SuccessMessage.jsx
+++ b/src/js/components/locations-configuration/SuccessMessage.jsx
@@ -36,7 +36,13 @@ const SuccessMessage = ({ setShowSuccessMessage, history, successMessageOpen }) 
             />
           </p>
           <div className="success-buttons-section d-flex flex-column">
-            <button type="button" className="btn btn-outline-primary" onClick={() => history.push(LOCATION_CONFIGURATION_URL.create())}>
+            <button
+              type="button"
+              className="btn btn-outline-primary"
+              onClick={() => {
+                window.location = LOCATION_CONFIGURATION_URL.create();
+              }}
+            >
               <Translate
                 id="react.locationsConfiguration.success.createAnother"
                 defaultMessage="Create another location"
--- a/src/js/consts/applicationUrls.js
+++ b/src/js/consts/applicationUrls.js
@@ -5,7 +5,7 @@
 export const CONTEXT_PATH = window.CONTEXT_PATH ?? '/openboxes';
 
 const DASHBOARD_URL = {
-  base: '/dashboard',
+  base: `${CONTEXT_PATH}/dashboard`,
 };
 
 const LOCATION_CONFIGURATION_URL = {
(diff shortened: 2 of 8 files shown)


awalkowiak (approved):

kchelstowski (approved):
So how does the basePath for BrowserRouter work? I would expect it to add the contextPath when we do `history.push`. That's weird...
I give an approve, but would be grateful for an eventual post-merge explanation if you are in the topic.

src/js/consts/applicationUrls.js
```diff
@@ -5,7 +5,7 @@
 export const CONTEXT_PATH = window.CONTEXT_PATH ?? '/openboxes';
 
 const DASHBOARD_URL = {
-  base: '/dashboard',
+  base: `${CONTEXT_PATH}/dashboard`,
```
kchelstowski, line 8: what I meant by my comment is that this part should not be necessary in my opinion - it should be enough to write `history.push('/dashboard')`

src/js/components/locations-configuration/SuccessMessage.jsx
```diff
@@ -36,7 +36,13 @@ const SuccessMessage = ({ setShowSuccessMessage, history, successMessageOpen })
             />
           </p>
           <div className="success-buttons-section d-flex flex-column">
-            <button type="button" className="btn btn-outline-primary" onClick={() => history.push(LOCATION_CONFIGURATION_URL.create())}>
+            <button
+              type="button"
+              className="btn btn-outline-primary"
+              onClick={() => {
+                window.location = LOCATION_CONFIGURATION_URL.create();
```
kchelstowski, line 43: have not noticed it before - why has it been changed from `history.push` to `window.location`? It's a React page, so it smells for me a bit :thinking: not knowing the reason for that change.

drodzewicz:
> So how does the basePath for BrowserRouter work? I would expect it to add the contextPath when we do `history.push`. That's weird... I give an approve, but would be grateful for an eventual post-merge explanation if you are in the topic.

It was decided in the previous PR to drop the approach of assigning `basename={CONTENCT_PATH}` on BrowserRouter since it creates different approach to creating URLs and is inconsistent.
Once we fully migrate our application to React we will be able to set the basename and remove the context path prefixes from  `urls.base`

kchelstowski:
@drodzewicz Oh, I have not noticed that you've in the end removed the `baseName` from the BrowserRouter in the previous PR.

src/js/components/locations-configuration/SuccessMessage.jsx
drodzewicz: The issue is that at this point in our application we are on URL `/openboxes/locationsConfiguration/create/:someId` and when trying to execute `history.push(LOCATION_CONFIGURATION_URL.create())` we are trying to redirect to `/openboxes/locationsConfiguration/create` but upon executing this history.push, nothing happens since from the point of view of react-router page doesn't change.

I assume, the reason for that is that we do not have a route defined for `/openboxes/locationsConfiguration` but only for `/openboxes/locationsConfiguration/:locationId` and we handle the logic ourselves for **with** and **without** id.

image

This issue will need to be investigated separately and is most likely a tech-debt.
My suggestion would be to explicitly define separate pages/steps in the browser router instead of relying on a single page with internal state tracking the page like id or step number.

For now all I can do is rely on window.location to completely refresh the page and redirect to `/openboxes/locationsConfiguration/create`

PR #4375, opened 2023-11-14
OBGM-71 Replace springcache plugin

alannadolny:
The description is pasted in the comments under the ticket: https://pihemr.atlassian.net/browse/OBGM-71?atlOrigin=eyJpIjoiMjU1ODIzYTk4M2NiNDgzMDg3NjNiYjE2NDQ3ZDM1MzEiLCJwIjoiaiJ9

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -588,7 +588,8 @@ dependencies {
     compile 'org.grails.plugins:browser-detection:3.4.0'
 
     // enable `import grails.plugin.cache.Cacheable`
-    implementation 'org.grails.plugins:cache'
+    implementation 'org.grails.plugins:cache:4.0.0'
+    compile 'org.grails.plugins:cache-ehcache:3.0.0'
 
     // enable `grails test-app -coverage`
     testImplementation('org.grails.plugins:code-coverage:2.0.3-3') {
--- a/build.gradle
+++ b/build.gradle
@@ -588,7 +588,7 @@ dependencies {
     compile 'org.grails.plugins:browser-detection:3.4.0'
 
     // enable `import grails.plugin.cache.Cacheable`
-    implementation 'org.grails.plugins:cache:4.0.0'
+    implementation 'org.grails.plugins:cache:4.0.3'
     compile 'org.grails.plugins:cache-ehcache:3.0.0'
 
     // enable `grails test-app -coverage`
--- a/grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
@@ -2,9 +2,7 @@ package org.pih.warehouse.api
 
 import grails.converters.JSON
 import grails.core.GrailsApplication
-import grails.plugin.cache.Cacheable
 import org.pih.warehouse.core.Location
-import org.pih.warehouse.core.Organization
 import grails.gorm.transactions.Transactional
 import org.pih.warehouse.core.User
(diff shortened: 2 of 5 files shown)


grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
```diff
@@ -106,7 +104,7 @@ class DashboardApiController {
     }
 
     def getFillRateDestinations() {
-        Location location = Location.get(params.locationId?:session.warehouse.id)
+        Location location = Location.get(params.locationId ?: session.warehouse.id)
```
jmiranda, line 107: @awalkowiak How do we feel about these automatic formatting changes?

jmiranda (changes requested):
We could discuss this on the tech huddle if necessary.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -102,7 +107,8 @@ class IndicatorDataService {
         return graphData
     }
 
-    GraphData getFillRate(Location location, def destination, def params) {
+    @Cacheable(value = "dashboardCache", key = { "getFillRate-${location.id}${destination?.id ?: ''}${params.querySize}${params.listFiltersSelected}${params.value}" })
```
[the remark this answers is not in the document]
alannadolny: For now, we can't use individual cache per indicator, because `@CacheEvict` cannot have an array of cache names as an argument, so we can't clear more than one cache using it. Because of that, we won't be able to refresh caches for all indicators.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -20,13 +23,15 @@ import org.pih.warehouse.shipping.Shipment
...
 @Transactional
+@GrailsTypeChecked(TypeCheckingMode.SKIP)
 class IndicatorDataService {
 
     def dashboardService
     def dataService
     GrailsApplication grailsApplication
     def messageService
 
+    @Cacheable(value = "dashboardCache", key = { "getExpirationSummaryData-${location.id}${params.querySize}" })
```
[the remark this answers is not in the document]
alannadolny: This example that you showed is available in grails-cache with version below 4.X.X. I tried to downgrade the version to make this key work, but there are some incompatibilities between this plugin and our other plugins/grails version. In the current version, it has to be a closure.
image

build.gradle
```diff
@@ -588,7 +588,8 @@ dependencies {
     compile 'org.grails.plugins:browser-detection:3.4.0'
 
     // enable `import grails.plugin.cache.Cacheable`
-    implementation 'org.grails.plugins:cache'
+    implementation 'org.grails.plugins:cache:4.0.0'
```
jmiranda, line 591: Was there a reason for choosing 4.0.0 instead of 4.0.3?

alannadolny: I was trying a lot of versions, and I forgot to change it back to 4.0.3. Already bumped up

grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
awalkowiak: I am ok with it

awalkowiak (approved):
@jmiranda lgtm. Please merge if you are ok with the current version.

jmiranda (approved):

PR #4378, opened 2023-11-15
OBGM-563 Remove unused CustomPropertyEditorRegistrar and editors

awalkowiak:
context: https://pihemr.atlassian.net/browse/OBGM-563?focusedCommentId=147401


drodzewicz (approved):

jmiranda (commented):
As mentioned in the ticket, it possible that parameter values passed as EntityTypeCode and IdentifierTypeCode might not be bound properly. Also we need to make sure data binding for other classes works within the REST API as well.

PR #4380, opened 2023-11-15
OBPIH-5930 Add Target Bin Location, Preferred Putaway Location and Default Replenishment Source columns to Inventorylevel report export

drodzewicz:
It looks like the values weren't mapped correctly to the columns.
This doesn't look like a migration related regression, because the **binLocation** and **opreferred** are existing values on InventoryLevel model but they are probably not the values that the user expects to see.

I am a little confused to why this happened, maybe some kind of refactor that happened some time ago?


drodzewicz:
@jmiranda As was discussed on last tech-huddle, instead of replacing the **preferred** and **binLcoation** columns I added the three new requested columns **Target Bin Location**, **Preferred Putaway Location** and **Default Replenishment Source** columns

jmiranda (approved):

alannadolny (approved):

PR #4381, opened 2023-11-16
OBPIH-5918 Fix missing manufacturer column value on I PO Item details tab

drodzewicz:


alannadolny (approved):

grails-app/views/order/_itemDetails.gsp
```diff
@@ -63,7 +63,7 @@
                                 ${orderItem?.productSupplier?.supplierCode}
                             </td>
                             <td class="center">
-                                ${orderItem?.productSupplier?.manufacturerName}
+                                ${orderItem?.productSupplier?.manufacturer}
```
jmiranda, line 66: @drodzewicz @awalkowiak This should be 
```
${orderItem?.productSupplier?.manufacturer?.name}
```

awalkowiak: `Organization.toString()` returns `name`, but I am ok with being specific here. (@drodzewicz)

jmiranda: Correct. I just want to avoid a bug if/when we change toString(). Since we're making a change to fix a bug, we might as well make the change so it doesn't lead to another bug down the road.

PR #4382, opened 2023-11-17
OBGM-71 Move toJsons for the rest of the methods to service

alannadolny:
On obdev3 I noticed one issue connected to the cache, which was not visible when I ran it locally. Two more indicators were throwing this issue with the cannot call `.toJson` method because we could not cast nested types in the cache to the originally created class, so I moved the rest of the `toJson`s to be sure that this issue is completely gone (I added a comment for each of that, of course)

It will be fixed when we will be able to use Cacheable on the controller method instead of service.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
@@ -94,13 +94,13 @@ class DashboardApiController {
         Location location = Location.get(params.locationId)
         Location destination = Location.get(params.destinationLocation)
         def fillRate = indicatorDataService.getFillRate(location, destination, params)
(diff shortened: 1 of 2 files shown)


awalkowiak (approved):

jmiranda (commented):
The toJson is just a Map. How/why are we returning a JSONObject?

alannadolny:
Uh, I was mistaken that this is JSONObject type (I didn't get any errors 🤔 ). Changed.

jmiranda:
@alannadolny Groovy can coerce anything to anything if there's a constructor for it. I'm guessing the internals of Groovy would just see that code :confused: and go 
```
return new JSONObject(mapObject)
```
and hope for the best.

PR #4383, opened 2023-11-17
OBPIH-5921 There is possibility to invoice more than shipped

alannadolny:
The validation takes place after entering the quantity - we are fetching information if the entered quantity is possible. If not, we can see a red border around the input field, and the next and save buttons are disabled.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -491,7 +491,7 @@ class UrlMappings {
         }
 
         // TODO: Investigate the proper way to handle validation as a REST resource
-        "/api/invoiceItems/$id/validation"(parseRequest: true) {
+        "/api/invoiceItems/$id/validation" {
             controller = "invoiceApi"
             action = [POST: "validateInvoiceItem"]
         }
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -480,6 +480,11 @@ class UrlMappings {
             action = [GET: "getPrepaymentItems"]
         }
 
+        "/api/invoices/validateInvoiceItem" {
+            controller = "invoiceApi"
+            action = [POST: "validateInvoiceItem"]
+        }
+
         "/api/invoiceStatuses"(parseRequest: true) {
             controller = { "invoiceApi" }
             action = [GET: "statusOptions"]
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -480,11 +480,6 @@ class UrlMappings {
             action = [GET: "getPrepaymentItems"]
         }
 
-        "/api/invoices/validateInvoiceItem" {
-            controller = "invoiceApi"
-            action = [POST: "validateInvoiceItem"]
-        }
-
         "/api/invoiceStatuses"(parseRequest: true) {
             controller = { "invoiceApi" }
             action = [GET: "statusOptions"]
@@ -493,6 +488,12 @@ class UrlMappings {
         "/api/invoiceTypeCodes"(parseRequest: true) {
             controller = { "invoiceApi" }
             action = [GET: "invoiceTypeCodes"]
+        }
+
+        // TODO: Investigate the proper way to handle validation as a REST resource
+        "/api/invoiceItems/$id/validation"(parseRequest: true) {
+            controller = "invoiceApi"
+            action = [POST: "validateInvoiceItem"]
         }
 
         // Stock Transfer API
--- a/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
@@ -20,6 +20,9 @@ import org.pih.warehouse.invoice.InvoiceItem
 import org.pih.warehouse.invoice.InvoiceType
 import org.pih.warehouse.invoice.InvoiceTypeCode
 import org.pih.warehouse.invoice.InvoiceStatus
+import org.pih.warehouse.product.Product
+import org.pih.warehouse.shipping.Shipment
+import org.pih.warehouse.shipping.ShipmentItem
 
 class InvoiceApiController {
 
@@ -182,5 +185,12 @@ class InvoiceApiController {
         Invoice invoice = Invoice.get(params.id)
         List<InvoiceItem> prepaymentItems = invoice.prepaymentItems
         render([data: prepaymentItems, totalCount: prepaymentItems.size()] as JSON)
+    }
+
+    def validateInvoiceItem() {
+        JSONObject jsonObject = request.JSON
+        InvoiceItem invoiceItem = InvoiceItem.load(jsonObject?.invoiceItemId)
+        Boolean isValid = invoiceService.validateInvoiceItem(invoiceItem, jsonObject?.quantity as Integer)
+        render([isValid: isValid] as JSON)
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
@@ -20,9 +20,6 @@ import org.pih.warehouse.invoice.InvoiceItem
 import org.pih.warehouse.invoice.InvoiceType
 import org.pih.warehouse.invoice.InvoiceTypeCode
 import org.pih.warehouse.invoice.InvoiceStatus
-import org.pih.warehouse.product.Product
-import org.pih.warehouse.shipping.Shipment
-import org.pih.warehouse.shipping.ShipmentItem
 
 class InvoiceApiController {
 
--- a/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
@@ -184,10 +184,11 @@ class InvoiceApiController {
         render([data: prepaymentItems, totalCount: prepaymentItems.size()] as JSON)
     }
 
-    def validateInvoiceItem() {
-        JSONObject jsonObject = request.JSON
-        InvoiceItem invoiceItem = InvoiceItem.load(jsonObject?.invoiceItemId)
-        Boolean isValid = invoiceService.validateInvoiceItem(invoiceItem, jsonObject?.quantity as Integer)
-        render([isValid: isValid] as JSON)
+    def validateInvoiceItem(InvoiceItem invoiceItem) {
+        if (!invoiceItem.validate(['quantity'])) {
+            throw new ValidationException("Invalid invoice item", invoiceItem.errors)
+        }
+
+        render(status: 200)
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
@@ -185,7 +185,7 @@ class InvoiceApiController {
     }
 
     def validateInvoiceItem(InvoiceItem invoiceItem) {
-        if (!invoiceItem.validate(['quantity'])) {
+        if (!invoiceItem.validate()) {
             throw new ValidationException("Invalid invoice item", invoiceItem.errors)
         }
 
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
@@ -82,8 +82,17 @@ class InvoiceItem implements Serializable {
         product(nullable: true)
         glAccount(nullable: true)
         budgetCode(nullable: true)
+        quantity(nullable: false, min: 0, validator: { quantity, obj ->
+            InvoiceItem.withNewSession {
+                InvoiceItem notUpdatedInvoiceItem = InvoiceItem.get(obj?.id)
+                if (notUpdatedInvoiceItem) {
+                    Boolean isValid = quantity + (obj?.shipmentItem?.quantityInvoiced - notUpdatedInvoiceItem?.quantity) <= obj?.shipmentItem?.quantity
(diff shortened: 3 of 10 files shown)


src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -337,6 +362,46 @@ class AddItemsPage extends Component {
...
+      }
+      return item;
+    });
+
+    this.setState(previousState => ({
+      values: {
+        ...previousState.values,
+        invoiceItems: mappedInvoiceItems,
+      },
+    }));
```
drodzewicz, line 402: use the `immutability-helper` it handles such nested state changes and will shorten the code and make it much more readable.
```js
const newValues = update(this.state.values, {
  invoiceItems: {
     [rowIndex]: { isValid: { $set:  data.isValid } }
  }
})
this.setState({ values: newValues })
```

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -351,12 +416,14 @@ class AddItemsPage extends Component {
         onSubmit={() => {}}
         mutators={{ ...arrayMutators }}
         initialValues={this.state.values}
-        render={({ handleSubmit, values }) => (
+        validate={() => this.validate()}
```
drodzewicz, line 419:
```suggestion
        validate={this.validate}
```

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -180,6 +200,11 @@ class AddItemsPage extends Component {
     this.removeItem = this.removeItem.bind(this);
     this.updateRow = this.updateRow.bind(this);
     this.saveInvoiceItems = this.saveInvoiceItems.bind(this);
+    this.validateInvoiceItem = this.validateInvoiceItem.bind(this);
+
+    this.debouncedInvoiceItemValidation = _.debounce((invoiceItem, rowIndex) => {
+      this.validateInvoiceItem(invoiceItem, rowIndex);
+    }, 1000);
```
drodzewicz, line 207:
```suggestion
    this.debouncedInvoiceItemValidation = _.debounce(this.validateInvoiceItem, 1000);
```

drodzewicz (changes requested):
Overall I am fine with the approach of "send request to validate" but I am sceptical.
I am not familiar with any patterns to handling these cases, but I am wondering if there is a way to attach the data we need, to do this validation on frontend (of course it should also validate on backend) but extracting more data might degrade performance of the get invoiceItem list, so I am not too sure.

alannadolny:
@drodzewicz regarding your thoughts:
To pull the maximum available quantity for validation we have to change the InoviceItem domain (there is no information about total invoiced quantity). We have this information in InvoiceItemCandidate, so there will be a little duplication between those two domains. Moreover, when we are adding InvoiceItem on the modal we have appropriate validation, so there is no need to pull these values again unless we change the quantity to the invoice (within the table on the add items step). Moreover, somebody can invoice the same item at the same time, so we won't have correct data if we are pulling it only at the beginning. We used a similar approach in the PO shipments (this approach with fetching inventory items)

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -127,14 +127,34 @@ const FIELDS = {
...
+              rowIndex,
+            });
+          },
+          onChange: (event) => {
+            debouncedInvoiceItemValidation({
+              invoiceItemId: values.invoiceItems[rowIndex].id,
+              quantity: event,
+              rowIndex,
+            });
+          },
```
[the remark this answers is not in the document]
alannadolny: I think the validation onChange is needed here because we don't have any information about validation until we lose focus. Of course, this validation is debounced, so we don't have a lot of requests to send.
Maybe I am wrong, It would be nice to hear opinions from others. @awalkowiak @jmiranda @kchelstowski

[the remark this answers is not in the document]
awalkowiak: @alannadolny Considering the fact, that there is a post to the `/api/invoices/validateInvoiceItem` endpoint I am considering requesting a change to do validation during the items / invoice save (and throwing exception if validation won't pass), to avoid adding the rpc endpoints, then you would not have to deal with that part (when to trigger it and on what circumstance)
cc @drodzewicz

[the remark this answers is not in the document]
jmiranda: Now that we went back to validating at the point of data entry, I am not sure what is more appropriate and user-friendly. But I think my main concern would be making sure that a user is not interrupted while they are entering a value here. It might make sense for entering a username for an account signup feature. 

"jmiranda" -> That username is taken
"jmiranda2" -> All good

But when entering a quantity, I would only be ok with a separate validation request on every keypress IFF the validation error was displayed without popups, toast messages, 

"100" -> Too much, my guy
delete a 0 
"10" -> Still not ok, dude
delete another 0 
"1" -> Alright, now we're talking

If this lead to three toast messages appearing on the screen or prompts that required a click to close, I would be annoyed.

However, if there was inline validation that might be less disruptive and ok. 

image

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -337,6 +360,43 @@ class AddItemsPage extends Component {
       });
   }
 
+  validate(values) {
+    const errors = {};
+    errors.invoiceItems = [];
+
+    _.forEach(values?.invoiceItems, (item, key) => {
+      if (_.has(item, 'isValid') && !item.isValid) {
+        errors.invoiceItems[key] = { quantity: 'react.invoice.errors.quantityToInvoice.label' };
```
jmiranda, line 369: Server should be returning this error message, for this field.

jmiranda (changes requested):
Check comments

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -490,6 +490,12 @@ class UrlMappings {
             action = [GET: "invoiceTypeCodes"]
         }
 
+        // TODO: Investigate the proper way to handle validation as a REST resource
+        "/api/invoiceItems/$id/validation"(parseRequest: true) {
```
jmiranda, line 494: AFAIK parseRequest is no longer applicable in Grails 3.

https://pihemr.atlassian.net/browse/OBGM-401?focusedCommentId=139786
https://stackoverflow.com/a/24048120/136597

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -82,8 +82,17 @@ class InvoiceItem implements Serializable {
         product(nullable: true)
         glAccount(nullable: true)
         budgetCode(nullable: true)
-
-        quantity(nullable: false, min: 0) // min = 0 for canceled items
+        quantity(nullable: false, min: 0, validator: { quantity, obj ->
+            InvoiceItem.withNewSession {
+                InvoiceItem notUpdatedInvoiceItem = InvoiceItem.get(obj?.id)
```
jmiranda, line 87: This scares me a little bit. 

I've only ever seen code that does a findBy or exists. 

https://github.com/mcroteau/greenfield-ecommerce/blob/974442fc4dbe2c7a79a8c825313d53f5cef23791/grails-app/domain/org/greenfield/AccountRole.groovy#L91

For now, make this object read-only so we're not doing anything untoward. 
```
InvoiceItem.read(obj.id)
```
My concern is that while we're creating a new Hibernate session we're still participating in the same transaction and I don't know how things will behave if something unforeseen happens.

If this does cause any issues while testing, perhaps we need a different way to get the original value 
* grails should have a property where it stores the original values (preferred)
* named query on InvoiceItem to return just the value we want i.e. original invoiceItem.quantity (hate it, but would be better than hydrating the full object)

jmiranda: I found what I was looking for ... getPersistentValue https://docs.grails.org/latest/ref/Domain%20Classes/getPersistentValue.html 

https://stackoverflow.com/questions/5005932/grails-custom-validator-based-on-a-previous-value-of-the-field

jmiranda: Let's use that if we can.

jmiranda (changes requested):
The query in constraints is probably going to cause an issue at some point, so let's try to refactor that if we can.

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -183,4 +183,12 @@ class InvoiceApiController {
         List<InvoiceItem> prepaymentItems = invoice.prepaymentItems
         render([data: prepaymentItems, totalCount: prepaymentItems.size()] as JSON)
     }
+
+    def validateInvoiceItem(InvoiceItem invoiceItem) {
+        if (!invoiceItem.validate(['quantity'])) {
```
kchelstowski, line 188: if we already touch the invoice item, shouldn't we try to validate it "deeply" (it means: not only the quantity, but whole object)?
If you agreed on validating the quantity only, then the method in my opinion should be called `validateInvoiceItemQuantity`, not `validateInvoiceItem`.

kchelstowski:
```suggestion
        if (!invoiceItem.validate()) {
```

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -82,8 +82,17 @@ class InvoiceItem implements Serializable {
         product(nullable: true)
         glAccount(nullable: true)
         budgetCode(nullable: true)
-
-        quantity(nullable: false, min: 0) // min = 0 for canceled items
+        quantity(nullable: false, min: 0, validator: { quantity, obj ->
```
kchelstowski, line 85: could you add types for those arguments? I'm always confused in the validator what is what (I know the names describe it, but still it shouldn't be a pain to change it).

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -82,8 +82,17 @@ class InvoiceItem implements Serializable {
...
+            InvoiceItem.withNewSession {
+                InvoiceItem notUpdatedInvoiceItem = InvoiceItem.get(obj?.id)
+                if (notUpdatedInvoiceItem) {
+                    Boolean isValid = quantity + (obj?.shipmentItem?.quantityInvoiced - notUpdatedInvoiceItem?.quantity) <= obj?.shipmentItem?.quantity
```
kchelstowski, line 89: for a better visibility let's define a shipment item above:
```groovy
ShipmentItem shipmentItem = obj?.shipmentItem
```
I would also consider adding a comment above `isValid` describing in a nutshell what we treat as a valid/invalid invoice item quantity.

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -256,6 +256,10 @@ class ShipmentItem implements Comparable, Serializable {
         return inventoryItem?.lotStatus == LotStatusCode.RECALLED
     }
 
+    Integer getQuantityInvoiced() {
```
kchelstowski, line 259: not a change request, but just FYI - this doesn't need to be a "heavier" `Integer`, but could be `int`, since this cannot be `null`. (You made a fallback for `0`.

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -337,6 +361,53 @@ class AddItemsPage extends Component {
...
+  }
+
+  async validateInvoiceItem({
+    invoiceItem,
+    rowIndex,
+  }) {
+    this.debouncedInvoiceItemValidation.cancel();
+
+    await invoiceItemApi.validateInvoiceItem(invoiceItem)
+      .then(() => {
```
kchelstowski, line 384: if you use the `async/await` approach, can't we use the `try/catch`?

src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -337,6 +361,53 @@ class AddItemsPage extends Component {
...
+        const updatedValues = update(this.state.values, {
```
kchelstowski, line 385: I honestly hate this `update` method and everything from this package, I'd prefer a simple `.map`, but saw the change request....

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
alannadolny: If there will be a need for validating more than this one property we can add it, for now, it's unnecessary

kchelstowski (approved):

drodzewicz (approved):

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -82,8 +82,19 @@ class InvoiceItem implements Serializable {
         product(nullable: true)
         glAccount(nullable: true)
         budgetCode(nullable: true)
-
-        quantity(nullable: false, min: 0) // min = 0 for canceled items
+        quantity(nullable: false, min: 0, validator: { Integer quantity, InvoiceItem obj ->
+                Integer notUpdatedQuantity = obj.getPersistentValue('quantity')
```
jmiranda, line 86: We still need to come up with a better name for this variable.
`quantityInvoicedOriginal` or `originalQuantityInvoiced` would be my vote. Or even `previousQuantityInvoiced`, `oldQuantityInvoiced` or `savedQuantityInvoiced` would suffice.

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
jmiranda: I'm unsure how I feel about this. @kchelstowski is right, but I don't want to change the scope of the method to the quantity. So I'd prefer keeping the scope at the invoice item, validating the whole object, and if necessary narrow the scope in the client. In other words, pull out just the quantity error message to display to the user.

jmiranda (changes requested):

awalkowiak (approved):

awalkowiak:
I had to remove my pending review twice because @alannadolny was faster with changes.

jmiranda (approved):
Looks good to me. The validation error message might require tweaking but I'd ask Kasia or Manon for that.

PR #4385, opened 2023-11-17
OBGM-310 Review configuration that was not migrated from Config.groovy

alannadolny:
I removed the things that were agreed to delete and I also removed the reference to the raven plugin in the auth controller


jmiranda (approved):
This looks good to me.

awalkowiak (approved):

PR #4386, opened 2023-11-20
OBPIH-5918 Access name attribute directly instead of using toString

drodzewicz:


awalkowiak (approved):

jmiranda (approved):
Thank you. Sorry for being a PITA.

PR #4387, opened 2023-11-21
OBPIH-5977 Display items in PO Template in the same order as entered during PO creation

alannadolny:


kchelstowski (approved):
a question - what makes it sorted in the view? I expected you to change it in two places, but I see that you've changed it only for the template. 
If it is sorted "by hand" in the view comparing the dates, this could be also shortened there to `.sort()` as you've overwritten the `compareTo` method.

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -111,4 +111,9 @@ class OrderAdjustment implements Serializable {
     Boolean getHasRegularInvoice() {
         return invoices.any { it.invoiceType == null || it.invoiceType?.code == InvoiceTypeCode.INVOICE }
     }
+
+    @Override
+    int compareTo(OrderAdjustment o) {
+        return dateCreated <=> o.dateCreated
```
jmiranda, line 117: Careful with this. This would disappear adjustments with the same dateCreated that are added to a TreeSet or TreeMap. 

> Note that the ordering maintained by a set (whether or not an explicit comparator is provided) must be consistent with equals if it is to correctly implement the Set interface. (See Comparable or Comparator for a precise definition of consistent with equals.) This is so because the Set interface is defined in terms of the equals operation, but a TreeSet instance performs all element comparisons using its compareTo (or compare) method, so two elements that are deemed equal by this method are, from the standpoint of the set, equal. The behavior of a set is well-defined even if its ordering is inconsistent with equals; it just fails to obey the general contract of the Set interface.

https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html

jmiranda: I usually do something like this 
```
    @Override
    int compareTo(OrderAdjustment obj) {
        return dateCreated <=> obj?.dateCreated ?:
                lastUpdated <=> obj?.lastUpdated ?:
                        id <=> party.id
    }
```

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -114,7 +114,7 @@ class DocumentTemplateService {
                     expectedShippingDate: orderItem?.estimatedReadyDate ? orderItem?.estimatedReadyDate?.format("MMM dd, yyyy") : ""
             ]
         }
-        def orderAdjustments = orderInstance?.orderAdjustments?.findAll { !it.canceled }?.collect { OrderAdjustment orderAdjustment ->
+        def orderAdjustments = orderInstance?.orderAdjustments?.findAll { !it.canceled }?.sort()?.collect { OrderAdjustment orderAdjustment ->
```
jmiranda, line 117: This feels like a case where the default sort order of the association would make sense.

Either in the domain class
```
class OrderAdjustment {
    static mapping = { 
        sort dateCreated : "desc"
    }
}
```
Or in the association on Order
```
class Order { 
    static mapping = {
        orderAdjustments(sort:'dateCreated', order:'desc')
    }
}
```

grails-app/views/order/_orderAdjustments.gsp
```diff
@@ -17,9 +17,9 @@
             </tr>
             </thead>
             <tbody>
-            <g:each var="orderAdjustment" in="${orderInstance.orderAdjustments?.sort { it.dateCreated }}" status="status">
+            <g:each var="orderAdjustment" in="${orderInstance.orderAdjustments?.sort()}" status="status">
```
jmiranda, line 20: Are there other places where we display order adjustments?

jmiranda (changes requested):

grails-app/views/order/_orderAdjustments.gsp
alannadolny: I can see them on `order/print`
```
orderInstance?.orderAdjustments.findAll { !(it.orderItem || it.canceled) }.sort { it.totalAdjustments }.reverse()
```

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
alannadolny: I followed your solution with `compareTo` with fallbacks

grails-app/views/order/_orderAdjustments.gsp
jmiranda: @alannadolny Ok cool. Can you include that in the ticket (@ Manon) with 1) a brief description of what is happening in that code 2) question on whether these need to be sorted as well?

jmiranda (approved):
LGTM. Please see comment below about what to add to the ticket for Manon's review.

awalkowiak (approved):

PR #4391, opened 2023-11-24
OBGM-130 Refactor database migration trigger in BootStrap.groovy to avoid having to update list of existing versions

alannadolny:
The description / Question is moved to the [ticket](https://pihemr.atlassian.net/browse/OBGM-130?atlOrigin=eyJpIjoiMjI4M2FiN2IwMTlkNDkwZjliMmQ0NWRmZWNiNWFhZDAiLCJwIjoiaiJ9)

### DIFF
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -654,10 +654,11 @@ class BootStrap {
             PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver()
             Resource[] resources = resolver.getResources("file:grails-app/migrations/*")
 
-            // Find directories with names matching current versions pattern
+            // Find directories with names matching current versions pattern which
+            // will match to any version number between 0.0.x to 999.999.x
             List<String> changelogVersions = resources
                     .collect { it.filename }
-                    .findAll { it.matches("[0-9]{1,3}.[0-9]{1,3}.x") }
+                    .findAll { it.matches("\\d{1,3}.\\d{1,3}.x") }
                     .reverse()
 
             // Exclude the newest changelog version, this one should be run separately
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -648,8 +648,10 @@ class BootStrap {
             liquibase = new Liquibase('views/drop-all-views.xml', new ClassLoaderResourceAccessor(), database)
             liquibase.update(null as Contexts, new LabelExpression());
 
-            // FIXME Not good that we'll need to update this on subsequent versions so this needs some more thought
-            List previousChangelogVersions = ["0.5.x", "0.6.x", "0.7.x", "0.8.x"]
+            // Find directories with names matching current versions pattern
+            List<String> changelogVersions = new File('grails-app/migrations').list().findAll { it.matches("[0-9].[0-9].x") }
+            // Exclude the newest changelog version, this one should be run separately
+            List<String> previousChangelogVersions = changelogVersions.tail()
 
             // Check if the executed changelog versions include one of the previous versions
             // and if so, then we need to keep running the old updates to catch up to 0.9.x
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -651,7 +651,7 @@ class BootStrap {
             // Find directories with names matching current versions pattern
             List<String> changelogVersions = new File('grails-app/migrations').list().findAll { it.matches("[0-9]{1,3}.[0-9]{1,3}.x") }
             // Exclude the newest changelog version, this one should be run separately
-            List<String> previousChangelogVersions = changelogVersions.tail()
+            List<String> previousChangelogVersions = changelogVersions.size() ? changelogVersions.tail() : []
 
             // Check if the executed changelog versions include one of the previous versions
             // and if so, then we need to keep running the old updates to catch up to 0.9.x
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -649,7 +649,7 @@ class BootStrap {
             liquibase.update(null as Contexts, new LabelExpression());
 
             // Find directories with names matching current versions pattern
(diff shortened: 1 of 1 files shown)


grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -605,8 +605,10 @@ class BootStrap {
             liquibase = new Liquibase('views/drop-all-views.xml', new ClassLoaderResourceAccessor(), database)
             liquibase.update(null as Contexts, new LabelExpression());
 
-            // FIXME Not good that we'll need to update this on subsequent versions so this needs some more thought
-            List previousChangelogVersions = ["0.5.x", "0.6.x", "0.7.x", "0.8.x"]
+            // Find directories with names matching current versions pattern
+            List<String> changelogVersions = new File('grails-app/migrations').list().findAll { it.matches("[0-9]{1,3}.[0-9]{1,3}.x") }
```
jmiranda, line 609: I just realized that you're looking at the changelog files rather than the database. We need to look at what changelogs have been executed already, not what could be executed.

jmiranda (changes requested):
Let's hold off on this one until we have a chance to tech huddle it.

grails-app/init/org/pih/warehouse/BootStrap.groovy
jmiranda: Sorry I confused this with the executedChangelogVersions above. So I think this is probably correct. I'm still worred about using the `File` API here because I don't know if that's going to work in all environments. 

I would either make use of `grailsResourceLocator`, `servletContext.getResourceAsStream` or something similar.

We use the former in LocalizationServer to pull in the appropriate message.properties for a locale.

Using the ServletContext would look something like this.

https://docs.grails.org/latest/ref/Servlet%20API/servletContext.html

jmiranda: @awalkowiak I know we discussed this today but I couldn't tell if you meant that you wanted to merge this as-is or push it off to the future. I'm ok with either.

grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -605,8 +607,18 @@ class BootStrap {
...
-            // FIXME Not good that we'll need to update this on subsequent versions so this needs some more thought
-            List previousChangelogVersions = ["0.5.x", "0.6.x", "0.7.x", "0.8.x"]
+            // Get all directories from /migrations
+            PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver()
+            Resource[] resources = resolver.getResources("file:grails-app/migrations/*")
+
+            // Find directories with names matching current versions pattern
+            List<String> changelogVersions = resources
+                    .collect { it.filename }
+                    .findAll { it.matches("[0-9]{1,3}.[0-9]{1,3}.x") }
```
jmiranda, line 617: Can you also add a comment to describe the regexp being used? Something like 
```
// Find directories with names matching current versions pattern which 
// will match to any version number between 0.0.x to 999.999.x
```
Also let's replace [0-9] with 
```
.findAll { it.matches("[[:digit:]]{1,3}.[[:digit:]]{1,3}.x") }
```

grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -605,8 +607,18 @@ class BootStrap {
...
+                    .reverse()
+
+            // Exclude the newest changelog version, this one should be run separately
+            List<String> previousChangelogVersions = changelogVersions.size() ? changelogVersions.tail() : []
```
jmiranda, line 621: Also, are we sure that the resource resolver will return the directories in the right order every time. Or that findAll will? Or should we add an explicit alphanumeric sort to make sure we get the list back in the proper order.

grails-app/init/org/pih/warehouse/BootStrap.groovy
alannadolny: It's interesting, because `[[:digit:]]` seems to be not working :thinking: 
image
neither with single `[`:
image
but it works with \d and just [0-9]:
image

grails-app/init/org/pih/warehouse/BootStrap.groovy
alannadolny: Directories are listed in the same order as they are in the /migrations directory. I created 0.4.x, 1.0.x, and 1.1.x to see how they will behave - each of those directories was placed in the right order when I created them:
image
0.4.x is the first of the list, and the 1.0.x and 1.1.x are at the end of the list, so because of that I am getting the list in the right order (from 0.4.x to 1.1.x).
So this is a list of changelog versions:
image
And the tail of it:
image

grails-app/init/org/pih/warehouse/BootStrap.groovy
jmiranda: Let's use \\d then.

jmiranda (commented):
One more thing to test (see below).

jmiranda:
Also, fix the conflict.

grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -605,8 +607,19 @@ class BootStrap {
...
+
+            // Find directories with names matching current versions pattern which
+            // will match to any version number between 0.0.x to 999.999.x
+            List<String> changelogVersions = resources
+                    .collect { it.filename }
+                    .findAll { it.matches("\\d{1,3}.\\d{1,3}.x") }
+                    .reverse()
+
+            // Exclude the newest changelog version, this one should be run separately
+            List<String> previousChangelogVersions = changelogVersions.size() ? changelogVersions.tail() : []
```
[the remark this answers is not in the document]
alannadolny: `.size` was working, but I changed this method to `.empty` to increase readability

awalkowiak (changes requested):
While working on other tickets related to db changelogs I noticed, that if you run the `install` changelogs on fresh db, and then current `0.9.0` changelogs, then on each app start bootstrap thinks that previous changelogs were not yet executed and will try to again install db from scratch. I think we need to expand the logic here to cover or look into either if there were any `install` changelogs executed or any `current version` changelogs executed, and based on that execute proper changelogs.

(part of the issues I had were caused after changing a branch and having different changeset IDs in the `install/changelogs...`, so normally it should be no problem to rerun install changelogs, because those executed would be skipped, so treat this rather as a "comment", not necessarily as a "change request", but I'd like to discuss it)

awalkowiak (approved):

PR #4393, opened 2023-11-27
OBGM-337 Recreate schema and data changelogs for new installs 

awalkowiak:
- regenerated schema after the 0.8.23 release ([OBGM-337](https://pihemr.atlassian.net/browse/OBGM-337))
- added missing data after the 0.8.23 release ([OBGM-202](https://pihemr.atlassian.net/browse/OBGM-202))
- cleaned up some data from the changelog with initial data insert ([OBGM-136](https://pihemr.atlassian.net/browse/OBGM-136))

[OBGM-337]: https://pihemr.atlassian.net/browse/OBGM-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OBGM-202]: https://pihemr.atlassian.net/browse/OBGM-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OBGM-136]: https://pihemr.atlassian.net/browse/OBGM-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


grails-app/migrations/install/changelog-create-schema.groovy
```diff
@@ -1600,7 +1604,7 @@ databaseChangeLog = {
 
             column(name: "quantity_ordered", type: "DECIMAL(65)")
 
-            column(name: "adjustments_count", type: "DECIMAL(45)")
+            column(name: "adjustments_count", type: "DECIMAL(44)")
```
jmiranda, line 1607: Can you explain the reason for this change? I'm not saying the original "45" was any better, just don't understand why 44 is any better?

I know these are views (and actually probably need to be removed from the create schema file) but we should actually come up with some sensible defaults for fields. 

* For counts, we probably want this to be an integer. 
* For currency values, I like to use DECIMAL(19, 2). But I just learned that to be compatible with GAAP accounting practices, then you need at least 4 decimal places to the right of the separator. 
* For quantity, eventually I want to use DECIMAL but INT is fine for now, especially for views.

https://stackoverflow.com/questions/13030368/best-data-type-to-store-money-values-in-mysql
https://rietta.com/blog/best-data-types-for-currencymoney-in/

jmiranda: As a reminder, the original changelog-create-schema.groovy is not special. We don't need to preserve it. It was created after the first rebase and then recreated for several rebases after that. But we don't need to keep it intact. In fact, I'm starting to think maybe it would be best to break it back up into the different objects (tables, fks, indices, unique constraints, etc)

grails-app/migrations/install/changelog-create-schema.groovy
```diff
@@ -2342,7 +2346,9 @@ databaseChangeLog = {
                 constraints(nullable: "false")
             }
 
-            column(name: "request_item_id", type: "CHAR(38)")
+            column(defaultValue: "", name: "request_item_id", type: "CHAR(38)") {
+                constraints(nullable: "false")
```
jmiranda, line 2350: It feels wrong to use defaultValue = "" and nullable = false. Do we need to use defaultValue? I assume this is part of the dbm-gorm-diff logic, but we should figure out if 1) we need it 2) if not, how to disable it.

grails-app/migrations/install/changelog-create-schema.groovy
```diff
@@ -3049,10 +3065,44 @@ databaseChangeLog = {
             column(name: "source_type", type: "VARCHAR(255)")
 
             column(name: "replenishment_type_code", type: "VARCHAR(255)")
+
+            column(name: "approved_by_id", type: "CHAR(38)")
+
+            column(name: "date_approved", type: "datetime")
```
jmiranda, line 3071: Capitalize the DATETIME.

grails-app/migrations/install/changelog-insert-data.xml
```diff
@@ -372,6 +372,46 @@
             <column name="name" value="Partially Received"/>
             <column name="sort_order" valueNumeric="3"/>
         </insert>
+        <insert tableName="event_type">
+            <column name="id" value="abb5b3d2-da3e-4a4c-9e3c-de139147f822"/>
```
jmiranda, line 376: Are these the same exact IDs as the ones created in develop?

jmiranda (commented):
My first thought while reviewing this was "I feel like I'm not seeing all of the changes I was expecting to see."

@awalkowiak Can you explain in the ticket how you generated this file? It really needs to be based off a version of the database created while running develop or master. I'm guessing there are changes missing if we don't go that route and we just add tables that were created recently on develop.

jmiranda:
> My first thought while reviewing this was "I feel like I'm not seeing all of the changes I was expecting to see."

I forgot that I reran the dbm-changelog script thing back in August. So that explains why there aren't as many changes as I expected.

grails-app/migrations/install/changelog-create-schema.groovy
awalkowiak: > Can you explain the reason for this change? I'm not saying the original "45" was any better, just don't understand why 44 is any better?

It's what was generated. I don't know why it chose 44 over 45. I can dig that topic a little bit.

grails-app/migrations/install/changelog-create-schema.groovy
awalkowiak: I think that it should be the former version (without defaultValue), but will investigate it a bit more.

grails-app/migrations/install/changelog-create-schema.groovy
awalkowiak: Sure, will do it (but I'll need to adjust it in the entire file, because it is lower in every case of `datetime`)

grails-app/migrations/install/changelog-insert-data.xml
awalkowiak: I used here exactly what is in the changelog adding this event type

awalkowiak:
> I forgot that I reran the dbm-changelog script thing back in August. So that explains why there aren't as many changes as I expected.

Yeah, I regenerated it on the master, and it basically regenerated only the parts that were added in the 0.8.23 (everything related to the "Request Approvals" feature

grails-app/migrations/install/changelog-create-schema.groovy
awalkowiak: > changelog-create-schema.groovy is not special. We don't need to preserve it. It was created after the first rebase and then recreated for several rebases after that. But we don't need to keep it intact. In fact, I'm starting to think maybe it would be best to break it back up into the different objects (tables, fks, indices, unique constraints, etc)

Ok, I'll split it into smaller chunks.

grails-app/migrations/install/changelog-create-schema.groovy
jmiranda: So that was auto-generated as well?

awalkowiak: Yup

drodzewicz (changes requested):
As a compromise to closing this ticket https://pihemr.atlassian.net/browse/OBGM-112 it was decided that we would use the opportunity of creating _new create schema changelog_ to update **role** table **name** column to be `nullable: false`, so please update the nullable constraint on the **role** table.

Additionally, I created a [pull request](https://github.com/openboxes/openboxes/pull/4368/files) that updated all of the missing role names, but I think it would also be nice if these changes would appear in the `changelog-insert-data.xsml`

awalkowiak:
@drodzewicz
> please update the nullable constraint on the role table.

Sure, will do it.
> I think it would also be nice if these changes would appear in the changelog-insert-data.xsml

but these are in the 0.9.0 already, so I think this is not required in the insert of initial data because it will be applied with all the 0.9 changelogs :thinking:

jmiranda (approved):

grails-app/conf/application.yml
```diff
@@ -539,7 +539,8 @@ openboxes:
 
     # Allow users to customize logo image url as well as label
     logo:
-        url: "/assets/openboxes_logo_40x40.jpg"
+        # By default, the context path is /openboxes so lets use it here, because we cannot reference it in another way
+        url: "/openboxes/assets/openboxes_logo_40x40.jpg"
```
jmiranda, line 543: Let's get rid of this change for now. If it actually needs to be fixed, let's do it in a separate PR against a ticket that focuses on this issue.

I assume that in places where logo.url is used, we can just deal with the context path ourselves in Grails. If we need the context path to be set for some reason, then we should do something like this for Spring Boot 1.3 (what Grails 3 uses)
```
server:
    context-path: /openboxes
```
... but we should also be able to override the context path at runtime using an environment variable 
```
export SERVER_CONTEXT_PATH=/openboxes
```
or command line argument
```
$ java -jar openboxes.war --server.context-path=/openboxes
```
But my hope was always that we'd deploy the application, the app name or context path would derived from the name of the war file and that would be passed into the application as-is. The only issue we were going to run into was that React didn't know the context path and that was going to need to be passed as a command line argument or environment variable.

If for some reasons I'm wrong, we should set this to be an external CDN URL? This would be a worst case scenario as it a little deceiving (we would essentially be adding one of those embedded images in emails that phones home to tell the sender who has opened their emails - I'm not a fan of those).

jmiranda (changes requested):
Everything looks good to me except this first change. If we need to deal with it, let's add a new ticket so there's a discussion.

awalkowiak:
Let's hold with merge on this one, I want to test it first after merging Alan's https://github.com/openboxes/openboxes/pull/4391

PR #4396, opened 2023-11-30
OBGM-736 Very slow import of many items on Add items to PO

alannadolny:
description: https://pihemr.atlassian.net/browse/OBGM-736?focusedCommentId=147954


jmiranda (commented):
I'm ok with this, but it doesn't feel like a robust solution. See discussion in ticket.

kchelstowski (approved):
as I mentioned in the ticket - I'm fine with that as long as it doesn't break anything

awalkowiak (approved):

PR #4399, opened 2023-12-01
OBGM-744 Fix ability to add adjustments to invoices

kchelstowski:


alannadolny (approved):

jmiranda (approved):
This looks good to me.

I would be remiss if I didn't mention that I feel slightly uncomfortable with these changes. I know almost nothing about the invoice feature at this point, but it seems we have lots of flexibility within the feature (how invoices are created, when invoices are paid, what data the invoice relates to i.e. shipments, shipment items, orders, order items). 

So adding even a small change to the validation logic, creates a branching of test cases that seems to be growing a bit unwiedy. 

If others feel uncomfortable as well, then I think we need to discuss it with Manon and Katarzyna to address it in some way before we run into a production bug that might be difficult to unravel. 

My only suggestion is to make sure we start codifying integration tests around this feature as soon as we can.

cc @kchelstowski @alannadolny @awalkowiak @drodzewicz

PR #4400, opened 2023-12-01
OBPIH-5990 Able to delete POSM when at least pending invoice exists

alannadolny:


grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -256,8 +256,8 @@ class ShipmentItem implements Comparable, Serializable {
         return inventoryItem?.lotStatus == LotStatusCode.RECALLED
     }
 
-    int getQuantityInvoiced() {
-        return invoiceItems.sum { it.quantity ?: 0 }
+    Integer getQuantityInvoiced() {
```
kchelstowski, line 259: out of curiosity - was there a reason for that change? `invoiceItems` seems to be a collection, so it should never be `null` - in the "worst case scenario" it should be an empty Set.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -671,5 +671,9 @@ class Shipment implements Comparable, Serializable {
         Event event = events.find { Event event -> event?.eventType?.eventCode == EventCode.CUSTOMS_RELEASE }
         return event?.eventDate
     }
+
+    Boolean hasInvoicedItem() {
```
kchelstowski, line 675: nitpicky: this could be a primitive `boolean` (`shipmentItems` can't be null)

kchelstowski (approved):

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
alannadolny: When I was trying to evaluate my code from the validating method I saw it was throwing an exception (as I remember, it was in case when I rolled back stock movement and deleted the connected invoice)

alannadolny:
@jmiranda @awalkowiak I changed the base branch to hotfix, as was planned in the ticket

jmiranda (approved):

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: @jmiranda agreed with you. I wouldn't mind (and actually I don't mind, that's why I've approved the PR without this change) the `Boolean` if the value could be null - in this case it can't - no matter what would happen.
I think I've raised this topic once there (or actually it was about naming): https://github.com/openboxes/openboxes/pull/4266#discussion_r1316064617

With that said I'd also be happy to discuss it, because I have read a lot of discussions, about this topic (as it's always been bothering me) with pros and cons of using both versions without a clear answer what's the best.

jmiranda: @kchelstowski Yeah I think the larger discussion around using wrappers (Integer) vs primitives (int) is probably one that should be investigated, discussed on a tech huddle, and added to our coding conventions. I've added that to a ticket for coding convention topics that need to be discussed https://pihemr.atlassian.net/browse/OBS-1406.

PR #4402, opened 2023-12-05
OBGM-745 Default dashboard configuration broken for new users 

alannadolny:
Description: https://pihemr.atlassian.net/browse/OBGM-745?focusedCommentId=148114

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -477,6 +477,10 @@ class UserService {
 
     def getDashboardConfig(User user, String id) {
         Gson gson = new Gson()
+        // Creating a deep copy of the dashboard configuration
+        // to avoid overriding config properties.
+        // Deep copy is created here using the gson library
+        // in the way mentioned in this article: https://www.baeldung.com/java-deep-copy
         def fullConfig = gson.fromJson(
                 gson.toJson(Holders.config.openboxes.dashboardConfig),
                 Object
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -9,8 +9,10 @@
  **/
 package org.pih.warehouse.core
 
+import com.google.gson.Gson
 import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
+import grails.util.Holders
 import grails.validation.ValidationException
 import groovy.sql.Sql
 import org.apache.commons.collections.ListUtils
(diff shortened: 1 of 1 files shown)


kchelstowski:
I'd be also good to figure out if marking this method as `@Transactional(readOnly = true)` would help - I'm wondering if this would create a separate transaction for that method, or it would use the current transaction and would just mark it as `readOnly` + if it would "come back" to the previous `readOnly` setting after execution of that method.

I'm just saying it as a potential spike/investigation ticket to figure that out.

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -474,11 +476,14 @@ class UserService {
...
-        def mainDashboardId = grailsApplication.config.openboxes.dashboardConfig.mainDashboardId
+        Gson gson = new Gson()
+        def fullConfig = gson.fromJson(
+                gson.toJson(Holders.config.openboxes.dashboardConfig),
+                Object
+        )
+        def mainDashboardId = Holders.config.openboxes.dashboardConfig.mainDashboardId
         def resultConfig = [
-                // Clone to not overwrite the main config in the line 422
-                dashboard: fullConfig.dashboards[id ?: mainDashboardId].clone(),
```
awalkowiak, line 481: Why did this stop working? What's the change here between Grails 1?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -474,11 +476,14 @@ class UserService {
     }
 
     def getDashboardConfig(User user, String id) {
-        def fullConfig = grailsApplication.config.openboxes.dashboardConfig
-        def mainDashboardId = grailsApplication.config.openboxes.dashboardConfig.mainDashboardId
+        Gson gson = new Gson()
+        def fullConfig = gson.fromJson(
```
awalkowiak, line 480: Is there a more elegant way to achieve this? If the `.clone()` does not work, and there is a need to do a deep copy.

awalkowiak (commented):
@kchelstowski I think the transactional should not be an issue because it is a config from runtime. I wonder why this worked on grails 1 and stopped working on grails 3 (I guess it was a shallow copy back then too, but I have not checked that on g1 yet)

alannadolny:
@kchelstowski `readOnly` in `@Transactional` is related to persisting entities in database only, it won't work here

grails-app/services/org/pih/warehouse/core/UserService.groovy
alannadolny: I didn't find any better solution for creating a deep copy, this one looks better for me, because we are currently using the gson library so I didn't have to install a new one.
Mainly I based on this article: https://www.baeldung.com/java-deep-copy

grails-app/services/org/pih/warehouse/core/UserService.groovy
alannadolny: I was trying to find out why it stopped working, but I didn't find anything.
Config properties can be overridden in both cases - in Grails 1 and Grails 3.
In Grails 1, it is also a shallow copy, so honestly, it shouldn't work either.
One difference is just in the implementation of configs - in Grails 1 it is ConfigObject, and in Grails 3 it is NavigableMap

grails-app/services/org/pih/warehouse/core/UserService.groovy
awalkowiak: @alannadolny ok, please add a comment above this line about what is happening here and link this article too.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4403, opened 2023-12-05
OBGM-149 Fix issues that came out during investigation of ajax tags

drodzewicz:
- added @Transactional to addProductGroupToProduct which fixes the persistence of the change to the database
- added nulsafe operator on productGroups which fixes 500 error thrown upon adding product to product group

these are small fixes that I came across when testing ajax tags

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -810,6 +810,7 @@ class ProductController {
      *
      * @return
      */
+    @Transactional
     def addProductGroupToProduct() {
         println "addProductGroupToProduct() " + params
         def product = Product.get(params.id)
--- a/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
@@ -49,6 +49,7 @@ class ProductController {
     def productMergeService
     UploadService uploadService
     def localizationService
+    ProductDataService productDataService
 
     static allowedMethods = [save: "POST", update: "POST"]
 
@@ -810,7 +811,6 @@ class ProductController {
      *
      * @return
      */
-    @Transactional
     def addProductGroupToProduct() {
         println "addProductGroupToProduct() " + params
         def product = Product.get(params.id)
@@ -820,7 +820,7 @@ class ProductController {
                 productGroup = new ProductGroup(name: params.productGroup, category: product.category)
             }
             product.addToProductGroups(productGroup)
-            product.save(failOnError: true)
+            productDataService.save(product)
         }
         render(template: 'productGroups', model: [productInstance: product])
     }
@@ -846,7 +846,6 @@ class ProductController {
     /**
      * Delete product group from database
      */
-    @Transactional
     def removeFromProductGroups() {
         println "removeFromProductGroup() " + params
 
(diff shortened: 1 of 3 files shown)


grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
```diff
@@ -810,6 +810,7 @@ class ProductController {
      *
      * @return
      */
+    @Transactional
```
alannadolny, line 813: Can we stop spreading `@Transactional` annotations over a lot of methods in controllers? We can use data services in Grails 3, which are not hard to create and they are the correct approach.

drodzewicz: Yes I know, I decided to just add the annotation because the rest of the ProductController is filled with similar methods with `@Transactional` so I went for a shortcut.
But if you feel strongly that no matter what we should not do it, then sure I can add a service method.

alannadolny: And now I can say "good job" honestly :clap:

grails-app/services/org/pih/warehouse/product/ProductGroupDataService.groovy
```diff
@@ -10,4 +10,6 @@ interface ProductGroupDataService {
     ProductGroup save(ProductGroup productGroup)
 
     ProductGroup get(String id)
+
+
```
kchelstowski, line 14: some spaces have creeped in here

grails-app/services/org/pih/warehouse/product/ProductDataService.groovy
```diff
@@ -0,0 +1,14 @@
...
+
+@Service(Product)
+interface ProductDataService {
+
+    void delete(String id)
+
+    Product save(Product product)
+
+    Product get(String id)
+
```
kchelstowski, line 13: nitpicky: there shouldn't this space before the `}`

kchelstowski (approved):

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/product/ProductController.groovy
jmiranda: While I agree with this in principle, I disagree with it in this case (this project) because we discussed this exact issue numerous times over the past 6 months. We have a ticket (OBGM-143) to deal with moving business logic from controllers to services and we discussed and agreed that most of these changes would be done post-migration because it added too much complexity during rebases.

Since we're done with rebasing, it's probably ok that you decided to do this refactoring. However, this particular PR still doesn't actually make the code any better, it just moves the transaction boundary from "around the controller action" to "inside the controller action". 

Instead, to make the code better we probably want to move all of the logic to a service method. But there's some complexity around that because we still need to figure out coding conventions like whether we want to lookup the Product and Product Group in the controller and pass those objects to the service or make it more straightforward by passing IDs as arguments and doing the lookups and persistence in the service.

Anyway, we can discuss the proper solution when we get to OBGM-143.

alannadolny: I agree that we discussed this issue numerous times, but don't you think it would be nice to move some of the "transactionals" when we have an opportunity to work on such cases? It should be a nice beginning of the future refactor and probably can help us avoid many bugs in the future because we won't rely on the fact that the method/controller is transactional. Because it is only ~10 minutes of work to create a simple data service with save/delete/get (additionally we can avoid lazy initialization exceptions thanks to that). Isn't that beautiful? :smile:

PR #4404, opened 2023-12-06
OBPIH-5990 Able to delete POSM when at least pending invoice exists

alannadolny:
I forgot to add the validation in the stock movement API controller (I added it only in the stock movement controller)

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -32,8 +32,6 @@ import org.pih.warehouse.requisition.RequisitionType
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentStatusCode
 
-import javax.xml.bind.ValidationException
-
 class StockMovementApiController {
 
     def dataService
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -32,6 +32,8 @@ import org.pih.warehouse.requisition.RequisitionType
 import org.pih.warehouse.shipping.Shipment
 import org.pih.warehouse.shipping.ShipmentStatusCode
 
+import javax.xml.bind.ValidationException
```
drodzewicz, line 35: I don't see this class being used anywhere in the changes, is it necessary?

alannadolny: Uh, I forgot to remove it 🫢

kchelstowski (approved):

drodzewicz (approved):

kchelstowski:
@drodzewicz I agree with you, but the problem is that we wouldn't be able to use `ValidationException` because it would expect us to pass the `errors`, and we don't have any at that state.
And I am also not sure if the `IllegalArgumentException` would fit well? But then, it would not have the `400` status, but `500`.

awalkowiak (approved):

PR #4405, opened 2023-12-06
OBGM-47 Fix small issues after refactor of Excel importers

drodzewicz:


grails-app/services/org/pih/warehouse/importer/LocationImportDataService.groovy
```diff
@@ -189,7 +189,10 @@ class LocationImportDataService implements ImportDataService {
         if (!address) {
             address = new Address()
         }
+
         address.properties = params
+        address.address = params.streetAddress ?: ''
+        address.address2 = params.streetAddress2 ?: ''
```
kchelstowski, line 195: that's weird....  I mean - the "frontend" should send proper params, not we should care what frontend (in this case the file) sends and parse it weirdly on the backend.
But I guess it was like that before and it is due to the fact that we parse the header from the excel to the param, and for use it would not be understood what `address2` means?

kchelstowski (approved):

awalkowiak (approved):

PR #4406, opened 2023-12-08
OBGM-750 Fix validation on InvoiceItem quantity when invoiceItem does not have a shipmentItem associated

drodzewicz:
when performing the validation on invoiceItem without shipmentItem attached
```groovy
Boolean isValid = quantity + (shipmentItem?.quantityInvoiced - originalQuantityInvoiced) <= shipmentItem?.quantity
```
Above code would throw an exception since we are trying to subtract a number from `null` (where null is shipmentItem?.quantityInvoiced)

### DIFF
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
@@ -86,13 +86,13 @@ class InvoiceItem implements Serializable {
             // If the invoice is a prepayment or the item is an order adjustment,
             // the validation below doesn't make sense, because
             // we do not have shipmentItem at this point.
-            if (obj.invoice.isPrepaymentInvoice || obj.orderAdjustment) {
+            if (obj.invoice?.isPrepaymentInvoice || obj.orderAdjustment) {
                 return true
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -86,13 +86,13 @@ class InvoiceItem implements Serializable {
             // If the invoice is a prepayment or the item is an order adjustment,
             // the validation below doesn't make sense, because
             // we do not have shipmentItem at this point.
-            if (obj.invoice.isPrepaymentInvoice || obj.orderAdjustment) {
+            if (obj.invoice?.isPrepaymentInvoice || obj.orderAdjustment) {
```
kchelstowski, line 89: is it possible for the InvoiceItem not to have the invoice?

kchelstowski (approved):

jmiranda (commented):
I don't have an opinion on the change, but I'm feeling very uncomfortable about this feature and want to discuss our strategy with this one on the next tech huddle.

awalkowiak (approved):

PR #4407, opened 2023-12-08
OBPIH-6008 Fix data binding in the ProductAssociation import that cau…

kchelstowski:
…sed the stale data being created in product availability table

Marking as do not merge, to determine what branch it should target (I assume hotfix.XXX?)


jmiranda (approved):
I'm ok with this change but just wanted to make sure there are no other properties being copied. If there are we should consider clearing them out before/after cloning. Even if they don't seem significant.

In addition, I want to make sure we add automated tests for this code as soon as possible.

kchelstowski:
This one is ready to be merged - I've provided changes to binding after the discussion from yesterday.

jmiranda (commented):
LGTM

PR #4408, opened 2023-12-08
OBGM-137 Remove checking if !hasExecutedAnyPreviousChangesets before …

awalkowiak:
…running install changelogs

### DIFF
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -14,7 +14,6 @@ import grails.util.Holders
 import liquibase.Contexts
 import liquibase.LabelExpression
 import liquibase.Liquibase
-import liquibase.changelog.ChangeSet
 import liquibase.changelog.DatabaseChangeLog
 import liquibase.database.DatabaseFactory
 import liquibase.database.jvm.JdbcConnection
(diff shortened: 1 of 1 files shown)


grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -664,20 +663,18 @@ class BootStrap {
...
-            // Check if the executed changelog versions include one of the previous versions
-            // and if so, then we need to keep running the old updates to catch up to 0.9.x
-            boolean hasExecutedAnyPreviousChangesets =
-                executedChangelogVersions.any { previousChangelogVersions.contains(it.version) }
-
-            // FIXME Remove !hasExecutedAnyPreviousChangesets once this goes to production
             //If nothing has been created yet, let's create all new database objects with the install scripts
-            List<ChangeSet> hasExecutedAnyChangesets = database.getRanChangeSetList()
-            if (!hasExecutedAnyChangesets || !hasExecutedAnyPreviousChangesets) {
+            if (!executedChangelogVersions) {
```
jmiranda, line 667: At first glance this looks very bad and not what we discussed. `hasExecutedAnyChangesets` is not the same as `executedChangelogVersions` and can't be used interchangably. 

Could you put the code back the way it was, and just remove the `|| !hasExecutedAnyPreviousChangesets` as discussed.
```
List<ChangeSet> hasExecutedAnyChangesets = database.getRanChangeSetList()
if (!hasExecutedAnyChangesets) { 
    ...
}
```
Thanks!

jmiranda (commented):
@awalkowiak This should not have been merged. Please revert the changes you made and either leave it for me or apply just the change we discussed on the call.

grails-app/init/org/pih/warehouse/BootStrap.groovy
awalkowiak: @jmiranda I did it on purpose. Here's why:
```
def executedChangelogVersions = LiquibaseUtil.getExecutedChangelogVersions()
...
liquibase = new Liquibase('views/drop-all-views.xml', new ClassLoaderResourceAccessor(), database)
...
boolean hasExecutedAnyPreviousChangesets =
    executedChangelogVersions.any { previousChangelogVersions.contains(it.version) }
...
List<ChangeSet> hasExecutedAnyChangesets = database.getRanChangeSetList()
if (!hasExecutedAnyChangesets || !hasExecutedAnyPreviousChangesets) {
```
Previously before checking the `database.getRanChangeSetList()` there was `views/drop-all-views.xml` installed which is also visible in the `database.getRanChangeSetList()` and due to that `hasExecutedAnyChangesets` was **always** `true`. So I could not just remove the `|| !hasExecutedAnyPreviousChangesets`, because the install changelogs would never be installed. Lets discuss it on the tech huddle

PR #4410, opened 2023-12-10
OBGM-752 Supported activities of location type are not assigned by default when create location on grails pages 

alannadolny:
When creating a new location the `locationInstance.supportedActivities` is null.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
@@ -123,7 +123,7 @@ class LocationController {
                     flash.message =
                         g.message(code: 'default.updated.message', args: [warehouse.message(code: 'location.label', default: 'Location'), locationInstance.id])
 
-                    if (locationInstance.supportedActivities.contains(ActivityCode.APPROVE_REQUEST.id)) {
+                    if (locationInstance.supportedActivities?.contains(ActivityCode.APPROVE_REQUEST.id)) {
                         flash.warning = g.message(code: 'location.supportedActivities.noRequestApprovers', args: [g.createLink(controller: "user", action: "list")])
                     }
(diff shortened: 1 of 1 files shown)


drodzewicz (approved):

awalkowiak (approved):

jmiranda:
@alannadolny This explanation doesn't make sense to me. Are you saying that update() action is being called on a newly instantiated but unsaved Location instance? And if so, why? If you're saying a persisted location has a NULL supported activities that is also very surprising as it should be an empty set.

awalkowiak:
@jmiranda 
>  Are you saying that update() action is being called on a newly instantiated but unsaved Location instance? And if so, why?

I think there are a couple of examples where there is an update or edit action used for create, and it was like that for a long time (I think this is a potential tech debt, we should improve when we will have time for it)

alannadolny:
@jmiranda 
> Are you saying that update() action is being called on a newly instantiated but unsaved Location instance
yes, we don't have create() for location, so we are calling update() action for saving

> If you're saying a persisted location has a NULL supported activities that is also very surprising as it should be an empty set.
It has null supported activities when it is not persisted - after persisting, it is a set.

jmiranda:
@awalkowiak @alannadolny can you point me to the code that calls update action when there is no location?

In the meantime, please do not merge this PR.

alannadolny:
@jmiranda location/edit.gsp, there is `<g:form method="post" action="update">`, each button inside this tag calls the update.

This PR was merged yesterday.

jmiranda:
Oh right. 

The thing I was worried about was that I wasn't seeing a new Location() anywhere, but it was due to the fact that we were doing that in the getLocation() service method.
```
def locationInstance = inventoryService.getLocation(params.id)
```

PR #4414, opened 2023-12-12
OBPIH-5977 Display items in PO Template in the same order as entered during PO creation

alannadolny:
Katarzyna asked me to add a sorting for order items also. Previously I added it only to order adjustments.

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
@@ -133,7 +133,7 @@ class DocumentTemplateService {
 
         def orderItemsGroupedByProductAndUom = orderInstance?.orderItems?.findAll {
             it.orderItemStatusCode != OrderItemStatusCode.CANCELED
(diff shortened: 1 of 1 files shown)


kchelstowski (approved):

grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -133,7 +133,7 @@ class DocumentTemplateService {
 
         def orderItemsGroupedByProductAndUom = orderInstance?.orderItems?.findAll {
             it.orderItemStatusCode != OrderItemStatusCode.CANCELED
-        }?.groupBy { [product: it.product, unitOfMeasure: it.unitOfMeasure] }?.collect { k, v ->
+        }?.sort()?.groupBy { [product: it.product, unitOfMeasure: it.unitOfMeasure] }?.collect { k, v ->
```
awalkowiak, line 136: I think this should not be an issue, but wanted to be cautious. Have you checked this template for a case when the PO has only order adjustments?

alannadolny: yup, it works

awalkowiak (approved):

PR #4416, opened 2023-12-12
OBGM-753 Empty string validation on name and code of organization

drodzewicz:
- added blank GORM validators to prevent from inserting empty string values
- added a consdition to search for exsiting organization by code and name which are not empty strings

### DIFF
--- a/grails-app/domain/org/pih/warehouse/core/Organization.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Organization.groovy
@@ -37,10 +37,10 @@ class Organization extends Party {
     }
 
     static constraints = {
-        code(nullable: false, unique: true,
+        code(nullable: false, blank: false, unique: true,
                 minSize: Holders.grailsApplication.config.openboxes.identifier.organization.minSize,
                 maxSize: Holders.grailsApplication.config.openboxes.identifier.organization.maxSize)
-        name(nullable: false, maxSize: 255)
+        name(nullable: false, blank: false, maxSize: 255)
         description(nullable: true, maxSize: 255)
         defaultLocation(nullable: true)
         active(nullable: false)
--- a/grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
@@ -10,7 +10,7 @@
 package org.pih.warehouse.core
 
 import grails.gorm.transactions.Transactional
-import grails.validation.ValidationException
+import org.apache.commons.lang.StringUtils
 
 @Transactional
 class OrganizationService {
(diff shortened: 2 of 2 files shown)


grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -45,7 +45,18 @@ class OrganizationService {
     }
 
     Organization findOrCreateOrganization(String name, String code, List<RoleType> roleTypes) {
-        Organization organization = Organization.findByCodeOrName(code, name)
+        Organization organization = org.pih.warehouse.core.Organization.find {
```
jmiranda, line 48: This doesn't feel right. I think I'd rather throw an exception if the code or name are empty than change all finders in the system to perform this type of search. Note that what we come up with here might be used throughout the system since there are a lot of places where we want to do a lookup by code or name.

If we clean up the OBNAV data and prevent users from entering blank values into code and name, then that should probably be enough.

jmiranda (changes requested):
As mentioned in the ticket, we just want to do the data clean up for now. We'll save code changes for when we actually come up with the scope and acceptance criteria.

drodzewicz:
Assuming the data in the database is correct and fixed (without organizations that have invalid empty **code**), the issue should not reoccur anymore, the changes in the PR should ensure that it would be impossible to create an organization with an empty **code** or **name** and throw an error if both passed arguments to `findOrCreateOrganization` are empty strings.

grails-app/domain/org/pih/warehouse/core/Organization.groovy
```diff
@@ -37,10 +37,10 @@ class Organization extends Party {
     }
 
     static constraints = {
-        code(nullable: false, unique: true,
+        code(nullable: false, blank: false, unique: true,
                 minSize: Holders.grailsApplication.config.openboxes.identifier.organization.minSize,
                 maxSize: Holders.grailsApplication.config.openboxes.identifier.organization.maxSize)
-        name(nullable: false, maxSize: 255)
+        name(nullable: false, blank: false, maxSize: 255)
```
kchelstowski, line 43: I assume whenever we make such a change, we should always align GORM constraints with DB constraints, hence a migration that adds this constraint should be added.
I'm not sure how the ddl would act in this case, because we do not have anything like "NOT EMPTY" in SQL and should probably do something like:
```sql
CHECK (name <> '')
```

drodzewicz: yep but we should keep in mind that CHECK is not available for `MySQL 5.7` and need to make sure that we are going to be using either version 8.^ or MariaDB

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -45,6 +45,9 @@ class OrganizationService {
     }
 
     Organization findOrCreateOrganization(String name, String code, List<RoleType> roleTypes) {
+        if (!name.trim() && !code.trim()) {
+            throw new IllegalArgumentException("Can't perform a search on Organization with both code and name arguments being an empty strings")
```
jmiranda, line 49: Change message to "Must provide non-empty code and name"

jmiranda (approved):

drodzewicz:
I pushed all of the changes that were discussed on the tech-huddle, so this PR can be finally reviewed
cc @awalkowiak @jmiranda @alannadolny @kchelstowski

awalkowiak (approved):

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -44,8 +44,25 @@ class OrganizationService {
...
+            eq("code", code)
+            ne("code", "")
+            isNotNull("code")
+        }
+        if (!organization) {
+            organization = Organization.createCriteria().list(max: 1) {
+                eq("name", name)
+                ne("name", "")
+                isNotNull("name")
+                order("dateCreated", "asc")
```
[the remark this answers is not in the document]
jmiranda: The current implementation of findOrganization just replicates what already existing with the dynamic query. But we should ask Kelsey and Manon whether this is appopriate since it basically says "return a record if we match on code or name" but the name could match multiple records that aren't at all related to the code that was passed in. Do we want to constrain the results by the unique code as well and create new records if that doesn't match.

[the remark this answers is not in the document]
jmiranda: This discussion was moved to ticket https://pihemr.atlassian.net/browse/OBGM-753

jmiranda (approved):

PR #4418, opened 2023-12-13
OBGM-752 Supported activities of location type are not assigned by default when create location on grails pages 

alannadolny:
Katarzyna asked me to check the issue with saving an address within this ticket

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
@@ -95,12 +95,6 @@ class LocationController {
                 params.supportedActivities = supportedActivities?.empty ? [ActivityCode.NONE.id] : supportedActivities
             }
 
-            // when we are passing address with id which is empty string or null
-            // then grails try to find an address with this id instead of creating a new one
-            if (!params.address.id) {
-                params.address.remove("id")
-            }
-
             locationInstance.properties = params
 
             if (!locationInstance.id && !locationInstance.organization) {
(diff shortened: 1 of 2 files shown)


grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
```diff
@@ -95,6 +95,12 @@ class LocationController {
                 params.supportedActivities = supportedActivities?.empty ? [ActivityCode.NONE.id] : supportedActivities
             }
 
+            // when we are passing address with id which is empty string or null
+            // then grails try to find an address with this id instead of creating a new one
+            if (!params.address.id) {
+                params.address.remove("id")
```
jmiranda, line 101: This is fine, but you could also make the inclusion of the parameter conditional.
```
<g:if test="${locationInstance?.address}">
    <g:hiddenField name="address.id" value="${locationInstance?.address?.id}"/>
</g:if>
```
I suspect adding this check to the controller is better because it hypothetically defends against more cases. But I really don't want to be doing stuff like this in all of our controllers.

So we should add a tech debt ticket to enable convertEmptyStringsToNull in application.yml.
```
grails.databinding.convertEmptyStringsToNull = true
```

jmiranda (commented):
I'm leaving this as a comment, not a change request because I don't know the right approach yet.

grails-app/controllers/org/pih/warehouse/core/LocationController.groovy
alannadolny: I tried to assign `null` to the address.id manually and it wasn't bound also, so I am unsure if changing this property will help, but I will take a look and let you know. I think I will go with this approach with the condition in the controller.

awalkowiak (approved):

PR #4420, opened 2023-12-14
OBGM-514 Make the quartz monitor render properly + add security rules 

kchelstowski:


grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
```diff
@@ -140,9 +141,12 @@ class RoleInterceptor {
...
     static Boolean needSuperuser(controllerName, actionName) {
-        superuserControllers?.contains(controllerName) || superuserActions[controllerName]?.contains(actionName) || superuserActions['*'].any {
-            actionName?.startsWith(it)
-        }
+        (superuserActions[controllerName]?.contains("*")
+            || superuserControllers?.contains(controllerName)
+            || superuserActions[controllerName]?.contains(actionName)
+            || superuserActions['*'].any {
+                actionName?.startsWith(it)
+            })
```
alannadolny, line 149: hmmm, do we have a styling rule about the placing of the logical operators?
I mean whether we should place them in the way you did, or:
```suggestion
        (superuserActions[controllerName]?.contains("*") ||
            superuserControllers?.contains(controllerName) ||
            superuserActions[controllerName]?.contains(actionName) ||
            superuserActions['*'].any {
                actionName?.startsWith(it)
            })
```

alannadolny (approved):
Great finding on the implementation -> compile!

grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
kchelstowski: According to Java coding conventions, a break should happen **before** the operator: https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html

I think I've raised this topic recently in one of the PRs, that we should discuss it and document it in our coding conventions.

drodzewicz, line 149:
```suggestion
            || superuserActions['*'].any(actionName?.startsWith(it))
            )
```
I think we can make it a little more compact without having to use `{ ... }` but it doesn't really matter I am fine with the way it looks.

drodzewicz (approved):

grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
kchelstowski: I agree, I have not realized that, because this code has been already there, the only thing I've introduced here is:
```groovy
superuserActions[controllerName]?.contains("*")
```
I could change it

kchelstowski: @drodzewicz I've applied your suggestion, but in fact, it should not be `any(...)`, but `any{...}`.

kchelstowski: by saying so I'm starting to to strongly agree with @awalkowiak that applying suggestions blindly via github sucks.

jmiranda (approved):
LGTM

PR #4421, opened 2023-12-14
OBPIH-5980 Data display error in Delivery Note when Pack level 1 is entered

alannadolny:


alannadolny:
I added a `do not merge` label to avoid merging this PR now, because it has to wait for 0.9.1

grails-app/views/deliveryNote/_printPage.gsp
```diff
@@ -65,7 +65,7 @@
                                     </g:each>
                                 </td>
                             </g:if>
-                            <g:if test="${requisitionItems.find { it.requisition?.shipment?.shipmentItems?.any { it.container }}}">
+                            <g:if test="${requisitionItems.find { it.requisition?.shipment?.shipmentItems?.any { it.container && it.container?.parentContainer }}}">
```
kchelstowski, line 68: wouldn't just:
```groovy
it.container?.parentContainer
```
work?

grails-app/views/deliveryNote/_printPage.gsp
```diff
@@ -65,7 +65,7 @@
                                     </g:each>
                                 </td>
                             </g:if>
-                            <g:if test="${requisitionItems.find { it.requisition?.shipment?.shipmentItems?.any { it.container }}}">
+                            <g:if test="${requisitionItems.find { it.requisition?.shipment?.shipmentItems?.any { it.container?.parentContainer }}}">
```
jmiranda, line 68: From what I can tell from the revision history, the code has gone from something like this
```
<g:if test="${shipmentItems.any { it.container }}">
```
to this 
```
<g:if test="${shipmentItems.any { it.container && it.container?.parentContainer }}">
```
and now to this 
```                            
<g:if test="${requisitionItems.find { it.requisition?.shipment?.shipmentItems?.any { it.container }}}">
```

I have no idea what this code is trying to achieve, so I feel like now would be a good time to move this logic to the domain. I just don't know exactly which object to interogate to get this information. Shipment? Requisition?

In either case, it should probably live outside of whatever loop we're iterating over in case the method is expensive (i.e. requires database integration). 
```
whateverObjectWeShouldInterogate.hasParentContainer()
whateverObjectWeShouldInterogate.hasChildContrainer()
```

jmiranda (changes requested):

awalkowiak (approved):

PR #4423, opened 2023-12-15
OBPIH-5981 Improve redirect for e-request edit action for a requestor…

kchelstowski:
… in a ward location (request approval feature) + refactor redirect logic for SM edit action

Since it's targeting the 0.9.1, and I had to touch this code, I decided to refactor this part to improve the readability of the redirect logic which was like this:
```groovy
if(stockMovement.isReturn) {
    redirect(controller: "stockTransfer", action: "edit", params: params)
} else if (stockMovement?.getStockMovementDirection(currentLocation) == StockMovementDirection.OUTBOUND && stockMovement?.requisition?.sourceType == RequisitionSourceType.ELECTRONIC) {
    redirect(action: "verifyRequest", params: params)
}
else if (stockMovement?.getStockMovementDirection(currentLocation) == StockMovementDirection.INBOUND) {

    if (stockMovement.isFromOrder) {
        redirect(action: "createCombinedShipments", params: params)
    } else if (stockMovement.requisition?.sourceType == RequisitionSourceType.ELECTRONIC) {
        if (stockMovement.requisition?.status == RequisitionStatus.CREATED) {
            redirect(action: "createRequest", params: params)
        } else {
            redirect(action: "verifyRequest", params: params)
        }
    } else {
        redirect(action: "createInbound", params: params)
    }
}
else {
    if (stockMovement.isFromOrder) {
        redirect(action: "createCombinedShipments", params: params)
    }
    redirect(action: "createOutbound", params: params)
}
```
and after my changes (I paste it since the changes in the github might be difficult to read):
```groovy
if (stockMovement.isReturn) {
    redirect(controller: "stockTransfer", action: "edit", params: params)
    return
}

StockMovementDirection stockMovementDirection = stockMovement?.getStockMovementDirection(currentLocation)

// Redirect logic for the e-request
if (stockMovement?.electronicType) {
    switch (stockMovementDirection) {
        case StockMovementDirection.OUTBOUND:
            redirect(action: "verifyRequest", params: params)
            break
        case StockMovementDirection.INBOUND:
            User currentUser = AuthService.currentUser
            RequisitionStatus requisitionStatus = stockMovement.requisition?.status
            // If a request's status is created or a user is a requestor logged in, in a location
            // with submit request and without manage inventory, and the request has waiting for approval
            // status, we want to redirect to add items page (OBPIH-5981)
            if (requisitionStatus == RequisitionStatus.CREATED
                || (userService.isUserRequestor(currentUser)
                    && currentLocation.downstreamConsumer
                    && requisitionStatus == RequisitionStatus.PENDING_APPROVAL)) {
                        redirect(action: "createRequest", params: params)
                        return
            }
            redirect(action: "verifyRequest", params: params)
            break
        default:
            break
    }
    return
}

// Redirect logic for the combined shipments feature
if (stockMovement.isFromOrder) {
    redirect(action: "createCombinedShipments", params: params)
    return
}

// Redirect logic for the regular stock movement feature
switch (stockMovementDirection) {
    case StockMovementDirection.OUTBOUND:
        redirect(action: "createOutbound", params: params)
        break
    case StockMovementDirection.INBOUND:
        redirect(action: "createInbound", params: params)
        break
    default:
        break
}
```
the part that is the scope of the ticket is this:
```groovy
if (requisitionStatus == RequisitionStatus.CREATED
      || (userService.isUserRequestor(currentUser)
          && currentLocation.downstreamConsumer
          && requisitionStatus == RequisitionStatus.PENDING_APPROVAL)) {
              redirect(action: "createRequest", params: params)
              return
  }
```


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -83,36 +86,62 @@ class StockMovementController {
...
+                                redirect(action: "createRequest", params: params)
+                                return
+                    }
+                    redirect(action: "verifyRequest", params: params)
+                    break
+                default:
+                    break
             }
+            return
         }
```
drodzewicz, line 127: something feels off here, I feel like this should have some default redirect

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -83,36 +86,62 @@ class StockMovementController {
...
+        // Redirect logic for the regular stock movement feature
+        switch (stockMovementDirection) {
+            case StockMovementDirection.OUTBOUND:
+                redirect(action: "createOutbound", params: params)
+                break
+            case StockMovementDirection.INBOUND:
+                redirect(action: "createInbound", params: params)
+                break
+            default:
+                break
```
drodzewicz, line 144: Same here, I think a default redirect should be present here.
`getStockMovementDirection` can return a `null` value and there is no redirect handler at the end, so I think the app would crash because no redirects are specified at the end.

drodzewicz (changes requested):
Thanks for doing the refactor, I had to dos some mental gymnastics to compare the old and the new versions of this redirect logic and it wasn't easy.

In general I think it looks good and does cover all of the old cases, but please take a look at the empty default cases, I think there should be something in case we land in that default case, as unlikely as it may seem.

drodzewicz:
@kchelstowski looks good, let's proceed with that

jmiranda (commented):
My first thought when seeing this was "rendering a view should not be this complicated". 

Can we reduce the complexity by creating separate URLs for the different types of requests / transfers and not using the stock movement controller as a central hub? If not, is there another way to reduce the complexity.

kchelstowski:
@jmiranda since for "Edit stock movement" our url is `/stockMovement/edit` and it's currently handled in the method I've refactored, to do it like you suggested, we would actually have to add a similar in terms of complexity logic in the `show.gsp` view of the stock movement, to determine what url the "Edit stock movement" should have.

Since the current else-if logic drove us nuts and I had to add one more case there, I decided to refactor it a bit, so it becomes more readable. 
Let me know if it's fine for now or if we'd want to have a different approach there.

awalkowiak (commented):
I am very "pro" these changes. I think this part was getting worse and worse.

@kchelstowski lets finalize the refactoring that you started, if there are any improvements left from Dariusz's review. And please make **double check locally** on every branch we had previously, that it is still working correctly.

@jmiranda I think that the issue here is that we rather wanted to keep various workflows under one stock movement. And it kept growing and expanding. I think ideally we could think about the separation of the pages or other type of the refactor, but there are to many of "what ifs". For now I would go with this cleanup for 0.9.1, since it is already done, and I think looks good. We can come back to this chunk of code, as tech debt and discuss on the tech huddle. I feel we might run into few use cases from past, where we get some answers why its like that. We need to dig it out a bit from tickets and perhaps test/use cases.

drodzewicz (approved):

awalkowiak (approved):

PR #4424, opened 2023-12-18
OBPIH-5840 Create a Dashboard indicator for number of requests waiting for approval

drodzewicz:


drodzewicz:
Pushing the first part of this ticket which can be reviewed, but waiting for feedback on questions left in the ticket comments.

grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
```diff
@@ -196,6 +198,12 @@ class DashboardApiController {
         render (numberData as JSON)
     }
 
+    def getRequestsPendingApproval() {
+        Location location = Location.get(params.locationId)
+        NumberData numberData = numberDataService.getRequestsPendingApproval(location, authService.currentUser)
+        render (numberData as JSON)
```
alannadolny, line 204:
```suggestion
        render(numberData as JSON)
```

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
```diff
@@ -210,6 +211,28 @@ class NumberDataService {
...
+            SELECT COUNT(distinct r.id) FROM Requisition r
+            WHERE r.origin = :location
+            AND r.sourceType = :sourceType
+            AND r.status  = :status
+            """,
+                [
+                        'location': location,
+                        'sourceType' : RequisitionSourceType.ELECTRONIC,
+                        'status' : RequisitionStatus.PENDING_APPROVAL,
+                ])
```
alannadolny, line 226: Wouldn't it be better to use createCritera here?

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
```diff
@@ -210,6 +211,28 @@ class NumberDataService {
...
+
+        String urlContextPath = ConfigHelper.contextPath
+        String redirectLink = currentUser.hasRoles(location, [RoleType.ROLE_REQUISITION_APPROVER])
+                ? "${urlContextPath}/stockMovement/list?direction=OUTBOUND&requisitionStatusCode=PENDING_APPROVAL&sourceType=ELECTRONIC"
+                : "${urlContextPath}/stockMovement/list?direction=OUTBOUND&sourceType=ELECTRONIC"
```
alannadolny, line 231: It looks odd to me, but I don't have any idea how to improve that, maybe someone else can come up with a better solution?

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
kchelstowski: I would go even further and give the data service method a shot (as I can see we already have the `RequisitionDataService`, so this would require implementing only a method)
```groovy
int countByOriginAndSourceTypeAndStatus(Location origin, RequisitionSourceType sourceType, RequisitionStatus status)
```

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
kchelstowski: I agree, imho we should also think about creating a url factory (similar to the frontend one) as a tech debt ticket.
In here, at the very least, I'd suggest using consts for the param values:
```groovy
?direction=StockMovementDirection.OUTBOUND&requisitionStatusCode=RequisitionStatus.PENDING_APPROVAL&sourceType=RequisitionSourceType.ELECTRONIC
```

drodzewicz: instead I decided to use the `createLink(...)` method which is provided by `ApplicationTagLib`.
That way we don;t have manually prepend the contextPath and we can use the enums and other values in the parameters which later builds the propper string link just like we want to.

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
drodzewicz: Previously I was just following the example of other NumberService methods which compute data with pure SQL statements, but you are right something like a `createCriteria()` looks a bit more friendly so I went with that approach

drodzewicz: I also tested your suggestion @kchelstowski with `countByOriginAndSourceTypeAndStatus()` but I realized that we need to cache this indicator and the keys are `location.id` and `currentUser.id` so either way i would need to call this `RequisitionDataService` method in a `NumberDataService` so it creates a little more clutter and methods that we have to maintain so I decided to go with `createCriteria()`

drodzewicz: Although now that I think about it, I remember you mentioning something that GORM Data Service methods are much faster than the standard criteria ones, so this might be a reason to use the DataService instead?

kchelstowski: @drodzewicz could you elaborate a bit on the fact that you couldn't use the data service method here? I don't mind the `createCriteria()` approach, but wanted to understand why the data service method didn't fit there?

kchelstowski: > Although now that I think about it, I remember you mentioning something that GORM Data Service methods are much faster than the standard criteria ones, so this might be a reason to use the DataService instead?

I think I've never said they are faster - it should probably evaluate to the same SQL query as the data service method, but imho they are just more readable and they don't require that much of "boilerplate" code - e.g. the `createCriteria` approach here took more lines than one-liner data service method.

jmiranda (approved):

awalkowiak (approved):

PR #4428, opened 2023-12-20
OBPIH-5774 Do not show inactive suppliers in the supplier select whil…

kchelstowski:
…e creating a PO


drodzewicz (approved):

alannadolny (approved):

jmiranda (approved):
I think we need a tech debt ticket where we identify other places in the code where we are returning inactive objects. Similarly, it would be interesting to see if there's a Hibernate mechanism that would allow us to handle inactive and soft-delete in a more central way.

For example,
https://www.baeldung.com/spring-jpa-soft-delete

I created a ticket for the tech debt 
https://pihemr.atlassian.net/browse/OBPIH-6051

awalkowiak (approved):

PR #4429, opened 2023-12-20
OBPIH-5826 comment indication approvals comment

drodzewicz:
follow the discussion in [the ticket](https://pihemr.atlassian.net/browse/OBPIH-5826)

* [x] add derived property currentEvent to stock movement 
* [x] use currentEvent to get the comment for a approval status if (status in approvalStatus) display currentEvent.comment
* [x] remove requisitonComments and recentRequisitionComments

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
@@ -888,13 +888,13 @@ class ShipmentController {
 
 
     def saveComment() {
-        def shipmentInstance = Shipment.get(params.shipmentId)
-        def recipient = (params.recipientId) ? User.get(params.recipientId) : null
-        def comment = new Comment(comment: params.comment, sender: session.user, recipient: recipient)
-        if (shipmentInstance) {
-            shipmentInstance.addToComments(comment).save(flush: true)
-            flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
+        Shipment shipmentInstance = Shipment.get(params.shipmentId)
+        if (!shipmentInstance) {
+            throw new Exception("Could not locate shipment with ID " + params.shipmentId)
         }
+        User recipient = (params.recipientId) ? User.get(params.recipientId) : null
+        shipmentService.addShipmentComment(shipmentInstance, params.comment,  session.user, recipient)
+        flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
         redirect(action: 'showDetails', id: params.shipmentId)
     }
 
--- a/grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
@@ -888,13 +888,9 @@ class ShipmentController {
 
 
     def saveComment() {
-        Shipment shipmentInstance = Shipment.get(params.shipmentId)
-        if (!shipmentInstance) {
-            throw new Exception("Could not locate shipment with ID " + params.shipmentId)
-        }
-        User recipient = (params.recipientId) ? User.get(params.recipientId) : null
-        shipmentService.addShipmentComment(shipmentInstance, params.comment,  session.user, recipient)
-        flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
+        User recipient = User.get(params.recipientId)
+        shipmentService.addShipmentComment(params.shipmentId, params.comment, session.user, recipient)
+        flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, params.shipmentId])}"
         redirect(action: 'showDetails', id: params.shipmentId)
     }
 
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -18,6 +18,7 @@ import org.apache.poi.hssf.usermodel.HSSFWorkbook
 import org.apache.poi.ss.usermodel.Cell
 import org.apache.poi.ss.usermodel.Row
 import org.hibernate.FetchMode
+import org.hibernate.ObjectNotFoundException
 import org.pih.warehouse.api.StockTransfer
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Comment
@@ -2377,10 +2378,14 @@ class ShipmentService {
         return referenceNumber
     }
 
-    Comment addShipmentComment(Shipment shipment, String content, User sender, User recipient) {
+    Comment addShipmentComment(String shipmentId, String content, User sender, User recipient) {
+        Shipment shipmentInstance = Shipment.get(shipmentId)
+        if (!shipmentInstance) {
+            throw new ObjectNotFoundException("Could not locate shipment with ID " + shipmentId)
+        }
         Comment comment = new Comment(comment: content, sender: sender, recipient: recipient)
         comment.save()
-        shipment.addToComments(comment)
+        shipmentInstance.addToComments(comment)
         return comment
     }
 }
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -2376,4 +2376,11 @@ class ShipmentService {
         }
         return referenceNumber
     }
+
+    Comment addShipmentComment(Shipment shipment, String content, User sender, User recipient) {
+        Comment comment = new Comment(comment: content, sender: sender, recipient: recipient)
+        comment.save()
+        shipment.addToComments(comment)
+        return comment
+    }
 }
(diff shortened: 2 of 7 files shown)


grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
```diff
@@ -888,13 +888,13 @@ class ShipmentController {
...
-        def recipient = (params.recipientId) ? User.get(params.recipientId) : null
-        def comment = new Comment(comment: params.comment, sender: session.user, recipient: recipient)
-        if (shipmentInstance) {
-            shipmentInstance.addToComments(comment).save(flush: true)
-            flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
+        Shipment shipmentInstance = Shipment.get(params.shipmentId)
+        if (!shipmentInstance) {
+            throw new Exception("Could not locate shipment with ID " + params.shipmentId)
         }
+        User recipient = (params.recipientId) ? User.get(params.recipientId) : null
```
kchelstowski, line 895: I know it's been there and you might've not realized that, but I think `User.get(params.recipientId)` should be enough.... unless there is a possibility that user's id would be `null`..

grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -2376,4 +2376,11 @@ class ShipmentService {
         }
         return referenceNumber
     }
+
+    Comment addShipmentComment(Shipment shipment, String content, User sender, User recipient) {
```
kchelstowski, line 2380: I don't mind it, but probably according to our code conventions, we shouldn't have more than 3 arguments (?) :thinking: 
I believe we need to figure out a way how to deal with such cases (command object??)

src/js/components/dashboard/StockRequestDashboard.jsx
```diff
@@ -31,6 +33,24 @@ const COLUMNS = [
     headerClassName: 'text-left font-weight-bold px-4 py-3',
     className: 'px-4 py-2',
     maxWidth: 200,
+    Cell: (row) => (
+      <span className="d-flex align-items-center">
+        <span className="d-inline-block text-overflow-ellipsis">
+          {row.original.displayStatus}
+        </span>
+        {row.original.recentComment && (
+        <Tooltip
```
kchelstowski, line 42: nitpicky: indent

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -179,6 +184,7 @@ class StockMovement implements Validateable{
             trackingNumber      : trackingNumber,
             driverName          : driverName,
             comments            : comments,
+            recentComment       : recentComment,
```
kchelstowski, line 187: I agree with you something doesn't feel right about that - I mean, we only care about the `shipmentComments` - what if we decided to care about the requisition comments also?
Would we then want to distinguish shipment/requisition recent comment or how would we want to approach that?
I don't know the scope much to be able to answer that question.

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -205,6 +211,15 @@ class StockMovement implements Validateable{
         ]
     }
 
+    Comment getRecentComment() {
+        if (shipmentComments?.size() > 0) {
+            return shipmentComments?.sort({ a, b ->
+                b.dateCreated <=> a.dateCreated
```
kchelstowski, line 217: shouldn't we look at `lastUpdated` instead of `dateCreated`? Maybe I don't know the scope very well, but I see that the `compareTo` method in `Comment` looks at the `lastUpdated` date.

If we do want to look at `lastUpdated`, a titbit: you could just write `.sort()`, since we have `compareTo` overwritten.

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
```diff
@@ -888,13 +888,13 @@ class ShipmentController {
...
     def saveComment() {
-        def shipmentInstance = Shipment.get(params.shipmentId)
-        def recipient = (params.recipientId) ? User.get(params.recipientId) : null
-        def comment = new Comment(comment: params.comment, sender: session.user, recipient: recipient)
-        if (shipmentInstance) {
-            shipmentInstance.addToComments(comment).save(flush: true)
-            flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
+        Shipment shipmentInstance = Shipment.get(params.shipmentId)
+        if (!shipmentInstance) {
+            throw new Exception("Could not locate shipment with ID " + params.shipmentId)
```
kchelstowski, line 893: maybe let's come up with a more specific exception? using the general `Exception` doesn't feel right to me.
Maybe we should start creating our own exceptions and define them? e.g. `NotFoundException`?

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
```diff
@@ -888,13 +888,13 @@ class ShipmentController {
...
         }
+        User recipient = (params.recipientId) ? User.get(params.recipientId) : null
+        shipmentService.addShipmentComment(shipmentInstance, params.comment,  session.user, recipient)
+        flash.message = "${warehouse.message(code: 'shipping.addedCommentToShipment.message', args: [params.comment, shipmentInstance.name])}"
```
kchelstowski, line 897: the problem with that is since the `ShipmentController` is `@Transactional`, you start and end the transaction boundary in the controller, hence if anything goes wrong while persisting the comment, you will still get the success message in the view, because the exception will happen at the end of the session, not while executing the `896`th line.

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
drodzewicz: yeah, this is up for debate what do we consider "the most recent comment" in my opinion it's based on date created since that's usually how it is sorted in many comment sections and messengers.

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
drodzewicz: yes, the whole comments hierarchy should be reevaluated like, which comments does the `StockMovement` inherit? The Shipment comments? Requisition comments? Order Comments?
Maybe we should mash all of those comments in a common list and extract the most recent one regardless of it's origin?

src/js/components/dashboard/StockRequestDashboard.jsx
drodzewicz: it's weird that ESLint did not throw an error on it, thanks for noticing

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
drodzewicz: I did not pay much attention toit, I'll try to mak sure that recipient is not null and if everything will be fine I'll fix it to be just `User.get(recipientId)`

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
drodzewicz: I believe `org.hibernate.ObjectNotFoundException` could be a good candidate

jmiranda: :+1:

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
jmiranda: I know we may have encountered issues like this in the past, but let's confirm this the actual behavior.

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
jmiranda: Yeah, we need to rewrite the APIs completely. This is a case where we've overloaded StockMovement to do so many things and it's showing signs of collapse.

jmiranda: Can you explain what "recentComment" is for? I know it's probably related to the requisition approval, but I want to know specific use case without digging through the code. I suspect this is a case where the codes needs to be more specific about what we're trying to accomplish i.e. getApprovalComment. And that makes me feel like we need to do something similar to synonyms where we need to classify comments according to their purpose instead of just getting the latest comment.

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -361,6 +376,7 @@ class StockMovement implements Validateable{
                 driverName: shipment.driverName,
                 trackingNumber: trackingNumber?.identifier,
                 comments: shipment.additionalInformation,
+                shipmentComments: shipment.comments,
```
jmiranda, line 379: If possible, I would like this to replace the "comments" property at some point. We can move "comments" to "additionalInformation" or try to figure out how we might allow the additionalInformation to be come an instance within the "comments" list. This is where we'd need to have a way to classify comments as we'd only ever want one of these "additionalInformation" comments to be added to stockMovement.comments.

grails-app/controllers/org/pih/warehouse/shipping/ShipmentController.groovy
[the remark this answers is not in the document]
jmiranda: Note: this comment was written weeks ago before I stumbled upon @kchelstowski's command object proposal. So this can probably be safely ignored if we go the command object route.

awalkowiak (approved):

PR #4431, opened 2023-12-20
OBPIH-6026 Changing quantities to 0 on very long stocklists breaks autosave

alannadolny:


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -38,6 +38,7 @@ import NotificationType from 'consts/notificationTypes';
 import RowSaveStatus from 'consts/rowSaveStatus';
 import apiClient from 'utils/apiClient';
 import { renderFormField } from 'utils/form-utils';
+import requestsQueue from 'utils/requestsQueue';
```
awalkowiak, line 41: I feel like this should be handled by proper lib. Quick googling on my phone, and I land on this suggestion https://github.com/axios/axios/issues/1010#issuecomment-326172188. I think we should investigate it briefly when we are back.

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
alannadolny: I used the TaskQueue proposed in this comment instead of the new Queue implementation

awalkowiak (approved):

PR #4432, opened 2023-12-22
OBPIH-5889 Order of statuses in Status filter of the requests list with approvals supported

alannadolny:


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -827,8 +827,12 @@ class StockMovementApiController {
...
     // Function for getting appropriate filter options based on current list and supporting requests approval
     List<RequisitionStatus> getOutboundRequisitionStatusCodes(Boolean isApprovalRequired, Boolean isElectronicType) {
-        // If a location doesn't have approval required, return listOutboundOptions no matter what list is displayed
+        // If a location doesn't have approval required, return listOutboundOptions when we are on outbound list
+        // but if we are on the requests list return listRequestsOptions
         if (!isApprovalRequired) {
+            if (isElectronicType) {
+                return RequisitionStatus.listRequestOptions()
+            }
             return RequisitionStatus.listOutboundOptions()
```
awalkowiak, line 836: I think this would look better as a one liner:
`return isElectronicType ? RequisitionStatus.listRequestOptions() : RequisitionStatus.listOutboundOptions()`

awalkowiak (changes requested):

kchelstowski (approved):
I was confused for a while (before I jumped into the ticket's comments), because the title of the PR and the ticket is "Order of statuses" and here I can see that you got rid of some options.
It'd be good next time to mention that there was some additional scope added while working on a ticket.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -827,9 +827,10 @@ class StockMovementApiController {
 
     // Function for getting appropriate filter options based on current list and supporting requests approval
     List<RequisitionStatus> getOutboundRequisitionStatusCodes(Boolean isApprovalRequired, Boolean isElectronicType) {
-        // If a location doesn't have approval required, return listOutboundOptions no matter what list is displayed
+        // If a location doesn't have approval required, return listOutboundOptions when we are on outbound list
+        // but if we are on the requests list return listRequestsOptions
         if (!isApprovalRequired) {
-            return RequisitionStatus.listOutboundOptions()
+            return isElectronicType ? RequisitionStatus.listRequestOptions() : RequisitionStatus.listOutboundOptions()
```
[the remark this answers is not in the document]
jmiranda: I think a large part of the problem is that Approval should have been implemented as a separate state machine. This is becoming clear to me because we should be able to apply the approval model to any of our activities: requests, purchase orders, supplier preferences, receipts, etc.

src/main/groovy/org/pih/warehouse/requisition/RequisitionStatus.groovy
```diff
@@ -80,14 +80,19 @@ enum RequisitionStatus {
         [CREATED, EDITING, APPROVED, PICKING, PICKED, CHECKING, ISSUED, CANCELED, PENDING, REQUESTED, DISPATCHED]
     }
 
-    // Options for request list when current location is supporting request approval (Added approved and waiting for approval)
+    // Options for request list when current location is supporting request approval (Added approved, rejected and waiting for approval)
```
jmiranda, line 83: Can we add some additional context for each of these to provide a reason why we are restricting the options? You could potentially find this information in the ticket.

jmiranda (commented):
Just commenting. Besides adding additional context to the RequisitionStatus list methods, this looks fine to me.

awalkowiak (approved):

PR #4436, opened 2023-12-29
OBPIH-6018 Add product sources list table

kchelstowski:
Screenshot from 2023-12-29 16-10-20

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -4264,6 +4264,22 @@ react.productSupplier.header.label=Product Sources List
 react.productSupplier.tabs.details.label=Details
 react.productSupplier.tabs.preferenceTypes.label=Preference Types
 react.productSupplier.createProductSource.label=Create Product Source
+react.productSupplier.listProductSources.label=List Product Sources
+react.productSupplier.error.productSupplierList.label=Unable to fetch product sources
+react.productSupplier.column.productCode.label=Product Code
+react.productSupplier.column.productName.label=Product Name
+react.productSupplier.column.sourceCode.label=Source Code
+react.productSupplier.column.preferenceType.label=Preference Type
+react.productSupplier.column.defaultPackSize.label=Default Pack Size
+react.productSupplier.column.packagePrice.label=Package Price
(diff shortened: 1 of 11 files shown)


grails-app/i18n/messages.properties
```diff
@@ -4264,6 +4264,22 @@ react.productSupplier.header.label=Product Sources List
...
+react.productSupplier.column.sourceCode.label=Source Code
+react.productSupplier.column.preferenceType.label=Preference Type
+react.productSupplier.column.defaultPackSize.label=Default Pack Size
+react.productSupplier.column.packagePrice.label=Package Price
+react.productSupplier.column.eachPrice.label=Each Price
+react.productSupplier.column.supplier.label=Supplier
+react.productSupplier.column.supplierCode.label=Supplier Code
+react.productSupplier.column.supplierProductName.label=Supplier Product Name
+react.productSupplier.column.dateCreated.label=Date Created
+react.productSupplier.column.status.label=Status
```
alannadolny, line 4280: I think the `column` part is not needed.
```
react.productSupplier.productCode.label=Product Code
react.productSupplier.productName.label=Product Name
react.productSupplier.sourceCode.label=Source Code
react.productSupplier.preferenceType.label=Preference Type
react.productSupplier.defaultPackSize.label=Default Pack Size
react.productSupplier.label=Package Price
react.productSupplier.eachPrice.label=Each Price
react.productSupplier.supplier.label=Supplier
react.productSupplier.supplierCode.label=Supplier Code
react.productSupplier.supplierProductName.label=Supplier Product Name
react.productSupplier.dateCreated.label=Date Created
react.productSupplier.status.label=Status
```
To be honest, I am not sure about the `productSupplier` also. Most of those translations can be potentially used in a lot of future redesigns, so I am thinking about having those under something similar like `react.default`.

kchelstowski: For all list pages we used to distinguish what's used for the columns, what's used for the table etc.
We once decided that "the same" labels could have different context in some places - this is why we also don't use the `react.default` for most of them.
I don't remember where, but probably even for some of the new list pages we've had a discussion about that.

PR #4438, opened 2024-01-02
OBPIH-5832 Product Source Redesign

jmiranda:
I'm creating this pull request so we can keep track of change requests from the already merged PRs. 

@awalkowiak This doesn't get merged until I am :100:


src/js/components/productSupplier/ProductSupplierTabs.jsx
drodzewicz, line 1: I think this component should be a bit more generic and receive the tabs configuration as props instead of defining it in the component specific to only Product Sources page.

During planning I foresaw the implementation for this component in a following way

1. Have a Generic Tabs component which we will be able to use on many different pages in the future.
2. Have it receive a tabs configuration
In this example something like
```jsx
const tabs = {
  [DETAILS_TAB]: {
    label: {
      id: 'react.productSupplier.tabs.details.label',
      defaultMessage: 'Details',
    },
  onClick: (key) => switchTab(key)
  render: () => <ComponentsToRenderOnTab1 />
  },
  [PREFERENCE_TYPES_TAB]: {
    label: {
      id: 'react.productSupplier.tabs.preferenceTypes.label',
      defaultMessage: 'Preference Types',
    },
   onClick: (key) => switchTab(key),
  render: () => <ComponentsToRenderOnTab2 />
  },
};
```
3. have it receive and active tab props like example
<Tabs  config={tabs} active={DETAILS_TAB} />
Tabs are something that will also be implemented in PO view page and it would be a shame if we would have to duplicate the code for it.

My preference would also be to use an existing library for tabs like https://www.npmjs.com/package/react-tabs

src/js/components/listPagesUtils/Tabs.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import useQueryParams from 'hooks/useQueryParams';
+import Translate from 'utils/Translate';
+
+const Tabs = ({ config }) => {
```
awalkowiak, line 8: Is the name of it too "generic"? Could we get a better name for it? Or do you feel like it is ok?

src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
```diff
@@ -0,0 +1,82 @@
...
+  ]);
+
+  const openConfirmationModal = (productSupplierId) => {
+    confirmationModal({
+      buttons: deleteConfirmationModalButtons(productSupplierId),
+      ...modalLabels,
+    });
+  };
+
+  const getActions = useCallback((productSupplierId) => (hasPermissionsToProductSourceActions(currentUser, isAdmin) ? [
```
awalkowiak, line 59: I think that this `hasPermissions` should be a "global" / more generic util or hook where you provide required permission and it checks if user has it or not. I don't feel I like writing a separate util or hook for each component/action sense :thinking:

src/js/wrappers/ListFilterFormWrapper.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+const ListFilterFormWrapper = ({ children, className }) => (
```
awalkowiak, line 5: There are 6 wrappers that are basically the same differing with only "default" classes that are added to the className:
1. `d-flex flex-column list-page-filters ${className}`
2. `d-flex justify-content-end buttons align-items-center ${className}`
3. `d-flex list-page-header ${className}`
4. `title-text p-3 d-flex justify-content-between align-items-center ${className}`
5. `list-page-list-section ${className}`
6. `d-flex flex-column list-page-main ${className}`

Do you really need all of them?

src/js/utils/permissionUtils.js
```diff
@@ -41,4 +42,13 @@ const canEditRequest = (currentUser, request, location) => {
   return isLocationDestination && isUserRequestor;
 };
 
+export const hasRole = (user, role) => user?.roles?.includes(role);
+
+// To have permissions for product source list actions
+// user has to be at least admin and has product manager permissions
+export const hasPermissionsToProductSourceActions = (user, isAdmin) => {
```
awalkowiak, line 49: I feel like this one can be more generic and reused for other things (or turned into hook)

src/js/utils/CustomConfirmModal.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+import PropTypes from 'prop-types';
+import { RiCloseFill } from 'react-icons/all';
+
+import Button from 'components/form-elements/Button';
+import Translate from 'utils/Translate';
+
+import 'utils/utils.scss';
+
+const CustomConfirmModal = ({
```
awalkowiak, line 11: What's that "Custom", why it can't be just "ConfirmModal"?

awalkowiak (commented):
I looked through the majority of frontend work, on Monday I'll look more through the backend (plus I'll look around in some react places I have just skimmed)

src/js/api/services/OrganizationApi.js
```diff
@@ -0,0 +1,7 @@
+import { GENERIC_API } from 'api/urls';
+import apiClient from 'utils/apiClient';
+
+export default {
+  // TODO: tech debt: Replace by the product api call instead of generic
```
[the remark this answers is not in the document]
kchelstowski: I think we discussed once and agreed, that we would eventually want to get rid of the generic services/controllers. 
The reason I used them was not to introduce many backend changes when I was working alone.

src/js/components/listPagesUtils/ListTitle.jsx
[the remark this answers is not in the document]
kchelstowski: I agree we should rethink the structure, because I also found it difficult to find a proper place for a few components/utils.

src/js/components/listPagesUtils/Tabs.jsx
kchelstowski: Initially it was named as "ProductSupplierTabs", but I got a suggestion from @drodzewicz, that we could eventually use this component for further redesigns (e.g. PO), hence the name should be more generic.

src/js/wrappers/ListFilterFormWrapper.jsx
kchelstowski: I agree at the moment it could be one, that would just receive className prop, but I think it is more maintainable with this approach, especially if we'd want to add any element in the specific wrapper and if we want to use it also while creating the forms.

src/main/groovy/org/pih/warehouse/product/ProductSupplierListParams.groovy
```diff
@@ -0,0 +1,24 @@
+package org.pih.warehouse.product
+
+import org.pih.warehouse.api.PaginationParams
+
+class ProductSupplierListParams extends PaginationParams {
```
jmiranda, line 5: @jmiranda to think about naming here.

src/js/api/services/OrganizationApi.js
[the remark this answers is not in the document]
jmiranda: Use generic API until we implement Resource annotation for organization.

jmiranda (commented):
Just to review the naming for the new params command class.

src/main/groovy/org/pih/warehouse/product/ProductSupplierListParams.groovy
jmiranda: Ok just change the name to PaginationCommand and ProductSupplierFilterCommand. 

I think pagination should support totalCount as well. 

In fact, for the pagination command, use this implementation as a guide.

https://github.com/bmihovski/grails3asciibooks/blob/64e1b6ee69260ae15af0a19ad409b5733ef9b756/asciibooks/src/main/groovy/com/asciibooks/commands/PaginationCommand.groovy

We probably don't need to handle sort/order now (that probably needs a separate class like SearchCommand anyway).

kchelstowski: @jmiranda thanks for the review. I will checkout the example you've attached. As for the `totalCount`, it would be difficult to include it, because it's not a user that we receive this param from, but it's counted after running the `createCriteria`, so I kinda do not see a use case for that currently.

PR #4439, opened 2024-01-04
OBPIH-6014 Product Sources list filters

kchelstowski:
Added TODOs to eventually replace the organization and product fetch by a call to the proper APIs instead of the generic one when you are back and we agree on an approach.
For now I used the generic api even though I hate it and I believe you also do, but in order for the list to work, it seems to be enough for a while. 

Screenshot from 2024-01-04 11-12-43

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
@@ -83,11 +83,11 @@ class SelectOptionsApiController {
         List<Map<String, String>> defaultPreferenceTypes = [
             [
                 id: ProductSupplierService.PREFERENCE_TYPE_MULTIPLE,
-                label: g.message(code: 'react.productSupplier.preferenceType.multiple.label')
+                label: g.message(code: "react.productSupplier.preferenceType.multiple.label", default: "Multiple")
             ],
             [
                 id: ProductSupplierService.PREFERENCE_TYPE_NONE,
-                label: g.message(code: 'react.productSupplier.preferenceType.none.label')
+                label: g.message(code: 'react.productSupplier.preferenceType.none.label', default: "None")
             ],
         ]
         List<Map<String, String>> preferenceTypes = genericApiService.getList(PreferenceType.class.simpleName, [:]).collect {
--- a/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
@@ -16,6 +16,7 @@ import org.pih.warehouse.core.PreferenceType
 import org.pih.warehouse.core.Tag
 import org.pih.warehouse.core.User
 import org.pih.warehouse.core.UserService
+import org.pih.warehouse.data.ProductSupplierService
 import org.pih.warehouse.glAccount.GlAccountService
 import org.pih.warehouse.product.Category
 import org.pih.warehouse.product.ProductCatalog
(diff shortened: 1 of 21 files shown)


src/js/hooks/list-pages/productSupplier/useProductSupplierListTableData.jsx
```diff
@@ -1,25 +1,43 @@
...
+  }) => {
+    const {
+      product,
+      supplier,
+    } = filterParams;
+    return _.omitBy({
+      offset: `${offset}`,
+      max: `${state.pageSize}`,
+      ...sortingParams,
+      ...filterParams,
```
alannadolny, line 28: probably it's based on my preferences, but I prefer spreading parameters at the end of the object (do not treat this as a change request)

src/js/hooks/list-pages/useCommonFilters.jsx
```diff
@@ -0,0 +1,26 @@
+import { useState } from 'react';
+
+import { useDispatch } from 'react-redux';
+import { useHistory } from 'react-router-dom';
+
+const useCommonFilters = () => {
```
alannadolny, line 6: What is the purpose of this hook? I am unsure if this hook reduces the amount of code in other components because you have to destructure the returned object, so it's also a lot of code. Basically, this hook doesn't contain any logic, just other hooks.

src/js/wrappers/ListFilterFormWrapper.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+const ListFilterFormWrapper = ({ children }) => (
+  <div className="d-flex flex-column list-page-filters">
```
alannadolny, line 6: I was thinking for a while about those wrappers that you created. In the beginning, I loved the idea, but now I am a little bit confused about whether "common styles" should be created using wrappers. I think each of those list pages, filters, etc. should have the same className, and we should have those styles in something like a "common" or "global" scss file.

src/js/hooks/list-pages/productSupplier/useProductSupplierListTableData.jsx
kchelstowski: if I had done it this way, it would eventually override the `product` and the `supplier` properties.

src/js/hooks/list-pages/useCommonFilters.jsx
kchelstowski: I realized that on all of the list pages, the code that is inside this hook is redundant. Starting from now, we could reuse this hook for the future list pages (and refactor the existing ones) instead of copying and pasting those `useStates` etc. that are there.

src/js/wrappers/ListFilterFormWrapper.jsx
kchelstowski: The purpose of the wrappers were to reduce the redundancy of the code for an eventual future refactor. If we ever decided to add/remove/rename a className of those divs, we'd have to do it only in a wrapper, not in every `*List.jsx` file.

kchelstowski: Another way to achieve the same result would be to store the classNames in a const. I went this way hoping you all would like the idea.

src/js/hooks/list-pages/productSupplier/useProductSupplierListTableData.jsx
alannadolny: For my whole life, I was sure that this thirdObj should return "thirdObj". I was sure that the parameters from spreading were always overridden by those typed "directly" 🤡🤡
image

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -79,12 +80,24 @@ class SelectOptionsApiController {
...
+        List<Map<String, String>> defaultPreferenceTypes = [
+            [
+                id: ProductSupplierService.PREFERENCE_TYPE_MULTIPLE,
+                label: g.message(code: "react.productSupplier.preferenceType.multiple.label", default: "Multiple")
+            ],
+            [
+                id: ProductSupplierService.PREFERENCE_TYPE_NONE,
+                label: g.message(code: 'react.productSupplier.preferenceType.none.label', default: "None")
+            ],
+        ]
```
kchelstowski, line 92: I decided to add the default options here, not on the frontend side, to avoid potential problems with translations that we've had for statuses options once (it was due to the fact that the labels had been rendered before the translation plugin or the redux's current locale was injected).
cc @jmiranda

alannadolny (approved):

PR #4441, opened 2024-01-04
OBPIH-6020 Preference Type Modal

alannadolny:


src/js/components/productSupplier/PreferenceTypeColumn.jsx
```diff
@@ -16,20 +18,49 @@ const getLabel = (productSupplierPreferences) => {
...
   return (
-    <span>
-      {_.isObject(label)
-        ? <Translate id={label.id} defaultMessage={label.defaultMessage} />
-        : label}
-    </span>
+    <>
+      <span
+        className={label?.className}
+        onClick={onCellClick}
```
kchelstowski, line 44: here it doesn't matter, because the definition of your method is `() => ...`, but be aware that if you had an argument, it would be called with the `e` from `span`.
I mean that your current code evaluets to:
```
onClick={e => onCellClick(e)}
```

src/js/components/productSupplier/PreferenceTypeColumn.jsx
```diff
@@ -16,20 +18,49 @@ const getLabel = (productSupplierPreferences) => {
...
+              <Translate id={label.id} defaultMessage={label.defaultMessage} />
+              {' '}
+              {label?.icon}
+            </>
+          )
+          : label}
+      </span>
+      <PreferenceTypeModal
+        productSupplierId={productSupplierId}
+        isOpen={!!preferenceTypeModalData.length}
```
kchelstowski, line 59: I don't mind this, but in general the double negations are difficult to read - how about doing it as:
```js
isOpen={Boolean(preferenceTypeModalData.length)}
```

src/js/components/productSupplier/modals/PreferenceTypeModal.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+      if (!destinationParty) {
+        return {
+          ...acc,
+          default: (
+            <p className="default-preference-type">
+              <span className="preference-type-location">Default: </span>
+              <span>{name}</span>
+            </p>
+          ),
```
kchelstowski, line 38: returning HTML like that feels weird to me - if it would not be difficult to change, how about just storing the `name` and `destinationParty.name` in the list and then mapping it in the JSX output (85th line)?
If it is more complicated than I think, then I guess you can skip it.

src/js/components/productSupplier/modals/PreferenceTypeModal.jsx
```diff
@@ -0,0 +1,135 @@
...
+          <RiCloseFill
+            size="32px"
+            className="cursor-pointer"
+            role="button"
+            onClick={closeModal}
+          />
+        </div>
+        <div className="preference-type-modal-list-container">
+          <div>
+            {mapPreferenceTypes(modalData).default}
```
kchelstowski, line 82: let's assign the result of the `mapPreferenceTypes(modalData)` to a variable, because doing it this way, you run this method twice - for the first time here, and for the second time in the 85th line.

kchelstowski:
Additionally I think it would be an interesting improvement if you somehow stored already visited product sources' preference type modal. 
How I see is just adding maybe an additional `useState` that would store the calculated data by the `mapPreferenceTypes` and the key would be the product source id.
By doing it, we'd avoid the "complexed" calculations  if you closed/opened the modal over and over again.
It's just a "loud thought", so if you find it interesting, proceed, if not, then we could skip it.

src/js/components/productSupplier/PreferenceTypeColumn.jsx
alannadolny: I don't have any preference about that, so I can change it

src/js/components/productSupplier/modals/PreferenceTypeModal.jsx
```diff
@@ -0,0 +1,132 @@
...
+        <div className="preference-type-modal-list-container">
+          <div>
+            <p className="default-preference-type">
+              <span className="preference-type-location">Default: </span>
+              <span>{mappedPreferenceTypes?.default?.name}</span>
+            </p>
+          </div>
+          <div className="preference-type-modal-list">
+            {mappedPreferenceTypes?.preferenceTypes?.map((preferenceType) => (
+              <p>
```
kchelstowski, line 74: you forgot to add the `key`

PR #4442, opened 2024-01-05
OBPIH-6021 Add an API endpoint to delete a product supplier + remove unnecessary url mapping 

kchelstowski:
It seems not to be necessary to add a url mapping for this (and it was not necessary to add it for the list method), as it's handled fine by the generic mappings:
```groovy
"/api/${resource}s"(parseRequest: true) {
    controller = { "${params.resource}Api" }
    action = [GET: "list", POST: "create"]
}

"/api/${resource}s/$id/status"(parseRequest: true) {
    controller = { "${params.resource}Api" }
    action = [GET: "status", DELETE: "deleteStatus", POST: "updateStatus"]
}

"/api/${resource}s/$id"(parseRequest: true) {
    controller = { "${params.resource}Api" }
    action = [GET: "read", POST: "update", PUT: "update", DELETE: "delete"]
}
```
By doing so, I've also applied @jmiranda 's suggestion from [here](https://github.com/openboxes/openboxes/pull/4435#discussion_r1437925142) to change the url to be in plural form (productSuppliers, not productSupplier).

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
@@ -13,4 +13,9 @@ class ProductSupplierApiController {
         List<ProductSupplierListDto> productSuppliers = productSupplierService.getProductSuppliers(filterParams)
         render([data: productSuppliers, totalCount: productSuppliers.totalCount] as JSON)
     }
+
+    def delete() {
+        productSupplierService.delete(params.id)
+        render status: 204
+    }
 }
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -113,11 +113,6 @@ class UrlMappings {
             action = [GET: "getInventoryItem"]
         }
 
-        "/api/productSupplier" {
-            controller = { "productSupplierApi" }
-            action = [GET: "list"]
-        }
-
         "/api/locations/locationTypes" {
             controller = { "locationApi" }
             action = [GET: "locationTypes"]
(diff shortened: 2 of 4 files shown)


grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -13,4 +13,9 @@ class ProductSupplierApiController {
         List<ProductSupplierListDto> productSuppliers = productSupplierService.getProductSuppliers(filterParams)
         render([data: productSuppliers, totalCount: productSuppliers.totalCount] as JSON)
     }
+
+    def delete() {
+        productSupplierService.delete(params.id)
+        render status: 204
```
alannadolny, line 19: As I know, the delete method from `@Service` won't throw an exception if the entity we want to delete doesn't exist. Shouldn't we return another status in this case? Or we won't tell the frontend part whether this entity is deleted?

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
kchelstowski: @alannadolny I described the decision under the ticket. I apologize not for tagging everyone there.

kchelstowski: It's definitely an important decision in terms of coding conventions, so I suppose we'll discuss it in details.

jmiranda: Yeah it's a good question. For what it's worth 204 = "No Content", so it's not saying "delete was successful". 

It looks like the Grails docs imply that they always return 204.
https://docs.grails.org/5.2.4/guide/REST.html#_implementing_the_delete_action

I would be interested to see what a scaffolded controller looks like when generated in Grails 3.

As for the REST convention, it seems like 204 on the first attempt and then 404 on subsequent attempts is an accepted convention (based on the upvotes).
https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time

I need to read a little bit more, but I feel like returning 204 would be fine for now (and maybe the proper solution because it's not leaking implementation details i.e. 404 is telling me something that I wouldn't otherwise know). With that said, 204 is misleading. 

So let's leave it as 204 for now. I'll add a tech debt label and we can come back to this when we've had some more time to investigate.

awalkowiak (approved):

PR #4444, opened 2024-01-09
OBPIH-5135 Browsers cannot change their password

alannadolny:
description: https://pihemr.atlassian.net/browse/OBPIH-5135?focusedCommentId=148998

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
@@ -10,6 +10,7 @@
 package org.pih.warehouse.user
 
 import grails.validation.ValidationException
+import org.apache.http.auth.AuthenticationException
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.LocationRole
 import org.pih.warehouse.core.LocationRoleDataService
@@ -270,6 +271,29 @@ class UserController {
             flash.message = "${warehouse.message(code: 'default.not.found.message', args: [warehouse.message(code: 'user.label'), params.id])}"
             redirect(action: "list")
         }
+    }
+
+    def changePassword() {
+        User user = userGormService.get(params?.id)
+        if (user) {
+            try {
+                userService.changePassword(user, params?.password, params?.passwordConfirm)
+                flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'user.label'), user.id])}"
+                redirect(action: "edit", id: user.id)
+            } catch (ValidationException e) {
+                // This read function is used to avoid getting lazy initialization exceptions in
+                // rendering the edit page, it is done like in the update function above
+                user = User.read(params.id)
+                user.errors = e.errors
+                render(view: "edit", model: [userInstance: user])
+            } catch (AuthenticationException e) {
+                flash.message = e.message
+                redirect(action: "edit", id: user.id)
+            }
+            return
+        }
+        flash.message = "${warehouse.message(code: 'default.not.found.message', args: [warehouse.message(code: 'user.label'), params.id])}"
+        redirect(action: "list")
     }
 
     def disableLocalizationMode() {
--- a/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/user/UserController.groovy
@@ -287,7 +287,7 @@ class UserController {
                 user.errors = e.errors
                 render(view: "edit", model: [userInstance: user])
             } catch (AuthenticationException e) {
-                flash.message = e.message
+                flash.error = e.message
                 redirect(action: "edit", id: user.id)
             }
             return
--- a/grails-app/services/org/pih/warehouse/core/UserService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/UserService.groovy
@@ -16,9 +16,10 @@ import grails.util.Holders
 import grails.validation.ValidationException
 import groovy.sql.Sql
 import org.apache.commons.collections.ListUtils
+import org.apache.http.auth.AuthenticationException
 import org.hibernate.sql.JoinType
 import org.pih.warehouse.auth.AuthService
-import org.pih.warehouse.core.Role
+import org.grails.plugins.web.taglib.ApplicationTagLib
 
 @Transactional
 class UserService {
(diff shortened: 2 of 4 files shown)


grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
```diff
@@ -20,6 +20,10 @@ class RoleInterceptor {
                                 'update', 'importData', 'receive', 'showRecordInventory', 'withdraw', 'cancel', 'change', 'toggle', 'exportAsCsv']
     def static changeControllers = ['createProductFromTemplate']
 
+    def static omittedControllersWithChangeActions = [
```
awalkowiak, line 23: I think it should be handled by the proper configuration of `openboxes.security.rbac.rules` in the `runtime.groovy`

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
jmiranda: Thank you @awalkowiak

jmiranda (approved):
LGTM

jmiranda:
Actually, I'm having second thoughts about approving this. I'll add some discussion to the ticket.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -234,6 +239,16 @@ class UserController {
 
         def userInstance = User.get(params.id)
         if (userInstance) {
+            // user is not allowed to edit another user's account unless the user has at least an admin role
+            if (authService?.currentUser?.id != params.id && !userService?.isUserAdmin(authService?.currentUser)) {
```
awalkowiak, line 243: @jmiranda Initially I suggested this as a further improvement to the issue here, but it still does not resolve the issue 100%, we still might have some potential problems with unwanted changes made by users to their accounts and their authorization parameters (through the postman or something like that).
I think that the easiest would be to introduce a separate user edit page stripped of the authorization tab for editing personal data and a password (or even only a simple "change password" form with dedicated action) that could be accessed by the user himself and an admin+ user and a separate (current version of) user edit page for admin+ users.

awalkowiak: What do you think? Or we are moving this into 9.1, and skipping this as a quick hotfix for now, since Manon said in a ticket that it is not as pressing as it was while Lesotho was going live?

jmiranda: @awalkowiak Yeah, the separate form / action for changing the password is what I was thinking we could do for now. In fact, I think we should do this regardless of how we implement the record-level authorization rules. 

Regarding the release, I would still move this to 0.9.1 because it doesn't seem urgent enough to hotfix. But that's just my vote. I would check again with Kelsey and Manon to see if they still agree that it can wait until March.

jmiranda: We also need to add a ticket to think through how we want to handle the record-level authorization. I actually think this might be possible by adding a ROLE_SELF to the minimum required roles or adding a boolean property to the map to allow the user to edit their own profile and/or password.

Note: I imagine we still want to allow superusers (maybe admins) as well as the user themself to edit their password and update profile information. We also probably need to take user active flag into account.

So I guess for now we don't need to do anything fancy in the interceptor, we can just implement logic in the controller action. But consider delegating the auth logic to a service.
```
userService.canUserEditUser()
```
:shrug:

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -272,6 +273,29 @@ class UserController {
...
+        User user = userGormService.get(params?.id)
+        if (user) {
+            try {
+                userService.changePassword(user, params?.password, params?.passwordConfirm)
+                flash.message = "${warehouse.message(code: 'default.updated.message', args: [warehouse.message(code: 'user.label'), user.id])}"
+                redirect(action: "edit", id: user.id)
+            } catch (ValidationException e) {
+                // This read function is used to avoid getting lazy initialization exceptions in
+                // rendering the edit page, it is done like in the update function above
+                user = User.read(params.id)
```
kchelstowski, line 286: ### Good practices:

If we considered good practices, it shouldn't be the way to go (solving LazyInitializationException by re-fetching an entity), but join the needed associations at the query level.
Seeing this code, something tells me that we must have some kind of OSIV mechanism that is associated with a session, and maybe it is turned off when catching an exception? :thinking: 
I've mentioned it recently in my PR: https://github.com/openboxes/openboxes/pull/4435#discussion_r1438102346
So in a nutshell - I would expect that when we have the OSIV filter disabled, we should get the Lazy every time we want to access some associations when not in a transaction boundary. 
I'd consider creating a spike ticket to investigate this part, because it's weird that in the `catch` we lose the associations.
Or maybe it's maybe related to `render` vs `redirect`? 
Could you check that?

### Not considering good practices:
Obviously we won't experience any performance issues due to that additional fetch, so I won't force you to go with the join query route, especially if you had just followed the code from the method above.

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -272,6 +273,29 @@ class UserController {
...
+                user.errors = e.errors
+                render(view: "edit", model: [userInstance: user])
+            } catch (AuthenticationException e) {
+                flash.message = e.message
```
kchelstowski, line 290: I think `flash.error` would fit better. Used in view like:
```gsp
<g:if test="${flash.error}">
  <div class="errors">${flash.error}</div>
</g:if>
```

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -51,14 +52,8 @@ class UserService {
...
         // setting the new password in the database
         userInstance.properties = params
-        if (params.changePassword && userInstance.password != params.password) {
-            userInstance.password = params?.password?.encodeAsPassword()
-            userInstance.passwordConfirm = params?.passwordConfirm?.encodeAsPassword()
-        } else {
-            // Needed to bypass the password == passwordConfirm validation
-            userInstance.passwordConfirm = userInstance.password
-        }
-
```
kchelstowski, line 61: have you made sure this method is not used for imports where we probably wouldn't want to remove this part?

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -92,6 +87,21 @@ class UserService {
...
+                    code: 'errors.accessDenied.message',
+                    default: 'Apologies, but you are not authorized to access this page or to perform this action.',
+            )
+            throw new AuthenticationException(errorMessage)
+        }
+
+        if (user.password != password) {
+            user.password = password?.encodeAsPassword()
+            user.passwordConfirm = passwordConfirm?.encodeAsPassword()
+            user.save(failOnError: true)
```
kchelstowski, line 102: an alternative way could be to use the save from the data service, as it throws validation exception when validation fails by default. (no need to change, just titbit).

grails-app/services/org/pih/warehouse/core/UserService.groovy
alannadolny: I was able to import users, in the import users template there is no password column
image

awalkowiak (approved):

PR #4459, opened 2024-01-17
OBPIH-5967 Can edit completed putaway when go back to edit page after complete

drodzewicz:
Follow the discussion [in the ticket](https://pihemr.atlassian.net/browse/OBPIH-5967?focusedCommentId=149178)

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
@@ -64,11 +64,15 @@ class PutawayApiController {
             throw new IllegalArgumentException("User must be logged into a location to perform putaway")
         }
 
+        Order order = Order.get(jsonObject.id)
+        if (order && Putaway.getPutawayStatus(order.status) == PutawayStatus.COMPLETED) {
+            throw new IllegalArgumentException("Can't update completed putaway")
+        }
+
         User currentUser = User.get(session.user.id)
 
         Putaway putaway = new Putaway()
         bindPutawayData(putaway, currentUser, currentLocation, jsonObject)
-        Order order
 
         // Putaway stock
         if (putaway?.putawayStatus?.equals(PutawayStatus.COMPLETED)) {


grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
```diff
@@ -52,6 +52,7 @@ class PutawayApiController {
             putawayItem.inventoryLevel = InventoryLevel.findByProductAndInventory(putawayItem.product, putaway.origin.inventory)
             putawayItem.quantityAvailable = productAvailabilityService.getQuantityOnHandInBinLocation(putawayItem.inventoryItem, putawayItem.currentLocation)
         }
+        response.setHeader("Cache-Control", "no-store")
```
alannadolny, line 55: why was this needed?

alannadolny: Oh yea, you cced Artur twice :smile: 
image
Now, I got it, thanks

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/PutawayApiController.groovy
drodzewicz: :see_no_evil:

awalkowiak: I am still baffled by the need for this

drodzewicz: I feel like it is something that we need to discuss and decide.
I wanted to show the issue and a potential solution.

If we feel like disabling the cache in the header is not the way to go, just a single validation should be enough to prevent users mess with old data

awalkowiak: Yeah, I feel like, having the validation should be enough. However, you can showcase or explain the cache store  tomorrow morning

awalkowiak (approved):

PR #4460, opened 2024-01-17
OBPIH-6043 Add export invoice details

kchelstowski:


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -30,8 +32,8 @@ class InvoiceService {
 
     def getInvoices(Map params) {
         // Parse pagination parameters
-        def max = params.max ? params.int("max") : 10
-        def offset = params.offset ? params.int("offset") : 0
+        def max = params.max ? params.int("max") : null
```
awalkowiak, line 35: Lets go back to the default pagination, and set it to null only if there is format=csv, and no max in params

awalkowiak (changes requested):

awalkowiak (approved):

PR #4464, opened 2024-01-19
OBPIH-6079 Items are not visible in outbounds after using save button

alannadolny:


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -996,9 +996,13 @@ class AddItemsPage extends Component {
...
-      this.requestsQueue.enqueueRequest(
-        saveItemsRequest(payload),
-      );
+      if (this.props.isAutosaveEnabled) {
+        this.requestsQueue.enqueueRequest(
+          saveItemsRequest(payload),
+        );
+      } else {
+        await saveItemsRequest(payload)();
+      }
```
kchelstowski, line 1005: lgtm but this currying looks weird to me, that it is required here `(payload)()` vs above `(payload)` - is it because of the implementation of the `TaskQueue` that it expects to get a function that it further runs, hence we need one more arrow function?

kchelstowski (approved):

PR #4467, opened 2024-01-22
OBPIH-6080 Do not display a toast message when there is a network error

kchelstowski:


src/js/utils/apiClient.jsx
```diff
@@ -97,6 +97,9 @@ export const handleError = (error) => {
       });
       break;
     default:
+      if (error?.code === 'ERR_NETWORK') {
```
alannadolny, line 100: Can you try if this works?
image

kchelstowski: @alannadolny thanks, I was searching for something like this, but I didn't get any IDE suggestions, so I assumed there are not static strings for that and I did not see a reason to create one on my own for that thing.

src/js/utils/apiClient.jsx
```diff
@@ -97,6 +97,9 @@ export const handleError = (error) => {
       });
       break;
     default:
+      if (error?.code === AxiosError.ERR_NETWORK) {
+        break;
```
awalkowiak, line 101: Add a comment why it's done like that.

awalkowiak (changes requested):

src/js/utils/apiClient.jsx
alannadolny: Yeah I also thought about avoiding creating our own "errors", so I looked into Axios implementation if they allow us to use their "errors" :smile:

kchelstowski:
@alannadolny as I explained, there is no response status for this type of error, this is why I handled it in the `default` case.
Screenshot from 2024-01-22 12-15-52

src/js/utils/apiClient.jsx
awalkowiak: @kchelstowski How about:
```
// We don't want to spam "network errors" popups
// when a user loses the connection, because 
// we are using the "Lost connection" message 
// for this, plus we are going to add a visual
// indicator for this <put a number of the ticket you created here>
```

awalkowiak (approved):

PR #4468, opened 2024-01-22
OBPIH-6037 Add bin location data to expiration reports

drodzewicz:


grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
-        def location = Location.get(session.warehouse.id)
-        def categorySelected = (params.category) ? Category.get(params.category) : null
-        def inventoryItems = dashboardService.getExpiredStock(categorySelected, location)
-        def categories = inventoryItems?.collect { it.product.category }?.unique()
-        def quantityMap = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
-        def expiredStockMap = [:]
-        inventoryItems.each { inventoryItem ->
-            expiredStockMap[inventoryItem] = quantityMap[inventoryItem]
+        Location location = Location.get(session.warehouse.id)
+        Category categorySelected = Category.get(params.category)
```
alannadolny, line 579: Shouldn't we use data services since we are on Grails 3?

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
+        List<InventoryItem> inventoryItems = dashboardService.getExpiredStock(categorySelected, location)
+        List<Category> categories = inventoryItems?.collect { it.product.category }?.unique()
+
+        List<Map> data
+        if (withBinLocation) {
+            data = productAvailabilityService.getQuantityOnHandByBinLocation(location)
+        } else {
+            data = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
+                    .collect{ key, val -> [ inventoryItem: key, quantity: val ] }
         }
```
alannadolny, line 591: Can't it be just written using a ternary operator?

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
 
     def listExpiringStock() {
-        def expirationStatus = params.status
-        def location = Location.get(session.warehouse.id)
-        def category = (params.category) ? Category.get(params.category) : null
-        def inventoryItems = dashboardService.getExpiringStock(category, location, expirationStatus)
-        def categories = inventoryItems?.collect { it?.product?.category }?.unique().sort {
+        String expirationStatus = params.status
+        Location location = Location.get(session.warehouse.id)
+        Category category = Category.get(params.category)
```
alannadolny, line 612: The same as above

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
-        inventoryItems.each { inventoryItem ->
-            expiringStockMap[inventoryItem] = quantityMap[inventoryItem]
+
+        List<Map> data
+        if (withBinLocation) {
+            data = productAvailabilityService.getQuantityOnHandByBinLocation(location)
+        } else {
+            data = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
+                    .collect{ key, val -> [ inventoryItem: key, quantity: val ] }
         }
```
alannadolny, line 626: the same as above

grails-app/views/inventory/listExpiredStock.gsp
```diff
@@ -114,50 +119,47 @@
...
 											<tr class="${(counter++ % 2) == 0 ? 'even' : 'odd'}">
 												<td class="center">
-													<g:checkBox id="${inventoryItem?.id }" name="inventoryItem.id"
-														class="checkbox" checked="${false }"
-															value="${inventoryItem?.id }" />
-
+													<g:checkBox id="${dataEntry.inventoryItem?.id }"
+																name="inventoryItem.id"
+																class="checkbox"
+																checked="${false }"
```
alannadolny, line 129:
```suggestion
																checked="${false}"
```

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
+        List<Category> categories = inventoryItems?.collect { it.product.category }?.unique()
+
+        List<Map> data
+        if (withBinLocation) {
+            data = productAvailabilityService.getQuantityOnHandByBinLocation(location)
+        } else {
+            data = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
+                    .collect{ key, val -> [ inventoryItem: key, quantity: val ] }
         }
+        data = data.findAll { inventoryItems.contains(it.inventoryItem) }
```
kchelstowski, line 592: Can't we do this filtering in the PA query? 
I think we could just pass the `inventoryItems` (or `inventoryItems.id`) to `getQuantityByBinLocation` (or the second method as well) and just add the `IN :inventoryItems` clause?

This could additionally improve the performance.

And a suggestion (regarding what version you prefer):
```groovy
List<Map> data = withBinLocation 
   ? productAvailabilityService.getQuantityOnHandByBinLocation(location) 
   : productAvailabilityService.getQuantityOnHandByInventoryItem(location)
         .collect{ key, val -> [ inventoryItem: key, quantity: val ] }
```

grails-app/views/inventory/listExpiredStock.gsp
```diff
@@ -47,10 +47,15 @@
 							</a>
 						</div>
 
-						<g:link params="[format:'csv',category:params.category]" controller="${controllerName}" action="${actionName}" class="button">
+						<g:link params="[format:'csv',category: params.category]" controller="inventory" action="listExpiredStock" class="button">
 							<img src="${resource(dir:'images/icons/silk',file:'disk.png')}" alt="${warehouse.message(code: 'default.button.download.label') }" style="vertical-align: middle"/>
 							&nbsp; <g:message code="default.button.downloadAsCSV.label" default="Download as CSV"/>
 						</g:link>
+
+						<g:link params="[format:'csv',category: params.category, withBinLocation: true]" controller="inventory" action="listExpiredStock" class="button">
```
alannadolny, line 55:
```suggestion
						<g:link params="[format: 'csv',category: params.category, withBinLocation: true]" controller="inventory" action="listExpiredStock" class="button">
```

grails-app/views/inventory/listExpiredStock.gsp
```diff
@@ -47,10 +47,15 @@
 							</a>
 						</div>
 
-						<g:link params="[format:'csv',category:params.category]" controller="${controllerName}" action="${actionName}" class="button">
+						<g:link params="[format:'csv',category: params.category]" controller="inventory" action="listExpiredStock" class="button">
```
alannadolny, line 50: there are a lot of params, maybe it would be more readable if we split those into new lines?

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
-            expiringStockMap[inventoryItem] = quantityMap[inventoryItem]
+
+        List<Map> data
+        if (withBinLocation) {
+            data = productAvailabilityService.getQuantityOnHandByBinLocation(location)
+        } else {
+            data = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
+                    .collect{ key, val -> [ inventoryItem: key, quantity: val ] }
         }
+        data = data.findAll { inventoryItems.contains(it.inventoryItem) }
```
kchelstowski, line 627: this seems to be a repetition of what is done in the `listExpiredStock`. Maybe we could create a private method in this controller (as you know me - I'd prefer to have that type of logic in the service, but no must, since this probably needs to be refactored anyway, anytime soon).

grails-app/views/inventory/listExpiredStock.gsp
alannadolny: \+ split into new lines

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
drodzewicz: sure, I can.
I see you guys really hate if/else statements... :sweat_smile: 
I feel like simple if/else most of the time is much more readable than a ternary.
Ternary is great for very short statements and this case kind of applies to that, although I was also considering the case where in the future we want to add another parameter which would call another service method and ternary in this case would be very bad.

But yeah, I can change it to ternary.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
drodzewicz: Sure I can do it.
I just didn't want to modify the old service methods for that but it probably is worth changing it.
I was following the old logic for getting the expired or expiring inventoryItems and filtering the PA by that but I agree, this should probably be done in the db.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
kchelstowski: from my perspective it was only a suggestion for you to decide, which approach you prefer there. I want everyone to feel comfortable with their code style, since your version is not "breaking" any rules.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
jmiranda: Let's come up with our conventions for data services before we go lock, stock, and barrel. But what exactly would be changing here? 

Also, what is going to happen if params.category is null here?

And just for giggles, we *could* use Location, Category, Boolean as arguments to this action and be done with the data binding.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
jmiranda: No @kchelstowski that is not how this works. It's ternary or nothing!!!!

jmiranda: Just kidding. I don't feel strongly either way. I do feel strongly that the code should be refactored and probably moved to the service layer and that the getExpiredStock should probably handle all or most of the work. But we can save that for another day in case it'll be too much work.

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
jmiranda: One thing to mention is that we should probably have this type of logic in a more central service, like ProductAvailabilityService. The DashboardService was created before our product availability and should probably be phased out or refactored (i.e. it should delegate to product availability service and other reporting services).

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -575,49 +575,70 @@ class InventoryController {
...
-            expiredStockMap[inventoryItem] = quantityMap[inventoryItem]
+        Location location = Location.get(session.warehouse.id)
+        Category categorySelected = Category.get(params.category)
+        Boolean withBinLocation = params.boolean("withBinLocation")
+
+        List<InventoryItem> inventoryItems = dashboardService.getExpiredStock(categorySelected, location)
+        List<Category> categories = inventoryItems?.collect { it.product.category }?.unique()
+
+        List<Map> data
+        if (withBinLocation) {
```
jmiranda, line 586: I haven't read the ticket, but is there still a case where we don't need or want the bin location data?

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
kchelstowski: @jmiranda in general the difference is that data service's read operations are run in a `readOnly` transaction. 
I once mentioned to the guys, that I needed to do a research about what happens if we are in a service (marked as `@Transactional`) and we'd run a method from data service that is `readOnly` - if the outer (service's) transaction would "swallow" that, hence the fetch would be done using our outer transaction, or if it would set our outer transaction as `readOnly` for a while.
In the end I think this `readOnly` doesn't matter at this point, because as soon as we run the:
```groovy
categoryDataService.get(params.category)
```
the readOnly transaction would commit at this point (so in further lines we could still update a category).

> Also, what is going to happen if params.category is null here?

The data service's get evaluates to the basic `Domain.get` anyway, hence it'd return `null` if an entity didn't exist.

I think the data services methods matter much, if we do not take care of a transaction lifecycle on our own, but in the end, they evaluate to basic domain methods like `Domain.get(...)`, `Domain.delete(...)`, `Domain.save(...)`, so I don't have a strong opinion about forcing to use data services for such simple cases.
I find the data services very helpful and powerful for stuff like fetch join query, etc. where it has helped us a lot already.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -603,27 +605,28 @@ class ProductAvailabilityService {
...
-            data = collectQuantityOnHandByBinLocation(results)
+        List results = ProductAvailability.createCriteria().list {
+            createAlias('inventoryItem', 'inventoryItem', JoinType.LEFT_OUTER_JOIN)
+            createAlias('binLocation', 'binLocation', JoinType.LEFT_OUTER_JOIN)
+            projections {
+                groupProperty("product", "product")
+                groupProperty("inventoryItem", "inventoryItem")
+                groupProperty("binLocation", "binLocation")
+                sum("quantityOnHand", "quantityOnHand")
+                sum("quantityAvailableToPromise", "quantityAvailableToPromise")
```
kchelstowski, line 621: have you made sure this would eventually return 0? I'm asking because of this:
```sql
sum(case when pa.quantityAvailableToPromise > 0 then pa.quantityAvailableToPromise else 0 end)
```

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -684,22 +687,22 @@ class ProductAvailabilityService {
...
-						where pa.location = :location
-						and pa.inventoryItem = ii
-						group by ii
-						""", [location: location])
-
-            results.each {
-                quantityMap[it[0]] = it[1]
+    Map<InventoryItem, Integer> getQuantityOnHandByInventoryItem(Location location, List<InventoryItem> inventoryItems = []) {
+        if (!location) {
+            throw new RuntimeException("Must specify location in order to get quantity on hand")
```
kchelstowski, line 692: was there a reason you refactored this part to throw an exception? Maybe there is a case where we'd still want to proceed with e.g. refresh PA, and by doing so, you'd break that operation.

drodzewicz: From what I have checked it seems we always require a location, but to be on the safe side I think I will revert it to returning an empty array instead of throwing an Exception.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
drodzewicz: a function that transforms this result `collectQuantityOnHandByBinLocation` handles the logic for setting a 0 for null values, but does not handle negative values...
Thanks for pointing this out, I think I am going to have to revert this to a simple SQL because it is quite hard to implement this logic with `CASE WHEN` in the criteria

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -684,22 +688,22 @@ class ProductAvailabilityService {
...
-						where pa.location = :location
-						and pa.inventoryItem = ii
-						group by ii
-						""", [location: location])
-
-            results.each {
-                quantityMap[it[0]] = it[1]
+    Map<InventoryItem, Integer> getQuantityOnHandByInventoryItem(Location location, List<InventoryItem> inventoryItems = []) {
+        if (!location) {
+            return []
```
kchelstowski, line 693: this method by default was returning an empty Map, not a List - even the returned type suggests it should be the Map.
I think this would not work as expected with the list.

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
drodzewicz: thanks, there was a similar case that returns a list so I messed it up.

kchelstowski (approved):

PR #4469, opened 2024-01-23
OBPIH-6035 New triangle icons for adjustments in stock history

alannadolny:


grails-app/domain/org/pih/warehouse/inventory/TransactionType.groovy
```diff
@@ -39,4 +39,8 @@ class TransactionType implements Serializable {
         String regex = "\\${Constants.LOCALIZED_STRING_SEPARATOR}"
         return name.split(regex)[0] == transactionTypeName.split(regex)[0]
     }
+
+    Boolean isAdjustment() {
+        compareName("Adjustment")
```
awalkowiak, line 44: Let's move `"Adjustment"` to the const, and please add a comment here why it's determined like that.

awalkowiak (changes requested):

drodzewicz (approved):

grails-app/domain/org/pih/warehouse/inventory/TransactionType.groovy
```diff
@@ -39,4 +39,9 @@ class TransactionType implements Serializable {
         String regex = "\\${Constants.LOCALIZED_STRING_SEPARATOR}"
         return name.split(regex)[0] == transactionTypeName.split(regex)[0]
     }
+
+    // Checking whether the transaction type is adjustment for displaying the appropriate icon on the stock history page
```
awalkowiak, line 43: @alannadolny I meant rather to put a comment here "**why** it's determined by doing `compareName(Constants.ADJUSTMENT)`", not "for what it is used"

PR #4475, opened 2024-01-24
OBPIH-5067 Sort items on Vendor filter on invoice list alphabetically 

alannadolny:


grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -112,6 +112,9 @@ class OrganizationService {
             if (params.active) {
                 eq('active', true)
             }
+            if (params.sort && params.order) {
+                order(params.sort, params.order)
+            }
```
kchelstowski, line 117: this would probably crash if a user provided a non-existing sort param, e.g. `&sort=test`. (Remember this endpoint is achievable also from API, not only from the UI).
I don't know if we wanna prevent that from happening and assume that no one would do so, but we used to do it like:
```groovy
if (params.sort == "name") {
  order("name", params.order ?: "asc")
}

kchelstowski (approved):

PR #4476, opened 2024-01-25
OBPIH-6020 Preference Type Modal (improvements after QA)

alannadolny:


src/js/components/productSupplier/modals/PreferenceTypeModal.jsx
```diff
@@ -14,6 +15,16 @@ const PreferenceTypeModal = ({
   modalData,
   productSupplierId,
 }) => {
+  useEffect(() => {
+    if (isOpen) {
+      document.body.style.overflowY = 'hidden';
```
awalkowiak, line 20: Is there no better way to achieve it?

awalkowiak: Ok I looked into the related issue on the `react-modal` repo and this is a suggested solution, I guess we can live with it...

PR #4477, opened 2024-01-25
OBPIH-5967 Add completed putaway validation before removing putawayItems

drodzewicz:


src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -325,7 +325,7 @@ class AddItemsPage extends Component {
   }
 
   async nextPage(formValues) {
-    const status = this.state.inboundReturn.status === StockTransferStatus.PLACED
+    const status = this.state.inboundReturn.status !== StockTransferStatus.PLACED
```
awalkowiak, line 328: I don't one thing here. The ticket mentions putaway, commit message stock transfer and the change is made in the inbound return. Which one is correct?

awalkowiak (changes requested):

src/js/components/returns/inbound/AddItemsPage.jsx
drodzewicz: This was supposed to be on the other branch https://github.com/openboxes/openboxes/pull/4478 sorry about that

awalkowiak (approved):

PR #4479, opened 2024-01-26
OBPIH-6022 Product Sources List Actions (improvements after QA)

alannadolny:


grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
```diff
@@ -113,6 +114,24 @@ class ErrorsController {
...
+                        code: "errors.existingAssociation.message",
+                        default: "Resource could not be deleted because of an existing association"
+                )
+
+                render([ errorCode: 500, errorMessages: [message]] as JSON)
+                return
+            }
+
+            Throwable root = ExceptionUtils.getRootCause(request.getAttribute('exception'))
+            render([errorCode: 500, errorsMessages: [root.getMessage()]])
```
kchelstowski, line 130: you have a typo here - `errorsMessages` instead of `errorMessages`.
By the way if we have a single message, I'd consider using `errorMessage`.

src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
```diff
@@ -17,6 +20,15 @@ const useProductSupplierActions = ({ fireFetchData }) => {
   const deleteProductSupplier = async (onClose, productSupplierId) => {
     try {
       await productSupplierApi.deleteProductSupplier(productSupplierId);
+      notification(NotificationType.SUCCESS)({
+        message: translate({
+          id: 'react.productSupplier.deleted.label',
+          defaultMessage: `Product Source ${productSupplierId} deleted`,
+          data: {
+            id: productSupplierId,
```
kchelstowski, line 28: wow, I've never seen this. :clap:

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
jmiranda: > you have a typo here - errorsMessages instead of errorMessages.

Good catch.

> By the way if we have a single message, I'd consider using errorMessage.

Can you elaborate on that? I assume our convention is to use the "errorMessages" key, so why would it be better to have a duplicate key "errorMessage" that requires the API client to check for values under both keys?

src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
jmiranda: Can you elaborate on that? What does "this" refer to?

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/core/ErrorsController.groovy
kchelstowski: @jmiranda we already have `errorMessage` - e.g. look at `handleValidationErrors` - we pass both `errorMessage` and `errorMessages`. Frontend side (error handler) is already prepared for both options, so we wouldn't have to introduce anything new. It's just that if we have a single message (like in this), not an array of messages (like for validation errors), we should pass `errorMessage` in my opinion, because in the end frontend would still destruct this array and make it a single message.

src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
kchelstowski: @jmiranda "this" refers to using `data` property of `transalate` method. I haven't known we could pass custom params for translation on the react side and then use it in the messages.properties like he did:
```groovy
react.productSupplier.deleted.label=Product Source ${id} deleted
```

awalkowiak (approved):

drodzewicz (approved):

PR #4480, opened 2024-01-29
OBPIH-6020 Preference Type Modal

alannadolny:
I included fixes after @awalkowiak's review in the product source feature branch in this PR

### DIFF
--- a/src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
+++ b/src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
@@ -75,8 +75,8 @@ const useProductSupplierActions = ({ fireFetchData }) => {
 
   const getActions = useCallback((productSupplierId) => (hasPermissions({
     user: currentUser,
-    roles: [RoleType.ROLE_PRODUCT_MANAGER],
     minimumRequiredRole: isAdmin,
+    supplementalRoles: [RoleType.ROLE_PRODUCT_MANAGER],
   }) ? [
       {
         defaultLabel: 'Edit',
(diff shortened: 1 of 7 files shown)


src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
```diff
@@ -60,23 +61,27 @@ const useProductSupplierActions = ({ fireFetchData }) => {
...
-    {
-      defaultLabel: 'Edit',
-      label: 'react.productSupplier.edit.label',
-      leftIcon: <RiPencilLine />,
-      onClick: () => {
-        window.location = PRODUCT_SUPPLIER_URL.edit(productSupplierId);
+  const getActions = useCallback((productSupplierId) => (hasPermissions({
+    user: currentUser,
+    roles: [RoleType.ROLE_PRODUCT_MANAGER],
+    minimumRequiredRole: isAdmin,
```
awalkowiak, line 67: Do these need to be separated? Could `ROLE_ADMIN` be moved to the roles, and then just checked if all required roles are there?

awalkowiak (changes requested):

src/js/hooks/list-pages/productSupplier/useProductSupplierActions.jsx
alannadolny: I think it's better to have it separated because it's a minimum required role, so when I pass the admin role, I can also be a superuser, and it still should be true. When I have to check the admin role within the roles argument I have to include the superuser here.

awalkowiak: @alannadolny Ok, but then call it `supplementalRoles` as it is on the backend, and maybe change the ordering of these params. So we have "required roles" first, and then "supplemental" ones.

awalkowiak:
@alannadolny plus fix the conflicts

awalkowiak (approved):

PR #4481, opened 2024-01-29
OBPIH-5905 New submenus in Inventory menu

alannadolny:
I included here the simple change with quotes mentioned in this ticket: https://pihemr.atlassian.net/browse/OBPIH-6067?atlOrigin=eyJpIjoiYmU0MTAxY2JjNGRiNDY2MThkMTQ0ZmM1ZmQ2NjU0MzkiLCJwIjoiaiJ9


grails-app/conf/runtime.groovy
```diff
@@ -137,7 +137,7 @@ openboxes {
             defaultLabel = "Analytics"
             menuItems = [
                 // TODO: Add option to include label 'beta'
-                [label: "inventory.browse.label", defaultLabel: "Browse Inventory", href: "/inventoryBrowser/index"],
+                [label: "inventory.view.label", defaultLabel: "Inventory Browser", href: "/inventoryBrowser/index"],
```
awalkowiak, line 140: `inventory.view.label` is `View Inventory`m your default is `Inventory Browser`, which one you are expecting here? If the later one, then add a separate message property, if the former one, then change the default label

grails-app/i18n/messages.properties
```diff
@@ -1192,7 +1192,7 @@ inventory.backToDashboard.label=Back to dashboard
 inventory.backToStockCard.label=Back to stock card
 inventory.belowMinimumLevel.label=Below minimum level
 inventory.belowReorderLevel.label=Below reorder level
-inventory.browse.label=Browse inventory
```
awalkowiak, line 1195: Is this one no longer used?

awalkowiak (changes requested):

awalkowiak (approved):

PR #4482, opened 2024-01-29
OBPIH-6014 Fix async select dropdown to be wider for longer labels

kchelstowski:
The point was to re-create the dropdown and its width when the options change, because before for the async select, the width has never dynamically changed regarding the options that were loaded.
Thanks for pair programing @drodzewicz @alannadolny

### DIFF
--- a/src/js/components/form-elements/FilterSelectField.jsx
+++ b/src/js/components/form-elements/FilterSelectField.jsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import _ from 'lodash';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


src/js/components/form-elements/FilterSelectField.jsx
```diff
@@ -66,7 +74,10 @@ const Menu = (props) => {
       target={inputContainer}
       container={document.getElementById('root')}
     >
-      <Dropdown inputContainerRec={inputContainer.getBoundingClientRect()} >
+      <Dropdown
+        inputContainerRec={inputContainer.getBoundingClientRect()}
+        hasOptions={!!props.options.length}
```
awalkowiak, line 79: Are `props.options` required? Is it possible it will be null? Should it be a part of `propTypes`?

kchelstowski: `options` prop from the react-select is an empty array by default.

awalkowiak (approved):

PR #4484, opened 2024-01-29
OBPIH-6044 Export invoice lines details on invoices list page

drodzewicz:


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -440,6 +440,11 @@ class UrlMappings {
         }
 
         // Invoice API
+        "/api/invoices/items"(parseRequest: true) {
+            controller = "invoiceApi"
+            action = [GET: "getAllInvoiceItems"]
+        }
+
```
kchelstowski, line 447: I don't mind it, but I think we used to do it via the export mapping, with a proper param (e.g. "orderItems").

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -127,8 +128,23 @@ class InvoiceApiController {
         return invoice
     }
 
+    def getAllInvoiceItems() {
+        Location location = Location.get(session.warehouse.id)
+        params.partyFromId = location?.organization?.id
+        List<InvoiceList> invoices = invoiceService.getInvoices(params)
```
kchelstowski, line 134: I spoke about that with @drodzewicz, but I want to make my point here, so that everyone can give their opinion - in my opinion it should be dealt directly via `InvoiceItem.createCriteria()` and 
```groovy
invoice {
  'in'("id" ,params.invoicesIds)
}
```
clause.
Plus I would monitor the performance here, because we might run into many N+1 - if we do, we could play with the data services and joins, but for now, I'd like to get others opinion about the approach of first getting the invoices and then getting the invoice items.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -460,4 +468,48 @@ class InvoiceService {
...
+
+        invoiceItems?.each { InvoiceItem invoiceItem ->
+            csv.printRecord(
+                    invoiceItem.invoice?.invoiceNumber,
+                    invoiceItem.invoice?.vendorInvoiceNumber,
+                    "${invoiceItem.invoice?.party?.code} ${invoiceItem.invoice?.party?.name}",
+                    invoiceItem.invoice?.currencyUom.name,
+                    invoiceItem.invoice?.status?.name(),
+                    "${invoiceItem.invoice?.partyFrom?.code} ${invoiceItem.invoice?.partyFrom?.name}",
+                    invoiceItem.invoice?.invoiceType?.code?.name() ?: "INVOICE",
```
kchelstowski, line 501: let's use the enum from `InvoiceTypeCode` in the fallback

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -460,4 +468,48 @@ class InvoiceService {
...
+                    invoiceItem.product?.productCode ?: "all",
```
kchelstowski, line 502: what does "all" mean here?

src/js/api/urls.js
```diff
@@ -34,6 +34,7 @@ export const STOCK_TRANSFER_REMOVE_ALL_ITEMS = id => `${STOCK_TRANSFER_BY_ID(id)
 
 // INVOICE
 export const INVOICE_API = `${API}/invoices`;
+export const INVOICE_ITEMS_API = `${API}/invoices/items`;
```
kchelstowski, line 37: if we agree this should be the url, then I think this could be shortened to:
```js
export const INVOICE_ITEMS_API = `${INVOICE_API}/items`;

File: src/js/components/invoice/InvoiceListTable.jsx [hunk 6]
```diff
@@ -145,7 +146,7 @@ const InvoiceListTable = ({
             <button
               type="button"
               className="dropdown-item"
-              onClick={() => downloadInvoices()}
+              onClick={downloadInvoices}
```
kchelstowski, line 149: I said it once to @alannadolny , but we need to be aware, that in this case, if we added any argument to the `downloadInvoices` method, this would be run with the `event` from the button.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
drodzewicz: yes this is probably the way to go, this should be on `/api/invoiceItems/`

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
drodzewicz: I don't think we are at risk of N+1 queries, if I am not missing anything this should be max 2 queries one for invoices and one for invoiceItems.

My goal was to reuse the query filter logic on  `invoiceService.getInvoices(params)` since we are expecting to get only those InvoiceItems that are a part of filtered invoices from invoice/list page, basically aggregate these queries.
First get invoices (filtered if necessary) and then based on these invoices get all of the invoiceItems

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -460,4 +468,48 @@ class InvoiceService {
...
+                    invoiceItem?.orderAdjustment ? invoiceItem?.description : invoiceItem.product?.name,
+                    invoiceItem.order?.orderNumber,
+                    invoiceItem.glAccount?.name,
+                    invoiceItem.budgetCode?.name,
+                    invoiceItem?.quantity,
+                    invoiceItem?.quantityPerUom,
+                    invoiceItem?.unitPrice ?: 0,
+                    invoiceItem?.totalAmount ?: 0,
```
kchelstowski, line 510: @drodzewicz I meant you might run into N+1 in this place, where you refer to many different entities and this would run N+1 for every invoice item. So my N+1 comment from above was not directly refering to the "two queries" you made, but overall.

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
drodzewicz: on invoice show page we have a similar case where we label adjustments as all since adjustments don't have a product code
https://github.com/openboxes/openboxes/blob/3627a1dd5fcf53edcb7f497ff5c07680b0d1e446/grails-app/views/invoice/_invoiceItems.gsp#L23-L25

drodzewicz:
Pushed some requested changes suggetsed by @kchelstowski and waiting for other developers (@alannadolny @awalkowiak), to review this PR and give an opinion on Kacpers comments

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
alannadolny, line 134: One more point here - should we pass the whole params object as an argument as it is done here?
```
List<InvoiceList> invoices = invoiceService.getInvoices(params)
```

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -460,4 +469,50 @@ class InvoiceService {
         }
         return csv
     }
+
+    CSVPrinter getInvoiceItemsCsv(List<InvoiceItem> invoiceItems) {
+        def g = Holders.grailsApplication.mainContext.getBean('org.grails.plugins.web.taglib.ApplicationTagLib')
```
alannadolny, line 474: Maybe it is worth to use move this to a function (like it is already done in some places)
```
    def getApplicationTagLib() {
        return Holders.grailsApplication.mainContext.getBean(ApplicationTagLib)
    }
```

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
alannadolny: @kchelstowski Yup, you're right, I also prefer to use our data services here (or just measure the performance and make a decision based on that)

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -490,6 +490,11 @@ class UrlMappings {
             action = [GET: "invoiceTypeCodes"]
         }
 
+        "/api/invoiceItems"(parseRequest: true) {
+            controller = "invoiceApi"
+            action = [GET: "getAllInvoiceItems"]
```
awalkowiak, line 495: I think this should be in a separate `InvoiceItemApiController`, because what I see the `getInvoiceItems` in the `InvoiceAPIController` is meant to get the invoice items from a specific invoice item, so if this one is meant to pull all invoice items from different invoices, then this should land in a separate controller.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
drodzewicz: we also have 
```groovy
"/api/invoiceItems/$id/validation" {
    controller = "invoiceApi"
    action = [POST: "validateInvoiceItem"]
}
```
so this will also have to end up in invoiceItemApicontroller

awalkowiak: Not really, I think this one can stay where it is, at least for now

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
awalkowiak: @drodzewicz On the other hand if this is for "import all" case you could also just do: `get invoices` -> `return invoices.invoiceItems`. Is the pagination required here? However if this is going to land in its own api controller, or the pagination params are required, then: 

I agree with @kchelstowski on that one in the long term. If you need to do some further invoice filtering you still can do it in the suggested by Kacper way as a:
```groovy
invoice {
  'in'('id', params.invoiceIds) // if there are invoice ids passed
  whatever other invoice filter you have there
}
```
Instead of doing `get invoices` -> ` get invoice items` by these invoices.

drodzewicz:
So after looking through the problem again, I came to the conclusion that my initial approach was wrong and this exact export should not have bee handled by a different endpoint targeting a  list of invoiceItems but instead I should have treated it as an expansion to invoice/list since we are interested in all of the filter parameters to persist from the invoice list to invoiceItem.

An exact case is handled in Purchase Order list
https://github.com/openboxes/openboxes/blob/3627a1dd5fcf53edcb7f497ff5c07680b0d1e446/grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy#L28-L44

So following the example of PO list I did the same thing, which is exporting a different document based on passed parameter `invoiceItems=true`.

Additionally as Kacper mentioned in previous comments, it would be nice to fetch invoice items eagerly for this export to avoid unnecessary N+1 queries to the database, so I added a parameter `fetchItemsEagerly` to specify the FetchMode eager for invoiceItems on invoicesList.
After comparing the export time of both cases here are some results.
Exporting 780 invoice items from 350 Invoices:
- with eager fetch: 1309 ms (avg)
- without eager fetch: 1347 ms (avg)

src/js/hooks/list-pages/invoice/useInvoiceListTableData.jsx
```diff
@@ -3,6 +3,7 @@ import { useDispatch } from 'react-redux';
 
 import { hideSpinner, showSpinner } from 'actions';
 import invoiceApi from 'api/services/InvoiceApi';
+import invoiceItemApi from 'api/services/InvoiceItemApi';
```
kchelstowski, line 6: I think it is not used?

drodzewicz: it's not, thanks

kchelstowski (approved):
thanks for the times @drodzewicz . Even though the difference is not that huge right now, we will get the benefit from it when the invoice table becomes bigger and bigger.

PR #4485, opened 2024-01-31
OBPIH-5932 Add validation to lot expiration date update when user clicks previous button on inbound returns

drodzewicz:
Same logic as for next page should be applied to previous page on inbound returns

### DIFF
--- a/src/js/components/returns/inbound/AddItemsPage.jsx
+++ b/src/js/components/returns/inbound/AddItemsPage.jsx
@@ -463,7 +463,7 @@ class AddItemsPage extends Component {
         return;
       }
     }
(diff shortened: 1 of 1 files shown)


src/js/components/returns/inbound/AddItemsPage.jsx
```diff
@@ -457,13 +457,14 @@ class AddItemsPage extends Component {
   }
 
   async previousPage(values, invalid) {
+    console.log(invalid)
```
awalkowiak, line 460: `console.log(invalid)`

awalkowiak (changes requested):

awalkowiak (approved):

PR #4486, opened 2024-01-31
OBPIH-6017+OBPIH-6071 Fix inconsistency in generating a product sourc…

kchelstowski:
…e code between creating manually/import vs PO


grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -26,6 +29,7 @@ class ProductSupplierController {
     def documentService
     def identifierService
     ProductSupplierDataService productSupplierGormService
+    ProductSupplierService productSupplierService
```
alannadolny, line 32: huh, that's interesting that we didn't have `productSupplierService` injected in the `productSupplierController` :smile: :smile:

alannadolny (approved):

PR #4487, opened 2024-01-31
OBPIH-6042 Add filter by date range to expiration reports

alannadolny:


awalkowiak:
@alannadolny there are some conflicts here

grails-app/controllers/org/pih/warehouse/inventory/InventoryController.groovy
```diff
@@ -574,13 +575,12 @@ class InventoryController {
...
-        def inventoryItems = dashboardService.getExpiredStock(categorySelected, location)
-        def categories = inventoryItems?.collect { it.product.category }?.unique()
-        def quantityMap = productAvailabilityService.getQuantityOnHandByInventoryItem(location)
-        def expiredStockMap = [:]
+    def listExpiredStock(InventoryReportCommand command) {
+        command.location = Location.get(session.warehouse.id)
+        List<InventoryItem> inventoryItems = dashboardService.getExpiredStock(command.category, command.location, command.startDate, command.endDate)
+        List<Category> categories = inventoryItems?.collect { it.product.category }?.unique()
+        Map<InventoryItem, Integer> quantityMap = productAvailabilityService.getQuantityOnHandByInventoryItem(command.location)
+        LinkedHashMap<InventoryItem, Integer> expiredStockMap = [:]
```
kchelstowski, line 583: `LinkedHashMap` is an implementation of `Map` interface. 
We should use interfaces as types if possible, hence `Map` should be the type here.
The same case as with Lists - you shouldn't define the type as `ArrayList<String> list = [...]`, but as `List<String> list = [...]`

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -163,6 +163,14 @@ class DashboardService {
             expiredStock = expiredStock.findAll { item -> item?.product?.category == category }
         }
 
+        if (startDate) {
+            expiredStock = expiredStock.findAll { item -> item?.expirationDate >= startDate }
+        }
+
+        if (endDate) {
+            expiredStock = expiredStock.findAll { item -> item?.expirationDate <= endDate }
+        }
```
kchelstowski, line 172: this feels odd to me - this should be probably filtered out at the query level, but I can see that you've followed the convention, so I don't have an opinion if it should be handled now or if we should treat it as a tech debt.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -175,7 +183,7 @@ class DashboardService {
      * @param threshold the threshold filter
      * @return a list of inventory items
      */
-    List getExpiringStock(Category category, Location location, String expirationStatus) {
+    List getExpiringStock(Category category, Location location, String expirationStatus, Date startDate, Date endDate) {
```
kchelstowski, line 186: just a comment - 4 arguments feels a lot to me, I think in the near future, we should rely on command classes more, since they give us more flexibility and better dev experience in my opinion.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -188,6 +196,14 @@ class DashboardService {
...
         }
 
+        if (startDate) {
+            expiringStock = expiringStock.findAll { item -> item?.expirationDate >= startDate }
+        }
+
+        if (endDate) {
+            expiringStock = expiringStock.findAll { item -> item?.expirationDate <= endDate }
+        }
+
```
kchelstowski, line 206: same comment as above

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -175,32 +184,40 @@ class DashboardService {
...
+        if (command.category) {
+            expiringStock = expiringStock.findAll { item -> item?.product?.category == command.category }
+        }
+
+        if (command.startDate) {
+            expiringStock = expiringStock.findAll { item -> item?.expirationDate >= command.startDate }
+        }
+
+        if (command.endDate) {
+            expiringStock = expiringStock.findAll { item -> item?.expirationDate <= command.endDate }
```
kchelstowski, line 205: I'm approving this one, but this needs to be refactored asap, especially if the performance is bad there.

kchelstowski (approved):

awalkowiak (approved):

PR #4491, opened 2024-02-05
OBPIH-6023 Add reusable Section and Subsection components

kchelstowski:
Peek 2024-02-05 14-24

### DIFF
--- a/src/js/components/Layout/v2/Subsection.jsx
+++ b/src/js/components/Layout/v2/Subsection.jsx
@@ -1,0 +1,60 @@
+import React, { useState } from 'react';
+
+import PropTypes from 'prop-types';
+import { RiArrowDownSLine, RiArrowUpSLine } from 'react-icons/ri';
+
+import Translate from 'utils/Translate';
+
+const Subsection = ({
+  title,
+  collapsable,
+  children,
+}) => {
+  // If a subsection is not collapsable, it is always expanded
+  // (collapsable: false --> expanded: true)
+  // If a subsection is collapsable, it is not expanded by default
+  // (collapsable: true --> expanded: false)
+  const [expanded, setExpanded] = useState(!collapsable);
+
+  const triggerCollapse = () => {
+    setExpanded(!expanded);
+  };
+
+  return (
+    <div className="v2-subsection">
+      <div className="subsection-title-wrapper">
+        <span
+          role="button"
+          tabIndex={0}
+          onClick={collapsable ? () => triggerCollapse() : null}
+          onKeyDown={collapsable ? () => triggerCollapse() : null}
+          style={collapsable ? { cursor: 'pointer' } : { cursor: 'unset' }}
+        >
+          <Translate id={title.label} defaultMessage={title.defaultMessage} />
+          {collapsable
+            && (expanded
+              ? <RiArrowUpSLine className="arrow-up" />
+              : <RiArrowDownSLine className="arrow-down" />)}
+        </span>
+      </div>
+      <div className={`subsection-body ${expanded ? 'subsection-body-expanded' : ''}`}>
+        {children}
+      </div>
+    </div>
+  );
+};
+
+export default Subsection;
+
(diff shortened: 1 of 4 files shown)


src/js/components/Layout/v2/Subsection.jsx
```diff
@@ -0,0 +1,60 @@
...
+          {collapsable
+            && (expanded
+              ? <RiArrowUpSLine className="arrow-up" />
+              : <RiArrowDownSLine className="arrow-down" />)}
+        </span>
+      </div>
+      <div className={`subsection-body ${expanded ? 'subsection-body-expanded' : ''}`}>
+        {children}
+      </div>
+    </div>
```
alannadolny, line 43: Maybe it is possible to reduce the amount of JS code inside the HTML...
Maybe you can create an object:
```
{
collapsable: {
onClick: ....,
onKeyDown: ...,
style: ...,
},
...
}
```
and spread it on this span?
I am just thinking out loud, you can skip it

alannadolny (approved):

src/js/components/Layout/v2/Subsection.jsx
```diff
@@ -0,0 +1,60 @@
...
+          role="button"
+          tabIndex={0}
+          onClick={collapsable ? () => triggerCollapse() : null}
+          onKeyDown={collapsable ? () => triggerCollapse() : null}
+          style={collapsable ? { cursor: 'pointer' } : { cursor: 'unset' }}
+        >
+          <Translate id={title.label} defaultMessage={title.defaultMessage} />
+          {collapsable
+            && (expanded
+              ? <RiArrowUpSLine className="arrow-up" />
```
drodzewicz, line 36: this can be a single csv with a class that rotates it when it is active
```jsx
<RiArrowUpSLine className={`arrow-up {expanded ? 'arrow-up--expanded' : ''}`} />
```
```css
.arrow-up--expanded {
  transform: rotate(180deg);
}
```

src/js/components/Layout/v2/Subsection.jsx
```diff
@@ -0,0 +1,60 @@
...
+}) => {
+  // If a subsection is not collapsable, it is always expanded
+  // (collapsable: false --> expanded: true)
+  // If a subsection is collapsable, it is not expanded by default
+  // (collapsable: true --> expanded: false)
+  const [expanded, setExpanded] = useState(!collapsable);
+
+  const triggerCollapse = () => {
+    setExpanded(!expanded);
+  };
```
drodzewicz, line 21: instead of a single `collapsable` prop I would also add `isExpanded: boolean` as an initial value, that way we can control if the section is collapsed or not for both states when it is **collapsable** or **not collapsable**, so something like:

```jsx
const Subsection = ({
  title,
  isExpanded,
  collapsable,
  children,
}) => ...

...
Subsection.defaultProps = {
  isExpanded: true,
  collapsable: true,
};
```

```jsx
const [expanded, setExpanded] = useState(isExpanded);
```

prevent any toggle action if section is not collapsible, maybe name `disabled` fro this prop sound better but collapsable should also be good.
```jsx
const triggerCollapse = () => {
  if (collapsable) {
   setExpanded(!expanded);
  }
};
```

drodzewicz (changes requested):

src/js/components/Layout/v2/Subsection.jsx
kchelstowski: Currently we are fine, because the `onClick` method would be `null` if a subsection is not collapseable, but I will add it just in case we would refactor something in the future.

kchelstowski: I've added additional prop `expandedByDefault` as you suggested, to make it more reusable in different scenarios. I didn't decide to go for `isExpanded`, not to mix `isExpanded` with `expanded`, as it might cause confusions.

drodzewicz (approved):

PR #4492, opened 2024-02-06
OBPIH-6101 Create a reusable Switch component

alannadolny:


src/js/components/form-elements/v2/Switch.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const onChangeValue = () => {
+    changeValue((previousValue) => {
+      onChange?.(!previousValue);
+      return !previousValue;
+    });
+  };
+
+  return (
+    <div id="switch-container">
```
awalkowiak, line 20: Why you are using `id` here? What if there are going to be multiple switch fields on the form for some usecase?

src/js/components/form-elements/v2/Switch.jsx
```diff
@@ -0,0 +1,80 @@
...
+      <label htmlFor="toggle" className="switch">
+        <input
+          id="toggle"
```
awalkowiak, line 23: I fell like this could use some sort of additional part of id passed in props :thinking:

awalkowiak (changes requested):

src/js/components/form-elements/v2/Switch.jsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from 'react';
+
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+import './style.scss';
+
+const Switch = ({ defaultValue, titles, onChange }) => {
```
kchelstowski, line 9: this should be ready to receive the `...field` prop from the `react-hook-form`. I can't exactly say now what all the props will be, so for now let's at least add it as `...rest` (let's name it as `...fieldProps`).

drodzewicz (approved):

kchelstowski (approved):

awalkowiak (approved):

PR #4495, opened 2024-02-07
OBPIH-5832 Rename pagination and filter command and implement validat…

kchelstowski:
…eable interface

The PR is addressing @jmiranda 's suggestions: https://github.com/openboxes/openboxes/pull/4438#discussion_r1480341156
Additionally, I've added `implements Validateable` for those commands, so we do not get warnings in the console, even though we do not need any validation for the filter params now.


alannadolny (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -46,12 +47,15 @@ class ProductSupplierService {
     ProductSupplierDataService productSupplierGormService
 
 
-    List<ProductSupplier> getProductSuppliers(ProductSupplierListParams params) {
+    List<ProductSupplier> getProductSuppliers(ProductSupplierFilterCommand params) {
+        if (params.hasErrors()) {
+            throw new ValidationException("Invalid params", params.errors)
+        }
```
drodzewicz, line 53: _Thinking out loud..._
My lazy butt is wondering if we should/could/want to create `validationInterceptor` that handles checking if a command object has errors and throw an exception instead of writing this logic everywhere ourselves.

drodzewicz (approved):

awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -46,12 +47,15 @@ class ProductSupplierService {
...
         // Store added aliases to avoid duplicate alias exceptions for product and supplier
         // This could happen when params.searchTerm and e.g. sort by productCode/productName is applied
         Set<String> usedAliases = new HashSet<>()
 
-        return ProductSupplier.createCriteria().list(max: params.max, offset: params.offset) {
+        return ProductSupplier.createCriteria().list(params.getPaginationParams()) {
```
jmiranda, line 58: command.paginationParams

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -46,12 +47,15 @@ class ProductSupplierService {
     ProductSupplierDataService productSupplierGormService
 
 
-    List<ProductSupplier> getProductSuppliers(ProductSupplierListParams params) {
+    List<ProductSupplier> getProductSuppliers(ProductSupplierFilterCommand params) {
```
jmiranda, line 50: ProductSupplierFilterCommand command

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
jmiranda: I thought the same thing with respect to request.exception (maybe we could handle both cases in the same interceptor). However, something doesn't feel right about this approach so I'd prefer to explore it as part of a separate spike.

jmiranda (changes requested):
minor changes

awalkowiak (approved):

PR #4496, opened 2024-02-07
OBPIH-6024 Create a reusable TextInput component for Product Sources create/edit form

alannadolny:


src/js/wrappers/InputWrapper.jsx
```diff
@@ -0,0 +1,91 @@
...
+              )}
+            >
+              <RiQuestionLine />
+            </Tooltip>
+          )}
+          {required && <span className="input-wrapper-asterisk">&#42;</span>}
+        </div>
+        {redirect && (
+          <div
+            onClick={() => history.push(redirect.redirectTo)}
```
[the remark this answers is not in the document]
alannadolny: Good point, maybe a better approach (instead of passing an additional flag) will be to pass the whole "redirect" function, it will increase the usability of this button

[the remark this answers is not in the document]
alannadolny: image
image
image
image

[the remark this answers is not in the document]
kchelstowski: @alannadolny 

> will be to pass the whole "redirect" function, it will increase the usability of this button

I was also considering such an approach, so it'd be fine to me.

src/js/wrappers/InputWrapper.jsx
```diff
@@ -0,0 +1,86 @@
...
+              )}
+        >
+          <RiQuestionLine />
+        </Tooltip>
+        )}
+        {required && <span className="input-wrapper-asterisk">&#42;</span>}
+      </div>
+      {button && (
+      <div
+        onClick={button.function}
```
drodzewicz, line 38:
```suggestion
        onClick={button.onClick}
```
let's follow the react naming pattern for callback props that handle onClick functionality

src/js/wrappers/InputWrapper.jsx
```diff
@@ -0,0 +1,86 @@
...
+    defaultMessage: PropTypes.string.isRequired,
+  }),
+  // Indicator whether the red asterisk has to be shown
+  required: PropTypes.bool,
+  // Title displayed above the field
+  title: PropTypes.shape({
+    id: PropTypes.string.isRequired,
+    defaultMessage: PropTypes.string.isRequired,
+  }),
+  // Hyperlink on the right side above the input
```
drodzewicz, line 70: after your last change this comment doesn't correspond to the usecase of this prop.

src/js/wrappers/InputWrapper.jsx
```diff
@@ -0,0 +1,86 @@
...
+  // Message displayed under the input
+  errorMessage: PropTypes.string,
+};
+
+InputWrapper.defaultProps = {
+  tooltip: null,
+  required: false,
+  title: null,
+  button: null,
+  errorMessage: null,
```
drodzewicz, line 85: is there a reason we need to specify default prop values as `null`?
by default props that are not passed are `undefined`

src/js/wrappers/InputWrapper.jsx
```diff
@@ -0,0 +1,86 @@
...
+          <RiQuestionLine />
+        </Tooltip>
+        )}
+        {required && <span className="input-wrapper-asterisk">&#42;</span>}
+      </div>
+      {button && (
+      <div
+        onClick={button.function}
+        role="presentation"
+        className="input-wrapper-redirect"
```
drodzewicz, line 40: after your last change `input-wrapper-redirect` is probably not an accurate description of this div

kchelstowski (approved):

drodzewicz (changes requested):

src/js/wrappers/InputWrapper.jsx
alannadolny: If a prop is not required, the linter shouts that we have to specify the default prop.
> ESLint: propType "..." is not required, but has no corresponding defaultProps declaration.(react/require-default-props)

drodzewicz: ok

awalkowiak (approved):

drodzewicz (approved):

PR #4497, opened 2024-02-08
OBPIH-6042 Add filter by date range to expiration reports (fix after QA)

alannadolny:


grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -163,7 +163,7 @@ class DashboardService {
             }
 
             if (command.startDate) {
-                ge("expirationDate", command.startDate)
+                gt("expirationDate", command.startDate + 1)
```
awalkowiak, line 166: @alannadolny Does this require both changes? `ge` -> `gt` and adding `+ 1`?

alannadolny: Yup, when I just changed `ge` -> `gt` I got the same results, It's probably due to hours comparison, but now it works fine.

awalkowiak: @alannadolny Then I think you have to still have `ge` if you add `+ 1`, because you are missing records that are from the `startDate + 1`, but don't have hours added

alannadolny: :heavy_check_mark:

awalkowiak (approved):

PR #4500, opened 2024-02-08
OBPIH-6100 Create a reusable Date component

alannadolny:
Current styling of the date component:
image
image
image
image
image
image
image

### DIFF
--- a/src/js/components/form-elements/v2/DateFieldInput.jsx
+++ b/src/js/components/form-elements/v2/DateFieldInput.jsx
@@ -12,23 +12,31 @@ const DateFieldInput = forwardRef(({
   className,
   disabled,
   ...props
-}, ref) => (
-  <div
-    className={`${disabled ? 'disabled' : ''} ${className}`}
-    {...props}
-    ref={ref}
-    tabIndex="0"
-    role="button"
-  >
-    <span>{value || placeholder}</span>
-    {!disabled
-      && (
-      <span className="form-element-icons-wrapper">
-        {value ? <RiCloseLine onClick={onClear} /> : <RiCalendarLine />}
-      </span>
-      )}
-  </div>
-));
+}, ref) => {
+  const disabledProps = disabled ? {
+    className: `disabled ${className}`,
+  } : {
+    className,
+    tabIndex: 0,
+    role: 'button',
+  };
+
+  return (
+    <div
+      {...props}
+      {...disabledProps}
+      ref={ref}
+    >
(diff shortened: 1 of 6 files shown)


src/js/components/form-elements/v2/TextInput.jsx
```diff
@@ -25,7 +25,7 @@ const TextInput = ({
   >
     <input
       disabled={disabled}
-      className={`text-input ${errorMessage ? 'has-errors' : ''}`}
+      className={`form-element-input ${errorMessage ? 'has-errors' : ''}`}
       placeholder={placeholder}
       {...fieldProps}
```
drodzewicz, line 30: same here, be careful what you are overriding, the same example as I pointed out before applies here, if input will have errorMessages and className, the className prop will overwrite the 
```className={`form-element-input ${errorMessage ? 'has-errors' : ''}`}```

**Review by drodzewicz — CHANGES_REQUESTED** [review]
overall looks good, but please make sure that the prop spread is in the right place that you expect it to be and that it overrides props that you expect to override

File: src/js/components/form-elements/v2/DateFieldInput.jsx [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+  disabled,
+  ...props
+}, ref) => {
+  const disabledProps = disabled ? {
+    className: `disabled ${className}`,
+  } : {
+    className,
+    tabIndex: 0,
+    role: 'button',
+  };
```
drodzewicz, line 22: could you elaborate on this part?
if input is disabled we want to remove the `role: button` from the div wrapper and `tabIndex: 0`

Why do we even need these attributes here? is it for the ability to **Tab** click and navigate with a keyboard between inputs or is there something more to it?

alannadolny: Divs are not focusable by default, so thanks to the `tabIndex: 0` and `role: button` I can access the `:focus` on the div element. When the input is disabled and the tabIndex and role are still present this div still can get a focus and something weird happens when trying to get focus on other date input elements.https://assets.grammarly.com/emoji/v1/1f64c.svg

kchelstowski (approved):
I feel like something is missing in terms of integrating it with `react-hook-form`, but I can't see yet what potentially is missing :smile: 
so from my perspective let that one sink in and we'll potentially fix the issues offhand.

drodzewicz (approved):

PR #4501, opened 2024-02-09
OBPIH-6037 Get only available quantity bin location in expiration report exports

drodzewicz:


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -644,9 +649,9 @@ class ProductAvailabilityService {
 						from ProductAvailability pa
 						left outer join pa.inventoryItem ii
 						left outer join pa.binLocation bl
-						where pa.location = :location and pa.quantityOnHand > 0
-						group by pa.product, pa.inventoryItem, pa.binLocation
-						""", [location: location])
+						where pa.location = :location and pa.quantityOnHand > 0""" +
+                        "${inventoryItems ? " and pa.inventoryItem.id in (:inventoryItems) " : " "}" +
+						"""group by pa.product, pa.inventoryItem, pa.binLocation""", arguments)
```
kchelstowski, line 654: Have you checked if it works fine with/without inventoryItems?
The indents here look weird, and the `"""` at the beginning of the line, could you maybe paste a screenshot how it really looks in the IDE?

drodzewicz: image

Yes, I have tested without inventoryItems, this method is also currently used in returning putaway candidate items and works fine after the change.
 the inventoryItems parameter has a default value of an empty list and that's the reason for the conditionals in the SQL query and the arguments.
If we are not using passed arguments in the query, we get an exception that's why I needed to condition it.

kchelstowski (approved):

alannadolny (approved):

PR #4502, opened 2024-02-09
OBPIH-6103 Create endpoint to create/edit product source details

kchelstowski:


src/main/groovy/org/pih/warehouse/product/ProductSupplierDetailsCommand.groovy
```diff
@@ -0,0 +1,43 @@
...
+        code(nullable: true)
+        manufacturer(nullable: true)
+        ratingTypeCode(nullable: true)
+        description(nullable: true)
+        manufacturerCode(nullable: true)
+        brandName(nullable: true)
+        productCode(nullable: true)
+        active(nullable: true)
+    }
+}
```
kchelstowski, line 43: Since we are about to create 3 separate endpoints for now, I've decided to go with the command approach to make sure, that when using details endpoint, a user will potentially be able only to edit details part and that the user won't be able to include any other properties, e.g. throwing a list of `productSupplierPreferences` that would be persisted if I used the `ProductSupplier` domain as a command class.

cc @awalkowiak @alannadolny @drodzewicz @jmiranda

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -355,4 +357,36 @@ class ProductSupplierService {
...
+    }
+
+    ProductSupplier updateDetails(ProductSupplierDetailsCommand command, String productSupplierId) {
+        validateProductSupplierDetails(command)
+        ProductSupplier productSupplier = productSupplierGormService.get(productSupplierId)
+        if (!productSupplier) {
+            throw new ObjectNotFoundException(productSupplierId, ProductSupplier.class.toString())
+        }
+        productSupplier.properties = command.properties
+        // TODO: To be replaced by a generic method created in OBPIH-6017 after merging into develop
```
kchelstowski, line 385: as you can see, the if statement below is repeated here, and in the `saveDetails` method. I've implemented a method for that recently, while working on OBPIH-6017 (assignSourceCode): https://github.com/openboxes/openboxes/pull/4486
that we will be able to use here as soon as the feature branch is merged into develop (upgrade to grails 3.3.10) branch.

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
         productSupplierService.delete(params.id)
         render status: 204
     }
+
+    def create(ProductSupplierDetailsCommand productSupplierCommand) {
+        ProductSupplier productSupplier = productSupplierService.saveDetails(productSupplierCommand)
+
+        response.status = HttpStatus.CREATED.value()
```
drodzewicz, line 37: This looks hardcore, I think it is safe to use numbers for HTTP status codes :sweat_smile:

drodzewicz (approved):

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
drodzewicz: I think I will retract my comment....

The reason I said this is that I usually dealt with HTTP codes in a number format and they are the same across all frameworks, and importing the HttpsStatus and extracting that exact number using `HttpStatus.CREATED.value()` was something that caught my attention, but thinking about this right now I think I see a point in this, this is probably easier to read and debug 

soooo... statement rollback, this looks right :+1:

src/main/groovy/org/pih/warehouse/product/ProductSupplierDetailsCommand.groovy
awalkowiak: I wonder if we could somehow use shared constraints, like here: https://grails.github.io/legacy-grails-doc/3.3.0/ref/Constraints/Usage.html so we have a "single source of truth" regarding constraints, and won't have to change two places whenever constraints will change on the domain level.

awalkowiak (commented):
I only have a question / discussion around constraints about how we want to move forward regarding sharing constraints between command and domain

src/main/groovy/org/pih/warehouse/product/ProductSupplierDetailsCommand.groovy
kchelstowski: @awalkowiak thanks, this feels interesting. I have not known this was possible. I will check it out and get back to you.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
         productSupplierService.delete(params.id)
         render status: 204
     }
+
+    def create(ProductSupplierDetailsCommand productSupplierCommand) {
```
awalkowiak, line 34: Is the naming `productSupplierCommand` here used on purpose (without `Details`)? If not I think it should match the command name

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
...
+        ProductSupplier productSupplier = productSupplierService.saveDetails(productSupplierCommand)
```
awalkowiak, line 35: I wonder if the word `create` would not be more suitable here instead of `save`, but I'll let @jmiranda decide what he prefers here. Other than that and one minor naming thing above it LGTM.

awalkowiak (commented):
Overall LGTM, just one minor thing that I'd like to hear @jmiranda's opinion.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
kchelstowski: you are right, this is inconsistency from my side, so answering - it was not intentional.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
kchelstowski: to include my opinion - I'd stick with `save`, since we are actually calling the db and saving the entity (and GORM's method is named `save`, not `create`, so to kinda follow the convention)
But I don't mind both, so let me know what you guys prefer.
I can also see that Grails' docs probably follow `save`/`update` pattern: https://docs.grails.org/latest/guide/REST.html

awalkowiak: Looking into this docs, it also suggests calling the action `save` if you do POST to the `/productSuppliers`, so then if you'd like to follow it, then you need to change the action name to `save`.

awalkowiak: Sounds good.

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -355,4 +357,36 @@ class ProductSupplierService {
...
     }
+
+    private static void validateProductSupplierDetails(ProductSupplierDetailsCommand command) {
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid product source", command.errors)
+        }
+    }
+
+    ProductSupplier saveDetails(ProductSupplierDetailsCommand command) {
+        validateProductSupplierDetails(command)
```
jmiranda, line 368:
```
ProductSupplier productSupplier = new ProductSupplier(command.properties)
if (command.hasErrors()) { 
    productSupplier.errors = command.errors
    return productSupplier
}

// do some stuff

if (!productSupplier.hasErrors()) { 
    productSupplier = productSupplierGormService.save(productSupplier)
}
return productSupplier
```

jmiranda: * should service throw exception 
* static vs compilestatic
    * @GrailsCompileStatic
    * @CompileStatic

awalkowiak: @jmiranda To add something to here - I feel like we should focus on building a proper rest API, and stick to one approach or convention (in this case I think we should stick to throwing exception or error and handling it in error controller instead returning products supplier object with errors copied from command), and not mix into it different things to "satisfy" other potential use cases (calls from other services or non API controllers)

cc @kchelstowski, @drodzewicz @alannadolny

src/main/groovy/org/pih/warehouse/product/ProductSupplierDetailsCommand.groovy
jmiranda: I found an even worse issue than what @kchelstowski was worried about. And we are definitely susceptible to this given that none of our domain or command constraints are marked as final.

https://grails.org/blog/2022-09-28-static-binding-bug.html

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
         productSupplierService.delete(params.id)
         render status: 204
     }
+
+    def create(ProductSupplierDetailsCommand productSupplierDetailsCommand) {
+        ProductSupplier productSupplier = productSupplierService.saveDetails(productSupplierDetailsCommand)
+
+        response.status = HttpStatus.CREATED.value()
+        render([data: productSupplier.toJson()] as JSON)
```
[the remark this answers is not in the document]
jmiranda: You'll also need to add one for ProductSupplierPreference.
```
        JSON.registerObjectMarshaller(ProductSupplierPreference) { ProductSupplierPreference productSupplierPreference ->
            return productSupplierPreference.toJson()
        }
```


And the supplier object in the ProductSupplier.toJson() is missing an ID, which initially prevented me from copying the JSON returned by a GET request and sending it in a PUT request.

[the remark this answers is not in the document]
jmiranda: Here's my attempt at the create endpoint. The validation exception should have a better message, but I hate that I need to pass a message there.
```
    def create(ProductSupplierDetailsCommand command) {
        if (command.hasErrors()) {
            throw new ValidationException("validation errors", command.errors)
        }

        ProductSupplier productSupplier = productSupplierService.saveProductSupplier(params.id, command)

        response.status = HttpStatus.CREATED.value()

        render([data: productSupplier] as JSON)
    }
```

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
...
+    }
+
+    def update(ProductSupplierDetailsCommand productSupplierDetailsCommand) {
+        ProductSupplier updatedProductSupplier = productSupplierService.updateDetails(productSupplierDetailsCommand, params.id)
```
jmiranda, line 42: And here's the update endpoint.
```
def update(ProductSupplierCommand command) {
    if (command.hasErrors()) {
        throw new ValidationException("validation errors", command.errors)
    }
    ProductSupplier productSupplier = productSupplierService.saveProductSupplier(params.id, command)

    render([data: productSupplier] as JSON)
}
```

jmiranda: I still don't like the params.id as an argument to this method, but understand why it was done that way.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -28,4 +30,17 @@ class ProductSupplierApiController {
...
+
+        render([data: updatedProductSupplier.toJson()] as JSON)
```
jmiranda, line 44: remove toJson

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -355,4 +357,36 @@ class ProductSupplierService {
...
         productSupplierGormService.delete(productSupplierId)
     }
+
+    private static void validateProductSupplierDetails(ProductSupplierDetailsCommand command) {
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid product source", command.errors)
+        }
+    }
+
+    ProductSupplier saveDetails(ProductSupplierDetailsCommand command) {
```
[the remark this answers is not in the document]
jmiranda: Here's my attempt at the generic save (used for create or update). 
```
ProductSupplier saveProductSupplier(String id, ProductSupplierCommand command) {
    ProductSupplier productSupplier
    if (id == null) {
        productSupplier = new ProductSupplier(command.properties)
    }
    else {
        productSupplier = productSupplierGormService.get(id)
        if (!productSupplier) {
            throw new ObjectNotFoundException(id, ProductSupplier.class.toString())
        }
        productSupplier.properties = command.properties
    }

    if (!productSupplier.code) {
        productSupplier.code =
            identifierService.generateProductSupplierIdentifier(command?.product?.productCode, command?.supplier?.code)
    }
    return productSupplierGormService.save(productSupplier)
}
```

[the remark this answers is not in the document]
jmiranda: Note: I am not handling the validation on the product supplier domain object. If there's a desire to throw a validationexception from the service, you'll need to add some code here to handle it.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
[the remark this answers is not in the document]
jmiranda: The reason I would prefer to use saveProductSupplier over saveDetails or save is because the ProductSupplierService will probably also house the code to save other objects like preferences and packages that are associated with a product supplier.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
jmiranda: Regarding the response for create and update, it is possible that we return the object as-is, but I am still working out the correct "response" for these endpoints based on best practices. There are a lot of good arguments for both sides.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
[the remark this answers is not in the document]
kchelstowski: Is there a deeper reason why you prefer Bootstrap over `toJson()` here? I think recently we've discussed that we wanted to avoid using Bootstrap and stick to `toJson()` for now even if it's not the best approach possible.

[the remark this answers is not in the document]
kchelstowski: The reason I went with `saveDetails` is for this method not to be confusing - doing it as `saveProductSupplier` in my opinion suggests, that we could update all product supplier's properties there - this is why I haven't named it like this.
But if you guys are ok with that, then I can change it.

kchelstowski:
@jmiranda 

> And fwiw, the danger seems to exist whether I use a command or domain class since the data binding eventually occurs on the domain object. We can certainly limit the reach with a command class, but we can still do things we might not want to allow.


Yeah, this is why I've mentioned many times that I am not fan of relying on domain classes for binding or inside a command class, because such "weirds" dangerous things could happen if we rely on domain classes, not on POJOs which we have control of most often with third party libraries like MapStruct which are very flexible and configurable for any use case.

So I am not surprised that we can edit the Product and Organization, because those are existing domain object that are bound to the transaction/session and any changes on them would be tracked.
An easy way to prevent from updating them would be to either create some generic method, or just write those two lines manually after binding to the product suppleier domain:
```groovy
productSupplier.properties = command.properties
productSupplier.product?.discard()
productSupplier.supplier?.discard()
```
and none changes would be persisted to product and supplier :slightly_smiling_face: 

I wouldn't agree that this makes no difference between using the command class and domain class for binding, because what I meant by "updating some additional properties" I meant rather those, that are not meant to be touched during the saving details process, e.g. binding `productSupplierPreferences`, `productPackages`, which would be possible if we used domain class, and is not possible with the command I've created.

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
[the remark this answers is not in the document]
jmiranda: Regarding your first comment, we should continue to use the JSON marshaller framework (including marshaller registration in BootStrap.groovy) until we have migrated to JSON views.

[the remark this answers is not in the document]
jmiranda: Regarding the second one, I understand your point. We just have a convention of using the name action<DomainClass>, like saveProduct, saveProductSupplier. 

We haven't really discussed having method names that imply scope. And I would say that the scope should be explicitly defined through better security measures, not though naming. But if we need 

If we want to adopt a new naming convention we should create a proposal on the new naming convention and agree on it before applying it. 

Here's an example of what a naming convention might look like. I don't like all of these convention, but I like the general idea of defining each convention, providing some justification and demonstration through a code example.
https://sergiodelamo.com/blog/grails-extra-conventions.html

src/main/groovy/org/pih/warehouse/product/ProductSupplierDetailsCommand.groovy
jmiranda: @awalkowiak Regarding your comment about shared constaints, which class should be the single source of truth.

awalkowiak: I'm guessing the domain in this case, and then we could override it if needed.

jmiranda:
@kchelstowski @awalkowiak i'll try to go through each discussion and determine whether it should be dealt with now or later.

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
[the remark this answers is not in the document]
jmiranda: @awalkowiak @kchelstowski I'll let you decide how you want to handle this 
* keep separate save and update methods
* use a single save method

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
jmiranda: I'd like to keep the code below if possible i.e. revert the assignSourceCode. I like the idea of abstracting the generation and assignment, but I'm not yet comfortable with how it's done in #4486 so we can postpone that for now.

jmiranda:
> An easy way to prevent from updating them would be to either create some generic method, or just write those two lines manually after binding to the product suppleier domain:
```
productSupplier.properties = command.properties
productSupplier.product?.discard()
productSupplier.supplier?.discard()
```
I understand your point, but this is not a sustainable approach. You need to do this everywhere you're using the "productSupplier.properties = ... " data binding mechanism. And if you add a new association to ProductSupplier, you need to remember to go back and discard the association. 

We either need to make this "discard" mechanism to be more automatic (i.e. annotation on the service method?) OR better yet we need the protection to happen during data binding phase in the controller. Either a feature within the @BindUsing annotation or explicit whitelisting properties (i.e. bindData(object, includes: [])) the client is allowed to update for a given service method. Or some other mechanism.
```
bindData(productSupplier, command.properties, [include: ['id', 'code', 'name', ...]])
```
We could potentially make this more palatable and DRY by create a command class interface to requires us to expose the properties to include and exclude via the command class.
```
bindData(productSupplier, command.properties, [include: command.includeProperties, exclude: command.excludeProperties])
```
> So I am not surprised that we can edit the Product and Organization, because those are existing domain object that are bound to the transaction/session and any changes on them would be tracked.

The point was that if you're binding objects to a domain or command, you can edit the properties of any object in their graph linked to that object. In order for command objects to be used to prevent this then your command object needs to bind only primitives and Strings (i.e. the PK of the object rather than the object itself).
```
String productId
String supplierId
``` 
Which, in my mind, defeats the purpose of using a command object for data binding. So again, there needs to be another way to protect against deep data binding. 
```
// This is a hypothetical for what i'm looking for
@BindUsing(onlyBindPrimaryKey=true)
Product product
```
However, I'm actually wondering whether the use of primitives + String in the command class would actually be all we need. 
```
productSupplier.properties = command.properties
```
It would be interesting to see if this approach yields what we're looking to do. It doesn't provide all of the validation we need at the command object level (i.e. we can have a productId in the command, but it might not actually map to a Product in the database), but that would be caught when we try to validate and save the domain. 

Anyway, that would be fine with me for now.

kchelstowski:
@jmiranda I throw one more possible option to avoid updating nested properties, e.g. we can just define a proper validator in the command's constraints and see if a property (let's say - product) is dirty:
```groovy
 static constraints = {
      ...
      product(validator: { Product product ->
          if (product?.dirty) {
              return ['validator.product.cantBeDirty']
          }
          return true
      })
  }

**Comment by jmiranda** [comment]
Ooh dang. That's interesting. It feels a bit heavy-handed (a slap instead of an ignore) and it requires maintainance for whenever we add new associations, but I think it's an interesting avenue to explore, especially if we can make it a little more reusable. 

ProductSupplierCommand.groovy
```
static constraints = { 
    product: shared: 'myFavoriteLittleDirtyConstraint'
    supplier: shared: 'myFavoriteLittleDirtyConstraint'
}
```

runtime.groovy
```
grails.gorm.default.constraints = {
  myFavoriteLittleDirtyConstraint = { value, object ->
     if (value?.dirty) {
         // TODO exercise for reader: return a specific error for each class (if it exists), generic error message if not
         return ['validator.myFavoriteLittleDirtyConstraint.message']
     }
     return true
  }
}
```

> but had no success in that.

Can you clarify what you meant by that? The product wasn't bound at all or wasn't read-only and still bound properties we didn't want? Remember the "read" doesn't make the object read-only as-in unmodifiable, it just disables the dirty detection and presumably automatic flush at session end.

Source: https://docs.grails.org/latest/ref/Domain%20Classes/read.html

However, this is the path I'd like to explore a little more. I'm guessing we can do some interesting things with the BindUsing annotations.
```
@BindUsing({ obj, source -> Product.read(source["id"])})
Product product
```
If not, we can explore creating our own annotation or extend the existing one to do what we want i.e. discard object.

kchelstowski:
> Can you clarify what you meant by that? The product wasn't bound at all or wasn't read-only and still bound properties we didn't want? 

Was not bound at all

kchelstowski:
@awalkowiak @jmiranda I've addressed suggestions marked with " :+1:  "

awalkowiak (approved):
I think all of the change requests were resolved. LGTM

PR #4505, opened 2024-02-13
OBGM-757 Modify 0.8.x README.md to redirect to current 0.9.x development

awalkowiak:


awalkowiak:
I am merging as is, this can be modified later according to the needs

PR #4508, opened 2024-02-16
OBPIH-6104 Add inputs and form logic for product sources details

kchelstowski:
Screenshot from 2024-02-16 11-04-28


src/js/components/productSupplier/create/BasicDetails.jsx
alannadolny, line 1: I think it is worth putting those subsections in a separate directory to differentiate them from other create-form components (the navigation will be easier)
I mean sth like: /create/section/BasicDetails.jsx

kchelstowski: sounds good

alannadolny (approved):

src/js/components/form-elements/v2/DateField.jsx
```diff
@@ -54,6 +50,7 @@ const DateField = ({
         utcOffset={0}
         placeholderText={placeholder}
         {...fieldProps}
+        onChange={(val) => fieldProps?.onChange?.(val?.format(DateFormat.MM_DD_YYYY))}
```
alannadolny, line 53: use the newly added format `DateFormat.MMM_DD_YYYY`

src/js/components/productSupplier/create/ProductSupplierFormMain.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import DetailsSection from 'components/productSupplier/create/sections/DetailsSection';
+
+import 'components/form-elements/DateFilter/DateFilter.scss';
+import 'react-datepicker/dist/react-datepicker.css';
```
drodzewicz, line 8: these imports should probably be in the datepicker component

src/js/components/productSupplier/create/subsections/BasicDetails.jsx
```diff
@@ -0,0 +1,197 @@
...
+          render={({ field }) => (
+            <TextInput
+              title={{ id: 'react.productSupplier.form.legacyCode.title', defaultMessage: 'Legacy Code' }}
+              tooltip={{
+                id: 'react.productSupplier.form.legacyCode.tooltip',
+                defaultMessage: 'Reference to this record in a previous or parallel purchasing system',
+              }}
+              {...field}
+            />
+          )}
```
drodzewicz, line 84: Do you think we should move these `<Controller name="" render {...fields}` into `TextInput`, `SelectField`, `DateField` components?
The instead of 
```jsx
<Controller
          name="legacyCode"
          control={control}
          render={({ field }) => (
            <TextInput
              title={{ id: 'react.productSupplier.form.legacyCode.title', defaultMessage: 'Legacy Code' }}
              tooltip={{
                id: 'react.productSupplier.form.legacyCode.tooltip',
                defaultMessage: 'Reference to this record in a previous or parallel purchasing system',
              }}
              {...field}
            />
          )}
```

we would do
```jsx
          <TextInput
             name="legacyCode"
            control={control}
            title={{ id: 'react.productSupplier.form.legacyCode.title', defaultMessage: 'Legacy Code' }}
            tooltip={{
              id: 'react.productSupplier.form.legacyCode.tooltip',
              defaultMessage: 'Reference to this record in a previous or parallel purchasing system',
            }}

          />
```
cc @alannadolny @kchelstowski

src/js/hooks/productSupplier/form/useProductSupplierForm.js
```diff
@@ -0,0 +1,150 @@
...
+        value: z.string(),
+        label: z.string(),
+      }).optional(),
+      manufacturerCode: z
+        .string()
+        .optional(),
+      brandName: z
+        .string()
+        .optional(),
+    });
```
drodzewicz, line 58: What do you think about moving this schema into a separate file instead of having it in a hook? 
cc @alannadolny

drodzewicz (changes requested):
My only real request change is moving imprt of CSS file for datepciker into datepicker component, other things are rather questions and future improvement like moving the Controller into Field components (maybe there is case to be made why we should not do it) and moving the validation schema in its own separate file - I think this will be much easier to maintain in the future but waiting for your opion

src/js/components/productSupplier/create/subsections/BasicDetails.jsx
kchelstowski: I get the point, but if it was up to me, I’d suggest sticking to calling the <Controller> with render for now, until we get used to the library and until we know any potential use cases we might have with the new inputs. Using <Controller> + render is not the only way to use the react-hook-form, hence if we added it to the inputs, any other approach would be impossible to use, or if we wanted to use input without the form, because it would always expect to receive the control prop.

src/js/hooks/productSupplier/form/useProductSupplierForm.js
kchelstowski: Sounds good and I was even thinking about it, but didn’t see that much value of it, but if you guys prefer it, I will move it to a separate file, since it doesn’t depend on React lifecycle.

src/js/components/productSupplier/create/subsections/BasicDetails.jsx
drodzewicz: That sounds good, I see what you mean.
Let's not try coming up with "clever abstract" ideas and stick with simple implementation for now, in the future this can easily be improved with more cases

src/js/hooks/productSupplier/form/useProductSupplierForm.js
drodzewicz: if you can then please move it, I think this will improve readability of this schema

src/js/components/productSupplier/create/subsections/BasicDetails.jsx
alannadolny: I am not sure about that, because of that we will be limited to the use of the `Controller` so, in a case when we would like to use our inputs without the form library it can cause some issues. I would leave it as it is.

src/js/hooks/productSupplier/form/useProductSupplierForm.js
alannadolny: That sounds good, but it doesn't look wrong to me at this moment.

awalkowiak: Plus these should be translatable, but that can be left for later.

kchelstowski: oh, yeah, I again forgot about it - so this changes the situation - to have them translatable, I need to use it in the hook, hence I'd need `translate()` method from the `useDispatch()`.

kchelstowski: Unless I'd suggest creating a separate hook for the validation schema -> `useProductSupplierValidation()` and put there the validation schema.
but I also agree let's do it "by the way"

drodzewicz (approved):

PR #4509, opened 2024-02-16
OBPIH-6052 validation for zero items to invoice in invoices workflow (post QA)

drodzewicz:
There were missing places where the same validation modal had to appear on invoice creation workflow like:
- When clicking the previous button
- when saving
- when saving & exiting

### DIFF
--- a/src/js/components/invoice/AddItemsPage.jsx
+++ b/src/js/components/invoice/AddItemsPage.jsx
@@ -53,11 +53,10 @@ const FIELDS = {
     loadMoreRows: ({ loadMoreRows }) => loadMoreRows(),
     isFirstPageLoaded: ({ isFirstPageLoaded }) => isFirstPageLoaded,
     // eslint-disable-next-line react/prop-types
-    addButton: ({ values, loadMoreRows, saveInvoiceItems }) => (
+    addButton: ({ values, loadMoreRows }) => (
       <InvoiceItemsModal
         btnOpenText="react.default.button.addLines.label"
         btnOpenDefaultText="Add lines"
-        onOpen={() => saveInvoiceItems(values)}
         invoiceId={values.id}
         onResponse={loadMoreRows}
       >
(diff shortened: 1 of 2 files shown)


src/js/components/invoice/AddItemsPage.jsx
```diff
@@ -53,11 +53,10 @@ const FIELDS = {
     loadMoreRows: ({ loadMoreRows }) => loadMoreRows(),
     isFirstPageLoaded: ({ isFirstPageLoaded }) => isFirstPageLoaded,
     // eslint-disable-next-line react/prop-types
-    addButton: ({ values, loadMoreRows, saveInvoiceItems }) => (
+    addButton: ({ values, loadMoreRows }) => (
       <InvoiceItemsModal
         btnOpenText="react.default.button.addLines.label"
         btnOpenDefaultText="Add lines"
-        onOpen={() => saveInvoiceItems(values)}
```
drodzewicz, line 60: This onOpen callback would fire a saveItems post request after we opened the dialog for invoice item candidates.
I was looking through PR's and tickets to find the reason behind this logic implementation and was not able to find anything.
It was implemented as part of OBPIH-3614.
To me this looks like a side effect that users might not expect especially with the new logic where we remove rows with 0 quantity.

If I have to leave this callback it would result in following behavior:
- user ads items
- user updates some items quantity to 0
- user opens dialog
-  save items request is sent 
- rows with 0 items disappear
result: user is confused about what just happened because he did;t click save or anything of sorts.

This looks like a weird side effect that should not happen when opening a dialog and I was not able to figure out any use cases for its existence.
So correct me if I am worng

drodzewicz: ok I can bring it back but in my opinion this will be not a pleasant experience for a user.
This looks like a side effect that user might not want to happen.
I will mention this case in the ticket and will want to hear an opinion after UAT.
I would suggest revisitng this case and coming up with some other mechanism or cases, but as part of a different ticket.

drodzewicz:
@awalkowiak I brought back this behavior of saving current items before displaying candidates.
Additionally I had to implement some logic to wait for user confirmation "that zero items will be deleted" and only then load the modal which is why I implemented new prop for rendering custom button through renderProps.

PR #4510, opened 2024-02-20
OBPIH-6109 Implement preference type variations subsection with the table

alannadolny:


src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -0,0 +1,90 @@
...
+              validityEndDate: '',
+              bidName: '',
+            })}
+            StartIcon={<RiAddLine className="button-add-icon" />}
+            defaultLabel="Add new"
+            label="Add new"
+          />
+        </div>
+        <DataTable
+          data={fields}
```
awalkowiak, line 52: I think it should be `rows`

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -0,0 +1,90 @@
...
+          columns={columns}
+          defaultPageSize={4}
+          pageSize={fields.length <= 4 ? 4 : fields.length}
```
awalkowiak, line 55: So if there will be 1000 rows, then all will be displayed at once?

src/js/consts/timeFormat.js
```diff
@@ -4,5 +4,5 @@ export const TimeFormat = {
 
 export const DateFormat = {
   MM_DD_YYYY: 'MM/DD/YYYY',
-  MMM_DD_YYYY: 'MMM DD, yyyy',
+  LL: 'll',
```
awalkowiak, line 7: What's wrong with the previous version? If you'd like to add `LL` (equivalent of `MMMM DD, yyyy`) which is also a valid case how would the key look like?

src/js/components/productSupplier/create/InvalidItemsIndicator.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { RiCheckboxCircleLine } from 'react-icons/all';
+
+import Translate from 'utils/Translate';
+
+const InvalidItemsIndicator = ({ className, errorsCounter }) => (
+  <div className={`invalid-items-indicator ${className}`}>
+    <span>
```
awalkowiak, line 10: This should be a button that will filter rows in the table

awalkowiak (changes requested):

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
alannadolny: yeah, and it will be scrollable

src/js/components/productSupplier/create/InvalidItemsIndicator.jsx
alannadolny: this button/indicator is handled in another ticket, I took it as the next one that I am working on

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -0,0 +1,90 @@
...
+import { RiAddLine } from 'react-icons/ri';
+
+import DataTable from 'components/DataTable';
+import Button from 'components/form-elements/Button';
+import Subsection from 'components/Layout/v2/Subsection';
+import InvalidItemsIndicator from 'components/productSupplier/create/InvalidItemsIndicator';
+import usePreferenceTypeVariationsColumns
+  from 'hooks/productSupplier/form/usePreferenceTypeVariationsColumns';
+
+const PreferenceTypeVariations = ({ control, errors: { productSupplierPreferences } }) => {
```
kchelstowski, line 15: I don't like the naming (productSuppllierPreferences). Let's stick to errors or at least pass it to the hook like:
```js
usePreferenceTypeVariationsColumns({
    errors: productSupplierPreferences,
    control,
    remove,
  });
```
and use it in the hook via `errors`.

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -0,0 +1,90 @@
...
+      <div className="preference-type-variations-subsection">
+        <div className="d-flex justify-content-end align-items-center mb-3">
+          <InvalidItemsIndicator className="mr-3" errorsCounter={_.filter(productSupplierPreferences)?.length} />
+          <Button
+            onClick={() => append({
+              destinationParty: {},
+              preferenceType: {},
+              validityStartDate: '',
+              validityEndDate: '',
+              bidName: '',
```
kchelstowski, line 44: let's maybe store those in a const to a better readability?

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -0,0 +1,90 @@
...
+            })}
+            StartIcon={<RiAddLine className="button-add-icon" />}
+            defaultLabel="Add new"
+            label="Add new"
```
kchelstowski, line 48: shouldn't it be a translation key?

src/js/consts/timeFormat.js
kchelstowski: I don't have any suggestion at the moment, I would assume that we won't use the "LL" in the near future :thinking: or just go with a more intuitive name (I haven't known what LL is tbh), so let's maybe go with convention of naming keys what they really display?

kchelstowski (changes requested):

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
awalkowiak: But IMHO it still should be paginated but loaded dynamically

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
awalkowiak: Ugh... I think it is misleading, but ok, if this is how it's named there, then fine.

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -12,18 +12,26 @@ import InvalidItemsIndicator from 'components/productSupplier/create/InvalidItem
...
 
-  const { columns } = usePreferenceTypeVariationsColumns({
-    productSupplierPreferences,
+  const { columns, translate } = usePreferenceTypeVariationsColumns({
+    errors,
     control,
     remove,
   });
 
+  const defualtTableRow = {
```
kchelstowski, line 27: `defualt` -> `default`

src/js/components/productSupplier/create/subsections/PreferenceTypeVariations.jsx
```diff
@@ -34,18 +42,12 @@ const PreferenceTypeVariations = ({ control, errors: { productSupplierPreference
...
+          <InvalidItemsIndicator className="mr-3" errorsCounter={_.filter(errors)?.length} />
           <Button
-            onClick={() => append({
-              destinationParty: {},
-              preferenceType: {},
-              validityStartDate: '',
-              validityEndDate: '',
-              bidName: '',
-            })}
+            onClick={() => append(defualtTableRow)}
```
kchelstowski, line 47: `defualt -> default`

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
```diff
@@ -150,9 +150,9 @@ const usePreferenceTypeVariationsColumns = ({ productSupplierPreferences, contro
         />
       ),
     },
-  ], [productSupplierPreferences]);
+  ], [errors]);
```
kchelstowski, line 153: include the dependencies I've mentioned in the previous comment

kchelstowski (changes requested):

kchelstowski (approved):

awalkowiak (approved):

PR #4511, opened 2024-02-20
OBPIH-6116 Create endpoint for uom options

kchelstowski:


grails-app/controllers/org/pih/warehouse/api/UnitOfMeasureApiController.groovy
alannadolny, line 1: Shouldn't it be in `SelectOptionsApiController`? :thinking:

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -446,6 +446,11 @@ class UrlMappings {
             action = [GET:"currencies"]
         }
 
+        "/api/unitOfMeasure/options" {
```
awalkowiak, line 449: I am not sure why the previous one is in a singular version, but we should stick to plurals I think `/api/unitOfMeasures`. Perhaps we could add a second plural version and mark somehow the previous one as deprecated.

awalkowiak (approved):

PR #4514, opened 2024-02-21
OBPIH-6102 Create reusable checkbox  form component

drodzewicz:
There are two modifications to the input wrapper
 first, since this checkbox component has the label on it's right side I added a _modified_ to the InputWrapper to have the ability to position this label top, bottom, left and right so it would be much more reusable to future form components.

I have also added a `<label />` tag and attributes `htmlFor` and `ids` for better semantics. This will help us target these input elements for automated testing so this change is required.

### DIFF
--- a/src/js/components/form-elements/v2/Checkbox.jsx
+++ b/src/js/components/form-elements/v2/Checkbox.jsx
@@ -6,14 +6,13 @@ import InputWrapper from 'wrappers/InputWrapper';
 
 import './style.scss';
 
-const TextInput = ({
+const CheckBox = ({
   title,
   tooltip,
-  required,
   button,
   disabled,
   errorMessage,
-  placeholder,
+  labelPosition,
   id,
   name,
   ...fieldProps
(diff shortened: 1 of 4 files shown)


src/js/components/form-elements/v2/Checkbox.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import InputWrapper from 'wrappers/InputWrapper';
+
+import './style.scss';
+
+const CheckBox = ({
```
kchelstowski, line 9: the file is named "Checkbox", while the component is named "CheckBox" - was it meant to be?

src/js/components/form-elements/v2/Checkbox.jsx
```diff
@@ -0,0 +1,82 @@
...
+}) => (
+  <InputWrapper
+    title={title}
+    tooltip={tooltip}
+    button={button}
+    errorMessage={errorMessage}
+    inputId={id || name}
+    labelPosition={labelPosition}
+  >
+    <div className="form-element-checkbox ">
```
kchelstowski, line 28: nitpicky: not necessary space at the end

src/js/components/form-elements/v2/TextInput.jsx
```diff
@@ -14,16 +14,21 @@ const TextInput = ({
...
     title={title}
     tooltip={tooltip}
     required={required}
     button={button}
+    inputId={id || name}
     errorMessage={errorMessage}
   >
     <input
+      id={id || name}
+      name={name}
```
kchelstowski, line 31: this anyway comes from the `...fieldProps` so I assume you only extracted this for a better readability?

src/js/wrappers/style.scss
```diff
@@ -1,4 +1,6 @@
 .input-wrapper-container {
+  display: flex;
+  flex-direction: column;
```
kchelstowski, line 3: this feels somewhat dangerous - doesn't it break the positioning of already existing fields?

src/js/wrappers/style.scss
```diff
@@ -32,6 +37,40 @@
...
   .input-wrapper-error-message {
+    order: 2;
     color: var(--color-red);
   }
 }
+
+.input-wrapper-label-position {
+  &-top .input-wrapper-title {
+    order: 0;
+  }
```
kchelstowski, line 48: nitpicky: there should be gaps between the definitions

src/js/components/form-elements/v2/TextInput.jsx
drodzewicz: I extracted `id` and `name` to also pass it to InputWrapper component, we are doing the filedProps spread only on input

src/js/wrappers/style.scss
drodzewicz: when adding all these new styles I specifically tested all other fields and everything should be fine, do you have any specific concerns why you think it would be dangerous?

PR #4516, opened 2024-02-22
OBPIH-6116 Refactor uom options endpoint to be able to receive type p…

kchelstowski:
…aram, which filters by uom class

I decided not to get rid of the currencies endpoint in case somebody uses this API.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -446,11 +446,6 @@ class UrlMappings {
             action = [GET:"currencies"]
         }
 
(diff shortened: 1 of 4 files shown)


drodzewicz (approved):

alannadolny (approved):

awalkowiak (changes requested):
Assuming you can now pass UnitOfMeasureType in the parameters, then the second currency endpoint you added is not required anymore. Please remove it (but leave the "deprecated" one)

awalkowiak (changes requested):
You need to also include a change on the front end

awalkowiak (approved):

PR #4517, opened 2024-02-23
OBPIH-6111 Add pricing section with inputs

drodzewicz:


src/js/api/services/UnitOfMeasureApi.js
```diff
@@ -1,6 +1,9 @@
-import { CURRENCIES_OPTIONS } from 'api/urls';
+import { CURRENCIES_OPTIONS, UNIT_OF_MEASURE_OPTIONS } from 'api/urls';
 import apiClient from 'utils/apiClient';
 
 export default {
   getCurrenciesOptions: () => apiClient.get(CURRENCIES_OPTIONS),
+  getUnitOfMeasureOptions: (type) => apiClient.get(UNIT_OF_MEASURE_OPTIONS, {
```
kchelstowski, line 6: shouldn't it follow our current convention with passing `config`, as `params` is not the only property we can pass to axios?

src/js/components/productSupplier/create/subsections/FixedPrice.jsx
```diff
@@ -0,0 +1,91 @@
...
+  <Subsection
+    title={{
+      label: 'react.productSupplier.form.subsection.fixedPrice',
+      defaultMessage: 'Fixed Price',
+    }}
+    collapsable={false}
+  >
+    <div className="form-grid-3">
+      <Controller
+        name="contractPriceEach"
```
kchelstowski, line 21: let's name it as `contractPricePrice`, because this is how currently it is named on the backend logic for import product sources. 
there is an association in `ProductSupplier`:
`ProductPrice contractPrice`, so I think that `contractPriceEach` would not be intuitive in this case

src/js/components/productSupplier/create/subsections/FixedPrice.jsx
```diff
@@ -0,0 +1,91 @@
...
+            }}
+            tooltip={{
+              id: 'react.productSupplier.form.contractPriceEach.tooltip',
+              defaultMessage: 'Fixed price per unit guaranteed by a contract with the supplier',
+            }}
+          />
+        )}
+      />
+      <Controller
+        name="priceValidUntil"
```
kchelstowski, line 41: let's name it as `contractPriceValidUntil`, as it is pointing to the `contractPrice` of `ProductSupplier`.

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
```diff
@@ -0,0 +1,159 @@
...
+import { fetchQuantityUoM } from 'actions';
+import SelectField from 'components/form-elements/v2/SelectField';
+import TextInput from 'components/form-elements/v2/TextInput';
+import Subsection from 'components/Layout/v2/Subsection';
+
+const PackageSpecification = ({ control, errors }) => {
+  const dispatch = useDispatch();
+
+  const {
+    quantityUoM,
```
kchelstowski, line 16: nitpicky, but let's name it as `quantityUom` and change the `UoMType` to `UomType`, as imho it's more JS' way of naming things.

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
```diff
@@ -0,0 +1,159 @@
...
+  } = useSelector(state => ({
+    quantityUoM: state.unitOfMeasure.quantity,
+  }));
+
+  useEffect(() => {
+    dispatch(fetchQuantityUoM());
+  }, []);
```
kchelstowski, line 23: I think lines 13-23 should belong to a custom hook, maybe `usePackageSpecification.js` and just return `quantityUom` from it.

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
```diff
@@ -0,0 +1,159 @@
...
+                id: 'react.productSupplier.form.uomCode.tooltip',
+                defaultMessage: 'The most common package purchased for this product',
+              }}
+              options={quantityUoM}
+              errorMessage={errors.uomCode?.message}
+            />
+          )}
+        />
+        <Controller
+          name="packageSize"
```
kchelstowski, line 55: let's name it as `productPackageQuantity`

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
```diff
@@ -0,0 +1,159 @@
...
+              }}
+              tooltip={{
+                id: 'react.productSupplier.form.packageSize.tooltip',
+                defaultMessage: 'The number of units per package',
+              }}
+            />
+          )}
+        />
+        <Controller
+          name="minimumOrderQuantity"
```
kchelstowski, line 75: let's name it s `minOrderQuantity`, as the field on the `ProductSupplier` domain is named like this.

src/js/consts/UoMType.js
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line import/prefer-default-export
```
kchelstowski, line 1: how about doing:
```js
const UomType = {...}

export default UomType;
```
so that you wouldn't have to disable the eslint?

kchelstowski: Moreover I'm starting to think, that maybe we should align the naming with the backend's enum, so to use a full name: `UnitOfMeasureType`. What do you think?

src/js/hooks/productSupplier/form/useProductSupplierForm.js
```diff
@@ -111,6 +121,19 @@ const useProductSupplierForm = () => {
...
+
+  // eachPrice is a computed value from packagePrice and packageSize
+  useEffect(() => {
+    if (packagePrice && packageSize) {
+      setValue('eachPrice', decimalParser(packagePrice / packageSize, 4));
+    } else {
+      setValue('eachPrice', '');
+    }
+  },
+  [packagePrice, packageSize]);
```
kchelstowski, line 135: If you decide to create the `usePackageSpecification` hook, I think this could be placed there, but you would have to return the `setValue` from this hook and pass it to the `PricingSection` from `ProductSupplierFormMain` component.
But on the other hand I don't mind keeping it there - I just think that such calculations that are strictly connected to a specific section, should be handled in this section's hook.

kchelstowski: I'm 50-50 about this one, so I'll let you guys decide.

src/js/hooks/productSupplier/form/useProductSupplierValidation.js
```diff
@@ -81,6 +81,45 @@ const useProductSupplierValidation = () => {
...
+        }, {
+          invalid_type_error: 'Default Source Package is required',
+          required_error: 'Default Source Package is required',
+        })
+        .required(),
+      packageSize: z
+        .coerce
+        .number({ required_error: 'Package size is required' })
+        .gte(1)
+        .nullable(),
```
kchelstowski, line 98: if it's on one hand required, why is it `nullable` additionally?

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
```diff
@@ -0,0 +1,159 @@
...
+  const dispatch = useDispatch();
+
+  const {
+    quantityUoM,
+  } = useSelector(state => ({
+    quantityUoM: state.unitOfMeasure.quantity,
+  }));
+
+  useEffect(() => {
+    dispatch(fetchQuantityUoM());
```
kchelstowski, line 22: I think we should add `AbortController`

kchelstowski (changes requested):

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
drodzewicz: You actually have a point about creating a separate hook for that chunk of code, but I think that this should not be a hook directly connected to a subsection, but an options hook, hear me out...

So currently in out app there are many instances where we do the exact thing as I did above for `quantityUom`.
We fetch these options, store them in redux and access them usually in the same component 

So I feel like it should be a pattern for us that we can follow. I suggest creating hooks for this specific case where we `dispatch(fetch...)` and return this from a selector in that custom hook.

I am unsure about the naming (as always) but my initial suggestion is to follow a naming for hook `use....Options` and the directory `option-data`.

What do you think?
@kchelstowski @alannadolny

src/js/hooks/productSupplier/form/useProductSupplierForm.js
drodzewicz: yes, I see what you mean but I feel like in this case, it is alright to calculate this value in our main hook for product sources. I don't think it adds any complexity to the readability of the code since it is the same place where we transform/manipulate the form data for the API call, but that's my opinion.

src/js/components/productSupplier/create/subsections/PackageSpecification.jsx
kchelstowski: @drodzewicz good point. I think we could name it like `useOptionsFetch`, we could pass there the functions:
`useOptionsFetch([fetchQuantityUom, ...])`
and it the hook itself just map through this list and run the functions.
Although please, remember about adding the AbortController, and also, as a dependency let's add the `locale` - recently we've forgotten to add it as a dependency, but let's reduce the amount of "translations bugs" that will come up when the crowdin is invoked again.

drodzewicz: I don't think we want to create an abstract/generic `useOptionsFetch` hook, I was thinking about a separate options hook for every other case like I pushed the example for QuantityUnitOfMeasure.
So some example are hooks like `useRatingTypeCodeOptions`, `useInvoiceStatusOptions` etc... that could potentially look the same as `useQuantityUnitOfMeasureOptions`

My reasoning for that is that I want to handle the fetch and a value return in the same hook, and handling a value return would be a little complicated to do in a generic way.

kchelstowski: Ok, I misunderstood the intentions.
What I've been thinking about was a generic fetch hook, to which we could pass fetch functions and potentially the dependency array.
I don't think we need to return the options from the hook directly if they are from Redux, since we have a global access for them, but we could also combine our ideas and leave your hook and create one more hook that would be called inside yours, just for fetching stuff (useEffect + dispatch). Let's wait for others to give their opinion on that one.

kchelstowski (approved):

awalkowiak (approved):

PR #4520, opened 2024-02-26
OBPIH-6110 Implement behavior for the X rows require your attention

alannadolny:


src/js/hooks/productSupplier/form/usePreferenceTypeVariationsFiltering.jsx
```diff
@@ -0,0 +1,32 @@
...
+  const [isFiltered, setIsFiltered] = useState(false);
+
+  // react-table triggers filtering on changing value in filter fields (under column header), when
+  // we want to filter after clicking button, we have to trigger changing filters value manually
+  const triggerFiltering = () => [{ value: isFiltered }];
+
+  // If row index is equal to index of invalid row and if the row is dirty
+  // it means it's invalid
+  const isRowValid = (rowIndex) => {
+    const isRowDirty = _.some(Object.values(updatedRows?.[rowIndex] || {}));
```
kchelstowski, line 15: what's the purpose of that method, what does it do behind the scene? An empty object is truthy, so I assume that this method will always return true, hence the `&& isRowDirty` might not make sense?

src/js/components/productSupplier/create/InvalidItemsIndicator.jsx
```diff
@@ -2,31 +2,79 @@ import React from 'react';
...
-      <Translate
-        id="react.productSupplier.form.invalidItemsIndicator.title"
-        defaultMessage="Item(s) require attention"
-      />
-    </span>
-  </div>
-);
+import Button from 'components/form-elements/Button';
+
+const invalidLines = {
```
kchelstowski, line 9: what's the purpose of those two consts? Could the name be more intuitive on what it does, not what it means?

kchelstowski: After getting to know the context, let's at least name it as `invalidLinesButton`/`validLinesButton`

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
```diff
@@ -28,89 +35,148 @@ const usePreferenceTypeVariationsColumns = ({ errors, control, remove }) => {
...
-              }
-              {...field}
-            />
-          )}
+          render={({ field }) => {
+            const hasErrors = !isFieldValid(row.index, 'destinationParty');
+            return (
+              <SelectField
+                options={buyers}
+                errorMessage={hasErrors}
```
kchelstowski, line 82: this is why I said something didn't feel right to me - imho it feels like a workaround. `errorMessage` is supposed to receive a string message and in this case we pass a boolean.
I think if we need such a case, we should think about making the `errorMessage` as optional prop and add a new one `hasErrors` that would handle the logic you need. 
I think it would be better than mixing the expected type of the `errorMessage` and adding `displayErrorMessage` prop.

src/js/hooks/productSupplier/form/useProductSupplierValidation.jsx
```diff
@@ -0,0 +1,113 @@
...
+        })),
+      });
+  };
+
+  return {
+    validationSchema,
+  };
+};
+
+export default useProductSupplierValidation;
```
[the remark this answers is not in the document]
alannadolny: to be honest, I don't know, but I am changing it back to `.js`

src/js/components/productSupplier/create/InvalidItemsIndicator.jsx
```diff
@@ -2,31 +2,79 @@ import React from 'react';
...
+  isFiltered,
+  trigger,
+}) => {
+  const { Icon, variant, wrapperClassName } = appliedFiltering({
+    buttonVariant: errorsCounter ? invalidLines : validLines,
+    isFiltered,
+  });
+
+  const handleOnFilterButtonClick = () => {
+    trigger('productSupplierPreferences');
```
awalkowiak, line 43: Perhaps it would be better to move it to const? Would it be possible to give this function a better name, so it's clear what it triggers?

src/js/components/form-elements/v2/SelectField.jsx
```diff
@@ -44,7 +45,7 @@ const SelectField = ({
   return (
     <InputWrapper
       title={title}
-      errorMessage={errorMessage}
+      errorMessage={displayErrorMessage ? errorMessage : null}
```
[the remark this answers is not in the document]
alannadolny: I will extend the topic

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsFiltering.jsx
alannadolny: This empty object is a fallback in `Object.values`, so it returns an empty array, and then we are checking if something is true within this array. True means that we entered a value in any field in the row.

src/js/components/productSupplier/create/InvalidItemsIndicator.jsx
alannadolny: I was bothered about the naming here, so thanks for the suggestion.

src/js/consts/timeFormat.js
```diff
@@ -4,5 +4,5 @@ export const TimeFormat = {
 
 export const DateFormat = {
   MM_DD_YYYY: 'MM/DD/YYYY',
-  MMM_DD_YYYY: 'MMM DD, yyyy',
+  MMM_DD_YYYY: 'MMM DD, YYYY',
```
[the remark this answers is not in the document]
alannadolny: This was my mistake during rebase, when I noticed that I didn't replace `LL  with `MMMM DD, yyyy` I just changed it by hand. It doesn't matter if this one is `yyyy` or `YYYY`

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsFiltering.jsx
kchelstowski: I didn't realize the fallback is inside the `( )`. I think this shortcut is not intuitive (not passing predicate to the `_.some` function). This is why I hate lodash for methods that are available in the pure JS :smiling_face_with_tear:

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
alannadolny: Basically, I added a `displayErrorMessage` prop to avoid having this `hasErrors` :laughing: 
The idea of this is to avoid passing `hasErrors` and `errorMessage` each time. For now, we can only pass `errorMessage` and this prop is an indicator of whether we have to display an error. The `displayErrorsMessage` is an additional prop turning on/off the captions under the input field. For me, this is a little bit easier to use than the solution you proposed. I am interested in what other developers think about that, so If they have similar feelings as you, I can change it.

cc @awalkowiak @drodzewicz

kchelstowski: on one hand you say:

> `errorMessage` and this prop is an **indicator of whether we have to display an error**.

so if `errorMessage` is the **indicator**, then what the `displayErrorMessage` is? Imho the `errorMessage` should NOT be the indicator, because the name says error message, so we expect it to be a string, not a boolean/an indicator. 

If we wanted somewhat a compromise, I'd suggest either (I think not recommended since you would have to change it in every place that has been done) changing from `errorMessage` to `error` object that would receive `message` and `showMessage` properties or......

I'll let the others speak, but I strongly feel like we should not mix the destination of the `errorMessage` to use it once as a string and once as somewhat an indicator and pass a boolean.

alannadolny: Eh, the `errorMessage` is a string, if it has content then the error is displayed, if it doesn't the the content is not visible (this is what I meant by indicator here).
The `displayErrorMessage` is just a prop for showing the caption under the input. If we don't pass it, then the error is visible by default.

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
```diff
@@ -28,89 +35,148 @@ const usePreferenceTypeVariationsColumns = ({ errors, control, remove }) => {
     dispatch(fetchBuyers());
   }, []);
 
+  const isFieldValid = (index, property) => {
+    const hasFieldErrors = errors?.[index]?.[property]?.message
+                   || errors?.[index]?.[property]?.id?.message;
```
drodzewicz, line 40: a lot of nested characters, can we split this up a bit :sweat_smile: 
```jsx
const fieldError = errors?.[index]?.[property];
const fieldErrorMessage = fieldError?.message || fieldError?.id.message;
```

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
kchelstowski: Yeah, it should be a string, but you actually pass there `hasErrors` which is a boolean.
I still don't feel convinced about the approach on that. @drodzewicz @awalkowiak what do you think?

drodzewicz (approved):

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
drodzewicz: @alannadolny you can do a workaround to Kacpers requirements :grin:  
```jsx
const errorMessage = isFieldValid(row.index, 'destinationParty') ? null : 'validation error';
return (
   <SelectField
     options={buyers}
     errorMessage={errorMessage}
     ...
```

But if we want to be strict with ourselves I think we would need to add an additional `isValid` prop which would trigger some error classes and such.
Basically, `errorMessage` prop should only be responsible for the error message and `isValid` for everything else like red border classes etc..

PR #4521, opened 2024-02-26
OBPIH-6208 Resurrect the old quartz admin page

awalkowiak:
@jmiranda I resurrected the old jobs/show template plus I copied the list template form quartz plugin to overwrite it and link rows to the jobs show pages.


kchelstowski (approved):

jmiranda:
Sorry, i meant to approve this one. I pulled down the PR to test and was happy with the direction you took. I noticed that you created an GSP to override the quartz list page. I didn't realize we could do that so 👏. It probably won't be a big deal but we just need to remember to update this whenever we upgrade the plugin version.

PR #4522, opened 2024-02-27
OBPIH-6166 Add endpoint for create/edit product package

kchelstowski:


grails-app/services/org/pih/warehouse/product/ProductPackageService.groovy
```diff
@@ -0,0 +1,62 @@
+package org.pih.warehouse.product
+
+import grails.gorm.transactions.Transactional
+import org.pih.warehouse.core.ProductPrice
+
+@Transactional
+class ProductPackageService {
```
kchelstowski, line 7: I feel somewhat weird about creating product package controller and service, because anyway I have to return `ProductSupplier` from the save method (because I mutate the product supplier also, e.g. `minOrderQuantity`), so I don't know if it's a "REST-y" solution to create a separate controller/service for this case if I do not return the `ProductPackage`, or if it should be handled in the product supplier service.
But since we've discussed to create its own API for that, I decided to go with that, but I can be easily convinced to move it to the `ProductSupplierService` and rename the methods. 
Let me know what you think.

grails-app/services/org/pih/warehouse/product/ProductPackageService.groovy
```diff
@@ -0,0 +1,62 @@
...
+            defaultProductPackage.product = command.productSupplier?.product
+            defaultProductPackage.uom = command.uom
+            defaultProductPackage.quantity = command.productPackageQuantity
+            productSupplier.addToProductPackages(defaultProductPackage)
+            productSupplier.defaultProductPackage = defaultProductPackage
+        }
+        // If product package price is not provided, skip assigning pricing data
+        if (!command.productPackagePrice) {
+            return
+        }
```
kchelstowski, line 39: this would mean that if we set the price once, we can't make it null anymore. But this is how it is done currently in the import, hence I kept the current behavior.

kchelstowski: Actually, it can't become null, but it can still be `0` which it can't with my solution, because I check the truthiness. I'm gonna change it to `== null`

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
```diff
@@ -107,7 +110,7 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
         active(nullable: true)
     }
 
-    ProductPackage getDefaultProductPackage() {
+    ProductPackage getDefaultProductPackageDerived() {
         return productPackages ? productPackages.sort { it.lastUpdated }.last() : null
```
alannadolny, line 114: maybe it is worth to create `compareTo` in the `ProductPackage` domain

grails-app/services/org/pih/warehouse/product/ProductPackageService.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+    ProductSupplier save(ProductPackageCommand command) {
+        setPackageData(command)
+        if (command.contractPricePrice) {
+            setContractPriceData(command.productSupplier, command.contractPricePrice, command.contractPriceValidUntil)
+        }
+        return command.productSupplier
+    }
+
+    private static void setPackageData(ProductPackageCommand command) {
```
alannadolny, line 17: why is this method static (and the one below also)?

src/main/groovy/org/pih/warehouse/product/ProductPackageCommand.groovy
```diff
@@ -0,0 +1,38 @@
...
+        productPackageQuantity(min: 0)
+        minOrderQuantity(nullable: true, min: 0.0)
+        productPackagePrice(nullable: true, min: 0.0)
+        contractPricePrice(nullable: true, min: 0.0)
+        contractPriceValidUntil(nullable: true, validator: { Date date ->
+            Date minDate = ConfigHelper.getMinimumExpirationDate()
+            if (date < minDate) {
+                return false
+            }
+            return true
```
alannadolny, line 35: can't you just return `date < minDate`?

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
kchelstowski: I just renamed the method, didn’t care about the content, because it was not in the scope of my work

grails-app/services/org/pih/warehouse/product/ProductPackageService.groovy
kchelstowski: Because the implementation doesn’t depend on actual instance of the service

src/main/groovy/org/pih/warehouse/product/ProductPackageCommand.groovy
kchelstowski: I probably could but would have to negate it, because otherwise I’d return true if date < minDate

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
alannadolny: Oh, that's true, didn't notice that, sry

grails-app/services/org/pih/warehouse/product/ProductPackageService.groovy
alannadolny: `setPackageData` sounds like a method that should set something on a currently existing instance, something feels wrong to me with that.

kchelstowski: it does on the product supplier instance that is bound to the command.

alannadolny (approved):

grails-app/domain/org/pih/warehouse/product/ProductSupplier.groovy
```diff
@@ -127,6 +130,46 @@ class ProductSupplier implements Serializable, Comparable<ProductSupplier> {
...
+        }
+        // If a default package of 1:1 association is not found, check for a default package "in an old way" (last updated)
+        if (defaultProductPackageDerived) {
+            return defaultProductPackageDerived?.productPrice
+                    ? (defaultProductPackageDerived?.productPrice?.price / defaultProductPackageDerived?.quantity)?.setScale(2, RoundingMode.HALF_UP)
+                    : 0.0
+        }
+        return 0.0
+    }
+
```
kchelstowski, line 172: lines 133-172 (and this commit in particular https://github.com/openboxes/openboxes/pull/4522/commits/fd98c02435f2c49683cfac23b45ce23a4a97d7dc) are applying the new behavior of the default product package to the product sources list.
At first we wanna search for a default product package that is set in the new association `ProductPackage defaultProductPackage` and if we do not find one, then we wanna search for the default product package "in an old way" to support backwards compatibility. 
It has been specified and requested in the comments of OBPIH-6116.

awalkowiak (approved):

PR #4523, opened 2024-02-28
OBPIH-6108 add preference type section with inputs

drodzewicz:


awalkowiak (approved):

src/js/components/productSupplier/create/subsections/DefaultPreferenceType.jsx
```diff
@@ -0,0 +1,133 @@
...
+                title={{
+                  id: 'react.productSupplier.form.defaultPreferenceType.title',
+                  defaultMessage: 'Default Preference Type',
+                }}
+                tooltip={{
+                  id: 'react.productSupplier.form.defaultPreferenceType.tooltip',
+                  defaultMessage: 'Company-wide purchasing preference for this supplier established through a competitive bid',
+                }}
+                options={preferenceTypes}
+                errorMessage={errors.defaultPreferenceType?.message}
```
alannadolny, line 45: One change related to the requested changes in my previous PR is missing. To display the error you need `hasErrors={true}` (only in select fields)

src/js/hooks/productSupplier/form/useProductSupplierForm.js
```diff
@@ -125,6 +111,7 @@ const useProductSupplierForm = () => {
       ratingTypeCode: values?.ratingTypeCode ? values.ratingTypeCode.id : null,
       productSupplierPreferences: values?.productSupplierPreferences,
       uom: values?.uom ? values.uom.id : null,
+      defaultPreferenceType: values?.defaultPreferenceType ? values.defaultPreferenceType.id : null,
```
alannadolny, line 114: Is `values?.defaultPreferenceType?.id` not enough? I think that when the `values?.defaultPreferenceType` will be empty, then `id` should be `null` / `undefined`

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
```diff
@@ -31,7 +31,6 @@ const usePreferenceTypeVariationsColumns = ({
   const translate = useTranslate();
 
   useEffect(() => {
-    dispatch(fetchPreferenceTypes());
     dispatch(fetchBuyers());
   }, []);
```
alannadolny, line 35: This is not a change request, but rather a loud thinking. Maybe we should create a separate hook, for fetching data, we can pass functions, as an argument, and within the hook, we will check if everything is fetched, and based on that show / hide the spinner.

alannadolny (changes requested):

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
kchelstowski: I'd suggest thinking asap about the react-query that would handle such things for us (loading (spinner), caching, etc.) 
cc @drodzewicz

kchelstowski (approved):

alannadolny (approved):

PR #4524, opened 2024-02-28
OBPIH-6111 add pricing section with inputs

drodzewicz:
One thing you may notice are changes to some other sections. This is because I have added the bootstrap grid system as a wrapper for our fields.
One of the reopen change requests I got was that my fields we not in the same layout as on the mock so I decided to use bootstrap grid system to solve this issue.
To not have different approaches in other sections I did the following modifications and also added `row` and `col-lg` classes

### DIFF
--- a/src/js/hooks/productSupplier/form/useProductSupplierForm.js
+++ b/src/js/hooks/productSupplier/form/useProductSupplierForm.js
@@ -139,6 +139,17 @@ const useProductSupplierForm = () => {
   },
   [packagePrice, productPackageQuantity]);
 
+  const uom = useWatch({ control, name: 'uom' });
+
+  // preselect value 1 when unit of measure Each is selected
+  useEffect(() => {
+    if (uom?.id === 'EA') {
+      setValue('productPackageQuantity', 1, { shouldValidate: true });
+    } else {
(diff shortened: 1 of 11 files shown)


src/js/hooks/productSupplier/form/useProductSupplierForm.js
```diff
@@ -126,14 +127,29 @@ const useProductSupplierForm = () => {
...
       setValue('eachPrice', '');
     }
   },
   [packagePrice, productPackageQuantity]);
 
+  const uom = useWatch({ control, name: 'uom' });
+
+  // preselect value 1 when unit of measure Each is selected
+  useEffect(() => {
+    if (uom?.id === 'EA') {
```
kchelstowski, line 146: this feels dangerous to me. Are we sure that on every instance this would have the id of `'EA'`? 
I think we'd be safer if we looked at `uom.code`, but we would have to include the code in the response of uoms and change the `getDefaultValues()` to include the code.
Unless @awalkowiak / @jmiranda know if this will always have id of `'EA'` on any instance, then we can let it go.

kchelstowski (approved):

PR #4527, opened 2024-03-04
OBPIH-6112 Add attributes section with inputs

alannadolny:


grails-app/services/org/pih/warehouse/product/AttributeService.groovy
```diff
@@ -0,0 +1,18 @@
...
+
+class AttributeService {
+
+    def list(EntityTypeCode entityTypeCode) {
+        String query = "select a from Attribute a"
+        Map<String, String> argumentsList = [:]
+
+        if (entityTypeCode) {
+            query += " join a.entityTypeCodes etc where etc = :entityTypeCode"
+            argumentsList += [entityTypeCode: EntityTypeCode.PRODUCT_SUPPLIER]
```
kchelstowski, line 13: shouldn't it be ?
```groovy
 argumentsList += [entityTypeCode: entityTypeCode]
```

src/js/components/productSupplier/create/sections/AttributesSection.jsx
```diff
@@ -0,0 +1,53 @@
...
+  const { attributesWithInputTypes } = useProductSupplierAttributes();
+
+  return (
+    <Section title={{ label: 'react.productSupplier.section.attributes.title', defaultMessage: 'Attributes' }}>
+      <div className="attributes-section gap-8 mt-4">
+        {
+          attributesWithInputTypes.map(({ attribute, Input }) => {
+            const errorMessage = errors?.[attribute?.id]?.message;
+            return (
+              <Controller
```
kchelstowski, line 19: `key` is missing

src/js/components/productSupplier/create/styles.scss
```diff
@@ -75,3 +75,8 @@
     font-weight: 400;
   }
 }
+
+.attributes-section {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
```
kchelstowski, line 81: wouldn't class `form-grid-3` work for you? It's the same as this one, with only difference that is `gap: 8px`

src/js/consts/entityTypeCode.js
```diff
@@ -0,0 +1,6 @@
+const entityTypeCode = {
```
kchelstowski, line 1: This should be defined as `EntityTypeCode`

src/js/hooks/productSupplier/form/useProductSupplierAttributes.js
```diff
@@ -0,0 +1,137 @@
...
+  const { attributes } = useSelector((state) => ({
+    attributes: state.productSupplier.attributes,
+  }));
+
+  useEffect(() => {
+    dispatch(fetchAttributes({
+      params: {
+        entityType: entityTypeCode.PRODUCT_SUPPLIER,
+      },
+    }));
```
kchelstowski, line 24: let's add an AbortController

src/js/hooks/productSupplier/form/useProductSupplierAttributes.js
```diff
@@ -0,0 +1,137 @@
...
+   * text input. When the attribute has the allowOther property set to true and has available
+   * options, then it should be a select field, but with the additional option "Other". In all
+   * other cases, it should be just a select field with the options fetched from the API.
+   */
+  const getAttributesWithInputTypes = (attributesToMap) =>
+    attributesToMap.map((attribute) => {
+      if (!attribute?.allowOther) {
+        return {
+          attribute: { ...attribute, options: mapAttributeOptions(attribute) },
+          Input: SelectField,
```
kchelstowski, line 85: I think this should be lower-case (I know long time ago I've introduced a StartIcon, but I believe it was wrong)

src/js/hooks/productSupplier/form/useProductSupplierAttributes.js
```diff
@@ -0,0 +1,137 @@
...
+    ? z.object({
+      value: z.string(),
+      label: z.string(),
+    }, errorProps)
+      .required()
+    : z.object({
+      value: z.string(),
+      label: z.string(),
+    })
+      .nullish()
```
kchelstowski, line 54: I think `.optional()` should also be added, otherwise it might still be considered as required, but I don't remember the exact case.

alannadolny: `optional`, `nullable`, and `nullish` are designated to cover a specific value, undefined or null. The `nullish` covers all of those cases. So the `nullish` includes `optional` within itself.
```
{
    optional?: string | undefined;
    nullable: string | null;
    nullish?: string | null | undefined;
}
```

src/js/hooks/productSupplier/form/useProductSupplierAttributes.js
alannadolny: If you mean the `SelectField` it's just our select component, it should be capitalized.

kchelstowski: no, I meant the property name

kchelstowski: ok, now I see the usage of that, so this makes sense.

kchelstowski (approved):

awalkowiak (approved):

PR #4534, opened 2024-03-06
OBPIH-5926 / OBPIH-6206 Delete product source preference records

alannadolny:


src/js/components/productSupplier/create/subsections/DefaultPreferenceType.jsx
```diff
@@ -1,21 +1,54 @@
...
+  const emptyPreferenceType = {
+    bidName: '',
+    validityStartDate: '',
+    validityEndDate: '',
+    preferenceType: '',
+  };
+
+  const afterDelete = () => {
+    setValue('defaultPreferenceType', emptyPreferenceType);
+  };
```
kchelstowski, line 41: not a must have, but I just feel like 27-41 lines could also be included in the hook you've created.
Another not "must have" (because it won't impact the performance at all) is that the `emptyPreferenceType` could be outside of the component, because its body never changes when a re-render happens (it is not dependent on the component life cycle)

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
```diff
@@ -0,0 +1,97 @@
...
+  preferenceTypeData,
+  afterDelete,
+  isDefaultPreferenceType,
+  onCancel,
+}) => {
+  const dispatch = useDispatch();
+
+  const isPreferenceTypeEmpty = !_.some(
+    Object.values(_.omit(preferenceTypeData || {}, 'id')),
+  );
```
kchelstowski, line 19: wouldn't it be cleaner to do it using `isEmpty`?
```js
const isPreferenceTypeEmpty = _.isEmpty(_.omit(preferenceTypeData || {}, 'id'))
```

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
```diff
@@ -0,0 +1,97 @@
...
+      },
+  };
+
+  const deletePreferenceType = async (onClose) => {
+    dispatch(showSpinner());
+    try {
+      if (preferenceTypeData?.id) {
+        await productSupplierApi.deleteProductSupplierPreference(preferenceTypeData?.id);
+      }
+      afterDelete?.();
```
kchelstowski, line 43: if we eventually don't call the delete (because of the if above), should we call the `afterDelete` anyway, or should it be inside the if also?

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
```diff
@@ -0,0 +1,97 @@
...
+    isPreferenceTypeEmpty,
+  };
+};
+
+export default useDeletePreferenceType;
+
+useDeletePreferenceType.propTypes = {
+  preferenceTypeData: PropTypes.shape({}),
+  afterDelete: PropTypes.func.isRequired,
+  isDefaultPreferenceType: PropTypes.bool,
```
kchelstowski, line 91: don't you miss `onCancel` here?

src/js/hooks/productSupplier/form/usePreferenceTypeVariationsColumns.jsx
```diff
@@ -19,6 +20,7 @@ const usePreferenceTypeVariationsColumns = ({
   updatedRows,
   triggerValidation,
 }) => {
+  const [selectedRow, setSelectedRow] = useState(null);
```
kchelstowski, line 23: let's add `index` to the name (or add a comment above that is index).

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
alannadolny: It's not the same.
As an empty preferenceType, we are considering { bidName: '', validityStartDate: '', ... } 
1. `isEmpty`
In this case, this function returns false because it checks if the object has any keys (it has, but with empty values)
2. `some` + `Object.values`
In this case, we are taking only the values of the keys, so it gives us an array like that ['', '', ...] and then we are checking the truthiness of this array, and because of the fact every value in this array is falsy, the whole array is also falsy.

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
alannadolny: yes, because if the row doesn't have the ID, it means it is a new row, so we don't have to make an API call, but we still would like to clear \ remove this row.

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
kchelstowski: yep, my bad, didn't understand the context before your explanation.

src/js/hooks/productSupplier/form/useDeletePreferenceType.js
kchelstowski: thanks, this is what I wanted to hear

kchelstowski (approved):

PR #4535, opened 2024-03-07
OBPIH-6112 Add attributes section with inputs (fixes after QA)

alannadolny:


grails-app/services/org/pih/warehouse/product/AttributeService.groovy
```diff
@@ -4,13 +4,13 @@ import org.pih.warehouse.core.EntityTypeCode
...
-    def list(EntityTypeCode entityTypeCode) {
+    def list(EntityTypeCode entityTypeCode, Boolean active = true) {
         String query = "select a from Attribute a"
         Map<String, String> argumentsList = [:]
 
         if (entityTypeCode) {
-            query += " join a.entityTypeCodes etc where etc = :entityTypeCode"
-            argumentsList += [entityTypeCode: entityTypeCode]
+            query += " join a.entityTypeCodes etc where etc = :entityTypeCode and active = :active"
+            argumentsList += [entityTypeCode: entityTypeCode, active: active]
```
alannadolny, line 13: I wasn't sure how I should deal with the active property separately from entityTypeCode here, because the syntax of this query forced me to append the active in the where clause

alannadolny:
I am marking this PR as a draft, to wait for Manon's response to proposed changes from Katarzyna.

grails-app/services/org/pih/warehouse/product/AttributeService.groovy
awalkowiak: @alannadolny one minor thing for now, if you have above `select a from Attribute a`, and you use it in join as `join a.entityTypeCodes` then please use also `a` here as `and a.active = :active` for clarity.

grails-app/services/org/pih/warehouse/product/AttributeService.groovy
```diff
@@ -4,13 +4,13 @@ import org.pih.warehouse.core.EntityTypeCode
...
 
-    def list(EntityTypeCode entityTypeCode) {
+    def list(EntityTypeCode entityTypeCode, Boolean active = true) {
         String query = "select a from Attribute a"
         Map<String, String> argumentsList = [:]
 
         if (entityTypeCode) {
-            query += " join a.entityTypeCodes etc where etc = :entityTypeCode"
-            argumentsList += [entityTypeCode: entityTypeCode]
+            query += " join a.entityTypeCodes etc where etc = :entityTypeCode and a.active = :active"
```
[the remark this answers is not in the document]
kchelstowski: if we suppose to have a default value for `active` we could do it like (pseudocode, check if it works/adjust it)
```groovy
String query = "select a from Attribute a"
String whereQuery = "where a.active = :active"
Map<String, String> argumentsList = [:]

if (entityTypeCode) {
        argumentsList += [entityTypeCode: entityTypeCode]
        query += " join a.entityTypeCodes etc"
        whereQuery += " and etc = :entityTypeCode"
}
argumentsList.active = active
query += whereQuery

return Attribute.executeQuery(query, argumentsList)

File: src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx [hunk 3]
```diff
@@ -0,0 +1,96 @@
...
+            size="24px"
+            className="cursor-pointer"
+            role="button"
+            onClick={onClose}
+          />
+        </div>
+        <div>
+          <TextInput
+            required
+            title={{ id: '', defaultMessage: 'Other' }}
```
kchelstowski, line 54: should the `id` be empty?

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
```diff
@@ -0,0 +1,96 @@
...
+            errorMessage={inputValue === '' && translate(
```
kchelstowski, line 55: what if `inputValue` would be `null`? Why don't we want to validate it then?

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
```diff
@@ -0,0 +1,96 @@
...
+          <Button
+            variant="transparent"
+            defaultLabel="Cancel"
+            label=""
+            onClick={onClose}
+          />
+          <Button
+            disabled={!inputValue}
+            defaultLabel="Confirm"
+            label=""
```
kchelstowski, line 73: should it be empty here and in the button above?

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/product/AttributeService.groovy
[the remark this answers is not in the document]
awalkowiak: :+1:

awalkowiak (changes requested):

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
drodzewicz, line 1: I think we can add a form with validation schema on this modal and 
```
setValue(`attributes.${selectedAttribute?.id}`, { id: attribute?.id, value, label: value });
```
on form submit.
This will require a little more code but is probably the right approach
cc @kchelstowski

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
```diff
@@ -0,0 +1,96 @@
...
+  const onClose = () => {
+    setInputValue(null);
+    close();
+  };
+
+  const setAttributeValue = (attribute, value) => () => {
+    setValue(`attributes.${selectedAttribute?.id}`, { id: attribute?.id, value, label: value });
+    setInputValue(null);
+    close();
+  };
```
drodzewicz, line 33: I think this doesn't have to be a double callback 
```suggestion
  const setAttributeValue = () => {
    setValue(`attributes.${selectedAttribute?.id}`, { id: selectedAttribute?.id, value, label: inputValue });
    setInputValue(null);
    close();
  };
```
am I wrong?

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
alannadolny: If the value is null, we are treating the input as not touched previously

src/js/components/productSupplier/modals/AddAttributeOptionModal.jsx
alannadolny: Yeah, it is probably the right approach, but I think it's a little overkill to use a new hook and new validation schema to handle the one input with a text value, with validation only for the presence/absence of the entered value.

drodzewicz (approved):

awalkowiak (approved):

PR #4539, opened 2024-03-11
OBPIH-6232 show hyperlink to source edit and stockcard on sources list page when user has sufficient roles

drodzewicz:
An additional case that I had to include in this ticket was conditionally rendered link to product sources edit if the user has sufficient roles (user is at least an admin and has product manager roles).

I have decided to create a custom hook `useUserHasPermission` to encapsulate the logic of retrieving user role data from redux and performing appropriate checks.

I also had to return location roles in the AppContext since I noticed that in a previous version of `hasPermission` util function we don't check for location roles on front end, only global roles.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
@@ -144,6 +144,7 @@ class ApiController {
         User user = User.get(session?.user?.id)
         Location location = Location.get(session.warehouse?.id)
         String highestRole = user.getHighestRole(location)
+        List currentLocationRoles = user.getRolesByCurrentLocation(location)?.roleType*.name()
         boolean isSuperuser = userService.isSuperuser(session?.user)
         boolean isUserAdmin = userService.isUserAdmin(session?.user)
         boolean isUserApprover = userService.hasRolePurchaseApprover(session?.user)
@@ -188,6 +189,7 @@ class ApiController {
             data: [
                 user                 : user,
                 location             : location,
+                currentLocationRoles: currentLocationRoles,
                 isSuperuser          : isSuperuser,
                 isUserAdmin          : isUserAdmin,
                 isUserApprover       : isUserApprover,
--- a/src/js/components/productSupplier/ProductSupplierListTable.jsx
+++ b/src/js/components/productSupplier/ProductSupplierListTable.jsx
@@ -8,6 +8,7 @@ import DateCell from 'components/DataTable/DateCell';
 import Button from 'components/form-elements/Button';
 import PreferenceTypeColumn from 'components/productSupplier/PreferenceTypeColumn';
 import { INVENTORY_ITEM_URL, PRODUCT_SUPPLIER_URL } from 'consts/applicationUrls';
+import RoleType from 'consts/roleType';
 import useProductSupplierActions from 'hooks/list-pages/productSupplier/useProductSupplierActions';
 import useProductSupplierListTableData from 'hooks/list-pages/productSupplier/useProductSupplierListTableData';
 import ActionDots from 'utils/ActionDots';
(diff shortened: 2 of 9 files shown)


src/js/hooks/useUserHasPermissions.js
```diff
@@ -0,0 +1,75 @@
...
+   * @type {function(RoleType[]): boolean}
+   */
+  const hasEveryRole = useCallback((roles) => {
+    const userRoles = Array.isArray(currentUser?.roles) ? currentUser?.roles : [];
+    const userLocationRoles = Array.isArray(currentLocationRoles) ? currentLocationRoles : [];
+
+    const allUserRoles = new Set([...userRoles, ...userLocationRoles]);
+
+    return _.every(roles, (role) => allUserRoles.has(role));
+  }, [currentUser]);
```
kchelstowski, line 43: shouldn't the `currentLocation` also be a dependency?

drodzewicz: yes, true.
And currentUser dependancy should be `[currentUser.id]`

kchelstowski: exactly :D and the `currentLocation` -> `currentLocation?.id`

kchelstowski (approved):

jmiranda (commented):
I'll let @awalkowiak decide on what to do here but I recommend moving the auth logic to the server and making it configurable.

src/js/hooks/useUserHasPermissions.js
```diff
@@ -0,0 +1,78 @@
...
+import RoleType from 'consts/roleType';
+
+/**
+ * Returns true or false if current logged-in user has all the roles specified in the parameters.
+ * The function checks if user has either global or location based roles.
+ * @param minRequiredRole {RoleType} - minimum required role
+ * @param supplementalRoles {RoleType[]} [d=[]] - list of supplemental required roles
+ * @returns {boolean}
+ */
+const useUserHasPermissions = ({ minRequiredRole, supplementalRoles = [] }) => {
```
awalkowiak, line 15: Should this be a replacement for the `permisionUtil` that we already have? Is there a case that we want to have both? I think we should have only one thing if that's possible.

src/js/components/productSupplier/ProductSupplierListTable.jsx
```diff
@@ -30,6 +32,11 @@ const ProductSupplierListTable = ({ filterParams }) => {
     exportProductSuppliers,
   } = useProductSupplierActions({ fireFetchData });
 
+  const isUserAdminWithProductManager = useUserHasPermissions({
```
awalkowiak, line 35: Let's change it to the suggested convention `canManageProducts` and let's stick to that version of variable names moving forward.

awalkowiak (changes requested):
@drodzewicz see my comments for the change requests (at least change the name of variable, and do a second one if will be a quick one, otherwise let me know and I'll put it under the TD ticket that I am going to create for this one)

@jmiranda I agree with you here, that it would be better to handle this on the backend, but on the other hand if we'd need to do a separate property for each combination of `canDoSomething` this feels odd to me too. I prefer moving this froward as is and do a `Spike` together with future permissions/security refactor to investigate how we should and want to resolve this. I am going to create a TD ticket and we can discuss it further on the tech huddle.

awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -144,6 +144,7 @@ class ApiController {
         User user = User.get(session?.user?.id)
         Location location = Location.get(session.warehouse?.id)
         String highestRole = user.getHighestRole(location)
+        List currentLocationRoles = user.getRolesByCurrentLocation(location)?.roleType*.name()
```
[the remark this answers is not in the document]
awalkowiak: I created this ticket for this one https://pihemr.atlassian.net/browse/OBPIH-6254

PR #4540, opened 2024-03-11
OBPIH-6108 Trigger validation on default preference type clearing (Fixes after QA)

drodzewicz:
- trigger validation on preferrence type select when clearing default preference type subsection
- on edit page of product sources, assign `undefined` instead of `null` on **validityStartDate**, **validityEndDate** and **bidName** because API returns `null` and Zod parses `null` to some date value when using `coerce.date()` 
- fix padding on form inputs and prevent inputs from jumping when errors appear under inputs (I have added `minHeight` on error message to prevent this height change when error appears)

### DIFF
--- a/src/js/hooks/productSupplier/form/useProductSupplierValidation.js
+++ b/src/js/hooks/productSupplier/form/useProductSupplierValidation.js
@@ -130,6 +130,7 @@ const useProductSupplierValidation = () => {
         .optional(),
       bidName: z
         .string()
+        .nullable()
         .optional(),
     })
       .refine(checkPreferenceTypeIsRequired, {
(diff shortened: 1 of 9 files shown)


src/js/hooks/productSupplier/form/useProductSupplierValidation.js
```diff
@@ -130,6 +130,7 @@ const useProductSupplierValidation = () => {
         .optional(),
       bidName: z
         .string()
+        .nullable()
```
kchelstowski, line 133: this can be either changed to `.nullish()` or removed at all, because you made a fallback for `?? undefined`

kchelstowski (approved):

alannadolny (approved):

PR #4543, opened 2024-03-13
OBPIH-6126 Rename supplier product name and site name to other generic names

drodzewicz:
It was requested to change the generic name for fields/columns _Supplier Product Name_ and _Site Name_ to **(Source) Name** and **Organization**

Additionally I have added jsdoc to our translation hook and utils function to describe what each parameter is doing because I find it confusing every time I want to remember which parameter takes in **data** **defaultMessage** etc...


grails-app/i18n/messages.properties
```diff
@@ -4303,7 +4304,7 @@ react.productSupplier.column.packagePrice.label=Package Price
 react.productSupplier.column.eachPrice.label=Each Price
 react.productSupplier.column.supplier.label=Supplier
 react.productSupplier.column.supplierCode.label=Supplier Code
-react.productSupplier.column.supplierProductName.label=Supplier Product Name
+react.productSupplier.column.name.label=(Source) Name
```
awalkowiak, line 4307: Why `(Source)` is in parenthesis? Was that requested?

awalkowiak (approved):

PR #4550, opened 2024-03-18
OBPIH-6092 Experimental fix to the issue with jquery tabs loading

awalkowiak:


grails-app/views/order/show.gsp
```diff
@@ -282,6 +282,17 @@
                     },
                     ajaxOptions: {
                       error: function(xhr, status, index, anchor) {
+                        var errorMessage = "Error loading tab: " + xhr.status + " " + xhr.statusText;
```
alannadolny, line 285: I assume it's copied from elsewhere, but
`var` -> `let`

awalkowiak: fixed

PR #4552, opened 2024-03-18
fixed bug on number format for Dutch language.

chetanmaharishi:


grails-app/i18n/messages_nl.properties
```diff
@@ -482,7 +482,7 @@ default.notes.label=Notities
 default.null.message=Eigenschap [{0}] van klasse [{1}] kan niet null zijn
 #default.number.format                 = 0
 default.numItems.label=\# items
-default.number.format=\#\#\#.\#\#\#.\#\#0,00
+default.number.format=###,###,##0.00
```
awalkowiak, line 485: I think all the `#` require backslash escape, because this file supports comments starting with `#`

awalkowiak (changes requested):

grails-app/i18n/messages_nl.properties
jmiranda: I would have guessed you were right, but check messages.properties and all of the other .properties files.

awalkowiak: :thinking:

awalkowiak: Yup, looks like you're right, then you can disregard this comment, and it lgtm

PR #4555, opened 2024-03-20
OBPIH-6210 Checked receipt line is received with quantity when edited to zero

alannadolny:


alannadolny:
I marking it as a draft, because I need one more confirmation from Manon about the proper behavior.

awalkowiak (approved):

PR #4556, opened 2024-03-20
OBPIH-6086 update actual ready date on multiple POs via import

drodzewicz:


grails-app/services/org/pih/warehouse/importer/PurchaseOrderActualReadyDateImportDataService.groovy
```diff
@@ -0,0 +1,128 @@
...
+        log.info "Validate data ${command.filename}"
+
+        command.data.eachWithIndex { params, index ->
+            Order order
+            OrderItem orderItem
+            Product product
+
+            // Product validation
+            if (!params["productCode"]) {
+                command.errors.reject("Row ${index + 1}: 'productCode' is required")
```
alannadolny, line 38: I am wondering whether those errors should be translatable. Do you think there's a need for that? Or we can leave it as is? :thinking:

src/main/groovy/org/pih/warehouse/importer/PurchaseOrderActualReadyDateExcelImporter.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    'E': 'actualReadyDate',
+            ]
+    ]
+
+    static Map propertyMap = [
+            "orderItemId"           : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "orderNumber"           : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "productCode"           : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "productName"           : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "actualReadyDate"       : ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
```
alannadolny, line 44: I am assuming it should be StringType instead of Date?

alannadolny (approved):

awalkowiak (approved):

PR #4570, opened 2024-03-28
OBPIH-6278 Do not allow uom qty different than 1, when chosen uom is …

kchelstowski:
…Each for the PO items import

### DIFF
--- a/grails-app/services/org/pih/warehouse/order/OrderService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderService.groovy
@@ -801,7 +801,7 @@ class OrderService {
                         UnitOfMeasure uom = uomParts.length > 1 ? UnitOfMeasure.findByCodeOrName(uomParts[0], uomParts[0]) : null
                         orderItem.quantityUom = uom
                         BigDecimal qtyPerUom = uomParts.length > 1 ? CSVUtils.parseNumber(uomParts[1], "unitOfMeasure"): null
-                        if (uom?.id == 'EA' && qtyPerUom != 1) {
+                        if (uom?.id == Constants.UOM_EACH_ID && qtyPerUom != 1) {
                             throw new IllegalArgumentException("Quantity per UoM must be 1, if selected UoM is Each")
                         }
                         if (!qtyPerUom || qtyPerUom < 1) {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -801,6 +801,9 @@ class OrderService {
                         UnitOfMeasure uom = uomParts.length > 1 ? UnitOfMeasure.findByCodeOrName(uomParts[0], uomParts[0]) : null
                         orderItem.quantityUom = uom
                         BigDecimal qtyPerUom = uomParts.length > 1 ? CSVUtils.parseNumber(uomParts[1], "unitOfMeasure"): null
+                        if (uom?.id == 'EA' && qtyPerUom != 1) {
```
awalkowiak, line 804: I think it is worth moving `'EA'` to `Constants.groovy` as `UOM_EACH_ID` or `EACH_ID`

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
kchelstowski: yeah, I always feel uncomfortable hard-coding strings, but I thought you guys would treat it as an overkill in that case :see_no_evil: but I'm happy we are on the same page

awalkowiak (approved):

grails-app/services/org/pih/warehouse/order/OrderService.groovy
jmiranda: I agree with converting string literals to constants in most cases, especially for things like PKs. The exception is when it makes it much more difficult to read the code. For example, if I need to go to the constants file in order to figure out what the code is doing then that would be worse then using the string literal.

But this rule should be in our coding conventions and perhaps should become a linter rule as well.

jmiranda: And in this particular case, I would prefer UOM_EA_ID over EA_ID because the latter would require me to go to the constants.groovy to figure out what's going on. So hopefully that was what we went with.

awalkowiak: @jmiranda Kacper went with even more descriptive -> `UOM_EACH_ID`

jmiranda: I saw that. :golf: :clap:

PR #4573, opened 2024-03-29
OBPIH-6288 Do not allow to re-generate source code when importing pro…

kchelstowski:
…duct sources and the code is not provided

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
@@ -208,7 +208,13 @@ class ProductSupplierService {
         if (!productSupplier) {
             productSupplier = new ProductSupplier(params)
         } else {
+            String sourceCode = productSupplier.code
             productSupplier.properties = params
+            // Do not allow to bind an empty source code, which caused the code to be re-generated for an existing product supplier (OBPIH-6288)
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -208,7 +208,13 @@ class ProductSupplierService {
         if (!productSupplier) {
             productSupplier = new ProductSupplier(params)
         } else {
+            String sourceCode = productSupplier.code
             productSupplier.properties = params
+            // Do not allow to bind an empty source code, which caused the code to be re-generated for an existing product supplier (OBPIH-6288)
+            // The code could be overwritten by the binding in the 212th line, hence we'd like to bring back the existing code, if the one in params is null/empty
```
alannadolny, line 214:
```
in the 212th line
```
If someone changes the code above, it won't be 212th line :smile:

alannadolny (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
awalkowiak: Yeah let's avoid doing that, I know I did it once or twice in the past, but let's refrain from adhering to line numbers in comments.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4576, opened 2024-04-05
OBPIH-6308 Delete row removes input and split lines in stock transfer process

alannadolny:


src/js/components/stock-transfer/StockTransferSecondPage.jsx
```diff
@@ -327,22 +327,42 @@ class StockTransferSecondPage extends Component {
     this.setState({ stockTransfer });
   }
 
+  getFilteredStockTransferItems(items, indexToDelete) {
```
awalkowiak, line 330: @alannadolny If the purpose of this function is to filter out the deleted items, then I would rename the function, if it can be used to filter out other items, not only the deleted ones, then I'd change the parameter name and the comments.

awalkowiak (approved):

PR #4579, opened 2024-04-10
OBPIH-6289 Add migration to remove faulty organization references from Product Supplier

drodzewicz:


alannadolny (approved):

awalkowiak (changes requested):
@drodzewicz wasn't it meant to be done directly on the prod's db? Not through migration?

drodzewicz:
> @drodzewicz wasn't it meant to be done directly on the prod's db? Not through migration?

Well yes, you might be right.
Since this was a rather simple database update I assumed that we could get away with a simple migration and affect all of the openboxes environment with this change, but now that I am thinking about it again, since this might be just PiH problem maybe a better approach would be to just update obnav directly.

I have left the comment in the ticket with the script and results from obnavstage https://pihemr.atlassian.net/browse/OBPIH-6289?focusedCommentId=151696 After I get your approval I can execute it on production

drodzewicz:
Closing this PR due to the fact that this issue should be handled directly on the obnav database since this looks like a PiH specific issue

PR #4580, opened 2024-04-11
OBPIH-6318 Review closed dependabot alerts and check frontend dependencies security vulnerabilities

alannadolny:


alannadolny:
Please, anyone reviewing this, be more careful about the libraries I removed. I checked the app, and everything works well, I removed them based on the depcheck. I also checked the docs of them and I didn't find any reasons why they were used in the app, but if something feels doesn't right for you, shout at me 🫣🫣

awalkowiak (approved):

PR #4583, opened 2024-04-15
OBPIH-6300 In progress status for fulfillment steps on requestor's dashboard

alannadolny:


src/main/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
```diff
@@ -32,6 +32,9 @@ enum StockMovementStatusCode {
     CANCELED(10),
     REJECTED(10, REJECTED),
 
+    // Status on the requestor's dashboard representing verifying, checked, picking and picked
+    IN_PROGRESS(3, PENDING),
```
awalkowiak, line 36: I wonder if this should not be the display status on its own (below the `PENDING`), without sort order, and should not take `PENDING` as its own displayStatusCode :thinking:

awalkowiak (commented):
Is this `getDisplayStatus()` only used on the requestor's dashboard?

I feel like this ticket is a really good candidate to write some unit tests around it.

src/main/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
alannadolny: > Is this getDisplayStatus() only used on the requestor's dashboard?
yes, it was introduced for mapping approval statuses for the requestor's dashboard

awalkowiak (approved):

PR #4586, opened 2024-04-17
OBPIH-6215 Refactor returns statuses (both inbound and outbound) and …

kchelstowski:
…align them on list page<->view page

### DIFF
--- a/grails-app/migrations/views/stock-movement-list-item.sql
+++ b/grails-app/migrations/views/stock-movement-list-item.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE VIEW stock_movement_list_item AS
-    SELECT 
+    SELECT
         r.id,
         r.name,
         r.request_number AS identifier,
@@ -34,8 +34,8 @@ CREATE OR REPLACE VIEW stock_movement_list_item AS
             LEFT JOIN
         shipment s ON s.requisition_id = r.id
     WHERE
-        r.is_template IS FALSE 
-    UNION ALL SELECT 
+        r.is_template IS FALSE
+    UNION ALL SELECT
         o.id,
         o.name,
         o.order_number AS identifier,
@@ -51,29 +51,29 @@ CREATE OR REPLACE VIEW stock_movement_list_item AS
         s.id AS shipment_id,
         s.current_status AS shipment_status,
         CASE
-            WHEN o.status IS NULL THEN 'PENDING'
-            WHEN o.status = 'PENDING' THEN 'PENDING'
-            WHEN o.status = 'PLACED' THEN 'REQUESTED'
+            WHEN o.status IS NULL THEN 'CREATED'
+            WHEN o.status = 'PENDING' THEN 'CREATED'
+            WHEN o.status = 'PLACED' THEN 'CREATED'
             WHEN o.status = 'APPROVED' THEN 'PICKING'
             WHEN o.status = 'CANCELED' THEN 'CANCELED'
             # These probably won't be used but we want to cover them just in case
-            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'COMPLETED' THEN 'DISPATCHED'
+            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'COMPLETED' THEN 'ISSUED'
             WHEN o.status = 'REJECTED' THEN 'CANCELED'
             ELSE 'PENDING'
         END AS status_code,
         NULL AS requisition_id,
         CASE
-            WHEN o.status IS NULL THEN 'PENDING'
-            WHEN o.status = 'PENDING' THEN 'PENDING'
-            WHEN o.status = 'PLACED' THEN 'REQUESTED'
+            WHEN o.status IS NULL THEN 'CREATED'
+            WHEN o.status = 'PENDING' THEN 'CREATED'
+            WHEN o.status = 'PLACED' THEN 'CREATED'
             WHEN o.status = 'APPROVED' THEN 'PICKING'
             WHEN o.status = 'CANCELED' THEN 'CANCELED'
             # These probably won't be used but we want to cover them just in case
-            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'COMPLETED' THEN 'DISPATCHED'
+            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'COMPLETED' THEN 'ISSUED'
             WHEN o.status = 'REJECTED' THEN 'CANCELED'
             ELSE 'PENDING'
         END AS status,
--- a/grails-app/migrations/views/stock-movement.sql
+++ b/grails-app/migrations/views/stock-movement.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE VIEW stock_movement AS
-    SELECT 
+    SELECT
         r.id,
         r.name,
         r.request_number AS identifier,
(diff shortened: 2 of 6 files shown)


src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -63,6 +63,7 @@ class StockMovement implements Validateable{
 
     StockMovementDirection stockMovementDirection
     StockMovementStatusCode stockMovementStatusCode
+    StockMovementType stockMovementType
```
awalkowiak, line 66: Why this was added here if it is not populated (keep in mind this StockMovement is not backed by the stock-movement view)

grails-app/views/returns/show.gsp
```diff
@@ -130,7 +130,14 @@
                                 <g:message code="stockMovement.status.label"/>
                             </td>
                             <td class="value">
-                                <format:metadata obj="${stockMovement?.status ?: (stockMovement?.shipment?.status?.code ?: stockMovement?.statusCode)}"/>
+                                <g:set var="isInbound" value="${stockMovement?.destination == currentLocation}" />
+                                %{-- If it's an inbound return, we want to display a shipment status--}%
```
awalkowiak, line 134: I feel like this should not be here because we wanted to move away from this type of logic in templates. Hopefully, this won't have to be drastically changed, after getting the results of the spike that @drodzewicz is working on.

grails-app/views/returns/show.gsp
```diff
@@ -130,7 +130,14 @@
                                 <g:message code="stockMovement.status.label"/>
                             </td>
                             <td class="value">
-                                <format:metadata obj="${stockMovement?.status ?: (stockMovement?.shipment?.status?.code ?: stockMovement?.statusCode)}"/>
+                                <g:set var="isInbound" value="${stockMovement?.destination == currentLocation}" />
```
awalkowiak, line 133: You could (should) use a `stockMovementDirection` here (but first look at the comment from the line below)

grails-app/migrations/views/stock-movement-list-item.sql
```diff
@@ -51,29 +51,29 @@ CREATE OR REPLACE VIEW stock_movement_list_item AS
         s.id AS shipment_id,
         s.current_status AS shipment_status,
         CASE
```
awalkowiak, line 53: I wonder if we should not move away from this part and move it to the StockMovement/OutboundStockMovement/StockMovementListItem classes, but I think this is part of @drodzewicz spike

grails-app/migrations/views/stock-movement-list-item.sql
```diff
@@ -51,29 +51,29 @@ CREATE OR REPLACE VIEW stock_movement_list_item AS
...
-            WHEN o.status = 'RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'COMPLETED' THEN 'DISPATCHED'
+            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'COMPLETED' THEN 'ISSUED'
             WHEN o.status = 'REJECTED' THEN 'CANCELED'
             ELSE 'PENDING'
         END AS status_code,
         NULL AS requisition_id,
         CASE
```
awalkowiak, line 67: Same comment as above

grails-app/migrations/views/stock-movement.sql
```diff
@@ -77,27 +77,27 @@ CREATE OR REPLACE VIEW stock_movement AS
         s.current_status AS shipment_status,
         rn.identifier AS tracking_number,
         CASE
```
awalkowiak, line 79: Same comment as above

grails-app/migrations/views/stock-movement.sql
```diff
@@ -77,27 +77,27 @@ CREATE OR REPLACE VIEW stock_movement AS
...
-            WHEN o.status = 'RECEIVED' THEN 'DISPATCHED'
-            WHEN o.status = 'COMPLETED' THEN 'DISPATCHED'
+            WHEN o.status = 'PARTIALLY_RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'RECEIVED' THEN 'ISSUED'
+            WHEN o.status = 'COMPLETED' THEN 'ISSUED'
             WHEN o.status = 'REJECTED' THEN 'CANCELED'
             ELSE 'PENDING'
         END AS status_code,
         NULL AS requisition_id,
         CASE
```
awalkowiak, line 92: Same comment as above

awalkowiak (commented):
I am not sure if this is change request or comment

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
kchelstowski: It has been added in case we returned the `StockMovement` to the show.gsp, not the `OutboundStockMovement`.
Since this property was missing in the `StockMovement`, I would get an exception when trying to evaluate this part of the code:
```gsp
 <g:if test="${stockMovement?.stockMovementType == StockMovementType.RETURN_ORDER && isInbound}">
```
because either `StockMovement` or `OutboundStockMovement` might hide under the `stockMovement`.

I know it is not populated and it would return `null` for any case, we don't set it explicitly, but in this case we would want it to just evaluate to false, instead of throwing the exception about the missing method.

grails-app/views/returns/show.gsp
kchelstowski: I could for the `StockMovement`, but could not for the `OutboundStockMovement`, as this property/method is missing there.

grails-app/migrations/views/stock-movement-list-item.sql
kchelstowski: For now I just went with what was requested in the ticket, so to map the order statuses to different statuses than to what they were mapped until now.

grails-app/migrations/views/stock-movement-list-item.sql
jmiranda: Holy shit. This makes me very nervous and is the antithesis of what I wanted to do with these tickets. This is "possible long-term" refactor, not "just get it working" refactor.

src/main/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
```diff
@@ -29,8 +29,9 @@ enum StockMovementStatusCode {
     PACKED(7, PENDING),
     REVIEWING(8, PENDING),
     DISPATCHED(9),
+    ISSUED(9),
```
jmiranda, line 32: I don't really understand why we're replacing DISPATCHED, but if you're going to replace DISPATCHED with ISSUED then you need to go :100:. DISPATCHED was the name of the status I gave for the stock movement state machine to differentiate it from the requisition. A requisition is ISSUED. In the case of an outbound shipment when the requisition is ISSUED it's also DISPATCHED as part of a shipment. We could have also used SHIPPED but I wanted to be generic in case we were dispatching to a ward or pharmacy within the hospital, in which case there isn't a need for shipment metadata.

jmiranda: What is the reasoning behind adding an ISSUED enum here? Would this be rectified by changing the localized message rendered for DISPATCHED state? If not, can you elaborate on what we're trying to accomplish. 

Reminder: We want to change as little as possible.

jmiranda (commented):
Aside: The Tech Hurdle: State Machine document that I'm drafting is going to propose a change that will allow us to (hopefully) represent all of the state machines in the system using a library like Spring Statemachine https://docs.spring.io/spring-statemachine/docs/4.0.0/reference/index.html#crashcourse. We might keep the enums, but this will allow us to define all of the states and transitions of each state machine. That means we'll be calling the state machine API to check whether we can transition to the next state rather than implementing that using a switch statement.

grails-app/migrations/views/stock-movement-list-item.sql
kchelstowski: As for this change I’m pretty confident, since this part of the code is responsible only for returns.

src/main/groovy/org/pih/warehouse/inventory/StockMovementStatusCode.groovy
kchelstowski: It was a request in the ticket, that they want a different mapping for particular order statuses. Manon has provided me a table with exact statuses we want to display for a particular order status. If you read the comments in the ticket and also the description, everything should be self-explanatory. 
Why I added the ISSUED to the enum is because of the fact, that I was getting an exception on the list page when trying to sort, and where eventually one of the 10 items has the ISSUED status. The exception was „No enum value [ISSUED] for StockMovementStatusCode”

PR #4587, opened 2024-04-19
OBPIH-6215 Implement displayStatus method for returns with util, that…

kchelstowski:
… can be extended to other workflows

### DIFF
--- a/src/main/groovy/util/StockMovementStatusHelper.groovy
+++ b/src/main/groovy/util/StockMovementStatusHelper.groovy
@@ -1,0 +1,83 @@
+package util
+
+import org.pih.warehouse.api.StockMovementDirection
+import org.pih.warehouse.api.StockMovementType
+import org.pih.warehouse.auth.AuthService
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.order.Order
+import org.pih.warehouse.order.OrderStatus
+import org.pih.warehouse.requisition.RequisitionStatus
+import org.pih.warehouse.shipping.Shipment
+import org.pih.warehouse.shipping.ShipmentStatusCode
+
+class StockMovementStatusHelper {
+
+    Order order
+
+    Shipment shipment
+
+    StockMovementType stockMovementType
+
+    Location origin
+
+    Location destination
+
+    RequisitionStatus defaultStatus
+
+
+    StockMovementDirection getStockMovementDirection(Location currentLocation) {
+        if (currentLocation == origin) {
+            return StockMovementDirection.OUTBOUND
+        }
+        if (currentLocation == destination || origin?.isSupplier()) {
+            return StockMovementDirection.INBOUND
+        }
+        return null
+    }
+
+    boolean isInbound() {
+        Location currentLocation = AuthService.currentLocation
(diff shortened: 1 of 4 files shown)


src/main/groovy/util/StockMovementStatusHelper.groovy
```diff
@@ -0,0 +1,83 @@
...
+        }
+        if (currentLocation == destination || origin?.isSupplier()) {
+            return StockMovementDirection.INBOUND
+        }
+        return null
+    }
+
+    boolean isInbound() {
+        Location currentLocation = AuthService.currentLocation
+        return getStockMovementDirection(currentLocation) == StockMovementDirection.INBOUND
```
alannadolny, line 40: Basically, if the `currentLocation` is used only once, I don't see a need to create a new variable for that :thinking: 
```suggestion
        return getStockMovementDirection(AuthService.currentLocation) == StockMovementDirection.INBOUND
```

src/main/groovy/util/StockMovementStatusHelper.groovy
```diff
@@ -0,0 +1,83 @@
...
+
+    StockMovementDirection getStockMovementDirection(Location currentLocation) {
+        if (currentLocation == origin) {
+            return StockMovementDirection.OUTBOUND
+        }
+        if (currentLocation == destination || origin?.isSupplier()) {
+            return StockMovementDirection.INBOUND
+        }
+        return null
+    }
```
alannadolny, line 36: Wouldn't it be better to somehow adjust the code to use the method from `StockMovement`? Or it is not possible? Maybe you can make it `static` and call it by `StockMovement.getStockMovementDirection`?
image

alannadolny (commented):
And maybe it would be worth to write some quick unit tests for that helper class?

kchelstowski:
> And maybe it would be worth to write some quick unit tests for that helper class?

I wouldn’t say there is a sense to do it now, since this can be extended and refactored for Dariusz’s needs for regular SMs.

alannadolny:
Ok, that sounds good to me

src/main/groovy/util/StockMovementStatusHelper.groovy
kchelstowski: I think it’s better to keep it in the helper class, because otherwise I’d have to refactor the existing method in the StockMovement to receive the locations in the params etc. 
I think this helper class should be „a central place” for any type of DTOs we have (StockMovement, OutboundStockMovement, etc.)

src/main/groovy/util/StockMovementStatusHelper.groovy
kchelstowski: For me it looked cleaner to keep it in a separate variable. It might not make a sense from the „memory” POV, but readability is also important. 
If you guys prefer not creating a variable for that, then I’ll change it.

src/main/groovy/util/StockMovementStatusHelper.groovy
alannadolny: :+1:

alannadolny (approved):

PR #4588, opened 2024-04-19
OBPIH-5813 Items statuses before fulfilment - Approvals statuses

alannadolny:
I decided to avoid modifying the original function for getting statuses because it can introduce regression (as I saw it was in calculating line items in requisition, in this case, we based on our previous "APPROVED" status)

### DIFF
--- a/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
@@ -194,6 +194,37 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
         }
     }
 
+    RequisitionItemStatus getDisplayStatus() {
+        if (isParentRequisitionPending()) {
+            return RequisitionItemStatus.PENDING
+        }
+
+        if (isParentRequisitionRejected()) {
+            return RequisitionItemStatus.CANCELED
+        }
+
+        if (isApproved() || parentRequisitionItem) {
+            return RequisitionItemStatus.APPROVED
+        }
+
+        if (isSubstituted()) {
+            return RequisitionItemStatus.SUBSTITUTED
+        }
+
+        if (isChanged()) {
+            return RequisitionItemStatus.CHANGED
+        }
+
+        if (isCanceled()) {
+            return RequisitionItemStatus.CANCELED
+        }
+
+        if (isCompleted()) {
+            return RequisitionItemStatus.COMPLETED
+        }
+
+        return RequisitionItemStatus.PENDING
+    }
 
     /**
      * We currently only support quantity change and substitution so there will be,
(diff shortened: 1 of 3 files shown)


src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+import org.pih.warehouse.requisition.RequisitionItemStatus
+import org.pih.warehouse.requisition.RequisitionItemType
+import org.pih.warehouse.requisition.RequisitionStatus
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Unroll
+class RequisitionItemSpec extends Specification implements DomainUnitTest<RequisitionItem> {
+
+    void 'RequisitionItem.getDisplayStatus() should return: #requisitionItemStatus for requisition status #requisitionStatus'() {
```
EWaterman, line 16: This is nitpicky but in Spock you want to have your declarations like:
```
given:
<assign your fields, mocks and stubs here>

when:
<call the method you're trying to test here>

then:
<do your asserts here>
```

or if you're just doing a single simple assert you can do: 

```
given:
<assign your fields, mocks and stubs here>

expect:
<call the method you're trying to test here and do the asserts at the same time>
```


Also, the general rule for your where: block is:

```
where:
givenFieldX | givenFieldY || expectedResultZ
```

so the fields on the left of the || is what you feed into the method in your `given:` block and the fields on the right is what you verify in the `then:` or `expect:` blocks.

Here's an example of everything together:

```
        given:
        domain.requisition = new Requisition()
        domain.requisition.status = requisitionStatus

        expect:
        domain.displayStatus == requisitionItemStatus

        where:
        requisitionStatus                  || requisitionItemStatus
        RequisitionStatus.PENDING_APPROVAL || RequisitionItemStatus.PENDING
        ...
```

EWaterman:
nice rework! These tests are awesome :+1:

awalkowiak (approved):

grails-app/domain/org/pih/warehouse/requisition/RequisitionItem.groovy
```diff
@@ -194,6 +194,37 @@ class RequisitionItem implements Comparable<RequisitionItem>, Serializable {
         }
     }
 
+    RequisitionItemStatus getDisplayStatus() {
```
awalkowiak, line 197: @alannadolny An afterthought. Is the above `getStatus()` still used? Seems like this one is adding two additional ifs at the top. Could we remove the other one and leave only this?

Btw GJ with tests.

awalkowiak: (Or if we are afraid that we can break something here after the two new ifs we could just call the other method.)

alannadolny: @awalkowiak done

PR #4592, opened 2024-04-23
OBPIH-6342 Added accessibility attributes for e2e testing

drodzewicz:
- gsp status bars for error role=alert for regular role=status
- User creation helper testids
- user impersonate banner


drodzewicz:
most of the files contain the same change which is adding aria-label and role="status" or role="alert" for `flash.message` on gsp pages.

Even though for now it was only required for user creation workflow I decided to just add these attributes to all of the `flash.messages` & `flash.errors`

alannadolny (approved):

awalkowiak (approved):

PR #4599, opened 2024-04-24
OBPIH-5978 Improve the date shipped field to show a user who shipped the SM, not the one, who created it.

kchelstowski:


grails-app/views/returns/show.gsp
```diff
@@ -278,18 +278,15 @@
                         <tr class="prop">
                             <td class="name">
                                 <g:message code="stockMovement.dateShipped.label"/>
-                                <g:if test="${stockMovement?.shipment?.status?.code==org.pih.warehouse.shipping.ShipmentStatusCode.PENDING}">
-                                    <small><g:message code="default.expected.label"/></small>
-                                </g:if>
                             </td>
                             <td class="value">
-                                <g:if test="${stockMovement?.dateShipped}">
+                                <g:if test="${stockMovement?.shipment?.hasShipped()}">
```
EWaterman, line 283: is this changing to a boolean? That's what we want to do? Same question for the stock movement step.

kchelstowski: hasShipped() method that is called, is a method on Shipment domain, which returned type is Boolean (it checks if any event on the shipment contains an event with the SHIPPED eventTypeCode)

kchelstowski: Before, we were supposed to display the Date shipped even for an SM that was not shipped, this is why before we were checking the dateShipped (even though the dateShipped in this case might be misleading, because it’s not in fact the dateShipped, but the expectedDeliveryDate if I remember correctly).
Now we want to display this date with the user, only if an SM is really shipped.

grails-app/views/returns/show.gsp
```diff
@@ -278,18 +278,15 @@
...
                                     <span title="${g.formatDate(date:stockMovement?.dateShipped)}">
                                         <g:formatDate format="MMMM dd, yyyy" date="${stockMovement?.dateShipped}"/>
                                     </span>
-                                    <g:if test="${stockMovement?.shipment?.createdBy}">
+                                    <g:if test="${stockMovement?.shipment?.shippedBy || stockMovement?.shipment?.createdBy}">
```
awalkowiak, line 287: Can you move this fallback to created by to the shipment's shippedBy method? It could be something like `if there is no shipped event and there is a case we want to return created by, then it would be returned`.

grails-app/views/returns/show.gsp
EWaterman: that makes sense. Thanks for the explanation!

EWaterman (approved):

grails-app/views/returns/show.gsp
kchelstowski: @awalkowiak I guess it should be `if there IS shipped event`, because otherwise we don't want to display any user (until the SM is shipped). 
I'm gonna push the change and let me know if we are on the same page

awalkowiak (approved):

PR #4602, opened 2024-04-25
OBPIH-6327 Data model should support "to pick" and "picked" stages

alannadolny:


grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy
```diff
@@ -40,7 +40,8 @@ class PicklistItem implements Serializable {
     InventoryItem inventoryItem
     Location binLocation
 
-    Integer quantity
+    Integer quantity // quantity = quantity picked
+    Integer quantityToPick
```
EWaterman, line 44: you don't need to add the field to the toJson method?

jmiranda: We need to add quantityPicked. The quantity property should be quantity to pick.

See OBKN https://github.com/openboxes/openboxes/blob/feature/obkn-integration/grails-app/domain/org/pih/warehouse/picklist/PicklistItem.groovy

jmiranda: I'd like to bring in some of the other fields (like picker and datePicked) as well. 

image

jmiranda: > you don't need to add the field to the toJson method?

Oh and to answer your question, Evan, we don't need to right now since in the PIH world quantity to pick (quantity) always equals quantity picked (quantityPicked). So essentially allocated and picked happen at exactly the same. This is not the ideal since there are important states and exceptions that occur between Allocated and Picked, but it's fine for PIH's default use case. At least for now. In addition, adding a state machine for the picking workflow would add overhead that would not likely not benefit for PIH until these events were important to track. That would change the minute PIH decided that it was important to allow short picks / overpicks or place a hold on a bin location when an item was not found in that location. 

Fwiw, the OBKN API (and the mobile app built on top) support a picking state machine that allows for a more complex picking workflow (including exceptions like partial pick, short picks, overpick, reallocation, etc). 

A typical state machine transition for a pick item might look like this.
```
Allocated -> To Pick -> Picking -> Picked
```

awalkowiak (commented):
Sorry for the confusion, I took wording from the ticket as the way we want to have it and added a comment around this (plus I was kinda influenced by the "not take look to obkn too much", and potentially it feels like we might need to change more places in future for example we will have to adjust the `getQuantityPickedByProductAndLocation` in `ProductAvailabilityService`, and probably few other places like that). I don't have an issue with having it the way it was. 

Additional to dos / next steps: 
- (as we discussed on slack internally) I think it would be worth to add to the migration a changeset changing the data from the past (that would at least populate quantity picked with the quantity field), especially if we are going to change the existing logic in other places to use `quantityPicked` instead of `quantity`
- as Even suggested please add these new fields to toJson
- I think you could add for now in the `createOrUpdatePicklistItem`:
```picklistItem.quantityPicked = quantity``` (that one is not that required, because this will be probably adjusted eitherway in the next ticket from this epic)
- you could also try to find places where the quantity from PicklistItem is being used, to prepare a list where it will have to be adjusted in the next and we can discuss on the tech huddle channel where and when we have to change it

alannadolny:
So, at this moment:
- I renamed the column `quantity` to `quantity_picked`
- I added a new column `quantity`, which has the data from the previous `quantity` column. In the case of new picklist items, it will be 0
- I added columns: picker_id and date_picked
- I added those property to the `toJson` method
- I changed the calls of `quantity` to  `quantityPicked`, I am not feeling comfortable with that, but this change is caused by the field renaming which was requested after the discussion on Slack / under PR (I am not sure If I changed it in every place, because it is quite hard to find all the occurrences)

PR #4604, opened 2024-04-27
OBPIH-6045 Updated openboxes.dbml file using schema-export and sql2db…

jmiranda:
…ml (see #4094)

### DIFF
--- a/openboxes.dbml
+++ b/openboxes.dbml
@@ -1,2719 +1,2429 @@
-Project openboxes {
-  database_type: 'MySQL'
-  Note: 'Data model for latest development version of OpenBoxes'
-}
-
-Table "DATABASECHANGELOG" {
-  "ID" varchar(63) [not null]
-  "AUTHOR" varchar(63) [not null]
-  "FILENAME" varchar(200) [not null]
-  "DATEEXECUTED" datetime [not null]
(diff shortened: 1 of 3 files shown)


jmiranda:
There's no need to comment on the PR, but I've added you as reviewers to brainstorm what we can do to automate the process documented in #4094. I don't want to automate committing changes to the dbml file here in the repo. I'd rather export that to GH packages or some temporary file store and then have github actions pull that artifact when running dbdocs.

EWaterman (approved):

openboxes.dbml
```diff
@@ -1,2719 +1,2429 @@
...
-  "ID" int(11) [pk, not null]
-  "LOCKED" tinyint(1) [not null]
-  "LOCKGRANTED" datetime [default: NULL]
-  "LOCKEDBY" varchar(255) [default: NULL]
-}
-
 Table "address" {
-  "id" char(38) [pk, not null, default: ""]
-  "version" bigint(20) [not null]
+  "id" varchar(255) [pk, not null]
```
jmiranda, line 2: Bummer. I just noticed that Hibernate's schema-export does do exactly what our database migrations do, so I'm going to need to run the migrations (maybe on a testcontainer) and then use mysqldump to get a more precise DDL file to convent to DBML.

PR #4607, opened 2024-04-29
OBPIH-6327 Data model should support "to pick" and "picked" stages

alannadolny:


awalkowiak (approved):

EWaterman (approved):

grails-app/migrations/0.9.x/changelog-2024-04-29-1200-alter-table-picklist-item-add-quantity-picked.groovy
```diff
@@ -0,0 +1,56 @@
...
+        preConditions(onError: "HALT", onFail: "MARK_RAN", onSqlOutput: "IGNORE") {
+            not {
+                foreignKeyConstraintExists(foreignKeyName: "fk_picker_person")
+            }
+        }
+
+        addForeignKeyConstraint(
+                baseColumnNames: "picker_id",
+                baseTableName: "picklist_item",
+                constraintName: "fk_picker_person",
```
awalkowiak, line 49: One thing that caught my eye before merging it. Could you change the name of this constraint to match a convention of having names of base table name and referenced table name -> `fk_picklist_item_person`

awalkowiak (changes requested):

grails-app/migrations/0.9.x/changelog-2024-04-29-1200-alter-table-picklist-item-add-quantity-picked.groovy
awalkowiak: Actually, your version makes more sense, let me think shortly about this one

awalkowiak: Yeah I was partially right. We usually do `fk + base table name + base column name (without _id)` for these constraint names, so I think this could be the way for this one, but your version `fk + base column name + reference table name` is ok too I guess

awalkowiak: As discussed internally on slack we are going to use a convention of `fk_baseTable_baseColumn_referenceTable` if the changelogs are not generated.

awalkowiak (approved):

jmiranda (commented):
Sorry, just got to this now. LGTM.

jmiranda (commented):
... although I'm starting to second guess myself on the naming of the picker property. Should have probably been pickedBy to stay consistent with other fields.

PR #4610, opened 2024-05-06
OBPIH-6153 Add ability to clear pick

kchelstowski:


src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -617,6 +622,25 @@ class PickPage extends Component {
     });
   }
 
+  async clearPicklist() {
+    const { values } = this.state;
+    const { picklist } = values;
```
alannadolny, line 627: Can't you just do:
```suggestion
    const { picklist } = this.state.values;
```
I do not see `values` in the code :thinking:

src/js/utils/apiClient.jsx
```diff
@@ -65,7 +65,10 @@ export const handleError = (error) => {
   const errorMessage = _.get(error, 'response.data.errorMessage', '');
   const errorMessages = _.get(error, 'response.data.errorMessages', []).join(', ');
   switch (error.response.status) {
+
     case 400: {
+      console.log("error message", errorMessage)
+      console.log("error messages", errorMessages)
```
alannadolny, line 71: :smile:

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
kchelstowski: yeah, you are right. Previously I used `stockMovementId` so I had to do it this way, and when changing it to `picklist`, I have not noticed I could shorten it. Thanks.

grails-app/controllers/org/pih/warehouse/api/PicklistApiController.groovy
```diff
@@ -0,0 +1,14 @@
...
+import org.pih.warehouse.picklist.PicklistService
+
+class PicklistApiController extends BaseDomainApiController {
+
+    PicklistService picklistService
+
+    def clearPicklist() {
+        picklistService.clearPicklist(params.id)
+
+        render status: 204
```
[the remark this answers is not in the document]
jmiranda: Here's a stackoverflow about deleting "some" of a resource collection, but it doesn't discuss response codes. 
https://stackoverflow.com/questions/37138720/restful-api-delete-some-of-a-resource-collection

If anyone finds good documentation on response codes for deleting a resource collection and in particular  first vs subsequent requests with different results i.e. deleted X items vs. resource collection has no items vs deleted 0 items because there were filters applied to the nested resource collection

* /api/picklists/:id/items
* /api/picklists/:id/items
* /api/picklists/:id/items?filter=value

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -86,6 +87,34 @@ class PicklistService {
...
+        }
+        itemsToRemove.each {
+            picklist.removeFromPicklistItems(it)
+        }
+
+
+        productAvailabilityService.refreshProductsAvailability(
+                picklist?.requisition?.origin?.id,
+                picklist?.requisition?.requisitionItems?.product?.id,
+                binLocations.unique(),
```
jmiranda, line 114: Make binLocations a Set<String> and we probably don't need to deal with the unique issue.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -86,6 +87,34 @@ class PicklistService {
...
+            binLocations.add(picklistItem.binLocation?.id)
+        }
+        itemsToRemove.each {
+            picklist.removeFromPicklistItems(it)
+        }
+
+
+        productAvailabilityService.refreshProductsAvailability(
+                picklist?.requisition?.origin?.id,
+                picklist?.requisition?.requisitionItems?.product?.id,
```
jmiranda, line 113: This is a bit misleading at first glance. I think it's doing this implicitly
```
picklist?.requisition?.requisitionItems?.collect { it?.product?.id } 
```
so let's make that explicit so it's slightly more readable and less confusing.
```
Set<String> productIds = picklist?.requisition?.requisitionItems?.collect { it?.product?.id } 
productAvailabilityService.refreshProductsAvailability(picklist?.requisition?.origin?.id,
                productIds, binLocations, false)
```
Note: The collect method might not allow the return to be a set, so if necessary turn it into a List and add .unique() at the end.

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -862,4 +864,10 @@ class StockMovementApiController {
         }
         render(status: 200)
     }
+
+    def clearPicklist() {
+        picklistService.clearPicklist(params.stockMovementId)
+
+        render(status: 204)
+    }
```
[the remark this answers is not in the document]
awalkowiak: If this is not needed right now, and we are going to rely only on the picklist API, then let's remove this for now.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -86,6 +87,34 @@ class PicklistService {
         productAvailabilityService.refreshProductsAvailability(orderItem?.order?.origin?.id, [orderItem?.product?.id], binLocations?.unique(), false)
     }
 
+    void clearPicklist(String id) {
```
awalkowiak, line 90: This is the second `void clearPicklist(String id)` in the code. The other one is in the `StockMovementService`. Please add a comment explaining that the parameter with id here is meant to be picklistItem id, and in the other one that it expects stock movement id. I would even consider deprecating the other one, but we have not discussed it so it is just a thought.

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -128,18 +129,22 @@ const FIELDS = {
...
+            <div className={!value ? 'text-danger' : null}>
+              {value.toLocaleString('en-US')}
+            </div>
+          ),
         },
       },
       buttonEditPick: {
-        label: 'react.stockMovement.editPick.label',
-        defaultMessage: 'Edit pick',
+        label: 'react.stockMovement.pick.label',
```
awalkowiak, line 140: There is one unknown here. Initially I thought this should be changed to pick only when we trigger clear picklist, but let's leave it as is for now, and let's see what Manon will say about that part.

awalkowiak (changes requested):

grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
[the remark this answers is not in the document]
kchelstowski: yeah, sorry for the confusion, I just forgot to remove it from there, after moving it to the `PicklistApiController`

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
kchelstowski: Yeah, this is doing the `.collect` behind the scenes. I've just used "groovy magic" :D but I'm fine with doing it implicitly

grails-app/controllers/org/pih/warehouse/api/PicklistApiController.groovy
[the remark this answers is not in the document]
kchelstowski: @jmiranda Definitely topic worth to extend, but from my understanding, since we do not return anything in the response, we should be fine with the `204`, since this is "No Content", even though we delete a collection.
I've always said I was "pro" a silent delete - so that e.g. even if there is nothing to delete or a resource of given id is not found, we still return 204.
This is about idempotency of HTTP methods. Looking at many resources, it feels like DELETE is idempotent, hence it should return the same response for a single request even if called many times, so in our case - even if we attempt to delete an empty collection, we should still get 204.
https://stackoverflow.com/a/6440374/23546798

> So regardless of whether you DELETE an existing resource, or attempt to DELETE a resource that does not exist, the server resource state is the same.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
kchelstowski: I will add the comments, but imho it's kinda self-explonatory that if I use the `PicklistService`, we are rather referring to the picklist id (unless the argument was named differently), and the same with the `StockMovementService` - if I use `id` as an argument, it's rather the SM id.

awalkowiak: Technically it is self-explanatory, but I feel like there was a lot of self-explanatory code in this repo, that is now not that self-explanatory :sweat_smile:

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
awalkowiak: I got confirmation, that your current version is correct. You can disregard the previous comment here.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
kchelstowski: It can't be a Set, because the product availability method requires a List, and it breaks, if the binLocations become the Set. I think refactoring it now (the PA method), or creating another overloaded method with Set instead of List does not make sense, unless you feel strong about it.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -86,6 +87,37 @@ class PicklistService {
...
+        }
+        // Store bin locations' id for product availability refresh
+        List<String> binLocations = []
+
+        Set<PicklistItem> picklistItems = picklist.picklistItems
+        List<PicklistItem> itemsToRemove = []
+        picklistItems.each { PicklistItem picklistItem ->
+            picklistItem.disableRefresh = true
+            itemsToRemove.add(picklistItem)
+            binLocations.add(picklistItem.binLocation?.id)
```
EWaterman, line 105: won't this add a null to the list if there's no binLocation? What's going to happen in the productAvailability if that happens?

kchelstowski: actually we have a use case for the `null` binLocation. It's the Default bin location. It's quite difficult to understand on first sight, so maybe either Justin or Artur could explain it in more details if you were curious.

kchelstowski: Essentially, looking at the stockcard of a product:
Screenshot from 2024-05-08 15-45-54

the first inventory item's binLocation is `null`.

EWaterman: gotcha. Thanks for the explanation!

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
awalkowiak: To be honest I expected something more like a javadoc with `[at]param`, but that can be adjusted in the next pr from that epic.

awalkowiak (approved):

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
kchelstowski: yeah, you are right, I will try to remember to change it to the javadoc.

PR #4611, opened 2024-05-07
OBPIH-6363 Validation and items requiring attention after clearing pick

alannadolny:
This one has to be merged after merging OBPIH-6153

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/PickPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/PickPage.jsx
@@ -348,7 +348,7 @@ class PickPage extends Component {
     return apiClient.get(url)
       .then((resp) => {
         const { totalCount } = resp.data;
-        const { associations } = resp.data.data;
+        const { associations, picklist } = resp.data.data;
         const printPicks = _.find(
           associations.documents,
           doc => doc.documentType === 'PICKLIST' && doc.uri.includes('print'),
@@ -357,6 +357,10 @@ class PickPage extends Component {
           totalCount,
           printPicksUrl: printPicks ? printPicks.uri : '/',
           sorted: false,
+          values: {
+            ...this.state.values,
+            picklist,
+          },
         }, () => this.props.hideSpinner());
       })
       .catch(() => this.props.hideSpinner());
@@ -567,7 +571,7 @@ class PickPage extends Component {
   validateEmptyPicks(formValues) {
     const emptyPicks = this.getRowsWithEmptyPicks(formValues.pickPageItems);
 
-    const pickPageItems = this.state.values.pickPageItems.map((item, index) => {
+    const pickPageItems = formValues.pickPageItems.map((item, index) => {
       if (emptyPicks.includes(index + 1)) {
         return { ...item, hasError: true };
       }
@@ -756,18 +760,24 @@ class PickPage extends Component {
                   className="float-right mb-1 btn btn-outline-secondary align-self-end btn-xs ml-1"
                 >
                   <span>
-                    <i className="fa fa-refresh pr-2"/>
-                    <Translate id="react.stockMovement.clearPick.label"
-                               defaultMessage="Clear pick"/>
+                    <i className="fa fa-refresh pr-2" />
+                    <Translate
+                      id="react.stockMovement.clearPick.label"
+                      defaultMessage="Clear pick"
+                    />
                   </span>
                 </button>
                 <label
                   htmlFor="csvInput"
                   className="float-right mb-1 btn btn-outline-secondary align-self-end ml-1 btn-xs"
                 >
-                  <span><i className="fa fa-download pr-2"/><Translate
-                    id="react.default.button.importTemplate.label"
-                    defaultMessage="Import template"/></span>
+                  <span>
+                    <i className="fa fa-download pr-2" />
+                    <Translate
+                      id="react.default.button.importTemplate.label"
+                      defaultMessage="Import template"
+                    />
+                  </span>
                   <input
                     id="csvInput"
                     type="file"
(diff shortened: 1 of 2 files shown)


src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -219,6 +219,7 @@ class PickPage extends Component {
       showAlert: false,
       alertMessage: '',
       itemFilter: '',
+      showOnlyErroredItems: false,
```
kchelstowski, line 222: wouldn't errored -> invalid work better?

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -525,6 +535,60 @@ class PickPage extends Component {
...
+      `The picked lot is missing at rows: ${emptyPickLinesNumber}`,
+      {
+        rows: emptyPickLinesNumber,
+      },
+    );
+
+    this.setState({
+      alertMessage,
+      showAlert: emptyPicks.length,
+    });
```
kchelstowski, line 566: I assume this is currently done like this (using `showAlert` flag), but it feels weird to me that we are eventually assinging an `alertMessage` to the state and in the end we are not displaying it.
Imho from the "logical point of view" we should do:
```js
if (emptyPicks.length) {
  this.setState({
      alertMessage,
      showAlert: true,
    });
}
```
but since you are not the one introducing the `showAlert` flag, treat it just as whining

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
alannadolny: It was also named like that previously, so I would stick to this name at this moment.

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -525,6 +535,60 @@ class PickPage extends Component {
       .catch(() => { this.props.hideSpinner(); });
   }
 
+  // Calculating which rows have quantity picked 0, and don't have subitems
```
awalkowiak, line 538: Please either add in the comment that the returned value is a list of indexes, not rows or change the name to indicate it returns row indexes.

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -641,39 +705,80 @@ class PickPage extends Component {
...
+                <button
+                  type="button"
+                  onClick={() => {
+                    this.validateEmptyPicks(values);
+                    this.setState({ showOnlyErroredItems: !showOnlyErroredItems });
+                  }}
+                  className={`float-right mb-1 btn btn-outline-secondary align-self-end btn-xs ml-3 ${showOnlyErroredItems ? 'active' : ''}`}
+                >
+                    <span>
+                      {this.getRowsWithEmptyPicks(values?.pickPageItems).length}
```
awalkowiak, line 750: Shouldn't this number be dynamically adjusted when the user is fixing lines?

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -641,39 +705,80 @@ class PickPage extends Component {
...
               <span className="buttons-container">
                 <FilterInput
                   itemFilter={itemFilter}
                   onChange={e => this.setState({ itemFilter: e.target.value })}
                   onClear={() => this.setState({ itemFilter: '' })}
                 />
+                <button
+                  type="button"
+                  onClick={() => {
+                    this.validateEmptyPicks(values);
```
awalkowiak, line 744: Is this required here?

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
alannadolny: it is

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
alannadolny: Yup, because this function sets `hasError` - an indicator of whether we have to hide/show a row

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -525,6 +535,60 @@ class PickPage extends Component {
       .catch(() => { this.props.hideSpinner(); });
   }
 
+  // Returns indexes of rows with quantity picked 0, and without subitems.
+  getRowsWithEmptyPicks(pickPageItems) {
+    return pickPageItems.reduce((acc, item, index) => {
+      // When the quantity picked is equal to 0 we have to check its subitems,
+      // because the item can be edited to 0, not cleared.
+      if (!item.quantityPicked && !item.picklistItems.length) {
+        return [...acc, index + 1];
```
drodzewicz, line 544: what is the reason for `index + 1` ?
As I am reading a function and the comment we get a list of indexes where quantity is zero or empty and we don't have pickListItems, but we are appending an index of the next item :thinking:

drodzewicz: I think I am going to answer this myself...
 this is for user readability 
"The picked lot is missing at rows: 1,2,3 ..."
instead of
"The picked lot is missing at rows: 0, 1,2 ..."

drodzewicz, line 544:
```suggestion
  getIndexesOfRowsWithEmptyPicks(pickPageItems) {
    return pickPageItems.reduce((acc, item, index) => {
      // When the quantity picked is equal to 0 we have to check its subitems,
      // because the item can be edited to 0, not cleared.
      if (!item.quantityPicked && !item.picklistItems.length) {
        return [...acc, index];
```

I saw Arturs comment suggesting adding a comment which you did, but I feel strongly that this should have a function name with clear intent, this should say that it returns indexes (normal indexes, 0,1,2,3...) and when we need to display these indexes to a user just do `indexes.map(it => it+1)`

I saw you use this same function later where you had to do this `indexes.includes(index + 1)` which doesn't sit right with me :/ 
Also the name might be confusing since `getRowsWithEmptyPicks` suggests that we return `rows` (object of values)

drodzewicz (changes requested):

drodzewicz (approved):

awalkowiak (approved):

PR #4614, opened 2024-05-09
OBGM-328 unit test cases for Grails 3

chetanmaharishi:
Please follow this PR also for previous commit
https://github.com/openboxes/openboxes/pull/4608

Now i have create one DomainUnit test case for TransactionType domain.

### DIFF
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -26,6 +26,7 @@ import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Event
 import org.pih.warehouse.core.EventCode
 import org.pih.warehouse.core.EventType
+import org.pih.warehouse.core.IdentifierService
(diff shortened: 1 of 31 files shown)


jmiranda:
@chetanmaharishi  As we discussed on Slack can you please create a PR for just the TransactionTypeSpec (including any other changes that it depends on). You can keep this branch lying around as we'll probably come back to it once we have a chance to review the reference implementation. I am expecting that we'll be able to go through each @ignore'd test and convert them one by one, following the conventions laid out by the TransactionTypeSpec.

cc @EWaterman

ewaterman:
Heads up that I'm going through each of these tests as I get time to slowly start reviving this branch

ewaterman:
whew okay I finally wrapped this app and now we have a ton more unit tests running.

It's a bit of a mess to look at since I renamed and moved a bunch of files but all the refactored tests now look like "proper" spock tests and are all passing.

@jmiranda as an FYI, since this PR only test changes I'm going to merge it in as is. In case you have any thoughts (I don't expect a review), feel free to leave comments here and we can address

PR #4617, opened 2024-05-14
OBPIH-6363 Validation and items requiring attention after clearing pick (Improvement after QA)

alannadolny:


src/js/components/stock-movement-wizard/outbound/PickPage.jsx
```diff
@@ -721,6 +727,8 @@ class PickPage extends Component {
       alertMessage,
     } = this.state;
     const { showOnly } = this.props;
+    const emptyPicksQuantity =
```
awalkowiak, line 730: `Count` or `Size` might be more suitable, because `Quantity` might mislead that this is some sort of item quantity.

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/outbound/PickPage.jsx
alannadolny: :+1:

awalkowiak (approved):

kchelstowski (approved):

PR #4619, opened 2024-05-14
Unit domain test case for TransactionType.

chetanmaharishi:
Unit test case for Transaction Type

### DIFF
--- a/src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
@@ -1,0 +1,91 @@
+package org.pih.warehouse.inventory
+
+import spock.lang.Unroll
+
+import grails.testing.gorm.DomainUnitTest
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class TransactionTypeSpec  extends Specification implements DomainUnitTest<TransactionType> {
+
+
+    @Test
+    @Unroll('TransactionType.validate() with name: #value should return #expected with errorCode: #expectedErrorCode')
+    void 'name field validation'() {
+        when:
+        domain.name = value
+
+        then:
+        assert domain.validate(['name']) == expected
+        assert domain.errors.getFieldError("name")?.getCode() == expectedErrorCode
+
+        where:
+        value                                             || expected | expectedErrorCode
+        null                                              || false    | 'nullable'
+        'this_is_big_string_to_test_max_length_255' * 255 || false    | 'maxSize.exceeded'
+        'Dummy Name'                                      || true     | null
+    }
+
+    @Test
+    @Unroll('TransactionType.validate() with description: #value should return #expected with errorCode: #expectedErrorCode')
+    void 'description field validation'() {
+        when:
+        domain.description = value
+
+        then:
+        assert domain.validate(['description']) == expected
+        assert domain.errors.getFieldError("description")?.getCode() == expectedErrorCode
+
+        where:
+        value                                             || expected | expectedErrorCode
+        null                                              || true     | null
+        'this_is_big_string_to_test_max_length_255' * 255 || false    | 'maxSize.exceeded'
+        'Dummy description'                               || true     | null
+    }
+
+    @Test
+    void 'all fields validation'() {
+        when:
+        domain
+
+        then:
+        assert !domain.validate()
+        assert domain.errors.allErrors.size() == 2
+    }
+
+    @Test
+    @Unroll('TransactionType.validate() with transactionCode: #value should return #expected with errorCode: #expectedErrorCode')
+    void 'transactionCode field validation'() {
+        when:
+        domain.transactionCode = value
+
+        then:
+        assert domain.validate(['transactionCode']) == expected
+        assert domain.errors.getFieldError("transactionCode")?.getCode() == expectedErrorCode
+
+        where:
+        value                 || expected | expectedErrorCode
+        null                  || false    | 'nullable'
+        TransactionCode.DEBIT || true     | null
+    }
+
+    @Test
+    @Unroll('TransactionType.isAdjustment() with isAdjustment: #value should return #expected')
+    void 'isAdjustment field validation'() {
+        when:
+        domain.name = value
+
+        then:
+        domain.isAdjustment() == expected
+
+        where:
+        value           || expected
+        'NO_Adjustment' || false
+        'Adjustment'    || true
+    }
+
+
+}


src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+
+        where:
+        value                                             || expected | expectedErrorCode
+        null                                              || true     | null
+        'this_is_big_string_to_test_max_length_255' * 255 || false    | 'maxSize.exceeded'
+        'Dummy description'                               || true     | null
+    }
+
+    @Test
+    void 'all fields validation'() {
```
EWaterman, line 50: what does the empty domain do here? What case is this testing?

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+        when:
+        domain.name = value
+
+        then:
+        domain.isAdjustment() == expected
+
+        where:
+        value           || expected
+        'NO_Adjustment' || false
+        'Adjustment'    || true
```
EWaterman, line 87: isn't it supposed to be of the format 'Adjustment|name'? So something like

```
value                      || expected
'testName'                 || false
'NoAdjustment|testName'    || false
'Adjustment|testName'      || true
```

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+        domain.description = value
+
+        then:
+        assert domain.validate(['description']) == expected
+        assert domain.errors.getFieldError("description")?.getCode() == expectedErrorCode
+
+        where:
+        value                                             || expected | expectedErrorCode
+        null                                              || true     | null
+        'this_is_big_string_to_test_max_length_255' * 255 || false    | 'maxSize.exceeded'
```
EWaterman, line 45: could also make it `'x' * 256` so that you're testing the actual max length value

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
```diff
@@ -0,0 +1,91 @@
+package org.pih.warehouse.inventory
+
+import spock.lang.Unroll
+
+import grails.testing.gorm.DomainUnitTest
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
```
EWaterman, line 8: please remove these junit imports. Also please remove the @Test annotations from each of the tests. You don't need them in Spock

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
chetanmaharishi: we are testing for empty domain. so it can throw error for all fields that has some validation. So in future if add any new field that can be catch here.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
chetanmaharishi: it can be. 
Basically it trying to find Adjustment field in string. so if you suggest i can make changes.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class TransactionTypeSpec  extends Specification implements DomainUnitTest<TransactionType> {
+
+
+    @Test
+    @Unroll('TransactionType.validate() with name: #value should return #expected with errorCode: #expectedErrorCode')
+    void 'name field validation'() {
```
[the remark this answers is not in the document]
jmiranda: @EWaterman I made some modifications to the tests and this is what the output looks like in the console. 

image

and the test report 

image

If we don't want data (especially long names) in the test spec description, then we could probably break the tests into "test with valid data" vs "tests against invalid data"

So, instead of this 
```
org.pih.warehouse.inventory.TransactionTypeSpec > Should return false with error code nullable when validating transaction type given name null PASSED

org.pih.warehouse.inventory.TransactionTypeSpec > Should return false with error code maxSize.exceeded when validating transaction type given name aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa PASSED

org.pih.warehouse.inventory.TransactionTypeSpec > Should return true with error code null when validating transaction type given name bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb PASSED
```
we'd see this
```
org.pih.warehouse.inventory.TransactionTypeSpec > Should return false with error code nullable when validating transaction type given an invalid name PASSED

org.pih.warehouse.inventory.TransactionTypeSpec > Should return false with error code maxSize.exceeded when validating transaction type given an invalid name PASSED

org.pih.warehouse.inventory.TransactionTypeSpec > Should return true with error code null when validating transaction type given a valid name PASSED
```
Or we could add a test description to the table 
```
adjective    || value        || expected | expectedErrorCode
null         || null         || false    | 'nullable'
really long  || 'a' * 256    || false    | 'maxSize.exceeded'
longish      || 'b' * 255    || true     | null
valid        || 'Dummy Name' || true     | null
```
So the test spec
```
def 'Should return #expected with error code #expectedErrorCode when validating transaction type given a #adjective name'() {
```
would resolve to 
```
org.pih.warehouse.inventory.TransactionTypeSpec > Should return false with error code maxSize.exceeded when validating transaction type given a really long PASSED
```

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
jmiranda: Basically, so we can check all of the errors at once in case we accidentally added a nullable:false validation that isn't being tested elsewhere. This should also catch validation that was removed. We should make sure this test includes a check on the actual fields / error codes returned in the Errors object, not just the count.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
jmiranda: I fixed this yesterday.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
jmiranda: I fixed this as well.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
jmiranda: done

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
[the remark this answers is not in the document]
EWaterman: yeah I agree in that I don't think we necessarily need to be ultra-verbose in specifying the exact inputs in the test name, but also I'm not sure if I'm a fan of needing to specify additional adjective fields just to populate the test name properly. I'm not sure it's the hugest deal either way. Ultimately I feel like as long as we stick to a general "X should return/do Y when Z" format, we're good.

src/test/groovy/unit/org/pih/warehouse/inventory/TransactionTypeSpec.groovy
EWaterman: okay that makes sense. I like the idea. It forces you to consider tests when you make a schema change. Though I agree it's probably a good idea to at least assert on each of the errors instead of just the number of errors

jmiranda: Yes, that makes sense as you may have removed one validation rule and added another.

chetanmaharishi:
These are final results after running test cases

1. org.pih.warehouse.inventory.TransactionTypeSpec > should return false with error code nullable when validating transaction type given name null PASSED

2. org.pih.warehouse.inventory.TransactionTypeSpec > should return false with error code maxSize.exceeded when validating transaction type given name aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa PASSED

3. org.pih.warehouse.inventory.TransactionTypeSpec > should return true with error code null when validating transaction type given name bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
bbbbbbbbbbbbbbbbbbbbb PASSED

4. org.pih.warehouse.inventory.TransactionTypeSpec > should return true with error code null when validating transaction type given name Dummy Name PASSED

5. org.pih.warehouse.inventory.TransactionTypeSpec > should return true with error code null when validating transaction type given description null PASSED

6. org.pih.warehouse.inventory.TransactionTypeSpec > should return false with error code maxSize.exceeded when validating transaction type given description aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa PASSED

7. org.pih.warehouse.inventory.TransactionTypeSpec > should return true with error code null when validating transaction type given description Dummy description PASSED

8. org.pih.warehouse.inventory.TransactionTypeSpec > should expect two validation errors when validating given an empty transaction type PASSED

9. org.pih.warehouse.inventory.TransactionTypeSpec > should expect the names to be the same PASSED

10. org.pih.warehouse.inventory.TransactionTypeSpec > should return false with error code nullable when validating transaction type given transactionCode null PASSED

11. org.pih.warehouse.inventory.TransactionTypeSpec > should return true with error code null when validating transaction type given transactionCode DEBIT PASSED

12. org.pih.warehouse.inventory.TransactionTypeSpec > should return true when given name Adjustment|fr:Ajustement|es:Ajustamiento PASSED

13. org.pih.warehouse.inventory.TransactionTypeSpec > should return false when given name NO_Adjustment PASSED

14. org.pih.warehouse.inventory.TransactionTypeSpec > should return true when given name Adjustment PASSED

15. org.pih.warehouse.inventory.TransactionTypeSpec > should return Wed May 29 22:43:35 IST 2024 and Wed May 29 22:43:35 IST 2024 when given name Testing Name and transactionCode DEBIT and
 dateCreated Wed May 29 22:43:35 IST 2024 and lastUpdated Wed May 29 22:43:35 IST 2024 PASSED

PR #4620, opened 2024-05-15
OBPIH-6356 Fixed bug causing fatal exception if sentry.dsn is not pro…

jmiranda:
…vided; added default sentry.properties file to classpath; removed old sentry plugin config;


src/main/resources/sentry.properties
```diff
@@ -0,0 +1 @@
+dsn=
```
EWaterman, line 1: does this work?

EWaterman (approved):

awalkowiak (approved):

src/main/resources/sentry.properties
jmiranda: Yes, it's the only thing that does.

jmiranda: I was planning to move all of the default config from SentryServletContainerInitializer to this file, but given how much of a pain in the neck this was I didn't want to do more than what was required. So I'll add a ticket to handle that and figuring out a way to get the SentryServletContainerInitializer to work in all environments or replace it with custom init method in BootStrap.groovy.

PR #4624, opened 2024-05-16
OBPIH-6368 Fix statuses mapping on outbound list page while filtering after state machine refactor + remove unused filters

kchelstowski:
… after state machine refactor + remove unused filters

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
@@ -46,6 +46,8 @@ class OutboundStockMovementListItem implements Serializable, Validateable {
     RequisitionSourceType sourceType // temporary sourceType field for ELECTRONIC and PAPER types
 
     StockMovementType stockMovementType
+
+    @Deprecated
     StockMovementStatusCode statusCode
 
     Requisition requisition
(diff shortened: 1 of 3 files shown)


drodzewicz:
wasn't this the whole point that PR ##4586 was closed?
We didn't want to change the statuses on the view to not ruin anything else in the system.

awalkowiak:
@drodzewicz @alannadolny as we discussed this internally on Slack, this was the easiest and cleanest way to fix this issue with filtering. I believe this was even mentioned on the planning call as a required solution. I know we decided at the start of this status-related work to avoid changing SQL vies, but I feel like this is the right approach here.
To achieve it in another way we'd had to do either a backward display status mapping or some additional property which feels like too much of unnecessary workarounds. 
I am ok with this approach as is. Let us know @jmiranda what do you think about it.

awalkowiak (changes requested):
@kchelstowski please add comments in this view about what is a `status_code`, what is a `status`, what it is used for, and why we see still `DISPATCHED` for `status_code` for cases we have `ISSUED` for `status`, plus overall why we are changing this one here, where this change came from, etc. Be as descriptive as possible (you could also add a link to the ticket)

kchelstowski:
@awalkowiak I've added comments + I've deprecated the `statusCode` as it seems not to be used anymore (and it seems to have never been used for the list page :exploding_head: ). The reason I've deprecated it and not removed it yet, is because I reckon this should be done in the cleanup ticket (I've also mentioned that in the comment)

awalkowiak (approved):

PR #4628, opened 2024-05-17
OBPIH-6332 Dashboard indicator for number of backdated shipments

alannadolny:
I asked Manon about adding the second indicator within this PR (the backdated inbound shipments indicator), so I am pushing it here


grails-app/controllers/org/pih/warehouse/api/DashboardApiController.groovy
```diff
@@ -214,4 +215,16 @@ class DashboardApiController {
         NumberData numberData = numberDataService.getOpenPurchaseOrdersCount(params)
         render(numberData as JSON)
     }
+
+    def getBackdatedOutboundShipments() {
+        Integer monthsLimit = (params.querySize as Integer) ?: 6
```
EWaterman, line 220: Please put this 6 month default in a constant. Or even better as a config variable.

EWaterman: Also, do we need to define a server-side max? Can the server properly handle the request if we get sent something like `querySize=9999`?

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1111,6 +1112,74 @@ class IndicatorDataService {
...
 
+    Map getBackdatedShipmentsChart(String query, String locationId, Integer monthsLimit) {
+        Map<String, Integer> data = [
+                '2 days': 0,
+                '3 days': 0,
+                '4 days': 0,
+                '5 days': 0,
+                '6 days': 0,
+                '7 days': 0,
+                '7+ days': 0,
```
kchelstowski, line 1123: I would suggest moving those "2 days" etc. to static constants in `Constants.groovy`, as this might be reused in the future.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1111,6 +1112,74 @@ class IndicatorDataService {
...
+        IndicatorData indicatorData = new IndicatorData(datasets, data*.key.toList())
+        GraphData graphData = new GraphData(indicatorData)
+
+        return graphData.toJson()
+    }
+
+    @Cacheable(value = "dashboardCache", key = { "getBackdatedOutboundShipments-${locationId}${monthsLimit}" })
+    Map getBackdatedOutboundShipmentsData(String locationId, Integer monthsLimit) {
+        String query = """
+        SELECT (
```
kchelstowski, line 1148: I guess the indentations are missing here

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1111,6 +1112,74 @@ class IndicatorDataService {
...
+                WHEN DATEDIFF(t.date_created, t.transaction_date) > 7 THEN '7+ days'
+                ELSE CONCAT(DATEDIFF(t.date_created, t.transaction_date), ' days')
+            END
+        ) as days_backdated, COUNT(t.id) as shipments  FROM shipment s
+        INNER JOIN transaction t ON t.incoming_shipment_id  = s.id
+        WHERE s.destination_id = :locationId 
+        AND t.date_created > :timeLimit
+        GROUP BY days_backdated
+        HAVING days_backdated > :daysOffset
+        """
```
kchelstowski, line 1178: the only difference between those two queries if origin vs destination. Can we somehow pass it as a variable here so that we can reuse the query in both methods?

alannadolny: I am not sure if this is a good idea, I would rather say string concatenation in SQL is a thing that we should avoid.
As an example:
https://stackoverflow.com/questions/23179329/why-is-concatenating-sql-strings-a-bad-idea

If anyone else would like me to use the string concatenation here I can do it, but I am not a big fan, and that's the reason I didn't do it before.

kchelstowski: I know cases when concatenation is a wrong idea, and this is why they introduced an ability to use params via `:param` and pass it as the second argument of a query method, but in this case:
1. you wouldn't use any variable that the user provides and that might break the query (e.g. `drop database openboxes;` :laughing: ), but a variable that you would pass,
2. I was not really thinking about the concatenation, but rather passing it as param (if possible, probably not), or via `${variable}`.

I'm not sure if any of those options are possible, hence I just asked if "we can".

grails-app/conf/runtime.groovy
```diff
@@ -73,6 +73,10 @@ openboxes.client.notification.autohide.delay = 8000
 // Autosave configuration (OBPIH-5493)
 openboxes.client.autosave.enabled = false
 
+// Backdata configuration (OBPIH-6332)
+openboxes.dashboard.backdate.daysOffset = 1
```
awalkowiak, line 77: Let's use `backdatedShipments` in these two names to match the indicators these are used for.

awalkowiak (changes requested):

src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -156,4 +156,15 @@ class Constants {
     static final String ADJUSTMENT = "Adjustment"
 
     static final String UOM_EACH_ID = "EA"
+
+    // Dashboard indicator axes
+    static final Map<String, Integer> backdataAxes = [
```
awalkowiak, line 161: Do you mean `axis` or `axes`?

alannadolny: axes

drodzewicz (approved):

src/main/groovy/org/pih/warehouse/core/Constants.groovy
awalkowiak: nvm, you're right, plural is `axes`

awalkowiak (approved):

kchelstowski (approved):

PR #4629, opened 2024-05-20
OBPIH-6332 Dashboard indicator for number of backdated shipments (Fix overriding config value)

alannadolny:


grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1114,7 +1114,7 @@ class IndicatorDataService {
     }
 
     Map getBackdatedShipmentsChart(String locationId, Integer monthsLimit, StockMovementDirection direction) {
-        Map<String, Integer> data = Constants.backdataAxes
+        Map<String, Integer> data = Constants.backdataAxes.clone()
```
awalkowiak, line 1117: I missed that one before. I think `Constants` is not the best place for things that can be changed. I looked into the previous request and it was `"I would suggest moving those "2 days" etc. to static constants in Constants.groovy, as this might be reused in the future"` - I think it meant only the keys, not entire map (correct me if I am wrong here @kchelstowski). 
Either way, I think this map should not be kept as is in the consts.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
kchelstowski: @awalkowiak yes, I meant the keys. I'm sorry @alannadolny I haven't noticed it in the previous PR.

awalkowiak (approved):

awalkowiak (approved):

PR #4630, opened 2024-05-20
OBPIH-6428 autowire api clients in tests

EWaterman:
JIRA: https://pihemr.atlassian.net/browse/OBPIH-6428

Currently our API tests are initializing their own XApi classes. This is fine, but a better approach would be to annotate those Api classes with @TestComponent and then in the test, pull them in via @Autowired. This way we only have to initialize the classes once for the entire tests, and by using @TestComponent (as opposed to @Component) we can guarantee that they’re only wired in for our tests.


src/integration-test/groovy/org/pih/warehouse/component/api/generic/GenericApiWrapper.groovy
EWaterman, line 1: This file had a spelling error so I renamed it. That's why it looks like it's new

src/integration-test/groovy/org/pih/warehouse/component/product/CategoryApiCRUDSpec.groovy
EWaterman, line 1: And this is what the change allows us to do. It's simple, but it saves us from having to init new XApi classes in each of our tests.

src/integration-test/groovy/org/pih/warehouse/component/api/base/AuthenticatedApiContext.groovy
EWaterman, line 1: These are components containing the default request specifications. It does common stuff like set the content-type and the session cookie so that we don't need to specify it for every single request.

We auto-wire them then in ApiSpec call loadContext to actually populate them since it needs to be done after the test user is authenticated.

src/integration-test/groovy/org/pih/warehouse/component/base/ApiSpec.groovy
```diff
@@ -45,6 +49,7 @@ import org.pih.warehouse.util.common.ResponseSpecUtil
  * a clean, known state for each test.
  */
 @Integration
+@Import(ComponentTestConfig.class)
```
EWaterman, line 52: This is where we pull the new config in

jmiranda (approved):
LGTM

Issue #165, opened 2016-11-16
Replace all inventory transactions with a correspondng adjustment transaction


jmiranda:
Currently, OB needs to calculate all quantity on hand values in the code rather than delegating to the database because of a design decision we made back when we started. We wanted to be able to back-enter transactions (adjustments, transfers, stock counts, etc) for some inventory items without that transaction leading to an increase/decrease in the current QoH of that inventory item and product. the (and requiring a possible offsetting adjustment later on in the transaction history). So we needed a way to record the known quantities and "reset" quantities for inventory items that existed in the past (without requiring us to actually record a 0 for all of those items. We came up with the idea of a Product Inventory transaction that would allow us to specify the quantity for all known inventory items without also needing to zero out all inventory items that are no longer in stock. This is working fine, but there's also an Inventory transaction which just allows a user to change the current QoH for a given inventory item. Because of this, trying to calculate the QoH for an entire product (i.e. usually consisting of several inventory items or lot numbers) is very difficult to do with a single database query. Storing/caching the QoH values for all inventory items (and products) is a workaround that we've adopted temporarily (see #100), but I believe this improvement will help us get back to calculating the QoH values on the fly.

Anyway, replacing the inventory transaction (i.e. a data migration for data that exists and a code change for all future inventory transactions) might be the only issue that prevents us from calculating QoH from the transaction tables as it would allow us to aggregate debits/credits and subtract from a known starting point (e.g. quantity at most recent cycle count for the product OR zero if the product has never been cycle counted)

Related resources:
* http://allenbrowne.com/appinventory.html
* http://stackoverflow.com/questions/29711351/how-to-handle-change-in-inventory-quantity-on-a-backdate

PR #4633, opened 2024-05-24
OBPIH-6328 Dashboard indicator for items with backdated shipments

alannadolny:


grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1168,6 +1169,83 @@ class IndicatorDataService {
...
+                AND tt.transaction_code = :transactionCode
+                GROUP BY ii.product_id 
+            ) as sc ON sc.product_id = ii.product_id
+            GROUP BY ii.product_id, sc.last_stock_count
+        '''
+
+        List<GroovyRowResult> results = dataService.executeQuery(query, [
+                locationId: location?.id,
+                inventoryId: location?.inventory?.id,
+                transactionCode: TransactionCode.PRODUCT_INVENTORY.name(),
```
awalkowiak, line 1208: I think this also should/could include `TransactionCode.INVENTORY`. I think it was used in the past for marking the record stock done through import inventory. At least we are still looking into this type while calculating transactions for product availability and the latest inventory dates indicator.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1168,6 +1169,83 @@ class IndicatorDataService {
...
+        ]
+
+        String query = '''
+            SELECT ii.product_id, COUNT(ii.product_id) as products, GROUP_CONCAT(bt.shipment_number SEPARATOR ' ') as shipments, DATE_FORMAT(sc.last_stock_count, "%d-%M-%Y") FROM (
+                SELECT t.id as transaction_id, s.shipment_number FROM shipment s
+                INNER JOIN `transaction` t ON t.outgoing_shipment_id  = s.id
+                WHERE DATE_ADD(t.transaction_date, INTERVAL :daysOffset DAY) < t.date_created
+                AND t.date_created > :timeLimit
+                AND s.origin_id = :locationId
+            ) AS bt
```
awalkowiak, line 1190: Please be a bit more descriptive here with `bt` and `sc` since it does not match any table name and it's hard to decode

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1168,6 +1169,83 @@ class IndicatorDataService {
...
+                inventoryId: location?.inventory?.id,
+                transactionCode: TransactionCode.PRODUCT_INVENTORY.name(),
+                daysOffset: Holders.config.openboxes.dashboard.backdatedShipments.daysOffset,
+                timeLimit: timeLimit,
+        ])
+
+        List<TableData> tableData = results.collect { new TableData(
+                it[0] as String,          // Item
+                it[2] as String,          // Shipments
+                it[3] as String ?: "None" // Last count
```
awalkowiak, line 1216: I think "Never" would fit here more, unless "None" was requested, then you can disregard this comment

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
alannadolny: I am not sure about that, this subquery for getting the last stock count date is the same as in the product view (I copied this part of the query from there).
So, this method is called in the view:
image
This is the method that contains the query:
image

But, if you are sure about it, I can change that

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
alannadolny: it was requested

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
jmiranda: Sorry just seeing this now. I would prefer NOT to include Inventory. We have removed all of the features that created this type of transaction so if there's data in the database it's very old and somewhat irrelevant. In other words, if a transaction exists that has an "INVENTORY" transaction code I'd be fine for the last stock count feature to return Never.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
jmiranda: @awalkowiak I guess this is the sort of detail we need to catch during "design".

One additional point (add to #3849)

> [!IMPORTANT] 
> **Convention** Don't use "magic strings" 
> Source: https://deviq.com/antipatterns/magic-strings

1. We should use an enum if there are multiple coded classification values (i.e. Never, Long Ago, Somewhat Recently, Recently) 
1. We should use a Constant if we need to support just a string literal in logic (i.e. Default, Never, None). 
1. We should allow constants and enums to be i18n'd, so there's a way to translate the actual string literal. 

Add it right under the default constant and call it never. 
```
static final String DEFAULT = "DEFAULT"
static final String NEVER = "NEVER"
```

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
awalkowiak: @jmiranda do we have all these types of decisions documented somewhere? I based my comment (suggestion) on the past behavior, and state of the current code where we are still looking into this transaction type, and there is some dead code in the `InventoryService` that is using it and it was a bit misleading. Perhaps we could at least add some comments about it in the `TransactionCode` enum?

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
awalkowiak: @jmiranda 
> I guess this is the sort of detail we need to catch during "design".

This particular label did not appear in the ticket at all, and it was just suggested by Manon outside the ticket. So it is hard to speak about design here if it was missed since the beginning. I am also not sure if I would spot it even with a proper dev design because I would be focusing on the SQL query itself, it is hard to tell.

awalkowiak: (added a topic around label and messages designing for a tech huddle)

awalkowiak (approved):

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
jmiranda: FWIW, in case it wasn't clear ... this was not criticism of what we should have done here but rather what patterns we should be looking for when we start to do proper design in the near future.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
jmiranda: > and state of the current code where we are still looking into this transaction type

I think you're misread what I said. 

> We have removed all of the features that created this type of transaction

If that's not true and there's still code that **creates** INVENTORY transactions then please flag them and let me know.

> do we have all these types of decisions documented somewhere?

Yes, this particular issue was documented before you guys were involved.
https://github.com/openboxes/openboxes/issues/165

The underlying issue is the QoH Calculation issue that is near and dear to my heart. It was recently documented here but there are probably countless discussions on this issue throughout tickets, documents, and community posts.
https://pihemr.atlassian.net/browse/OBPIH-6353

The pull request for #165 is this epic roller coaster. 
https://github.com/openboxes/openboxes/pull/710

I was only able to get to a place where we could convert the INVENTORY transactions to ADJUSTMENT transactions on a product by product basis. The idea was to eventually move this to an automated feature that would convert everything, but I wasn't comfortable with that since there were some edge cases that I wasn't absolutely sure about. 

So I pushed that code and converted all of the transactions in the PIH and SBHF databases. But I have been holding out hope that I'd someday get become independently wealth so I could fund my own development priorities or have time to work on it as tech debt using PIH's budget. However, it's pretty far down on the tech debt priority list, so I think it'll probably be refactored out completely once we have time to work on the Cycle Count feature and the data model refactorings around that. 

> Perhaps we could at least add some comments about it in the TransactionCode enum?

Yes, that's probably a good idea.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1168,6 +1169,83 @@ class IndicatorDataService {
...
+        Date timeLimit = LocalDate.now().minusMonths(monthsLimit).toDate()
+        Map<String, Integer> amountOfItemsWithBackdatedShipments = [
+            (Constants.ONE)            : 0,
+            (Constants.TWO)            : 0,
+            (Constants.THREE)          : 0,
+            (Constants.FOUR_OR_MORE)   : 0,
+        ]
+
+        String query = '''
+            SELECT ii.product_id, COUNT(ii.product_id) as products, GROUP_CONCAT(backdated_transaction.shipment_number SEPARATOR ' ') as shipments, DATE_FORMAT(stock_count.last_stock_count, "%d-%M-%Y") FROM (
```
jmiranda, line 1184: This formatting is premature. We should return the date (or NULL) and then let the caller / client handle the rendering logic (i.e. formatting in a specific locale). And what I didn't realize was that we could just return NULL and allow the renderer to conditionally render the value properly. So probably don't even need the constant, definitely want to return a date and have the renderer handle the date format.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1168,6 +1169,83 @@ class IndicatorDataService {
...
+        results.each {
+            if (it[1] >= 4) {
+                amountOfItemsWithBackdatedShipments[Constants.FOUR_OR_MORE] += 1
+                return
+            }
+
+            amountOfItemsWithBackdatedShipments[it[1] as String] += 1
+        }
+
+        Table table = new Table("Item", "Shipments", "Last Count", tableData.toList())
```
jmiranda, line 1228: These need to be translatable strings. We do not need to handle that here, but we need to provide this as feedback in the ticket to make sure that Manon (at least) knows that these labels are hard-coded (along with the None label) and cannot be translated.

jmiranda: One way to tackle the translatable label problem is to require the ticket creator to specify multiple translations for each string at the beginning to make it clear that there's a i18n requirement.

PR #4640, opened 2024-05-27
OBPIH-6341 Fixes for export pick document

drodzewicz:
Include rows for requisition items that don't have picklist items and return quantity 0 for those lines

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
@@ -113,7 +113,7 @@ class PicklistController {
                 "${g.message(code: 'inventoryItem.lotNumber.label')}": it?.inventoryItem?.lotNumber ?: "",
                 "${g.message(code: 'inventoryItem.expirationDate.label')}": it?.inventoryItem?.expirationDate ? it.inventoryItem.expirationDate.format(Constants.EXPIRATION_DATE_FORMAT) : "",
                 "${g.message(code: 'inventoryItem.binLocation.label')}": it?.binLocation?.name ?: "",
-                "${g.message(code: 'default.quantity.label')}": it?.quantity ?: "",
+                "${g.message(code: 'picklist.quantity.label')}": it?.quantity ?: "",
             ]
         }
 
@@ -155,7 +155,7 @@ class PicklistController {
                     "${g.message(code: 'inventoryItem.lotNumber.label')}": "",
                     "${g.message(code: 'inventoryItem.expirationDate.label')}": "",
                     "${g.message(code: 'inventoryItem.binLocation.label')}": "",
-                    "${g.message(code: 'default.quantity.label')}": it?.requisitionItem?.quantity ?: "",
+                    "${g.message(code: 'picklist.quantityToPick.label')}": it?.requisitionItem?.quantity ?: "",
             ]
         }
 
--- a/grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
@@ -98,22 +98,39 @@ class PicklistController {
         String format = params.get("format", "csv")
 
         List<PickPageItem> pickPageItems = stockMovementService.getPickPageItems(params.id, null, null)
-        List<PicklistItem> picklistItems = pickPageItems.picklistItems?.flatten()
 
         // We need to create at least one row to ensure an empty template
-        if (picklistItems?.empty) {
-            picklistItems.add(new PicklistItem())
+        if (pickPageItems?.empty) {
+            pickPageItems.add(new PickPageItem())
         }
 
-        def lineItems = picklistItems.collect {
-            [
-                "${g.message(code: 'default.id.label')}": it?.requisitionItem?.id ?: "",
-                "${g.message(code: 'product.productCode.label')}": it?.requisitionItem?.product?.productCode ?: "",
-                "${g.message(code: 'product.name.label')}": it?.requisitionItem?.product?.name ?: "",
-                "${g.message(code: 'inventoryItem.lotNumber.label')}": it?.inventoryItem?.lotNumber ?: "",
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -94,28 +94,57 @@ class PicklistController {
...
+
+        List<Map> lineItems = pickPageItems.collectMany { pickPageItem ->
+            if (pickPageItem.picklistItems.size() > 0) {
+                return pickPageItem.picklistItems.collect { picklistItem ->
+                    return [
+                            id: picklistItem?.requisitionItem?.id ?: "",
+                            productCode: picklistItem?.requisitionItem?.product?.productCode ?: "",
+                            productName: picklistItem?.requisitionItem?.product?.name ?: "",
+                            lotNumber: picklistItem?.inventoryItem?.lotNumber ?: "",
+                            expirationDate: picklistItem?.inventoryItem?.expirationDate,
```
alannadolny, line 130: expirationDate doesn't need fallback as the other fields?

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -94,28 +94,54 @@ class PicklistController {
         render(template: "/picklist/print", model: [requisition: requisition, picklist: picklist, location: location, order: params.order])
     }
 
+    private Map formatPicklistItemPickExport(Map item) {
```
jmiranda, line 97: From what I gather, we only need to do the translation once when we generate the CSV column headers. This isn't an expensive operation now that we're using Crowdin, but it would be if we were going to the database every time. But I do think this would make more sense at the point of rendering (i.e. generate CSV, generate Excel).

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -94,28 +94,54 @@ class PicklistController {
...
+        return [
+                "${g.message(code: 'default.id.label')}": item?.id ?: "",
```
kchelstowski, line 99: let's add default messages

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -94,28 +94,54 @@ class PicklistController {
...
-                "${g.message(code: 'inventoryItem.binLocation.label')}": it?.binLocation?.name ?: "",
-                "${g.message(code: 'default.quantity.label')}": it?.quantity ?: "",
+
+        List<Map> lineItems = pickPageItems.collectMany { pickPageItem ->
+            if (pickPageItem.picklistItems.size() > 0) {
+                return pickPageItem.picklistItems.collect { picklistItem ->
+                    return [
+                            id: picklistItem?.requisitionItem?.id,
+                            productCode: picklistItem?.requisitionItem?.product?.productCode,
+                            productName: picklistItem?.requisitionItem?.product?.name,
```
kchelstowski, line 128: I'm wondering what about synonyms here, if we want to include them or not, so to return `displayName ?: name`?

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -94,28 +94,54 @@ class PicklistController {
...
-            [
-                "${g.message(code: 'default.id.label')}": it?.requisitionItem?.id ?: "",
-                "${g.message(code: 'product.productCode.label')}": it?.requisitionItem?.product?.productCode ?: "",
-                "${g.message(code: 'product.name.label')}": it?.requisitionItem?.product?.name ?: "",
-                "${g.message(code: 'inventoryItem.lotNumber.label')}": it?.inventoryItem?.lotNumber ?: "",
-                "${g.message(code: 'inventoryItem.expirationDate.label')}": it?.inventoryItem?.expirationDate ? it.inventoryItem.expirationDate.format(Constants.EXPIRATION_DATE_FORMAT) : "",
-                "${g.message(code: 'inventoryItem.binLocation.label')}": it?.binLocation?.name ?: "",
-                "${g.message(code: 'default.quantity.label')}": it?.quantity ?: "",
+
+        List<Map> lineItems = pickPageItems.collectMany { pickPageItem ->
```
kchelstowski, line 122: I'm trying to imagine the use case for the `.collectMany` here? 
From what I understand: https://blog.mrhaki.com/2011/09/groovy-goodness-transform-items-into.html
a basic `.collect` should also work for you?

drodzewicz: `collectMany `is for collecting nested lists.
If you look at the first part of the collect I return a list of collected picklists `pickPageItem.picklistItems.collect`.
So using just a `collect` will result in a following value eg. `[ [a,b,c], [n,c,g] ]` if I want to use a simple collect I would need to `flatten()` at the end

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
drodzewicz: sure, I was not thinking about it since I just copy pasted the old approach that did not have the default messages

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
drodzewicz: yes I understand, but the problem is that this was pretty much the old way of generating csv documents and I just moved this part into it's own function to avoid some dangerous code repetition.

The `dataService.generateCsv()` takes in the `List<Map>` and based on the first data entry generates the headings based on `keys`.

In my previous implementation for csv exporters I had considered generating headings in the `generateCSV ` function, but since we didn't want to proceed with a new the approach for exporters I reverted my changes and used the "old way" using `dataService`.

So summing up, I do see the problem for calling `g.message` for each row of `List<Map>` but I am not sure if we should spend time improving it right now or should we wait for a design (from idea garden ticket) and refactor it then.

drodzewicz: I have updated my appraoch and decided to generate headings before collecting the whole `List<Map>` and then use that translated string as a key, that way we call g.message once and us that value in the collect. 
_Check for latest updated files_

awalkowiak (approved):

PR #4642, opened 2024-05-28
OBPIH-6395 Add reusable useWizard hook

kchelstowski:
Peek 2024-05-28 15-44


src/js/tests/hooks/useWizard.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+    expect(result.current[0].key).toEqual(3);
+  });
+
+  it('wizard goes to previous step', () => {
+    const { result } = renderHook(() => useWizard({ initialKey: 4, steps }));
+
+    act(() => {
+      result.current[1].previous();
+    });
```
alannadolny, line 52: one nitpicky thing - the naming of tests (look into this comment) https://github.com/openboxes/openboxes/pull/3701#discussion_r1045926241

kchelstowski: Where is that convention documented? I can't any resource opting for the "should" convention.

alannadolny: do we have any conventions documented? :laughing: :laughing: :laughing: :laughing:

src/js/components/stock-movement-wizard/fullOutbound/FullOutboundImport.jsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+  ], []);
+
+  const stepsTitles = steps.map((step) => ({
+    title: step.title,
+    key: step.key,
+  }));
+
+  const [
+    Step,
```
alannadolny, line 39: Can you get `Component` and `key` by destructuring?
I would also like to make the object returned from the hook flat. It doesn't make sense to add an additional object.
``` 
return [
    Step,
      set,
      first,
      next,
      previous,
      last,
  ];
```

kchelstowski: imho it's cleaner to distinguish the step and the util hook functions, this is why I made it like this.

grails-app/conf/runtime.groovy
```diff
@@ -257,6 +257,7 @@ openboxes {
                     defaultLabel: "Stock Movement",
                     menuItems: [
                         [label: "outbound.create.label", defaultLabel: "Create Outbound Movements", href: "/stockMovement/createOutbound?direction=OUTBOUND"],
+                        [label: "outbound.fullOutbound.import.label", defaultLabel: "Import Completed Outbound", href: "/stockMovement/fullOutboundImport"],
```
awalkowiak, line 260: Perhaps it could be a simple: `outbound.import.label`?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -66,6 +66,10 @@ class StockMovementController {
         render(view: "/common/react", params: params)
     }
 
+    def fullOutboundImport() {
```
awalkowiak, line 69: Do we have to have a "full" keyword everywhere?

src/js/components/wizard/v2/WizardStepsV2.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import 'components/wizard/WizardSteps.scss';
+
+const WizardStepsV2 = ({ steps, currentStepKey }) => {
```
awalkowiak, line 7: What has changed here compared to `WizardSteps`?

src/js/components/stock-movement-wizard/fullOutbound/FullOutboundImport.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { useMemo } from 'react';
+
+import FullOutboundImportConfirm from 'components/stock-movement-wizard/fullOutbound/FullOutboundImportConfirm';
+import FullOutboundImportDetails from 'components/stock-movement-wizard/fullOutbound/FullOutboundImportDetails';
+import WizardStepsV2 from 'components/wizard/v2/WizardStepsV2';
+import useWizard from 'hooks/useWizard';
+
+const FullOutboundImport = () => {
+  const FullOutboundImportStep = {
```
awalkowiak, line 9: `FullOutboundImportSteps`

src/js/components/stock-movement-wizard/fullOutbound/FullOutboundImport.jsx
```diff
@@ -0,0 +1,56 @@
...
+      key: 'DETAILS',
+      // TODO: Make titles translatable in OBPIH-6329
+      title: 'Create',
+    },
+    CONFIRM: {
+      key: 'CONFIRM',
+      title: 'Confirm',
+    },
+  };
+  const steps = useMemo(() => [
```
awalkowiak, line 20: This feels a bit weird. Can we just have an object with full steps structure that would be transformed to list of values? Feels like this part could be simplified

src/js/hooks/useWizard.js
```diff
@@ -0,0 +1,73 @@
...
+    setKey(steps[0]?.key);
+  };
+
+  const last = () => {
+    const lastIdx = steps.length - 1;
+    setKey(steps[lastIdx]?.key);
+  };
+
+  const next = () => {
+    const nextStepIdx = steps.findIndex((s) => s.key === key) + 1;
```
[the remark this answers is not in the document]
awalkowiak: I also think I would prefer Alan's suggestion.

src/js/components/wizard/v2/WizardStepsV2.jsx
kchelstowski: I answered that question in one of the comments below: https://github.com/openboxes/openboxes/pull/4642/files#r1617300455

src/js/hooks/useWizard.js
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  const first = () => {
+    setKey(steps[0]?.key);
+  };
+
+  const last = () => {
+    const lastIdx = steps.length - 1;
+    setKey(steps[lastIdx]?.key);
+  };
```
drodzewicz, line 19: It's cool that you added these but are we ever going to use them?
I don't care if the stay or they go, I am just not familiar with any use case where we will want to jump to first and last steps.

src/js/tests/hooks/useWizard.test.jsx
drodzewicz: When Alan was implementing tests a while ago, this was a request from me, _"It should..."_ is a very common pattern for test name that explains the behavior and expected result like
"should return initial key"
"should advance to the next step when next is called"
"should go to the previous step when previous is called'"
"should go to the specific step when set is called"

drodzewicz (changes requested):

jmiranda (commented):
@kchelstowski Wait does this mean I can finally click on the wizard steps to go through the flow? I'm so confused. And excited.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: Yeah I was going to say we really should keep things simple. Try not to use PIH terminology if there's a way to make it simple. Imagine some old Greek guys having a conversation and then distill their response. 

> Socrates: What does it do? 
> Kacperian: It provides a way to import a stock movement including all of its dependent objects like requests, picklists, etc. 

:sparkles: 
```
importStockMovement()
```
:sparkles:

src/js/hooks/useWizard.js
kchelstowski: We already have such use cases - e.g. if we have an SM in progress and it is supposed to be on "X" step, we go to step "X".

src/js/components/wizard/v2/WizardStepsV2.jsx
```diff
@@ -0,0 +1,47 @@
...
+  const isCurrentStep = (iteratedStep) => iteratedStep.key === currentStepKey;
+  return (
+    <div className="steps-main-box">
+      <div className="steps-inside-wrapper">
+        {steps.map((step) => (
+          <div
+            key={step.key}
+            className={`step-container ${isCurrentStep(step) ? 'active' : ''}`}
+            data-testid={isCurrentStep(step) ? 'active' : 'inactive'}
```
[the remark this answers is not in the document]
kchelstowski: It was copied from the existing `WizardSteps`, so I did not even realize and thought about it, but since you are testing ninja, I trust you :laughing:

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
jmiranda: or 
```
importOutboundStockMovement()
```

src/js/hooks/useWizard.js
drodzewicz: yes, but I am not talking about function `set` but `last` and `first`

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: I would go with the 2nd option, especially if we are supposed to distinguish the SMs in the future for inbound, outbound etc.

src/js/hooks/useWizard.js
kchelstowski: maybe now we don't have the use case, but they felt cute and I also wanted to have a complete hook, that would cover any of the future cases, so I kept them :sadsmile: 
Let me know if I should remove them in your opinion guys.

drodzewicz (approved):

PR #4644, opened 2024-05-29
OBPIH-6387 Run MySQL in a testcontainer for integration tests

EWaterman:
The bulk of the leg work to get testcontainers set up was done by Justin, this is just to actually hook it up.

There's two methods of wiring up the test database to use testcontainers. You can do it directly via jdbc url, or via annotations. I chose via url because it's the smallest change, it automatically picks up the db config we have set in application.yml, and it immediately applies to all integration tests without issue.

More info here: https://java.testcontainers.org/modules/databases/jdbc/

If we wanted to do it the annotation way, it'd look something like this:

```
@Testcontainers
class Test extends Specification {

    @Shared // Marking it shared allows us to preserve the database across tests.
    MySQLContainer mySQLContainer = new MySQLContainer(DockerImageName.parse("mysql:5.7"))
            .withLogConsumer(new Slf4jLogConsumer(logger))
            .withDatabaseName('openboxes')
            .withUsername('openboxes')
            .withPassword('openboxes')

    void setupSpec() {
        mySQLContainer.start()
        assert mySQLContainer.isCreated()
        assert mySQLContainer.isRunning()
    }

    void cleanupSpec() {
        mySQLContainer.stop()
    }
}
```

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -269,9 +269,9 @@ environments:
             logAbandoned: true
             logSql: false
             logSlowQueries: false
-            username: "openboxes"
-            password: "openboxes"
-            url: jdbc:tc:mysql:///openboxes?serverTimezone=UTC&useSSL=false  # https://java.testcontainers.org/modules/databases/jdbc/
+            # https://java.testcontainers.org/modules/databases/jdbc/
+            # Note that testcontainers auto-disable SSL when connecting via jdbc https://github.com/testcontainers/testcontainers-java/pull/561
+            url: jdbc:tc:mysql:5.7.44:///openboxes?TC_MY_CNF=testcontainers
             driverClassName: org.testcontainers.jdbc.ContainerDatabaseDriver
     production:
         dataSource:
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -271,7 +271,8 @@ environments:
             logSlowQueries: false
             username: "openboxes"
             password: "openboxes"
-            url: jdbc:mysql://localhost:3306/openboxes?serverTimezone=UTC&useSSL=false
+            url: jdbc:tc:mysql:///openboxes?serverTimezone=UTC&useSSL=false  # https://java.testcontainers.org/modules/databases/jdbc/
+            driverClassName: org.testcontainers.jdbc.ContainerDatabaseDriver
     production:
         dataSource:
             dumpQueriesOnException: false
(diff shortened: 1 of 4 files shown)


EWaterman:
Once this is merged I'll update our documentation on test running to make it clear that by default this will run against a container and not your local db

grails-app/conf/application.yml
```diff
@@ -271,7 +271,8 @@ environments:
             logSlowQueries: false
             username: "openboxes"
             password: "openboxes"
-            url: jdbc:mysql://localhost:3306/openboxes?serverTimezone=UTC&useSSL=false
+            url: jdbc:tc:mysql:///openboxes?serverTimezone=UTC&useSSL=false  # https://java.testcontainers.org/modules/databases/jdbc/
```
jmiranda, line 274: - Move comemnt above 
- See if settings are actually getting applied here (try useSSL=true or some obvious setting)
- If the settings aren't getting passed through to the driver then we should set them in a mysql cnf override file as specified in docs (https://java.testcontainers.org/modules/databases/mysql/)
- Find out what is the default MySQL test container version and add that as either a comment or explicit 
- Write docs if we figure out how to actually get into the test container

jmiranda (approved):
LGTM

grails-app/conf/application.yml
jmiranda: * Test whether we can override configuration via env variables and system properties
   * export DATASOURCE_URL = 
   * grails -DdataSource.url= test-app
* If it doesn't work, then we need to do some passthrough shinanegans with gradle to make it work properly https://stackoverflow.com/questions/28196042/how-override-grails-configuration-from-command-line

jmiranda: With the config I want to 

1. make it work out of the box without any changes
2. use passthrough in build.gradle 
3. do something like this in application.yml
```
config:
    property: System.get("config.property", defaultValue)
```
I would prefer 1 or 2, and would like to stay away from 3 like the plague because it would require a code change to application.yml every time we add a new property.

EWaterman:
@jmiranda you were right the settings weren't getting applied. I added an override my.cnf file like you mentioned and that worked.

The default version is MySQL 57 so I changed it to use a specific version for clarity.

I think I'll be changing this shortly to use the annotation based config instead of just the jdbc connection url since that is far easier to make customizable for the scenario where we want to be able to swap around the db that we use, but I'll make a separate task for that one.

I'll be updating our testing docs to mention all this.

PR #4646, opened 2024-06-03
OBPIH-6396 Full Outbound Import - Reusable file select field (frontend)

alannadolny:


src/js/components/form-elements/v2/FileSelect.jsx
```diff
@@ -0,0 +1,152 @@
...
+    fieldProps.onChange?.(multiple ? acceptedFiles : acceptedFiles[0]);
+  }, []);
+
+  const translate = useTranslate();
+
+  const getFileExtension = (file) => file.path.split('.')?.[1];
+
+  const validateFileType = (file) => {
+    if (!allowedExtensions.length || allowedExtensions.includes(getFileExtension(file))) {
+      return null;
```
kchelstowski, line 32: I think for clarity of validation methods we used to have, this should eventually return `true` if the validation passes, which I think it does in this case if I understand it correctly:
"If we don't have any allowed extensions specified or the file extension is in the allowed ones, the validation should pass (currently returning null)"?

src/js/components/form-elements/v2/FileSelect.jsx
```diff
@@ -0,0 +1,152 @@
...
+    };
+  };
+
+  const {
+    getRootProps, getInputProps, open, acceptedFiles, fileRejections,
+  } = useDropzone({
+    onDrop,
+    noClick: true,
+    noKeyboard: true,
+    validator: validateFileType,
```
kchelstowski, line 51: but I guess this hook from the library expects to return a falsy value if the validation passes, hence my previous comment doesn't make sense, since we are not able to adjust that?

kchelstowski (commented):
My question would be if you checked the compatibility of the `FileSelect` you've implemented with both: react-hook-form and zod and if so, if there are any obstacles we need to know before trying to use one in the upcoming tickets?

src/js/components/form-elements/v2/FileSelect.jsx
```diff
@@ -0,0 +1,152 @@
...
+        {data.size}
+        {' '}
+        bytes)
+        {file?.errors?.length ? (
+          <ul>
+            {file.errors.map((e) => (
+              <li key={e.code}>{e.message}</li>
+            ))}
+          </ul>
+        ) : null}
```
drodzewicz, line 72: I would maybe do something like:
```suggestion
        <div>
          <span className="mx-1" >{data.path}<span>
          <span className="mx-1" >({data.size})<span>
        </div>
        {!!file?.errors?.length ? (
          <ul>
            {file.errors.map((e) => (
              <li key={e.code}>{e.message}</li>
            ))}
          </ul>
        ) }
```

But current solution is fine

drodzewicz (approved):

src/js/components/form-elements/v2/FileSelect.jsx
kchelstowski: @drodzewicz I think you skipped the fallback (`:`), or you meant to do it via `&&`. Honestly I'd avoid double negation, since it is usually confusing. 
I would prefer Alan's version.

drodzewicz: Sure, Idc

src/js/components/form-elements/v2/FileSelect.jsx
alannadolny: The `null` is required by the `react-dropzone`. Basically, this validation blocks you from uploading files with the wrong extensions.

src/js/components/form-elements/v2/FileSelect.jsx
alannadolny: This is not the validation for our libraries, I answered it in the comment above. I tested it with our libraries and it is working fine. In case someone has some problems with adjusting it, it can also be done by the workaround by passing the `setValue` function.

PR #4650, opened 2024-06-04
OBPIH-6365 ability to create edit a pick via import with incorrect stock cases

drodzewicz:


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1152,69 +1152,90 @@ class StockMovementService {
...
+                )
             }
-            pickPageItem.picklistItems.add(new PicklistItem(
-                    requisitionItem: requisitionItem,
-                    inventoryItem: availableItem.inventoryItem,
-                    binLocation: availableItem.binLocation,
-                    quantity: params.quantity,
-                    sortOrder: pickPageItem.sortOrder
-            ))
+            createMissingShipmentItem(pickPageItem.requisitionItem)
```
drodzewicz, line 1237: changes implemnted here meant to follo the way for updating picklist like on 
`StockMovementItemApiController > updatePicklist` where we update individual requisition item.

One problem that I have noticed is that previously importing picklist the old way would not update the shipmentItems where we need to additionally do this: `createMissingShipmentItem(requisitionitem)`

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1484,7 +1505,7 @@ class StockMovementService {
                 }
 
                 // Remove from picklist
-                if (picklistItem.quantity <= 0) {
+                if (picklistItem.quantity) {
```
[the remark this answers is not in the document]
drodzewicz: thansk for pointing that out, there was additionally condition that I didn't need anymore and I removed too much so it's a "typo type" mistake

awalkowiak (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,8 +1132,8 @@ class StockMovementService {
         }
     }
 
-    void validatePicklistListImport(List<PickPageItem> pickPageItems, Boolean supportsOverPick, List<ImportPickCommand> picklistItems) {
-        Map<String, List> groupedItems = picklistItems.groupBy { it.id }
+    void validatePicklistListImport(StockMovement stockMovement, List<PickPageItem> pickPageItems, List<ImportPickCommand> picklistItems) {
+        Map<String, List> picklistItemsGroupedByRequisitionItem = picklistItems.groupBy { it.id }
```
jmiranda, line 1136: Thank you! This is way more clear.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1121,9 +1121,9 @@ class StockMovementService {
                         code: it.code,
                         name: it.name,
                         lotNumber: it.lotNumber,
-                        expirationDate: new Date(it.expirationDate),
+                        expirationDate: it.expirationDate ? new Date(it.expirationDate) : null,
                         binLocation: it.binLocation,
-                        quantity: Integer.parseInt(it.quantity)
+                        quantity: it.quantity.isNumber() ? Integer.parseInt(it.quantity) : null
```
jmiranda, line 1126: Do we want to return a validation error if it's not a number? Or is it ok to handle the quantity as null? I think the distinction is that if null eventually becomes 0 and 0 is a valid quantity, then it might be confusing to the user who entered accidentially entered "100 ea" to have no picklist items for that line. 

I'm looking at this in a vacuum so I don't have a good sense of the expected behavior or next steps but I would look downstream to see what behavior we've implemented (understanding that the behavior might change in the future) and make sure that we handle it as explicitly as possible, rather than coercsing input to something the user may not have intended.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1152,69 +1152,90 @@ class StockMovementService {
             }
             if (pickPageItem) {
                 AvailableItem availableItem = pickPageItem.getAvailableItem(data.binLocation, data.lotNumber)
+                ApplicationTagLib g = grailsApplication.mainContext.getBean(ApplicationTagLib.class)
 
                 if (!availableItem) {
-                    String lotNumberErrorMessage = data.lotNumber ?: "empty"
-                    String binLocationErrorMessage = data.binLocation ?: "empty"
+
+                    String lotNumberName = data.lotNumber ?: g.message(code: "default.label", default: "Default")
```
jmiranda, line 1159: I used Constants.DEFAULT_LOT_NUMBER and Constants.DEFAULT_BIN_LOCATION, but this is much better. 

However, let's do this to be consistent.
```
String lotNumberName = data.lotNumber ?: g.message(code: "default.noLotNumber.label", default: Constants.DEFAULT_LOT_NUMBER)
```

jmiranda:
```
String binLocationName = data.binLocation ?: g.message(code: "default.noBinLocation.label", default: Constants.DEFAULT_BIN_LOCATION)
```

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: Oh shit, good catch. That's very important.

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -379,6 +379,14 @@ class AvailableItem {
 
 enum AvailableItemStatus {
     AVAILABLE, PICKED, RECALLED, HOLD, NOT_AVAILABLE
+
+    static getList() {
```
jmiranda, line 383: Follow the existing convention for listing enum values and rename the second method to be more readable.
```
static list() { ... }
static listUnavailable() { ... }
```

jmiranda (approved):
You don't need to incorporate the requested changes before merging, but I would like to explore the question about validating quantity as a discussion in the ticket. I will make the other requested changes in my OBPIH-6331 PR later today.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: We probably need to document this somewhere or centralize it i.e. any time we muck with the picklist it automatically updates the shipment. 

> One problem that I have noticed is that previous implementation of importing picklist would not update the shipmentItems where we need to additionally do this: createMissingShipmentItem(requisitionitem)

Additionally, are you saying this is a current bug in production? If you're not sure we should definitely add this as a discussion point to the ticket with the advise that we probably want to investigate this as an OBS data issue i.e. comparing picklist items to shipment items to ensure they include the same quantity. If not, this is going to be a pain to clean up.

drodzewicz: Most likely yes, since my previous implementation for the pick import was based on the old import I would assume that the same issue occurred before.

PR #4651, opened 2024-06-05
OBPIH-6365 Add validation on quantity field

drodzewicz:
This PR was created to fix some points mentioned in #4650 

As you can see it's not so straightforward to add a validation of type (not a number) on a quantity, let me explain.

So the case is that a user uploads a file with a potential "invalid" quantity number.
We want our API to process this data and return to the user all of the errors that come up in the File.

Now if we set **quantity** as a type _Integer_, we lose the ability to even get to the custom validators because it will throw an exception something along the lines of `Can't cast String to an Integer` in constructor when assigning `this.quantity = quantity`, which is why I changed the type of the quantity so we can assign the quantity and then our validator can validate the value.

Additionally, I have added a `getQuantityAsNumber` so we don't have to cast String to Integer every time we want to use this value, not sure if you guys will support this solution.

At the moment this is the best I could come up with so if you guys have any ideas then let me know

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
+++ b/grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
@@ -11,7 +11,9 @@ class ImportPickCommand implements Validateable {
     String lotNumber
     Date expirationDate
     String binLocation
-    String quantity
+    String quantityAsText
+
+    static transients = ['quantity']
 
     static constraints = {
         id(nullable: false, blank: false)
@@ -25,27 +27,18 @@ class ImportPickCommand implements Validateable {
         })
         expirationDate(nullable: true)
         binLocation(nullable: true)
-        quantity(nullable: false, validator: { val, obj ->
+        quantityAsText(nullable: false, validator: { val, obj ->
             if (!NumberUtils.isNumber(val)) {
-                return ['notANumber.error']
+                return ['invalid.error']
             }
             if (Integer.parseInt(val) < 0) {
                 return ['negative.error']
             }
             return true
         })
-        quantityAsNumber(nullable: true)
     }
 
-    void setQuantity(String quantity) {
-        this.quantity = quantity
-    }
-
-    void setQuantity(Integer quantity) {
-        this.quantity = String.valueOf(quantity)
-    }
-
-    Integer getQuantityAsNumber() {
-        return NumberUtils.isNumber(this.quantity) ? Integer.parseInt(this.quantity) : null
+    Integer getQuantity() {
+        return NumberUtils.isNumber(this.quantityAsText) ? Integer.parseInt(this.quantityAsText) : null
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
+++ b/grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
@@ -11,7 +11,7 @@ class ImportPickCommand implements Validateable {
     String lotNumber
     Date expirationDate
     String binLocation
-    Integer quantity
+    String quantity
 
     static constraints = {
         id(nullable: false, blank: false)
@@ -25,6 +25,27 @@ class ImportPickCommand implements Validateable {
         })
         expirationDate(nullable: true)
         binLocation(nullable: true)
-        quantity(nullable: false, min: 0)
+        quantity(nullable: false, validator: { val, obj ->
+            if (!NumberUtils.isNumber(val)) {
+                return ['notANumber.error']
+            }
+            if (Integer.parseInt(val) < 0) {
+                return ['negative.error']
+            }
+            return true
+        })
+        quantityAsNumber(nullable: true)
+    }
+
+    void setQuantity(String quantity) {
+        this.quantity = quantity
+    }
+
+    void setQuantity(Integer quantity) {
+        this.quantity = String.valueOf(quantity)
+    }
+
+    Integer getQuantityAsNumber() {
+        return NumberUtils.isNumber(this.quantity) ? Integer.parseInt(this.quantity) : null
     }
 }
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -1884,6 +1884,8 @@ picklistItem.quantity.label=Quantity
 
 #Picklist import
 importPickCommand.lot.scientificNotation.error=Lot numbers must not be specified in scientific notation. Please reformat field with lot number to a number format
+importPickCommand.quantity.notANumber.error=Invalid quantity: "{2}". Please enter a valid number.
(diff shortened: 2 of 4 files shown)


grails-app/i18n/messages.properties
```diff
@@ -1884,6 +1884,8 @@ picklistItem.quantity.label=Quantity
 
 #Picklist import
 importPickCommand.lot.scientificNotation.error=Lot numbers must not be specified in scientific notation. Please reformat field with lot number to a number format
+importPickCommand.quantity.notANumber.error=Invalid quantity: "{2}". Please enter a valid number.
+importPickCommand.quantity.negative.error=Invalid quantity: "{2}". Quantity must not be a negative number.
 importPickCommand.requisitionItem.notFound.error=Requisition item with id: {0} not found
 importPickCommand.availableItem.notFound.error=The stock entry you selected: lot "{0}" and bin location "{1}" does not exist. Please review pick.
 importPickCommand.availableItem.notAvailable.error=The stock entry you selected: lot "{0}" and bin location "{1}" is not available. Please review pick.
```
jmiranda, line 1891: In order to keep things consistent, let's change this to "unavailable" as well.

jmiranda (commented):
Let me know what you think. Also, if you've moved on I can work on the request changes as part of OBPIH-6331.

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
```diff
@@ -25,6 +25,27 @@ class ImportPickCommand implements Validateable {
         })
         expirationDate(nullable: true)
         binLocation(nullable: true)
-        quantity(nullable: false, min: 0)
+        quantity(nullable: false, validator: { val, obj ->
```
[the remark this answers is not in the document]
jmiranda: I see what you mean now regarding this comment.

> Now if we set quantity as a type Integer, we lose the ability to even get to the custom validators because it will throw an exception something along the lines of Can't cast String to an Integer in constructor when assigning this.quantity = quantity, which is why I changed the type of the quantity so we can assign the quantity and then our validator can validate the value.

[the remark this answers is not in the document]
jmiranda: So i think what you have is the right approach for now. 

We could/should explore if there's a way to do some of the parsing / validation logic in the CSVMapReader to avoid these kinds of workarounds. Or use a different library. 

But that's out of scope for this ticket.

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
```diff
@@ -25,6 +25,27 @@ class ImportPickCommand implements Validateable {
...
+            if (!NumberUtils.isNumber(val)) {
+                return ['notANumber.error']
+            }
+            if (Integer.parseInt(val) < 0) {
+                return ['negative.error']
+            }
+            return true
+        })
+        quantityAsNumber(nullable: true)
```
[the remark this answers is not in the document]
jmiranda: More to the point, is this necessary to resolve some other issue? This feels like it might be related to the fact that the getQuantityAsNumber() getter should probably have a property defined in the transients list.

jmiranda:
> Additionally, I have added a getQuantityAsNumber so we don't have to cast String to Integer every time we want to use this value, not sure if you guys will support this solution.
> 
> At the moment this is the best I could come up with so if you guys have any ideas then let me know

I pushed a commit that reverses those quantity mappings, so that the ugly property name is assigned the raw string field (String quantityAsText) rather than the field we'd like to expose (Integer quantity).

Otherwise, the PR looks good.

jmiranda:
If/when this gets merged, please use a merge commit and NOT "squash and merge". I recreated my branch for OBPIH-6331 and I'm basing it off this branch.

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
```diff
@@ -11,7 +11,9 @@ class ImportPickCommand implements Validateable {
     String lotNumber
     Date expirationDate
     String binLocation
-    Integer quantity
+    String quantityAsText
+
+    static transients = ['quantity']
```
kchelstowski, line 16: fyi: in Grails 3 we don't need to declare a transient anymore. The getter is enough.

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -380,11 +380,11 @@ class AvailableItem {
 enum AvailableItemStatus {
     AVAILABLE, PICKED, RECALLED, HOLD, NOT_AVAILABLE
 
-    static getList() {
+    static list() {
```
kchelstowski, line 383: have you made sure it is not used anywhere with the `getList()` version?

kchelstowski (commented):
as far I as I understand the context, I'm fine with your solution, but for me, getting an exception "cannot cast ...." would be self-explanatory that I've done something wrong with the number.

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
drodzewicz: Yep I am sure, It was me who introduced this variable in the last PR

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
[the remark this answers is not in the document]
drodzewicz: it was necessary because `getQuantityAsNumber()` returned a `null` and without it I got an error that `getQuantityAsNumber` cant be a nul value.
And as Kacper mentioned earlier, we don't have to define a transient for this getter

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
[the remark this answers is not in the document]
drodzewicz: This definitely sounds like something we should discuss in the design of importers ticket (OBID idea garden ticket)

grails-app/controllers/org/pih/warehouse/picklist/ImportPickCommand.groovy
jmiranda: I think I prefer transients over setters but we can discuss.

kchelstowski: We don't need a setter neither. Since Grails 2.0 we can declare a getter without having to introduce a transient field:
https://docs.grails.org/6.1.2/ref/Domain%20Classes/transients.html

> As of Grails 2.0 if there is only a getter or only a setter method, you don’t need to declare the property name of the method in the transients list. Only typed fields and get/set pairs that form a property but shouldn’t be persisted need to go in the transients list.

PR #4654, opened 2024-06-06
OBPIH-6420 Undo edit functionality in cleared pick mode

alannadolny:


src/js/api/urls.js
```diff
@@ -146,3 +146,4 @@ export const PRODUCT_SUPPLIER_ATTRIBUTE_BATCH = `${PRODUCT_SUPPLIER_ATTRIBUTE_AP
 
 export const PICKLIST_API = `${API}/picklists`;
 export const PICKLIST_CLEAR = id => `${PICKLIST_API}/${id}/items`;
+export const PICKLIST_REVERT_PICK = (id, itemId) => `${PICKLIST_CLEAR(id)}/${itemId}`;
```
jmiranda, line 149: We should be more obvious here to avoid confusion / regression later. If there's a baseUrl shared by multiple endpoints, let's define the base URL. Also
```
export const PICKLIST_API = `${API}/picklists`;
export const PICKLIST_ITEM_API = `${API}/picklists/${id}/items`;
export const PICKLIST_ITEM_CLEAR = id => `${PICKLIST_ITEMS_API}`;
export const PICKLIST_ITEM_REVERT = (id, itemId) => `${PICKLIST_ITEMS_API(id)}/${itemId}`;
```

jmiranda (commented):
@awalkowiak has final say on what needs to be changed here.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
...
     }
+
+    void revertPick(String id, String itemId) {
+        Picklist picklist = Picklist.get(id)
+        RequisitionItem requisitionItem = RequisitionItem.get(itemId)
+        List<PicklistItem> picklistItemsToRemove = PicklistItem.findAllByRequisitionItem(requisitionItem)
+
+        if (!picklist) {
+            throw new ObjectNotFoundException(id, Picklist.class.toString())
+        }
```
kchelstowski, line 335: I would call this just after attempting to get a picklist, to avoid some unnecessary queries (e.g. the one from 331th), if they are not gonna be used either way.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
             order("lastUpdated", "desc")
         }
     }
+
+    void revertPick(String id, String itemId) {
+        Picklist picklist = Picklist.get(id)
+        RequisitionItem requisitionItem = RequisitionItem.get(itemId)
```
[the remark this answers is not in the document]
awalkowiak: @jmiranda Yup, I agree that the current solution with `DELETE /api/picklists/$id/items/$itemId` and passing requisition item id as `itemId` is not perfect and might be misleading or cause some confusion. When I was looking at the code I was thinking about moving the current `revertPick` (with some minor adjustments if needed) under `DELETE /api/requisitionItems/:id/picklistItems` or `DELETE /api/stockMovementItems/:id/picklistItems` (the first one was my first thought, but considering it is stock movement related, then I think the latter one is more suiting)

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
...
+        List<PicklistItem> picklistItemsToRemove = PicklistItem.findAllByRequisitionItem(requisitionItem)
```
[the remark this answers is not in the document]
awalkowiak: Should be just `requisitionItem.picklistItems`

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
...
+            throw new ObjectNotFoundException(id, Picklist.class.toString())
+        }
+
+        if (!picklistItemsToRemove?.size()) {
+            throw new ObjectNotFoundException(itemId, PicklistItem.class.toString())
+        }
+
+        List<String> binLocations = []
+        picklistItemsToRemove.each {
+            picklist.removeFromPicklistItems(it)
```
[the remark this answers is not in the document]
awalkowiak: Yup, it has to be checked and disabled if needed.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
...
+            binLocations.add(it.binLocation?.id)
+            it.delete(flush: true)
+        }
+
+        productAvailabilityService.refreshProductsAvailability(
+                picklist?.requisition?.origin?.id,
+                picklistItemsToRemove?.requisitionItem?.product?.id?.unique(),
```
awalkowiak, line 350: If you have `requisitionItem`, then having `[requisitionItem?.product?.id]` is enough, because there is no way there will be different products for picklist items for the same requisition item.

grails-app/services/org/pih/warehouse/picklist/PicklistService.groovy
```diff
@@ -322,4 +324,32 @@ class PicklistService {
...
+                binLocations,
+                false
```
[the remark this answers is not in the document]
awalkowiak: @jmiranda Force refresh deletes the line from the table first, I think we don't have to do it here and just doing the line update is enough.

src/js/api/urls.js
awalkowiak: :+1: this could be improved

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -339,6 +339,11 @@ class UrlMappings {
             action = [DELETE: "clearPicklist"]
         }
 
+        "/api/picklists/$id/items/$itemId" {
```
awalkowiak, line 342: This endpoint needs to be improved as it was discussed in the comment in service and on slack

awalkowiak (changes requested):

awalkowiak (changes requested):

awalkowiak (approved):

PR #4656, opened 2024-06-07
OBPIH-6329 Add details step of full outbound import form

kchelstowski:
Screenshot from 2024-06-07 13-42-31

### DIFF
--- a/src/js/components/form-elements/v2/DateField.jsx
+++ b/src/js/components/form-elements/v2/DateField.jsx
@@ -6,13 +6,13 @@ import DatePicker from 'react-datepicker';
 
 import DateFieldInput from 'components/form-elements/v2/DateFieldInput';
 import { DateFormat, TimeFormat } from 'consts/timeFormat';
+import useTranslate from 'hooks/useTranslate';
 import InputWrapper from 'wrappers/InputWrapper';
 import RootPortalWrapper from 'wrappers/RootPortalWrapper';
 
 import 'react-datepicker/dist/react-datepicker.css';
 import 'components/form-elements/DateFilter/DateFilter.scss';
 import './style.scss';
-import useTranslate from 'hooks/useTranslate';
 
 const DateField = ({
   title,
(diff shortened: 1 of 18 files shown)


src/js/components/form-elements/v2/DateField.jsx
```diff
@@ -25,17 +25,27 @@ const DateField = ({
...
 
   const formatDate = (dateToFormat) => {
     if (!dateToFormat) {
       return null;
     }
+    if (showTimeSelect) {
+      return moment(new Date(dateToFormat), DateFormat.MMM_DD_YYYY_HH_MM_SS);
+    }
 
     return moment(new Date(dateToFormat), DateFormat.MMM_DD_YYYY);
```
alannadolny, line 50: It can be reduced using ternary operator

src/js/components/stock-movement-wizard/outboundImport/OutboundImportTitle.jsx
alannadolny, line 1: I think I would move this component to `OuboundImportHeader.jsx`, basically this one is short and has no logic, so I don't see any benefits of separating those.

src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -0,0 +1,53 @@
...
+    origin: undefined,
+    destination: undefined,
+    requestedBy: undefined,
+    dateRequested: undefined,
+    dateShipped: moment(new Date()).format(DateFormat.MMM_DD_YYYY_HH_MM_SS),
+    shipmentType: undefined,
+    trackingNumber: undefined,
+    comments: undefined,
+    expectedDeliveryDate: undefined,
+    packingList: undefined,
```
alannadolny, line 21: Are those `undefined` necessary to use these values as defaults?

src/js/utils/zodUtils.js
```diff
@@ -0,0 +1,14 @@
...
+ * @return true if the date is not from the future; false if the date is from the future
+ * @param pickedDate
+ */
+const validateFutureDate = (pickedDate) => {
+  const date = moment(pickedDate);
+  const today = moment(new Date());
+  return date.startOf('day').isSameOrBefore(today.startOf('day'));
+};
+
+export default validateFutureDate;
```
alannadolny, line 14: I think I would move it to `form-utils.js` instead of creating the next "utils" file

src/js/components/form-elements/v2/DateField.jsx
kchelstowski: I would say it doesn't reduce much, and it's just a matter of preference, as it's a simple one if statement. 
Anyway, changed to ternary.

src/js/components/stock-movement-wizard/outboundImport/OutboundImportTitle.jsx
kchelstowski: Removed

src/js/hooks/outboundImport/useOutboundImportForm.js
kchelstowski: For now yeah, the eventual adjustments (if we will even support the "edit" functionality for this feature), should be dealt with in the "integrate backend & frontend" tucket.

drodzewicz (approved):

alannadolny (approved):

PR #4658, opened 2024-06-10
OBPIH-6331 Inferring bin locations in pick import

jmiranda:


jmiranda:
Although it's not done, y'all can start to review this one.

grails-app/controllers/org/pih/warehouse/allocation/AllocationRequest.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    static constraints = {
+        product nullable:false
+        inventoryItem nullable:true
+        requisitionItem nullable:true
+        availableItems nullable:false
+        suggestedItems nullable:true
+        quantityRequired nullable:false
+        picklistItemCommand nullable:false
+    }
```
alannadolny, line 28: Can you add whitespaces after the `:`? And we can also add brackets.

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -218,22 +219,46 @@ class PicklistController {
...
         }
 
-        String csv = new String(importFile.bytes)
-        List<ImportPickCommand> importedLines = stockMovementService.parsePickCsvTemplateImport(csv)
+//        MultipartFile importFile = command.importFile
+//        if (importFile.isEmpty()) {
+//            throw new IllegalArgumentException("File cannot be empty")
+//        }
+
+//        String csv = new String(importFile.bytes)
```
alannadolny, line 245: I assume this will be deleted later because you are still working on that

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -285,6 +285,25 @@ class Location implements Comparable<Location>, java.io.Serializable {
         Location.list().findAll { it.isDepotWardOrPharmacy() }.sort { it.name }
     }
 
+    Location getDefaultLocation() {
+        return null
+    }
```
alannadolny, line 290: Is there something missing? I can't see any comment here

grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -453,6 +454,42 @@ class Product implements Comparable, Serializable {
...
+                if (lotNumber) {
+                    eq("lotNumber", lotNumber)
+                } else {
+                    or {
+                        isNull("lotNumber")
+                        eq("lotNumber", "")
+                    }
+                }
+            }
+        } as InventoryItem
```
alannadolny, line 489: It can be only my own preference, but I prefer to use the `return` keyword

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -323,6 +324,10 @@ class AvailableItem {
         quantityAvailable(nullable: true)
     }
 
+    String getBinLocationName() {
+        binLocation?.name?:Constants.DEFAULT_BIN_LOCATION_NAME
```
alannadolny, line 328:
```suggestion
        binLocation?.name ?: Constants.DEFAULT_BIN_LOCATION_NAME
```

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
jmiranda: Yes, but thanks for the reminder. I just wanted to make sure I tested an invalid file before removing it.

grails-app/domain/org/pih/warehouse/core/Location.groovy
jmiranda: Yeah, this is awkward at best. I think this was just a quick hack to get it working and I was expecting to get back to it. I am trying to provide a representation of a default location. But the default location is just a Location that is null. 

I would love to do this in a better way and open to ideas. A javadoc comment is necessary. And maybe this needs to be a static method, returning a constant? 

```
static Location getDefaultLocation() { 
    return Constants.DEFAULT_LOCATION 
}
```
or something like this? 
```
final static Location DEFAULT_LOCATION = null;
static Location getDefaultLocation() { 
    return Location.DEFAULT_LOCATION; 
}
```

grails-app/domain/org/pih/warehouse/product/Product.groovy
jmiranda: Yeah, same.  It was there but i was getting a weird IDE error so while I was trying to fix that I think I accidentally removed the "return"
image

grails-app/domain/org/pih/warehouse/core/Location.groovy
alannadolny: The first static method returning a constant looks great

grails-app/domain/org/pih/warehouse/product/Product.groovy
alannadolny: Probably adding `as InventoryItem` behind the closure should solve the issue, but I am not sure.

jmiranda: Yes it does. The `as InventoryItem` doesn't feel right either though. But I was just saying that i tried a few things (including removing the return) so that's why it was in that state. Should be resolved now.

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+        // FIXME Remove this if it's no longer being used
+        // Configuration property that allows overpick for a given location
+        Boolean supportsOverpick = command.location.supports(ActivityCode.ALLOW_OVERPICK)
 
-                if (availableItem && AvailableItemStatus.listUnavailable().contains(availableItem.status)) {
-                    String lotNumberName = data.lotNumber ?: g.message(code: "default.noLotNumber.label", default: Constants.DEFAULT_LOT_NUMBER)
-                    String binLocationName = data.binLocation ?: g.message(code: "default.noBinLocation.label", default: Constants.DEFAULT_BIN_LOCATION_NAME)
+        // TODO Do we handle the case where ID is null?
+        // Group the picklist items from the CSV by requisition item ID
+        Map<String, List> picklistItemsGroupedByRequisitionItem = command.picklistItems.groupBy { it.id }
```
awalkowiak, line 1161: Is the `id` on the `PicklistItemCommand` a requisition item's `id`?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
 
+        // Iterate over the parent requisition items
+        picklistItemsGroupedByRequisitionItem.each { requisitionItemId, picklistItems ->
+
+            // Iterate over each separate pick
+            picklistItems.each { PicklistItemCommand data ->
```
awalkowiak, line 1167: Does `PicklistItemCommand` represent already picked item, item to be picked or both cases?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+                Integer quantityRequired = data.quantity
+
+                // FIXME Add proper comment to explain what this is for
+                // Base validation, creates errors object?
+                data.validate()
+
+                // TODO What happens if ID is null. We should throw an exception if we don't have a good answer.
+                //  skip validation if id is empty since most of the validation relies on an existing requisition id
+                // skip validation if id is empty since mos tof the validation relies on an existing requisition id
+                if (!data.id) {
```
awalkowiak, line 1180: looks like it cannot be empty because id on `PicklistItemCommand` has constraints: `nullable: false, blank: false`

grails-app/controllers/org/pih/warehouse/picklist/PicklistImportDataCommand.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    StockMovement stockMovement
+    List<PickPageItem> pickPageItems
+    List<PicklistItemCommand> picklistItems
+
+    PicklistImportDataCommand() {
+        // FIXME this probably shouldn't be necessary, but i was getting
+        //  a validation error on location (possibly when the command object
+        //  was initially bound).
+        location = AuthService.currentLocation
```
awalkowiak, line 18: I think it should be a StockMovement's origin location instead current location. What do you think?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+                                    [data.lotNumber] as Object[],
+                                    "Bin location cannot be empty for multiple rows with the same lot number {0}"
+                            )
+                        }
+                    }
+
+                    // OBPIH-6331 Resolve bin location, if a bin location value was provided. Otherwise,
+                    // if the user-provided bin location is null or ambiguous then we need to apply
+                    // allocation rules
+                    Location internalLocation = command.location.getInternalLocation(data.binLocation)
```
awalkowiak, line 1229: Is it technically possible (now or in the future, through API) that someone will be able to import an outbound not from the "current location"? This is why I am wondering if we should not start using origin instead of current location as a `command.location` :thinking:

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+
+        Product product = allocationRequest.product
+        InventoryItem inventoryItem = allocationRequest.inventoryItem
+        Integer quantityRequired = allocationRequest.quantityRequired
+        List<AvailableItem> availableItems = allocationRequest.availableItems
+        PicklistItemCommand picklistItemCommand = allocationRequest.picklistItemCommand
+
+        boolean isInDefaultLocation = availableItems.every { it.isDefaultLocation }
+        boolean isInReceivingLocations = availableItems.every { it.isReceivingLocation }
+        boolean isInBinLocations = availableItems.every { it.isBinLocation && !it.isDefaultLocation }
```
awalkowiak, line 1370: Are these three variables used anywhere

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+        log.info "Available items " + availableItems.size()
+        availableItems.each {
+            log.info "Available items " + it.toJson()
+        }
+        Integer countInVirtualAndBinLocations = availableItems?.findAll { it?.isVirtualLocation || it?.isBinLocation }?.size()
+        log.info "123countInVirtualAndBinLocations " + countInVirtualAndBinLocations
+        Integer countInBinLocations2 = availableItems?.findAll { it?.isBinLocation }?.size()
+        log.info "countInBinLocations " + countInBinLocations2
+        Integer countInVirtualLocations = availableItems?.findAll { it?.isVirtualLocation }?.size()
+        log.info "countInVirtualLocations " + countInVirtualLocations
```
awalkowiak, line 1416: Are these required too? Or the scenario above is still to be finished?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
+            // TODO If there's enough quantity available in default location, then
+            //  allocate from default location.
+
+            // Scenario 3c: if lot has stock entries only in hold bins
+            // TODO If lot has stock entries only in hold bin, then throw an exception.
+
+            // Scenario 4: Any other scenario (stock in “real” bin and virtual bin, stock in multiple real bins)
+            // TODO If there's not enough available quantity in default + receiving
+            //  locations, then throw an exception
+            boolean isAllStockInReceivingOrDefault = availableItems.every { it.isReceivingLocation || it.isDefaultLocation }
```
awalkowiak, line 1498: Is there an order of precedence for this scenario which is picked first? Plus for this scenario shouldn't we also look at `stock in “real” bin` looks like you only look at receiving or default

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
-                AvailableItem availableItem = pickPageItem.getAvailableItem(params.binLocation, params.lotNumber)
-                Picklist picklist = stockMovement.requisition?.picklist
-                // find existing picklist to update
-                PicklistItem picklistItem = picklist.picklistItems.find {
-                    it.inventoryItem?.id == availableItem.inventoryItem?.id &&
-                            it.binLocation?.id == availableItem.binLocation?.id
+            // TODO Need to test this thoroughly as it does not feel like a good idea. For example,
+            //  we don't necessarily want to clear picklist if there's no picklist items to import
+            //  so this might need to go inside a check on !picklistItemsToImport.empty
+            clearPicklist(pickPageItem.requisitionItem)
```
awalkowiak, line 1531: Yeah I think this might cause unnecessary product availability refresh if there are no existing picklist items at this stage

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
```diff
@@ -218,22 +219,42 @@ class PicklistController {
...
 
-        MultipartFile importFile = command.importFile
-        if (importFile.isEmpty()) {
-            throw new IllegalArgumentException("File cannot be empty")
-        }
+        // Bind stock movement based on provided ID
+        command.stockMovement = stockMovementService.getStockMovement(params.id)
+
+        // Bind user-provided location or current location
+        command.location = command.location ?: Location.get(session.warehouse.id)
```
drodzewicz, line 228: (Just thinking out loud)

I wonder if in case of importing, we will ever need current logged in location :thinking: 
For example, case with supported activity **Allow Overpick** I belive we should check this activity not on current logged in location but on origin because the origin location should decide if it supports this activity.

This was just one example but I guess any future validation will involve either the **origin** or **destination** locations, the current logged in location will probably be a single case that will not allow importing items if `current location != origin`.

Do I have a point or am I going off the track?

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1132,106 +1140,438 @@ class StockMovementService {
...
-                            [lotNumberName, binLocationName] as Object[],
-                            "There is no item available with: lot \"${lotNumberName}\" and bin location \"${binLocationName}\""
-                    )
-                }
+        // The parent requisition items that picklist items are associated with
+        List<PickPageItem> pickPageItems = command.pickPageItems
+
+        // FIXME Remove this if it's no longer being used
+        // Configuration property that allows overpick for a given location
+        Boolean supportsOverpick = command.location.supports(ActivityCode.ALLOW_OVERPICK)
```
drodzewicz, line 1157: like I mentioned in the comment before, this should probably be checked for origin location and not current logged-in.

grails-app/controllers/org/pih/warehouse/picklist/PicklistController.groovy
awalkowiak: Yup, I agree with you. I also think we should base it on the origin in this case. I brought this up in the PicklistImportDataCommand too.

awalkowiak:
Closing this one because this is stale branch, and the feature itself was already merged

PR #4660, opened 2024-06-11
OBPIH-6372 full outbound import confirm page frontend

drodzewicz:
This PR should be merged after #4656 because I based my approach on Kacpers details step


drodzewicz:
There are some commits from crowding OBPIH-6328 and OBPIH-6365 most likely because #4656 is not rebased with develop.

alannadolny (approved):

src/js/components/stock-movement-wizard/outboundImport/OutboundImportTitle.jsx
alannadolny, line 1: I think this component was removed in Kacper's PR

alannadolny (changes requested):
:point_up_2:

awalkowiak (approved):

src/js/components/stock-movement-wizard/outboundImport/subsections/OutboundImportShipmentDetails.jsx
```diff
@@ -0,0 +1,153 @@
...
+                disabled
+                title={{
+                  id: 'react.stockMovement.shipDate.label',
+                  defaultMessage: 'Ship date',
+                }}
+                placeholder={{
+                  id: 'react.default.dateInput.placeholder.label',
+                  default: 'Select a date',
+                }}
+                errorMessage={errors.dateShipped?.message}
```
kchelstowski, line 76: I think you missed `showTimeSelect` prop.

src/js/components/stock-movement-wizard/outboundImport/subsections/OutboundImportShipmentDetails.jsx
```diff
@@ -0,0 +1,153 @@
...
+                  defaultMessage: 'Tracking Number',
+                }}
+                errorMessage={errors.trackingNumber?.message}
+              />
+            )}
+          />
+        </div>
+        <div className="col-lg-3 px-2 pt-2 d-lg-block d-md-none" />
+        <div className="col-lg-3 px-2 pt-2 d-lg-block d-md-none" />
+        <div className="col-lg-3 col-md-6 px-2 pt-2">
```
kchelstowski, line 119: what are those two empty divs responsible for?

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
```diff
@@ -73,9 +77,19 @@ const OutboundImport = () => {
     <PageWrapper>
       <WizardStepsV2 steps={stepsTitles} currentStepKey={Step.key} />
       <OutboundImportHeader />
-      <form onSubmit={handleSubmit(onSubmit)}>
+      <form onSubmit={handleSubmit(onSubmitStockMovementDetails)}>
         {is(OutboundImportStep.DETAILS.key) && (<Step.Component {...detailsComponentProps} />)}
-        {is(OutboundImportStep.CONFIRM.key) && (<Step.Component previous={previous} />)}
+      </form>
+      <form onSubmit={handleSubmit(onConfirmImport)}>
```
kchelstowski, line 83: nice idea with having two separate forms, triggering the same `handleSubmit` - I assume we will be able easily to manipulate whether we trigger the validation endpoint or the save (onSubmitStockMovementDetails would trigger the validation endpoint and the onConfirmImport would trigger the save)

src/js/components/stock-movement-wizard/outboundImport/subsections/OutboundImportShipmentDetails.jsx
drodzewicz: Let me share a screenshot for how the confirm page looks like first
image
So there are supposed to be a space between two input, and these two empty divs just act as a placeholders empty spaces for the grid.

To answer your potential question "Why did you not use **space-between**?" it's because when using space-between it will not have the same effect and will result in these two fields be of different size than the ones above since they are going to be pushed to the smallest element size which is different to the auto-fit size of the grid elements.

kchelstowski: sounds good

PR #4662, opened 2024-06-11
OBPIH-6486: hibernate errors running tests against different contexts

EWaterman:
https://pihemr.atlassian.net/browse/OBPIH-6486

This change doesn't actually fix the issue, but it does make tests pass while we figure out the cause. There's something strange going on with how hibernate is loading our domains in the app context during tests that is preventing us from having some tests use some context config and other tests use another config.

I also took the opportunity to restructure our packages. Now it works like:
image
with API, slice, and smoke tests being separated into their own packages which we can run independently if we want.


src/integration-test/groovy/org/pih/warehouse/common/base/IntegrationSpecConfig.groovy
EWaterman, line 1: This is the only real change. The comment in the todo and in the ticket should hopefully explain the situation.

Eventually I want to break this class up into ApiSpecConfig, SliceSpecConfig, SmokeSpecConfig... which can each do their own component scan for whatever test components they need, but I spent hours trying to get that to work and just couldn't find out what was causing the hibernate errors.

EWaterman:
Got the :green_circle: from Justin offline. Merging.

PR #4663, opened 2024-06-12
OBPIH-6329 Full outbound import - Details step (Fixes after QA)

kchelstowski:
- origin is filled by default with the current location
- date shipped validation (can't be date from the past)
- expected delivery date validation (can't be before the date shipped)

### DIFF
--- a/src/js/utils/form-utils.jsx
+++ b/src/js/utils/form-utils.jsx
@@ -158,12 +158,12 @@ export const validateFutureDate = (pickedDate) => {
 };
 
 /**
- * Method to check if first date is later than the second date
- * @return true if the first date is later than the second date
+ * Method to check if first date is later or same as the second date
+ * @return true if the first date is later or same as the second date
  * @param laterDate
  * @param earlierDate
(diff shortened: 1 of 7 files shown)


src/js/utils/form-utils.jsx
```diff
@@ -156,3 +156,15 @@ export const validateFutureDate = (pickedDate) => {
...
   return date.startOf('day').isSameOrBefore(today.startOf('day'));
 };
+
+/**
+ * Method to check if first date is later than the second date
+ * @return true if the first date is later than the second date
+ * @param laterDate
+ * @param earlierDate
+ */
+export const validateOneAfterAnotherDate = (laterDate, earlierDate) => {
```
awalkowiak, line 166: `validateOneAfterAnotherDate` - "Method to check if first date is later than the second date",  but you check `isSameOrAfter`.

awalkowiak (changes requested):

awalkowiak (approved):

PR #4670, opened 2024-06-13
OBPIH-6378 Ability to edit pick via pick import

alannadolny:


grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -1209,6 +1217,8 @@ class StockMovementService {
                 return
             }
 
+            picklistService.revertPick(requisitionItemId)
```
awalkowiak, line 1220: Two  questions:
1. Should this always be triggered? Feels like if this `importPicklistItems` method is used for create picklist items, then this would trigger an unnecessary refresh on product availability even if there is nothing being removed (reverted).
2. Side note/question - Is there a difference between clearPicklist from this service and revertPick from picklistService? I missed that when there was a PR for undo edited picklist item feature.

I terms of change requests here. I think the first point **should** be improved, and the second **can potentially** be improved. If it is **exactly** the same, then we can think about potentially removing one or at least writing some comments explaining which is used where and why.

awalkowiak (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
jmiranda: > Side note/question - Is there a difference between clearPicklist from this service and revertPick from picklistService? I missed that when there was a PR for undo edited picklist item feature.
>
> clearPicklist -> Remove all picklist items from the picklist 

I don't think this is true. clearPicklist has a variation that clears picklist items for a single requisition item. Which seems to be the same as the revertPick method. 

So I agree with Artur on the path forward. We should take a look at the two methods (revertPick vs clearPicklist) to see if they are identical and then consolidate / refactor from there.

Also, my branch already dealt with this issue using clearPicklist, so I wish you guys had reviewed my changes before doing this. 
```
// TODO Does the order of operations matter (remove shipment items vs picklist items)
// Remove existing shipment items and picklist items
removeShipmentItemsForModifiedRequisitionItem(pickPageItem.requisitionItem)

// TODO Need to test this thoroughly as it does not feel like a good idea. For example,
//  we don't necessarily want to clear picklist if there's no picklist items to import
//  so this might need to go inside a check on !picklistItemsToImport.empty
clearPicklist(pickPageItem.requisitionItem)
```
> [!important]
> Lesson learned. We probably should have detected that these tickets were going to be adding to the same code and created a feature branch for all of these picklist import tickets. I am dealing with a bunch of rebase issues that I fear that this second rebase is going to cause some regressions. I would appreciate @alannadolny, @awalkowiak, and @drodzewicz to review the recent import picklist merges and try to help me massage them into the Inferring Bin Location branch (OBPIH-6331-part-deux).

PR #4676, opened 2024-06-16
Allow testcontainers to be used in development #4764

jmiranda:


ewaterman:
@jmiranda is this still something you're planning on experimenting with?

PR #4681, opened 2024-06-18
OBPIH-6410 Allow configuring country specific localization

awalkowiak:
"Some" context https://pihemr.atlassian.net/browse/OBPIH-6410?focusedCommentId=153567

### DIFF
--- a/grails-app/i18n/messages_es_MX.properties
+++ b/grails-app/i18n/messages_es_MX.properties
@@ -1,0 +1,1 @@
+default.number.format=\# \# \#, \# \# \#, \# \# \#0.00
--- a/grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
@@ -52,7 +52,7 @@ class LocalizationService {
      * @return
      */
     Locale getLocale(String languageCode) {
-        return languageCode ? new Locale(languageCode) : currentLocale
+        return languageCode ? LocalizationUtil.getLocale(languageCode) : currentLocale
     }
 
     /**
@@ -69,51 +69,11 @@ class LocalizationService {
      */
     Properties getMessagesProperties(Locale locale) {
         Properties messagesProperties = new Properties()
-        def messagesPropertiesFilename = (locale && locale.language != "en" && locale.language != 'null') ? "messages_${locale.language}.properties" : "messages.properties"
+        def messagesPropertiesFilename = (locale && locale.language != "en" && locale.language != 'null') ? "messages_${locale.toString()}.properties" : "messages.properties"
 
         def resource = grailsResourceLocator.findResourceForURI('classpath:' + messagesPropertiesFilename)
         messagesProperties.load(resource.inputStream)
 
         return messagesProperties
-    }
-
-    Properties getMessagesPropertiesWithPrefix(String prefix, Locale locale) {
-        return getMessagesProperties(locale).findAll {
-            it.key.startsWith(prefix)
-        }
-    }
-
-    /**
-     * Get messages properties while running app using deployed WAR.
-     *
-     * @param messagesPropertiesUrl
-     * @return
-     */
-    Properties getMessagesPropertiesFromResource(String messagesPropertiesUrl) {
-        Properties properties = new Properties()
-        def inputStream = ServletContextHolder.servletContext.getResourceAsStream(messagesPropertiesUrl)
-        if (inputStream) {
-            String messagesPropertiesString = IOUtils.toString(inputStream)
-            properties.load(new StringReader(messagesPropertiesString))
-            inputStream.close()
-        }
-        return properties
-    }
-
-    /**
-     * Get messages properties while running app using grails run-app.
-     *
-     * @param messagesPropertiesUrl
-     * @return
-     */
-    Properties getMessagesPropertiesFromClasspath(String messagesPropertiesUrl) {
-        Properties properties = new Properties()
-        File messagesPropertiesFile = new ClassPathResource(messagesPropertiesUrl)?.getFile()
-        if (messagesPropertiesFile.exists()) {
-            InputStream messagesPropertiesStream = new FileInputStream(messagesPropertiesFile)
-            properties.load(messagesPropertiesStream)
-            messagesPropertiesStream.close()
-        }
-        return properties
     }
 }
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -1449,7 +1449,7 @@ class ProductService {
 
     Product addSynonymToProduct(String productId, String synonymTypeCodeName, String synonymValue, String localeName) {
         Product product = Product.get(productId)
-        Locale locale = localeName ? new Locale(localeName) : null
+        Locale locale = localeName ? LocalizationUtil.getLocale(localeName) : null
         SynonymTypeCode synonymTypeCode = synonymTypeCodeName ? SynonymTypeCode.valueOf(synonymTypeCodeName) : SynonymTypeCode.ALTERNATE_NAME
         Synonym synonym = new Synonym(name: synonymValue, locale: locale, synonymTypeCode: synonymTypeCode)
         product.addToSynonyms(synonym)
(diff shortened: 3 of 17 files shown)


jmiranda:
@awalkowiak Please give me a few days to get to this one.

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -26,9 +26,41 @@ class LocalizationUtil {
         return localizationService.currentLocale
     }
 
+    /**
```
EWaterman, line 29: An alternative to this would be to extend the Locale class and give it another constructor but I don't know if there's much advantage of that over what you've done

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1449,7 +1449,7 @@ class ProductService {
 
     Product addSynonymToProduct(String productId, String synonymTypeCodeName, String synonymValue, String localeName) {
         Product product = Product.get(productId)
-        Locale locale = localeName ? new Locale(localeName) : null
+        Locale locale = localeName ? LocalizationUtil.getLocale(localeName) : null
```
EWaterman, line 1452: Not a question to be resolved here, but I'm wondering if there's a fancier way to do this so that we don't need to bake all this localization logic into our services.

It'd be awesome if we could mark a Domain field as "localized" and then whenever we do a look up on that field it'd automatically also search for the synonyms.

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -26,9 +26,41 @@ class LocalizationUtil {
...
+            locale.setDefault(new Locale(language))
+            return locale
+        }
+
+        locale = new Locale(localeCode)
+        locale.setDefault(new Locale(Holders.config.openboxes.locale.defaultLocale ?: "en"))
+        return locale
+    }
+
     static List<Locale> getSupportedLocales() {
```
EWaterman, line 61: Since locales are static, we could consider caching them in a map (maybe in the LocalizationService) so that we don't need to reconstruct them every time we do a lookup. Currently, every time we call this method we're re-initializing all of the Locale objects. That's not a big deal since it's not really a lot of computation work, but still a good practice IMO.

If we wanted to get really fancy, we could even build that map on app startup based on the locales we have defined in our properties.

grails-app/i18n/messages_es_MX.properties
```diff
@@ -0,0 +1 @@
+default.number.format=\# \# \#, \# \# \#, \# \# \#0.00
```
EWaterman, line 1: is this being used yet?

grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
```diff
@@ -69,51 +69,11 @@ class LocalizationService {
      */
     Properties getMessagesProperties(Locale locale) {
         Properties messagesProperties = new Properties()
-        def messagesPropertiesFilename = (locale && locale.language != "en" && locale.language != 'null') ? "messages_${locale.language}.properties" : "messages.properties"
+        def messagesPropertiesFilename = (locale && locale.language != "en" && locale.language != 'null') ? "messages_${locale.toString()}.properties" : "messages.properties"
```
EWaterman, line 72: in what situations do we want `locale.toString()` vs locale.getDisplayName()`? I see we use both

grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
```diff
@@ -69,51 +69,11 @@ class LocalizationService {
...
 
         def resource = grailsResourceLocator.findResourceForURI('classpath:' + messagesPropertiesFilename)
         messagesProperties.load(resource.inputStream)
 
         return messagesProperties
     }
-
```
EWaterman, line 79: unused code?

awalkowiak: yep

grails-app/services/org/pih/warehouse/core/LocalizationService.groovy
awalkowiak: We want to find a message properties file corresponding to the current locale. `toString()` concatenates language code with country code together with an underscore (or returns just language code if no country code). So for Spanish, we are getting just `es` -> `messages_es.properties` and for Mexican Spanish, we will get `es_MX` -> `messages_es_MX.properties`

grails-app/i18n/messages_es_MX.properties
awalkowiak: This was an "initial" request from the ticket, so I just added it here to not leave it empty

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -26,9 +26,41 @@ class LocalizationUtil {
...
+     * @param localeCode String with locale code represented as single language code (ex.: "en", "es"),
+     *                   or combined with country code (ex.: "en_GB", "es_MX").
+     * @return Locale
+     * @exception NullPointerException (Locale throws NPE when all params passed to constructor are null)
+     * */
+    static Locale getLocale(String localeCode) {
+        Locale locale
+
+        if (localeCode?.contains("_")) {
+            String[] localeCodeParts = localeCode.split("_")
```
kchelstowski, line 48: nitpicky: I'd add the `"_"` separator to the Constants.

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -26,9 +26,41 @@ class LocalizationUtil {
...
+            String language = localeCodeParts[0]
+            String country = localeCodeParts[1]
+            locale = new Locale(language, country)
+            locale.setDefault(new Locale(language))
+            return locale
+        }
+
+        locale = new Locale(localeCode)
+        locale.setDefault(new Locale(Holders.config.openboxes.locale.defaultLocale ?: "en"))
```
kchelstowski, line 57: I would move the `new Locale(...) ?: "en"` to a separate variable for a better readability.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
awalkowiak: Yeah, I guess this is a more Synonyms related topic suiting for a TD discovery or some OBIG

grails-app/views/admin/showSettings.gsp
```diff
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.core.RoleType" %>
+<%@ page import="org.pih.warehouse.LocalizationUtil; org.pih.warehouse.core.RoleType" %>
```
kchelstowski, line 1: nitpicky: I know it was done automaticallty, but we used to separate the imports in gsp templates to be one import per line

grails-app/views/user/create.gsp
```diff
@@ -1,5 +1,5 @@
 
-<%@ page import="org.pih.warehouse.core.User" %>
+<%@ page import="org.pih.warehouse.LocalizationUtil; org.pih.warehouse.core.User" %>
```
kchelstowski, line 2: same as above

kchelstowski (commented):
+ I would make sure none adjustments have to be done in the `InitializationInterceptor`, but looking quickly at the code, I think none have to be done.

grails-app/views/admin/showSettings.gsp
awalkowiak: Yeah this is my laziness since I was doing the automatic import and noticed it while creating the PR it's in one line, I'll fix that

awalkowiak:
@kchelstowski 
> I would make sure none adjustments have to be done in the InitializationInterceptor, but looking quickly at the code, I think none have to be done.

Initially, I was making the changes there too, because I was pulling through this new util every "defaultLocale" as well, but I decided to skip that one for three reasons:
- I just assumed `localizationModeLocale` is crowdin-specific and it will always be `ach` for us
- I just assumed no one would have country-specific as the default for now (at least for now)
- I want to improve the `supportedLocales` in some further (ideally) post-release ticket/changes

kchelstowski:
@awalkowiak I think even if the users will have country-specific language, this should be handled already by the:
```groovy
session?.locale ?: session.user?.locale
```
assuming the session locale would be set correctly (which you've probably done)
so I think the only edge case I saw there is the last fallback:
```groovy
new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
```
but as long as no one has their defaultLocale set to `es`, we should be fine.

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
alannadolny, line 57: I think it would be better to use "en" from the `Locale` class instead of hardcoding it.

kchelstowski (approved):

alannadolny (approved):

awalkowiak:
@jmiranda I am merging this one, but feel free to leave a post-merge review

PR #4682, opened 2024-06-20
OBPIH-6520 Set up for date localization in DD/Month/YYYY format for Es and Mx locales

alannadolny:
Usage:
1. G tag
```
<g:formatDate date="${new Date()}" formatName="${org.pih.warehouse.DateFormatName.CUSTOM.id}"/>
<g:formatDate date="${new Date()}" formatName="${org.pih.warehouse.DateFormatName.DEFAULT.id}"/>
<g:formatDate date="${new Date()}" formatName="${org.pih.warehouse.DateFormatName.EXPIRY.id}"/>

```
2. Grails util
```
LocalizationUtil.formatDate(new Date(), DateFormatName.EXPIRY)
LocalizationUtil.formatDate(new Date(), DateFormatName.CUSTOM)
LocalizationUtil.formatDate(new Date(), DateFormatName.DEFAULT)
```
3. React component
```
<FormatDate date={new Date()} formatName={DateFormatName.DEFAULT} />
<FormatDate date={new Date()} formatName={DateFormatName.EXPIRY} />
<FormatDate date={new Date()} formatName={DateFormatName.CUSTOM} />
```

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -357,9 +357,9 @@ default.credits.label=Credits
 default.balance.label=Balance
 default.dashboard.label=Dashboard
 default.data.label=Data
-default.date.format=dd/MMM/yyyy hh:mm:ss a z
-custom.date.format=MM/dd/yyyy
-expiry.date.format=MM/yyyy
+datetime.format=dd/MMM/yyyy hh:mm:ss a z
+date.format=MM/dd/yyyy
+monthyear.format=MM/yyyy
 default.date.label=Date
 default.dateCreated.label=Date Created
 default.dateFrom.label=From
@@ -3236,9 +3236,9 @@ react.default.dateInput.placeholder.label=Select a date
 react.default.form.field.required.label=${0} is required
 
 # Date formats
-react.default.defaultDate.format=MM/DD/yyyy hh:mm:ss a z
-react.default.customDate.format=MM/DD/yyyy
-react.default.expiryDate.format=MM/yyyy
+react.default.datetime.format=MM/DD/yyyy hh:mm:ss a z
+react.default.date.format=MM/DD/yyyy
+react.default.monthyear.format=MM/yyyy
 
 # Notifications
 react.notification.connectivity.offline.label=Lost Connection
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -358,6 +358,8 @@ default.balance.label=Balance
 default.dashboard.label=Dashboard
 default.data.label=Data
 default.date.format=dd/MMM/yyyy hh:mm:ss a z
+custom.date.format=MM/dd/yyyy
+expiry.date.format=MM/yyyy
 default.date.label=Date
 default.dateCreated.label=Date Created
 default.dateFrom.label=From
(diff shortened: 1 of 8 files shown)


kchelstowski (approved):

grails-app/i18n/messages.properties
```diff
@@ -3232,6 +3234,12 @@ react.default.select.typeToSearch.label=Type to search
 react.default.noOptions.label=No options
 react.default.dateInput.placeholder.label=Select a date
 react.default.form.field.required.label=${0} is required
+
+# Date formats
+react.default.defaultDate.format=MM/DD/yyyy hh:mm:ss a z
+react.default.customDate.format=MM/DD/yyyy
+react.default.expiryDate.format=MM/yyyy
```
EWaterman, line 3241: we're only changing the format to dd/MMMMM/yyyy for ex_MX? Do we not want all locals to have the new format?

awalkowiak: I think that others should have a default right now.

grails-app/utils/org/pih/warehouse/DateFormatName.groovy
```diff
@@ -0,0 +1,21 @@
...
+// 1. Add new format in message.properties file
+// 2. Add the newly created label to this enum
+// This enum is used in:
+// 1. G tag - <g:formatDate date={your date} formatName={id of enum property} />
+// 2. LocalizationUtil.formatDate(your date, id of enum property)
+
+enum DateFormatName {
+    DEFAULT('default.date.format'),
+    CUSTOM('custom.date.format'),
+    EXPIRY('expiry.date.format'),
```
EWaterman, line 14: I'm not sure if this is the types we want. DEFAULT and CUSTOM are more generic sounding while EXPIRY is for a very specific purpose (product expiration). It's mixing and matching a bit.

What are you imagining is the behaviour if we add a new feature X that also needs to display only month and year? Would it re-use the EXPIRY type or would it need to create its own X type that also formats MMM/yyyy?

My instinct tells me that we'd be better off with more directly descriptive types (ex: DATE_TIME, FULL_DATE. MONTH_YEAR) but let me know your thoughts

grails-app/utils/org/pih/warehouse/DateFormatName.groovy
```diff
@@ -0,0 +1,21 @@
...
+
+    final String id
```
EWaterman, line 16: I think something like "property" is more accurate of a field name

grails-app/utils/org/pih/warehouse/DateFormatName.groovy
EWaterman: and if you want to avoid having DEFAULT be in the enum itself you could move it to a constant in the enum:

```
public final DateFormatName DEFAULT = DATE_TIME
```

grails-app/utils/org/pih/warehouse/DateFormatName.groovy
```diff
@@ -0,0 +1,21 @@
...
+// 1. Add new format in message.properties file
+// 2. Add the newly created label to this enum
+// This enum is used in:
+// 1. G tag - <g:formatDate date={your date} formatName={enum property} />
+// 2. LocalizationUtil.formatDate(your date, enum property)
+
+enum DateFormatName {
+    DATE_TIME('default.date.format'),
+    FULL_DATE('custom.date.format'),
+    MONTH_YEAR('expiry.date.format'),
```
EWaterman, line 14: thanks for making these changes but now the actual properties themselves don't match.

I was thinking we'd have something like:
```
datetime.format=DD/MMM/yyyy hh:mm:ss a z
date.format=DD/MMM/yyyy
monthyear.format=MMM/yyyy
```
the monthyear one is hard to decide on what a good wording would be but maybe it's clear enough like that.

PR #4395, opened 2023-11-29
OBGM-508 Improve product availability refresh performance

kchelstowski:

PR #4687, opened 2024-06-24
[OBPIH-6479, OBPIH-6483] PR labeler, PR template description

EWaterman:
This is a low-priority change so feel free to take your time reviewing it.

Tickets: https://pihemr.atlassian.net/browse/OBPIH-6483 https://pihemr.atlassian.net/browse/OBPIH-6479

Background info: https://pihemr.atlassian.net/wiki/spaces/OB/pages/3066658817/DISCOVERY+OBPIH-6376+Standardize+GitHub+branching+practices#Pull-Request-Rules

Added a new GitHub action for automatically labelling a pull request based on the branch name and what files are changed. You can see some of the labels being automatically applied in this PR.

The goal is to be able to determine the type of change made by a PR just by looking at its labels. This will hopefully make reviewing easier in that we can filter reviews by things like frontend/backend and also will make searching for reviews, discussions, and issues easier. We can add more labels in the future if we want.

I also added a pull request template `pull_request_template.md` that will auto fill for every new PR. That way we can ensure all PRs are clearly describing their changes and are following our code guidelines


.github/release.yml
```diff
@@ -3,21 +3,32 @@
 changelog:
```
EWaterman, line 3: This is for when we generate our release notes. It'll split the commits into sections based on the labels that were on their PRs.

It was already existing from a previous PR so I just fixed it to support the updated labels.

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
+### :sparkles: Description of Change
```
EWaterman, line 1: This template will be auto-generated for every new PR. Feel free to suggest other things to add but maybe we can review it as a team in a tech huddle. I just wanted to have something for now

drodzewicz:
Look good to me! :rocket: 
One more thing that we should probably include is some kind of labeling for the PR of which release is this PR addressing.
There were many times in the past where I was going through git-logs and reviewing old PR's to see the reasoning for the change and wondered which release did this address, so maybe some **text in the description** or a **tag** would be a good addition
@EWaterman

EWaterman:
@drodzewicz Yeah that would be useful. I'd definitely like for it to be automatic since I don't want people to have to remember what version we're on and to add the label. Maybe we could create a GitHub Action that pulls the `version` setting from `gradle.properties` https://github.com/openboxes/openboxes/blob/develop/gradle.properties#L8C1-L8C8 then adds a label using that version but I don't know how easy that is.

I do know that GitHub has milestones, which is essentially like releases. I see we've used them before https://github.com/openboxes/openboxes/pull/4395. Maybe something we could consider again.

jmiranda:
Yeah that's a great idea @drodzewicz and something I hadn't thought of for PRs. @EWaterman milestones is definitely the way to go there. I used them a bit with issues but never really considered it for PRs since we can trace back to the release using the ticket. It would be great to use milestones on PRs but one thing that might be surprisingly difficult is keeping the milestone for a PR in sync with the Jira fixVersion.

EWaterman:
I'd like to handle release/milestone labelling in a separate PR. I imagine that will require some thought since we're not really utilizing milestones right now. I do like the idea of associating PRs to releases, it'll just require some investigation.

Also for reference, this is what the current PR template looks like:

Screenshot from 2024-07-11 21-02-03

PR #4690, opened 2024-06-25
OBPIH-6447 Make the shipped by fallback to last updated, not created …

kchelstowski:
…by for old requests

### DIFF
--- a/grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
+++ b/grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
@@ -415,6 +415,7 @@ class Shipment implements Comparable, Serializable {
     User getShippedBy() {
         Event shippedEvent = events?.find { it.eventType?.eventCode == EventCode.SHIPPED }
         if (shippedEvent) {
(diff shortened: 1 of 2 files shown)


kchelstowski:
Added the "do not merge" label to let Katarzyna collect a few requests that she'll be able to compare the changes on, which she wouldn't be able to, if the PR got merged first.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -415,6 +415,9 @@ class Shipment implements Comparable, Serializable {
     User getShippedBy() {
         Event shippedEvent = events?.find { it.eventType?.eventCode == EventCode.SHIPPED }
         if (shippedEvent) {
+            if (requisition?.electronicType) {
```
EWaterman, line 418: can you add a comment here explaining why we want to do this just for e-requisitions? From the description of the ticket, this is just a more accurate default value in that scenario

EWaterman (approved):

awalkowiak (approved):

PR #4691, opened 2024-06-25
OBPIH-6461 Scroll issue on add items page after import template 

alannadolny:


src/main/groovy/org/pih/warehouse/api/PaginationCommand.groovy
```diff
@@ -22,6 +22,10 @@ class PaginationCommand implements Validateable {
         return Math.max(offset ?: 0, 0)
     }
 
+    Integer getOriginalMax() {
```
EWaterman, line 25: maybe "getUncappedMax" "getAbsoluteMax" is more clear.

grails-app/controllers/org/pih/warehouse/core/StockMovementItemsParamsCommand.groovy
```diff
@@ -11,6 +11,13 @@ class StockMovementItemsParamsCommand extends PaginationCommand implements Valid
 
     Boolean refreshPicklistItems = true
 
+    @Override
+    Integer getMax() {
+        // Overriding getter for max value from pagination command to be able
+        // to get all of the items, instead of just 10 (10 was a default value in pagination command)
+        return super.originalMax
```
EWaterman, line 18: it looks like 10 is the default value for the original getMax if "max" is undefined. Is the issue that the client is not providing a max in the request? I think having the client provide a value would be better than bypassing our global max value for pagination.

Or is the issue that the client is trying to provide a value >100, which isn't allowed because of the hard cap? If so, it seems this solution is just a roundabout way to bypass doing any pagination on the request. If that's the case, I'd suggest one of the following:

1) Don't make this a PaginationCommand
2) Have the client properly support pagination of the items, and while it's fetching the stock movement, loop until there's no more items to fetch.

2 is my preference if possible since it's generally good to keep our requests to a known size.

grails-app/controllers/org/pih/warehouse/core/StockMovementItemsParamsCommand.groovy
```diff
@@ -11,6 +11,13 @@ class StockMovementItemsParamsCommand extends PaginationCommand implements Valid
 
     Boolean refreshPicklistItems = true
 
+    @Override
+    Integer getMax() {
```
kchelstowski, line 15: In my opinion we shouldn't override that. 
If it turned out, that the extension of `PaginationCommand` doesn't fit here, we should get rid of that extension instead of having to override methods in order to "cheat" the `PaginationCommand`.

What I would suggest, but I assume not everyone will like that idea, is to create a `PaginationSomething` interface, that you would implement here :smile: - then, I'd be fine with such override. 
Another option is - after creating an interface, we could specify two pagination commands - one, that would have the default values, and one, that wouldn't have them (I know that in ideal world we would have this behavior aligned).
P.S. I wouldn't have an idea how to name those, so I'm fine with just implementing the interface by yourself here.

src/main/groovy/org/pih/warehouse/core/PaginationParams.groovy
```diff
@@ -0,0 +1,23 @@
...
+
+    Integer offset
+
+    Integer getMax() {
+        return max
+    }
+
+    Integer getOffset() {
+        return offset
+    }
```
kchelstowski, line 17: I think those two getters are not necessary, because Groovy/Grails adds them by default.

awalkowiak (approved):

PR #4692, opened 2024-06-25
OBPIH-6520, OBPIH-6412 Dates localization

alannadolny:
This ticket contains code for:
- OBPIH-6520 Set up for date localization in DD/Month/YYYY format for Es and Mx locales 
(Previously was in: https://github.com/openboxes/openboxes/pull/4682)
- OBPIH-6412 Localize date on inbound workflow
(Previously was in: https://github.com/openboxes/openboxes/pull/4689)

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -357,9 +357,6 @@ default.credits.label=Credits
 default.balance.label=Balance
 default.dashboard.label=Dashboard
 default.data.label=Data
-datetime.format=dd/MMM/yyyy hh:mm:ss a z
-date.format=MM/dd/yyyy
-monthyear.format=MM/yyyy
 default.date.label=Date
 default.dateCreated.label=Date Created
 default.dateFrom.label=From
@@ -656,6 +653,13 @@ document.url.label=URL
 documents.label=Documents
 documentType.label=Document Type
 documentTypes.label=Document Types
+
+# Date formats
+default.date.format=dd/MMM/yyyy hh:mm:ss a z
+common.date.format=MM/dd/yyyy
+expiry.date.format=MM/yyyy
+fullmonth.date.format=MMMM dd, yyyy
+abbreviatedmonth.date.format=dd/MMM/yyyy
 
 # Email
 email.errorReportSubject.message=Bug Report
@@ -3236,9 +3240,9 @@ react.default.dateInput.placeholder.label=Select a date
 react.default.form.field.required.label=${0} is required
 
 # Date formats
-react.default.datetime.format=MM/DD/yyyy hh:mm:ss a z
-react.default.date.format=MM/DD/yyyy
-react.default.monthyear.format=MM/yyyy
+react.default.defaultDate.format=MM/DD/YYYY HH:mm:ss Z
+react.default.commonDate.format=MM/DD/YYYY
+react.default.expiryDate.format=MM/YYYY
 
 # Notifications
 react.notification.connectivity.offline.label=Lost Connection
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -357,7 +357,9 @@ default.credits.label=Credits
 default.balance.label=Balance
 default.dashboard.label=Dashboard
 default.data.label=Data
-default.date.format=dd/MMM/yyyy hh:mm:ss a z
+datetime.format=dd/MMM/yyyy hh:mm:ss a z
+date.format=MM/dd/yyyy
+monthyear.format=MM/yyyy
 default.date.label=Date
 default.dateCreated.label=Date Created
 default.dateFrom.label=From
@@ -3232,6 +3234,12 @@ react.default.select.typeToSearch.label=Type to search
 react.default.noOptions.label=No options
 react.default.dateInput.placeholder.label=Select a date
 react.default.form.field.required.label=${0} is required
+
+# Date formats
+react.default.datetime.format=MM/DD/yyyy hh:mm:ss a z
+react.default.date.format=MM/DD/yyyy
+react.default.monthyear.format=MM/yyyy
+
 # Notifications
 react.notification.connectivity.offline.label=Lost Connection
 react.notification.connectivity.offline.message=You are now offline. Your changes will be saved when the connection is restored.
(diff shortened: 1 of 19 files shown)


grails-app/views/stockMovement/_summary.gsp
```diff
@@ -1,3 +1,4 @@
+<%@ page import="org.pih.warehouse.DateFormatName" %>
```
jmiranda, line 1: This should be in a package like "core". Unless there's a better place for it.

jmiranda: Or "util" if people feel more strongly about that.

grails-app/views/stockMovement/_summary.gsp
```diff
@@ -91,37 +92,60 @@
                         <g:if test="${!stockMovement?.dateRequested }">
                             <span class="dateRequested">
                                 <warehouse:message code="stockMovement.dateRequested.label"/>:
-                                <label><format:date obj="${stockMovement?.dateRequested}"/></label>
+                                <label>
+                                    <g:formatDate
+                                            formatName="${DateFormatName.ABBREVIATED_MONTH.property}"
+                                            date="${stockMovement?.dateRequested}"
```
jmiranda, line 98: I need more time to think about this and review what's currently happening, but I'm going to post the rest of my review as-is.

grails-app/i18n/messages.properties
```diff
@@ -655,6 +654,13 @@ documents.label=Documents
 documentType.label=Document Type
 documentTypes.label=Document Types
 
+# Date formats
+datetime.format=dd/MMM/yyyy hh:mm:ss a z
+date.format=MM/dd/yyyy
+monthyear.format=MM/yyyy
+fullmonth.date.format=MMMM dd, yyyy
+abbreviatedmonth.date.format=dd/MMM/yyyy
```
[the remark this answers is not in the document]
alannadolny: Basically, I just changed the labels because of the request in this PR: https://github.com/openboxes/openboxes/pull/4682

grails-app/i18n/messages.properties
```diff
@@ -3232,6 +3238,12 @@ react.default.select.typeToSearch.label=Type to search
 react.default.noOptions.label=No options
 react.default.dateInput.placeholder.label=Select a date
 react.default.form.field.required.label=${0} is required
+
+# Date formats
+react.default.datetime.format=MM/DD/YYYY HH:mm:ss Z
+react.default.date.format=MM/DD/YYYY
+react.default.monthyear.format=MM/YYYY
```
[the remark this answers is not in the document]
alannadolny: It's necessary because we are pulling only the labels starting with `react`... And the `react.default` are the labels we always pull.
And yes, there is a small change between the formats.

grails-app/views/stockMovement/_summary.gsp
alannadolny: I didn't know where it should be placed, so thanks for this comment.

jmiranda: Yeah let's move it to `src/main/groovy/org/pih/warehouse/core` since that's where most of the enums are located. And let's change the name of the enum to just DateFormat. Or if that's problematic because of java.text.DateFormat, then maybe DateFormatCode. Something is bothering me about DateFormatName. If anyone has any other suggestions please add it here.

grails-app/i18n/messages.properties
```diff
@@ -655,6 +654,13 @@ documents.label=Documents
 documentType.label=Document Type
 documentTypes.label=Document Types
 
+# Date formats
+datetime.format=dd/MMM/yyyy hh:mm:ss a z
+date.format=MM/dd/yyyy
```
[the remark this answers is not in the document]
jmiranda: I have confirmed that the order of precedence is: `date.format` > `default.date.format` > 'yyyy-MM-dd HH:mm:ss z' so date.format **would** override default.date.format.

https://github.com/grails/grails-gsp/blob/cae0d8182705bf1481ae53c5e2d81511356e4640/grails-plugin-gsp/src/main/groovy/org/grails/plugins/web/taglib/FormatTagLib.groovy#L181

grails-app/views/stockMovement/_summary.gsp
jmiranda: I'm trying something new with respect to making decisions as a team, so please take this poll when you have a minute.

https://github.com/openboxes/openboxes/discussions/4693

@alannadolny @awalkowiak @EWaterman @drodzewicz @kchelstowski

grails-app/i18n/messages.properties
```diff
@@ -357,7 +357,9 @@ default.credits.label=Credits
 default.balance.label=Balance
 default.dashboard.label=Dashboard
 default.data.label=Data
-default.date.format=dd/MMM/yyyy hh:mm:ss a z
+datetime.format=dd/MMM/yyyy hh:mm:ss a z
+date.format=MM/dd/yyyy
```
awalkowiak: This should stay as a `default.date.format` due to order of precedence resolving date formats

awalkowiak (commented):
Test review of specific commit

grails-app/i18n/messages.properties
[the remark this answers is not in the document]
jmiranda: Ok, let's do that.

Add this back 
```
default.date.format=dd/MMM/yyyy hh:mm:ss a z
```

And remove datetime.format and date.format.

grails-app/i18n/messages.properties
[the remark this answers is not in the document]
jmiranda: @EWaterman I see your point, but I think we're going to have a problem either way. The format/type also gets unruly when you look at all of the different date formats we're currently using.

What's missing from this discussion is a discovery of the actual date format patterns used and whether we need them all. And if not, how to standardize and consolidate. That is a rather large refactoring that I don't think we have time or energy to undertake that right now. 

> [!IMPORTANT]
> TL;DR And the fact that the initial discovery did not include a discussion on what's currently used and whether/how we should consolidate makes me realize that we shouldn't be going down the path we're heading now. We should take a step back, change all of our hard-coded formats that use MM to use MMM instead and call it a day. I'm asking Kelsey if this is ok for now.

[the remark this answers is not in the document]
jmiranda: > the fact that the initial discovery did not include a discussion on what's currently used and whether/how we should consolidate

To be clear, this is not necessarily a problem of the discovery, but a combination of the aim of the discovery i.e. the requirements, objectives, open questions). And to a large extent that falls on us (@jmiranda @EWaterman @awalkowiak). We definitely need to be more thorough with expected outcomes and what questions should be explored. And I'm hoping that will be a collaborative and probably iterative process i.e. as discovery is being done, more questions and avenues should arise. We just need to set the discovery on the right path.

PR #4697, opened 2024-06-26
OBPIH-6410 Improve country specific locale handling

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -209,6 +209,7 @@ class ApiController {
                 timezone             : timezone,
                 minimumExpirationDate: minimumExpirationDate,
                 activeLanguage       : currentLocale.toString(),
+                activeLanguageTag    : currentLocale.toLanguageTag(),
```
kchelstowski, line 212: I assume this is done, so that in case of `es-MX`, the `toLanguageTag()` would return `es`, whereas `toString()` was returning `es-MX`?

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,41 @@ class LocalizationApiController {
...
-        Locale selectedLocale = languageCode ? LocalizationUtil.getLocale(languageCode) : currentLocale
+        Locale selectedLocale = languageCode ? StringUtils.parseLocaleString(languageCode) : currentLocale
 
-        // Get the default message properties as well as the message properties for the selected locale
+        // Get the system default message properties
         Properties defaultMessageProperties = localizationService.getMessagesProperties(defaultLocale)
+
+        // Get the fallback language default message properties if it is country specific language
+        Properties fallbackMessageProperties = null
+        if (selectedLocale.country && supportedLocales.contains(selectedLocale.language)) {
```
kchelstowski, line 38: I assume the only Locale for now, that would have the `country` not `null` is the `es-MX`, and this is the case for checking the `selectedLocale.country`?
Maybe it's worth having a comment about that (just a suggestion, because I needed a few seconds to figure out that part)

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,41 @@ class LocalizationApiController {
...
+            fallbackMessageProperties = localizationService.getMessagesProperties(fallbackLocale)
+        }
+        // Get the message properties for the selected locale
         Properties selectedMessageProperties = localizationService.getMessagesProperties(selectedLocale)
 
         // Get all translations for the given prefix and locale from the database
-        List<Localization> localizedMessages = prefix ? Localization.findAllByCodeIlikeAndLocale("${prefix}%", selectedLocale.language) :
-                Localization.findAllByLocale(selectedLocale.language)
+        List<Localization> localizedMessages = prefix ? Localization.findAllByCodeIlikeAndLocale("${prefix}%", selectedLocale.toString()) :
+                Localization.findAllByLocale(selectedLocale.toString())
```
kchelstowski, line 47: The same question as under the `activeLanguageTag` but in the opposite way

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -27,43 +27,9 @@ class LocalizationUtil {
...
-            return locale
-        }
-
-        locale = new Locale(localeCode)
-        String defaultLocaleCode = Holders.config.openboxes.locale.defaultLocale
-        Locale defaultLocale = defaultLocaleCode ? new Locale(defaultLocaleCode) : Locale.ENGLISH
-        locale.setDefault(defaultLocale)
-        return locale
-    }
-
```
kchelstowski, line 63: did you get rid of it and replaced it by the `StringUtils`, because it handles exactly what this util did?

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
```diff
@@ -376,7 +376,7 @@ const mapStateToProps = state => ({
   translate: translateWithDefaultMessage(getTranslate(state.localize)),
   currencyCode: state.session.currencyCode,
   allStatuses: state.purchaseOrder.statuses,
-  locale: state.session.activeLanguage,
+  locale: state.session.activeLanguageTag,
```
kchelstowski, line 379: what's the reason for that change? I mean:
1. why do we have to change it to the `activeLanguageTag`
2. if we even have to, why does it only take place on the PO list?

src/main/groovy/util/CustomMessageSource.groovy
```diff
@@ -0,0 +1,23 @@
...
+
+class CustomMessageSource extends ResourceBundleMessageSource {
+
+    /**
+     * Custom message source resolver to deal with country specific locales.
+     * For example when we have `es_MX` we want to first fallback to the
+     * language locale -> `es`, and then to system default locale -> `en`
+     * */
+    @Override
+    protected MessageFormat resolveCode(String code, Locale locale) {
```
kchelstowski, line 14: is it a method that is now currently used behind the scenes when resolving the translations?

kchelstowski (commented):
so far looks good to me, those are only open questions, so that I can understand the context better.

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
awalkowiak: No. The language tag will be `es-MX` for Mexican Spanish, and this is required to be able to use it in the js ```Number.toLocaleString()``` or in the `moment.js` in that format.
toString returns `es_MX` which cannot be used in `toLocaleString`

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
awalkowiak: Sure I can improve the comment a little bit more

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
awalkowiak: This will resolve into `es_MX` and i think this is how it should look like in this case

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
awalkowiak: Yup

src/js/components/purchaseOrder/PurchaseOrderListTable.jsx
awalkowiak: I explained it in your first question (`Number.toLocaleString()` does not handle `es_MX` format)

src/main/groovy/util/CustomMessageSource.groovy
awalkowiak: Yes

grails-app/conf/spring/resources.groovy
```diff
@@ -0,0 +1,10 @@
+package spring
+
+import util.CustomMessageSource
+
+beans = {
+    messageSource(CustomMessageSource) {
```
jmiranda, line 6: Can you add a comment to explain the purpose of this bean as well as any other details regarding the implementation that might be helpful / pertinent for future us?

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -177,7 +177,7 @@ class ApiController {
                 supportedActivities.contains(ActivityCode.AUTOSAVE.name())
         def defaultLocale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
         def supportedLocales = locales.collect {
-            def defaultName = LocalizationUtil.getLocale(it)?.getDisplayName(currentLocale ?: defaultLocale)
+            def defaultName = StringUtils.parseLocaleString(it)?.getDisplayName(currentLocale ?: defaultLocale)
```
jmiranda, line 180: I wouldn't have been mad if you kept LocalizationUtil.getLocale(String) and implemented the method using StringUtils.parseLocaleString().

I'm guessing I was the original author of this line and I'm now looking at it like ... what in the hell. 
```
def defaultName = StringUtils.parseLocaleString(it)?.getDisplayName(currentLocale ?: defaultLocale)
```

Oh right, I think we were trying to translate the display name of the locale as it would look in that locale. For the Locale selector in the footer. That's right. 

* Acholi 
* عربى 
* Deutsche 
* English
* Español
* Suomi
* Français
* Kreyò
* lItaliano
* Português
* Chinese

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,41 @@ class LocalizationApiController {
     GrailsApplication grailsApplication
 
     def list() {
-        String languageCode = params.lang
+        String languageCode = params.languageCode
```
jmiranda, line 24: Is this so we don't have conflicts with the spring locale interceptor?

grails-app/controllers/org/pih/warehouse/user/UserController.groovy
```diff
@@ -301,7 +301,7 @@ class UserController {
         // If we are in localization mode and want to change the language through the footer on React side
         // we want to make sure we set the language to the clicked one, not for the previous locale
         // we want to set the prev locale only if we disable localization mode through "Disable localization mode" button
-        Locale locale = params?.locale ? LocalizationUtil.getLocale(params.locale) : session?.previousLocale
+        Locale locale = params?.locale ? StringUtils.parseLocaleString(params.locale) : session?.previousLocale
```
jmiranda, line 304: Now that I'm seeing this used more and more, do we really need to keep the current implementation of LocalizationUtil.getLocale()? It seems like it might be better to just leave all of these LocalizationUtil calls and change the underlying implementation to use StringUtils.parseLocaleString().
```
static Locale getLocale(String localeCode) {
    return StringUtils.parseLocaleString(localeCode)
}
```

You could even add a lenient flag to the method to switch to the more strict parseLocale method which expects a BCP 47 language tag.
 
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#parseLocale(java.lang.String)

jmiranda (commented):
Here's my first pass of the code review. I'll get back to the rest in the morning.

src/main/groovy/util/CustomMessageSource.groovy
jmiranda: I'm concerned about the side effects, but I love the idea. We can also probably come up with a better name. :)

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
awalkowiak: Yeah this triggers right now unnecessary language change when we are pulling translation after already doing the language change. Once we fully migrate to the changing locale with `lang` param this place will be adjusted accordingly, but for now this is unnecessary.

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,42 @@ class LocalizationApiController {
...
         String prefix = params.prefix
 
         String[] supportedLocales = grailsApplication.config.openboxes.locale.supportedLocales
 
+        Locale defaultLocale = Locale.default
```
jmiranda, line 29: How do you feel about moving this to LocalizationService.getDefaultLocale()?
```
LocalizationService.getDefaultLocale() {
    return Locale.default
```
That way if we want to change the behavior we can do that in one place. For example I think the default locale should actually be the one we specify in the config, not what the system thinks. 

By the way, we should document and figure out the different ways to set the default locale and what the precedence should be. It might be the case that the Application config should actually just use the JVM argument or system locale by default. Probably worth a discovery ticket.

* System default locale (sudo update-locale LANG=fr_FR.UTF-8)
* JVM argument (-Duser.language=en -Duser.region=US)
* Application Config [application.yml] (locale.defaultLocale)
* Application Config override [openboxes.yml] (local.defaultLocale)

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,42 @@ class LocalizationApiController {
...
-        // Get the default message properties as well as the message properties for the selected locale
+        // Get the system default message properties
         Properties defaultMessageProperties = localizationService.getMessagesProperties(defaultLocale)
+
+        // Get the country fallback language default message properties if it is country specific language
+        // (Example: if we have es_MX we want to have messages fallback in this precedence es_MX -> es -> en)
+        Properties fallbackMessageProperties = null
+        if (selectedLocale.country && supportedLocales.contains(selectedLocale.language)) {
+            Locale fallbackLocale = new Locale(selectedLocale.language)
+            fallbackMessageProperties = localizationService.getMessagesProperties(fallbackLocale)
```
jmiranda, line 41: I assume this is so that we can have messaage_es_MX.properties be as thin as possible, with just the message. 

But did we find out how Crowdin actually works with respect to variations of language? 

If Crowdin is going to copy all of the default (en) messages into messages_es_MX.properties by default then I think it might be worth exploring how to manage this outside of OpenBoxes i.e. we'll probably need to manually copy message_es.properties to message_es_MX.properties. The thing I worried about here is that we're doing something that we haven't put much thought into yet.

If you feel confident with this solution, let's but we should some documentation around 
* What we know about Crowdin's handling of locale + variations (region, country, etc)
* What we know about how Java work with respect to Locale + variations 
* What we know about how Message Sources work within Spring Boot 
* What we know about how Message Sources work within Grails

awalkowiak: > But did we find out how Crowdin actually works with respect to variations of language?

Not entirely (the part that I am missing and I am not sure about is how to set secondary secondary source language). However, since we wanted to rely on Spanish translations and just modify a couple of messages we should rely on the the built-in custom Localization feature and not worry about crowdin (for now I skipped adding it there, because feels like this is not required)

> If Crowdin is going to copy all of the default (en) messages into messages_es_MX.properties by default then I think it might be worth exploring how to manage this outside of OpenBoxes i.e. we'll probably need to manually copy message_es.properties to message_es_MX.properties. 

Right now the `es_MX` is not added to crowding, and it should not overwrite anything. And the purpose of `fallbackLocale` is purely to first pull Spanish messages before English messages. This is the same thing that is currently happening in the `CustomMessageResolver`. When we pivot towards using `lang` to change locales then probably we could also modify this localization API to rely on the message code resolver too.

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
```diff
@@ -21,30 +21,42 @@ class LocalizationApiController {
...
+        }
+        // Get the message properties for the selected locale
         Properties selectedMessageProperties = localizationService.getMessagesProperties(selectedLocale)
 
         // Get all translations for the given prefix and locale from the database
-        List<Localization> localizedMessages = prefix ? Localization.findAllByCodeIlikeAndLocale("${prefix}%", selectedLocale.language) :
-                Localization.findAllByLocale(selectedLocale.language)
+        List<Localization> localizedMessages = prefix ? Localization.findAllByCodeIlikeAndLocale("${prefix}%", selectedLocale.toString()) :
+                Localization.findAllByLocale(selectedLocale.toString())
```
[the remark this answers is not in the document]
jmiranda: For the record, these suggestions ^^ are for the longer term solution not for this PR. We should create a discovery or tech debt for any/all of the bullets. In the very near future we should definitely try to add a test or two to give us a sense of confidence that it will be working as expected.

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
jmiranda: If it works the way it did, then no change is required. 

And I agree that we should consolidate the additional scope we've identified, add into a ticket or tickets and work on them based on the trigger points I mentioned in one of the other comments.

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
jmiranda: My last comment was being written before your last comment, so it's not a response to that. 

All of what you wrote above is very helpful. Thank you. If we haven't already we should try to add as much of that information to the source code as comments (not sure where). If there's not a logical place for these comments we should at least create a discovery page for all of the known knowns and known unknowns we've identified from this feature.

jmiranda: > Right now the es_MX is not added to crowdin,

If nothing else, this information needs to be conveyed to Manon and Kelsey (and subsequently their translators) to make sure we don't translate when logged into the es_MX locale. We should even consider adding a feature to disable translation in this locale (if possible) OR have a process to deal with translations that may come through unsupported locales (again, if that's even possible).

src/main/groovy/util/CustomMessageSource.groovy
jmiranda: I can't think of a good name myself, but wondering if you have any more insight into what it should be called. Maybe `CascadingResourceBundleMessageSource`?

Otherwise, let's go with CustomResourceBundleMessageSource.

src/main/groovy/util/CustomMessageSource.groovy
```diff
@@ -0,0 +1,23 @@
...
+        MessageFormat messageFormat = super.resolveCode(code, locale)
+        // If message not found try to fallback to language locale first (if it is country
+        // specific locale)
+        if (!messageFormat && locale?.country) {
+            messageFormat = super.resolveCode(code, new Locale(locale.language))
+        }
+        return messageFormat ?: super.resolveCode(code, Locale.ENGLISH)
```
jmiranda, line 21: Are we sure we need to specify a locale here? Does this resolve the code in the default messages.properties or messages_en.properties? Since message_en.properties doesn't exist, I'm guessing this might be equivalent so should be no harm. But I think we should know what's actually happening.

src/main/groovy/util/CustomMessageSource.groovy
jmiranda: Turns out that HierarchicalMessageSource is an interface implemented by ResourceBundleMessageSource. So now my head is spinning :face_with_diagonal_mouth: Does that change anything for you @awalkowiak? Does ResourceBundleMessageSource not do the cascading through the resource bundles?

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/HierarchicalMessageSource.html

grails-app/controllers/org/pih/warehouse/api/LocalizationApiController.groovy
awalkowiak: When you are entering the crowdin localization mode you have to select one of the configured in crowdin languages, you cannot enter es_MX to translate it, so we should be safe here. But I will let Kelsey know about it, so we are aligned

PR #4700, opened 2024-06-27
OBPIH-6520 Set up for date localization in DD/Month/YYYY format for Es and Mx locales

alannadolny:


grails-app/controllers/org/pih/warehouse/core/DateFormat.groovy
```diff
@@ -0,0 +1,22 @@
...
+// 1. G tag - <g:formatDate date={your date} formatName={enum property} />
+// 2. LocalizationUtil.formatDate(your date, enum property)
+
+enum DateFormat {
+    COMMON('common.date.format'),
+    EXPIRY('expiry.date.format'),
+    FULL_MONTH('fullmonth.date.format'),
+    ABBREVIATED_MONTH('abbreviatedmonth.date.format'),
+
+    final String property
```
jmiranda, line 17: `property` is ambiquous so let's go with `code` instead. While also ambiquous there's precendence for using the `code` when dealing with localization keys (see validation errors, namely BeanPropertyBindingResult/ObjectError).

grails-app/i18n/messages.properties
```diff
@@ -655,6 +654,13 @@ documents.label=Documents
 documentType.label=Document Type
 documentTypes.label=Document Types
 
+# Date formats
+default.date.format=dd/MMM/yyyy hh:mm:ss a z
+common.date.format=MM/dd/yyyy
+expiry.date.format=MM/yyyy
+fullmonth.date.format=MMMM dd, yyyy
+abbreviatedmonth.date.format=dd/MMM/yyyy
```
jmiranda, line 662: As I mentioned, I don't like this convention. It's too specific and you've broken the convention immediately upon adding the date formats to es-MX.
```
fullmonth.date.format=dd/MMM/yyyy
```

Is there a better way of doing this that allows us to choose a format by type/style, while also being non-committal on the actual implementation?

It's a built in feature using type and style attributes rather than formatName. 
```
<g:formatDate date="${date}" type="date" style="SHORT"/>
<g:formatDate date="${date}" type="time" style="LONG"/>
<g:formatDate date="${date}" type="datetime" style="MEDIUM"/>
<g:formatDate date="${date}" type="datetime" style="LONG" timeStyle="SHORT"/>
```

So if we're going down this path, then let's use type and style and then you get to go with type/style convention while also being somewhat customizable. It limits the combinations allowed, but gives us a rigidity / convention that I think we're looking for. 

Here's the path forward. We have two options. 

1. we hard-code MMM into all date formats and leave this larger refactoring for later
1. we continue down the refactoring path and we use a combination of the following, but use type and style for most date, time, and datetime formatting
    * use type and style format almost everything
        * type=date,time,datetime
        * style=short,medium,long
    * use format name for things where there's a pretty standard convention based on purpose
        * expiry.date.format (MMM/yyyy)
        * auditing.date.format image
        * auditing.datetime.format image
    * use neither accept the default for all other dates where the default format is ok

jmiranda (commented):
Partial review to discuss with Artur.

jmiranda:
Here are my discovery notes (thought process) for this and why I don't think we're ready for making these change.

---

I'm now going to walk through my thought process again. While this is going to be verbose and tedius, it would be helpful for everyone to read it because this is how discovery should be done. This outlines how I thought about the problem and how I need you to think about these problems going forward.

Here's a quick snapshot of the date formats used just in our GSPs. 
```
$ grep "formatDate" | grep -Po "format=\".*?\"" grails-app/views/*/*.gsp | cut -d "=" -f2 | sort | uniq -c | sort -nr 
     40 "MM/dd/yyyy"
     38 "d MMM yyyy"
     29 "hh:mma"
     28 "MMM d, yyyy"
     15 "dd/MMM/yyyy"
     14 "MMMMM dd, yyyy hh:mma"
     10 "MMMM dd, yyyy"
      5 "MMM d, yyyy hh:mma"
      5 "MMM dd, yyyy"
      3 "MMM yyyy"
      3 "MMM d, yyyy  hh:mma"
      3 "MMM dd, yyyy hh:mma z"
      3 "d MMMMM yyyy hh:mma"
      3 "d MMMMM yyyy  hh:mma"
      2 "MMM d hh:mma"
      2 "d MMM yyyyy"
      2 "dd MMM yyyy"
      1 "MMMMM dd, yyyy"
      1 "MMM dd HH:mm z"
      1 "d MMMMM yyyy hh:mm a z"
```

Taking 4 of the top 5, how should we classify these if we needed each as a separate format? By purpose or format? 
```
     40 "MM/dd/yyyy"
     38 "d MMM yyyy"
     28 "MMM d, yyyy"
     15 "dd/MMM/yyyy"
```

If we go by format, we'd be left with something like this ... 

* MONTH_DAY_YEAR
* DAY_MONTHNAME_YEAR
* MONTHNAME_DAY_YEAR
* DAY_MONTHNAME_YEAR_WITH_SLASH

Using a "purpose" convention would also put us into a similar spot. 

* default.date.format
* yourFavorite.date.format
* someOther.date.format
* better.default.date.format

Seeing this sparks the following questions

* Do we actually need all of these unique date formats? 
* Is there a way to consolidate around a classification (i.e. type or purpose)? 
* Or should we use the convention @EWaterman suggests and create separate date formats for each format used?

We can't answer those questions because we did not do a thorough discovery. We at least should have understood that part of the work for this ticket was going to be consolidating date formats. Not having that as part of the discovery causes a lot of problems because there's too much noise to come up with a proper design. 

In my mind, there will probably be about 5 classifications we need right out of the box. 

* Display date and time together 
* Display just the date on its own 
* Display just the time, split from the date, but usually displayed below or nearby. 
* Display an expiration date 
* Display a long friendly date with day and month names for reports 

What we call them is not important to me. What is important is whether these five date format classifications are enough to satisfy all of our date formatting (display) needs? 

Or do we need different variations of the date formats for different purposes. That is a probably a question for Manon and Kelsey.

But here's how the date formats explode when there's no standard or convention from the beginning. 

* Display date and time together 
    * 01/01/2025 12:00 AM EDT
    * January 01, 2025 12:00:00 AM EDT
    * Jan 1, 2025 00:00:00 EDT
* Display just the date on its own 
    * Jan 01, 2025
    * 01/01/2025
    * 01/Jan/2025
* Display just the time, split from the date, but usually displayed below or nearby. 
    * 00:00:00 EDT
    * 12:00:00 AM EDT
* Display an expiration date 
    * 01/2025
    * Jan 2025
    * January 2025
* Display a long friendly date with day and month names for reports 
   * Monday, January 1st, 2025 12:00:00 AM EDT
* yada yada yada 

I could see the potential and make the argument for using all of these date formats in different contexts. But we need to start from a single date format that works for the five cases above and then make explicit decisions on whether the rest are needed.

From there, we can ask the question 

* How should we classify / name these different date formats? By purpose? By implementation? A mix of both?

We could then just create the default formats for each of these 
```
default.date.format=d/MMM/yyyy hh:mm:ss a z
default.datetime.format=d/MMM/yyyy hh:mm:ss a z
default.time.format=hh:mm:ss a z
common.date.format=d/MMM/yyyy
expiry.date.format=MMM/yyyy
auditing.date.format=MMM d, yyyy
auditing.datetime.format=d/MMM/yyyy hh:mm:ss a z
friendly.date.format=EEEEE, MMMMM d, yyyy
```
> [!NOTE]
> One problem that I don't have an answer for is that default.date.format is a misnomer. Grails uses this as the default when displaying date and time, so we need to have a default date format for just date. For that I was using common.date.format, but I'm open to suggestions.

Now, we need to ask questions to consider whether we need variations of those date formats.
* How many different formats do we need for that? 
* What's the default for each? 
* Does we always want to include timezone? 
* Do we want to display military time or do we want to always use the AM/PM designation?

Does one default date format like dd MMM yyyy work for everyone?

The discovery process should have had an expectation set to investigate how many different date formats we use. It's my fault that we didn't specify that as an explicit expectation. I didn't really think of it until I did the discovery myself, so to some degree this is just one of those things that I expect to happen as part of discovery. But I will try to do a better job of setting expecations ahead of time.

We would have also seen that **most** of the date formats already include the month name (Jan) instead of the month number (01) and should have asked the question ... can we just ALWAYS display the month name for now? That would solve the immediate requirement while giving us time to do a thorough discovery.

PR #4703, opened 2024-06-27
OBPIH-6403 Unhide rejected requests from the requestor dashboard (war…

kchelstowski:
…d-depot case)

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
@@ -432,15 +432,18 @@ class StockMovementService {
         Date createdAfter = params.createdAfter ? Date.parse("MM/dd/yyyy", params.createdAfter) : null
         Date createdBefore = params.createdBefore ? Date.parse("MM/dd/yyyy", params.createdBefore) : null
         List<ShipmentType> shipmentTypes = params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
+        Location currentLocation = AuthService.currentLocation
 
         PagedResultList shipments = Shipment.createCriteria().list(max: max, offset: offset) {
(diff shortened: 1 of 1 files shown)


awalkowiak (approved):

grails-app/services/org/pih/warehouse/inventory/StockMovementService.groovy
```diff
@@ -432,15 +432,18 @@ class StockMovementService {
         Date createdAfter = params.createdAfter ? Date.parse("MM/dd/yyyy", params.createdAfter) : null
         Date createdBefore = params.createdBefore ? Date.parse("MM/dd/yyyy", params.createdBefore) : null
         List<ShipmentType> shipmentTypes = params.list("shipmentType") ? params.list("shipmentType").collect{ ShipmentType.read(it) } : null
+        Location currentLocation = AuthService.currentLocation
 
         PagedResultList shipments = Shipment.createCriteria().list(max: max, offset: offset) {
-            // OBPIH-6403: We want to hide SMs with requisition of status REJECTED from the inbound list
+            // OBPIH-6403: We want to hide SMs with requisition of status REJECTED from the inbound list (only for the depot-depot case!!)
```
EWaterman, line 438: I'm going to approve this review, but I think it'd be good for us to start working through ways to solve this kind of problem more generally.

I notice we have a ton of conditional blocks in our code that cause different behaviours depending on what every specific state we're in (such as this one wanting to hide rejected SMs but only for depot-depot). Putting it all in the code like this makes us pretty inflexible to change in the future. I think we'd benefit from trying to make this more data driven whenever possible.

For example, I think it'd be great if we could have a sort of StockMovementConfig table that can define some settings around things like this via data. We could let it specify the conditions on which a stock movement is hidden, and then use those settings here when fetching the shipment. Essentially it moves all these if blocks into data so that they can be more fluidly configured on a case by case basis. Then we can have an admin page that lets us configure these settings per location.

Anyways, this is more just me dumping my thoughts. No need to consider that for this change.

PR #4710, opened 2024-07-01
OBPIH-6457 Localize PO import/export

kchelstowski:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -840,9 +843,11 @@ class OrderService {
                     orderItem.recipient = recipient ? personService.getPersonByNames(recipient) : null
 
                     def estReadyDate = null
+                    Locale locale = LocalizationUtil.currentLocale
+                    SimpleDateFormat readyDateFormat = new SimpleDateFormat(LocalizationUtil.getLocalizedOrderImportDateFormat(locale))
```
awalkowiak, line 847: I am not sure about this part. What if someone changed a locale after doing export and while filling in the template with dates that follow the format that is in the header? Should we catch this case and throw an exception that the current locale does not match the one that the export was made on? Or we don't care about it and rely on training? Feels like this **might** cause some confusion and support issues and perhaps this is not the best approach.

kchelstowski: I feel like we should not cover this part. As I told you, getting the info about header in this method would be a pain in the neck, hence I assumed that if someone is using `es_MX`, he/she would use the same template for export and import. 

On the other hand we can say - ok, if we coverered this edge case, then what if someone uses `es_MX`, expects to import the date in the `dd/MM/yyyy` format, but unfortunately the header contains `MM/dd/yyyy` - what should we prioritize then - the template's header or the actual locale? My mind says the locale is more important that the header :thinking:

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -148,4 +149,16 @@ class LocalizationUtil {
...
     }
+
+
+    static boolean isSpanishMexico(Locale locale) {
+        return locale.language == "es" && locale.country == "MX"
+    }
+
+    static String getLocalizedOrderImportDateFormat(Locale locale) {
+        if (isSpanishMexico(locale)) {
+            return Constants.EUROPEAN_DATE_FORMAT
```
awalkowiak, line 160: Perhaps it is worth having separate const for this purpose and not use the `EUROPEAN_DATE_FORMAT`?

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -148,4 +149,16 @@ class LocalizationUtil {
...
+        }
+        return Constants.EXPIRATION_DATE_FORMAT
```
awalkowiak, line 162: Same as above

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -148,4 +149,16 @@ class LocalizationUtil {
     static String getDefaultString(String str) {
         return getLocalizedString(str, null)
     }
+
+
+    static boolean isSpanishMexico(Locale locale) {
+        return locale.language == "es" && locale.country == "MX"
```
awalkowiak, line 155: I wonder if this could look a bit better, but I guess for now at least I'm ok with it.

Other options:
```groovy
return locale == new Locale("es", "MX")
```
or
```groovy
return locale == LocalizationUtils.getLocale("es-MX") // or es_MX, both works
```

grails-app/services/org/pih/warehouse/order/OrderService.groovy
awalkowiak: I would say that the header is more important in this case because it contains the format that is meant to be used in the column. 

But yeah I agree feels like this might cause more headaches than it is required and what you have is enough as a quick solution.

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
kchelstowski: sure, if the proper `equals` method is implemented in the `Locale` class, so that your suggestions will work, I'll change it

src/test/groovy/unit/org/pih/warehouse/core/LocalizationUtilTest.groovy
```diff
@@ -6,35 +6,52 @@
...
+
+		Locale spanish = new Locale("es")
+
+
+		assert LocalizationUtil.isSpanishMexico(spanishMexico)
+
+		assert !LocalizationUtil.isSpanishMexico(english)
+
+        assert !LocalizationUtil.isSpanishMexico(spanish)
+	}
```
kchelstowski, line 54: FYI: it just looks weird on github. Locally the formatting looks fine:
Screenshot from 2024-07-02 10-16-25

grails-app/services/org/pih/warehouse/order/OrderService.groovy
alannadolny: >I would say that the header is more important in this case because it contains the format that is meant to be used in the column.
+1

grails-app/utils/org/pih/warehouse/LocalizationUtil.groovy
```diff
@@ -148,4 +149,16 @@ class LocalizationUtil {
     static String getDefaultString(String str) {
         return getLocalizedString(str, null)
     }
+
+
+    static boolean isSpanishMexico(Locale locale) {
+        return locale == new Locale("es", "MX")
+    }
```
alannadolny, line 156: I feel like it should be more generic. Going this path we should create a function for checking every locale we have?

kchelstowski: I don't expect methods for all of the locales in the near future. I think for now this method is self-explanatory and it's not a good time to overthink it, is it?  
If we wanted to do it generic it would be probably as simple as passing an additional param like `localeToCompare`, but the question is, if we feel strong that we should do it this way?

awalkowiak (commented):
I am merging this as is. We can discuss on the tech huddle if this is a short-term or a long-term solution, and if this requires some further improvements after the release.

PR #4712, opened 2024-07-01
OBPIH-6347 enable search by synonym in putaway original product name

drodzewicz:


src/js/components/put-away/PutAwayPage.jsx
```diff
@@ -369,10 +378,13 @@ class PutAwayPage extends Component {
...
-    const val = filter.id === 'inventoryItem.expirationDate'
-      ? this.props.formatLocalizedDate(rowData, DateFormat.COMMON)
-      : rowData;
+    let val = row[filter.id];
+    if (filter.id === 'inventoryItem.expirationDate') {
+      val = this.props.formatLocalizedDate(val, DateFormat.COMMON);
+    }
+    if (filter.id === 'product') {
+      val = val ? `${val.name} ${val.displayNameOrDefaultName}` : null;
+    }
```
kchelstowski, line 387: just of out curiosity - how does it work in the background? I can see such a filter method for the first time and it looks (at least) weird.

src/js/components/put-away/SplitLineModal.jsx
```diff
@@ -202,7 +201,22 @@ class SplitLineModal extends Component {
...
+            <h3 className="font-weight-bold">
+              <span>{this.props.putawayItem.product.productCode}</span>
+              <Tooltip
+                arrow="true"
+                delay="150"
+                duration="250"
+                hideDelay="50"
+                className="text-overflow-ellipsis"
+                disabled={this.props.putawayItem.product?.name ==
+                  this.props.putawayItem.product?.displayNameOrDefaultName}
```
kchelstowski, line 213: 1. why `==` instead of `===`
2. can't we just check if `disabled={!this.props.putawayItem.product?.displayNameOrDefaultName}`

drodzewicz: double equals instead of a single is just a typo, but as for `disabled={!this.props.putawayItem.product?.displayNameOrDefaultName}`
displayNameOrDefaultName is always populated, jus tlike the name suggests it returns displayName when it is present or default name

src/js/components/put-away/PutAwayPage.jsx
drodzewicz: it's a filter method that is build ion ReactTable. When we want to filter the table column we must check for the key (column accessory ) like
```if (filter.id === 'product')```
and return a boolean value if criteria matches the row value

src/js/components/put-away/SplitLineModal.jsx
kchelstowski: oh, I'm sorry, I have not noticed you used the `displayNameOrDefaultName`.

kchelstowski (approved):

PR #4716, opened 2024-07-02
OBPIH-6410 Make Locale name in the footer customizable

awalkowiak:


grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
```diff
@@ -183,7 +183,7 @@ class ApiController {
         def defaultLocale = new Locale(grailsApplication.config.openboxes.locale.defaultLocale ?: "en")
         def supportedLocales = locales.collect {
             def defaultName = LocalizationUtil.getLocale(it)?.getDisplayName(currentLocale ?: defaultLocale)
-            def name =  messageSource.getMessage("locale.${it}.label", null,  defaultName, (currentLocale ?: defaultLocale))
+            def name =  "${warehouse.message(code: "locale.${it}.label", default: defaultName)}"
```
alannadolny, line 186: shouldn't it be g.message?

alannadolny (changes requested):

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
awalkowiak: No, because `g.message` does not fall back to the custom localization feature. As you can see, the other line was also a `g.message`, but I want to allow customization of these labels. There was a request from Kelsey to somehow better distinguish `Español` from `Español (México)` with for example `Español (Spain)` or something like that, and I thought it would be better to make this part customizable instead of hard coding these translations into message properties.

alannadolny: Hmmm ok, so I am fine with that, but we have to remember that we wanted to get rid of `warehouse:message`

alannadolny (approved):

kchelstowski (approved):

drodzewicz (approved):

grails-app/controllers/org/pih/warehouse/api/ApiController.groovy
jmiranda: @awalkowiak LGTM but just wanted to include two points to add context to the discussion.

## TL;DR: None of these things are actionable - just FYIs
1. Consolidate the logic to localize the locales
1. To use g:message or warehouse:message? 

## The deets

### 1. Consolidate the logic to localize the locales

As you are likely aware (I assume you based your solution on _footer.gsp), _footer.gsp does this already. I just wanted to remind you about it since we might want to consolidate this logic into a centralized utility class at some point. Not now, but I'll add a tech debt entry for it. 

### 2. To use g:message or warehouse:message?

Leave it as warehouse:message for now. At least until we decide whether supporting custom translations is worthwhile. 

This is debatable and I've gone back and forth on this one over the years. warehouse:message is the taglib that pulls from the database if custom localization is enabled. This feature became less important once we moved to Crowdin, but I like the idea of keeping it around for times where users want to override a particular translation without having to create their own custom locale (e.g. Product Code -> SKU, each -> piece). So I'm ok with keeping this feature (and maybe expanding it) but we should discuss how we want it to work and plan around it. 

Here's what I would want to do at a minimum 
* Build an actual database-backed message source
* Refactor the localization table to allow for more usages (i.e. synonyms, referential data, etc).
* Cache the messages within the database table for faster retrieval
* Convert the Message taglib with "warehouse" namespace to use either the default namespace "g" (my preference) or a namespace more suitable like "ob" or "openboxes". Then find-and-replace "warehouse:message" with "g:message"

Aside: The "warehouse" namespace was chosen before we came up with the name OpenBoxes.

awalkowiak: @jmiranda these are good points. 
1. This would be a good further improvement. As it was a specific request from Kelsey, I wanted to add a solution that will require the least amount of changes and complications, but you have a fair point here, that we should have one logic pulling footer data.
2. I feel like this feature is quite useful for custom implementations, and I'd rather suggest expanding this in the future, rather than getting rid of it.

PR #4720, opened 2024-07-03
OBPIH-6449 add attributes for tests

drodzewicz:


alannadolny (changes requested):
⠀

alannadolny (approved):

jmiranda (commented):
> [!Important]
> TL;DR: We should add a discussion about Accessibility to our tech huddle topics (or as a Discovery) to make sure we're heading in the right direction. 

@drodzewicz :face_with_head_bandage: Please don't hit me ... (and please note that I do not know the answer to this question, nor have a strong opinion about it).

... but do we need to localize these Aria labels? If we're going to put in the effort to handle accessibility, we should also keep language accessibility in mind.

Here's one opinion on the subject

> ARIA attributes that include human readable text are part of it and need to be translated.

Source: https://stackoverflow.com/questions/73163108/should-aria-attributes-also-be-translated-if-the-page-is-using-translations

Secondly, since this is a change that could impact the application (not just tests) ...

1. Can you explain the \<div\> to \<section\> change? 
2. Do you know if there's any impact related to styles with this change? There are definitely places in the CSS where we use `div.<className>`.

I might argue that the div should stay, but we could wrap it with a section? But if we know definitively that this change is not going to cause any issues, then I'm ok with it.

drodzewicz:
> Can you explain the `<div>` to `<section>` change?
> Do you know if there's any impact related to styles with this change? There are definitely places in the CSS where we use div.`<className>`.

Regarding changes to some tag I am well aware of some of the implications that they can cause and when changing it to another tag I was checking if it didn't break anything styling or js behavior.
The reason why I was changing some of these tags from `<div />` to for eg. `<section />` is to have a more abstract query on testing side on some areas of the page.
Basically, when writing element selectors I try to be as abstract as possible so that any class change or a little location change of the element does not break the test. So instead of writing selectors like `table#line-items-table > button:nth-child(2)` playwright gives me the ability to select an element by its role and label like for example `getByRole('button', { name: 'Submit' })`, so any time this button changes styles or any other insignificant change my test won't break because we are looking at an abstract structure of the document a looking for a button with label submit.

An alternative approach is also add a `data-testid` which is also very common in frontend testing, but I thought that since I have the opportunity I can fix some of the semantics of the document.
We can talk about this on the tech-huddle and if we come to the conclusion that we "don't want to risk" any document change I can stick with adding label attributes and `data-testid`

> do we need to localize these Aria labels?

It's a good point I had thought about it but didn't dive deep into this idea, I don't have an opinion at the moment so we can discuss this on the tech-huddle as well 
@jmiranda

awalkowiak (approved):
lgtm, but if needed we can discuss the topics above on the tech huddle

PR #4722, opened 2024-07-03
OBPIH-6535 Custom PO Template - Mexico

alannadolny:


grails-app/services/org/pih/warehouse/core/DocumentTemplateService.groovy
```diff
@@ -146,7 +148,18 @@ class DocumentTemplateService {
                 unitOfMeasure: k.unitOfMeasure,
                 unitPrice: v?.unitPrice?.first(),
                 quantity: v?.sum { it.quantity },
-                totalPrice: v?.sum { it.totalPrice() }
+                totalPrice: v?.sum { it.totalPrice() },
+                budgetCode: v?.budgetCode?.first()?.code ?: "",
+                expectedShippingDate: v?.estimatedReadyDate?.first() ? v?.estimatedReadyDate?.first()?.format(Constants.EUROPEAN_DATE_FORMAT) : ""
```
alannadolny, line 153: I had to hardcode the format here, because I wasn't able to use the formatting in the document directly, because it's rendered as a table, so calling `expectedShippingDate?anyFunction` is called on the sequence, not on the one element. I also tried using some functions that I found in the Freemarker documentation, but we have too old version of this plugin.

In the rest of the places that needed formatting, I used the formatting function that I found in docs:
```
${maxExpectedReadyDate?string["dd/MM/yyyy"]}
```

awalkowiak (approved):

PR #4728, opened 2024-07-08
OBPIH-6551 Rename the edit button in the Spanish locale in the receiving workflow

alannadolny:


grails-app/i18n/messages_es.properties
```diff
@@ -3183,7 +3183,7 @@ react.default.button.next.label=Siguiente
 react.default.button.submitRequest.label=Submit request
 react.default.button.submitForApproval.label=Submit for approval
 react.default.button.finish.label=Finalizar
-react.default.button.edit.label=Editar sección
+react.default.button.edit.label=Editar
```
awalkowiak, line 3186: I think this should be done through Crowdin, because I think it will want to overwrite it back to the previous version

awalkowiak (changes requested):

PR #4737, opened 2024-07-10
OBPIH-6425 Make the header labels in the export of PO shipment transl…

kchelstowski:
…atable

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
@@ -184,7 +184,7 @@ class CombinedShipmentItemApiController {
             }
         }
 
-        String filename = g.message(code: "orderItem.template.header.label", default: "Order-items-template")
+        String filename = g.message(code: "orderItem.template.fileName.label", default: "Order-items-template")
 
         response.setHeader("Content-disposition", "attachment; filename=\"${filename}.csv\"")
         render(contentType: "text/csv", text: sw.toString(), encoding: "UTF-8")
--- a/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
@@ -11,6 +11,8 @@ package org.pih.warehouse.api
 
 import grails.plugins.csv.CSVWriter
 import grails.converters.JSON
+import grails.util.Holders
+import org.grails.plugins.web.taglib.ApplicationTagLib
 import org.grails.web.json.JSONObject
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.importer.ImportDataCommand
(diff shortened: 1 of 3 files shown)


grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -112,23 +114,29 @@ class CombinedShipmentItemApiController {
...
     }
 
+    ApplicationTagLib getApplicationTagLib() {
+        return Holders.grailsApplication.mainContext.getBean(ApplicationTagLib)
+    }
+
     def exportTemplate() {
         def sw = new StringWriter()
 
+        ApplicationTagLib g = applicationTagLib
```
alannadolny, line 124: Is there a reason for that change? I prefer using applicationTagLib instead of g in the code below.

grails-app/i18n/messages.properties
```diff
@@ -3843,6 +3843,7 @@ react.combinedShipments.selectOrders.label=Select orders...
 react.combinedShipments.selectProduct.label=Select product...
 react.combinedShipments.availableItems.label=Available order items
 react.combinedShipments.blankTemplate.label=Blank import template
+react.combinedShipments.template.header.label=Order-items-template
```
alannadolny, line 3846: We have to remember that this one won't be testable for now

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
EWaterman: yeah I generally try to avoid abbreviated variable names to avoid confusion

EWaterman: yeah that makes sense to be consistent with what our gsp files are doing. Do you know if that `<g:..>` is something we started or if it's a built in grails thing?

grails-app/i18n/messages.properties
EWaterman: This is for the name of the file that we generate, right? Why is the label `template.header.label` and not `template.fileName.label`? Or am I misunderstanding

kchelstowski: @EWaterman yeah, good point, all the time I had in mind the "header" word, as I was translating the headers in the template, but here, this definitely should be `fileName`.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
kchelstowski: @EWaterman it's a built in grails thing. 
So are `<g:if>`, `<g:hasErrors>` etc.

kchelstowski:
@EWaterman I totally forgot about this one, hence I've just pushed the minor change you proposed.

grails-app/controllers/org/pih/warehouse/api/CombinedShipmentItemApiController.groovy
```diff
@@ -176,7 +184,9 @@ class CombinedShipmentItemApiController {
             }
         }
 
-        response.setHeader("Content-disposition", "attachment; filename=\"Order-items-template.csv\"")
+        String filename = g.message(code: "orderItem.template.header.label", default: "Order-items-template")
```
EWaterman, line 187: is this code correct?

kchelstowski: thanks for catching that - I forgot I included it also in the response header.

EWaterman (approved):

PR #4738, opened 2024-07-11
OBPIH-6565 Fix ability to set time properly on ship date field in req…

kchelstowski:
…uests/outbounds + fix display of the time there

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/SendMovementPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/SendMovementPage.jsx
@@ -94,6 +94,8 @@ const SHIPMENT_FIELDS = {
     attributes: {
       localizeDate: true,
       localizedDateFormat: DateFormat.DEFAULT,
+      // It's necessary for properly setting up the time part of the localized date format
+      dateFormat: 'MM/DD/YYYY HH:mm Z',
       required: true,
       showTimeSelect: true,
       autoComplete: 'off',


alannadolny (approved):

src/js/components/stock-movement-wizard/outbound/SendMovementPage.jsx
```diff
@@ -94,6 +94,8 @@ const SHIPMENT_FIELDS = {
     attributes: {
       localizeDate: true,
       localizedDateFormat: DateFormat.DEFAULT,
+      // It's necessary for properly setting up the time part of the localized date format
+      dateFormat: 'MM/DD/YYYY HH:mm Z',
```
EWaterman, line 98: I'm not fully caught up on the date formatting stuff. is there a reason for re-adding the dateFormat attribute instead of adding `localizedDateFormat: DateFormat.SOMETHING_ELSE`?

kchelstowski: This dateFormat is responsible for what is displayed, and the other one (DateFormat.SOMETHING_ELSE) is responsible for what is sent to the backend (or in the opposite, I don’t remember now).
I was only re-using the logic, that @alannadolny had implemented for this, so he would be able to answer you in more details, or to correct me if I’m wrong.
My understanding is that even if we display a date in Mexico, we still send it to save in a „standardized” format (MM/dd/yyyy), and this is why we need both formats there.

alannadolny: It's because we had to use the old date picker component, which is hard to modify. This part is needed to appropriately pick the time (there were some issues with moment.js because at the beginning we had only labels for the format, not the format, and the moment.js had some trouble dealing with that).

PR #4744, opened 2024-07-12
OBPIH-6570 Status jumps from edit to pick after refresh

alannadolny:


kchelstowski (approved):
I'm actually surprised, that when a method's definition has two arguments, in this case `(callback, status)`, and you call it only with one argument, it would still be called without any exception, and the second argument will evaluate to undefined.

src/js/components/returns/outbound/AddItemsPage.jsx
```diff
@@ -323,7 +324,7 @@ class AddItemsPage extends Component {
 
   dataFetched = false;
 
-  saveAndTransition(callback) {
+  saveAndTransition(callback, status) {
```
EWaterman, line 327: Having a default value could be clearer: `saveAndTransition(callback, status=null) {`

Though I haven't worked with react much so maybe it's well enough understood behaviour that it'll be undefined if not provided

EWaterman (approved):

drodzewicz (approved):

PR #4745, opened 2024-07-12
OBPIH-6533 Hide Import complete outbound from menu

drodzewicz:
Add ability to enable and disable individual navItems in the config

### DIFF
--- a/grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
+++ b/grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
@@ -87,6 +87,9 @@ class MegamenuService {
     List buildAndTranslateMenuItems(List menuItems, User user, Location location) {
         def builtMenuItems = []
         menuItems.each {
+            if (it.enabled == false) {
+                return
+            }
             def minRole = it.minimumRequiredRole
             if (minRole && !userService.isUserInRole(user, minRole)) {
                 return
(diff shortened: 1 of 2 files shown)


kchelstowski (approved):
we were supposed to only remove this subsection, but since you came up with a solution to hide a specific subsection, this looks good to me

grails-app/services/org/pih/warehouse/api/MegamenuService.groovy
```diff
@@ -87,6 +87,9 @@ class MegamenuService {
     List buildAndTranslateMenuItems(List menuItems, User user, Location location) {
         def builtMenuItems = []
         menuItems.each {
+            if (it.enabled == false) {
```
EWaterman, line 90: how is this working? If any menu item is disabled we disable all menu items?

kchelstowski: @EWaterman only the one that is `enabled: false`. When we return from the `.each` loop, it will continue with the iterations - this `.each` goes through the subsections of a section, hence in this case only the Import Completed Outbound would be hidden

drodzewicz: yes, exactly like Kacper explained.
This is the same logic we have on nav item sections, we just didn't have this logic for individual items, so this was basically copy and paste

EWaterman: hunh interesting. I was assuming the return would break the whole loop and not just be a continue.

EWaterman (approved):

PR #4746, opened 2024-07-15
OBPIH-6559 Fix creation of duplicate shipments while editing an inbou…

kchelstowski:
…nd return

I will post an explanation of what is going on in the ticket

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -286,6 +286,8 @@ class Order implements Serializable {
     }
 
     def getShipments() {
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -286,7 +286,11 @@ class Order implements Serializable {
     }
 
     def getShipments() {
```
EWaterman, line 288: can you add a brief comment in the code here as to why we need to filter out nulls? It's a complex enough situation to not immediately be obvious.

EWaterman (approved):

drodzewicz (approved):

PR #4748, opened 2024-07-16
OBPIH-6576 Remove case with returning ward locations for superuser on inbound

drodzewicz:


grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -149,28 +149,15 @@ class LocationService {
...
-            }
-            if (direction == "OUTBOUND") {
-                return outboundMovementLocations
-            }
-        } else {
-            if (direction == "INBOUND") {
-                return locations.findAll {
-                    it.locationType.locationTypeCode == LocationTypeCode.SUPPLIER || !it.supports(ActivityCode.MANAGE_INVENTORY)
-                }
+        if (direction == "INBOUND") {
```
alannadolny, line 152: LGTM, but I think we can improve the code (I know it was done like that previously), maybe you can use the StockMovemntDirection enum?

alannadolny (approved):

grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -148,29 +149,15 @@ class LocationService {
...
-                return outboundMovementLocations
-            }
-        } else {
-            if (direction == "INBOUND") {
-                return locations.findAll {
-                    it.locationType.locationTypeCode == LocationTypeCode.SUPPLIER || !it.supports(ActivityCode.MANAGE_INVENTORY)
-                }
+        if (direction == StockMovementDirection.INBOUND.name()) {
+            return locations.findAll {
+                it.locationType.locationTypeCode == LocationTypeCode.SUPPLIER
```
EWaterman, line 154: what's the effect of removing the `!it.supports(ActivityCode.MANAGE_INVENTORY)` check for superuser inbound movements? Is this what was including wards? Are there other types of locations that would also be filtered out now because of this?

drodzewicz: Yes, the lack of `MANAGE_INVENTORY` activity means it's a "Ward".
I asked Katarzyna if returning only supplier location on inbound is what we expect and she has confirmed that we do not  expect any other location there. This whole if statement that I have removed was solely for returning ward locations for superuser, since the ticket specifies that we don't want to do that anymore I have removed it so superuser and regular user receives the same result.

EWaterman: As I mentioned in the standup, we probably want to understand the reason why that line was specifically added in before we remove it. Looks like it that whole block was added as a part of this bug fix: https://pihemr.atlassian.net/browse/OBPIH-2764.

Are you able to follow the steps there to make sure removing the line doesn't reproduce the issue? From the ticket the steps are:

> reproduction steps:
> 1. Create Inbound stock movement, set as origin a depot with manage inventory permission
> 2. Add some items and try to go to next page.
> 
> expected results:
> 
> User should be able to go to edit page and then to all other stock movement pages.

I'm going to approve the change but I think it would be best to verify that the change isn't reintroducing a bug.

EWaterman (approved):

kchelstowski (approved):

PR #4750, opened 2024-07-17
OBPIH-6577 Return non template document types in add document views

drodzewicz:


src/main/groovy/org/pih/warehouse/core/DocumentCode.groovy
```diff
@@ -25,7 +25,6 @@ enum DocumentCode {
 
     static list() {
         [
-                SHIPPING_TEMPLATE,
```
kchelstowski, line 28: why has it been removed from there? isn't it used anywhere else?

alannadolny (approved):

src/main/groovy/org/pih/warehouse/core/DocumentCode.groovy
drodzewicz: there were two `SHIPPING_TEMPLATE` in the `list()` figured that was a mistake and removed one of them

drodzewicz:
> This change only affects the dropdown where you select the document type right? Not the actual upload itself?
@EWaterman I have added an additional validation for template types on these document uploads.
All of them use the same endpoint which is kind of "Generic" so I only had to add this validation in one place.

P.S. I kind of hate this endpoint and would love for it to be refactored some day, but since it works I guess we don't care about it at  the moment :smile: so don;t be too harsh on the way I did this validation.

src/main/groovy/org/pih/warehouse/core/DocumentCode.groovy
EWaterman: normally I'd say we shouldn't do this change since there's no direct bug associated with it and we're already mostly through testing but I don't actually see this `list()` call being used anywhere so likely it's fine.

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
```diff
@@ -207,10 +210,17 @@ class DocumentController {
...
                     documentNumber: command.documentNumber,
-                    documentType: DocumentType.get(typeId))
+                    documentType: documentType)
+
+            documentInstance.validate()
+
+            List<DocumentType> nonTemplateDocumentTypes = documentService.getNonTemplateDocumentTypes()
+            if (!nonTemplateDocumentTypes.contains(documentType)) {
+                documentInstance.errors.reject("documentType", "Template types are not allowed for this document upload")
+            }
```
EWaterman, line 220: You're right I'm not a fan of having to do this :laughing: I'm assuming you're doing it to avoid needing more code change below where it's checking the domain validator errors. Pretty hacky hehe.

I'd think a more proper solution would be to add a custom `DocumentValidator` that wraps the existing validator on the domain and then also performs more complex checks like this one. Then we could just call `documentValidator.validate()` and it'll all be handled by a single call. That's not the simplest of changes though so if you don't want to do that right now can you add a todo here asking for a future refactor?

One thing to check. In the ticket it says "users should not be able to create new docs of this type. That should be limited to admins (prob superuser)." How are admins adding template documents currently? Is it via this API? If so, this change will prevent them from doing so

drodzewicz: I believe none of the users should be able to add template type of documents in these places (invoice, stockMovement, PO, etc... ), we have a dedicated page for adding all sorts of documents that each user can access so we are not limiting anything.
I believe the goal for this ticket was to prevent users from adding template documents on individual invoices, stock movements etc.. because they are accessible globally on other invoices and stockmovement and not limited to the one where it was uploaded.

EWaterman: yes agreed, I just want to confirm that the general page for adding documents doesn't also use this flow. Can you verify that with your change you can still navigate to the /openboxes/document/list page and upload a template document?

drodzewicz: The add document pages/subpages that I implemented a change for use the 
`DocumentController` > `saveDocument` or `updateDocument`
while the document/list uses the
`DocumentController` > `save` or `update`
which I have not touched

EWaterman: that's not confusing at all :laughing: thanks for confirming

EWaterman (approved):

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
```diff
@@ -207,10 +210,17 @@ class DocumentController {
                     contentType: file.contentType,
                     extension: FileUtil.getExtension(file.originalFilename),
                     documentNumber: command.documentNumber,
-                    documentType: DocumentType.get(typeId))
+                    documentType: documentType)
+
+            documentInstance.validate()
+
+            List<DocumentType> nonTemplateDocumentTypes = documentService.getNonTemplateDocumentTypes()
+            if (!nonTemplateDocumentTypes.contains(documentType)) {
```
kchelstowski, line 218: this could be handled in the `validator` of the `Document`, but since we are tight on time and the guys approved, I will merge it as is.

kchelstowski (approved):

PR #4751, opened 2024-07-17
OBPIH-6581 Fix redirect when attaching a document to a stock movement

kchelstowski:


grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
```diff
@@ -189,6 +193,10 @@ class DocumentController {
         def requestInstance = Requisition.get(command.requestId)
         def productInstance = Product.get(command.productId)
         def invoiceInstance = Invoice.get(command.invoiceId)
+        def stockMovement = outboundStockMovementService.getStockMovement(command.stockMovementId)
```
alannadolny, line 196: Why def? :cry:

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
```diff
@@ -257,6 +265,10 @@ class DocumentController {
         } else {
             log.info "Document is too large"
             flash.message = "${warehouse.message(code: 'document.documentTooLarge.message')}"
+            if (stockMovement) {
+                redirect(controller: 'stockMovement', action: 'show', id: stockMovement.id)
+                return
+            }
             if (shipmentInstance) {
                 redirect(controller: 'stockMovement', action: 'show', id: command.shipmentId)
                 return
```
alannadolny, line 274: So if you are working on it, can you remove those disgusting elses when there are returns already?

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
```diff
@@ -578,8 +594,8 @@ class DocumentController {
         Map model = [document: document, inventoryItem: inventoryItem, location: location]
         String body = templateService.renderTemplate(document, model)
 
-        response.contentType = 'image/png'    
-        // TODO Move labelary URL to application.yml 
+        response.contentType = 'image/png'
```
alannadolny, line 597: can't it be combined somehow in the response builder below?
I mean here:
```
        response.outputStream << Request.Post('http://api.labelary.com/v1/printers/8dpmm/labels/4x6/0/')
            .bodyString(body, ContentType.APPLICATION_FORM_URLENCODED)
            .execute()
```

alannadolny (approved):
lgtm, but I have a few questions

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
kchelstowski: because it can be either `OutboundStockMovement` or `StockMovement`

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
kchelstowski: honestly I would prefer not to refactor this causing a headache code right now, just before the release, as this is a generic method that I would have to test for many workflows (SMs, invoices, etc.).

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
kchelstowski: I have not changed this part of the code. This is just IntelliJ that did some formatting changes (trailing spaces or something like that)

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
EWaterman: essentially all you're checking here is that the provided `command.stockMovementId` is actually associated with a real stock movement right? All subsequent checks could just use `command.stockMovementId` if we wanted.

I get why you did it this way but I'm not the hugest fan of having to rope stock movement specific checks into the generic DocumentController. I know that's more a consequence of how this is all designed in the first place though so I don't expect you to change it. Maybe this could be refactored some day to have the command provide a single field that tells us what redirect it wants.

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
EWaterman: I agree. Like my previous comment says, this class definitely could use a refactor but now is probably not the best time.

EWaterman (approved):

grails-app/controllers/org/pih/warehouse/core/DocumentController.groovy
kchelstowski: @EWaterman generally, we have an access to `stockMovementInstance` from the document view model, but the problem is that if we want to eventually send it in the `uploadDocument` form as the `<g:hiddenField>`, it expects us to set a "raw" value (string, integer, boolean), so I could only set the id (string), so essentially we do the second call for the SM in this flow :man_facepalming: 
I don't like it as well, but yeah... this is GSP.....

PR #4754, opened 2024-07-18
OBPIH-6562 Localize dates in list pages

alannadolny:


kchelstowski (approved):

grails-app/i18n/messages_es_MX.properties
```diff
@@ -5,3 +5,4 @@ default.number.format=\#\#\#,\#\#\#,\#\#0.00
 react.default.defaultDate.format=DD/MMM/YYYY HH:mm:ss Z
 react.default.commonDate.format=DD/MMM/YYYY
 react.default.expiryDate.format=MMM/YYYY
+react.default.displayDate.format=DD/MMM/YYYY
```
EWaterman, line 8: sorry I don't remember where the conversation on this ended up previously. Why do we want a different display format for Mexico specifically?

alannadolny: In this case, I won't break the current date format for every other locale on the list pages. Previously we didn't consider changing formats on the list pages, so this format is a counterpart of the format for Mexico. Previously we were mainly discussing the grails side of things, the discussion about naming was about naming design (purpose vs description of format)

PR #4755, opened 2024-07-19
[OBPIH-6585] Facts table seems not to build on staging

EWaterman:
https://pihemr.atlassian.net/browse/OBPIH-6585

`LocationType.locationTypeCode` is currently a nullable field, but `LocationDimension.locationTypeCode` is a non-null field. If we ever have a location type with a null location type code, it causes column constraint failures when generating location dimensions for all locations and causes the whole action to fail.

Screenshot from 2024-07-17 11-18-21

This change is to make `LocationType.locationTypeCode` non-nullable.

### DIFF
--- a/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
+++ b/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
@@ -11,7 +11,7 @@ databaseChangeLog = {
         // bad records instead, but this is the safer option in case those location types are accidentally being used.
         update(tableName: "location_type") {
             column(name: "location_type_code", value: "DEPOT")
-            where("location_type IS NULL")
+            where("location_type_code IS NULL")
         }
 
         addNotNullConstraint(columnDataType: "VARCHAR(100)", columnName: "location_type_code", tableName: "location_type")
--- a/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
+++ b/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
@@ -1,12 +1,17 @@
 databaseChangeLog = {
 
-    changeSet(author: "ewaterman", id: "1721321391197-94", objectQuotingStrategy: "LEGACY") {
-        preConditions(onError: "HALT", onFail: "MARK_RAN", onSqlOutput: "IGNORE") {
-            and {
-                columnExists(columnName: "location_type_code", tableName: "location_type")
+    changeSet(author: "ewaterman", id: "180720241200-0", objectQuotingStrategy: "LEGACY") {
 
-                sqlCheck("""SELECT COUNT(*) FROM location_type WHERE location_type_code IS NULL""", expectedResult: "0")
-            }
+        preConditions(onError: "HALT", onFail: "MARK_RAN", onSqlOutput: "IGNORE") {
+            columnExists(columnName: "location_type_code", tableName: "location_type")
+        }
+
+        // It is invalid behaviour for a location type to not have a code, so we shouldn't have any in production, but
+        // we assign any existing ones an arbitrary code (in this case DEPOT) just in case. We could try to delete the
+        // bad records instead, but this is the safer option in case those location types are accidentally being used.
+        update(tableName: "location_type") {
+            column(name: "location_type_code", value: "DEPOT")
+            where("location_type IS NULL")
         }
 
         addNotNullConstraint(columnDataType: "VARCHAR(100)", columnName: "location_type_code", tableName: "location_type")
--- a/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
+++ b/grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
@@ -1,0 +1,14 @@
+databaseChangeLog = {
+
+    changeSet(author: "ewaterman", id: "1721321391197-94", objectQuotingStrategy: "LEGACY") {
+        preConditions(onError: "HALT", onFail: "MARK_RAN", onSqlOutput: "IGNORE") {
+            and {
+                columnExists(columnName: "location_type_code", tableName: "location_type")
+
+                sqlCheck("""SELECT COUNT(*) FROM location_type WHERE location_type_code IS NULL""", expectedResult: "0")
+            }
+        }
+
+        addNotNullConstraint(columnDataType: "VARCHAR(100)", columnName: "location_type_code", tableName: "location_type")
+    }
+}
--- a/grails-app/views/locationType/create.gsp
+++ b/grails-app/views/locationType/create.gsp
@@ -41,7 +41,6 @@
                             </td>
                             <td valign="top" class="value ${hasErrors(bean: locationTypeInstance, field: 'description', 'errors')}">
                                 <g:select name="locationTypeCode" class="chzn-select-deselect" from="${org.pih.warehouse.core.LocationTypeCode.list()}"
-                                          noSelection="['':'']"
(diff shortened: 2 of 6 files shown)


kchelstowski (commented):
lgtm, but feels like one of the tests failed:
Screenshot from 2024-07-19 09-58-41

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
```diff
@@ -0,0 +1,14 @@
+databaseChangeLog = {
+
+    changeSet(author: "ewaterman", id: "1721321391197-94", objectQuotingStrategy: "LEGACY") {
```
alannadolny, line 3: basically, the ID of the changeset should correspond to the date in the same way you did in the changelog name.
So, as an example, the file name of one of the changelog files:
image
it's 2024-02-22-1300
and the changesets within it are:
```
<changeSet author="kchelstowski" id="220220241300-0">
<changeSet author="kchelstowski" id="220220241300-1">
```
220220241300-0 -> 22-02-2024-1300 and the "0" is a number of the changeset, so the next one has 1, and the next should be 2, and so on.

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
```diff
@@ -0,0 +1,14 @@
...
+        preConditions(onError: "HALT", onFail: "MARK_RAN", onSqlOutput: "IGNORE") {
+            and {
+                columnExists(columnName: "location_type_code", tableName: "location_type")
+
+                sqlCheck("""SELECT COUNT(*) FROM location_type WHERE location_type_code IS NULL""", expectedResult: "0")
```
alannadolny, line 8: I think there should be just one quote instead of three.

grails-app/views/locationType/create.gsp
```diff
@@ -41,7 +41,6 @@
                             </td>
                             <td valign="top" class="value ${hasErrors(bean: locationTypeInstance, field: 'description', 'errors')}">
                                 <g:select name="locationTypeCode" class="chzn-select-deselect" from="${org.pih.warehouse.core.LocationTypeCode.list()}"
-                                          noSelection="['':'']"
```
alannadolny, line 44: What is the default value at this moment? Doesn't it throw any error?

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
alannadolny, line 8: What will happen when there are already location types with code null? Shouldn't it be also fixed?

EWaterman: yeah I was thinking about whether or not we want to update all rows that have a code null to have some default code. There aren't any locations like this on obnav, obnavstage, or obdev3 so this will run fine there, but it might still be best to provide a default just in case. The reason I didn't was because I didn't want to silently change behaviour somewhere, but since we're considering location types without a code to be a bug, maybe it's fine. It just felt a little arbitrary picking one at random. Thoughts?

grails-app/views/locationType/create.gsp
EWaterman: It defaults to the first item in the list. This is how we do it in party roles (which is where I got the idea from).

image

See how there's no X button for Party so you always have to select something. It can't be empty.

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
EWaterman: agreed, I originally had this on multiple lines (hence the triple quotes) but then condensed it to one so the triple quote isn't needed anymore

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
EWaterman: why DDMMYYYYTTTT and not YYYMMDDTTTT like we have in the name? Anyways, I'll fix it like you said.

drodzewicz: tradition

src/test/groovy/unit/org/pih/warehouse/core/LocationControllerTests.groovy
EWaterman, line 1: I decided to just refactor this whole test class while I was here since it's really old.

I still need to figure out how to properly stub Grails' `PagedResultList` objects (it's really annoying they don't give you an easy way to do this) but I'm going to leave that as a problem for another day.

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
EWaterman: Just to update this thread, I changed the script to update any existing location types that don't have a code to use DEPOT. Again, this shouldn't affect any real data (there's no cases of this in prod) but doing it this way makes the script run without errors on any dev environments that happen to have bad data in them

jmiranda:
I haven't had a chance to review this yet but the first thing I thought was this is too big of a change this late in the release process. So please don't merge until we have that discussion. 

In the meantime since this is a data issue we should clean up the data on prod / staging and refresh the fact / dimension tables.

EWaterman:
@jmiranda yeah I agree. It was only a data issue on stage (it's fixed now) so at this point the change can probably be moved to develop because it's not currently an issue on prod and this is just preventative.

grails-app/migrations/0.9.x/changelog-2024-07-18-1200-alter-table-location-type-make-location-type-code-non-null.groovy
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog = {
+
+    changeSet(author: "ewaterman", id: "180720241200-0", objectQuotingStrategy: "LEGACY") {
```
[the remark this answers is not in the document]
jmiranda: Convention: Also, start the changeset ID from 1 instead of 0 (180720241200-1). I usually reserve the 0 for when a changeset needs to be inserted before the others. However, the ordering of the IDs doesn't really matter as the changesets should be executed in the order they appear, so do whatever feels ok and make sure the IDs are unique. I just wanted to point out the convention.

> [!TIP] 
> 
> Creating these unique IDs is a pain so way back in the day I created a bash script that generates a changeset ID using the date command. There's probably a better way to do this, but it's worked for me. 

Here's the script in case anyone wants to use it. I'll add this to the dbm/liquibase documentation ticket as well. 
```
#!/bin/bash
OUTPUT=$(date +%s%N | cut -b1-13)
echo $OUTPUT
echo -n $OUTPUT | xclip -selection c
```
Copy to $HOME/bin.
```
[jmiranda@jmiranda-ThinkPad-W540 ~]$ ls -al /home/jmiranda/bin/generateChangeSetId.sh 
-rwxrwxr-x 1 jmiranda jmiranda 96 Apr 11  2013 /home/jmiranda/bin/generateChangeSetId.sh
```
NOTE: Holy $#!% I wrote this script over a decade ago. Let me know if there's a better (standard) way to do this (i.e. one improvement would probably to create a bash alias).

Execute bash script (make sure HOME/bin is in the PATH)
```
[jmiranda@jmiranda-ThinkPad-W540 ~]$ generateChangeSetId.sh 
1721751106382

```

EWaterman:
Based on Justin's comments I'm going to re-implement this change the proper way using some sort of default enum value for location type code. I'll re-open a new PR against develop so closing this one.

PR #4757, opened 2024-07-19
OBPIH-6581 Fix displayed status in add comment workflow

kchelstowski:


grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -354,8 +354,11 @@ class StockMovementController {
     }
 
     def addComment() {
-        Requisition requisition = Requisition.get(params.id)
-        StockMovement stockMovement = StockMovement.createFromRequisition(requisition)
+        def stockMovement
+        stockMovement = outboundStockMovementService.getStockMovement(params.id)
+        if (!stockMovement) {
+            stockMovementService.getStockMovement(params.id)
+        }
```
alannadolny, line 361: Basically, it can be shortened to:
```
def stockMovement = outboundStockMovementService.getStockMovement(params.id) ?: stockMovementService.getStockMovement(params.id)
```
You know, that I don't like mutating values, so this idea is a result of my hatred. If you would like you can also shorten it just to:
```
def addComment() {
   [
      stockMovement: outboundStockMovementService.getStockMovement(params.id) ?: stockMovementService.getStockMovement(params.id),
      comment: new Comment()
   ]
}
```
In this case, you don't have to use def :smile:

alannadolny (approved):

drodzewicz (approved):

kchelstowski:
Merging it, as this one is pressing for Kasia. @EWaterman let me know if you have any suggestions that I could eventually address "post-merge".

PR #4758, opened 2024-07-19
OBPIH-6592 Investigate issue with Items with backdated shipments since last stock count indicator

alannadolny:


grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1188,7 +1188,7 @@ class IndicatorDataService {
         String query = '''
             SELECT 
                 product_code,
-                COUNT(product_code) as products,
+                COUNT(DISTINCT backdated_shipment) as products,
```
kchelstowski, line 1191: what's the context of that? Why do we name the number of shipment numbers as "products"?

alannadolny: To be honest I don't know why I named it like that previously, because it's the number of backdated shipments in which the exact product appears.

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
```diff
@@ -1188,7 +1188,7 @@ class IndicatorDataService {
         String query = '''
             SELECT 
                 product_code,
-                COUNT(product_code) as products,
+                COUNT(DISTINCT backdated_shipment) as shipments_with_backdated_product,
```
EWaterman, line 1191: won't changing `products` to `shipments_with_backdated_product` affect whoever is using the result of this query? Where is this query getting parsed?

EWaterman: Looks like line 1267 is where this is accessed and it's by index so the rename is probably fine. Please retest though just in case.

EWaterman (approved):

kchelstowski (approved):

grails-app/services/org/pih/warehouse/dashboard/IndicatorDataService.groovy
alannadolny: works well

PR #4769, opened 2024-07-25
OBPIH-6547 Bump-up unit tests coverage

alannadolny:


src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    void 'issueRequisition should throw an exception when inventory service is unable to save local transfer'() {
+        given:
+        Requisition requisition = new Requisition(id: 1)
+        String comment = "Comment to requisition with errors"
+
+        and:
+        Picklist.metaClass.static.findByRequisition = { Requisition foundRequisition -> return new Picklist(requisition: foundRequisition)
+        }
```
EWaterman, line 81: nitpick: I'd either move the '}' to the above line or move the return statement down to its own line. Same comment for the next test.

It's weird that grails and gorm doesn't really provide a method of stubbing domain methods. This feels weird to have to do but I couldn't find a better solution from googling.

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    @Shared
+    IdentifierService identifierService
+
+    void setupSpec() {
+        mockDomain Requisition
+        mockDomain Picklist
+    }
+
+    void setup() {
+        inventoryService = Spy(InventoryService) {
```
EWaterman, line 36: why Spy? Does it not still work if you make it a Stub? From what I see in the docs, we should generally avoid using Spy if at all possible. https://grails.org/blog/2018-06-22.html

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+
+        when:
+        Requisition copyOfRequisition = service.cloneRequisition(requisition)
+        Map jsonOfOriginalRequisition = requisition.toJson()
+        Map jsonOfCopiedRequisition = copyOfRequisition.toJson()
+
+        then:
+        'Copy of ' + jsonOfOriginalRequisition['name'] == jsonOfCopiedRequisition['name']
+        jsonOfOriginalRequisition.remove('name')
+        jsonOfCopiedRequisition.remove('name')
```
EWaterman, line 156: kinda weird but I can't think of a better solution off the top of my head outside of just asserting on each key of the map.

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -13,88 +16,346 @@ import spock.lang.Unroll
...
+        domain.approveQuantity()
+
+        then:
+        thrown(ValidationException)
+
+        where:
+        quantity | quantityCanceled
+        10       | 11
+        20       | 20
+        21       | 50
```
EWaterman, line 176: what about quantityCancelled null and 0?

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -13,88 +16,346 @@ import spock.lang.Unroll
...
+        Boolean isRequisitionCanceled = requisitionItem.isCanceled()
+
+        then:
+        isCanceled == isRequisitionCanceled
+
+        where:
+        quantityCanceled | quantity | modificationItem      | substitutionItem      | requisitionItems        || isCanceled
+        3                | 3        | null                  | null                  | null                    || true
+        3                | 2        | null                  | null                  | null                    || false
+        3                | 2        | Mock(RequisitionItem) | null                  | null                    || false
```
EWaterman, line 301: shouldn't quantity for these lines be 3? Otherwise they wouldn't be cancelled regardless.

EWaterman: or I guess a better question is do we also want to add lines where quantity is 3 for each of these combos and assert that isCanceled is true

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -13,88 +16,346 @@ import spock.lang.Unroll
...
+    void "RequisitionItem.quantityNotCanceled() should calculate #quantityNotCanceled when #quantity and #quantityCanceled are passed"() {
+        given:
+        domain.quantity = quantity
+        domain.quantityCanceled = quantityCanceled
+
+        when:
+        Integer calculatedQuantityNotCanceled = domain.quantityNotCanceled()
+
+        then:
+        calculatedQuantityNotCanceled == quantityNotCanceled
```
EWaterman, line 188: When you have simple one line asserts like this you can do:
```
expect:
domain.quantityNotCanceled() == quantityNotCanceled
```

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -13,88 +16,346 @@ import spock.lang.Unroll
...
+
+        when:
+        Boolean isItemSubstituted = domain.isSubstituted()
+
+        then:
+        isSubstituted == isItemSubstituted
+
+        where:
+        quantityCanceled | substitutionItem      || isSubstituted
+        0                | null                  || false
```
EWaterman, line 343: probably also worth having a line where quantityCancelled is > 0 but isSubstituted is still false if there are no substitutionItem

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
```diff
@@ -13,88 +16,346 @@ import spock.lang.Unroll
...
+
+        then:
+        isCanceledDuringPick == isCanceled
+
+        where:
+        status                      | modificationItemQuantityPicked || isCanceled
+        RequisitionStatus.PICKING   | 0                              || true
+        RequisitionStatus.ISSUED    | 0                              || true
+        RequisitionStatus.VERIFYING | null                           || false
+        RequisitionStatus.PICKING   | null                           || false
```
EWaterman, line 326: could be nice to add some test cases around the modificationItem being set since I see that's used in the isCanceledDuringPick() method. Maybe something similar to what you did for the last test you have in this file. Up to you though.

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
alannadolny: > nitpick: I'd either move the '}' to the above line or move the return statement down to its own line. Same comment for the next test.

I did it like that before and the run autoformatting in Intellij :smile:

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
alannadolny: This line: 
```
1 * inventoryService.generateTransactionNumber()
```
doesn't work when I change it to Stub, but I think this one check doesn't make too much sense

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionItemSpec.groovy
alannadolny: Comparing null with value doesn't throw an exception, I can add this case in the test above

src/test/groovy/unit/org/pih/warehouse/requisition/RequisitionServiceSpec.groovy
jmiranda: > It's weird that grails and gorm doesn't really provide a method of stubbing domain methods.

There are. We just need to do some more work on testing to start using them. I'm hoping a test-a-thon exercise will help us uncover more testing best practices.

kchelstowski (approved):

PR #4771, opened 2024-07-29
OBPIH-6197 Update documentation for official 0.9.x release 

jmiranda:
Docs are currently published here
https://docs.openboxes.com/en/obpih-6197/


docs/admin-guide/migration/in-place/index.md
```diff
@@ -0,0 +1,186 @@
...
+=== "Tomcat"
+    
+    This documentation does not officially support enabling HTTPS/SSL on Tomcat, but there's 
+    nothing preventing you from using this feature.  
+
+## Upgrade Application
+
+### Backup Database 
+
+### Configuration
```
awalkowiak, line 173: In 0.8.x we were using both `openboxes-config.properties` and `openboxes-config.groovy` and for 0.9.x we switched to having only `openboxes-config.groovy` on PIH instances of OB, but since both still should work as before I am not sure if there is anything that needs to be mentioned here about that.

awalkowiak, line 173: Perhaps adding something about JDBC driver settings and dataSource dialect here would be useful, since you mentioned above that for 0.9.x the supported dbs are mysql8 and mariadb

docs/admin-guide/migration/parallel/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### Install Dependencies
+
+Use the latest Installation Guide to install dependencies.
+
+#### Configure Server(s)
+
+Use the latest Configuration Guide to configure your server(s).
+
+### Migrate Database
```
EWaterman, line 48: what about any data that was inserted since the backup was taken? Not much point to making it rolling deployable if that can't be saved. Maybe I'm misunderstanding though

EWaterman: we talked about this over slack. I was assuming this "parallel" option meant rolling deploy / no downtime deploy. Really it just means installing a new instance in parallel then migrating over when you're ready.

EWaterman: I think what's throwing things off for me is the "Switch Over" step is glossing over quite a lot. I think what you're suggesting is taking two backups: One at the start of the process that we can upload to the new 0.9.x instance whenever we're ready to test the instance, and then a second backup while actually switching over to the new instance while the old instance is down to capture any new data that was added since then.

So we'd want the "Disable Production Environment" step to be before we do the second backup (not sure if you're still writing all this out).

docs/admin-guide/migration/index.md
```diff
@@ -0,0 +1,57 @@
...
+properly, and you're done.
+
+OpenBoxes 0.8.x releases were tied to the following tech stack. This upcoming 0.9.x release(s) support a bit more variability.
+
+| Dependency         | 0.8.x Supported Versions | 0.9.x Supported Versions   |
+|:-------------------|:-------------------------|----------------------------|
+| Operating System   | Ubuntu 18.04             | Ubuntu 22.04               |
+| Java               | Java 7                   | Java 8                     |
+| Application Server | Tomcat 7                 | Tomcat 8.5, **Tomcat 9**   |
+| Database           | MySQL 5.7                | **MySQL 8**, MariaDB 10.11 |
```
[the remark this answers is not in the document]
jmiranda: yeah that's a good point. at some point in the docs i think i said that MySQL 5.7 should work but we don't officially support it. if that statement is not in the docs, we should make that explicit.

docs/admin-guide/migration/in-place/index.md
```diff
@@ -0,0 +1,186 @@
...
+    #### Install nginx apt
+    
+        sudo apt-get install nginx -y
+    
+    #### Configure SSL
+    
+        sudo touch /etc/nginx/conf.d/default.conf
+        cat 9_nginx_config.template > /etc/nginx/conf.d/default.conf
+        echo "Bruteforce creating missing files in the /etc/letsencrypt"
+        curl https://raw.githubusercontent.com/certbot/certbot/master/certbot-nginx/certbot_nginx/_internal/tls_configs/options-ssl-nginx.config > /etc/letsencrypt/options-ssl-nginx.conf
```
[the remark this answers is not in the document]
jmiranda: fixed

docs/admin-guide/migration/in-place/index.md
```diff
@@ -0,0 +1,186 @@
...
+    
+            Include configuration changes for mod_jk or mod_http or whatever. 
+
+=== "nginx"
+    
+    #### Install nginx apt
+    
+        sudo apt-get install nginx -y
+    
+    #### Configure SSL
```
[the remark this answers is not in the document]
jmiranda: if you can figure out how to work that in, then great. if you'd prefer to remove the "in place" migration instructions because there are a lot of gotchas, then i'd be ok with that. and i've been thinking similarly since starting down this path.

docs/admin-guide/migration/known-issues/index.md
```diff
@@ -0,0 +1,5 @@
+# Known Issues
+
+* GitHub > MySQL 8 issues 
+* https://pihemr.atlassian.net/browse/OBS-1609
```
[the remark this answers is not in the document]
jmiranda: yeah these links were more of a placeholder to remind me to pull out information from the ticket. but i'm just going to comment them out for now.

docs/admin-guide/migration/in-place/index.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Configuration
+
+### Swap Tomcat
+
+## Validation 
+
+* Check logs to make sure application has started that all database migrations have been executed without errors.
+* Check the application 
+* Check Sentry (if applicable)
```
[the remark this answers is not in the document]
jmiranda: yup, that was a section i wanted to include in the configuration documentation. however, i probably won't have time before going on vacation so i'll work on that when i get back.

docs/admin-guide/migration/in-place/index.md
jmiranda: i am advocating for openboxes.yml for new installs (https://docs.openboxes.com/en/obpih-6197/admin-guide/installation/ubuntu2204/configuration/), but there's nothing explicitly stated for parallel and in-place migrations. for the migrations, we should probably state that config should be copied as-is onto the new server / new directory structure.

docs/admin-guide/migration/parallel/index.md
jmiranda: to the point about "parallel" vs "in-place" migration, i am using those terms because those were the most common. however i still don't like "parallel" as it causes confusion. my first attempt at this was "clean slate" vs "in place", which i don't like because "clean slate" carries a conotation that is not true (i.e. we're starting fresh, including the database). but now i'm also seeing that "side-by-side" is also used to describe "parallel" migrations, so i might use that breakdown instead.

it's after struggling for

jmiranda: regarding the "switch over" part, i'll just say that in general there's a lot missing so i might need some help structuring the missing buckets and maybe even writing some of these sections because my brain is mush after working on this for so long.

docs/admin-guide/migration/in-place/index.md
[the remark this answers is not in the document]
awalkowiak: @jmiranda I figured out that the missing part was to install the proper certbot plugin for nginx:
```
sudo apt-get install python3-certbot-nginx
```

awalkowiak (approved):

PR #4772, opened 2024-07-29
OBPIH-6497 [backend] Determine encumbered and invoiceable items or adjustments

alannadolny:


grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -108,6 +110,14 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
...
     }
 
+    Boolean getIsInvoiceable() {
+        return !isInvoiced && order.isPlaced()
+    }
+
+    Boolean getIsEncumbered() {
+        return !isInvoiced && hasPrepaymentInvoice
+    }
+
```
kchelstowski, line 120: I know we've done that in some places, but I don't like the naming like `getIs`.
Booleans should be named as `isSomething`.
Moreover, those don't have to be marked transients anymore, so you could just have a method `isInvoicable()`.

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -283,10 +287,28 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
...
+        return !isCompletelyInvoiced()
+    }
+
     Boolean isPending() {
         return !isCompletelyFulfilled()
     }
 
+    Boolean getHasNotInvoicedAdjustment() {
+        return orderAdjustments.any { !it.isInvoiced }
+    }
```
kchelstowski, line 310: same as above

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
alannadolny: Yeah, but the naming convention with get, allows us to use the `.getIsEncumbered()` in this way: `.isEncumbered`, so I prefer this groovy syntactic sugar

kchelstowski: I don't see anything bad in using `.isEncumbered()` vs forcing to access a property via the field name (`isEncumbered`). I feel like naming booleans using `getIs`, `getHas` is an anti-pattern we should avoid.
https://rules.sonarsource.com/java/tag/convention/RSPEC-2047/

alannadolny: This is groovy, not Java :smile: 
The rule that you send is another case. They say that we should use `hasBar`, not `getBar`

kchelstowski: yeah, and why should we use `getBar` for a boolean field?

alannadolny: not `getBar`, but `getIsBar`, and then we can use it `isBar`

kchelstowski: I will repeat it again, and you will do whatever you want - `getIs` feels like antipattern, and I don't see anything wrong in using `isBar()` in this case.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -456,8 +456,24 @@ class Order implements Serializable {
         return !hasInvoice && isPrepaymentRequired && hasActiveItemsOrAdjustments
     }
 
+    /*
```
EWaterman, line 459: I think you need javadocs to start with `/**` (two asterisks) to have them work correctly in IDEs and other tools. Good job with the documentation here though!

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -456,8 +456,24 @@ class Order implements Serializable {
...
+    */
     Boolean getCanGenerateInvoice() {
-        return hasPrepaymentInvoice && isShipped() && !hasRegularInvoice
+        // If prepayment invoice doesn't exist we can't generate invoice
+        if (!hasPrepaymentInvoice) {
+            return false
+        }
+
+        return orderItems.any {
+            it.hasNotInvoicedAdjustment || // condition related to first point
```
awalkowiak, line 473: What if the order adjustment is not tied to any order item? This is possible too

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -108,6 +110,14 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
         return invoiceItems*.invoice.unique()
     }
 
+    Boolean getIsInvoiceable() {
+        return !isInvoiced && order.isPlaced()
+    }
+
+    Boolean getIsEncumbered() {
+        return !isInvoiced && hasPrepaymentInvoice
```
awalkowiak, line 118: Should this be also `order.isPlaced()` to be considered encumbered or not?

awalkowiak (changes requested):
First comment is a request change, second one is just a question

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+            getHasPrepaymentInvoice() >> hasPrepaymentInvoice
+        }
+        OrderItem orderItem = Spy(OrderItem) {
+            getIsEncumbered() >> isEncumbered
+            getIsInvoiceable() >> isInvoiceable
+        }
+        Set<OrderItem> orderItems = [
+                Mock(OrderItem),
+                orderItem,
+                Mock(OrderItem)
```
EWaterman, line 74: what do these mock order items give us?

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+        order.setOrderItems(orderItems)
+
+        expect:
+        order.canGenerateInvoice == canGenerateInvoice
+
+        where:
+        hasPrepaymentInvoice | orderItemStatusCode          | isEncumbered || canGenerateInvoice
+        false                | OrderItemStatusCode.CANCELED | false        || false
+        true                 | OrderItemStatusCode.CANCELED | false        || false
+        true                 | OrderItemStatusCode.PENDING  | true         || false
```
EWaterman, line 58: what about testing the scenario where `Order.stats == PLACED`? then you would be able to generate an invoice right?

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
EWaterman: I'm used to seeing fields like:
```
invoiceable
encumbered
```
with getters and setters like
```
isInvoiceable
setInvoiceable
isEncumbered
setEncumbered
```
From the groovy docs: https://groovy-lang.org/objectorientation.html#_property_naming_conventions

> ...generated getters and setters will have a name formed by capitalizing the property name and adding a get or set prefix (or optionally "is" for a boolean getter)... isEmpty might be the getter method name for a property named empty.

So with that setup in Groovy we could do `boolean isOrderInvoiceable = order.isInvoiceable()` as Kacper says. That's my personal preference but if you disagree Alan feel free to do as you wish.

jmiranda: Java beans use the same convention. 

So since we keep having this conversation 

https://github.com/openboxes/openboxes/pull/4266#pullrequestreview-1627749462

... let's just use the java/groovy convention from now one. 

So no more "is" prefix in the property name. 

For me, I prefer using Groovy's implicit getter method
```
if (order.invoiceable)
```
over the explicit getter methods
```
if(order.getInvoiceable())
```
or 
```
if(order.isInvoiceable())
```

kchelstowski: The problem with `getIs` is that the property itself would be named with the `is` prefix, e.g. `isInvoiceable` instead of `invoiceable`. 
https://stackoverflow.com/a/5322730

> Your field should not be called isCurrent in the first place. Is is a verb and verbs are inappropriate to represent an Object's state. Use an adjective instead, and suddenly your getter / setter names will make more sense:

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
alannadolny: I added those mocked order items, because we have to have at least one order item fulfilling the tested case, so I am testing if one item of those three matches the tested condition.

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
alannadolny: Order status cannot be placed. The available options are:
```
[PENDING, COMPLETED, CANCELED, BACKORDER]
```

kchelstowski (approved):

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
EWaterman: I guess my question was more how does the test treat these mocks? Do they have quantity 0?

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
EWaterman: that's OrderItemStatusCode right? I meant OrderStatus

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
```diff
@@ -0,0 +1,107 @@
...
+        ]
+        order.setOrderAdjustments(orderAdjustments)
+
+        expect:
+        order.canGenerateInvoice == canGenerateInvoice
+
+        where:
+        hasPrepaymentInvoice || canGenerateInvoice
+        false                || false
+        true                 || false
```
EWaterman, line 27: shouldn't this condition return true? You have an un-invoiced adjustment that has no associated order items

alannadolny: that's interesting, why this test doesn't fail

drodzewicz (approved):

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
alannadolny: Oh, got the reason, I didn't take order status into account (orderAdjustment.isInvoiceable takes into account the fact if order is placed)

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -456,8 +457,37 @@ class Order implements Serializable {
         return !hasInvoice && isPrepaymentRequired && hasActiveItemsOrAdjustments
     }
 
+    /**
+        Should return true if at least one PO item is invoiceable.
+        A PO item is invoiceable if the order has a prepayment invoice and at least one of those requirements are fulfilled:
+            1. There is an order adjustment on that item that hasn’t already been invoiced
```
awalkowiak, line 463: I think the first point is misleading. 

> What is an invoiceable item?
a shipped item or a placed adjustment; any placed adjustment is pulled as invoiceable on the partial final invoice generated (but it can be removed or edited) or a canceled line

**Order adjustment** should be invoiceable if the order has a prepayment invoice and is placed.
**Order item** should be invoiceable if it has prepayment invoice and:
1. is partially shipped (only this partial quantity not yet invoiced is invoiceable)
2. items canceled that are not yet invoiced

Imho there is no need to check in order item is invoiceable if it has `NotInvoicedAdjustment`, I think the order adjustments should be separated from order items and checked as it is written above.

(to be confirmed)

awalkowiak (approved):

PR #4774, opened 2024-07-31
OBPIH-6498 Show confirmation dialog when trying to generate a partial invoice

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6498](https://pihemr.atlassian.net/browse/OBPIH-6498)

**Description:**
Add a confirmation popup when users is trying to generate a partial invoice.

---

video: https://jam.dev/c/5b9137ba-ff96-4506-b86c-5fe6281ef3ac
image

[OBPIH-6498]: https://pihemr.atlassian.net/browse/OBPIH-6498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -518,6 +518,10 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Boolean isFullyInvoiceable() {
+        return orderItems.every{ it -> it.invoiceable }
+    }
+
     // isInbound is temporary distinction between outbound and inbound used only for Outbound and Inbound Returns
     Boolean isInbound(Location currentLocation) {
         return isReturnOrder && destination == currentLocation && origin != currentLocation
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -519,7 +519,7 @@ class Order implements Serializable {
     }
 
     Boolean isFullyInvoiceable() {
-        Set encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
+        Set<OrderItem> encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
         return encumberedOrderItems && encumberedOrderItems.every{ it -> it.invoiceable }
     }
 
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -518,9 +518,26 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Set<OrderItem> getFullyInvoiceableOrderItems() {
+        return orderItems.findAll{ it.invoiceable && it.quantityRemaining == 0 }
+    }
+
+    Set<OrderAdjustment> getFullyInvoiceableAdjustments() {
+        return orderAdjustments.findAll{ it.invoiceable }
+    }
+
     Boolean isFullyInvoiceable() {
-        Set<OrderItem> encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
-        return encumberedOrderItems && !encumberedOrderItems.any{ it -> !it.invoiceable }
+        Boolean areItemsFullyInvoiceable = false
+
+        if (orderItems) {
+            areItemsFullyInvoiceable = orderItems.size() == fullyInvoiceableOrderItems.size()
+        }
+
+        if (orderAdjustments) {
+            areItemsFullyInvoiceable = orderAdjustments.size() == fullyInvoiceableAdjustments.size()
+        }
+
+        return areItemsFullyInvoiceable
     }
 
     // isInbound is temporary distinction between outbound and inbound used only for Outbound and Inbound Returns
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -518,26 +518,16 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
-    Set<OrderItem> getFullyInvoiceableOrderItems() {
-        return orderItems.findAll{ it.invoiceable && it.quantityRemaining == 0 }
-    }
-
-    Set<OrderAdjustment> getFullyInvoiceableAdjustments() {
-        return orderAdjustments.findAll{ it.invoiceable }
-    }
-
     Boolean isFullyInvoiceable() {
-        Boolean areItemsFullyInvoiceable = false
-
-        if (orderItems) {
-            areItemsFullyInvoiceable = orderItems.size() == fullyInvoiceableOrderItems.size()
+        if (!orderItems && !orderAdjustments) {
+            return false
         }
 
-        if (orderAdjustments) {
-            areItemsFullyInvoiceable = orderAdjustments.size() == fullyInvoiceableAdjustments.size()
-        }
+        Boolean areAllOrderItemsInvoiceable = !orderItems ?: orderItems.every{ it.invoiceable && it.quantityRemaining == 0 }
 
-        return areItemsFullyInvoiceable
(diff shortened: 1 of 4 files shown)


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -518,6 +518,10 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Boolean isFullyInvoiceable() {
+        return orderItems.every{ it -> it.invoiceable }
```
EWaterman, line 522: I believe you want the order item check here to be invoiceable || has already been invoiced.

If we've previously generated an invoice for some of the order items, those will no longer be invoiceable but shouldn't cause the pop-up if the remaining items are invoiceable

grails-app/views/order/_summary.gsp
```diff
@@ -354,6 +352,22 @@
...
         setTimeout(fetchOrderDerivedStatus, ${grailsApplication.config.openboxes.purchaseOrder.derivedStatusFetch.delay});
+
+        $('#generate-invoice').on('click', function() {
+          if (${!hasRoleInvoice}) {
+            alert("${g.message(code:'errors.noPermissions.label')}")
+            return
+          }
+
+          if (${!orderInstance.fullyInvoiceable}) {
+            const dialogmessage = '${warehouse.message(code: 'invoice.partialInvoice.message',default: 'Not all lines on this purchase order are available to invoice. Generate partial invoice?')}';
```
alannadolny, line 363: why not cammel case?

grails-app/domain/org/pih/warehouse/order/Order.groovy
drodzewicz: you are correct, I made some changes that should reflect this behavior.
I decided to check invoicebale against encumbered items which should make sense, right? 

Additionally added a simple unit test to reinforce this method

src/test/groovy/unit/org/pih/warehouse/order/OrderSpec.groovy
```diff
@@ -109,4 +111,29 @@ class OrderSpec extends Specification implements DomainUnitTest<Order> {
...
+        mockOrderItem(true, false)  | mockOrderItem(true, true)     || false
+        mockOrderItem(false, true)  | mockOrderItem(false, true)    || false
+        mockOrderItem(false, false) | mockOrderItem(false, false)   || false
+    }
+
+    private OrderItem mockOrderItem(boolean encumbered, boolean invoiceable) {
+        return Mock(OrderItem) {
+            isEncumbered() >> encumbered
+            isInvoiceable() >> invoiceable
+            toString() >> "encumbered: $encumbered, invoiceable: $invoiceable"
```
EWaterman, line 136: clever overriding the toString() here!

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -518,6 +518,11 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Boolean isFullyInvoiceable() {
+        Set encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
```
kchelstowski, line 522: why not specifing the type of the Set? `Set<OrderItem>`?

kchelstowski:
```suggestion
        Set<OrderItem> encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
```

grails-app/views/order/_summary.gsp
```diff
@@ -354,6 +352,22 @@
...
+
+        $('#generate-invoice').on('click', function() {
+          if (${!hasRoleInvoice}) {
+            alert("${g.message(code:'errors.noPermissions.label')}")
+            return
+          }
+
+          if (${!orderInstance.fullyInvoiceable}) {
+            const dialogMessage = '${warehouse.message(code: 'invoice.partialInvoice.message', default: 'Not all lines on this purchase order are available to invoice. Generate partial invoice?')}';
+            if(!confirm(dialogMessage)) {
```
kchelstowski, line 364:
```suggestion
            if (!confirm(dialogMessage)) {
```

grails-app/domain/org/pih/warehouse/order/Order.groovy
kchelstowski: and one more question - does it parse it to the Set correctly? In general the `.findAll` returns `List<T>`

drodzewicz: The IDE suggestion was that it returned a `Set<OrderItems>`

EWaterman: Any particular reason to not do it in one loop with something like:
`return !orderItems.any{ it -> it.encumbered && !it.invoiceable }`

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -518,6 +518,11 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Boolean isFullyInvoiceable() {
+        Set<OrderItem> encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
+        return encumberedOrderItems && encumberedOrderItems.every{ it -> it.invoiceable }
```
EWaterman, line 523: Any particular reason to not do it in one loop with something like:
`return !orderItems.any{ it -> it.encumbered && !it.invoiceable }`

drodzewicz: the example you have provided fails on cases where none of our items is encumbered.
If none of them are encumbered it returns `false` for `.any{}` and a negation returns `true` for `isFullyInvoiceable` which does not make sense which is why I added an additional check for if there are any encumbered items.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -518,6 +518,11 @@ class Order implements Serializable {
         return orderType?.isTransferOrder()
     }
 
+    Boolean isFullyInvoiceable() {
+        Set<OrderItem> encumberedOrderItems = orderItems.findAll{ it -> it.encumbered }
+        return encumberedOrderItems && !encumberedOrderItems.any{ it -> !it.invoiceable }
```
awalkowiak, line 523: Would not `encumberedOrderItems?.every { it -> it.invoiceable }` be cleaner? Plus this should also look into orderAdjustments, because if you won't have order items, just order adjustments, then you will always get the warning. On top of that `invoiceable` checks if item is invoiceable, not fully invoiceable, as you can see, there is 
`quantityAvailableToInvoice > 0` and in you case you are more interested in if `quantityAvailableToInvoice == quantity remaining to invoice`

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/order/Order.groovy
drodzewicz: good point on the adjustments but as for `return encumberedOrderItems?.every { it -> it.invoiceable }` this will not work since it is not the same as  `return encumberedOrderItems && encumberedOrderItems.every { it -> it.invoiceable }`
The first check for `encumberedOrderItems` is for a case where we do `.every` on an empty list that will return `true` and we don't want that

drodzewicz:
The tests are going to fail ATM I am working on adding some new ones since the logic for checking if Order is fully invoiceable has changed.

But feel free to give you approval or disapproval for the current version of the `isFullyInvoiceable`
@awalkowiak

awalkowiak (approved):

grails-app/domain/org/pih/warehouse/order/Order.groovy
EWaterman: Last thought: What about adjustments? If we have items that are encumbered and non-invoiceable, but we have an invoiceable adjustment, we'd still want the confirmation to pop up, right?

PR #4775, opened 2024-08-02
OBPIH-6527 Full outbound import - validation & save endpoints 

kchelstowski:

> *The PR contains the validation (and import file) endpoints as promised. The creation/save endpoint is still in progress.*

**Link to GitHub issue or Jira ticket: https://pihemr.atlassian.net/browse/OBPIH-6527**

**Description:**

---

> 

---

- [x] Not Applicable

---


grails-app/migrations/0.9.x/changelog-2024-08-02-1200-add-location-name-index.xml
alannadolny, line 1: do we have a final decision to create migrations in XML instead of Groovy? I forgot how this discussion went.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -1025,6 +1025,21 @@ class ProductAvailabilityService {
...
+    Integer getQuantityAvailableToPromiseForProductInBin(InventoryItem inventoryItem, Location location, Location binLocation) {
+        return ProductAvailability.createCriteria().get {
+            projections {
+                property("quantityAvailableToPromise")
+            }
+            eq("inventoryItem", inventoryItem)
+            eq("location", location)
+            if (binLocation) {
+                eq("binLocation", binLocation)
+                return
```
alannadolny, line 1037: Does this return work? :smile:

grails-app/utils/org/pih/warehouse/CommandUtils.groovy
```diff
@@ -21,4 +24,17 @@ class CommandUtils {
...
+     * Util to build and return localized errors grouped by field in the format:
+     * { fieldName1: ["some error"], fieldName2: ["some error 2"] }
+     * @param errors
+     * @return
+     */
+    static Map<String, List<String>> buildErrorsGroupedByField(Errors errors) {
+        ApplicationTagLib g = Holders.grailsApplication.mainContext.getBean(ApplicationTagLib)
+        errors.allErrors
+                .groupBy { FieldError error -> error.field }
+                .collectEntries { [it.key, it.value.collect{ error -> g.message(error: error)}] } as Map<String, List<String>>
```
alannadolny, line 38:
```suggestion
                .collectEntries { [it.key, it.value.collect { error -> g.message(error: error)}] } as Map<String, List<String>>
```

alannadolny: We can also move the `g.message` to the new line.

src/main/groovy/org/pih/warehouse/importer/PackingListExcelImporter.groovy
```diff
@@ -0,0 +1,84 @@
...
+            columnMap: [
+                    A: "palletName",
+                    B: "boxName",
+                    C: "productCode",
+                    D: "product.name",
+                    E: "lotNumber",
+                    F: "expirationDate",
+                    G: "binLocation",
+                    H: "quantityPicked",
+                    I: "recipient"
```
alannadolny, line 27: Is the indentation correct?

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+    Date expirationDate
+
+    @BindUsing({ obj, source ->
+        Location location = Location.findByNameOrDefault(source['binLocation'])
+        // If location is not found, but we provided bin location name, it means, the location was not found,
+        // and we want to indicate that, instead of falling back to default (null)
+        // without this, we would then search e.g. for quantity available to promise for a product in the default bin location
+        if (!location && source['binLocation'] && !source['binLocation'].equalsIgnoreCase(Constants.DEFAULT_BIN_LOCATION_NAME)) {
+            // We want to indicate, that a bin location for given name has not been found.
+            obj.binLocationFound = false
```
alannadolny, line 37: wouldn't it be better to move it to something like `EmptyStringsToNullBinder.groovy`? I would like to avoid adding a lot of code between properties, seeing this I am expecting to see methods under this, not the rest of the properties. But it can be just my opinion.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+        quantityPicked(min: 0, validator: { Integer quantityPicked, ImportPackingListItem item ->
+            if (!item.binLocationFound) {
+                return ['binLocationNotFound']
+            }
+            ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
+            InventoryItem inventoryItem = item.product?.getInventoryItem(item.lotNumber)
+            if (!inventoryItem) {
+                return ['inventoryItemNotFound']
+            }
+            Location currentLocation = AuthService.currentLocation
```
alannadolny, line 86: is there a benefit from assigning it to the variable? The name of the property already suggests that it is a current location.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+
+    @BindUsing({ obj, source -> Product.findByProductCode(source['product']) })
+    Product product
+
+    String lotNumber
+
+    Date expirationDate
+
+    @BindUsing({ obj, source ->
+        Location location = Location.findByNameOrDefault(source['binLocation'])
```
awalkowiak, line 31: You should look for a bin location name under a specific parent. I think this might be the main reason why you got the performance issues. Bin location name is unique under a specific parent but can be duplicated globally, in this version you can't be sure that you found the bin location that you wanted to find.

awalkowiak: Well maybe it is not the main reason of performance issues, but definitely looking by name is not enough in terms what you are trying to find here.

kchelstowski: @awalkowiak yeah, something smelled to me as well. I mean - the performance issues are surely due to the fact that we have not had any index for the name column (because we search for only one location, and searching via id, that has index works fast), but if I understand you correctly, I should look for a location under origin's locations, so:
`origin.getInternalLocation(source['binLocation'])`?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -1025,6 +1025,21 @@ class ProductAvailabilityService {
         return quantityAvailableToPromiseByProductNotInBin ?: 0
     }
 
+    Integer getQuantityAvailableToPromiseForProductInBin(InventoryItem inventoryItem, Location location, Location binLocation) {
+        return ProductAvailability.createCriteria().get {
+            projections {
+                property("quantityAvailableToPromise")
+            }
+            eq("inventoryItem", inventoryItem)
+            eq("location", location)
```
kchelstowski, line 1034: have just realized after @awalkowiak 's comment about binLocation, that I should probably check for the quantity atp for the origin, not the current location?

kchelstowski: in 99% scenarios, the current location would be origin as it is filled in by default, but I assume we want to allow changing the origin by a user?

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
kchelstowski: yeah, since we don't have any additional statements below (except `isNull` which we would not want to be executed if bin location is present), this works fine.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: yeah, this could be moved to a method, for now I've just focused on the functionality, but when I will do the final cleanup, I will take this suggestion into account.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: probably there isn't, it just felt better for me to pass a variable to the method below instead of passing the `AuthService.currentLocation`, but I could agree that for this particular case it probably doesn't make sense to create the variable

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
awalkowiak: Yes if possible please rely on the origin and destination directly

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
awalkowiak: Yes this one would be good

grails-app/migrations/0.9.x/changelog-2024-08-02-1200-add-location-name-index.xml
```diff
@@ -0,0 +1,16 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="kchelstowski" id="020820241200-1">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="location_name_idx"/>
+            </not>
+        </preConditions>
+        <createIndex indexName="location_name_idx" tableName="location" unique="false">
```
awalkowiak, line 12: If you look for bin by parent and name, then this should not be required

src/main/groovy/org/pih/warehouse/importer/PackingListExcelImporter.groovy
```diff
@@ -0,0 +1,84 @@
...
+    void importData(ImportDataCommand command) {
+        throw new UnsupportedOperationException("This operation is not supported")
+    }
+
+
+    List<Map> toJson() {
+        return data.collect { [
+                palletName: it?.palletName,
+                boxName: it?.boxName,
+                product: it["productCode"],
```
jmiranda, line 76: Let's call the property "productCode" in order to ensure there's no confusion. 

The other fields (like binLocation) are fine, because there's ambiguity to what data will be in the Bin Location column (i.e. id, code, name of bin location). 

But there's no ambiquity in the Product Code column.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+            if (!inventoryItem) {
+                return ['inventoryItemNotFound']
+            }
+            Location currentLocation = AuthService.currentLocation
+            Integer quantity = productAvailabilityService.getQuantityAvailableToPromiseForProductInBin(inventoryItem, currentLocation, item.binLocation)
+            if (!quantity) {
+                return ['noStock']
+            }
+            if (quantityPicked > quantity) {
+                return ['overPick', quantityPicked]
```
jmiranda, line 92: name this `overpick` since it's an actual word and not two words "over pick"). This is similar to "putaway". It's putaway, not "put away".

jmiranda: Sources:
* https://www.merriam-webster.com/dictionary/overpick
* https://docs.oracle.com/en/cloud/saas/supply-chain-and-manufacturing/24b/famml/overpick-for-sales-orders-transfer-orders-and-work-orders.html#s20077043

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+            if (!item.binLocationFound) {
+                return ['binLocationNotFound']
+            }
+            ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
+            InventoryItem inventoryItem = item.product?.getInventoryItem(item.lotNumber)
+            if (!inventoryItem) {
+                return ['inventoryItemNotFound']
+            }
+            Location currentLocation = AuthService.currentLocation
+            Integer quantity = productAvailabilityService.getQuantityAvailableToPromiseForProductInBin(inventoryItem, currentLocation, item.binLocation)
```
jmiranda, line 87: I understand the desire to have this here in teh constraints. But this seems dangerous. I think I'd prefer if the validateData was doing this validation. One benefit would be that we'd inject the productAvailabilityService into the Importer once vs once each time domain.validate() is called (likely more than once per domain). 

With that said, there an argument for allowing the domain to call the service. 
https://stackoverflow.com/a/3654488/136597
https://stackoverflow.com/a/2460029/136597

But perhaps we could inject the service into the domain once
https://stackoverflow.com/a/9699609/136597

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+            if (!quantity) {
+                return ['noStock']
```
jmiranda, line 89: let's call this one 
`stockout` or `unavailable`. 

ANd make sure that !quantity returns true for both 0 and < 0.

grails-app/migrations/0.9.x/changelog-2024-08-02-1200-add-location-name-index.xml
jmiranda: Now that I've read through the entire review I'm understanding this a bit better. And while Artur is right that it probably isn't necessary for this PR, we **should** have an index on location name. However, we need to look into how we're using the name column when querying locations. 

For example, if we always query name and another column together then we'd want to create a composite index. Even if you do some queries with both columns and some queries with just one column, MySQL should use the composite index. So it's better to have the composite index then two separate indexes.

So while there's definitely a need to create an index on location name, we might want to investigate this as a separate issue as long as the performance issue (alluded to below) is no longer an issue.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -1025,6 +1025,21 @@ class ProductAvailabilityService {
         return quantityAvailableToPromiseByProductNotInBin ?: 0
     }
 
+    Integer getQuantityAvailableToPromiseForProductInBin(InventoryItem inventoryItem, Location location, Location binLocation) {
+        return ProductAvailability.createCriteria().get {
+            projections {
+                property("quantityAvailableToPromise")
```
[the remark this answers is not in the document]
kchelstowski: @jmiranda this is the issue I've once fixed: https://pihemr.atlassian.net/browse/OBPIH-6008
I've run your query on my local db, and then checked all the columns how they look like.
Here is an example:
```
id                                  |version|bin_location_id                 |inventory_item_id               |location_id                     |product_id|product_code|lot_number  |bin_location_name|quantity_allocated|quantity_on_hand|date_created       |last_updated       |quantity_on_hold|quantity_available_to_promise|
------------------------------------+-------+--------------------------------+--------------------------------+--------------------------------+----------+------------+------------+-----------------+------------------+----------------+-------------------+-------------------+----------------+-----------------------------+
0ac391ad-4d6d-4a27-bc01-77139d1fd332|      0|09f5e116831aefb00183377b1e5d58c4|09f5e1167c3c9801017c4ba7755715fe|09f5e1167e6ee8ec017e9c3c480f6f43|10001     |21064       |MWM062620N47|R-307CFC         |                 0|               0|2023-08-30 18:09:01|2023-08-30 18:09:01|               0|                            0|
f47db6bf-6ee9-4d73-8a3a-f7c58e39f5dc|     22|09f5e116831aefb00183377b1e5d58c4|09f5e1167c3c9801017c4ba7755715fe|09f5e1167e6ee8ec017e9c3c480f6f43|10001     |10001       |MWM062620N47|R-307CFC         |                 0|               0|2023-09-18 21:01:21|2023-10-25 21:18:03|               0|                            0|
```
so what needs to be done on production is to refresh PA on the locations where those rows exist

src/main/groovy/org/pih/warehouse/importer/PackingListExcelImporter.groovy
kchelstowski: I called it `product`, so that I would not have to do any mapping on the frontend side in order to bind the product in the `ImportPackingListItem`. 
I will change that, and will just adjust the frontend, so that for the validation endpoint we would send this property as:
```js
product: record.productCode
```

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: We decided not to perform the validation at import file level, hence we can't do it in the `validateData` method, as it would require us to refactor the whole thing now.
The `productAvailabilityService` is not injected per each item, since for each item we have the same memory reference to the service, so for each item it uses "the same" instance of product availability service.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: good point about the quantity < 0. I thought it's not possible to have negative quantity in the product availability table.

src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -98,36 +59,51 @@ const useOutboundImportForm = ({ next }) => {
...
+        fulfillmentDetails: basicDetails,
+        packingList: items,
+        sendingOptions,
+      }).then(() => apiClient.post('/api/fulfillments', {
+        fulfillmentDetails: basicDetails,
+        packingList: items,
+        sendingOptions,
+      })).then((response) => {
+        window.location = STOCK_MOVEMENT_URL.show(response.data.data.id);
+      });
```
kchelstowski, line 105: @awalkowiak don't look at this file at all. It's just a :spaghetti: on my testing purposes. I will make it cleaner etc in the frontend ticket.
I decided not to remove it though, to make the feature testable for Kasia and Manon

kchelstowski:
This one is ready for review

grails-app/controllers/org/pih/warehouse/api/FulfillmentApiController.groovy
```diff
@@ -0,0 +1,50 @@
...
+        errors.sendingOptions = CommandUtils.buildErrorsGroupedByField(command.sendingOptions.errors)
+
+        command.packingList.each { ImportPackingListItem item ->
+            if (item.hasErrors()) {
+                // Build errors for each row separately. Row id is the key.
+                // e.g. { 2: { binLocation: ["Bin location cannot be null"] } }
+                errors.packingList[item.rowId] = CommandUtils.buildErrorsGroupedByField(item.errors)
+            }
+        }
+    }
```
EWaterman, line 49: I like the thought process here. Maybe someday we can find a way to generalize this so that we can have a common format for our error responses.

src/main/groovy/org/pih/warehouse/outbound/ShippingRequest.groovy
```diff
@@ -0,0 +1,27 @@
...
+    ShipmentType shipmentType
+
+    String trackingNumber
+
+    Date expectedDeliveryDate
+
+    static constraints = {
+        expectedShippingDate(validator: { value, obj ->
+            // Check if shipping date happens before the expected delivery date (if delivery date specified)
+            if (obj.expectedDeliveryDate && value.after(obj.expectedDeliveryDate + 1)) {
```
EWaterman, line 19: why is the + 1 needed here? Is it adding a day?

src/main/groovy/org/pih/warehouse/importer/PackingListExcelImporter.groovy
```diff
@@ -0,0 +1,84 @@
...
+    void importData(ImportDataCommand command) {
+        throw new UnsupportedOperationException("This operation is not supported")
+    }
+
+
+    List<Map> toJson() {
+        return data.collect { [
+                palletName: it?.palletName,
+                boxName: it?.boxName,
+                productCode: it["productCode"],
```
EWaterman, line 76: why is productCode being accessed differently?

grails-app/i18n/messages.properties
```diff
@@ -1073,6 +1074,29 @@ fulfillmentItem.available.label=Available
 fulfillmentItem.packed.label=Packed
 fulfillmentItem.picked.label=Picked
 fulfillmentItem.requested.label=Requested
+
+# Messages related to outbound import
+fulfillmentRequest.origin.invalid=Origin can't be the same as destination
+fulfillmentRequest.destination.invalid=Destination can't be the same as origin
```
EWaterman, line 1080:
```suggestion
fulfillmentRequest.origin.sameAsDestination=Origin can't be the same as destination
fulfillmentRequest.destination.sameAsOrigin=Destination can't be the same as origin
```
Because "invalid" is very general but the error is a very specific error scenario.

grails-app/i18n/messages.properties
```diff
@@ -1073,6 +1074,29 @@ fulfillmentItem.available.label=Available
...
+fulfillmentRequest.destination.nullable=Destination cannot be null
```
EWaterman, line 1081:
```suggestion
fulfillmentRequest.destination.nullable=Destination cannot be empty
```
I feel like a non-developer would understand "empty" more than "null". Same comment for all these.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -0,0 +1,98 @@
...
+            }
+            return ['inventoryItemNotFound', lotNumber, item.product?.productCode]
+        })
+        expirationDate(nullable: true)
+        binLocation(nullable: true)
+        quantityPicked(min: 0, validator: { Integer quantityPicked, ImportPackingListItem item ->
+            if (!item.binLocationFound) {
+                return ['binLocationNotFound']
+            }
+            ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
```
EWaterman, line 82: I'm not a big fan of having to pull a service in like this to do validation. IMO request objects should be simple and not depend on any other large classes.

My preferred approach to handling situations like this is to add an `ImportPackingListValidator` that can autowire in whatever classes it needs.
- https://docs.spring.io/spring-framework/reference/core/validation/validator.html
- https://stackoverflow.com/questions/41996143/spring-service-method-and-a-complex-validation-logic-rules

That way request objects can act more like simple POJOs. But maybe there's a better way to achieve this via Grails that I don't know about.

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
```diff
@@ -70,4 +101,252 @@ class FulfillmentService {
 
         return command
     }
+
+    StockMovement createOutbound(ImportPackingListCommand command) {
+        // First create a requisition
```
EWaterman, line 106: IMO you probably don't need a lot of these comments since they're just reiterating what the method names are (which means you have good method names). My general approach is to have self-documenting code (ie writing code in a way that makes the functionality obvious by just reading the code) and then add comments if it's still unclear.

Not a big deal though and better to be overly verbose than have very unclear code.

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
```diff
@@ -70,4 +101,252 @@ class FulfillmentService {
...
+        }
+        if (trackingNumber) {
+            name.append("${separator}${trackingNumber}")
+        }
+        if (description) {
+            name.append("${separator}${description}")
+        }
+        return name
+                .toString()
+                .replace(Constants.SPACE_SEPARATOR, "")
```
EWaterman, line 143: Not really important but `SPACE_SEPARATOR` is somewhat misleading here since you're actually just trimming all white space from a string and it has nothing to do with the separator character.

I'd make this a `StringUtils.removeWhiteSpace(String s)` utility method that we can re-use anywhere.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
EWaterman: I'm okay with doing it this way for now, but I would like to brainstorm some potential longterm solutions

grails-app/controllers/org/pih/warehouse/api/FulfillmentApiController.groovy
kchelstowski: @EWaterman we actually have a generalized version - we have errors handler (ErrorsController). I'd say this is just a new case, where instead of throwing a popup error, we want to associate an error with a particular row in the table.

src/main/groovy/org/pih/warehouse/outbound/ShippingRequest.groovy
kchelstowski: @EWaterman yeah, a similar addition is implemented in the Shipment domain.

src/main/groovy/org/pih/warehouse/importer/PackingListExcelImporter.groovy
kchelstowski: good question - previously I was probably accessing some deeper associations (or copied it from somewhere) and didn't realize I've mixed the type of access. I will unify it.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: @EWaterman I didn't want to introduce a validator class or proceed with the validation in the service for this one property, not to mix two approaches - `validator`, and a custom validation in a validator class.
I know injecting a service into a domain is debatable, but since Grails' recommended validation happens in the domain (validator), I've gone this way. 
As I mentioned somewhere above - this doesn't "recreate" the product availability service, or anything like this, because the PA service is singleton, and we always get the same reference no matter which item we are iterating through.

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
kchelstowski: @EWaterman I also had a doubt whether to put those comments or not, but thought that maybe it's better to explain the iteration explicitly.

src/main/groovy/util/StringUtil.groovy
```diff
@@ -52,5 +53,9 @@ class StringUtil {
         return null
     }
 
+    static String removeWhiteSpace(String s) {
```
awalkowiak, line 56: Why not `StringUtils.deleteWhitespace` or `StringUtils.deleteSpace`?

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListErrors.groovy
```diff
@@ -0,0 +1,10 @@
+package org.pih.warehouse.outbound
+
+class ImportPackingListErrors {
+
+    Map<String, List<String>> fulfillmentDetails
+
+    Map<String, List<String>> sendingOptions
+
+    Map<String, Map<String, Map<String, List<String>>>> packingList = new HashMap<>()
```
awalkowiak, line 9: This nested map is confusing, perhaps a comment with an example could be useful.

awalkowiak (commented):
Two minor things

src/main/groovy/util/StringUtil.groovy
kchelstowski: I went with @EWaterman suggestion

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListErrors.groovy
kchelstowski: I believe I introduced a comment in the controller itself: 
```groovy
// Build errors for each row separately. Row id is the key.
 // e.g. { 2: { binLocation: ["Bin location cannot be null"] } }
errors.packingList[item.rowId] = CommandUtils.buildErrorsGroupedByField(item.errors)
```
but I don't mind adding it also here for clarity

grails-app/controllers/org/pih/warehouse/api/FulfillmentApiController.groovy
jmiranda: Can we move the build errors into CommandUtils as well?

grails-app/controllers/org/pih/warehouse/outbound/PackingListController.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+    UploadService uploadService
+
+    def importPackingList(ImportDataCommand command) {
+        MultipartFile importFile = command.importFile
+        File localFile = uploadService.createLocalFile(importFile.originalFilename)
+        importFile.transferTo(localFile)
+        DataImporter packingListImporter = new PackingListExcelImporter(localFile.absolutePath)
+
+        render([data: packingListImporter.toJson() ] as JSON)
```
jmiranda, line 20: Interesting idea to add the toJson to the importer object. I would have expected this to be the resposibility of the command object.  I need to think about this more, but it feels like the importer should be responsible for parsing the data and the command object should be responsible for rendering the data. 

I could be wrong. Maybe the importer is a glorified command object itself? Anyone else have any thoughts?

In general, it feels like we'll want to rememeber this when we refactor import/export since we want to be able to import and export in different formats. I'm not sure if the importer / exporter should be responsible for that. Instead we delegate that to a class that implements a Strategy pattern (or maybe Adapter pattern). 

https://pihemr.atlassian.net/browse/OBIG-4

jmiranda (commented):
@jmiranda to move comments to the new PR. 

https://github.com/openboxes/openboxes/pull/4786

grails-app/controllers/org/pih/warehouse/api/FulfillmentApiController.groovy
```diff
@@ -0,0 +1,50 @@
...
+    static void buildErrors(ImportPackingListCommand command, ImportPackingListErrors errors) {
+        // Build errors instance for each sub-step (fulfillmentDetails, sendingOptions, list of items)
+        // Errors will be grouped by the sub-step and by fields, e.g.: { fulfillmentDetails: { origin: ["Cannot be null"] } }
+        errors.fulfillmentDetails = CommandUtils.buildErrorsGroupedByField(command.fulfillmentDetails.errors)
+        errors.sendingOptions = CommandUtils.buildErrorsGroupedByField(command.sendingOptions.errors)
+
+        command.packingList.each { ImportPackingListItem item ->
+            if (item.hasErrors()) {
+                // Build errors for each row separately. Row id is the key.
+                // e.g. { 2: { binLocation: ["Bin location cannot be null"] } }
```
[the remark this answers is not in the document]
kchelstowski: @jmiranda having the `rowId` as the key makes it easier to access.
 If we went with the approach you've just proposed, we would have to do parse it on the frontend using some `Object.entires`/`Object.keys` in order to access all errors of a row, which would make the operation O(n^2), not O(1) as it is now (accessing a key of a Map is O(1)), so I expect that we could experience some performance issues if there were a lot of rows.

grails-app/controllers/org/pih/warehouse/outbound/PackingListController.groovy
kchelstowski: As for the importer itself, I was just following the pattern we already have, as it was painful to make it work. I agree there is a bit of mess with the exporters/importers and this should be refactored at some point, but answering your question I'd say we could treat the importer class as the command, since it contains the data

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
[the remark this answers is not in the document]
jmiranda: @kchelstowski Although you've fixed the data issue, the problem is that this scenario can still happen. So we need to understand how it happens and be defensive about how to handle it.

@awalkowiak Let me know if you have any thoughts. It might not be fixable but we shouldn't ignore it if it's a possibility.

PR #4776, opened 2024-08-02
OBPIH-6536 [backend] Create partial invoice for prepay PO

alannadolny:

> *A concise summary of what is being changed. Please provide enough context for reviewers to be able to understand the change and why it is necessary.*

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6536

**Description:**

:x: 

---
:x:

---

- [x] Backend automation tests ([unit](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013869579/Backend+Unit+Testing+Standards), [API](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013738522/Backend+Integration+Testing+Standards), smoke)
- [ ] Not Applicable

Check whether only not invoiced items/adjustments are visible on the Create Invoice page

---

- [x] The change has tests that prove the issue is fixed / the feature works as intended

2nd condition is met - the context is available in the ticket


grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
```diff
@@ -107,7 +108,7 @@ class InvoiceController {
             redirect(action: "list")
         }
 
-        Invoice invoice = invoiceService.generateInvoice(order)
+        Invoice invoice = prepaymentInvoiceService.generateInvoice(order)
```
EWaterman, line 111: What method is getting called when generating an invoice for a PO that wasn't pre-paid? If this API is used in non-prepay related flows we don't want to have it use the prepaymentInvoiceService

src/test/groovy/unit/org/pih/warehouse/invoice/InvoiceServiceSpec.groovy
```diff
@@ -0,0 +1,45 @@
...
+import org.pih.warehouse.product.Product
+import org.pih.warehouse.shipping.ShipmentItem
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Unroll
+class InvoiceServiceSpec extends Specification implements ServiceUnitTest<InvoiceService>, DataTest {
+
+    @Override
+    Class<?>[] getDomainClassesToMock() {
```
EWaterman, line 18: why is this needed? You can't just call mockDomains in setup?

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -260,6 +278,14 @@ class ShipmentItem implements Comparable, Serializable {
         return invoiceItems?.sum { it.quantity ?: 0 } ?: 0
     }
 
+    Integer getQuantityToInvoice() {
+        return quantity - quantityInvoiced
```
EWaterman, line 282: What's your plan for eventually handling cancellations/returns?

quantityInvoiced loops through all the invoices and sums the quantity, which works fine for the case where there are no quantity adjustments but... if a shipment item has quantity 3 and we cancelled or returned the shipment, we generate an invoice with quantity 0. In that scenario quantityInvoiced for the shipment item is 0, and isInvoiceable will be true (which is incorrect). We'll need some way to note that even if we invoice for less than the quantity shipped for that shipment item, we've still "fully invoiced" it.

Side note: I remember chatting with Manon about whether we need to be able to split a single shipment item up into multiple invoices (Ex: we have a shipment item with quantity 3 which has an invoice item with quantity 1 and another invoice item with quantity 2) and I thiiiiink she said we don't need the ability to do that (probably worth confirming).

So maybe it'd just be a matter of changing the getQuantityInvoiced method to return 0 for shipment items without an invoice and shipmentItem.quantity for shipment items that do have an invoice. So we ignore the quantity in the invoiceItem for the purpose of determining if the shipment item has been fully invoiced. (Which we can do because we can safely assume that each shipmentItem will only ever have a single invoiceItem so if it was invoiced at all, it's done.)

I know the work to implement cancellations isn't covered by this ticket, I just want to make sure it'll eventually be able to work within this system.

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
EWaterman: nvm I see the first line in the method checks that it has a prepayment invoice already.

Have you considered moving these methods to a PrepaymentInvoiceController since you're also doing that for the service methods?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -0,0 +1,69 @@
...
+import org.pih.warehouse.order.OrderAdjustment
+import org.pih.warehouse.order.OrderItem
+import org.pih.warehouse.order.OrderItemStatusCode
+import org.pih.warehouse.shipping.ShipmentItem
+import grails.gorm.transactions.Transactional
+
+@Transactional
+class PrepaymentInvoiceService {
+
+    InvoiceService invoiceService
```
EWaterman, line 14: Generally my preference would be to not have services call other services but instead move common code to shared util classes (Ex: InvoiceNumberGenerator, InvoiceItemGenerator (or whatever names)). It breaks the code up into more manageable units (easier testing too).

I don't expect you to to make any kind of change like that here though. Just noting down my thoughts.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -462,6 +464,14 @@ class Order implements Serializable {
         return !hasInvoice && isPrepaymentRequired && hasActiveItemsOrAdjustments
     }
 
+    Set<OrderItem> getInvoiceableOrderItems() {
+        return orderItems.findAll { it.invoiceable }
+    }
+
+    Set<OrderAdjustment> getInvoiceableAdjustment() {
```
EWaterman, line 471:
```suggestion
    Set<OrderAdjustment> getInvoiceableAdjustments() {
```

grails-app/controllers/org/pih/warehouse/invoice/InvoiceController.groovy
alannadolny: I think it's not worth it, to be honest. Later we will add PrepaymentInvoiceApiController, so this one will have only one endpoint.

PR #4778, opened 2024-08-06
OBPIH-6497 [backend] Determine encumbered and invoiceable items or adjustments (fix after QA)

alannadolny:


grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -285,14 +286,23 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
         return quantityInvoicedInStandardUom >= quantity
     }
 
+    Integer getInvoicedQuantity() {
+        return allInvoiceItems
+                ?.findAll { !it?.invoice?.isPrepaymentInvoice }
+                ?.sum { it.quantity } ?: 0
```
awalkowiak, line 292: Could you add a comment that this quantity is already in standard uom? Here or in the `getQuantityAvailableToInvoice`, so it is clear?

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -285,14 +286,23 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
...
         return canceled
                 ? null
-                : quantityShippedInStandardUom - quantityInvoicedInStandardUom
+                : quantityShippedInStandardUom - invoicedQuantity
     }
 
     Boolean isInvoiceable() {
-        return encumbered && (canceled || quantityAvailableToInvoice > 0)
+        return hasPrepaymentInvoice &&
+                !hasRegularInvoice &&
```
awalkowiak, line 303: Is this `!hasRegularInvoice` correct? What if an order item will be split across multiple shipments and then multiple final invoices (= multiple regular invoices)?

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
awalkowiak: btw, for order adjustments it makes sense, but I am not sue if it makes sense for order items

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -109,7 +109,7 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
     }
 
     Boolean isInvoiceable() {
-        return !isInvoiced && order.placed
+        return hasPrepaymentInvoice && !hasRegularInvoice && order.placed
```
EWaterman, line 112: Is this because `isInvoiced` used for non-prepay flows?

src/test/groovy/unit/org/pih/warehouse/order/OrderAdjustmentSpec.groovy
```diff
@@ -21,11 +22,11 @@ class OrderAdjustmentSpec extends Specification implements DomainUnitTest<OrderA
...
         where:
-        isInvoiced | orderStatus         || isInvoiceable
-        true       | OrderStatus.PENDING || false
-        false      | OrderStatus.PLACED  || true
-        true       | OrderStatus.PLACED  || false
-        false      | OrderStatus.PENDING || false
+        orderStatus         | hasRegularInvoice  || isInvoiceable
+        OrderStatus.PENDING | false              || false
+        OrderStatus.PLACED  | false              || true
+        OrderStatus.PENDING | false              || false
```
EWaterman, line 28: hasRegularInvoice should be true here right? Otherwise it;s the same as the first case

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
EWaterman: yes an order item can be split across multiple shipments, but I'm not sure if we support a shipment item being split across multiple invoices. I think we expect that a shipment item can only be invoiced once. This is because if we want to invoice a quantity 2 but a quantity 4 was shipped, then that indicates 2 qty was returned (or whatever) and the shipment item should then be considered fully invoiced (even though we only invoiced quantity 2 of 4).

EWaterman: btw here's an example of what an invoice would look like if we have multiple shipments for each order item

5580b76c-fef9-4d3d-9b25-7a7770d73d7a

awalkowiak:
I will be taking over this one

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak: `isInvoiced` looks for posted invoices, and here we want to have both posted and not yet posted invoices

kchelstowski (approved):

jmiranda (commented):
Can we please follow the convention of quantity?
```
quantityDescriptionOfQuantity
```
like
```
quantityOrdered
quantityShipped
quantityOnHand
quantityAvailableToPromise
```
I've seen a few times where descriptor leak into the beginning of the property.
```
invoicedQuantity
postedQuantityInvoicedInStandardUom
```
If there's a good reason to do this please let me know ... otherwise it should always be 
```
quantityDescribeWhatTheQuantityIsFor
```

PR #4779, opened 2024-08-06
OBPIH-6499 [backend] Calculate inverse items

alannadolny:
Because I am going ooo, I am pushing my branch with the next steps:
Inverse items are created during the creation of regular, but relations are missing, so:
Check if adding relations (to shipment items etc.) to inverse items will break logic, if so, check how to avoid that, or do it another way. We have to add those relations because some data is missing while displaying invoice items on React pages. After fixing the missing data, inverse the quantity (should be a simple change in calculation + remember about prepaid / 50, etc.)


awalkowiak:
I will be taking over this one

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -36,6 +35,7 @@ class InvoiceItem implements Serializable {
 
     String id
     Invoice invoice
+    InvoiceItemType invoiceItemType
```
awalkowiak, line 38: @jmiranda let me know if you prefer this one as a separate `InvoiceItemType` domain that has `name` and `code` as  `InvoiceItemTypeCode` enum instead straight away using enum here, and I'll adjust that

grails-app/migrations/0.9.x/changelog-2024-08-05-1300-add-invoice-item-type-to-invoice-item.xml
```diff
@@ -0,0 +1,16 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="anadolny" id="050820241300-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="invoice_item" columnName="invoice_item_type"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="invoice_item">
+            <column name="invoice_item_type" type="CHAR(38)" />
```
EWaterman, line 13: Is this CHAR(X) what we typically do for our enum fields?

grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -316,64 +316,6 @@ class InvoiceService {
         return invoiceItem
     }
 
-    InvoiceItem createFromOrderItem(OrderItem orderItem) {
```
awalkowiak, line 319: If anyone wonders, I moved these to `PrepaymntInvoiceService`

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -188,11 +203,14 @@ class InvoiceItem implements Serializable {
                 uom: unitOfMeasure,
                 amount: amount,
                 unitPrice: unitPrice,
-                totalAmount: totalAmount,
-                totalPrepaymentAmount: totalPrepaymentAmount,
                 orderAdjustment: orderAdjustment,
                 productName: product?.name,
-                displayNames: product?.displayNames
+                displayNames: product?.displayNames,
+                type: invoiceItemType?.name(),
```
EWaterman, line 209: if you're calling name() here why do we need the type field in the enum?

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
```diff
@@ -192,7 +192,7 @@ class InvoiceApiController {
         render([data: invoice?.toJson()] as JSON)
     }
 
-
+    @Deprecated
```
EWaterman, line 195: can you add a comment to all these deprecated methods explaining why they're being deprecated? (Or are you planning on fully removing them in a later commit?)

grails-app/migrations/0.9.x/changelog-2024-08-05-1300-add-invoice-item-type-to-invoice-item.xml
awalkowiak: I did not notice this CHAR(38) when I picked this up, I think usually we do VARCHAR(255) for enum values, CHAR(38) is for ids - I can change that

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
awalkowiak: It is built on the `Enum`

grails-app/controllers/org/pih/warehouse/api/InvoiceApiController.groovy
awalkowiak: Sure, will do it

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -13,6 +13,40 @@ class PrepaymentInvoiceService {
 
     InvoiceService invoiceService
 
+    Invoice generatePrepaymentInvoice(Order order) {
```
EWaterman, line 16: where are these methods being called? I don't see any other controllers

awalkowiak: In the `generatePrepaymentInvoice` action in the `InvoiceController`

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -23,47 +57,150 @@ class PrepaymentInvoiceService {
...
+            invoice.addToInvoiceItems(inverseItem)
         }
 
         return invoice.save()
     }
 
-    Invoice generatePrepaymentInvoice(Order order) {
-        if (order.orderItems.any { it.hasInvoices } || order.orderAdjustments.any { it.hasInvoices }) {
-            throw new Exception("Some order items or order adjustments for this order already have been invoiced")
+    InvoiceItem createFromOrderItem(OrderItem orderItem) {
```
EWaterman, line 87: suggestion: make all these createFromX methods private. The only thing that needs to be public is whatever we're exposing to external sources (the controllers). That way it's clearer what the entrypoints to the class are

grails-app/domain/org/pih/warehouse/invoice/InvoiceItemType.groovy
```diff
@@ -0,0 +1,14 @@
+package org.pih.warehouse.invoice
+
+enum InvoiceItemType {
+
+    INVERSE('Inverse'),
```
awalkowiak, line 5: @jmiranda Let me know if other types come to your mind you'd like to have added here

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
EWaterman: yes but I mean is InvoiceItemType.type being used anywhere? name() just uses the actual name of the enum itself.

Personally I prefer to define the toString() method of enums and use that (and in this case you probably want to return the type field). It lets us be more intentional about how our items get stringified.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -13,6 +13,40 @@ class PrepaymentInvoiceService {
...
+        if (order?.orderItems?.empty && order?.orderAdjustments?.empty) {
+            throw new Exception("No order items or order adjustments found for given order")
+        }
+
+        BigDecimal prepaymentPercent = order.paymentTerm?.prepaymentPercent?:100
+
+        order.activeOrderItems.each { OrderItem orderItem ->
+            InvoiceItem invoiceItem = createFromOrderItem(orderItem)
+            // This is prepayment item, we need to adjust the amount according to the payment terms
+            invoiceItem.amount = invoiceItem.amount * (prepaymentPercent / 100)
```
EWaterman, line 34: didn't you say amount wasn't being used?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -13,6 +13,40 @@ class PrepaymentInvoiceService {
...
+
+        Invoice invoice = invoiceService.createFromOrder(order)
+        invoice.invoiceType = InvoiceType.findByCode(InvoiceTypeCode.PREPAYMENT_INVOICE)
+        invoiceService.createOrUpdateVendorInvoiceNumber(invoice, order.orderNumber + Constants.PREPAYMENT_INVOICE_SUFFIX)
+
+        if (order?.orderItems?.empty && order?.orderAdjustments?.empty) {
+            throw new Exception("No order items or order adjustments found for given order")
+        }
+
+        BigDecimal prepaymentPercent = order.paymentTerm?.prepaymentPercent?:100
```
EWaterman, line 29: This 100 should probably be in a `DEFAULT_PAYMENT_PERCENT` constant?

Actually (and I know this probably is pre-existing code), now that I think about it, is this even safe to do? These are the payment terms we have:

```
| id    | version | name                 | code  | description | date_created        | last_updated        | prepayment_percent | days_to_payment |
|-------|---------|----------------------|-------|-------------|---------------------|---------------------|--------------------|-----------------|
| COD   | 0       | Cash On Delivery     | COD   | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | NULL               | NULL            |
| NET30 | 0       | Net 30               | NET30 | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | NULL               | 30              |
| PIA   | 0       | Payment In Advance   | PIA   | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | NULL               | NULL            |
| PP100 | 0       | Prepayment 100       | PP100 | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | 100                | NULL            |
| PP25  | 0       | Prepayment 25/Net 30 | PP25  | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | 25                 | 30              |
| PP50  | 0       | Prepayment 50/Net 30 | PP50  | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | 50                 | 30              |
| US    | 0       | Upon shipment        | US    | NULL        | 2023-08-14 15:02:59 | 2023-08-14 15:02:59 | NULL               | 30              |

```

So I'm hoping we're not allowing the user to use non-prepayment terms here. Not sure how PIA differs from PP100 but I'm assuming that's why we need this null check.

Anyways, this isn't new code so I'm not asking for a refactor, just noting it down as it seemed a bit strange.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -23,47 +57,150 @@ class PrepaymentInvoiceService {
...
+                invoiceItemType: InvoiceItemType.REGULAR
+        )
+        invoiceItem.addToOrderAdjustments(orderAdjustment)
+        return invoiceItem
+    }
 
-        order.activeOrderAdjustments.each { OrderAdjustment orderAdjustment ->
-            InvoiceItem invoiceItem = invoiceService.createFromOrderAdjustment(orderAdjustment)
-            invoice.addToInvoiceItems(invoiceItem)
+    InvoiceItem createInverseItemForOrderItem(OrderItem orderItem) {
```
EWaterman, line 158: if this is only for items with no shipments (ie cancelled ones), it's probably best to be clear in the method name. `createInverseItemForCancelledOrderItem()` or something

src/test/groovy/unit/org/pih/warehouse/invoice/InvoiceServiceSpec.groovy
EWaterman, line 1: why delete? Is it not worth still testing the old non-prepay flows?

src/test/groovy/unit/org/pih/warehouse/invoice/PrepaymentInvoiceServiceSpec.groovy
```diff
@@ -50,52 +52,31 @@ class PrepaymentInvoiceServiceSpec extends Specification implements ServiceUnitT
         null != service.generateInvoice(order)
     }
 
-    void 'PrepaymentInvoiceService.generateInvoice() should create invoice not including already invoiced items'() {
+    void 'PrepaymentInvoiceService.createFromShipmentItem() should calculate quantity to invoice: #quantityToInvoice when quantity is: #quantity and quantity invoiced: #quantityInvoiced'() {
```
EWaterman, line 55: not worth testing generatePrepaymentInvoice() (or whatever) here?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -23,47 +57,150 @@ class PrepaymentInvoiceService {
...
+                quantityUom: orderItem?.quantityUom,
+                quantityPerUom: orderItem.quantityPerUom ?: 1,
+                unitPrice: orderItem.unitPrice,
+                // For non-canceled order items invoice item amount is equal to (order item quantity * unit price)
+                amount: (orderItem.quantity ?: 0) * (orderItem.unitPrice ?: 0.0),
+                invoiceItemType: InvoiceItemType.REGULAR
+        )
+        invoiceItem.addToOrderItems(orderItem)
+        return invoiceItem
+    }
```
EWaterman, line 119: You could simplify this down to something like this if you want:
```
  private InvoiceItem createFromOrderItem(OrderItem orderItem) {
      // For non-canceled order items invoice item amount is equal to (order item quantity * unit price)
      int amount = orderItem.orderItemStatusCode == OrderItemStatusCode.CANCELED
              ? 0 : (orderItem.quantity ?: 0) * (orderItem.unitPrice ?: 0.0)

      InvoiceItem invoiceItem = new InvoiceItem(
              quantity: orderItem.quantity,
              product: orderItem.product,
              glAccount: orderItem.glAccount ?: orderItem.product.glAccount,
              budgetCode: orderItem?.budgetCode,
              quantityUom: orderItem?.quantityUom,
              quantityPerUom: orderItem.quantityPerUom ?: 1,
              unitPrice: orderItem.unitPrice,
              amount: amount,
              invoiceItemType: InvoiceItemType.REGULAR
      )
      invoiceItem.addToOrderItems(orderItem)
      return invoiceItem
  }
```
I know you probably did it this way intentionally to keep the cancelled and non-cancelled flow separate but it's probably fine to condense it at this point since it's just one field that's different. Up to you though.

src/test/groovy/unit/org/pih/warehouse/invoice/InvoiceServiceSpec.groovy
awalkowiak: It would be empty. I moved the `createFromShipmentItem` test to `PrepaymentInvoiceServiceSpec` since I moved these creates to the other service

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: This `amount` is the one that is going to be used from now (this is a field on invoice item). The `totalAmount` transient was the deprecated one.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: Added the constant, but yeah this was an old calculation, I think it was done like that on purpose - if a term does not have a prepayment percentage we always prepay full amount

src/test/groovy/unit/org/pih/warehouse/invoice/PrepaymentInvoiceServiceSpec.groovy
awalkowiak: To be honest there were so many breaking changes, that I was just only fixing what was not working. I am going to add a ticket to revisit tests after the whole implementation on the backend is sealed after QA, because going back and forth on these tests at this stage is frustrating

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: Yeah I think I am going to leave it as is for now (could be adjusted together with tests coverage ticket)

awalkowiak:
@EWaterman I replied (and made change requests) to your comments, but don't be mad at me for resolving conversations, I just kept track of resolved requests with this.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItemType.groovy
```diff
@@ -0,0 +1,18 @@
...
+    PREPAYMENT('Prepayment'),
+    REGULAR('Regular')
+
+    String type
+
+    String toString() {
+        return name()
```
EWaterman, line 12:
```suggestion
        return type
```
Either that or I'd just remove the type field entirely since it's not being used anywhere but ultimately it's not a big deal.

EWaterman (approved):

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
jmiranda: Yeah, now that I'm seeing how it's being used, I don't think InvoiceItemType is correct. An invoice item type would describe what the invoice item is for i.e. product, service, adjustment, subscription, labor, tax, discount, fees, etc.

What we're talking about here is something I still don't really understand. But it feels like it might be the invoice type (prepayment, standard, final) and should go at the Invoice level. 

This is why I wish we had done the "from scratch" design instead of trying to build on what we had (at least from the data model perspective) so that we could have spent more time thinking about the proper way to do this. Frankly, I don't have a good enough understanding of it so I guess I'll leave it up to you. But I would really consider moving this "type" to the invoice and derive the "inverse" logic based on, I guess whether the item is in an invoice with InvoiceTypeCode.FINAL (and whatever other)

In either case, the safer option would be to use the enum here since we could always migrate to using a type domain that maps to the enum later.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItemType.groovy
jmiranda: If we added InvoiceTypeCode.FINAL, could the INVERSEness of an item be derived?  I assume that the other two (prepayment, regular) could already be derived from the InvoiceType. So even if there's a need to store whether an invoice item is an "inverse" item, we could do so with just a boolean field.

Side note: If/when you add REGULAR InvoiceTypeCode it should be called STANDARD.

Source: https://www.freshbooks.com/hub/invoicing/types-of-invoices

awalkowiak: @jmiranda Yeah, to be honest, this part was not designed properly, because we wanted to go with calculating all the time and we switched to keeping inverse items on the final invoice right before we started the development. So something that was my **suggested direction** in case we want to go with storing inverse items somehow was cemented in my brain as the version we want to do. As deriving item type from invoice type does not work for me here (due to mixing both standard/regular items with inverse items), I decided that the boolean fits here more, and this is what I implemented for now (we can still revise it again in the final feature PR).

alannadolny:
LGTM, but I can't approve this PR because I created it :smile:

grails-app/domain/org/pih/warehouse/invoice/InvoiceItemType.groovy
awalkowiak: @EWaterman i replaced InvoiceItemType with inverse boolean

PR #4781, opened 2024-08-08
OBPIH-6497 Fix comparing invoice quantities in UoM and in standard Uom

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6497

**Description:**
This is a fix to issues found during QA. Turned out that we had a few places that were mixing up quantities in UoM and quantities in standard UoM. 

---

- [X] Backend automation tests ([unit](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013869579/Backend+Unit+Testing+Standards), [API](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013738522/Backend+Integration+Testing+Standards), smoke)
- [ ] Not Applicable

---

- [X] The change has tests that prove the issue is fixed / the feature works as intended

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/Order.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/Order.groovy
@@ -480,7 +480,7 @@ class Order implements Serializable {
             3. Quantity greater than 0 of the item has been shipped and not all of that quantity has already been invoiced.
     **/
     Boolean getCanGenerateInvoice() {
-        // If prepayment invoice doesn't exist we can't generate invoice
+        // If prepayment invoice for this order doesn't exist we can't generate invoice
         if (!hasPrepaymentInvoice) {
             return false
         }
@@ -495,13 +495,11 @@ class Order implements Serializable {
     }
 
     Boolean hasInvoiceableOrderItem() {
-        return orderItems.any { it.invoiceable }
+        return orderItems.any { it.canBeOnRegularInvoice() && it.invoiceable }
     }
 
     Boolean hasInvoiceableOrderAdjustment() {
-        return orderAdjustments.any {
-            it.invoiceable
-        }
+        return orderAdjustments.any { it.canBeOnRegularInvoice() && it.invoiceable }
     }
 
     def getActiveOrderItems() {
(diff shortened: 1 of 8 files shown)


kchelstowski (approved):

EWaterman:
@awalkowiak can you explain the difference between standard UOM vs. non-standard and when we want to use each? I'm not clear which is supposed to be used in any given scenario

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -108,8 +108,23 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
...
 
+    /**
+     * Adjustment can be on a regular invoice if it is not canceled or if it is canceled, but has prepayment
+     * */
+    Boolean canBeOnRegularInvoice() {
+        if (canceled) {
+            return hasPrepaymentInvoice
+        }
+
+        return true
```
EWaterman, line 119: can be simplified to `return !canceled || hasPrepaymentInvoice` if you want

grails-app/domain/org/pih/warehouse/order/OrderItem.groovy
```diff
@@ -320,16 +321,29 @@ class OrderItem implements Serializable, Comparable<OrderItem> {
...
 
+    /**
+     * Item can be on a regular invoice if it is not canceled or if it is canceled, but has prepayment
+     * */
+    Boolean canBeOnRegularInvoice() {
+        if (canceled) {
+            return hasPrepaymentInvoice
+        }
+
+        return true
```
EWaterman, line 332: same here. Can be simplified to return !canceled || hasPrepaymentInvoice if you want

awalkowiak:
@EWaterman sure: 
Order items have quantity, unit of measure, and quantity per unit of measure that is not always equal to one. Things are ordered and invoiced in "unit of measure" quantity (which is equal to `quantity`), but shipments and receipts are done in standard uom quantity (which is equal to `quantity from order * quantity per uom`), so if we are comparing invoiced quantity with the shipped quantity we have to be sure we are comparing correct values (in this case either both compared quantities have to be `quantity in standard uom` or both to be in `uom quantity`). 

I know that there are some plans for the upcoming release(s) to make this less confusing.

grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -495,13 +495,11 @@ class Order implements Serializable {
     }
 
     Boolean hasInvoiceableOrderItem() {
-        return orderItems.any { it.invoiceable }
+        return orderItems.any { it.canBeOnRegularInvoice() && it.invoiceable }
```
EWaterman, line 498: is there a reason to not move the `canBeOnRegularInvoice` methods to within the `invoiceable` methods? Do we ever need to know if an item/adjustment is invoiceable outside of the context of generating a regular invoice for it?

awalkowiak: It has to be separate for checking if the order `isFullyInvoiceble`. The `canBeOnRegularInvoice` still could be part of `invoiceable` for this case that you pointed out, but since I needed it separately for another place, I decided not to combine these together (at least for now).

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak: Yeah, when I was doing the improvements to fix some edge cases, I wanted to keep it as clear as possible, to avoid any further confusion. I want to add a separate subtask for adding more tests (and improving currently existing ones) that will be done after the solution is working properly, then this could be changed to a one-liner there, if this is preferable.

PR #4791, opened 2024-08-19
[OBPIH-6626] [Bug-Fix] Prevent multiple full outbound imports from happening

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6626](https://pihemr.atlassian.net/browse/OBPIH-6626)

**Description:**
TLDR; of the issue: After submitting the full outbound request user is waiting for the redirect to happen and is able to click "Finish" button since after first submit it is not disabled and as a result for each click submits a "NEW"  full outbound stock movement.

My solution was to display the spinner overlay which should prevent the user from interacting with the UI while the request is being processed.

---

image

---

- [X] Not Applicable

---

[OBPIH-6626]: https://pihemr.atlassian.net/browse/OBPIH-6626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/hooks/outboundImport/useOutboundImportForm.js
+++ b/src/js/hooks/outboundImport/useOutboundImportForm.js
@@ -14,10 +14,13 @@ import { STOCK_MOVEMENT_URL } from 'consts/applicationUrls';
 import NotificationType from 'consts/notificationTypes';
 import { DateFormat } from 'consts/timeFormat';
 import useOutboundImportValidation from 'hooks/outboundImport/useOutboundImportValidation';
+import useSpinner from 'hooks/useSpinner';
 import useTranslate from 'hooks/useTranslate';
 
 const useOutboundImportForm = ({ next }) => {
   const translate = useTranslate();
+  const spinner = useSpinner();
+
   const { validationSchema } = useOutboundImportValidation();
   const { currentLocation } = useSelector((state) => ({
     currentLocation: state.session.currentLocation,
@@ -108,6 +111,7 @@ const useOutboundImportForm = ({ next }) => {
         'content-type': 'multipart/form-data',
       },
     };
+    spinner.show();
     const importPackingListResponse = await packingListApi.importPackingList(formData, config);
     const { basicDetails, sendingOptions } = buildDetailsPayload(values);
 
@@ -144,7 +148,7 @@ const useOutboundImportForm = ({ next }) => {
       const tableDataGrouped = groupTableDataByErrors(e.response.data);
       setTableData([...tableDataGrouped.itemsWithErrors, ...tableDataGrouped.itemsWithoutErrors]);
     }
-
+    spinner.hide();
     next();
   };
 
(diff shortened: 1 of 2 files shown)


src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -166,6 +171,7 @@ const useOutboundImportForm = ({ next }) => {
       // If the save went sucessfully, redirect to the stock movement view page
       window.location = STOCK_MOVEMENT_URL.show(response.data?.data?.id);
     } catch (e) {
+      spinner.hide();
```
alannadolny, line 174: Is this `hide` only in the catch block? What happens if everything goes well?

drodzewicz: if everything goes well we are going to get redirected to the show page of the stock movement.

When we call 
```js
window.location = STOCK_MOVEMENT_URL.show(response.data?.data?.id);
```
it still takes some time and if we hide the spinner overlay before redirecting, then user will still be able to click the **finish** button and the issue will not be resolved.
I decided to keep the show overlay until we are redirected, I believe this looks like a natural behavior for the user - it's like after clicking finish we are waiting for the show page to load.

alannadolny (approved):

awalkowiak (approved):

src/js/hooks/outboundImport/useOutboundImportForm.js
kchelstowski: I think the `hide` should be called in the finally. Because of redux-persist we might then come back to the React page with the show state of the spinner.

kchelstowski:
one comment above

kchelstowski:
Moreover I think that the backend should also prevent it. I had that in mind, but since I focused on happy case, I didn't make such validations, but I strongly believe we should also have some kind of "if shipment exists, do not allow to create", etc.
I will try to have that in mind when doing final cleanup.

src/js/hooks/outboundImport/useOutboundImportForm.js
drodzewicz: `spinner` is blacklisted and is not stored in the redux-presist

drodzewicz:
@kchelstowski I might be wrong but I am not sure how would that be possible? is there any identifier that we can use to say that this outbound is already created? But I don't think so...
From my perspective backend is working properly, until the very last step we are not modifying an existing stock movement  but creating a new one based on the provided parameters

kchelstowski:
@drodzewicz after adding the spinner, it won't be possible from the frontend perspective, and what I meant is rather something "nice to have" especially if we ever want to support editing an outbound through that workflow - this would then require refactor of API to make additional checks, but we can live without that for now.

src/js/hooks/outboundImport/useOutboundImportForm.js
kchelstowski: ok, I was to lazy to check it myself, so thanks for confirming.

PR #4796, opened 2024-08-20
[OBPIH-6612] add inverse prepayment invoice items for existing data

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6612

**Description:** Now that we're storing inverse prepayment invoice items on the server, we need to migrate our existing non-prepayment (aka final) invoices to include those inverses as well since historically we've generated the inverses on demand. See ticket for details.

---

- [x] Backend automation tests ([unit](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013869579/Backend+Unit+Testing+Standards), [API](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3013738522/Backend+Integration+Testing+Standards), smoke)
- [ ] Not Applicable

**Description of test plan (if applicable):** This is going to require thorough testing to make sure that we don't muck up any of our environments by running this script. We'll need to populate a local db with some old data then run the migration to ensure everything looks correct.

---

- [x] The change has tests that prove the issue is fixed / the feature works as intended


EWaterman:
@awalkowiak lets chat about this PR either during tech huddle or a JEAM meeting. I'm hoping we can refine this a bit since it's pretty crude in its current state. Maybe we can build a lot of this logic into the PrepaymentInvoiceService (or maybe a new PrepaymentInvoiceMigrationService) and unit test the heck out of it to know that it works.

It also doesn't seem to run on app startup so I need to figure out why...

awalkowiak:
@EWaterman 
> It also doesn't seem to run on app startup so I need to figure out why...

You have to add this migration file in the [0.9.x/changelog.xml](https://github.com/openboxes/openboxes/blob/develop/grails-app/migrations/0.9.x/changelog.xml). Looks like (at least in this PR) it is missing there.

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                }
+
+                /*
+                 * For every pre-existing prepayment invoice, for every invoice item, add an inverse invoice item
+                 * to the final invoice of the order.
+                 *
+                 * For existing data from before the partial invoicing feature was introduced, ALL non-prepayment
+                 * invoices will ALWAYS be the final invoice, meaning every order will have one prepayment invoice
+                 * and one non-prepayment (aka final) invoices.
```
EWaterman, line 72: I'm second guessing myself. Is this even true? Against develop I was just able to generate two separate invoices for an order with two items, one invoice per item...

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -110,7 +110,7 @@ class InvoiceItem implements Serializable {
         quantityPerUom(nullable: false)
         amount(nullable: true)
         unitPrice(nullable: true)
-        inverse(nullable: true)
+        inverse(nullable: false)
```
EWaterman, line 113: @awalkowiak the migration script used to be looping through all existing invoice items and marking them as `inverse = false` which I realize is the same as doing this and since the code will probably break if this value isn't set, then I think this makes sense to make as non-null.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,108 @@
...
+        // Else if the invoice item is for a canceled order item.
+        OrderItem orderItem = prepaymentInvoiceItem.orderItem
+        if (orderItem.orderItemStatusCode == OrderItemStatusCode.CANCELED) {
+            InvoiceItem inverseItem = prepaymentInvoiceService.createInverseItemForCanceledOrderItem(orderItem)
+            return inverseItem ? [inverseItem] : []
+        }
+
+        // Else the invoice item is for a non-canceled order item. The item might be split across multiple shipments
+        // so we need to loop the shipment items and generate an inverse for each.
+        // TODO: This was copied from PrepaymentInvoiceService. Why do we need to do this? Shouldn't it be a single inverse for each prepay, regardless of if it was split over multiple shipments?
```
EWaterman, line 93: @awalkowiak highlighting this for discussion

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
awalkowiak: :thinking: Did you try it on the staging too? I think there might be some initial implementation of this feature on the develop (allowing to generate it, so perhaps that why you could do it)

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
```diff
@@ -0,0 +1,87 @@
...
+                        return
+                    }
+                    Order order = orders[0]
+
+                    // Find the regular invoice that we will be populating with the inverse items. We expect that
+                    // there is always only a single (final) invoice. If there is no regular invoice, or it already
+                    // has inverse items, skip it.
+                    List<Invoice> regularInvoices = order.invoices.findAll { it.isRegularInvoice}
+                    if (!regularInvoices || regularInvoices.size() != 1) {
+                        totalOrdersWithBadRegInvoice++
```
awalkowiak, line 65: Technically lack of regular invoice does not indicate it is "bad", someone might just generate prepayment invoice and did not manage to generate final

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
```diff
@@ -0,0 +1,87 @@
...
+                        return
+                    }
+
+                    prepaymentInvoiceMigrationService.generateInverseInvoiceItems(prepaymentInvoice, regularInvoice)
+
+                    totalInvoicesMigrated++
+                }
+
+                println("Invoice migration complete!")
+                println("${totalInvoicesMigrated} invoices were migrated.")
```
awalkowiak, line 80: I'd add a count of `Invoice.findAllByInvoiceType(prepaymentInvoiceType)`, to have a better grasp if these numbers are too high or too low comparing to total prepayment invoices

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
awalkowiak: Partial invoicing had to made from the shipment items, but for this case, where we have only one final invoice, I think we could just make 100% copy prepayment item (well actually, 99%, because amount will multiplied by `-1`, and quantity 0 for canceled items)

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,108 @@
...
+            prepaymentInvoice.invoiceItems.find { it.amount == null }.each { InvoiceItem prepaymentInvoiceItem ->
+                prepaymentInvoiceItem.amount = computePrepaymentInvoiceItemAmount(prepaymentInvoiceItem)
+                prepaymentInvoiceItem.save(failOnError: true, flush: true)
+            }
+        }
+    }
+
+    // TODO: Can we merge this with PrepaymentInvoiceService? That way if the amount calculation ever changes, it will get picked up automatically here.
+    private BigDecimal computePrepaymentInvoiceItemAmount(InvoiceItem prepaymentInvoiceItem) {
+        BigDecimal prepaymentPercent = (prepaymentInvoiceItem.order.paymentTerm?.prepaymentPercent ?: Constants.DEFAULT_PAYMENT_PERCENT) / 100
```
awalkowiak, line 40: In my currently pending I moved this to transient on order (you could rebase it once its merged)

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,108 @@
...
+     */
+    void generateInverseInvoiceItems(Invoice prepaymentInvoice, Invoice finalInvoice) {
+        prepaymentInvoice.invoiceItems.each { InvoiceItem prepaymentInvoiceItem ->
+            List<InvoiceItem> inverseItems = createInverseInvoiceItems(prepaymentInvoiceItem, finalInvoice)
+            inverseItems?.each { finalInvoice.addToInvoiceItems(it) }
+        }
+        finalInvoice.save(failOnError: true, flush: true)
+    }
+
+    private List<InvoiceItem> createInverseInvoiceItems(InvoiceItem prepaymentInvoiceItem, Invoice finalInvoice) {
```
awalkowiak, line 76: I **think** it should be just a clone of prepaymentInvoiceItem with modified quantity and amount according to needs

awalkowiak:
Did you try to run it locally? How was the performance?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
EWaterman: good point. Should quantity be 0 though? Looking at `PrepaymentInvoiceService.createInverseItemForCanceledOrderItem` I see it does `inverseItem.quantity = prepaymentItem.quantity`

https://github.com/openboxes/openboxes/blob/feature/OBPIH-6398-partial-invoices-for-prepaid-po/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy#L164

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,102 @@
...
+            finalInvoice.addToInvoiceItems(inverseItem)
+        }
+        return finalInvoice.save(failOnError: true, flush: true)
+    }
+
+    private InvoiceItem createInverseInvoiceItem(InvoiceItem prepaymentInvoiceItem) {
+        // Because we already computed the amount field for prepayment invoice items in step 1, and because for
+        // existing data there will always be a one-to-one mapping of prepayment invoice item to final invoice item,
+        // all we need to do to create the inverse item is copy the prepayment invoice item and inverse the amount.
+        InvoiceItem inverseItem = new InvoiceItem()
```
EWaterman, line 82: @awalkowiak I don't think this is the best way to clone the invoice item but I couldn't find much in the code about how we typically handle cloning domain objects (do we even do it?).

I'm thinking the best solution is a clone() method on the domain but it'd be nice if we just had some sort of default behaviour that we could wire into all our domain objects by default that will allow us to deep-clone things like relationships to other tables.

I didn't want to implement something like that before discussing it though (especially since GORM scares me) so I'm curious about your thoughts

EWaterman:
@awalkowiak I haven't been able to performance test this yet since I haven't tried setting my local up with prod data. Maybe we can chat tomorrow about what it takes to get that set up (and I can document the process for the future).

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -191,6 +191,39 @@ class InvoiceItem implements Serializable {
...
+            shipmentItem.addToInvoiceItems(clone)
+        }
+        for (OrderItem orderItem : orderItems) {
+            clone.addToOrderItems(orderItem)
+            orderItem.addToInvoiceItems(clone)
+        }
+        for (OrderAdjustment orderAdjustment : orderAdjustments) {
+            clone.addToOrderAdjustments(orderAdjustment)
+            orderAdjustment.addToInvoiceItems(clone)
+        }
```
alannadolny, line 223: Can't it be just `shipmentItem.setInvoicetems()` instead of looping over all items?

+ I am not sure how it works in Grails, but is there a need to copy the relation for both sides? If you set it only for the "owner" of the relation, it will also be set for the second object, but after closing the session? @kchelstowski is a hibernate master, so probably he should know more about that :smile:

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
```diff
@@ -0,0 +1,93 @@
...
+                    if (!orders || orders.size() != 1) {
+                        totalBadOrders++
+                        return
+                    }
+                    Order order = orders[0]
+
+                    // Find the regular invoice that we will be populating with the inverse items. For pre-exisitng
+                    // data, there is always only a single (final) invoice. If there is no regular invoice, or it
+                    // already has inverse items, skip it.
+                    List<Invoice> regularInvoices = order.invoices.findAll { it.isRegularInvoice}
```
alannadolny, line 67:
```suggestion
                    List<Invoice> regularInvoices = order.invoices.findAll { it.isRegularInvoice }
```

grails-app/migrations/0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy
```diff
@@ -0,0 +1,93 @@
...
+                    if (!regularInvoices) {
+                        totalOrdersNotYetFinalInvoiced++
+                        return
+                    }
+                    Invoice regularInvoice = regularInvoices[0]
+                    if (regularInvoices.size() > 1 || regularInvoice.invoiceItems.any { it.inverse}) {
```
alannadolny, line 73:
```suggestion
                    if (regularInvoices.size() > 1 || regularInvoice.invoiceItems.any { it.inverse }) {
```

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,83 @@
...
+     * feature was introduced.
+     */
+    void updateAmountForPrepaymentInvoiceItems() {
+        InvoiceType prepaymentInvoiceType = InvoiceType.findByCode(InvoiceTypeCode.PREPAYMENT_INVOICE)
+
+        Invoice.findAllByInvoiceType(prepaymentInvoiceType).each { Invoice prepaymentInvoice ->
+            // The amount field will be null for all pre-existing invoices since the field was not being used.
+            prepaymentInvoice.invoiceItems.find { it.amount == null }.each { InvoiceItem prepaymentInvoiceItem ->
+                prepaymentInvoiceItem.amount = computePrepaymentInvoiceItemAmount(prepaymentInvoiceItem)
+                prepaymentInvoiceItem.save(failOnError: true, flush: true)
```
alannadolny, line 30: Is this flush needed?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,83 @@
...
+     * This method assumes that because there is both a prepayment invoice and a final invoice, that the order
+     * associated with both invoices has been fully invoiced, and so there's no need to check each item of the
+     * prepayment invoice individually, they should ALL need to have inverse items created for them.
+     */
+    Invoice generateInverseInvoiceItems(Invoice prepaymentInvoice, Invoice finalInvoice) {
+        prepaymentInvoice.invoiceItems.each { InvoiceItem prepaymentInvoiceItem ->
+            InvoiceItem inverseItem = createInverseInvoiceItem(prepaymentInvoiceItem)
+            finalInvoice.addToInvoiceItems(inverseItem)
+        }
+        return finalInvoice.save(failOnError: true, flush: true)
```
alannadolny, line 69: and this flush?

src/test/groovy/unit/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationServiceSpec.groovy
```diff
@@ -0,0 +1,272 @@
...
+
+        where:
+        quantity | unitPrice | prepaymentPercent | adjustmentPercent | adjustmentAmount || expectedAmount
+        1        | 1.0       | 100               | 100               | 1                || 1
+        2        | 1.0       | 100               | 100               | 1                || 1
+        1        | 2.0       | 100               | 100               | 1                || 1
+        1        | 1.0       | 50                | 100               | 1                || 0.5
+        1        | 1.0       | 100               | 50                | 1                || 1
+        1        | 1.0       | null              | 100               | 1                || 1
+
```
alannadolny, line 197: Is there an empty line, or is it just odd GitHub formatting?

alannadolny (commented):
Maybe I am missing something, but is there a check to see if this migration was run? I mean it looks like it can take a lot of time to run on a bigger db, so maybe it will be beneficial to check if there is a need to run that before.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
```diff
@@ -0,0 +1,83 @@
...
+import org.pih.warehouse.order.OrderAdjustment
+import org.pih.warehouse.order.OrderItem
+import org.pih.warehouse.order.OrderItemStatusCode
+
+/**
+ * This service exists for the sole purpose of performing data migrations for changelog 0.9.x/2024-08-20-0000.
+ * The logic was only put in a service class so that it could be unit tested. Do not use it in regular flows.
+ */
+@Transactional
+class PrepaymentInvoiceMigrationService {
```
awalkowiak, line 15: I started wondering if it is a good idea to move this logic to a separate service, because we are removing the liquibase's complaints about checksum if someone modifies something in this service

awalkowiak (commented):
I think we need to have some extensive logging, which PO we are updating right now and how many there are still left. I would also wrap the update logic in some try-catch and log into the console the POs that failed to update (I am basing this idea on the Alan's recent issue with his local db)

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
awalkowiak: Plus I am getting more and more convinced that it should be a manual trigger from `data migration` page in app instead of a db migration ;/

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
EWaterman: good point. I'm also very curious about this :eyes:

src/test/groovy/unit/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationServiceSpec.groovy
EWaterman: Yeah I put empty lines to split up the logical groups (where adjustmentAmount is null vs not-null). I can probably add a comment here to explain why.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
EWaterman: I was thinking there'd be some GORM weirdness since we're updating these invoice items and then fetching them pretty soon after in step 2 of the migration but I'll test it out without the flush. Maybe GORM can handle the case gracefully. If not, maybe there's a way to just flush the cache at the end.

It's also making me wonder if GORM is batching these saves together or if we'll need to do it explicitly.

I think all this is highlighting the fact that I know very little about how GORM works behind the scenes...

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
EWaterman: this one is probably safe to not flush and just let GORM persist it as it would normally.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceMigrationService.groovy
EWaterman: that's a fair point, though I'd hope we wouldn't need to ever modify this file and if we did I don't see a world where we'd want to rerun it. The only reason it's in here is so that I could unit test it.

As to making this a manual trigger, yes we discussed it before and that's likely a safer option (that or just auto-migrate any invoices when they're saved if they haven't been migrated yet). We'd need to refactor the partial invoicing code to be able to support old data, but it'd mean we're not running this questionably long migration step on startup.

EWaterman:
@awalkowiak @alannadolny 

The change runs the full way through with my latest commit. Looks like it had something to do with the way the code was looping over the prepayment invoices. Maybe something to do with iterator vs loops, I'm not totally sure.

Here's the output I got on my local with stage data:
```
Prepayment Invoice migration completed in 12755 milliseconds!
761 prepayment invoices/orders were found.
694 orders were migrated.
6 were skipped due to no migration being needed
0 were skipped due to a bad order.
```
which is much faster than I thought it would be... I'm slightly suspicious and also wondering where the other 61 prepayment invoices are (I have a log statement for each case so I have no idea...)

I still have to check how the invoices are looking in the frontend (my local is behaving all wonky with the stage data).

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
kchelstowski: @alannadolny @EWaterman this version is fine, because we loop through the shipmentItems, not the invoiceItems, so for each SHIPMENT item we set the clone.
The same with the loops below.

EWaterman:
I poked around in the UI for some invoices that were migrated and all the ones I looked at seem to look correct to me. That said, I'm noticing the migration is not setting the amount field for any old non-prepay, non-inverse invoice items, would this be a problem?

I'd appreciate if someone else could run the migration locally to confirm it works for them as well. @awalkowiak @alannadolny 

Artur, I know you wanted to revisit the topic of whether we should not do this migration, so if you have that conversation Monday please loop me in (I'll be OOO),

image

EWaterman:
@alannadolny to answer an earlier question, I thought about it, but it's quite difficult to construct a reasonable precondition for these migrations. For step 1, we could technically add a check like:
```
    preconditions(onFail:'MARK_RAN') {
            sqlCheck(expectedResult: '0') {
                <count all invoice items where amount is null joined in invoice where invoice type is prepayment>
            }
    }
```
but that's doing essentially the same lookup that the migration is doing and so wouldn't really save us much. And step 2 of the migration is even harder to build a precondition for. Anyways, if there's nothing to migrate, the script will likely run fast enough since it's just doing a few selects, but it is a valid concern.

EWaterman:
@awalkowiak I'm going to remove the do not merge flag in case any decisions are made on Monday, but I'd like at the very least to have one more person test on their local first.

jmiranda (commented):
I'm going to try to address all of the points made in the PR and anything else I could think of.

1. I like the idea of moving the migration code to its own service. Even though it's only going to be executed once it's better to have the logic in one place. 
2. Whether we go with a liquibase changeset or manual data migration trigger is up to you guys. If we think there's even a remote chance that existing implementations will have prepayment invoices we should probably go with the former.
3. The one thing I wanted to see in the PR was whether we had logic to detect whether this migration had already been executed. Not a precondition, but the check mentioned in Line 78-79 where we skip invoices that already have inverse items. One word of caution on this. If you get to a place where you want to allow rerunning the migration (due to an error that might occur in the middle) you might want to change the logic to "does every invoice item have an inverse" so the process can be rerun over the invoice that error'd.
4. I wasn't ever really concerned about the performance of this because there's not much happening. I'm just glad that was actually the case.
5. Are we sure the service, executed via Bootstrap.groovy (during database migrations phase), is participating in a transaction? Can you run some tests on what happens when an exception occurs while processing an invoice within the loop? 
6. While not necessary, the one way to rollback the operation would be to remove all inverse invoice items for an invoice. In fact, you could set up the migration to execute ALWAYS and as long as the check logic was "do all invoice items have an invoice", you could safely execute the inverse operation.

EWaterman:
@jmiranda 

3. That's a fair concern. It might be somewhat complex to determine since we'd need to fetch all invoice items of an invoice at once and see if there's a matching inverse for each item, but maybe it's worth it to do that extra work just to be sure we don't ever get in a halfway state. If this is all working in a transaction, it shouldn't be an issue since it should get rolled back. I'll check.

5. Good question. I'll test.

6. I'm not sure it's safe to rollback by deleting inverse items. I don't think the system could handle that properly since it now expects the inverse items to be there. Maybe it'd be fine though and we can keep generating inverse items the old way if there aren't already inverses. Maybe @awalkowiak can comment on this.

EWaterman:
Okay I tested the code and it does seem to be properly transactional (I threw an error at the end of the migration and all amount fields were set back to null).

I believe this solves concerns 3 since if there's ever an error (there shouldn't be), the script will rollback all inverse items it created and so would be impossible to get stuck in a place where some of a prepayment invoice's items have an inverse and some don't.

As far as I'm concerned I think this can be merged once we've had a chance to retest against Manon's data on stage.

awalkowiak (approved):

PR #4800, opened 2024-08-22
OBPIH-6502 Add API endpoints to edit prepaid invoice items

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6502

**Description:**
 Add API endpoints to edit prepaid invoice items

---

- [X] Not Applicable

Will be tested with frontend changes

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -538,7 +538,7 @@ class UrlMappings {
             action = [POST: "updateItems"]
         }
 
-        "/api/prepaymentInvoiceItems/$id/" {
+        "/api/prepaymentInvoiceItems/$id" {
             controller = "prepaymentInvoiceItemApi"
             action = [POST: "update", DELETE: "delete"]
         }
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -533,9 +533,14 @@ class UrlMappings {
             action = [POST: "validateInvoiceItem"]
         }
 
+        "/api/prepaymentInvoices/$id/invoiceItems" {
+            controller = "prepaymentInvoiceApi"
+            action = [POST: "updateItems"]
+        }
+
         "/api/prepaymentInvoiceItems/$id/" {
             controller = "prepaymentInvoiceItemApi"
-            action = [DELETE: "delete"]
+            action = [POST: "update", DELETE: "delete"]
         }
 
         // Stock Transfer API
--- a/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceApiController.groovy
@@ -9,21 +9,18 @@
  **/
 package org.pih.warehouse.api
 
-import org.grails.web.json.JSONObject
+import org.grails.web.json.JSONArray
 
-class PrepaymentInvoiceItemApiController {
+class PrepaymentInvoiceApiController {
 
     def prepaymentInvoiceService
 
-    def delete() {
-        prepaymentInvoiceService.removeInvoiceItem(params.id)
-        render status: 204
-    }
-
-    def update() {
-        JSONObject jsonObject = request.JSON
-        Integer quantity = jsonObject.get('quantity')
-        prepaymentInvoiceService.updateInvoiceItemQuantity(params.id, quantity)
+    /**
+     * Bulk prepaid invoice items edit. Only can change quantity of given item
+     * */
+    def updateItems() {
+        JSONArray invoiceItems = request.JSON
+        prepaymentInvoiceService.updateItemsQuantity(params.id, invoiceItems)
         render status: 200
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceApiController.groovy
@@ -10,17 +10,18 @@
 package org.pih.warehouse.api
 
 import org.grails.web.json.JSONArray
+import org.pih.warehouse.invoice.PrepaymentInvoiceService
 
 class PrepaymentInvoiceApiController {
 
-    def prepaymentInvoiceService
+    PrepaymentInvoiceService prepaymentInvoiceService
 
     /**
      * Bulk prepaid invoice items edit. Only can change quantity of given item
      * */
     def updateItems() {
         JSONArray invoiceItems = request.JSON
-        prepaymentInvoiceService.updateItemsQuantity(params.id, invoiceItems)
+        prepaymentInvoiceService.updateItems(params.id, invoiceItems)
         render status: 200
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceItemApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceItemApiController.groovy
@@ -10,10 +10,11 @@
 package org.pih.warehouse.api
 
 import org.grails.web.json.JSONObject
+import org.pih.warehouse.invoice.PrepaymentInvoiceService
 
 class PrepaymentInvoiceItemApiController {
 
-    def prepaymentInvoiceService
+    PrepaymentInvoiceService prepaymentInvoiceService
 
     def delete() {
         prepaymentInvoiceService.removeInvoiceItem(params.id)
--- a/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
@@ -26,19 +26,17 @@ class PrepaymentInvoiceService {
             throw new Exception("No order items or order adjustments found for given order")
         }
 
-        BigDecimal prepaymentPercent = (order.paymentTerm?.prepaymentPercent ?: Constants.DEFAULT_PAYMENT_PERCENT) / 100
-
         order.activeOrderItems.each { OrderItem orderItem ->
             InvoiceItem invoiceItem = createFromOrderItem(orderItem)
             // This is prepayment item, we need to adjust the amount according to the payment terms
-            invoiceItem.amount = invoiceItem.amount * prepaymentPercent
+            invoiceItem.amount = invoiceItem.amount * order.prepaymentPercent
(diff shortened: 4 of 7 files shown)


grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -256,4 +253,67 @@ class PrepaymentInvoiceService {
...
+        InvoiceItem inverseItem = findInverseItem(invoiceItem)
+        if (inverseItem) {
+            OrderItem orderItem = invoiceItem?.shipmentItem?.orderItems?.find { it }
+            InvoiceItem prepaymentItem = orderItem.invoiceItems.find { it.isPrepaymentInvoice }
+            // find quantity that is still available to inverse and add to the current quantity from this inverse item
+            // this needs to be checked because we can also increase invoiced quantity here
+            Integer quantityAvailableToInverse = inverseItem.quantity + getQuantityAvailableToInverse(orderItem, prepaymentItem)
+            Integer quantityToInverse = quantity > quantityAvailableToInverse ? quantityAvailableToInverse : quantity
+            inverseItem.quantity = quantityToInverse
+            inverseItem.amount = quantityToInverse * prepaymentItem.unitPrice * orderItem.order.prepaymentPercent * (-1)
```
awalkowiak, line 308: @EWaterman I still remember you wanted to move it to separate method but I was not sure how to actually do it. Is something like that what you'd expect here:
```groovy
BigDecimal calculateAmount(Integer quantity, BigDecimal unitPrice, BigDecimal percentage = 1, Integer inverse = 1) {
    return quantity * unitPrice * percentage * inverse
}
```
and then just passing what we have available or need to use in calculation, or do you see here something a bit more complex in this method?

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -533,9 +533,14 @@ class UrlMappings {
             action = [POST: "validateInvoiceItem"]
         }
 
+        "/api/prepaymentInvoices/$id/invoiceItems" {
```
EWaterman, line 536: I'm not generally a fan of the camelcase URIs. I usually try to stick to a feature/sub-feature/entity-based system with something like: `/api/invoices/prepayment/$id/items` or `/api/prepayment-invoices/$id/invoice-items`. Up to you though

grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceApiController.groovy
```diff
@@ -0,0 +1,26 @@
...
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.api
+
+import org.grails.web.json.JSONArray
+
+class PrepaymentInvoiceApiController {
+
+    def prepaymentInvoiceService
```
EWaterman, line 16: in my opinion we should use proper types unless it's absolutely necessary to use def

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -533,9 +533,14 @@ class UrlMappings {
...
+            controller = "prepaymentInvoiceApi"
+            action = [POST: "updateItems"]
+        }
+
         "/api/prepaymentInvoiceItems/$id/" {
             controller = "prepaymentInvoiceItemApi"
-            action = [DELETE: "delete"]
+            action = [POST: "update", DELETE: "delete"]
```
EWaterman, line 543: If I'm doing an update only on the provided fields I usually use PATCH over POST (I think that's the generally accepted convention) but I'm not really bothered either way.

You do have a trailing slash in the URI which is probably best to remove for consistency's sake but up to you.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -256,4 +253,67 @@ class PrepaymentInvoiceService {
...
         return relatedObject?.invoiceItems?.find { InvoiceItem it -> it.inverse && it.invoice == invoice }
     }
+
+    /**
+     * Bulk editing invoice items quantity on final invoice for prepaid POs.
+     * We are allowing to edit only the quantity here. After editing invoice items quantity we have to
+     * change quantity on the corresponding inverse item.
+     *
+     * */
+    void updateItemsQuantity(String invoiceId, List items) {
```
EWaterman, line 263: It might be best to make this method more generic (ie call it updateItems instead of updateItemsQuantity). Right now we only allow you to update the quantity field but that could change in the future.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -256,4 +253,67 @@ class PrepaymentInvoiceService {
...
+        if (quantity <= 0) {
+            throw new IllegalArgumentException("Quantity to change needs to be higher than 0")
+        }
+        if (!invoiceItem?.shipmentItem) {
+            throw new IllegalArgumentException("Invoice item is missing shipment item. Cannot edit quantity.")
+        }
+        Integer quantityAvailableToInvoice = invoiceItem.shipmentItem.quantityToInvoice
+        if (quantity > quantityAvailableToInvoice + invoiceItem.quantity) {
+            throw new IllegalArgumentException("Cannot update quantity to higher value than available to invoice")
+        }
```
EWaterman, line 294: I asked a similar question in the other PR, but what about cancelled items? You shouldn't be able to update the invoice quantity right? Is that already handled by the quantityToInvoice field? Would it just be 0 in that case?

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
EWaterman: I think the thing I'm mainly worried about is that we calculate amount all over the place and it'd be nice to have a calculateAmountForPrepaymentItem(), calculateAmountForRegularItem() and calculateAmountForInverseItem() methods just to have it unified.

I haven't thought it through fully so this might not be the best idea, but they could maybe just take in the invoice item itself and do all the lookups they need so that the caller doesn't need to know what to do.

The logic here is just a bit tough to navigate with all the transients and chained calls so it'd be great if there was a simple method of calculating it that brought it all together

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -256,4 +253,67 @@ class PrepaymentInvoiceService {
...
+            Integer quantityAvailableToInverse = inverseItem.quantity + getQuantityAvailableToInverse(orderItem, prepaymentItem)
+            Integer quantityToInverse = quantity > quantityAvailableToInverse ? quantityAvailableToInverse : quantity
+            inverseItem.quantity = quantityToInverse
+            inverseItem.amount = quantityToInverse * prepaymentItem.unitPrice * orderItem.order.prepaymentPercent * (-1)
+            return
+        }
+
+        // If there is a case that we did not found inverse item, but it was made available to inverse now after
+        // editing or removing invoice items on invoices, lets try to create it here
+        inverseItem = createInverseItemForShipmentItem(invoiceItem?.shipmentItem, invoiceItem)
```
EWaterman, line 314: nitpick: I'd put this case in the if block (and change the condition to `if (!inverseItem)`) since it's much simpler and so the big block above doesn't need to be nested inside an if block.

grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceItemApiController.groovy
```diff
@@ -17,4 +19,11 @@ class PrepaymentInvoiceItemApiController {
         prepaymentInvoiceService.removeInvoiceItem(params.id)
         render status: 204
     }
+
+    def update() {
+        JSONObject jsonObject = request.JSON
+        Integer quantity = jsonObject.get('quantity')
+        prepaymentInvoiceService.updateInvoiceItemQuantity(params.id, quantity)
+        render status: 200
+    }
```
alannadolny, line 28: btw, from the frontend perspective, update/post methods should return the newly updated state, because we always have to refetch the new data. It's not a change request, because in this one case, on the frontend it's not necessary to fetch the data again, but in most cases, we should.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
alannadolny, line 294: Should those messages be translatable? I think it can be helpful for users.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
awalkowiak: We don't have a written convention for these yet (camel vs kebab case), hence I am following for now what we have in other cases (ex. `stockMovements`, `partialReceiving`, etc)

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: Ok, I think I am going to leave it as is for now, but I'll think about this as a standalone improvement in separate ticket/pr

grails-app/controllers/org/pih/warehouse/api/PrepaymentInvoiceItemApiController.groovy
awalkowiak: I agree, but here we have a bit more complex update, where we are passing one invoice item, but in reality we are updating two, so should we return both here or only the one that was sent in requests? Due to this, I decided to keep it simple and return only status, at least for now.

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -256,4 +255,79 @@ class PrepaymentInvoiceService {
...
+     * We are allowing to edit only the quantity here. After editing invoice items quantity we have to
+     * change quantity on the corresponding inverse item.
+     *
+     * */
+    void updateItems(String invoiceId, List items) {
+        def g = grailsApplication.mainContext.getBean('org.grails.plugins.web.taglib.ApplicationTagLib')
+        Invoice invoice = Invoice.get(invoiceId)
+        if (!invoice) {
+            String defaultMessage = "Cannot find invoice with id: ${invoiceId}"
+            throw new IllegalArgumentException(g.message(code: "invoice.cannotFind.label", args: [invoiceId], default: defaultMessage))
```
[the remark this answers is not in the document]
awalkowiak: I think we don't have clear convention

PR #4802, opened 2024-08-23
[OBPIH-6627] keep progress when reload on confirm page of full outbound import

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6627](https://pihemr.atlassian.net/browse/OBPIH-6627)

**Description:**
A requested feature for the _Full Outbound Import_ was to implement an ability for a user to refresh the page and not lose the filled-out form as we have in other workflows.

One big difference between full outbound import and other create stock movement workflows is that in the latter we save the data in the database and upon refreshing the page we full this created Pendning stock movement,but as for full outbound import, we are only saving it at the end so when we refresh the page we lose all the data.

To somehow save this filled-out data I decide to store it in the `sessionStorage` which is one of the browser storages that we can access.
I decided to use `sessionStorage` and `localStorage` because of the one main difference, sessionStorage expires once the tab is closed, unlike localStorage which has to be cleared manually. Since we are working with temporary data that we don't care to save for too long `sessionStorage` seemed like the approach to take. 
To manage sessionStorage I implemented a custom `useSessionStorage` hook. kind of mimicking the `useState`

Another big change was a slight modification to the `useWizzard` hook.
Since we are dealing with page refreshes, we need to know what step we are refreshing, If user opens `CONFIRM` step and there is cached data we want to render that step with prefilled cached data, and if user opens `DETAILS` page and there is available cached data we want to assume that user wants to do a new import and we should not load cached data.

So I modified the `useWizzard` hook so we switch between steps based on the URL query parameter `step` instead of a value held in `state`

---

_- No screenshots -_

---

- [X] Not Applicable

---

[OBPIH-6627]: https://pihemr.atlassian.net/browse/OBPIH-6627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alannadolny (approved):

src/js/hooks/useSessionStorage.js
```diff
@@ -0,0 +1,9 @@
+const useSessionStorage = (key, defaultValue) => {
+  const val = JSON.parse(sessionStorage.getItem(key)) ?? defaultValue;
+  const setValue = (newValue) => sessionStorage.setItem(key, JSON.stringify(newValue));
+  const clearValue = () => sessionStorage.removeItem(key);
+
+  return [val, setValue, clearValue];
```
kchelstowski, line 6: wouldn't it be better to return it as object, so we would be able to get rid of the confusing:
```js
const [cachedData,]
```
it would get even more weird if we wanted to use the first and the third property:
```js
const [cachedData, null, clearValue]
```

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
```diff
@@ -70,6 +65,13 @@ const OutboundImport = () => {
     trigger,
   } = useOutboundImportForm({ next });
 
+  /** Redirect to first step if there is no cached data */
+  useEffect(() => {
+    if (_.isEmpty(cachedData) && !is(OutboundImportStep.DETAILS)) {
+      navigateToStep(OutboundImportStep.DETAILS);
+    }
+  }, []);
```
kchelstowski, line 73: let's move it to a separate small hook named e.g. useCachedData

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
```diff
@@ -78,15 +80,27 @@ const OutboundImport = () => {
...
+   * where the required XLS file is lost. The file was already validated earlier,
+   * so we bypass validation and submit the form directly to retain progress.
+   *
+   * Related ticket OBPIH-6627 (Keep filled form progress when refreshing the page)
+   */
+  const handleConfirmSubmitForm = (submitMethod) => (event) => {
+    event.preventDefault();
+    submitMethod(getValues());
+  };
+
```
kchelstowski, line 94: I kinda do not get this method. Why do we want to skip validation? And which validation exactly? Is it the frontend validation of the file from previous step? Even if so, how do you distinguish here whether the submit happens from the refreshed page or from regular usage of the workflow?

src/js/components/stock-movement-wizard/outboundImport/OutboundImportStep.js
```diff
@@ -0,0 +1,6 @@
+const OutboundImportStep = {
+  DETAILS: 'DETAILS',
+  CONFIRM: 'CONFIRM',
+};
+
+export default OutboundImportStep;
```
kchelstowski, line 6: let's move it to `consts` directory, where we place all the enums.

src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -10,38 +10,69 @@ import { useSelector } from 'react-redux';
...
+    };
+
+    /** Load cached data:
+     * If we have saved cached data in the session-storage
+     * we want to load cached data only on CONFIRM step,
+     * otherwise we clear the session-storage data since that means that user has abandoned the form
+     * and that cached data is not relevant anymore
+    * */
+    if (!_.isEmpty(cachedData)) {
+      if (OutboundImportStep.CONFIRM === queryParams?.step) {
```
kchelstowski, line 59: I will probably get to this below, but why do we suddenly need to store the step in the query params, when we should have access to the current step from the `useWizard` hook?

src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -10,38 +10,69 @@ import { useSelector } from 'react-redux';
...
+        const { fulfillmentDetails, sendingOptions } = cachedData;
+        values.description = fulfillmentDetails?.description;
+        values.destination = fulfillmentDetails?.destination;
+        values.requestedBy = fulfillmentDetails?.requestedBy;
+        values.dateRequested = fulfillmentDetails?.dateRequested;
+        values.shipmentType = sendingOptions?.shipmentType;
+        values.trackingNumber = sendingOptions?.trackingNumber;
+        values.comments = sendingOptions?.comments;
+        values.expectedDeliveryDate = sendingOptions?.expectedDeliveryDate;
```
kchelstowski, line 68: what about packing list?

src/js/hooks/useSessionStorage.js
drodzewicz: I modeled this `useSessionStorage` hook after `useState` where we destruct the value and setter from an array and accessing nth value like `[,,clearFunction] = useSessionStorage()` is rather common and is aligned with react conventions.

I don't feel like one solution is better than the other, it's what we prefer, both are correct.

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
drodzewicz: values from `localStorage` and `sessionStorage` are not related to state and will not trigger the component rerender even if you put the in the dependency array.
Also these values are synchronous unlike useEffect which is asynchronous which means that these values will be available in the hook without a delay.

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
drodzewicz: The problem is that I don't cache the uploaded file `packingList` which is a required value in the validation schema,
So after loading all of the values from `sessionStorage` I am missing the `packingList` File which prevents me from submitting the form.

I figured that there are no more fields that we need to fill out on the CONFIRM step and all of the validation should have happened in the previous step so we can assume that the data that we have on confirm step is correct and as a workaround we can skip the validation.

src/js/hooks/outboundImport/useOutboundImportForm.js
drodzewicz: since we are dealing with page refreshes I needed know what page are we reloading, if we were to use just state for storing current step it would be cleared after the refresh and we wouldn't know if we need to load the cached data (when we are on CONFIRM step) or not load it when we are for example on DETAILS step

src/js/hooks/outboundImport/useOutboundImportForm.js
```diff
@@ -10,38 +10,69 @@ import { useSelector } from 'react-redux';
...
+      } else {
+        clearCachedData();
```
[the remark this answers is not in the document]
drodzewicz: yes, that's correct
Unlike refreshing on the CONFIRM step I believe there is no case a user would want to refresh this step and instead should be treated as refreshing and clearing the whole form

src/js/hooks/useWizard.js
```diff
@@ -1,54 +1,78 @@
...
+  };
+
+  /** Compute current active wizard step
+   * Determine current wizard step based on URl query "step" parameter
+   * Otherwise use provided initial key
+   * Or default to the first step
+   */
+  const currentStepKey = useMemo(() =>
+    parsedQueryParams.step || initialKey || steps[0]?.key,
+  [parsedQueryParams.step, steps, initialKey]);
```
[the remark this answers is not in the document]
drodzewicz: we are sure

src/js/hooks/useSessionStorage.js
kchelstowski: ok, for me it just looks weird when we have a comma without anything after that, but if you feel like it's better, then I'm fine, it was not a change request

src/js/components/stock-movement-wizard/outboundImport/OutboundImport.jsx
kchelstowski: Ok, I supposed this was the case, thanks for explaining

src/js/hooks/outboundImport/useOutboundImportForm.js
kchelstowski: thanks for explaining, I forgot to delete this comment after I figured out below what the case is.

src/js/hooks/outboundImport/useOutboundImportForm.js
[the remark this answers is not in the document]
kchelstowski: Has it been confirmed with Kasia that this should be the scenario?

kchelstowski (approved):

PR #4803, opened 2024-08-26
OBPIH-6500 [frontend] Modifications to the invoice confirmation page (fix after QA)

alannadolny:

**Description:**

The reopening was caused by duplicating some invoice items / missing some invoice items. Pagination on the frontend fetched items in parts:
0-10 items
10-20 items
20-30 items
....
Each data fetching was in a different order, so each request can duplicate items.

### DIFF
--- a/grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
@@ -101,11 +101,13 @@ class InvoiceService {
             invoiceItems = InvoiceItem.createCriteria().list(max: max.toInteger(), offset: offset.toInteger()) {
                 eq("invoice", invoice)
                 order("inverse", "asc")
+                order("id", "asc")
             }
         } else {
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/invoice/InvoiceService.groovy
```diff
@@ -101,11 +101,13 @@ class InvoiceService {
             invoiceItems = InvoiceItem.createCriteria().list(max: max.toInteger(), offset: offset.toInteger()) {
                 eq("invoice", invoice)
                 order("inverse", "asc")
+                order("id", "asc")
```
awalkowiak, line 104: Perhaps sorting by product code here would be better?

alannadolny: No, we can invoice a few products with the same product code and the same issue will be visible.

kchelstowski (approved):

drodzewicz (approved):

PR #4805, opened 2024-08-28
OBPIH-6532 [frontend] Ability to edit invoice item (fixes after QA)

alannadolny:
In this PR I fixed:
- Blocking the "post invoice" button when the invoice contains canceled items
- Fix error while clicking exit in empty invoices
- Fix calculating invoice value
- Fix the UI issue with displaying action menu


package.json
```diff
@@ -104,6 +104,7 @@
     "react-table": "6.8.6",
     "react-table-hoc-fixed-columns": "2.3.4",
     "react-tabs": "2.3.1",
+    "react-tiny-popover": "6.0.5",
```
EWaterman, line 107: Can you summarize what this library is giving us over our previous way of handling pop-up messages? Do you think we should be using this approach for all our pop-up messages going forward?

awalkowiak (approved):

package.json
alannadolny: This library is a replacement for our component called "Action Dots":
image
Previously we created it using Bootstrap, but when this component is placed within the table (basically that's all of our use cases) we had to use a workaround - we didn't let the popup touch the header (because the popup is not visible outside the table) or anything else (the table has fixed row size, so there was always enough space for that). But in the invoices workflow, we can have a table with only one item, and then the popups are not visible. Fixing it with styling is a really hard challenge, and in this case, I was not able to use portals, because the popup is handled by Bootstrap in a different way and there is an error that the portal can't find a defined container. So yeah, from now we should use the component created by me for displaying dropdowns with a menu. For any other popups, I think it depends on the case, but I prefer using this library to be consistent across the whole app.

drodzewicz (approved):

src/js/utils/ContextMenu.jsx
```diff
@@ -0,0 +1,117 @@
...
+                  && <Translate id={action.label} defaultMessage={action.defaultLabel} />}
+              </>
+            );
+            const elementType = getActionItemType(action);
+            const link = elementType === actionItemType.LINK || actionItemType.REACT_LINK
+              ? buildLink(action, id)
+              : '';
+
+            return (
+              <React.Fragment key={action.label}>
```
kchelstowski, line 54: fyi: we can now use the shortened version: `<> </>`

kchelstowski (commented):
lgtm, nice research about the library

PR #4807, opened 2024-08-28
OBPIH-6633 Prevent Browser users from creating full outbound via import

alannadolny:


drodzewicz (approved):

awalkowiak (approved):

kchelstowski (commented):
lgtm, but I'm wondering why we don't hide the full outbound import in the menu for such users in the same way we do it for e.g. invoices, when you are not an invoice user.

PR #4813, opened 2024-08-30
OBPIH-6513+OBPIH-6514 Indicator for items requiring attention + include file row…

kchelstowski:
… number in the table

### DIFF
--- a/src/js/hooks/outboundImport/useOutboundImportFiltering.js
+++ b/src/js/hooks/outboundImport/useOutboundImportFiltering.js
@@ -1,0 +1,12 @@
+import { useState } from 'react';
+
+const useOutboundImportFiltering = () => {
+  const [isFiltered, setIsFiltered] = useState(false);
+
+  return {
+    setIsFiltered,
+    isFiltered,
+  };
+};
+
+export default useOutboundImportFiltering;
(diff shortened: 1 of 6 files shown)


src/js/hooks/outboundImport/useOutboundImportFiltering.js
```diff
@@ -0,0 +1,12 @@
+import { useState } from 'react';
+
+const useOutboundImportFiltering = () => {
+  const [isFiltered, setIsFiltered] = useState(false);
+
+  return {
+    setIsFiltered,
+    isFiltered,
+  };
+};
```
alannadolny, line 10: Maybe it's worth moving to this hook more logic related to filtering? Generally speaking, having a hook with just one setState is not worth creating this hook :smile: 
I think you can move to this hook actions like:
```
setIsFiltered((prev) => !prev);
```
and call it toggleFiltering
or things like that:
```
isFiltered
          ? (itemsWithErrors ?? [])
          : (itemsInOrder ?? []
```

drodzewicz (approved):

awalkowiak (approved):

src/js/hooks/outboundImport/useOutboundImportFiltering.js
awalkowiak: Other than this set state hook, lgtm

alannadolny (approved):

PR #4820, opened 2024-09-03
OBPIH-6644 Fix calculating inverse item after reducing ordered quantity

awalkowiak:
https://pihemr.atlassian.net/browse/OBPIH-6644

**Description:**
When an order item had quantity reduced after making a prepayment, then it was not fully inversed. There was a need for a fix while calculating inverse quantity to take a look if the ordered item is both fully shipped and fully invoiced, and then add inverse item with quantity available to inverse.

---
- [ ] Not Applicable

According to reproduction steps

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
@@ -182,7 +182,13 @@ class PrepaymentInvoiceService {
             return null
         }
         InvoiceItem inverseItem = createFromShipmentItem(shipmentItem)
-        Integer quantity = invoiceItem.quantity >= quantityInverseable ? quantityInverseable : invoiceItem.quantity
+        Integer quantity
+        if (orderItem.isCompletelyFulfilled() && orderItem.isFullyInvoiced()) {
+            // If quantity if fully shipped and fully invoiced set full inverse quantity
+            quantity = quantityInverseable
+        } else {
+            quantity = invoiceItem.quantity >= quantityInverseable ? quantityInverseable : invoiceItem.quantity
+        }
         inverseItem.quantity = quantity
         inverseItem.inverse = true
         inverseItem.unitPrice = prepaymentItem.unitPrice


kchelstowski (approved):

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -182,7 +182,13 @@ class PrepaymentInvoiceService {
...
         }
         InvoiceItem inverseItem = createFromShipmentItem(shipmentItem)
-        Integer quantity = invoiceItem.quantity >= quantityInverseable ? quantityInverseable : invoiceItem.quantity
+        Integer quantity
+        if (orderItem.isCompletelyFulfilled() && orderItem.isFullyInvoiced()) {
+            // If quantity if fully shipped and fully invoiced set full inverse quantity
+            quantity = quantityInverseable
+        } else {
+            quantity = invoiceItem.quantity >= quantityInverseable ? quantityInverseable : invoiceItem.quantity
+        }
```
alannadolny, line 191: In the code above, we use returns instead of else, so maybe to keep the code consistent, it would be better to use the same approach here?

+ What do you think about writing some simple unit tests? :sadsmile:

awalkowiak: In the code above we return nulls and we don't want to execute code that is below. Here we want to determine the quantity that is going to be used in the latter part of the code

awalkowiak: > + What do you think about writing some simple unit tests? :sadsmile:

When it is QA'ed we can think about bumping up test coverage in entire partial invoice feature

alannadolny: > When it is QA'ed we can think about bumping up test coverage in entire partial invoice feature

good idea, but we can't forget about that as always

PR #4822, opened 2024-09-03
OBPIH-6644 Fix calculating inverse item after reducing ordered quantity (invoice item edit)

awalkowiak:
https://pihemr.atlassian.net/browse/OBPIH-6644

**Description:**
This is an improvement to my previous PR (https://github.com/openboxes/openboxes/pull/4820). I forgot about invoice item editing on confirm page...

---

- [ ] Not Applicable

According to the test plan

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
@@ -182,13 +182,7 @@ class PrepaymentInvoiceService {
             return null
         }
         InvoiceItem inverseItem = createFromShipmentItem(shipmentItem)
-        Integer quantity
-        if (orderItem.isCompletelyFulfilled() && orderItem.isFullyInvoiced()) {
-            // If quantity if fully shipped and fully invoiced set full inverse quantity
-            quantity = quantityInverseable
-        } else {
-            quantity = invoiceItem.quantity >= quantityInverseable ? quantityInverseable : invoiceItem.quantity
-        }
+        Integer quantity = getQuantityToInverse(orderItem, invoiceItem.quantity, quantityInverseable)
         inverseItem.quantity = quantity
         inverseItem.inverse = true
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -349,4 +343,13 @@ class PrepaymentInvoiceService {
...
     }
+
+    Integer getQuantityToInverse(OrderItem orderItem, Integer quantityInvoiced, Integer quantityInverseable) {
+        if (orderItem.isCompletelyFulfilled() && orderItem.isFullyInvoiced()) {
+            // If quantity if fully shipped and fully invoiced set full inverse quantity
+            return quantityInverseable
+        }
+
+        return quantityInvoiced >= quantityInverseable ? quantityInverseable : quantityInvoiced
+    }
```
alannadolny, line 354: looks like a method that should be placed in the domain, what do you think @awalkowiak?

kchelstowski: I'd say it's difficult to determine in which domain it should be placed, because in most of those we operate on an order item, so it would be weird to place it in the `InvoiceItem` domain, especially that all the values passed in this method are not transients, but some values calculated in some other methods, e.g. `getQuantityToInverse` that is already placed in this service a few lines above.

I will let you guys debate on that and make a final decision

kchelstowski (approved):

drodzewicz (approved):

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: Technically it potentially could be on the order item, but since it did not rely only on the data that could be pulled from the order item or prepayment item easily (it relies also on the specific invoice item quantity, currently created or edited) I think it is more suiting here.

PR #4824, opened 2024-09-03
OBPIH-6636 Trigger order summary refresh after editing existing item quantity

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6636

**Description:**
The order summary materialized view was not refreshing when the order items quantity was edited.

---

- [ ] Not Applicable

According to the reproduction steps

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
@@ -766,6 +766,10 @@ class OrderController {
             if (!orderService.isOrderEditable(orderItem.order, session.user)) {
                 throw new UnsupportedOperationException("${warehouse.message(code: 'errors.noPermissions.label')}")
             }
+            if (params.quantity && orderItem.quantity != (params.quantity as Integer)) {
+                // if existing item's quantity is edited we have to trigger the order summary refresh
+                orderItem.disableRefresh = false
+            }
             orderItem.properties = params
             orderItem.refreshPendingShipmentItemRecipients()
         }


grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -766,6 +766,10 @@ class OrderController {
             if (!orderService.isOrderEditable(orderItem.order, session.user)) {
                 throw new UnsupportedOperationException("${warehouse.message(code: 'errors.noPermissions.label')}")
             }
+            if (params.quantity && orderItem.quantity != (params.quantity as Integer)) {
```
alannadolny, line 769: maybe use `params.int('quantity')`?

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
```diff
@@ -766,6 +766,10 @@ class OrderController {
...
+                // if existing item's quantity is edited we have to trigger the order summary refresh
+                orderItem.disableRefresh = false
```
EWaterman, line 771: can you explain why this is needed? Instinctively, I'd assume that if the orderItem's quantity is updated, the view would know that it'd need to be refreshed because a field is changing. Why does quantity need a special check here? If other params (ie not quantity) are updated, does the view get refreshed?

awalkowiak: > the view would know that it'd need to be refreshed because a field is changing.

That's correct for the view, but we have a materialized view for storing order summary (effectively a table in db).
Refreshing order summary for order **items** changes is disabled by default. It's refreshed only when a new item is added to the order. Item's quantity edit here did not trigger refresh on the order itself. In a result, the view has/calculates proper values, but the materialized view has old data, hence need to trigger this one at this level.

> Why does quantity need a special check here? If other params (ie not quantity) are updated, does the view get refreshed?

Only quantity matters for the order summary. Hence no need to trigger it after editing other properties

EWaterman (approved):

grails-app/controllers/org/pih/warehouse/order/OrderController.groovy
kchelstowski, line 769: titbit: you could use `orderItem.isDirty('quantity')`

kchelstowski (approved):

PR #4831, opened 2024-09-09
OBPIH-6668 Responsiveness in very long shipments with split line appears and disappears

alannadolny:

**Link to GitHub issue or Jira ticket:** [OBPIH-6668](https://pihemr.atlassian.net/browse/OBPIH-6668?atlOrigin=eyJpIjoiOTE3NjZjZDAxNTg2NGI1YWJjYTYxZTFiMmY5OGJhYmIiLCJwIjoiaiJ9)

**Description:**
The issue with disappearing new lines was caused by managing the state in the component and in the form state at the same time, so when we were fetching new lines, we overrode the already created new lines.

[OBPIH-6668]: https://pihemr.atlassian.net/browse/OBPIH-6668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
@@ -338,7 +338,6 @@ class AddItemsPage extends Component {
   setLineItems({
     response,
     startIndex,
-    fetchedLineItems,
     append,
   }) {
     const { data } = response.data;
@@ -360,7 +359,6 @@ class AddItemsPage extends Component {
 
     this.loadMoreRows(append)({
       startIndex: startIndex + this.props.pageSize,
-      fetchedLineItems: fetchedLineItems + lineItemsData.length,
     });
 
     this.props.hideSpinner();
@@ -589,7 +587,7 @@ class AddItemsPage extends Component {
 
   // At this moment we are fetching the whole stock movement at once (it already includes stock movement items)
   // So this function is not used at this moment, but can be helpful in case of future improvements
-  loadMoreRows = (append) => ({ startIndex, fetchedLineItems }) => {
+  loadMoreRows = (append) => ({ startIndex }) => {
     this.setState({
       isFirstPageLoaded: true,
     });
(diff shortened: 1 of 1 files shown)


src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -350,29 +335,35 @@ class AddItemsPage extends Component {
...
+    append?.('lineItems', lineItemsData);
+
+    if (data.length < 10) {
       this.props.hideSpinner();
+      return;
+    }
+
+    this.loadMoreRows(append)({
+      startIndex: startIndex + this.props.pageSize,
+      fetchedLineItems: fetchedLineItems + lineItemsData.length,
```
kchelstowski, line 363: at first thought I would think that `fetchedLineItems` actually store the items, not the size of the items, so how about renaming it to `fetchedItemsSize`/`fetchedLineItemsSize`/`fetchedLineItemsCount`?

kchelstowski (commented):
as a side note: in some places it's a bit difficult to spot, what is actually changed and what are just the lint fixes, so I would not call the fix eslint on a whole file, or at least indicate in a PR what lines are the "real" changes" and which could be skipped.

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
alannadolny: > changed and what are just the lint fixes, so I would not call the fix eslint on a whole file
btw, I think we should run the linter for the entire project some time ago.

alannadolny:
``` 
fetchedLineItems: fetchedLineItems + lineItemsData.length,
```
I noticed that it's not used anymore, so I am deleting it :smile:

awalkowiak (approved):

PR #4832, opened 2024-09-09
OBPIH-6671 Fix amount outstanding on orders report calculation

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6671

**Description:**
There was an issue with calculating the report when there were multiple (partial) invoice items per shipment item (the quantity shipped was unnecessarily duplicated). Fixed it by grouping rows by shipment item and summing the quantity invoiced.

### DIFF
--- a/grails-app/services/org/pih/warehouse/report/ReportService.groovy
+++ b/grails-app/services/org/pih/warehouse/report/ReportService.groovy
@@ -897,10 +897,7 @@ class ReportService implements ApplicationContextAware {
                         WHEN shipment.current_status IN ('SHIPPED', 'PARTIALLY_RECEIVED', 'RECEIVED') THEN IFNULL(shipment_item.quantity / order_item.quantity_per_uom, 0)
                         ELSE 0
                     END AS quantity_shipped,
-                    CASE 
-                        WHEN invoice.date_posted IS NULL THEN 0
-                        ELSE IFNULL(SUM(invoice_item.quantity), 0) 
-                    END AS quantity_invoiced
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -914,7 +911,7 @@ class ReportService implements ApplicationContextAware {
                     AND (invoice.invoice_type_id != :prepaymentInvoiceId OR invoice.invoice_type_id IS NULL)
                     AND (invoice_item.inverse IS NULL OR invoice_item.inverse = FALSE)
                     ${additionalFilter}
-                GROUP BY o.id, order_item.id, shipment_item.id, invoice_item.id
+                GROUP BY o.id, order_item.id, shipment_item.id
```
EWaterman, line 914: why were we grouping by invoice item before? It seems like doing that would only ever result in one line per grouping and so make the SUM in line 902 pointless. What you've done here looks correct. I suppose the only reason it was working before was that an order item only ever had one invoice item so it just happened to work?

EWaterman (approved):

grails-app/services/org/pih/warehouse/report/ReportService.groovy
awalkowiak: Yeah I am not sure. IMHO it should not have been done since the beginning, and I am not sure why it was done. I think the regular invoices (not prepaid) were broken because of this before, but this bug was hard to spot.

kchelstowski (approved):

alannadolny (approved):

PR #4835, opened 2024-09-10
OBPIH-6675 Errors cached despite leaving outbound import workflow

kchelstowski:

**Link to GitHub issue or Jira ticket: https://pihemr.atlassian.net/browse/OBPIH-6675?atlOrigin=eyJpIjoiZDA1OTBlNmVjN2U5NDk5ZmFhOGY1YWE3MmFkMmYwOTMiLCJwIjoiaiJ9**

**Description:**

The problem here was that the cached data was cleared, but two `useEffects` were affecting themselves, hence before they were cleared, the cached data was sent for validation, as it was not empty yet.
To fix that, we want to load the cached data only if we are on the confirm step.

### DIFF
--- a/src/js/hooks/outboundImport/useOutboundImportForm.js
+++ b/src/js/hooks/outboundImport/useOutboundImportForm.js
@@ -267,8 +267,16 @@ const useOutboundImportForm = ({ next }) => {
     }
   };
 
+  const handleLoadCachedData = () => {
+    if (OutboundImportStep.CONFIRM === queryParams?.step) {
+      loadCachedData();
+      return;
+    }
+    clearCachedData();
+  };
+
   useEffect(() => {
-    loadCachedData();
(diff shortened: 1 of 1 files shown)


src/js/hooks/outboundImport/useOutboundImportForm.js
drodzewicz, line 1: Good job!
Could you also remove the 
```js
...
else {
  clearCachedData();
}
```
from the `useMemo` for default values

First I think it is unnecessary to call it there anymore since now we are calling `clearCahcedData` in the `useEffect` hook (what you have impement)
And also it is probably wrong to call any methods with side-effect in a `useMemo` hook (my bad 🙈 )

drodzewicz (changes requested):

drodzewicz (approved):

PR #4842, opened 2024-09-16
OBPIH-6690 Add option to edit unit price on invoice items made from order adjustments

alannadolny:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6690

**Description:**
I adjusted the hook to be able to edit the unit price of order adjustments.

---

image

### DIFF
--- a/src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
+++ b/src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
@@ -11,7 +11,7 @@ import useTranslate from 'hooks/useTranslate';
 
 const useInvoicePrepaidItemsTable = ({
   invoiceItems,
-  updateInvoiceItemQuantity,
+  updateInvoiceItemData,
   invoiceId,
   refetchData,
 }) => {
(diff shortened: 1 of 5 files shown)


src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
```diff
@@ -118,10 +134,35 @@ const useInvoicePrepaidItemsTable = ({
...
 
-  const updateRowQuantity = (rowId, quantity) =>
+  // validation for order adjustments
+  const validateUnitPrice = (row) => {
+    if (row?.unitPrice === 0 || Math.abs(row?.unitPrice) > Math.abs(row?.amount)) {
+      setInvalidRows((rows) => ([...rows, row?.id]));
+      return translate('react.invoice.errors.unitPrice.label', 'Wrong amount to invoice value');
+    }
+
+    setInvalidRows((rows) => rows.filter((rowId) => row.id !== rowId));
```
kchelstowski, line 144: this feels a bit weird - why can't we just update the invalid rows when there is one?
In other words:

- if there are validation errors, append the invalid item to the list

currently it looks like all of the rows are marked as invalid, and if there is not validation error for a row, we filter it out from the invalid rows list :thinking: ?

kchelstowski (approved):

src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
alannadolny: If the row is invalid we are adding it to the `invalidRows`:
```
setInvalidRows((rows) => ([...rows, row?.id]));
```
The line that you are pointed removes the line from `setInvalidRows` if there is no error (the user already changed the value)

drodzewicz (approved):

PR #4846, opened 2024-09-18
OBPIH-6674 Unable to update bc and unit price through PO import when other items inluced in invoice

alannadolny:


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -869,7 +869,13 @@ class OrderService {
                     if (currentLocation.isAccountingRequired() && !code) {
                         throw new IllegalArgumentException("Budget code is required.")
                     }
-                    BudgetCode budgetCode = BudgetCode.findByCode(code)
+
+                    // There can be more than one budget code with the same code, despite the fact that the code is unique from the domain perspective.
+                    // As a fix for that we would like to have only one active budget code with the same code, so we have to find only active
+                    // budget codes. In case when there is only one budget code, and this one is inactive we have to throw a validation error,
```
kchelstowski, line 875: if you say so, why don't you throw this exception if the `foundBudgetCodes.first()` would be inactive?

kchelstowski (commented):
I think we should revisit as soon as possible our ddl and add the constraints to the database. This is probably the reason why the duplicate exists - it must've been updated/inserted through the DB. I can't see any unique constraint on the `budget_code` table.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
alannadolny: It's threw a few lines below

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -869,7 +869,13 @@ class OrderService {
...
+                    // so we can't just look for BudgetCode.findAllByCodeAndActive(code, true);
+                    List<BudgetCode> foundBudgetCodes = BudgetCode.findAllByCode(code)
+                    BudgetCode budgetCode = foundBudgetCodes.size() > 1 ? foundBudgetCodes.find { it.active } : foundBudgetCodes.first()
```
awalkowiak, line 878: If we have a case with multiple active, then we still might have the issue with accidentally updating the budget code on the item, perhaps it would be worth throwing an exception if we have multiple active budget codes with the same code?

EWaterman:
I agree with Kacper that I'd like to prioritize a proper fix for this.

I was curious to know how many cases of duplicate code that we have in prod:
```
select code, count(*) from budget_code group by code having count(*) > 1;
+------------+----------+
| code       | count(*) |
+------------+----------+
| HAUNR.0380 |        2 |
| PH0650.022 |        2 |
| SI0385.055 |        2 |
| SI0716.001 |        2 |
+------------+----------+
4 rows in set (0.211 sec)
```
and the offenders in detail
```
+----------------------------------+------------+----------------------------------------------------+----------------------------------------------------+-----------------+---------------------+---------------------+----------------------------------+----------------------------------+--------+
| id                               | code       | name                                               | description                                        | organization_id | date_created        | last_updated        | created_by_id                    | updated_by_id                    | active |
+----------------------------------+------------+----------------------------------------------------+----------------------------------------------------+-----------------+---------------------+---------------------+----------------------------------+----------------------------------+--------+
|  SI0716.001                      | SI0716.001 | Meds and Plumpy Nuts                               | Meds and Plumpy Nuts                               | NULL            | 2021-01-15 00:00:00 | 2021-01-15 00:00:00 | 8a8a9e9666194c89016635c637370b9b | 8a8a9e9666194c89016635c637370b9b | true   |
| 09f5e1167944cbae017aaac49b965793 | SI0385.055 | GAC- ANC                                           | GAC- ANC                                           | NULL            | 2021-07-15 15:24:09 | 2021-07-15 15:24:27 | 8a8a9e9666194c8901663a87e9a80b9c | 8a8a9e9666194c8901663a87e9a80b9c | true   |
| 09f5e1167b98d5fd017b9c867f1c0a35 | HAUNR.0380 | HUM Facility Maintenance - Waste Management        | HUM Facility Maintenance - Waste Management        | NULL            | 2021-08-31 14:04:25 | 2024-09-09 17:01:27 | 09f5e1167751da090177b70696d04f2d | 8a8a9e9665c4f59d0165c5101f6d0001 | false  |
| 09f5e1167b98d5fd017ba21c0a4f3da6 | PH0650.022 | Other - Immokalee                                  | Other - Immokalee                                  | NULL            | 2021-09-01 16:05:52 | 2024-09-09 17:02:02 | 09f5e1167751da090177b70696d04f2d | 8a8a9e9665c4f59d0165c5101f6d0001 | false  |
| HAUNR.0380                       | HAUNR.0380 | HUM Facility Maintenance - Waste Management        | HUM Facility Maintenance - Waste Management        | NULL            | 2025-01-04 00:00:00 | 2025-01-04 00:00:00 | 8a8a9e9666194c89016635c637370b9b | 8a8a9e9666194c89016635c637370b9b | true   |
| PH0650.022                       | PH0650.022 | Other - Immokalee                                  | Other - Immokalee                                  | NULL            | 2021-01-15 00:00:00 | 2021-01-15 00:00:00 | 8a8a9e9666194c89016635c637370b9b | 8a8a9e9666194c89016635c637370b9b | false  |
| SI0385.055                       | SI0385.055 | Social Support for Vulnerable patient and children | Social Support for Vulnerable patient and children | NULL            | 2025-01-04 00:00:00 | 2025-01-04 00:00:00 | 8a8a9e9666194c89016635c637370b9b | 8a8a9e9666194c89016635c637370b9b | true   |
| SI0716.001                       | SI0716.001 | Meds and Plumpy Nuts                               | Meds and Plumpy Nuts                               | NULL            | 2025-01-04 00:00:00 | 2025-01-04 00:00:00 | 8a8a9e9666194c89016635c637370b9b | 8a8a9e9666194c89016635c637370b9b | true   |
+----------------------------------+------------+----------------------------------------------------+----------------------------------------------------+-----------------+---------------------+---------------------+----------------------------------+----------------------------------+--------+
```
with only `SI0716.001` and `SI0385.055` having multiple active.

Given that it seems to be a very rare occurrence maybe we can fix what we have on prod manually (though we'd still potentially need a migration to fix any non-PIH environments).

While I think we should add unique constraint, I don't see a simple way to fix existing data via migration given that we'd need a way to pick which code should be the active one, which isn't immediately obvious from just looking at the data.

grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -869,7 +869,19 @@ class OrderService {
...
+                    // budget codes. In case when there is only one budget code, and this one is inactive we have to throw a validation error,
+                    // so we can't just look for BudgetCode.findAllByCodeAndActive(code, true);
+                    List<BudgetCode> foundBudgetCodes = BudgetCode.findAllByCode(code)
+                    List<BudgetCode> activeBudgetCodes = foundBudgetCodes.findAll { it.active }
+
+                    if (activeBudgetCodes.size() > 1) {
+                        throw new IllegalArgumentException("Found more than one active budget code with the same code.")
+                    }
+
+                    BudgetCode budgetCode = activeBudgetCodes.size() == 0 ? foundBudgetCodes[0] : activeBudgetCodes.first()
```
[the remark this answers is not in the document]
alannadolny: > you're mixing your access methods here ([0] and x.first())

I did it on purpose, the `.first` method throws an error when there is no element, `[0]` doesn't so it will be null, and then we will get the information that we didn't find any budget code.

If there is more than one inactive budget code, we will get the error that the budget code is inactive, independently of which one inactive budget code the system will pick.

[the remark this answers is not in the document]
EWaterman: oh neat I didn't actually know that groovy made index-based accesses null safe. :+1:

EWaterman (approved):

awalkowiak (approved):

PR #4848, opened 2024-09-19
OBPIH-6679 Fix inferring bin location when providing a lot number

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6679](https://pihemr.atlassian.net/browse/OBPIH-6679)

**Description:** Had to modify the inferring bin location logic a bit due to following problem.
- inferring would work on out of stock bins, which is why I had to sue the product availability service which handles this search properly
- Were some issues with inferring default bin location
- providing a wrong bin location with correct lot would result in the algorithm trying to infer the bin location based on lot which we do not want in this situation and want to display the incorrect bin

---

[OBPIH-6679]: https://pihemr.atlassian.net/browse/OBPIH-6679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
+++ b/src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
@@ -28,11 +28,12 @@ class ImportPackingListItem implements Validateable {
         if (source['lotNumber']) {
             return source['lotNumber']
         }
+        ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
 
         // otherwise infer inventory item based on provided bin-location
-        Location internalLocation = Location.findByNameAndParentLocation(source['binLocation'], obj.origin)
         Product product = Product.findByProductCode(source['product'])
-        List<ProductAvailability> items = ProductAvailability.findAllByProductAndBinLocationAndLocation(product, internalLocation, obj.origin)
+        List<ProductAvailability> items = productAvailabilityService.getAvailableBinLocations(obj.origin, product?.productCode)
+                .findAll { it?.binLocation?.name == source['binLocation'] }
 
         // infer lot-number only if there is a single possible inventory
         if (items.size() == 1) {
@@ -48,22 +49,24 @@ class ImportPackingListItem implements Validateable {
     @BindUsing({ obj, source ->
         // Handle inferring bin location when it is not provided by the user
         if (!source['binLocation']) {
-            Product product = Product.findByProductCode(source['product'])
-            List<ProductAvailability> items = ProductAvailability.findAllByProductAndLotNumberAndLocation(product, source['lotNumber'], obj.origin)
+            ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
 
+            Product product = Product.findByProductCode(source['product'])
+            List<ProductAvailability> items = productAvailabilityService.getAvailableBinLocations(obj.origin, product?.productCode)
+                    .findAll { it?.inventoryItem?.lotNumber == source['lotNumber']}
             // Infer bin location only if there is a single possible inventory
             if (items.size() == 1) {
                 return items.first().binLocation
             }
 
             // If default bin is in stock then return it
-            if (items.size() > 0 && items.any{ it.binLocation == null }) {
+            if (items.size() > 0 && items.any { it.binLocation == null }) {
                 // null value assumes that this is a default bin location
                 return null
             }
 
             // Any attempt at inferring bin location failed and bin location was not located
-            obj.binLocationFound = false
+            return null
         }
 
         Location internalLocation = Location.findByNameAndParentLocation(source['binLocation'], obj.origin)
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -28,11 +28,12 @@ class ImportPackingListItem implements Validateable {
...
             return source['lotNumber']
         }
+        ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
 
         // otherwise infer inventory item based on provided bin-location
-        Location internalLocation = Location.findByNameAndParentLocation(source['binLocation'], obj.origin)
         Product product = Product.findByProductCode(source['product'])
-        List<ProductAvailability> items = ProductAvailability.findAllByProductAndBinLocationAndLocation(product, internalLocation, obj.origin)
+        List<ProductAvailability> items = productAvailabilityService.getAvailableBinLocations(obj.origin, product?.productCode)
+                .findAll { it?.binLocation?.name == source['binLocation'] }
```
kchelstowski, line 36: since there is a lot of those methods in the product availability service, I won't force that, but I would prefer to have this filtering in the query itself, instead of calling `.findAll` just after doing the fetch.
Please take a look if there is a method that would return product availability for pair: `origin`, `product`, `binLocation` and use it if possible.
Otherwise, if the performance is OK for an item, that has stock in a lot of bins, I'm fine with that, not to implement another comprehensive method in that service.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -46,26 +47,44 @@ class ImportPackingListItem implements Validateable {
...
+            // Infer bin location only if there is a single possible inventory
+            if (items.size() == 1) {
+                return items.first().binLocation
+            }
+
+            // If default bin is in stock then return it
+            if (items.size() > 0 && items.any { it.binLocation == null }) {
+                // null value assumes that this is a default bin location
+                return null
+            }
```
kchelstowski, line 66: is this check needed, if you return null below anyway?

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -121,7 +140,7 @@ class ImportPackingListItem implements Validateable {
         })
         binLocation(nullable: true, validator: { Location binLocation, ImportPackingListItem item ->
             if (!item.binLocationFound) {
-                return ['binLocationNotFound', binLocation.name]
+                return ['binLocationNotFound', binLocation?.name]
```
kchelstowski, line 143: I don't mind it, but I'm wondering if you had run into the nullpointer here? I think it's not possible, because the only possible option to execute this code is if you return the dummy location

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -46,26 +47,44 @@ class ImportPackingListItem implements Validateable {
...
-            obj.binLocationFound = false
-            // returns a dummy location object to add more context in a response of what location was not found
-            return new Location(name: source['binLocation'])
+        // Handle inferring bin location when it is not provided by the user
+        if (!source['binLocation']) {
+            ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
+
+            Product product = Product.findByProductCode(source['product'])
+            List<ProductAvailability> items = productAvailabilityService.getAvailableBinLocations(obj.origin, product?.productCode)
+                    .findAll { it?.inventoryItem?.lotNumber == source['lotNumber']}
```
kchelstowski, line 56: a similar comment like above

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
awalkowiak: I don't think we have a one right now that does a search for specific bin locaitons

awalkowiak (approved):
Beside the one unnecessary if with return null that Kacper already mentioned it LGTM.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
drodzewicz: yeah, we don't and I would need to implement a new service method for that.
I don't think we are dealing with too much data at this point and maybe a simple `findAll` search should be enough

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
drodzewicz: yeah, you are right, I had a little different approach before and this was necessary there.

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
drodzewicz: you are right, I had a NPE but I had used a different approach before where I returned a a `null` value with setting this flag as `false` which resulted in a NPE.
I can remove this nullsafe operator now

kchelstowski (approved):

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
kchelstowski: ok, I guess we can just monitor and live with it for now, but I wanted to make a point, that this usually should not be the way to go.

alannadolny (approved):

PR #4849, opened 2024-09-19
OBPIH-6690 Add option to edit unit price on invoice items made from order adjustments (fix after QA)

alannadolny:


awalkowiak (approved):

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -246,6 +246,7 @@ class InvoiceItem implements Serializable {
                 inverse: inverse,
                 isCanceled: orderItem?.canceled ?: orderAdjustment?.canceled,
                 quantityAvailableToInvoice: quantityAvailableToInvoice,
+                amountAvailableToInvoice: orderAdjustment?.amountAvailableToInvoice,
```
EWaterman, line 249: do we only need this field for adjustments? It'll be null for regular invoice items. I know these are just for display but intuitively it feels like both fields should work for both cases (with quantity being 1 or 0 for adjustments).

All these fields are starting to feel confusing to me... Shouldn't amountAvailableToInvoice be based on the amount field? Maybe a transient would apply best here (though I think this class is starting to get overloaded with those).

src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
```diff
@@ -133,7 +133,11 @@ const useInvoicePrepaidItemsTable = ({
 
   // validation for order adjustments
   const validateUnitPrice = (row) => {
-    if (row?.unitPrice === 0 || Math.abs(row?.unitPrice) > Math.abs(row?.amount)) {
+    const amountAvailableToInvoice =
+      Math.abs(row?.amountAvailableToInvoice) + Math.abs(row?.amount);
```
EWaterman, line 137: why plus?

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
awalkowiak: > do we only need this field for adjustments? It'll be null for regular invoice items. I know these are just for display but intuitively it feels like both fields should work for both cases (with quantity being 1 or 0 for adjustments).

This is InvoiceItem's toJson. So in reality this handles all cases, not only adjustments

> All these fields are starting to feel confusing to me... Shouldn't amountAvailableToInvoice be based on the amount field? Maybe a transient would apply best here (though I think this class is starting to get overloaded with those).

Yeah this might be confusing, the unit price and amount for order adjustments, but I would say those two are the same (potentially only differs with sign). So perhaps it is misleading in some places. Tomorrow I'll take a deeper look at the partial invoicing of adjustments, if there is some potential for some simplification.

src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
awalkowiak: I think this is a case where we have "amount available **outside** this invoice item" **plus** "amount **from this** item" to determine full available amount

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
EWaterman: > This is InvoiceItem's toJson. So in reality this handles all cases, not only adjustments

So we'd need amountAvailableToInvoice to look at the shipment or order item as well then right?

> Yeah this might be confusing, the unit price and amount for order adjustments, but I would say those two are the same (potentially only differs with sign). So perhaps it is misleading in some places. Tomorrow I'll take a deeper look at the partial invoicing of adjustments, if there is some potential for some simplification.

Sounds good!

src/js/hooks/invoice/useInvoicePrepaidItemsTable.jsx
EWaterman: I just thought it'd be amount available to invoice = current amount available to invoice - amount we're invoicing for in this row

PR #4851, opened 2024-09-19
OBPIH-6689 Fix amount available to inverse calculation

awalkowiak:

**Description:**
Previously this value was wrongly calculated. Instead of subtraction, it should be addition.

### DIFF
--- a/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
@@ -438,6 +438,17 @@ class PrepaymentInvoiceService {
 
     BigDecimal getAmountAvailableToInverse(OrderAdjustment orderAdjustment, InvoiceItem prepaymentItem) {
         BigDecimal inversedAmount = orderAdjustment.inversedAmount
-        return Math.abs(prepaymentItem.amount) > Math.abs(inversedAmount) ? prepaymentItem.amount - inversedAmount : 0
+        BigDecimal prepaymentAmount = prepaymentItem.amount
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -438,6 +438,17 @@ class PrepaymentInvoiceService {
...
+            return 0
+        }
+
+        // To determine what is current amount available to inverse we have to add prepayment amount to currently
+        // inversed amount (because both values have different signs). Examples:
+        // prepaymentItem.amount = 10, orderAdjustment.inversedAmount = -3, expected available to inverse = 7
+        // prepaymentItem.amount = -5, orderAdjustment.inversedAmount = 2, expected available to inverse = -3
+        // The actual inverse item sign will be switched while saving the inverse, becaue now we only determine
+        // how much of prepayment item's amount is available for inversing.
+        return prepaymentAmount + inversedAmount
```
EWaterman, line 452: for future evan to understand: the amount available to invoice is the absolute difference between the prepayment amount and the amount that has already been invoiced.

EWaterman: is the scenario you listed above realistic? If an adjustment is prepaid as a negative adjustment, when would we need to be able to regular invoice it later as a positive adjustment? Seems like a "discount" adjustment wouldn't make sense to be invoiced as an additional cost and vice versa for "tax" (or whatever) adjustments.

That said, I suppose it's fine to be able to support that case since it doesn't seem to be complicating calculations too much.

alannadolny (approved):

drodzewicz (approved):

PR #4852, opened 2024-09-19
OBPIH-6726 Bugfix to allow pre-migration prepayment invoices to be edited

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6726

**Description:** See ticket for the full breakdown. This fix is to ensure that when migrating existing invoices we're properly creating an inverse invoice item for each matching *shipment* item in a regular invoice. Previously we were only checking the prepayment invoice item, which is based on *order item* and so had no shipment association. This worked fine for cancelled items and adjustments, but for orders with shipments, the invoice needs to be built off of the regular invoice item (which itself is built from shipment items).

A diagram to help me understand the complex relationships going on here:

image


awalkowiak (commented):
On a first sight it looks good to me, but I'd like to test it locally with my recent changes as well, before merging

awalkowiak (approved):

PR #4843, opened 2024-09-16
OBPIH-6688 modify method checking if order adjustment is invoiceable

drodzewicz:
### :sparkles: Description of Change

> *A concise summary of what is being changed. Please provide enough context for reviewers to be able to understand the change and why it is necessary. If the issue/ticket already provides enough information, you can put "See ticket" as the description.*

**Link to GitHub issue or Jira ticket:** [OBPIH-6688](https://pihemr.atlassian.net/browse/OBPIH-6688)

**Description:**
The is invoiced transient value should handle case where we are trying to check if adjusment is invoiceable when there is only partially invoiced amount.

---
### :camera: Screenshots & Recordings (optional)

> *If this PR contains a UI change, consider adding one or more screenshots here or link to a screen recording to help reviewers visualize the change. Otherwise, you can remove this section.*

[OBPIH-6688]: https://pihemr.atlassian.net/browse/OBPIH-6688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4854, opened 2024-09-20
OBPIH-6688 Check if adjusment was invoiced as canceled

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6688](https://pihemr.atlassian.net/browse/OBPIH-6688)

**Description:**
In the previous PR for this ticket #4843 I have removed the condition checking if the adjustment is invoiceable if it was canceled with 0 amount, I assumed it is going to be covered by the newly implemented logic.
After QA it appears that this case was not handled too well and we need to check for it in the method.  

---


grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -180,6 +180,11 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
             return hasPrepaymentInvoice && !hasRegularInvoice && order.placed && !invoicedFullAmount
         }
 
+        // check if by any chance it was invoiced as cancelled
+        if (totalAdjustments == 0) {
+            return order.placed && !hasRegularInvoice
```
EWaterman, line 185: is there a case where we do have a regular invoice but it doesn't invoice for the zero adjustment? Wouldn't in that case the adjustment still be invoiceable?

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -180,6 +180,11 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
             return hasPrepaymentInvoice && !hasRegularInvoice && order.placed && !invoicedFullAmount
         }
 
+        // check if by any chance it was invoiced as cancelled
```
awalkowiak, line 183: I think the if check below makes sense, but I am not sure if the comment here makes sense. If it was invoiced as canceled then it will have a regular invoice. Plus looking at Katarzyna's comment in the ticket, there was a case about the total adjustments being 0, this does not mean it was canceled. Am I missing something here?

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
drodzewicz: I don't think there is such a case.
I was not aware of the case for invoicing adjustments with amount 0 and this check should handle it.

So in my understanding, if an adjustment already has a regular incoice it was probably already invoiced in some way and since totalAdjusmten amount specified is zero we can assume it was invoiced as 0

EWaterman (approved):

awalkowiak (approved):

alannadolny (approved):

PR #4855, opened 2024-09-20
OBPIH-6689 Rely on unit price instead amount while calculating inverse items

awalkowiak:

**Description:**
This is basically a changing approach to determining inverse items. Previously we were relying on the amount field, which while comparing to unit prices could have different sign and could be reduced by prepayment percentage. To simplify things we should always rely on the unit price all the time while determining inverse items.

### DIFF
--- a/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
+++ b/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
@@ -151,9 +151,9 @@ class PrepaymentInvoiceService {
                 quantity: orderAdjustment?.canceled ? 0 : 1,
                 quantityUom: orderAdjustment.orderItem?.quantityUom,
                 quantityPerUom: orderAdjustment.orderItem?.quantityPerUom ?: 1,
-                unitPrice: orderAdjustment.amountAvailableToInvoice,
-                // For non-canceled order adjustments amount is equal to the total adjustments
-                amount: orderAdjustment?.canceled ? 0 : orderAdjustment.amountAvailableToInvoice
+                unitPrice: orderAdjustment.unitPriceAvailableToInvoice,
+                // For non-canceled order adjustments amount is equal to the total adjustments (which is kept as unit price)
+                amount: orderAdjustment?.canceled ? 0 : orderAdjustment.unitPriceAvailableToInvoice
         )
         invoiceItem.addToOrderAdjustments(orderAdjustment)
         return invoiceItem
@@ -203,8 +203,11 @@ class PrepaymentInvoiceService {
             return null
         }
 
-        BigDecimal amountToInverse = getAmountAvailableToInverse(orderAdjustment, prepaymentItem)
-        if (!amountToInverse) {
+        BigDecimal unitPriceToInverse = getUnitPriceAvailableToInverse(
+                prepaymentItem.unitPrice,
+                orderAdjustment.inversedUnitPrice
+        )
+        if (!unitPriceToInverse) {
             return null
         }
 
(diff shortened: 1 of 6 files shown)


alannadolny (approved):

src/test/groovy/unit/org/pih/warehouse/invoice/PrepaymentInvoiceServiceSpec.groovy
```diff
@@ -347,47 +347,50 @@ class PrepaymentInvoiceServiceSpec extends Specification implements ServiceUnitT
...
-        preapymentItemAmount   | inversedAmount   | amountAvailableToInverse
-        1.0                    | -1.0             | 0.0
-        -1.0                   | 1.0              | 0.0
-        1.0                    | 0.0              | 1.0
-        -1.0                   | 0.0              | -1.0
-        0.0                    | 0.0              | 0.0
-        10                     | -3               | 7
-        -5                     | 2                | -3
-        -5                     | 5                | 0
+        preapymentItemUnitPrice | inversedUnitPrice | unitPriceAvailableToInverse
```
EWaterman, line 385: minor: prepaymentItemUnitPrice

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
```diff
@@ -432,18 +438,11 @@ class PrepaymentInvoiceService {
...
-        // To determine what is current amount available to inverse we have to add prepayment amount to currently
-        // inversed amount (because both values have different signs). Examples:
-        // prepaymentItem.amount = 10, orderAdjustment.inversedAmount = -3, expected available to inverse = 7
-        // prepaymentItem.amount = -5, orderAdjustment.inversedAmount = 2, expected available to inverse = -3
-        // prepaymentItem.amount = -5, orderAdjustment.inversedAmount = 5, expected available to inverse = 0
-        // The actual inverse item sign will be switched while saving the inverse, because now we only determine
-        // how much of prepayment item's amount is available for inversing.
-        return prepaymentItem.amount + orderAdjustment.inversedAmount
+    BigDecimal getUnitPriceAvailableToInverse(BigDecimal unitPrice, BigDecimal inversedUnitPrice) {
+        return Math.abs(unitPrice) > Math.abs(inversedUnitPrice) ? unitPrice - inversedUnitPrice : 0
```
EWaterman, line 446: reintroducing this discussion from the other PR: are we allowing a scenario where unit price and inversed unit price have different signs? Like I said previously, that case doesn't really make sense, but if a user gets confused and puts in the wrong sign, we might get into undesirable scenarios here.

Ex: unit price is 1 and inversed unit price is -1. We should return 2 but we return 0.

Maybe because of this we shouldn't allow the signs to differ.

EWaterman:
I'm still wrapping my head around this change. Is this going to impact how the migration works?

awalkowiak:
> I'm still wrapping my head around this change. Is this going to impact how the migration works?

@EWaterman I **think** it should not affect the migration, because this in majority affects/targets editing invoice items and inverse items. I'll try to double check that while testing your migration locally

grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy
awalkowiak: > are we allowing a scenario where unit price and inversed unit price have different signs? 

No. **Both** will have the same sign (can be negative or positive but both will have the same, and this is already validated here https://github.com/openboxes/openboxes/blob/release/0.9.2-hotfix1/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy#L400). That's why I changed to the unit price from amount, because with the amount it was a wild ride here because amount could have different signs.

awalkowiak: > are we allowing a scenario where unit price and inversed unit price have different signs? 

No. **Both** will have the same sign (can be negative or positive but both will have the same, and this is already validated here https://github.com/openboxes/openboxes/blob/release/0.9.2-hotfix1/grails-app/services/org/pih/warehouse/invoice/PrepaymentInvoiceService.groovy#L400). That's why I changed to the unit price from amount, because with the amount it was a wild ride here because amount could have different signs.

drodzewicz (approved):

PR #4857, opened 2024-09-23
OBPIH-6691 Add partially invoiced order summary status for order adjustments

alannadolny:
**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6691

### DIFF
--- a/grails-app/migrations/views/order-summary-helper-views.sql
+++ b/grails-app/migrations/views/order-summary-helper-views.sql
@@ -105,29 +105,32 @@ SELECT
         ELSE 1
         END AS quantity_ordered,
     quantity_invoiced,
-    IFNULL(
-            order_adjustment_amount,
+    ABS(
+        IFNULL(
+        order_adjustment_amount,
             IF(order_adjustment_percentage IS NOT NULL,
                IF(order_item_id IS NOT NULL,
                   order_item_subtotal * (order_adjustment_percentage / 100),
                   order_subtotal * (order_adjustment_percentage / 100)),
-               0)
+               0
+            )
+        )
     ) as total_adjustments,
     invoice_item_amount as order_adjustment_amount
 FROM (
          SELECT
-             `order`.id                        	    	 AS order_id,
-             order_adjustment.id 		      	    	 AS adjustment_id,
-             order_adjustment.canceled 		    	   	 AS order_adjustment_canceled,
-             order_adjustment.amount     			     AS order_adjustment_amount,
-             invoice_item.amount 						 AS invoice_item_amount,
-             `order`.order_number              		     AS order_number,
-             order.status                      		     AS order_status,
-             invoice_item.id				        	 AS invoice_item_id,
-             order_adjustment.percentage 			 	 AS order_adjustment_percentage,
-             order_adjustment.order_item_id 	         AS order_item_id,
-             order_item.quantity * order_item.unit_price AS order_item_subtotal,
-             SUM(order_total.order_total)				 AS order_subtotal,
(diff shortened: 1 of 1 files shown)


grails-app/migrations/views/order-summary-helper-views.sql
```diff
@@ -246,9 +267,10 @@ CREATE OR REPLACE VIEW order_summary AS (
...
-                WHEN (SUM(items_and_adjustments_union.quantity_ordered) + SUM(items_and_adjustments_union.adjustments_count)) = (SUM(items_and_adjustments_union.quantity_invoiced) + SUM(items_and_adjustments_union.adjustments_invoiced)) THEN 'INVOICED'
-                WHEN (SUM(items_and_adjustments_union.quantity_ordered) + SUM(items_and_adjustments_union.adjustments_count)) > 0 AND (SUM(items_and_adjustments_union.quantity_invoiced) + SUM(items_and_adjustments_union.adjustments_invoiced) > 0) THEN 'PARTIALLY_INVOICED'
+                WHEN SUM(items_and_adjustments_union.quantity_ordered) = 0 AND SUM(items_and_adjustments_union.invoiced_amount) = 0 THEN NULL
+                WHEN SUM(items_and_adjustments_union.quantity_ordered) = SUM(items_and_adjustments_union.quantity_invoiced) AND SUM(items_and_adjustments_union.total_adjustments) = SUM(items_and_adjustments_union.invoiced_amount) THEN 'INVOICED'
+                WHEN (SUM(items_and_adjustments_union.quantity_ordered) < SUM(items_and_adjustments_union.quantity_invoiced) AND SUM(items_and_adjustments_union.quantity_ordered) > 0)
```
awalkowiak, line 272: I think this `SUM(union.quantity_ordered) < SUM(union.quantity_invoiced` should have the opposite sign. It's partially invoiced if only part of the ordered quantity is invoiced --> `ordered > invoiced`   
plus this part: `AND SUM(items_and_adjustments_union.quantity_ordered) > 0` should rather check if `quantity_invoiced > 0`

grails-app/migrations/views/order-summary-helper-views.sql
```diff
@@ -246,9 +267,10 @@ CREATE OR REPLACE VIEW order_summary AS (
...
+                         OR (SUM(items_and_adjustments_union.invoiced_amount) < SUM(items_and_adjustments_union.total_adjustments) AND SUM(items_and_adjustments_union.total_adjustments) > 0) THEN 'PARTIALLY_INVOICED'
```
awalkowiak, line 273: `AND SUM(items_and_adjustments_union.total_adjustments) > 0` this should be as well `invoiced_amount > 0`

awalkowiak (changes requested):

awalkowiak (changes requested):
Few key points from my local testing / local review:
1. As mentioned in previous review, the `PARTIALLY_INVOICED` checks are wrong and need to be improved.
2. If nothing is invoiced, then order status should not be `partially invoiced`, but should properly reflect the **current** order state.
3. The invoiced amount calculation does not take into account if invoice is posted (here we only count only posted invoices).
4. With the current solution there might be an ssue when we have an order adjutement with amount 0, that was invoiced for this amount (it should be treated as invoiced properly). Probably you could utilize the still available adjustment count and invoiced adjustment count on top of amount comparison for this case.

awalkowiak (approved):
There still might be one case that is not working, but I am merging it as is (disregarding the one problematic case, it looks good) since there is a lot to QA here.

PR #4860, opened 2024-09-24
OBPIH-6740 Move liquibase migrations to a single entrypoint changelog

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6740

**Description:** There are more details in the ticket, but essentially this change moves the migration logic that we had in Bootstrap into the entrypoint changelog. I've also taken the chance to clean up how we call our clean install migrations vs regular upgrade migrations to hopefully simplify the process somewhat. This approach lets us take advantage of the migration plugin's properties (which will be convenient for tests to be able to override the changelog, providing its own that will insert test data). For example:

```
grails:
    plugin:
      databasemigration:
          changelogFileName: changelog.groovy


environments:
  test:
      plugin:
          databasemigration:
              changelogFileName: changelog-test.groovy

```

You'll also notice that I added a 'final' tagged changeset to each release folder. We don't strictly need to use tags for this change to work, but it is a more proper solution in my opinion. It's easy enough to revert though if we choose not to use them.


EWaterman:
@jmiranda take a look at this when you get a chance. I've only tested the happy path on my local so it's not ready to merge yet but I wanted to get something into review so that we could talk about it. If we like it, and want to move forward with a change like this, we can open it up to the rest of the team and start testing it more properly.

src/main/groovy/util/TaggedMigrationVersion.groovy
```diff
@@ -0,0 +1,60 @@
+package util
+
+import org.apache.commons.lang3.math.NumberUtils
+
+/**
+ * Represents an app version from the perspective of a tagged liquibase changeSet.
+ * By convention these should map to directories under the grails-app/migrations folder (Ex: '0.9.x').
+ */
+class TaggedMigrationVersion implements Comparable<TaggedMigrationVersion> {
```
EWaterman, line 9: This is just a convenience class to make comparing versions easier

EWaterman:
okay, I see that this is failing for integration tests with:
```
WARN  groovy.sql.Sql - Failed to execute: SELECT EXISTS (SELECT 1 FROM DATABASECHANGELOG) because: Table 'test.DATABASECHANGELOG' doesn't exist
    2024-09-24 21:26:36,369 [Test worker] ERROR o.s.boot.SpringApplication - Application startup failed
    liquibase.exception.ChangeLogParseException: java.sql.SQLSyntaxErrorException: Table 'test.DATABASECHANGELOG' doesn't exist
```
I'll try to figure out why migrations aren't running properly for test containers but we can still review this change as normal for now.

grails-app/migrations/changelog.groovy
```diff
@@ -1,9 +1,75 @@
...
+ * don't want to assume that it always brings us to the latest release because we may have forgotten or decided not to
+ * update/regenerate the 'install' migrations during the release process.
+ */
+final TaggedMigrationVersion CLEAN_INSTALL_VERSION = new TaggedMigrationVersion('0.8.x')
+
+/*
+ * List all of our tagged migration versions. Any new tagged versions should be added to this list.
+ * By convention, these should map to the migrations directories.
+ */
+final List<TaggedMigrationVersion> ALL_VERSIONS = [
```
EWaterman, line 16: I thought about having it read all the folder names in the /migrations folder to extract these versions so that we wouldn't need to update this list with every release but that felt pretty hacky so I left it like this. We're already expecting that specific folder structure though so maybe it's fine to do.

EWaterman:
I've tested this change on my local 0.9.x setup and on clean installs via integration tests. It could be useful to further test this for 0.8.x installs just to be extra safe though

jmiranda:
I promise I'll get back to reviewing this one soon.

jmiranda:
I encountered a weird issue when running this against a 0.8.x database. 
```
2025-01-13 15:12:59,467 ERROR [main] liquibase.changelog.ChangeSet: Change Set 0.8.x/changelog-2018-05-30-2300-create-dimensional-model.xml::1546354459620-35::jmiranda (generated) failed.  Error: Table 'openboxes_demo2.transaction_type_dimension' doesn't exist [Failed SQL: (1146) ALTER TABLE openboxes_demo2.transaction_type_dimension ADD CONSTRAINT FK7FA87A22B3FB7111 FOREIGN KEY (transaction_type_id) REFERENCES openboxes_demo2.transaction_type (id)]
```

Upon further investigation, it turns out the database was not getting restored completely as the mysql import was stopping around the shipment tables. 

I hadn't noticed it the first time around, but upon re-running the restore process off the demo database dump I saw the following error. 
```
[jmiranda@ThinkPad-P17 ~]$ mysql openboxes_demo < openboxes_demo.sql 
ERROR 1273 (HY000) at line 5071: Unknown collation: 'utf8mb4_0900_ai_ci'
```
For context, I'm running the mysqldump on MySQL 8 and the MySQL import on MariaDB, so there's something wonky between the two. However, the weird thing is that none of our tables should be using utf8mb4 since we always default to utf8 which is a pointer to utf8mb3. 

Long story short, I need a bit more time to debug this. 

At the very least the changelog seems to be executing the expected versions, so I think once I think we'll be good once I get passed this.
```
2025-01-13 15:12:55,768 INFO  [main] org.pih.warehouse.BootStrap: Executing migrations ...
Executing migrations for release version: 0.8.x
Executing migrations for release version: 0.9.x
```

One caveat is that I was hoping to see the same logging from before (in particular, the lines that output executeChangelogVersions and upgradeChangelogVersions). Is it possible to restore those? It's just helpful for debugging in case we encounter issues in the future. You can make them DEBUG if necessary.
```
2025-01-13 16:07:03,366 INFO  [main] org.pih.warehouse.BootStrap: executedChangelogVersions: [[version:install], [version:0.8.x], [version:0.9.x], [version:0.5.x], [version:0.6.x], [version:0.7.x], [version:views]]
2025-01-13 16:07:03,366 INFO  [main] org.pih.warehouse.BootStrap: Dropping all views (will rebuild after migrations complete)...
2025-01-13 16:07:04,218 INFO  [main] l.c.StandardChangeLogHistoryService: Reading from openboxes.DATABASECHANGELOG
2025-01-13 16:07:04,627 INFO  [main] liquibase.changelog.ChangeSet: Custom SQL executed
2025-01-13 16:07:04,627 INFO  [main] liquibase.changelog.ChangeSet: ChangeSet views/drop-all-views.xml::1633402273161-1::jmiranda ran successfully in 371ms
2025-01-13 16:07:06,870 INFO  [main] org.pih.warehouse.BootStrap: upgradeChangeLogVersions: [0.5.x, 0.6.x, 0.7.x, 0.8.x]
2025-01-13 16:07:06,873 INFO  [main] org.pih.warehouse.BootStrap: Running upgrade changelog ...
2025-01-13 16:07:08,229 INFO  [main] l.c.StandardChangeLogHistoryService: Reading from openboxes.DATABASECHANGELOG
2025-01-13 16:07:08,780 INFO  [main] org.pih.warehouse.BootStrap: Running latest updates
2025-01-13 16:07:09,419 INFO  [main] l.c.StandardChangeLogHistoryService: Reading from openboxes.DATABASECHANGELOG
```

EWaterman:
@jmiranda first off, thanks for testing these thoroughly.

Some of the log statements are necessarily gone because I moved the logic from BootStrap.groovy to inside the root changelog.groovy itself, which I don't think I can log from (I'm actually not certain about that). I can definitely log before and after the migrations run, so I'll try to add some more helpful log statements there at least. They shouldn't be too bloated so I don't mind them staying as INFO.

EWaterman:
@jmiranda I updated the logs to the following. Turns out the changelog can in fact log properly:
```
2025-01-14 08:39:29,874 INFO  [main] org.pih.warehouse.BootStrap: Executing migrations ...
2025-01-14 08:39:30,401 INFO  [main] changelog: Dropping all views. They will be rebuilt after migrations complete...
2025-01-14 08:39:30,447 INFO  [main] changelog: Current migration version is 0.9.x. Skipping older migrations. They've already ran.
2025-01-14 08:39:30,451 INFO  [main] changelog: Executing migrations for release version: 0.9.x
2025-01-14 08:39:31,774 INFO  [main] l.c.StandardChangeLogHistoryService: Reading from openboxes.DATABASECHANGELOG
2025-01-14 08:39:31,972 INFO  [main] liquibase.changelog.ChangeSet: Custom SQL executed
2025-01-14 08:39:31,972 INFO  [main] liquibase.changelog.ChangeSet: ChangeSet views/drop-all-views.xml::1633402273161-1::jmiranda ran successfully in 144ms
2025-01-14 08:39:31,992 INFO  [main] liquibase.changelog.ChangeSet: SQL in file views/product-demand-view.sql executed
2025-01-14 08:39:31,992 INFO  [main] liquibase.changelog.ChangeSet: ChangeSet views/changelog.xml::1580848680306-1::jmiranda ran successfully in 12ms
...all other changesets...
```

jmiranda (approved):
I finally tested the three cases I wanted to cover
* empty database
* existing 0.8.x database 
* existing 0.9.x database

... and everything worked as expected.

jmiranda:
With respect to the issue I encountered with the 0.8.x database, it looks like MySQL 8 set the character set to utf8mb4 for one table on the demo server. I'm not sure how that happens, but will check to see if obdev1 and/or any of the other 0.8.x databases to see if they have any tables with the wrong character set.
```
mysql> SELECT table_name, CCSA.character_set_name FROM information_schema.`TABLES` T,        information_schema.`COLLATION_CHARACTER_SET_APPLICABILITY` CCSA WHERE CCSA.collation_name = T.table_collation   AND T.table_schema = "openboxes_demo" and character_set_name != 'utf8mb3' \G
*************************** 1. row ***************************
        TABLE_NAME: stockout_fact
CHARACTER_SET_NAME: utf8mb4
1 row in set (0.00 sec)
```

PR #4865, opened 2024-09-26
OBPIH-6734 Scroll down to bottom position of the page when redo import

kchelstowski:

**Description:**

---

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -45,6 +45,7 @@ import {
   SET_ACTIVE_CONFIG,
   SET_OFFLINE,
   SET_ONLINE,
+  SET_SCROLL_TO_BOTTOM,
   SHOW_INFO_BAR,
   SHOW_INFO_BAR_MODAL,
   SHOW_SPINNER,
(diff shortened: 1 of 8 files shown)


drodzewicz:
I don't think I understand why was the approach with redux store necessary, I might be missing something so could you explain?

To me, looking at the current code, it looks like we could handle everything in `OutboundImport` component just like handling redirect to the details step when there is no cached data available
```jsx
  /** Redirect to first step if there is no cached data */
  useEffect(() => {
    if (_.isEmpty(cachedData) && !is(OutboundImportStep.DETAILS)) {
      navigateToStep(OutboundImportStep.DETAILS);
    }
  }, []);
```
so basically following the logic, if we have cached data available and we are on Details step, then scroll next button into view
```jsx
  useEffect(() => {
    if (!_.isEmpty(cachedData) && is(OutboundImportStep.DETAILS)) {
      buttonRef.current.scrollIntoView();
    }
  }, [currentStepKey]);
```

kchelstowski:
@drodzewicz the problem with that is then I would have to store the `useRef` in this component, and pass it to the Details component. The result of that would be that whenever we show the confirm page, the reference would become `undefined`, as the button would not be rendered by DOM, and it would be difficult to manipulate the state of the reference.
As for the Redux, I decided to use a reducer to avoid passing the props two levels down, and to be able to use a custom hook in one single component instead of mixing approach of passing props a few levels down + creating a custom hook that would use those then.

drodzewicz (approved):

PR #4866, opened 2024-09-27
OBPIH-6747 support empty lot number in import that matches inventory item defined as null

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6747](https://pihemr.atlassian.net/browse/OBPIH-6747)

**Description:**
The inferring bin Location and lot number required a bit of a refactor, I decided to move many parts into standalone service methods for better maintainability and readability because this feature has grown a bit.

Additionally since  there are many different inferring/binding scenarios I am working on implementing tests for these methods

---

[OBPIH-6747]: https://pihemr.atlassian.net/browse/OBPIH-6747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
@@ -393,11 +393,11 @@ class FulfillmentService {
 
         // If bin location is not provided then first check if inventory with default lot number is available
         if (!binLocationName) {
-            AvailableItem itemsWithDefaultLot = productAvailabilityService
+            AvailableItem itemWithDefaultLot = productAvailabilityService
                     .getAvailableItemWithDefaultLots(obj.origin, productCode)
             // only infer if there is one possible value
-            if (itemsWithDefaultLot) {
-                return itemsWithDefaultLot?.inventoryItem?.lotNumber
+            if (itemWithDefaultLot) {
+                return itemWithDefaultLot?.inventoryItem?.lotNumber
             }
         }
 
--- a/grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
@@ -13,6 +13,7 @@ import grails.gorm.transactions.Transactional
 import grails.util.Holders
 import grails.validation.ValidationException
 import org.apache.commons.lang3.StringUtils
+import org.pih.warehouse.api.AvailableItem
 import org.pih.warehouse.api.StockMovement
 import org.pih.warehouse.auth.AuthService
 import org.pih.warehouse.core.Constants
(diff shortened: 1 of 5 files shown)


grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
```diff
@@ -378,4 +379,71 @@ class FulfillmentService {
 
         return debitTransaction
     }
+
+    String bindOrInferLotNumber(ImportPackingListItem obj, Map source) {
```
kchelstowski, line 383: I can see that the only property from `obj` we use is the `origin`. Let's change the arguments to be `Location origin, Map source`, since we do not need anything else from the `obj`.

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
```diff
@@ -378,4 +379,71 @@ class FulfillmentService {
...
+
+        // If lot number is provided then just return this value and bind it
+        // Otherwise try to infer the lot number
+        if (lotNumber) {
+            return lotNumber
+        }
+
+        // If bin location is not provided then first check if inventory with default lot number is available
+        if (!binLocationName) {
+            AvailableItem itemsWithDefaultLot = productAvailabilityService
```
kchelstowski, line 396: shouldn't it be `item` instead of `items`?

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
```diff
@@ -378,4 +379,71 @@ class FulfillmentService {
...
+
+        return availableItem?.inventoryItem?.lotNumber
+    }
+
+    Location bindOrInferBinLocation(ImportPackingListItem obj, Map source) {
+        String lotNumber = source['lotNumber']
+        String productCode = source['product']
+        String binLocationName = source['binLocation']
+
+        ProductAvailabilityService productAvailabilityService = Holders.grailsApplication.mainContext.getBean(ProductAvailabilityService)
```
kchelstowski, line 416: This service is already injected in the `FulfillmentService`, you don't have to inject it again here.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -743,6 +743,68 @@ class ProductAvailabilityService {
         return availableBinLocations
     }
 
+    AvailableItem getAvailableItemWithDefaultLots(Location origin, String productCode) {
```
kchelstowski, line 746: I think the name of this method doesn't fit to what the method is actually doing. We should either think about the name again, or at least add a comment above the method.
My point is that if you find 3 items, in fact there is 3 available items found, but the method would return `null`, which is a bit misleading.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -743,6 +743,68 @@ class ProductAvailabilityService {
...
+            !it?.inventoryItem?.lotNumber // Null or empty lot number
+        }
+        // only infer if there is one possible value
+        if (itemsWithDefaultLot.size() == 1) {
+            return itemsWithDefaultLot.first()
+        }
+        return null
+    }
+
+    AvailableItem inferAvailableItemByBinLocation(Location origin, String productCode, String binLocationName) {
```
kchelstowski, line 760: I think `get` fits more than `infer` - we should change it to `get` here or use `infer` in all methods.

grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -743,6 +743,68 @@ class ProductAvailabilityService {
...
+    AvailableItem inferAvailableItemByLotNumber(Location origin, String productCode, String lotNumber) {
+        Product product = Product.findByProductCode(productCode)
+        List<AvailableItem> availableItems = getAvailableBinLocations(origin, product?.id)
+
+        // if bin location is not provided and lot is provided
+        // then look for it in available stock
+        List<AvailableItem> availableItemsWithProvidedLotNumber = availableItems.findAll {
+            it?.inventoryItem?.lotNumber == lotNumber
+        }
+        if(availableItemsWithProvidedLotNumber.size() == 1) {
```
kchelstowski, line 786: nitpicky: space between if and (

src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy
```diff
@@ -23,62 +22,17 @@ class ImportPackingListItem implements Validateable {
...
-        }
-
-        // Final fallback if provided location was not located
-        // we want to indicate, that a bin location for given name has not been found.
-        obj.binLocationFound = false
-        // return a dummy location object to add more context in a response of which location was not found
-        return new Location(name: source['binLocation'])
+    @BindUsing({ obj, source  ->
+        FulfillmentService fulfillmentService = Holders.grailsApplication.mainContext.getBean(FulfillmentService)
+        return fulfillmentService.bindOrInferBinLocation(obj as ImportPackingListItem, source as Map)
```
kchelstowski, line 35: instead of casting it here, I think you could assign the type in the `obj, source`, like:
```groovy
@BindUsing({ ImportPackingListItem obj, Map source -> })

drodzewicz: I am not sure why but for some reason this doesn't work.
When doing this it can't find the method which is why I resorted to the approach of casting the value

grails-app/services/org/pih/warehouse/fulfillment/FulfillmentService.groovy
drodzewicz: let's leave it as they are since the methods are basically the `@UseBinding` methods but moved to the service.
The other binding method requires this `obj` since we set `obj.binLcoationNotFound` flag on it and it would be weird where one has this argument and the other binding method doens't

kchelstowski (approved):

PR #4871, opened 2024-10-01
OBPIH-6693 fix adjustment invoice statuses

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6693](https://pihemr.atlassian.net/browse/OBPIH-6693)

**Description:**

---

[OBPIH-6693]: https://pihemr.atlassian.net/browse/OBPIH-6693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
@@ -128,7 +128,7 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
      * */
     BigDecimal getInvoicedUnitPrice() {
         return invoiceItems?.sum {
-            if (it.invoice?.isRegularInvoice && !it.inverse) {
+            if (it.invoice?.datePosted != null && it.invoice?.isRegularInvoice && !it.inverse) {
                 return it.unitPrice
             }
             return 0
--- a/grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
+++ b/grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
@@ -59,7 +59,6 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
     static transients = [
         'totalAdjustments',
         'postedPurchaseInvoiceItems',
-        'isInvoiced',
         "invoices",
         "hasInvoices",
         "hasPrepaymentInvoice",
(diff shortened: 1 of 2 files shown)


grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -128,7 +132,7 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
      * */
     BigDecimal getInvoicedUnitPrice() {
         return invoiceItems?.sum {
-            if (it.invoice?.isRegularInvoice && !it.inverse) {
+            if (it.invoice?.datePosted != null && it.invoice?.isRegularInvoice && !it.inverse) {
```
awalkowiak, line 135: I don't like it here. This might break the `getUnitPriceAvailableToInvoice`. I think this should be a separate transient (invoiced unit price on posted invoices)

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
awalkowiak: (as you can see, on the `OrderItem` there are both `quantityInvoiced` and `postedQuantityInvoiced` for the same reason)

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -100,27 +99,28 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
         }
     }
 
-    Boolean getIsInvoiced() {
-        return !postedPurchaseInvoiceItems.empty
+    Boolean getIsFullyInvoiced() {
+        return Math.abs(unitPriceOnPostedInvoices) >= Math.abs(totalAdjustments)
     }
```
drodzewicz, line 104: I removed the `isInvoiced` transient because it is not used anywhere anymore and if we will ever need this in the future I think we are going to use the derived status like we use in other domains.

grails-app/domain/org/pih/warehouse/order/OrderAdjustment.groovy
```diff
@@ -100,27 +99,28 @@ class OrderAdjustment implements Serializable, Comparable<OrderAdjustment> {
...
 
     OrderAdjustmentInvoiceStatus getDerivedPaymentStatus() {
-        Boolean fullyInvoiced = invoicedUnitPrice == totalAdjustments
-
-        if (!isInvoiced) {
+        if (postedPurchaseInvoiceItems.empty) {
```
awalkowiak, line 107: Could you add some comments inside this method (especially on the latter ifs checks)? Looking at the ordering of the if checks it overall makes sense and should work now, but I am afraid, that if someone starts adjusting here something, then it might easily get broken.

awalkowiak (approved):

PR #4872, opened 2024-10-02
OBPIH-6271 Add an ability to delete existing default preference type …

kchelstowski:
…via import

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
@@ -306,13 +306,13 @@ class ProductSupplierService {
         productSupplierPreference.preferenceType = preferenceType
         productSupplierPreference.comments = comments
 
-        Date globalPreferenceTypeValidityStartDate = validityStartDate ? Constants.EXPIRATION_DATE_FORMATTER.parse(validityStartDate) : null
+        Date globalPreferenceTypeValidityStartDate = validityStartDate ? Constants.MONTH_DAY_YEAR_DATE_FORMATTER.parse(validityStartDate) : null
 
         if (globalPreferenceTypeValidityStartDate) {
             productSupplierPreference.validityStartDate = globalPreferenceTypeValidityStartDate
         }
 
-        Date globalPreferenceTypeValidityEndDate = validityEndDate ? Constants.EXPIRATION_DATE_FORMATTER.parse(validityEndDate) : null
+        Date globalPreferenceTypeValidityEndDate = validityEndDate ? Constants.MONTH_DAY_YEAR_DATE_FORMATTER.parse(validityEndDate) : null
 
         if (globalPreferenceTypeValidityEndDate) {
             productSupplierPreference.validityEndDate = globalPreferenceTypeValidityEndDate
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
@@ -272,8 +272,26 @@ class ProductSupplierService {
 
         PreferenceType preferenceType = params.globalPreferenceTypeName ? PreferenceType.findByName(params.globalPreferenceTypeName) : null
 
+        assignDefaultPreferenceType(productSupplier,
+                preferenceType,
+                params.globalPreferenceTypeComments,
+                params.globalPreferenceTypeValidityStartDate,
+                params.globalPreferenceTypeValidityEndDate)
+
+        if (!productSupplier.code && !productSupplier.id) {
+            assignSourceCode(productSupplier, supplier)
+        }
+        return productSupplier
+    }
+
+    void assignDefaultPreferenceType(ProductSupplier productSupplier,
+                 PreferenceType preferenceType,
+                 String comments,
+                 String validityStartDate,
+                 String validityEndDate) {
+        ProductSupplierPreference productSupplierPreference = productSupplier.getGlobalProductSupplierPreference()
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -272,34 +272,51 @@ class ProductSupplierService {
...
+
+    void assignDefaultPreferenceType(ProductSupplier productSupplier,
+                 PreferenceType preferenceType,
+                 String comments,
+                 String validityStartDate,
+                 String validityEndDate) {
+        ProductSupplierPreference productSupplierPreference = productSupplier.getGlobalProductSupplierPreference()
         if (preferenceType) {
-            ProductSupplierPreference productSupplierPreference = productSupplier.getGlobalProductSupplierPreference()
+            SimpleDateFormat dateFormat = new SimpleDateFormat("MM/dd/yyyy")
```
alannadolny, line 294: Can you replace the hardcoded date format with the one placed in `Constants.groovy`?

alannadolny: You can also use the SimpleDateFormat from this file, it's already declared here

jmiranda: Good point @alannadolny 

Using SimpleDateFormat back in Grails 1.3.9 / Java 7 was a no-brainer since it was all we really had. I think we attempted to use Joda Time at some point to avoid some of the badness of the Java 7 date libraries. In Java 8 I think the datetime and formatting libraries are a bit better, so we could potentially use those. Wrapping whatever we do around a utility / service might be a good idea so we can codify our convention in one place.

So here are my recommendations. 
- [ ] Add Alan's recommendation to our current coding convention (using SimpleDateFormat, Constants, configuration) to make this clear to all developers. 
- [ ] If we're calling this assignDefaultPreferenceType() in a loop, we should try to find a way to share the date formatter across iterations of the loop (static variable, pass in as an argument) as this could be expensive.
- [ ] Consider adding a wrapper (utility, service) to abstract these methods so we can update the implementation without having to refactor a lot of code.
- [ ] I would also recommend using the interface / abstract class on the assignment side. 
        ```
        DateFormat dateFormat = new SimpleDateFormat("MM/dd/yyyy")
        ```
- [ ] Create a discovery ticket to figure out best practices for date formatting in Java 8 and beyond. 

None of these are urgent recommendations, so no need to do them here in this PR and I'll defer to @EWaterman @awalkowiak to decide whether any of these gets done at all.

kchelstowski: @alannadolny @jmiranda It was only a copy-paste of already existing code, hence I didn't pay much attention to that, but obviously I can change it to the existing constant.

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -272,34 +272,51 @@ class ProductSupplierService {
...
                 productSupplierPreference.validityStartDate = globalPreferenceTypeValidityStartDate
             }
 
-            def globalPreferenceTypeValidityEndDate = params.globalPreferenceTypeValidityEndDate ? dateFormat.parse(params.globalPreferenceTypeValidityEndDate) : null
+            Date globalPreferenceTypeValidityEndDate = validityEndDate ? dateFormat.parse(validityEndDate) : null
 
             if (globalPreferenceTypeValidityEndDate) {
                 productSupplierPreference.validityEndDate = globalPreferenceTypeValidityEndDate
             }
+            return
```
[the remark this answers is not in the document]
kchelstowski: could you elaborate what would be cleaner? in my opinion doing `if (!preferenceType)` at the beginning wouldn't change much, because the logic of assigning preferences would have to stay the same, so the only change would be, that this logic wouldn't be nested inside the `if` statement, unless I miss something :thinking:

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
EWaterman: Yeah I think we'd benefit from a mini discovery to standardize how we format our dates and then create some kind of date util class as Justin mentioned that abstracts away the specific implementation in case we need to change it at some point.

For this ticket though since it's not new functionality, I think we stick with Alan's recommendation.

awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -272,34 +273,50 @@ class ProductSupplierService {
...
+        }
 
-            if (globalPreferenceTypeValidityStartDate) {
-                productSupplierPreference.validityStartDate = globalPreferenceTypeValidityStartDate
-            }
+        productSupplierPreference.preferenceType = preferenceType
+        productSupplierPreference.comments = comments
 
-            def globalPreferenceTypeValidityEndDate = params.globalPreferenceTypeValidityEndDate ? dateFormat.parse(params.globalPreferenceTypeValidityEndDate) : null
+        Date globalPreferenceTypeValidityStartDate = validityStartDate ? Constants.EXPIRATION_DATE_FORMATTER.parse(validityStartDate) : null
```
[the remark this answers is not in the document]
kchelstowski: I was supposed to use an existing constant. I would say technically it might be considered as the expiration date, as the validity date is somewhat a synonym, isn't it? :thinking:

EWaterman (approved):

PR #4874, opened 2024-10-03
OBPIH-6691 Add partially invoiced order summary status for order adjustments

alannadolny:


grails-app/migrations/views/order-summary-helper-views.sql
```diff
@@ -221,119 +194,151 @@ CREATE OR REPLACE VIEW order_item_summary AS (
...
+       adjustments_invoiced,
+       items_ordered,
+       items_shipped,
+       items_received,
+       items_invoiced,
+       order_status,
+       shipment_status,
+       receipt_status,
+       payment_status,
+       total_adjustments,
```
awalkowiak, line 214: Do you need `total_adjustments` and `invoiced_adjustments_amount` here?

PR #4880, opened 2024-10-07
OBPIH-6746 Value in amount column of PO Invoice tab for regular POs

alannadolny:


grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -146,34 +144,6 @@ class InvoiceItem implements Serializable {
         return orderAdjustment ? orderAdjustment.description : product?.name
     }
 
-    /**
-     * Total shipment item value
-     * @deprecated From now should rely on the amount field instead always calculating it
-     */
-    def getTotalAmount() {
```
EWaterman, line 153: This is being used in `InvoiceService.getInvoiceItemsCsv`. You should hopefully just need to switch to use amount instead of totalAmount

awalkowiak: Yup, @alannadolny please replace it in the service as well

grails-app/migrations/0.9.x/changelog-2024-10-04-1610-assign-amount-to-invoice-items.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="anadolny" id="041020241610-0">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <columnExists tableName="invoice_item" columnName="amount"/>
+                <not>
```
EWaterman, line 10: Weird that liquibase requires this clunky syntax and doesn't have a better solution like `expectedResult=">0"` or `<sqlCheck notExpectedResult="0">`

grails-app/migrations/0.9.x/changelog-2024-10-04-1610-assign-amount-to-invoice-items.xml
```diff
@@ -0,0 +1,29 @@
...
+                    <sqlCheck expectedResult="0">
+                        SELECT COUNT(item.id) FROM invoice_item item
+                        WHERE item.amount IS NULL
+                    </sqlCheck>
+                </not>
+            </and>
+        </preConditions>
+
+        <sql>
```
EWaterman, line 19: can you add a comment here clarifying that this is only allowed because we've already migrated all prepayment invoices and so the only invoice items with a null amount at this point should be non-prepayment ones

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -247,10 +217,6 @@ class InvoiceItem implements Serializable {
                 isCanceled: orderItem?.canceled ?: orderAdjustment?.canceled,
                 quantityAvailableToInvoice: quantityAvailableToInvoice,
                 unitPriceAvailableToInvoice: orderAdjustment?.unitPriceAvailableToInvoice,
-                // Total amount and total prepayment amount are deprecated and amount field
-                // should be used instead (OBPIH-6398, OBPIH-6499)
-                totalAmount: totalAmount,
```
awalkowiak, line 252: I wonder if technically these two fields should stay for now, because this was expected as a part of `getPrepaymentItems`. I am not sure if we have a defined API endpoint deprecation flow. However, there is really high probability we were the only ones using it.

grails-app/migrations/0.9.x/changelog-2024-10-04-1610-assign-amount-to-invoice-items.xml
```diff
@@ -0,0 +1,29 @@
...
+            UPDATE invoice_item AS item
+            JOIN (
+                SELECT id, quantity * IFNULL(unit_price, 0) AS calculated_amount
+                FROM invoice_item
+            ) AS calculated_amount ON calculated_amount.id = item.id
+            SET item.amount = calculated_amount.calculated_amount
```
awalkowiak, line 25: It's nitpicky, but `calculated_amount.calculated_amount` looks weird.

grails-app/migrations/0.9.x/changelog-2024-10-04-1610-assign-amount-to-invoice-items.xml
EWaterman: Also, you can't just do:
```
UPDATE invoice_item
SET amount = quantity * IFNULL(unit_price, 0)
WHERE amount IS NULL;
```

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
alannadolny: so what's the decision here? should I revert those properties?

PR #4881, opened 2024-10-08
OBPIH-6275 Export boolean values as true/false booleans, instead of a…

kchelstowski:
…n empty string for false boolean

**Description:**
Initially I wanted to just parse booleans to string for the product sources export, not to impact the generic export builder, but it turned out, that the validation while importing the template back, would prevent the active field from being imported, as it would be passed as string, not as a boolean.
The change would potentially impact e.g. the Locations export, but I'd say there is nothing wrong with that, as it also fixes the same problem in the locations export, that would be probably addressed in a ticket anyway.

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DocumentService.groovy
@@ -404,9 +404,11 @@ class DocumentService {
     void createExcelRow(HSSFSheet sheet, int rowNumber, Map dataRow) {
         Row excelRow = sheet.createRow(rowNumber)
         dataRow.keySet().eachWithIndex { columnName, index ->
-            def cellValue = dataRow.get(columnName) ?: ""
-            // POI can't handle objects so we need to convert all objects to strings unless they are numeric
-            if (!(cellValue instanceof Number)) {
+            def actualValue = dataRow.get(columnName)
+            // We can't just check the truthiness of the actualValue, because the false boolean would be evaluated to an empty string
+            def cellValue = actualValue == null ? "" : actualValue
+            // POI can't handle objects so we need to convert all objects to strings unless they are numeric or boolean
+            if (!(cellValue instanceof Number) && !(cellValue instanceof Boolean)) {
                 cellValue = cellValue.toString()
             }
             excelRow.createCell(index).setCellValue(cellValue)
--- a/grails-app/services/org/pih/warehouse/data/DataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/DataService.groovy
@@ -702,13 +702,13 @@ class DataService {
                 } else if (element.defaultValue && !value) {
                     value = element.defaultValue
                 }
-                // We can't just check the truthiness of the actualValue, because the false boolean would be evaluated to an empty string
+                // We can't just check the truthiness of the value, because the false boolean would be evaluated to an empty string
                 properties[fieldName] = value == null ? "" : value
             } else {
                 // to access object value by key we must use the object.get(key) instead of object[key]
                 // because using the object[key] will throw an error when trying to export data using the batch controller
                 value = object.get(element) ?: element.tokenize('.').inject(object) { v, k -> v?."$k" }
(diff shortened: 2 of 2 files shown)


grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -702,12 +702,14 @@ class DataService {
                 } else if (element.defaultValue && !value) {
                     value = element.defaultValue
                 }
-                properties[fieldName] = value ?: ""
+                // We can't just check the truthiness of the actualValue, because the false boolean would be evaluated to an empty string
```
EWaterman, line 705: "actualValue" -> "value" in these two comments

kchelstowski: Thanks, I copied it from the other file and forgot to change it

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
```diff
@@ -404,9 +404,11 @@ class DocumentService {
...
         Row excelRow = sheet.createRow(rowNumber)
         dataRow.keySet().eachWithIndex { columnName, index ->
-            def cellValue = dataRow.get(columnName) ?: ""
-            // POI can't handle objects so we need to convert all objects to strings unless they are numeric
-            if (!(cellValue instanceof Number)) {
+            def actualValue = dataRow.get(columnName)
+            // We can't just check the truthiness of the actualValue, because the false boolean would be evaluated to an empty string
+            def cellValue = actualValue == null ? "" : actualValue
+            // POI can't handle objects so we need to convert all objects to strings unless they are numeric or boolean
+            if (!(cellValue instanceof Number) && !(cellValue instanceof Boolean)) {
```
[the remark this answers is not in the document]
kchelstowski: We can still pass booleans as strings for some imports. This method is used only for exports, they don't touch the import, so nothing would be broken in terms of the import.
The only change is that now, if e.g. a location/product source is inactive, it won't be exported as an empty string, but with the actual boolean value:
Screenshot from 2024-10-08 16-35-17
vs
Screenshot from 2024-10-08 16-36-10

This change actually would helps users, that they won't eventually have to fill those empty cells with `false`, as the exported file would already contain the value, that is supposed to be passed during the import.

EWaterman (approved):

grails-app/services/org/pih/warehouse/data/DataService.groovy
```diff
@@ -702,12 +702,14 @@ class DataService {
                 } else if (element.defaultValue && !value) {
                     value = element.defaultValue
                 }
-                properties[fieldName] = value ?: ""
+                // We can't just check the truthiness of the value, because the false boolean would be evaluated to an empty string
+                properties[fieldName] = value == null ? "" : value
```
alannadolny, line 706: can't we just use elvis operator here?

grails-app/services/org/pih/warehouse/core/DocumentService.groovy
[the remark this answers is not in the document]
kchelstowski: @awalkowiak this would affect e.g. location export, and I mentioned it in the description, but as I said, I believe this could be somewhat expected, as this would fix the same issue in the location export as well (users wouldn't have to fill empty cells with `false`), as the exported file would be compatible to be imported.
Moreover, in the ticket I had an info bar:

> Check Locations export as well, and other exports with active/inactive status

so I believe it would be expected to be done in those as well, but I haven't asked Manon explicitly about that.

TL:DR - I will ask Manon to make sure it's fine to impact/fix this issue in other exports as well and will reach out to you guys.

grails-app/services/org/pih/warehouse/data/DataService.groovy
kchelstowski: @alannadolny no we can't, I explained it in the comment above, that we couldn't just check the truthiness of the value, because the `false` boolean would be evaluated to an empty string, which we want to avoid.

PR #4883, opened 2024-10-08
OBPIH-6631 Add a bit more comments in product merge service

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6631

**Description:**
A bit of comments in the product merge service

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
@@ -39,20 +39,62 @@ class ProductMergeService {
     @Transactional
     def mergeProduct(String primaryId, String obsoleteId) {
         /**
-         * ================================
-         * === Swap Product's relations ===
-         * ================================
+         * ====================================
+         * ============= Preface ==============
+         * ====================================
+         * This method implements a product merge functionality. The main (primary) product will retain all
+         * transaction data and other relationships. Secondary (obsolete) product will be deactivated and all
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
```diff
@@ -140,6 +191,7 @@ class ProductMergeService {
             // Swap assemblyProduct to primary
             obsoletedComponent.assemblyProduct = primary
             // Note: needs flush because of "User.locationRoles not processed by flush"
+            // (this was an issue on grails 1.3.9, might not be a case anymore)
```
alannadolny, line 194: Maybe it will be worth creating a ticket for checking that and removing if this is unnecessary?

alannadolny (approved):

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
kchelstowski, line 194: nitpicky and not related/not important now, but I'm actually curious why `User.locationRoles` would even have an impact on those saves + if we were to remove the flush, we could get rid of those explicit saves at all, as we update an existing entity and all of the changes are tracked by Hibernate

kchelstowski (approved):

grails-app/services/org/pih/warehouse/product/ProductMergeService.groovy
awalkowiak: Added it to this one https://pihemr.atlassian.net/browse/OBPIH-5484

PR #4885, opened 2024-10-09
OBPIH-5182 IdentifierService overhaul

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-5182

**Description:**

Wowie this is a big change. Sorry. ~~I don't expect you to review this in its entirety before we get a chance to discuss it in a tech huddle.~~ Tech huddle chat has happened

The main focus of this ticket is to do the following:
- Add a retry mechanism when generating ids in case the one we generate already exists
- Check for uniqueness across tables when generating ids that are shared across tables. Specifically:
  - We use shipment number as location number when creating internal locations from shipments
  - We use order number as shipment number when creating a shipment from an order
  - We use request number as shipment number when performing stock movements

It was a pain to achieve the above since the code was quite segmented across classes, but also IdentifierService was bloated with feature specific logic that each had their own way of generating ids, so I opted to refactor the flow.

The logic is largely the same, but now instead of doing `IdentifierService.generateInvoiceIdentifer()` we do `InvoiceIdentifierService.generate()`. The feature specific services all extend from the IdentifierService. This gives us the following:
- All id generation methods share the same feature agnostic base flow, meaning they all get the benefits of all the id generation tools (custom delimiters, format overrides, random sections...)
- It's easy to hook in additional functionality to the feature specific services if needed
- It makes adding new feature services really easy (just a few methods to override)
- It keeps feature specific logic isolated from the other features

Still TODO:
- The logic in ProductIdentifierService, OrganizationIdentifierService, and PurchaseOrderIdentifierService need to be generalized and moved into IdentifierService so that we can support sequential ids more generally.
  - I won't be doing this in this task because it's big enough already so those classes are just using the old code as it was.
- ~~Writing unit tests on all refactored the id generation flow~~
- ~~Writing integration / smoke tests to verify that the current config can generate ids without error~~


src/main/groovy/org/pih/warehouse/core/identification/BlankIdentifierResolver.groovy
```diff
@@ -0,0 +1,76 @@
...
+ * Handles generating identifiers for all instances of an entity that don't yet have one.
+ *
+ * Used by the AssignIdentifierJob as a backup retry mechanism to ensure all entities have a valid id.
+ *
+ * Classes that implement this one will almost certainly also extend from IdentifierService. The only reason this trait
+ * is kept separate from the base service is to allow services to opt out of providing the functionality if they're
+ * not called by the AssignIdentifierJob.
+ */
+@Slf4j
+trait BlankIdentifierResolver<T extends GormEntity> {
```
EWaterman, line 16: This trait allows us to generalize supporting the AssignIdentifierJob that batch generates ids for rows where the id is null

grails-app/services/org/pih/warehouse/product/ProductIdentifierService.groovy
```diff
@@ -0,0 +1,199 @@
...
+import org.pih.warehouse.core.Constants
+import org.pih.warehouse.core.IdentifierGeneratorTypeCode
+import org.pih.warehouse.core.IdentifierService
+import org.pih.warehouse.core.identification.BlankIdentifierResolver
+import org.pih.warehouse.core.identification.IdentifierGeneratorParams
+
+// TODO: We should be able to move most of this logic into the identifier service. That way we can support sequential
+//       id generation more generally. We'd need an abstract method there for fetching the next sequence number to use,
+//       and a new ${sequence} config option to know where to put the sequence.
+class ProductIdentifierService extends IdentifierService implements BlankIdentifierResolver<Product> {
```
EWaterman, line 16: A majority of this code is just copied out from IdentifierService and some other places. I couldn't get around to refactoring this as a part of this change because it has such a complex configuration.

See the TODO here for how we could refactor in the future.

grails-app/services/org/pih/warehouse/order/PurchaseOrderIdentifierService.groovy
```diff
@@ -0,0 +1,70 @@
...
+import org.pih.warehouse.core.IdentifierService
+import org.pih.warehouse.core.IdentifierTypeCode
+import org.pih.warehouse.core.Organization
+import org.pih.warehouse.core.identification.IdentifierGeneratorParams
+import org.pih.warehouse.shipping.Shipment
+
+// TODO: We should be able to move this sequential logic into the identifier service. That way we can support sequential
+//       id generation more generally. We'd need an abstract method there for fetching the next sequence number to use,
+//       and a new ${sequence} config option to know where to put the sequence.
+class PurchaseOrderIdentifierService extends IdentifierService {
```
EWaterman, line 15: A majority of this code is just copied out from IdentifierService. I couldn't get around to refactoring this as a part of this change.

See the TODO here for how we could refactor in the future.

grails-app/services/org/pih/warehouse/core/OrganizationIdentifierService.groovy
```diff
@@ -0,0 +1,73 @@
+package org.pih.warehouse.core
+
+import grails.core.GrailsApplication
+import org.apache.commons.lang.WordUtils
+
+// TODO: Try to merge this logic into IdentifierService. We'd need to support a new ${abbreviation} keyword in the
+//       format that would get injected to based on the same logic that we have here. Then we'd add a new
+//       identifier.x.abbreviation.field = "path.to.field" config that we can use to extract the field from the entity.
+//       Or if that's too challenging, we can simply return the string to use via abstract method in IdentifierService.
+class OrganizationIdentifierService {
```
EWaterman, line 10: A majority of this code is just copied out from IdentifierService and some other places. I couldn't get around to refactoring this as a part of this change because it has such a complex configuration.

All I really did here was add comments to clarify the code.

See the TODO here for how we could refactor in the future.

grails-app/services/org/pih/warehouse/core/ConfigService.groovy
```diff
@@ -0,0 +1,37 @@
+package org.pih.warehouse.core
+
+import grails.core.GrailsApplication
+import grails.util.Holders
+
+/**
+ * Handles fetching property values from the app configuration.
+ */
+class ConfigService {
```
EWaterman, line 9: This was an attempt to unify our approach to fetching config.

It's based on best practices defined here: https://grails.org/blog/2016-08-31.html

It also makes mocking config in tests a lot easier.

grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
```diff
@@ -20,13 +19,10 @@ class AssignIdentifierJob {
...
 
-        Product.withNewSession {
-            identifierService.assignProductIdentifiers()
-            identifierService.assignShipmentIdentifiers()
-            identifierService.assignReceiptIdentifiers()
-            identifierService.assignOrderIdentifiers()
-            identifierService.assignRequisitionIdentifiers()
-            identifierService.assignTransactionIdentifiers()
-        }
+//        Product.withNewSession {
```
EWaterman, line 22: Note to self: test if we still need this!

grails-app/conf/runtime.groovy
```diff
@@ -4,38 +4,43 @@ import org.pih.warehouse.core.IdentifierGeneratorTypeCode
...
-openboxes.identifier.order.format = Constants.DEFAULT_ORDER_NUMBER_FORMAT
-openboxes.identifier.product.format = Constants.DEFAULT_PRODUCT_NUMBER_FORMAT
-openboxes.identifier.productSupplier.format = Constants.DEFAULT_PRODUCT_SUPPLIER_NUMBER_FORMAT
+openboxes.identifier.retry.max = 10
+openboxes.identifier.numeric = "0123456789"
+openboxes.identifier.alphabetic = "ABCDEFGHJKMNPQRSTUVXYZ"
+openboxes.identifier.alphanumeric = "0123456789ABCDEFGHJKMNPQRSTUVWXYZ"
+openboxes.identifier.sequenceNumber.format = "00000"
+
+// Default identifier
```
EWaterman, line 13: Note that I added "default" format and random fallbacks. This simplifies configuration for features that don't need a custom format

grails-app/services/org/pih/warehouse/core/IdentifierService.groovy
EWaterman, line 1: This is where the bulk of the change is. I generalize the generate flow here and split out any feature specific stuff to feature services. It's hard to parse in diff view so I suggest reviewing the raw file.

awalkowiak:
(review still in progress)

grails-app/services/org/pih/warehouse/product/ProductIdentifierService.groovy
awalkowiak: Yeah this one is though, perhaps we could get more tests for it? This is not a change request for now, but might be an entry point to bigger refactor of that part

awalkowiak (commented):
There still might be some places that are not thoroughly reviewed by me, but it overall lgtm and I think this requires some good testing locally and/or on dev instance

grails-app/services/org/pih/warehouse/core/OrganizationIdentifierService.groovy
```diff
@@ -0,0 +1,75 @@
...
+    GrailsApplication grailsApplication
+
+    /**
+     * Generates a unique id that resembles the given name.
+     *
+     * Ex: Given a name "Big Bad Guys" it might generate "BBG", or "BB0" if "BBG" already exists (up to "BB9").
+     * Ex: Given a name "Biggie" it might generate "BIG", or "BI0" if "BIG" already exists (up to "BB9").
+     */
+    String generate(String name) {
+        Integer minSize = grailsApplication.config.getProperty('openboxes.identifier.organization.minSize', Integer)
```
jmiranda, line 23: Potential Tech Huddle conversation to be had ... do we provide the config properties as part of the Params / Context object passed to the generate method OR do we always inject these into the service using the @Value annotation.

grails-app/services/org/pih/warehouse/core/LocationIdentifierService.groovy
```diff
@@ -0,0 +1,14 @@
...
+class LocationIdentifierService extends IdentifierService {
+
+    @Override
+    String getEntityKey() {
+        return "location"
+    }
+
+    @Override
+    protected Integer countDuplicates(String locationNumber) {
+        return Location.countByLocationNumber(locationNumber)
```
jmiranda, line 12: Seeing this implementation makes me wonder if the appropriate method name for this is ...
```
Integer countByIdentifier(String identifier);
```
That way it's more precise and we're not implying anything by it's invocation. However, maybe the implication ("I want to know how many duplicates are there") is the point.

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
```diff
@@ -67,7 +67,7 @@ class OrganizationService {
             organization = new Organization()
             organization.name = name
             organization.partyType = PartyType.findByCode(Constants.DEFAULT_ORGANIZATION_CODE)
-            organization.code = code?:identifierService.generateOrganizationIdentifier(name)
+            organization.code = code?:organizationIdentifierService.generate(name)
```
jmiranda, line 70: This does not have to happen now, but passing the organization object here (either in params or as an argument on its own) seems like the right approach.

grails-app/services/org/pih/warehouse/data/ProductSupplierIdentifierService.groovy
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Generates a new product supplier code.
+     */
+    String generate(String productCode, String organizationCode) {
+        return generate(IdentifierGeneratorParams.builder()
+                .customKeys([
+                        'productCode': productCode,
+                        'organizationCode': organizationCode,
```
jmiranda, line 26: Seeing customKeys being used for the first time, I'm wondering if there's a more intutive name for that. customKeys implies that we need to pass in a list of map keys. 

You can leave it for now, but something as simple as data or customData might suffice.

jmiranda (commented):
Sigh. I'm still not done but I wanted to post my comments so you have a chance to review.

grails-app/services/org/pih/warehouse/core/OrganizationIdentifierService.groovy
EWaterman: The TODO describes my goal here. I want to have a single generate() method that is driven entirely by properties. I'm imagining this being configured like:
```
openboxes.identifier.organization.format = "\${abbreviation}"
openboxes.identifier.organization.abbreviation.field = "name"
openboxes.identifier.organization.minSize = 2
openboxes.identifier.organization.maxSize = 3
```
then run with:
```
Organization org = new Organization(name: "A Good Name")
String id = organizationIdentifierService.generate(org)
id == "AGN"
```

grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
EWaterman: No unfortunately... I think I'm going to leave it as is for now, but I do think it's worth investigating in the future if this is actually a good idea. I'm thinking it'd be better to split them all into their own sessions/transactions.

grails-app/services/org/pih/warehouse/requisition/RequisitionIdentifierService.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+class RequisitionIdentifierService extends IdentifierService implements BlankIdentifierResolver<Requisition> {
+
+    @Override
+    String getPropertyKey() {
+        return "requisition"
+    }
+
+    @Override
+    protected Integer countDuplicates(String requestNumber) {
```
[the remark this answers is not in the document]
EWaterman: for 1 and 2, yes can do. Thanks for the context on the why. Maybe we can add a dev task in the next release to refactor this properly, or at least a discovery to determine the work required. The nice thing about the ids being the same is that it's simpler to cross-reference them, but yes if we're refactoring to not need them, we could remove the check here on other tables.

Another option would be to not check for duplicates in other tables at all, and instead of simply assigning shipmentNumber=requestNumber, we do something like:
```
shipmentNumber = shipmentIdentifierService.generate(IdentifierGeneratorParams.builder()
                .formatOverride("\${custom.requestNumber}-\${random}")
                .customKeys(['requestNumber': requisition.requestNumber])
                .build())
```
So given requestNumber ABC123, shipmentNumber would be ABC123-123456.

grails-app/services/org/pih/warehouse/data/ProductSupplierIdentifierService.groovy
EWaterman: I like customProperties

grails-app/services/org/pih/warehouse/core/OrganizationService.groovy
EWaterman: yeah I agree. Just didn't want to do the refactor at this point since it was already getting unwieldy.

grails-app/services/org/pih/warehouse/core/LocationIdentifierService.groovy
EWaterman: yeah I wanted to be specific about what we need, which is "how many copies of this id already exist in the database". Though I suppose countByIdentifier implies something similar

grails-app/services/org/pih/warehouse/core/OrganizationIdentifierService.groovy
EWaterman: yeah we should discuss. I think @Value is the recommended approach in 99% of cases. There are situations here where we need to fetch individual settings by feature type, and for those it makes more sense to use the config service since we don't want to assume which feature settings are defined since they'll vary between implementations.

also the only reason it's using the old way here is because I didn't want to change the code until I had time to revisit and properly refactor it.

grails-app/services/org/pih/warehouse/product/ProductIdentifierService.groovy
```diff
@@ -0,0 +1,201 @@
...
+     * or for configured config.openboxes.identifier.product.format
+     */
+    private String generateRandomProductIdentifier(ProductType productType) {
+        // If the product type is null or the product type does not have a custom identifier,
+        // then generate product code from DEFAULT_PRODUCT_NUMBER_FORMAT
+        if (!productType || !productType.productIdentifierFormat) {
+            return generate()
+        }
+
+        // if the product type does not contain sequential part, then generate identifier basing on custom format
```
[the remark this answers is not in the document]
EWaterman: this is all copy-pasted code so I left everything as is, including the comments

grails-app/services/org/pih/warehouse/product/ProductIdentifierService.groovy
```diff
@@ -0,0 +1,201 @@
...
+        return generate(IdentifierGeneratorParams.builder()
+                .formatOverride(productType.productIdentifierFormat)
+                .build())
+    }
```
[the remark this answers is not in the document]
EWaterman: yeah I tried to change as little of this class as possible when I brought it in so all that flow copy-pasted as is. I eventually want to refactor all of this so that it is muuuch simpler to use, but I'm saving that for future work

src/main/groovy/org/pih/warehouse/core/identification/BlankIdentifierResolver.groovy
```diff
@@ -0,0 +1,76 @@
...
+                String identifier = generateForEntity(entity)
+                if (identifier == null) {
+                    log.error("Failed to generate a unique identifier for entity ${entityType}. Terminating job early to save CPU in case we're running out of ids.")
+                    return
+                }
+
+                log.info("Assigning identifier ${identifier} to ${entityType} with id: ${entity.id}")
+                setIdentifierOnEntity(identifier, entity)
+
+                if (!entity.merge(flush: true, validate: false)) {
```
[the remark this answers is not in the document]
EWaterman: this is copy-pasted from how it was working before. I didn't want to change any of the logic, just wanted to make it commonized. We should definitely revisit it though because needing to flush and skip validation here is definitely suspect.

src/main/groovy/org/pih/warehouse/core/identification/BlankIdentifierResolver.groovy
```diff
@@ -0,0 +1,76 @@
...
+     */
+    abstract String generate(IdentifierGeneratorParams params=null)
+
+    /**
+     * Generates an identifier for the given entity.
+     * Can be overridden by child classes if they require additional logic when setting unassigned identifiers.
+     */
+    String generateForEntity(T entity) {
+        return generate(IdentifierGeneratorParams.builder()
+                .templateEntity(entity)
```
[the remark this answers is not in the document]
EWaterman: I actually originally had the generate method passing in the entity as a requirement but it got complicated because there were some cases where two different entities were sharing the same format for generating ids so it got wonky with all the typing.

I was struggling with the balance of "this should be driven 100% via properties" and "well we need to provide certain params in certain situations directly in the code". With the custom params it's kinda inevitable that we need to provide them via code, but for everything else it can be determined purely via properties.

I suppose I was trying too hard to make the identifiers not need to rely on a specific entity (by making it optional), but given that the property config is entity specific, maybe we should just lean in and say an identifier service implementation is for a specific entity, full stop. Any entity that uses an identifier must define their own service.

grails-app/conf/runtime.groovy
jmiranda: Ooh, that's a good point re: maintaining backwards compatibility. I don't actually think we know if anyone has overriden these properties so maybe I was wrong. Maybe we need to keep them for now. We can discuss or I'll leave it up to you to decide. 

Ideally, we would provide a mechanism to "deprecate" config properties to allow config property changes like this. We could keep a mapping of old config properties (to their superseding property) for backwards-compatibility. 

One brute force way to deal with these changes would be to keep track of the deprecated config properties in a list somewhere and throw an exception during startup if we encounter any of them in the grailsApplication.config. The better approach would be to fallback to these deprecated values (if they exist) but that adds complexity for testing and lifecycle management (when do they die).

grails-app/conf/runtime.groovy
```diff
@@ -2,41 +2,49 @@ import org.pih.warehouse.core.ActivityCode
...
+openboxes.identifier.default.format = "\${random}"  // By default, simply generate a random id
+openboxes.identifier.default.random.template = "NNNLLL"
+openboxes.identifier.default.random.condition = RandomCondition.ALWAYS  // By default, we always fill in random sections
+openboxes.identifier.default.delimiter = "-"
+openboxes.identifier.default.prefix.enabled = true
+
+// Transaction identifier
+openboxes.identifier.transaction.random.template = "AAA-AAA-AAA"
```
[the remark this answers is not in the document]
EWaterman: interesting. We could definitely look at something like that to make the config simpler, but I think regardless we should consider this an admin feature that needs public documentation. Do you need custom ids? Here's how to define that for your implementation.

grails-app/conf/runtime.groovy
EWaterman: In general, we could do something like fetch the new format, then if we find nothing fetch the old format. It's two lookups but only if you're on the old format.

In this case though I agree it's unlikely anyone has modified these properties so I'll just change them.

jmiranda: Sounds good.

grails-app/services/org/pih/warehouse/product/ProductIdentifierService.groovy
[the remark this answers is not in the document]
jmiranda: Sounds good.

jmiranda:
@EWaterman The tests are :woman_cook: :kiss:  - nice work. The only lingering doubt for is that we have lots of code changes so lots of avenues for regression. I think we're beyond the point at which we can turn back but if you have any lingering doubts, perhaps you and I can enumerate those fears so it's easier for testing. We can bring them up with Manon and let her decide on whether this gets merged or not.

Aside: @awalkowiak One thing we should have done from the beginning was deploy this branch to its own server for testing. I completely forgot to do that, but perhaps that's an option if Evan would prefer. 

Otherwise, if Evan feels confident, let's merge it into develop and start testing it out.

EWaterman:
@jmiranda I regression tested the identifiers on my local today (and did end up catching a few edge case errors that I fixed and wrote some more automation tests for).

I think because this is going to develop (and not a release branch) I feel confident enough about the change in that it won't break any flows and the ids all look correct. But yes we can definitely enumerate the regression testing steps to perform. I'll add that to the ticket.

One thing I'm not certain about is POs. Mine generates `PO-MA-000000` but everything I see on prod is `INTPO000000`. However looking at the old code, it seems like it should also generate with a format like mine. I'll chat with the team about it tomorrow before I merge anything in.

EWaterman:
answered my own question. PIH environments have set this in their openboxes-config.groovy:
```
openboxes.identifier.product.format = 'LLNNN'  // overrides Constants.DEFAULT_PRODUCT_NUMBER_FORMAT='LLNN'
openboxes.identifier.purchaseOrder.format = "\${destinationPartyCode}PO\${sequenceNumber}"
```
We can talk to tomorrow on how we want to handle resolving that

PR #4889, opened 2024-10-10
OBPIH-6226 Add PO UoM in the PO receipt workflow

alannadolny:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6226


grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -12,6 +12,7 @@ package org.pih.warehouse.shipping
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Person
+import org.pih.warehouse.core.UnitOfMeasure
```
kchelstowski, line 15: it seems not be used

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -296,6 +297,14 @@ class ShipmentItem implements Comparable, Serializable {
         return quantityToInvoiceInStandardUom > 0
     }
 
+    OrderItem getOrderItem() {
+        return orderItems[0]
```
kchelstowski, line 301: I'm wondering why this relationship is many-to-many

grails-app/i18n/messages.properties
```diff
@@ -3642,12 +3642,13 @@ react.partialReceiving.lotSerialNo.label=Lot/Serial No.
...
 react.partialReceiving.recipient.label=Recipient
-react.partialReceiving.shipped.label=Shipped
+react.partialReceiving.shippedInPo.label=Shipped (in PO)
+react.partialReceiving.shippedEach.label=Shipped (Each)
 react.partialReceiving.received.label=Received
 react.partialReceiving.receiving.label=Receiving
 react.partialReceiving.check.label=Check
 react.partialReceiving.toReceive.label=To receive
-react.partialReceiving.receivingNow.label=Receiving now
+react.partialReceiving.receivingNowEach.label=Receiving now (each)
```
kchelstowski, line 3651: I think for this case, the key could stay `receivingNow`.

src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -253,10 +253,21 @@ const TABLE_FIELDS = {
...
+      uom: {
+        type: (params) => <LabelField {...params} />,
+        label: 'react.partialReceiving.shippedInPo.label',
+        defaultMessage: 'Shipped (in PO)',
+        multilineHeader: true,
+        flexWidth: '1',
+        getDynamicAttr: (props) => ({
+          hide: !props.values?.isShipmentOrderBased,
+        }),
+      },
```
kchelstowski, line 265: shouldn't this field be disabled?

src/main/groovy/org/pih/warehouse/api/PartialReceipt.groovy
```diff
@@ -41,6 +41,10 @@ class PartialReceipt {
         return partialReceiptItems
     }
 
+    Boolean isShipmentOrderBased() {
+        return shipment.orders.size() > 0
```
kchelstowski, line 45: wasn't it supposed to be only applied to the shipment from PO?
if so, this should be changed, because currently it would return true also for returns and putaways.

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -73,6 +74,8 @@ class ShipmentItem implements Comparable, Serializable {
             "unavailableQuantityPicked",
             "paymentTerm",
             "quantityInvoiced",
+            "orderItem",
+            "uom",
```
kchelstowski, line 78: I don't have a strong opinion on that, but I can see that in some places we use `uom` and in some places we use `unitOfMeasure` - I think it would be good to align on one version, as more PRs related to that will come.

kchelstowski: For now, I'd say that `unitOfMeasure` is more relevant, because it's currently used in `OrderItem` and `Product` domain.

src/main/groovy/org/pih/warehouse/api/PartialReceiptItem.groovy
```diff
@@ -99,7 +104,8 @@ class PartialReceiptItem {
                 quantityRemaining               : quantityRemaining,
                 cancelRemaining                 : cancelRemaining,
                 quantityOnHand                  : quantityOnHand,
-                comment                         : comment
+                comment                         : comment,
+                uom                             : displayUom,
```
kchelstowski, line 108: the same comment as in the transients

src/js/components/receiving/PartialReceivingPage.jsx
awalkowiak: I think LabelField is "disabled" by default

src/main/groovy/org/pih/warehouse/api/PartialReceipt.groovy
awalkowiak: Could you just use `shipment.isFromPurchaseOrder` instead, if it is only for POs?

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -296,6 +297,14 @@ class ShipmentItem implements Comparable, Serializable {
         return quantityToInvoiceInStandardUom > 0
     }
 
+    OrderItem getOrderItem() {
+        return orderItems[0]
+    }
+
+    String getUom() {
```
awalkowiak, line 304: Looks like your and @drodzewicz are adding the same thing, So it would be to align on the same version (plus naming-wise I would say it should be `getUnitOfMeasure`)

awalkowiak (approved):

grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
EWaterman: I just made this same comment on the other pr...

drodzewicz (approved):

kchelstowski (approved):

PR #4891, opened 2024-10-10
OBPIH-6225 Add PO UoM in the PO shipment workflow

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6225](https://pihemr.atlassian.net/browse/OBPIH-6225)

**Description:**
On add items step:
- Added column with information about Uom (PO UoM)
- replaced previous quantity input (each) with quantity input (per UOM)
- Added  Quantity (each) column that is dynamically calculated based on provided pack size and input of quantity per UOM

On Send page
- Added quantity picked per Uom that displays amount of packs and uom picked
- renamed quantity picked column header to Quantity Picked (each)

---

image

image

[OBPIH-6225]: https://pihemr.atlassian.net/browse/OBPIH-6225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


grails-app/domain/org/pih/warehouse/shipping/ShipmentItem.groovy
```diff
@@ -224,6 +224,11 @@ class ShipmentItem implements Comparable, Serializable {
         return orderItem ? orderItem.quantityPerUom : 1
     }
 
+    String getUnitOfMeasure() {
+        OrderItem orderItem = OrderItem.get(this.orderItemId)
```
kchelstowski, line 228: Looking at how the `orderItemId` transient is implemented, I don't see a point in fetching the order item here. I'd rather just go with `orderItems[0]`

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -131,6 +133,8 @@ class StockMovementItem {
                 quantityCanceled          : quantityCanceled,
                 quantityRevised           : quantityRevised,
                 quantityPicked            : quantityPicked,
+                unitOfMeasure             : unitOfMeasure,
+                packsRequested            : quantityRequested / packSize,
```
kchelstowski, line 137: are you sure we don't have to do any additional checks in terms of dividing those two values, like checking is `packSize` is always `> 0`?

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -159,17 +159,49 @@ const FIELDS = {
           },
         }),
       },
-      quantityRequested: {
+      packsRequested: {
         type: TextField,
-        label: 'react.stockMovement.quantity.label',
-        defaultMessage: 'Qty',
+        label: 'react.stockMovement.quantityPerUom.label',
+        defaultMessage: 'Quantity (per UoM)',
```
awalkowiak, line 165: I think this should be something different. "Packs requested" is something different than "quantity per uom" ("quantity per uom" is "pack size")

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -131,6 +140,8 @@ class StockMovementItem {
                 quantityCanceled          : quantityCanceled,
                 quantityRevised           : quantityRevised,
                 quantityPicked            : quantityPicked,
+                unitOfMeasure             : unitOfMeasure,
+                packsRequested            : packsRequest,
```
awalkowiak, line 144: Can you call it `packsRequested` as well instead of `packsRequest`?

awalkowiak (changes requested):

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
drodzewicz: this name was suggested to me by Manon, I don't have any opinion about it so this needs further confirmation

src/js/components/stock-movement-wizard/combined-shipments/AddItemsPage.jsx
```diff
@@ -159,17 +159,49 @@ const FIELDS = {
...
+        },
+        getDynamicAttr: ({ rowIndex, values }) => ({
+          formatValue: () => {
+            const packsRequested = _.toInteger(
+              _.get(values, `lineItems[${rowIndex}].packsRequested`),
+            );
+            const packSize = _.toInteger(
+              _.get(values, `lineItems[${rowIndex}].packSize`),
+            );
+            return packsRequested * packSize;
```
alannadolny, line 201: Can you just use `values.lineItems[rowIndex].packsRequested` with nullsafes instead of this lodash function?

src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
```diff
@@ -202,11 +202,27 @@ const SUPPLIER_FIELDS = {
...
+        label: 'react.stockMovement.quantityPickedPOUom.label',
+        defaultMessage: 'Qty Picked (in PO UoM)',
+        multilineHeader: true,
+        fixedWidth: '100px',
+        attributes: {
+          disabled: true,
+        },
+        getDynamicAttr: ({ rowIndex, tableItems }) => ({
+          formatValue: () => {
+            const row = _.get(tableItems, `[${rowIndex}]`, {});
```
alannadolny, line 216: the same here

src/main/groovy/org/pih/warehouse/api/StockMovementItem.groovy
```diff
@@ -78,6 +80,13 @@ class StockMovementItem {
         return stocklistItem?.quantity ?: null
     }
 
+    BigDecimal getPacksRequested () {
+        if (packSize == null || packSize == 0) {
```
alannadolny, line 84: so when we are checking if packSize is null or 0, we can use !packSize

drodzewicz: I made other changes but I'd prefer to leave this as it is now, I feel like this way is more explicit to show what are we checking instead of `!packSize`

awalkowiak (approved):

PR #4892, opened 2024-10-11
OBPHI-6759 Rollback of invoice doesn't reverse order status 

alannadolny:


grails-app/migrations/views/order-summary-helper-views.sql
```diff
@@ -309,27 +309,27 @@ CREATE OR REPLACE VIEW order_summary AS (
...
+                               0                                                                             AS quantity_invoiced,
+                               0                                                                             AS items_invoiced,
+                               IFNULL(SUM(order_adjustment_payment_statuses.quantity_invoiced), 0)           AS adjustments_invoiced,
+                               SUM(DISTINCT total_adjustments.total_adjustments)                             AS total_adjustments,
+                               IFNULL(SUM(ABS(order_adjustment_payment_statuses.invoiced_amount)), 0)        AS invoiced_adjustments_amount
                   FROM `order`
                            LEFT OUTER JOIN order_adjustment ON order_adjustment.order_id = `order`.id
                            LEFT OUTER JOIN (
-                                SELECT adjustment_id, SUM(invoiced_amount) as invoiced_amount
+                                SELECT adjustment_id, SUM(invoiced_amount) as invoiced_amount, IF(SUM(quantity_invoiced) > 0, 1, 0) as quantity_invoiced
```
awalkowiak, line 332: Does this work properly for the partial amount invoicing? Let's say you have the adjustment invoiced in two or more invoices, and if it is fully and not yet fully invoiced?

alannadolny: Yeah, it works, because we are looking into the amount and the quantity_invoiced that you are pointing out here.
When we are starting invoicing an adjustment it is already included in this count, but splitting it between a few invoices won't increase the counter, so we are not able to have more adjustments than we had on the PO as it was possible at the beginning. It's a way for handling adjustments with 0 amount because we are not able to split them between invoices, so as an example:
3 adjustments, first with 5 amount, second with 5, and the last one with 0:
- I am invoicing the first adj, but only 3 amount -> total amount: 10, invoiced amount 3, quantity invoiced 1, adj count: 3 [PARTIALLY_INOVICED]
- I am invoicing the first adj second time,  2 amount -> total amount: 10, invoiced amount 5, quantity invoiced 1, adj count: 3 [PARTIALLY_INOVICED]
- I am invoicing the second adj, full amount -> total amount: 10, invoiced amount 10, quantity invoiced 2, adj count: 3 [PARTIALLY_INOVICED]
- I am invoicing the last adj -> total amount: 10, invoiced amount 10, quantity invoiced 3, adj count: 3 [INVOICED]

awalkowiak (approved):

PR #4899, opened 2024-10-17
OBPIH-6226 Add PO UoM in the PO receipt workflow

alannadolny:


src/main/groovy/org/pih/warehouse/api/PartialReceiptItem.groovy
```diff
@@ -71,7 +71,7 @@ class PartialReceiptItem {
     }
 
     String getDisplayUnitOfMeasure() {
-        return "${shipmentItem?.quantityPerUom?.toInteger()} ${shipmentItem?.unitOfMeasure}"
+        return "${shipmentItem?.orderItem?.quantity} ${shipmentItem?.unitOfMeasure}"
```
awalkowiak, line 74: @alannadolny Should this show how much was shipped or how much was ordered?

alannadolny: ordered, shipped is already multiplied (qty * qty per uom)
when I order 5 quantities of something in a pack of 3, I want to display:
5 PACK/3, and then the quantity shipped is displayed in "Shipped (each)" column as 15

awalkowiak: But what if it is partial shipment of part of the ordered quantity?

awalkowiak (approved):

PR #4900, opened 2024-10-17
OBPIH-6267 Do not select edited lines to receive automatically in par…

kchelstowski:
…tial receiving workflow

**Description:**

---

### DIFF
--- a/src/js/components/receiving/PartialReceivingPage.jsx
+++ b/src/js/components/receiving/PartialReceivingPage.jsx
@@ -735,6 +735,15 @@ class PartialReceivingPage extends Component {
     }));
   }
 
+  static mapContainers(containers, parentIndex, items) {
+    return containers.map((container, idx) => {
+      if (idx === parentIndex) {
+        return { ...container, shipmentItems: [...container.shipmentItems, ...items] };
+      }
+      return container;
+    });
+  }
+
   /*
    * Saves changes made in edit line modal and updates data.
    * @param {object} editLines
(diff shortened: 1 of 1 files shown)


src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -735,33 +735,134 @@ class PartialReceivingPage extends Component {
...
+        return {
+          ...acc,
+          originalItem: item,
+        };
+      }
+      return {
+        ...acc,
+        itemsAfterCurrentRow: [...acc.itemsAfterCurrentRow, item],
+      };
+    }, { itemsBeforeCurrentRow: [], originalItem: null, itemsAfterCurrentRow: [] });
```
alannadolny, line 758: I think you can simplify that code using `_.groupBy` from lodash

src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -735,33 +735,134 @@ class PartialReceivingPage extends Component {
...
+      if (idx === parentIndex) {
+        const { shipmentItems } = container;
+        return {
+          ...container,
+          shipmentItems: PartialReceivingPage.mapShipmentItems(shipmentItems, rowIndex, items),
+        };
+      }
+      return container;
+    });
+  }
```
alannadolny, line 806: I think it would be better to move this function to some utils instead of creating a static function ... static functions in React feels odd to me

alannadolny (commented):
Looking at those changes, I am getting more and more confident that this should be refactored asap, because each of our changes is rather fixing using duct tape, not in the way it should be done.

src/js/components/receiving/PartialReceivingPage.jsx
kchelstowski: I don't know how I would be able to use the `_.groupBy` from lodash here - from what I can see in the [docs](https://lodash.com/docs/4.17.15#groupBy), I don't find it useful in that specific case, where I base the "key" on whether the index is lower, equal or higher than x.
I believe that this `.reduce` is quite readable, isn't it?

PR #4901, opened 2024-10-18
OBPIH-6369 Receiving : Receipt not complete after using "Edit" to change item

alannadolny:


src/test/groovy/unit/org/pih/warehouse/receiving/ReceiptServiceSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+@Unroll
+class ReceiptServiceSpec extends Specification implements ServiceUnitTest<ReceiptService>, DataTest {
+
+    @Shared
+    ShipmentService shipmentService
+
+    void setupSpec() {
+        mockDomain Receipt
+        mockDomain Shipment
```
EWaterman, line 25: for future reference, you can do `mockDomains(Receipt, Shipment)`

src/test/groovy/unit/org/pih/warehouse/receiving/ReceiptServiceSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+        shipmentService = Mock(ShipmentService) {
+            createShipmentEvent(_, _, _, _) >> void
+        }
+        service.shipmentService = shipmentService
+    }
+
+    void 'savePartialReceiptEvent should create RECEIVED event when partial receiving is not supported'() {
+        given:
+        Location location = Stub(Location) {
+            supports(_ as ActivityCode) >> false
```
EWaterman, line 38: Is there any reason not to specify the actual enum value here? It makes things a bit more clear in my opinion and also guarantees that we're entering the proper flow.

Same for all the other tests.

```suggestion
            supports(ActivityCode.PARTIAL_RECEIVING) >> false
```

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -287,26 +294,39 @@ class ReceiptService {
...
 
-        if (shipment.isFullyReceived()) {
-            if (!shipment.wasReceived()) {
-                shipmentService.createShipmentEvent(shipment,
-                        receipt?.actualDeliveryDate,
-                        EventCode.RECEIVED,
-                        shipment.destination)
-            }
-        } else {
+        if (!shipment.wasReceived() && (!shipment.destination.supports(ActivityCode.PARTIAL_RECEIVING) || shipment.isFullyReceived())) {
```
EWaterman, line 313: why are we even allowing the flow to continue here if partial receiving is disabled for the location? Shouldn't we error instead?

Based on the fact that we don't error I'm guessing it's because regular receiving flows also use this method, which is misleading IMO based on the method name if that's true. Anyways, not new code so nothing I expect you to change.

src/test/groovy/unit/org/pih/warehouse/receiving/ReceiptServiceSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+import org.pih.warehouse.receiving.Receipt
+import org.pih.warehouse.receiving.ReceiptService
+import org.pih.warehouse.shipping.Shipment
+import org.pih.warehouse.shipping.ShipmentService
+import spock.lang.Shared
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Unroll
+class ReceiptServiceSpec extends Specification implements ServiceUnitTest<ReceiptService>, DataTest {
```
EWaterman, line 18: Great job writing tests!

It'd probably also be worth testing these scenarios:
1. partial receiving is enabled, it was partially received before, and now it's fully received
2. partial receiving is enabled, it was partially received before, and it still is (no events should happen)
3. partial receiving is enabled, it was fully received before, and it still is (no events should happen)

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
EWaterman, line 313: also, I'm not totally sure about this new flow. The old flow looked correct to me. Mainly:
```
// We're fully receiving
if (fully received now) {
  if (it was not already in fully received status) {
    create fully received event
  }
}
// We're partially receiving
else if (it was not already in partially received status) {
  create partial received event
}
```
so I'm wondering if there's an issue not with this class but with `shipment.isFullyReceived()`. Or maybe I'm just misunderstanding. I haven't had a chance to dig much deeper into the code

alannadolny: I don't know why we call this feature "partial" everywhere, because all those functions simultaneously support partial / not partial receiving. That's the reason for allowing it here; if partial receiving is disabled, we can receive it only once.

awalkowiak (approved):

PR #4916, opened 2024-10-25
OBPIH-6707 Indicator when quantities in split line do not sum up to o…

kchelstowski:
…riginal qty

**Description:**

---

### DIFF
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -3678,6 +3678,7 @@ react.partialReceiving.error.invalidDate.label=This date is invalid. Please ente
 react.partialReceiving.error.expiryWithoutLot.label=Items with an expiry date must also have a lot number.
 react.partialReceiving.error.lotAndExpiryControl.label=Both lot number and expiry date are required for this item.
 react.partialReceiving.confirmReceive.notSupportingPartialReceiving.message=You have not entered a value in the receiving quantity for ${count} line/lines on this shipment. These lines will be received as zero. You will not be able to go back and receive them later. Do you want to continue?
+react.partialReceiving.mismatchingQuantityShipped.label=The total edited quantity does not match the original quantity shipped.
 
 # PutAway
 react.putAway.createPutAway.label=Create Putaway
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -3678,7 +3678,7 @@ react.partialReceiving.error.invalidDate.label=This date is invalid. Please ente
 react.partialReceiving.error.expiryWithoutLot.label=Items with an expiry date must also have a lot number.
 react.partialReceiving.error.lotAndExpiryControl.label=Both lot number and expiry date are required for this item.
 react.partialReceiving.confirmReceive.notSupportingPartialReceiving.message=You have not entered a value in the receiving quantity for ${count} line/lines on this shipment. These lines will be received as zero. You will not be able to go back and receive them later. Do you want to continue?
-react.partialReceiving.mismatchingQuantityShipped.label=The total edited quantity does not match the original quantity shipped.
+react.partialReceiving.error.mismatchingQuantityShipped.label=The total edited quantity does not match the original quantity shipped.
 
 # PutAway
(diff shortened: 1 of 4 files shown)


grails-app/i18n/messages.properties
```diff
@@ -3678,6 +3678,7 @@ react.partialReceiving.error.invalidDate.label=This date is invalid. Please ente
 react.partialReceiving.error.expiryWithoutLot.label=Items with an expiry date must also have a lot number.
 react.partialReceiving.error.lotAndExpiryControl.label=Both lot number and expiry date are required for this item.
 react.partialReceiving.confirmReceive.notSupportingPartialReceiving.message=You have not entered a value in the receiving quantity for ${count} line/lines on this shipment. These lines will be received as zero. You will not be able to go back and receive them later. Do you want to continue?
+react.partialReceiving.mismatchingQuantityShipped.label=The total edited quantity does not match the original quantity shipped.
```
alannadolny, line 3681: I think we should follow one pattern when naming those labels. I saw that we are using `.error.` in the labels.
So maybe we can call it like: `react.partialReceiving.error.mismatchingQuantityShipped.label`

src/js/components/receiving/modals/EditLineModal.jsx
```diff
@@ -86,10 +86,22 @@ class EditLineModal extends Component {
     } = props;
     const dynamicAttr = getDynamicAttr ? getDynamicAttr(props) : {};
     const attr = { ...attributes, ...dynamicAttr };
+    const groupedShipmentItems = this.groupShipmentItems();
+    const shipmentItemsQuantityMap = Object.entries(groupedShipmentItems)
+      .reduce((acc, [key, value]) =>
+        ({
+          ...acc,
+          [key]: value.reduce((sum, curr) => (sum + _.toInteger(curr.quantityShipped)), 0),
```
alannadolny, line 94: you can just use `sum` or `sumBy` instead of using nested reduce.

src/js/components/receiving/modals/EditLineModal.jsx
```diff
@@ -108,11 +120,25 @@ class EditLineModal extends Component {
...
 
+  groupShipmentItems() {
+    const { containers } = this.props.values;
+    /**
+     * containers are built like: [{..., shipmentItems: []}]
+     * so in the end we have to flat the result,
+     * as after mapping, the result would look like
+     * [[<shipmentItem>, <shipmentItem>], [<shipmentItem>, <shipmentItem>]]
+     */
+    const shipmentItems = containers.map((container) => container.shipmentItems).flat();
```
alannadolny, line 131: what about `_.flatten`?

src/js/components/receiving/modals/EditLineModal.jsx
```diff
@@ -187,7 +220,56 @@ class EditLineModal extends Component {
     });
   }
 
+  calculateQuantityShippedSum(values) {
+    const { shipmentItemQuantityShippedSum } = this.state;
+    const originalItem = values.find((item) => item.rowId);
+    const qtyShippedSumFromModal =
+      values.reduce((acc, curr) => acc + _.toInteger(curr.quantityShipped), 0);
```
alannadolny, line 227: `_sum` or `_sumBy`

grails-app/i18n/messages.properties
kchelstowski: yeah I agree, I never feel comfortable adding a new label because I always wonder if any pattern is applicable there.

src/js/components/receiving/modals/EditLineModal.jsx
kchelstowski: I don't see any advantage of that, since `flat()` does the same

alannadolny: you're right, so I think you can use `_.flatMap` instead :smile:

kchelstowski: :thinking: I swear I checked whether the `.flatMap` exists in JS, but couldn't find it, and suddenly it exists :laughing:

awalkowiak (approved):

PR #4917, opened 2024-10-25
OBPIH-6775 Add information bar about import workflow on normal outbound workflow

alannadolny:


src/js/components/form-elements/FieldArrayComponent.jsx
```diff
@@ -123,9 +123,9 @@ class FieldArrayComponent extends Component {
                         <div>
                           {this.props.translate(config.headerTooltip, config.headerDefaultTooltip)}
                         </div>
-                      ) : (config.label &&
+                      ) : (headerLabel &&
```
kchelstowski, line 126: I'm a bit scared of such changes to generic components. Have you made sure nothing else broke?

awalkowiak: +1, @alannadolny?

drodzewicz (approved):

src/js/components/form-elements/FieldArrayComponent.jsx
alannadolny: yup, it works

PR #4923, opened 2024-10-31
OBPIH-6707 Fix quantity shipped indicator on edit modal when a brand …

kchelstowski:
…new row is added to the table

**Description:**
The comments before calling the `groupShipmentItems` sort of describe the issue that was caught by Jakub, but in a nutshell - constructor is called only once, so in the constructor I tried to rely on initial values of the form (original qty), and then, while opening a modal, relying on the current values of the form. The problem was that when we add a new row at the end (e.g. we edit the last item), this modal has not yet been built in the background, hence when creating it and having already edited an item (with decreased/increased qty shipped), the new row would look at it as the original qty and the indicator would not be shown in that line.
To fix this, I need to store the initial receipt candidates values separately, not to rely on the values of the form in the constructor.

---

### DIFF
--- a/src/js/components/receiving/PartialReceivingPage.jsx
+++ b/src/js/components/receiving/PartialReceivingPage.jsx
@@ -579,7 +579,10 @@ class PartialReceivingPage extends Component {
     return apiClient.get(url)
       .then((response) => {
         this.setState({ values: {} }, () => {
-          this.setState({ values: parseResponse(response.data.data) });
+          this.setState({
+            values: parseResponse(response.data.data),
+            initialReceiptCandidates: parseResponse(response.data.data),
+          });
         });
       })
       .catch(() => this.props.hideSpinner());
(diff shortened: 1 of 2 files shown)


src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -579,7 +579,10 @@ class PartialReceivingPage extends Component {
     return apiClient.get(url)
       .then((response) => {
         this.setState({ values: {} }, () => {
-          this.setState({ values: parseResponse(response.data.data) });
+          this.setState({
+            values: parseResponse(response.data.data),
+            initialReceiptCandidates: parseResponse(response.data.data),
```
alannadolny, line 584: why do we need to clear the values and then assign a value to this property?

kchelstowski: ask the one implementing it initially :see_no_evil: 
I didn't even see this clearing before, as I only cared about the place where I needed to pass the additional property.
Let's leave it as it is, because I don't know what potentially would be broken by that....

alannadolny: :thinking:

awalkowiak: > why do we need to clear the values and then assign a value to this property? 

This is a really good question, this clearing values seems to be pointless here, and I don't know/remember any reason why it went in. I guess we could remove it and just check out if the page loads correctly.

> ask the one implementing it initially 🙈

:/

awalkowiak (approved):

PR #4925, opened 2024-11-04
OBPIH-6765 Add "Desired Date of Delivery" for E-Requests

alannadolny:


grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -46,6 +46,7 @@ class Requisition implements Comparable<Requisition>, Serializable {
     Date dateDelivered
     Date dateIssued
     Date dateReceived
+    Date desiredDeliveryDate
```
awalkowiak, line 49: Looks like our convention is `date<Something>`, but I don't know how this should look like for this one. @jmiranda do you have any preference or suggestion?

src/main/groovy/org/pih/warehouse/api/StockMovement.groovy
```diff
@@ -171,6 +172,7 @@ class StockMovement implements Validateable{
             dateCreated         : dateCreated?.format("MM/dd/yyyy"),
             dateShipped         : dateShipped?.format("MM/dd/yyyy HH:mm XXX"),
             expectedDeliveryDate: expectedDeliveryDate?.format("MM/dd/yyyy HH:mm XXX"),
+            desiredDeliveryDate : requisition?.desiredDeliveryDate?.format("MM/dd/yyyy"),
```
EWaterman, line 175: why requisition and not the desiredDeliveryDate field you added?

grails-app/views/email/_approvalsAlert.gsp
```diff
@@ -11,11 +12,32 @@
...
-                       requisition.requestedBy,
-                       g.createLink(uri: redirectUrl, absolute: true),
+                args="[requisition.destination]"
+        />
+        %{--
+             If there is a desired delivery date we want to add a comma at the end of this message, because there is a
+             second part of the sentence. If there is no delivery date, we just finish the sentence after requestCreatedBy
+             message. It cannot be included in HTML alone because there will be additional space around the punctuation mark.
+         --}%
+        <g:set var="punctuationMark" value="${requisition.desiredDeliveryDate ? ',' : '.'}" />
```
awalkowiak, line 22: I get the idea here, but this looks odd. Would it not be clearer if it was just a `{requisition.desiredDeliveryDate ? ',' : '.'}` in between two `g:message` tags instead of an additional argument?

awalkowiak: or the other option would be always having the 
```You have received a new request from: {0}, created by: {1}, desired date of delivery: {2}. Click <a href="{2}">here</a> to review your requests pending your approval.```
and `{2}` would get `requisition.desiredDeliveryDate ?: 'unknown'`, or something like that.

File: src/main/groovy/org/pih/warehouse/api/StockMovement.groovy (hunk 2 above)
Comment (kchelstowski) [reply src/main/groovy/org/pih/warehouse/api/StockMovement.groovy:175]: I think we wanted to get rid of formatting a date on the backend and let the frontend handle the eventual formatting.
How about removing `.format` here and formatting it with Moment on the frontend?

File: grails-app/views/email/_approvalsAlert.gsp (hunk 3 above)
Comment (alannadolny) [reply grails-app/views/email/_approvalsAlert.gsp:22]: To your second comment: I asked Manon about that before and she answered it would be better to not show the date if it is null.
To your first comment: As you can see I left a comment "It cannot be included in HTML alone because there will be additional space around the punctuation mark."

Comment (awalkowiak) [reply grails-app/views/email/_approvalsAlert.gsp:22]: ok

File: grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy (hunk 1 above)
Comment (jmiranda) [reply grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy:49]: @awalkowiak @alannadolny I will have an answer for you today.

Comment (jmiranda) [reply grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy:49]: In the future, let's flag all tickets that require data model changes during the planning poker activity. All data model changes (no matter how small) must go through a design process. And this should almost always be done before we start working on the feature. 

cc @awalkowiak @EWaterman

Comment (jmiranda) [reply grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy:49]: I think dateDeliveryRequested makes the most sense. 

Besides the naming, the reason I mentioned "design" was because I would have implemented this a little more generically at the data model and API level. The requested delivery date should also be available at the requisition line item level, even if the UI of the current feature does not support it.

File: grails-app/migrations/0.9.x/changelog.xml [hunk 4]
```diff
@@ -17,4 +17,5 @@
     <include file="0.9.x/changelog-2024-08-19-1200-alter-table-invoice-item-set-invoice-non-null.xml" />
     <include file="0.9.x/changelog-2024-08-20-0000-migrate-prepayment-invoices-add-amount-and-inverse-items.groovy" />
     <include file="0.9.x/changelog-2024-10-04-1610-assign-amount-to-invoice-items.xml" />
+    <include file="0.9.x/changelog-2024-10-28-1330-add-desired-delivery-date-field.xml" />
```
jmiranda, line 20: make cure the prefix here is "add-column-..."

awalkowiak (approved):

PR #4931, opened 2024-11-08
OBPIH-5739 Fix filtering for handling requirements in products list page

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -362,17 +362,19 @@ class ProductService {
             }
 
             if (!handlingRequirements.empty) {
-                if (handlingRequirements.contains(ProductField.COLD_CHAIN)) {
-                    eq("coldChain", true)
-                }
-                if (handlingRequirements.contains(ProductField.CONTROLLED_SUBSTANCE)) {
-                    eq("controlledSubstance", true)
-                }
-                if (handlingRequirements.contains(ProductField.HAZARDOUS_MATERIAL)) {
-                    eq("hazardousMaterial", true)
-                }
-                if (handlingRequirements.contains(ProductField.RECONDITIONED)) {
-                    eq("reconditioned", true)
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -362,17 +362,19 @@ class ProductService {
...
+                    }
+                    if (handlingRequirements.contains(ProductField.CONTROLLED_SUBSTANCE)) {
+                        eq("controlledSubstance", true)
+                    }
+                    if (handlingRequirements.contains(ProductField.HAZARDOUS_MATERIAL)) {
+                        eq("hazardousMaterial", true)
+                    }
+                    if (handlingRequirements.contains(ProductField.RECONDITIONED)) {
+                        eq("reconditioned", true)
+                    }
```
alannadolny, line 377: What do you think about doing it more generic? I mean that now when we are adding a new handling requirement we have to add it to the product domain, to the ProductField enum, and to this place to make the filtering work. So maybe we can add a function for listing all of the handling requirements - something similar to what we have in the RequisitionStatus enum (listOubtoundOptions, listOutboundOptionsWhenApprovalRequired, listRequestOptionsWhenApprovalRequired, etc.). Then we can do a loop over this list and if the current requirement from the loop is in handling requirement we can do `eq(elemenetFromLoop, true)`, so doing it this way we don't have to handle adding new requirements in this place. (I believe we can do a loop here)

cc @awalkowiak

awalkowiak: (my thought - you have valid point, but imho this is fine for now, because I think there are no plans to add new handling requirements :thinking:)

kchelstowski: I had the same feeling, initially I considered doing this with a generic method but then I realized that we wouldn’t achieve much by that, since those are only 4 lines of the code with a low risk to expand

PR #4932, opened 2024-11-08
OBPIH-6099 Fix issue with global recipient on outbound autosave

drodzewicz:

**Link to GitHub issue or Jira ticket:** /[OBPIH-6099](https://pihemr.atlassian.net/browse/OBPIH-6099)

**Description:** Issue was that on outbound workflow when autosave was enabled this globalr ecipient did not work propetly

Current autosave behacior only includes logic for individual edits on the row, since we are working with a "bulk" update I had to add some modifications like thi `editAll` prop

---

[OBPIH-6099]: https://pihemr.atlassian.net/browse/OBPIH-6099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -1065,7 +1065,7 @@ class AddItemsPage extends Component {
 
     // We don't want to save the item during editing or
     // when there is an error in line
-    if (isEdited) {
+    if (isEdited || editAll) {
       this.debouncedSave();
       return;
     }
--- a/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
+++ b/src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
@@ -189,6 +189,15 @@ const NO_STOCKLIST_FIELDS = {
               onChange={(val) => {
                 if (val) {
                   setRecipientValue(val);
+                  saveProgress({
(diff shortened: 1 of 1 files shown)


src/js/components/stock-movement-wizard/outbound/AddItemsPage.jsx
```diff
@@ -1034,6 +1043,10 @@ class AddItemsPage extends Component {
     // rowIndex is present
     const isEdited = rowIndex !== undefined;
     const itemsWithStatuses = values.lineItems.map((item) => {
+      if (editAll) {
+        return { ...item, rowSaveStatus: RowSaveStatus.PENDING };
+      }
+
```
alannadolny, line 1049: I can see that the line below has an additional condition when `RowSaveStatus.PENDING` is applied. Have you checked what will happen when editing the same row multiple times?

drodzewicz: in that regard everything should work properly.
Addityionaly I have added the `editAll` condition to the following statement.
```jsx
if (isEdited || editAll) {
  this.debouncedSave();
  return;
}
```
this should ensure a safe items saveing step if there are pending save processes from previous actions.

alannadolny (approved):

PR #4933, opened 2024-11-08
OBPIH-6649 Add support for Jacoco code coverage reporting

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6649

**Description:** Generates a code coverage report using Jacoco whenever tests are run and then uploads them to Codecov when commits are pushed to PRs or when PRs are merged to develop.

Codecov dashboard: https://app.codecov.io/gh/openboxes/openboxes/

image

image

And this is what the raw HTML Jacoco reports that codecov is built from look like (we probably won't use these much except for local testing):
Screenshot from 2024-11-07 17-34-05

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -31,18 +31,16 @@ jobs:
       # tests into a separate job and run the matrix only on that job (because unit tests don't care about the
       # underlying database and so running them against multiple databases is pointless), but matrix jobs are run
       # in parallel so it wouldn't save us any running time to do that. Plus it would overcomplicate the flow.
-      - name: Run All Tests    
-        if: ${{ strategy.job-index != 0 }}
+      - name: Run All Tests
         run: TEST_DATABASE=${{ matrix.testdatabase }} ./grailsw test-app --non-interactive --no-daemon
 
       # Tests are run for each database in the above matrix but we only want to generate a code coverage report once.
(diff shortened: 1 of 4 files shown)


EWaterman:
The Codecov report comment here should be more interesting once we merge to develop and it has something to compare to. I still need to configure exactly what the comment displays (I want to make sure it does a diff on changed files and doesn't show the whole coverage) but I'll leave that for the next PR.

awalkowiak (commented):
Overall, looks good to me. Can we get a list of all files that it looks into? Is it everything from the `build/classes/groovy/main`? Locally I see some changelogs there and taglibs. 5% coverage seems small? It kinda aligns with the ~5% that we had in the 0.8.x, but feels like we bumped up the coverage by at least a few percentage points recently? :thinking:

EWaterman:
@awalkowiak yeah everything under `build/classes/groovy/main` is included. Jacoco calculates coverage slightly differently than Cobretura does (which is what we were using in Grails 1 but has since fallen out of support) so I'm not surprised the percentage is somewhat off. We could do some more filtering out if we want (such as excluding all *TagLib) but I think it's fine that the number isn't perfect. The main goal is to have coverage gradually going up.

EWaterman:
Actually you were right! The filter condition I had on closures was causing all closures to appear as 0% covered. We're at 7.5% now which feels more accurate.

image

jmiranda (approved):
I am speechless. :heart_hands:

PR #4940, opened 2024-11-13
OBPIH-6804 Build shipment history to display shipment events

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/core/Event.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Event.groovy
@@ -18,7 +18,7 @@ package org.pih.warehouse.core
  *{eventDate: 1/1/2010, eventLocation: Boston, eventType: SHIPPED}*
  *  Shipment #2 Arrived at Customs on 5/5/2010:
  *{eventDate: 5/5/2010, eventLocation: Customs, eventType: ARRIVED}*/
-class Event implements Comparable, Serializable {
+class Event implements Comparable, Serializable, Historiable {
 
     String id
     Date eventDate                // The date and time on which the Event occurred
@@ -40,6 +40,18 @@ class Event implements Comparable, Serializable {
         eventLocation(nullable: true)
         createdBy(nullable: true)
         comment(nullable: true)
+    }
+
+    @Override
+    List<HistoryItem<Event>> getHistory() {
+        HistoryItem<Event> historyItem = new HistoryItem<>(
+                identifier: id,
+                date: eventDate,
+                associatedLocation: eventLocation,
+                parentObject: this,
+                eventCode: eventType?.eventCode
+        )
+       return [historyItem]
     }
 
     String toString() { return "$eventType $eventLocation on $eventDate" }
--- a/grails-app/domain/org/pih/warehouse/core/Event.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Event.groovy
@@ -18,7 +18,7 @@ package org.pih.warehouse.core
  *{eventDate: 1/1/2010, eventLocation: Boston, eventType: SHIPPED}*
  *  Shipment #2 Arrived at Customs on 5/5/2010:
  *{eventDate: 5/5/2010, eventLocation: Customs, eventType: ARRIVED}*/
-class Event implements Comparable, Serializable, Historiable {
+class Event implements Comparable, Serializable, Historizable {
 
     String id
     Date eventDate                // The date and time on which the Event occurred
@@ -43,13 +43,22 @@ class Event implements Comparable, Serializable, Historiable {
     }
 
     @Override
+    ReferenceDocument getReferenceDocument() {
+        return new ReferenceDocument(
+                label: eventType?.description,
+                id: id,
+                identifier: id,
+        )
+    }
+
+    @Override
     List<HistoryItem<Event>> getHistory() {
         HistoryItem<Event> historyItem = new HistoryItem<>(
                 identifier: id,
                 date: eventDate,
-                associatedLocation: eventLocation,
-                parentObject: this,
-                eventCode: eventType?.eventCode
+                location: eventLocation,
+                eventCode: eventType?.eventCode,
+                referenceDocument: getReferenceDocument()
         )
        return [historyItem]
     }
--- a/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
+++ b/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
@@ -10,13 +10,16 @@
 package org.pih.warehouse.receiving
 
 import grails.util.Holders
+import org.pih.warehouse.core.EventCode
+import org.pih.warehouse.core.Historiable
+import org.pih.warehouse.core.HistoryItem
 import org.pih.warehouse.core.Person
 import org.pih.warehouse.inventory.Transaction
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.RefreshOrderSummaryEvent
 import org.pih.warehouse.shipping.Shipment
 
-class Receipt implements Serializable, Comparable<Receipt> {
+class Receipt implements Serializable, Comparable<Receipt>, Historiable {
 
     def publishRefreshEvent() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshOrderSummaryEvent(this))
@@ -88,5 +91,14 @@ class Receipt implements Serializable, Comparable<Receipt> {
         return receiptItems?.sort(receiptItemsComparator)
     }
 
-
+    @Override
+    List<HistoryItem<Receipt>> getHistory() {
+        HistoryItem<Receipt> historyItem = new HistoryItem<>(
+                identifier: receiptNumber,
+                date: actualDeliveryDate,
+                associatedLocation: shipment.destination,
+                parentObject: this
+        )
+        return [historyItem]
+    }
 }
--- a/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
+++ b/grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
@@ -10,16 +10,16 @@
 package org.pih.warehouse.receiving
 
 import grails.util.Holders
-import org.pih.warehouse.core.EventCode
-import org.pih.warehouse.core.Historiable
+import org.pih.warehouse.core.Historizable
 import org.pih.warehouse.core.HistoryItem
 import org.pih.warehouse.core.Person
+import org.pih.warehouse.core.ReferenceDocument
 import org.pih.warehouse.inventory.Transaction
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.RefreshOrderSummaryEvent
 import org.pih.warehouse.shipping.Shipment
 
-class Receipt implements Serializable, Comparable<Receipt>, Historiable {
+class Receipt implements Serializable, Comparable<Receipt>, Historizable {
 
     def publishRefreshEvent() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshOrderSummaryEvent(this))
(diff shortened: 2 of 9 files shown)


src/main/groovy/org/pih/warehouse/core/HistoryItem.groovy
```diff
@@ -0,0 +1,15 @@
...
+
+    Date date
+
+    // FIXME: I didn't know what name suits the best for the location
+    Location associatedLocation
+
+    T parentObject
+
+    // FIXME: Introduced this property to be able to determine which history item is pointing to e.g. received event, which one to shipped event etc.
+    EventCode eventCode
```
EWaterman, line 14: So the idea is that this HistoryItem is generic enough to be applied to all features that need history? What if we need the history item to contain some feature-specific data?

Maybe it'd be more flexible if each feature was forced to implement their own ShipmentHistoryItem which extends HistoryItem.

Or maybe that's not needed and I'm overthinking it. I suppose my question is what does the future of this class look like?

jmiranda (commented):
Adding comments ahead of final review.

src/main/groovy/org/pih/warehouse/core/HistoryItem.groovy
```diff
@@ -0,0 +1,15 @@
+package org.pih.warehouse.core
+
+class HistoryItem<T> {
+    String identifier
+
+    Date date
+
+    // FIXME: I didn't know what name suits the best for the location
+    Location associatedLocation
```
jmiranda, line 9: Let's just call this location for now. Or facility if we want to be specific about what we're talking about.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -697,5 +697,51 @@ class Shipment implements Comparable, Serializable {
...
+    List<HistoryItem> getHistory() {
+        List<HistoryItem> histories = []
+        // First collect history of CREATED event
+        histories.add(new HistoryItem<Shipment>(
+                identifier: shipmentNumber,
+                date: dateCreated,
+                associatedLocation: origin,
+                parentObject: this,
+                // FIXME: Don't know what event code suits the best for CREATED
+                eventCode: EventCode.SCHEDULED
```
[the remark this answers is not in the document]
jmiranda: Adding EventCode.CREATED sounds good to me.

src/main/groovy/org/pih/warehouse/core/HistoryItem.groovy
```diff
@@ -0,0 +1,15 @@
+package org.pih.warehouse.core
+
+class HistoryItem<T> {
+    String identifier
```
jmiranda, line 4: The details from the "reference document" probably move into the ~HistoryItemLink~ ~HistoryItemDetails~ ReferenceDocument class (id, name, URL) discussed below. 

> [!NOTE]  
> A reference document is a generic term used in ERPs and WMSs to generically define the object that 

I'm also fine with those properties living in the HistoryItem object but at some point (when we do the same for stock history) we'll probably have two of these objects (one for the transaction, one for the reference document) so we'll want to refactor at that point anyway.

We probably don't need an identifier here, just the URL (which requires the UUID) and label (which requires the identifier) used for displaying the anchor tag. So here's how i would probably model the generic history item for now.

HistoryItem
```
Date date
Location location
ReferenceDocument referenceDocument
```
ReferenceDocument (alternatives: HistoryItemLink, HistoryItemDetails)
```
// Required to render a link
String label // label to display with URL
String url // URL for the link

// These might be useful, but probably aren't necessary
String id  // uuid assigned by database
String identifier  // human readable identifier
String name  // long human unreadable name
String description // extra description if necessary
```

The actual reference document class (Shipment, Receipt) should be responsible for generating these objects because they will know how to generate a link to the correct resource (i.e. stock movement details page)
```
<a href="${historyItem.referenceDocument.url}">${historyItem.referenceDocument.label}</a>
```

We're already doing this, but just wanted to show why it's necessary and suggest there might be room for improvement here. For example, the Shipment.getHistory() is pretty gnarly (my fault) so maybe there's a way to delegate some of that work to the dependent objects.

grails-app/domain/org/pih/warehouse/core/Event.groovy
```diff
@@ -18,7 +18,7 @@ package org.pih.warehouse.core
  *{eventDate: 1/1/2010, eventLocation: Boston, eventType: SHIPPED}*
  *  Shipment #2 Arrived at Customs on 5/5/2010:
  *{eventDate: 5/5/2010, eventLocation: Customs, eventType: ARRIVED}*/
-class Event implements Comparable, Serializable {
+class Event implements Comparable, Serializable, Historiable {
```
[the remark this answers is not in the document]
EWaterman: Historizable seems fine to me

grails-app/domain/org/pih/warehouse/receiving/Receipt.groovy
```diff
@@ -10,13 +10,16 @@
 package org.pih.warehouse.receiving
 
 import grails.util.Holders
+import org.pih.warehouse.core.EventCode
+import org.pih.warehouse.core.Historiable
+import org.pih.warehouse.core.HistoryItem
```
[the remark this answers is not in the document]
jmiranda: I'm actually fine with core for now.

grails-app/domain/org/pih/warehouse/core/Event.groovy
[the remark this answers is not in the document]
jmiranda: Same. I think if we end up building more history-related features we can switch from a generic Historizable to a more feature-specific naming convention around *HistoryProvider so like StockHistoryProvider, EventHistoryProvider, AuditHistoryProvider each defining getHistory() methods that return a list of HistoryItem objects subclasses for each feature.

kchelstowski:
It's ready for final review.
An example of history object:
```json
{
	"1": {
		"date": "2023-11-23T13:40:00Z",
		"eventCode": {
			"enumType": "org.pih.warehouse.core.EventCode",
			"name": "SHIPPED"
		},
		"identifier": "842ZYG",
		"location": {
			"id": "0186",
			"name": "Imres",
			"description": null,
			"locationNumber": "",
			"locationGroup": null,
			"parentLocation": null,
			"locationType": {
				"id": "4",
				"name": "Supplier|fr:Fournisseurs",
				"description": "Supplier",
				"locationTypeCode": "SUPPLIER"
			},
			"sortOrder": 0,
			"hasBinLocationSupport": false,
			"hasPackingSupport": false,
			"hasPartialReceivingSupport": false,
			"hasCentralPurchasingEnabled": false,
			"organizationName": "Imres",
			"organizationCode": "US-0186",
			"backgroundColor": "f79ee1",
			"zoneName": null,
			"zoneId": null,
			"active": true,
			"organization": {
				"id": "0186",
				"name": "Imres",
				"description": null,
				"code": "US-0186",
				"dateCreated": "2008-07-06T00:00:00Z",
				"lastUpdated": "2008-07-06T00:00:00Z",
				"defaultLocation": null,
				"partyType": {
					"id": "ORG",
					"name": "Organization",
					"code": "ORG",
					"partyTypeCode": "ORGANIZATION"
				},
				"roles": [
					{
						"id": "0186S",
						"roleType": "ROLE_SUPPLIER",
						"startDate": "1980-01-01T00:00:00Z",
						"endDate": null
					}
				],
				"sequences": {
					"PURCHASE_ORDER_NUMBER": ""
				}
			},
			"manager": null,
			"address": {
				"address": "Larserpoortweg 26",
				"address2": "PO Box 214",
				"city": "8200 AE Lelystad",
				"country": "",
				"description": "PO Ship-to: Direct to Site\r\nPayment terms: Net 30",
				"postalCode": "10003",
				"stateOrProvince": ""
			},
			"supportedActivities": [
				"FULFILL_ORDER",
				"EXTERNAL",
				"SEND_STOCK"
			]
		},
		"referenceDocument": {
			"description": "RWPO000021 po16",
			"id": "cab2b1fa8bf9e8c3018bfc6917fd0124",
			"identifier": "842ZYG",
			"label": "842ZYG",
			"name": "Imres-Rwink.Pharm-23Nov2023-RWPO000021po16",
			"url": "/stockMovement/show/cab2b1fa8bf9e8c3018bfc6917fd0124"
		}
	}
}
```

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -697,5 +697,63 @@ class Shipment implements Comparable, Serializable {
     Boolean hasChildContainer() {
         shipmentItems?.any { it?.container }
     }
+
+    Set<Event> getCustomEvents() {
+        return events.findAll { it.eventType?.eventCode?.name()?.startsWith("CUSTOMS") }
```
awalkowiak, line 702: As discussed on slack, please add  `listSystemEventTypeCodes` and `listCustomEventTypeCodes` to the enum, and use the `listCustomEventTypeCodes` here

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -697,5 +697,63 @@ class Shipment implements Comparable, Serializable {
...
+            histories.add(new HistoryItem<Shipment>(
+                    identifier: shipmentNumber,
+                    date: dateShipped(),
+                    location: origin,
+                    eventCode: EventCode.SHIPPED,
+                    referenceDocument: shipmentReferenceDocument,
+            ))
+        }
+        // Then collect history of a received event if any
+        if (wasReceived()) {
```
awalkowiak, line 740: Should not be partially received before this one?

awalkowiak (changes requested):

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
jmiranda: Yes, but it shouldn't matter since we *should* sort before displaying. With that said, i think dding methods to HistoryItem to ensure the correct ordering would be a good idea as well.

src/main/groovy/org/pih/warehouse/core/Historizable.groovy
```diff
@@ -0,0 +1,7 @@
+package org.pih.warehouse.core
+
+interface Historizable {
+    List<HistoryItem> getHistory()
+
+    ReferenceDocument getReferenceDocument()
```
jmiranda, line 6: Interesting idea. However, I'm not sure I would put the reference document here since reference document is associated with the HistoryItem. In fact, each HistoryItem could have one or more ReferenceDocuments. For now we just might have the one, but the Received and Partially Received events might have a Receipt reference document, while Shipped might have a Shipment. I think the way it works out right now, Shipment is the reference document for each of these events but that's only the case because we don't have a Receipt details page that we want to show to the user.

jmiranda: There is a ReferenceDocument in HistoryItem, so I see what you're doing here. Every class should decide how it should be rendered as a HistoryItem, including how we might render it as a link. So this method is just telling the class to provide an implementation for both levels (for the row and for a link within the row).

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -697,5 +697,63 @@ class Shipment implements Comparable, Serializable {
...
+    List<HistoryItem> getHistory() {
+        List<HistoryItem> histories = []
+        ReferenceDocument shipmentReferenceDocument = getReferenceDocument()
+        // First collect history of CREATED event
+        histories.add(new HistoryItem<Shipment>(
+                identifier: shipmentNumber,
+                date: dateCreated,
+                location: origin,
+                eventCode: EventCode.CREATED,
+                referenceDocument: shipmentReferenceDocument
```
jmiranda, line 727: This is going to be weird looking, but if you remove Line 720 and change this line to 
```
referenceDocument: referenceDocument
```
it should work.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: It depends on if we want to keep the ordering in this method already. I assumed this could be sorted in every possible way on the view side, but I can change the ordering.

src/main/groovy/org/pih/warehouse/core/Historizable.groovy
kchelstowski: Yeah, for now I assumed that a "workflow" - receipt, shipment would have only one type of reference document, especially that we do not have a receipt view page.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: yeah it will, I think I kept that, because initially I had a few reference documents in the `getHistory()` method, before I decided to move each creator to a proper domain (Shipment, Receipt).

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: added

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: I switched the order

awalkowiak (approved):

PR #4941, opened 2024-11-14
OBPIH-6813 shipped in po column when edits to shipped quantities in receiving

drodzewicz:

**Link to GitHub issue or Jira ticket:** [OBPIH-6813](https://pihemr.atlassian.net/browse/OBPIH-6813)

**Description:**
Main goal of this ticket was to display a propper Shipped (In PO UoM) quantity when splitting items.

Before if we had an item with shipped quantity `12` that was labeled with UoM `3 PK/4` and we split the item in two items by 8 and 4 we still got the same label but instead we want it to be `2 PK/4`  and  `1 PK/4` 

---

image

[OBPIH-6813]: https://pihemr.atlassian.net/browse/OBPIH-6813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -261,15 +261,38 @@ const TABLE_FIELDS = {
...
-        getDynamicAttr: (props) => ({
-          hide: !props.values?.isShipmentFromPurchaseOrder,
-        }),
+        getDynamicAttr: ({ values, parentIndex, rowIndex }) => {
+          const shipmentItem = _.get(
+            values,
+            `containers[${parentIndex}].shipmentItems[${rowIndex}]`,
+            {},
+          );
+          const packsRequested = _.round(shipmentItem?.packsRequested, 1);
```
awalkowiak, line 277: Is this rounding correct? Have you confirmed that behavior? This would give the same result for PK/6 and shipped 1 in each, and for PK/5 and shipped 1 in each, which is 0.2. Feels like rounding to two decimal places would be more accurate.

awalkowiak: Yeah on second thought I thing this should be 2 decimal places, feels like quarters (`.25`) might be a potential common case. In the current version this would be rounded up to `0.3`.

kchelstowski (approved):

src/js/components/receiving/PartialReceivingPage.jsx
```diff
@@ -261,15 +261,38 @@ const TABLE_FIELDS = {
...
         },
-        getDynamicAttr: (props) => ({
-          hide: !props.values?.isShipmentFromPurchaseOrder,
-        }),
+        getDynamicAttr: ({ values, parentIndex, rowIndex }) => {
+          const shipmentItem = _.get(
+            values,
+            `containers[${parentIndex}].shipmentItems[${rowIndex}]`,
+            {},
+          );
```
alannadolny, line 276: can you use nullsafe here?

drodzewicz: I had some issues when using 
```
 const shipmentItem = values?.containers[parentIndex]?.shipmentItems[rowIndex] ?? {};
```

I guess the values are very deeply nested and nulish operator does not handle it very well

PR #4945, opened 2024-11-15
OBPIH-6733 Re-add release information to Sentry

EWaterman:


EWaterman:
Had a minor breakthrough last night while randomly digging into Sentry.init flows. This is a marginally better approach to how we were setting the release before in that it also works when testing locally. I also added some more documentation around how Sentry.init works so that we're not confused in the future.

I'm going to merge in as is because I tested on obdev5 already

PR #4947, opened 2024-11-19
OBPIH-6472 Reintroduce shipment events tab in stock movement UI

alannadolny:

image

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
@@ -25,6 +25,7 @@ import org.pih.warehouse.core.Document
 import org.pih.warehouse.core.DocumentCommand
 import org.pih.warehouse.core.DocumentService
 import org.pih.warehouse.core.DocumentType
+import org.pih.warehouse.core.HistoryItem
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.Comment
 import org.pih.warehouse.core.User
@@ -453,6 +454,15 @@ class StockMovementController {
         def stockMovement = getStockMovement(params.id)
         def receiptItems = stockMovementService.getStockMovementReceiptItems(stockMovement)
         render(template: "receipts", model: [receiptItems: receiptItems])
+    }
+
+    def events() {
+        StockMovement stockMovement = getStockMovement(params.id)
+        List<HistoryItem> historyItems = stockMovement.shipment.getHistory()
+        render(
+                template: "events",
+                model: [historyItems: historyItems, shipmentId: stockMovement?.shipment?.id]
+        )
     }
 
     // Used by SM show page 'tabs' actions - packing list, documents and receipts
--- a/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
@@ -458,7 +458,7 @@ class StockMovementController {
 
     def events() {
         StockMovement stockMovement = getStockMovement(params.id)
-        List<HistoryItem> historyItems = stockMovement.shipment.getHistory()
+        List<HistoryItem> historyItems = stockMovement?.shipment?.getHistory() ?: []
         render(
                 template: "events",
                 model: [historyItems: historyItems, shipmentId: stockMovement?.shipment?.id]
--- a/grails-app/domain/org/pih/warehouse/core/Location.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Location.groovy
@@ -10,6 +10,7 @@
 package org.pih.warehouse.core
 
 import grails.util.Holders
+import org.hibernate.sql.JoinType
 import org.pih.warehouse.inventory.Inventory
 import org.pih.warehouse.inventory.InventorySnapshotEvent
 import org.pih.warehouse.inventory.RefreshProductAvailabilityEvent
@@ -362,6 +363,18 @@ class Location implements Comparable<Location>, java.io.Serializable {
 
     Boolean isApprovalRequired() {
         return supports(ActivityCode.APPROVE_REQUEST)
+    }
+
+    static List<Location> listNonInternalLocations() {
+        return createCriteria().list {
+            createAlias("locationType", "locationType", JoinType.LEFT_OUTER_JOIN)
+            and {
+                eq('active', true)
+                not {
+                    'in'('locationType.locationTypeCode', [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION])
+                }
+            }
+        }
     }
 
     Map toBaseJson() {
--- a/grails-app/domain/org/pih/warehouse/core/Location.groovy
+++ b/grails-app/domain/org/pih/warehouse/core/Location.groovy
@@ -371,7 +371,7 @@ class Location implements Comparable<Location>, java.io.Serializable {
             and {
                 eq('active', true)
                 not {
-                    'in'('locationType.locationTypeCode', [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION])
+                    'in'('locationType.locationTypeCode', LocationTypeCode.listInternalTypeCodes())
                 }
             }
         }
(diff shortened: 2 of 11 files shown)


grails-app/views/stockMovement/_events.gsp
```diff
@@ -0,0 +1,137 @@
...
+                <td valign="top" class="name">
+                    <label>
+                        <g:message code="default.location.label" default="Location" />
+                    </label>
+                </td>
+                <td valign="top" class="value ${hasErrors(bean: eventInstance, field: 'location', 'errors')}">
+                    <g:select id="eventLocation.id"
+                              name="eventLocation.id"
+                              noSelection="['': warehouse.message(code: 'default.selectOne.label')]"
+                              from='${Location.list()}'
```
awalkowiak, line 96: Ooof, that `Location.list()` might be an overkill. On obdev3 this gives us `51646` options with all locations in the system (even inactive). What should be a potential location to choose here? Depots, Wards, Internal Locations? Can we somehow narrow this or do an async select?

grails-app/views/stockMovement/_events.gsp
```diff
@@ -0,0 +1,137 @@
...
+            </tr>
+            %{-- Event date picker --}%
+            <tr class="prop">
+                <td valign="top" class="name">
+                    <label>
+                        <g:message code="event.eventDate.label" default="Event date"/>
+                    </label>
+                </td>
+                <td valign="top" class="value ${hasErrors(bean: eventInstance, field: 'eventDate', 'errors')}">
+                    <g:datePicker name="eventDate" value="${eventInstance?.eventDate}" precision="minute"/>
```
awalkowiak, line 111: Should this be required? Or is later set with the current datetime if empty?

grails-app/views/stockMovement/_events.gsp
awalkowiak: Additional question - should this be required? On the old event tab, it looks like it is preselected with origin and cannot be cleared.

alannadolny: > AND the Location field selects the value “null” by default

grails-app/views/stockMovement/_events.gsp
alannadolny: This picker cannot be empty on form submission. It's preset with the current date.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -455,6 +456,15 @@ class StockMovementController {
         render(template: "receipts", model: [receiptItems: receiptItems])
     }
 
+    def events() {
+        StockMovement stockMovement = getStockMovement(params.id)
```
kchelstowski, line 460: I would use the `stockMovementService`'s method to be consistent with the show view page.

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -455,6 +456,15 @@ class StockMovementController {
...
+        List<HistoryItem> historyItems = stockMovement.shipment.getHistory()
```
kchelstowski, line 461: don't we need nullsafes here?

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -196,12 +196,12 @@ class Shipment implements Comparable, Serializable, Historizable {
...
         })
 
-        // a shipment can't have two events with the same event code (this should be the case for the current event codes: CREATED, SHIPPED, RECEIVED)
-        // we may want to change this in the future?
+        // a shipment can't have two system events with the same event code
         events(validator: { events ->
-            events?.collect({
-                it.eventType?.eventCode
-            })?.unique({ a, b -> a <=> b })?.size() == events?.size()
+            List<Event> systemEvents = events.toList().findAll {
```
kchelstowski, line 201: a titbit, that you probably didn't have to parse it to list, since you have the `.findAll` method available also for the Set.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -196,12 +196,12 @@ class Shipment implements Comparable, Serializable, Historizable {
             })?.unique({ a, b -> a <=> b })?.size() == referenceNumbers?.size()
         })
 
-        // a shipment can't have two events with the same event code (this should be the case for the current event codes: CREATED, SHIPPED, RECEIVED)
-        // we may want to change this in the future?
+        // a shipment can't have two system events with the same event code
```
kchelstowski, line 199: actually what about partially_received events?

src/main/groovy/org/pih/warehouse/core/HistoryItem.groovy
```diff
@@ -10,4 +10,8 @@ class HistoryItem<T> {
     ReferenceDocument referenceDocument
 
     EventCode eventCode
+
+    Comment comment
+
+    User createdBy
```
kchelstowski, line 16: (my bad) I can see redundant `identifier` in this domain (it was moved to the `ReferenceDocument`).
Could you please remove the `identifier` from the `HistoryItem` and the assignments of it (Shipment domain, Receipt domain, Event domain).

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
kchelstowski: For now I guess it doesn't matter, because we build history items for partially_received events looking at the `receipts`, not the `events

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: actually see how the logic is handled for the `def show()` , I would use the same snippet here:
```
// Pull Outbound Stock movement (Requisition based) or Outbound or Inbound Return (Order based)
def stockMovement = outboundStockMovementService.getStockMovement(params.id)
// For inbound stockMovement only
if (!stockMovement) {
    stockMovement =  stockMovementService.getStockMovement(params.id)
}
```

alannadolny: I would leave the code as is here, the `getStockMovement` method was introduced not that long ago, and I can see there is an additional logic inside
```
    // Used by SM show page 'tabs' actions - packing list, documents and receipts
    def getStockMovement(String stockMovementId) {
        def stockMovement
        // Pull stock movement in "old fashion" way to bump up performance a bit (instead of getting OutboundStockMovement) for Non-Returns
        def order = Order.get(stockMovementId)
        if (order) {
            stockMovement = outboundStockMovementService.getStockMovement(stockMovementId)
        } else {
            stockMovement = stockMovementService.getStockMovement(stockMovementId)
        }

        return stockMovement
    }
```

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
alannadolny: When I am removing the `.toList()` I am getting:
```
Cannot cast object '[Shipped|fr:Exp Rwinkwavu Pharmacy Warehouse on 2024-08-23 16:02:22.0]' with class 'java.util.TreeSet' to class 'java.util.List' due to: groovy.lang.GroovyRuntimeException: Could not find matching constructor for: java.util.List(org.pih.warehouse.core.Event)
```

kchelstowski: Isn’t it due to the fact that you declared the returned type as List<Event>?

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
kchelstowski: ok, if this method is used for other tabs as well, I'm fine with that

grails-app/domain/org/pih/warehouse/core/Location.groovy
```diff
@@ -364,6 +365,18 @@ class Location implements Comparable<Location>, java.io.Serializable {
         return supports(ActivityCode.APPROVE_REQUEST)
     }
 
+    static List<Location> listNonInternalLocations() {
+        return createCriteria().list {
+            createAlias("locationType", "locationType", JoinType.LEFT_OUTER_JOIN)
+            and {
+                eq('active', true)
+                not {
+                    'in'('locationType.locationTypeCode', [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION])
```
kchelstowski, line 374: you could use `LocationTypeCode.listInternalTypeCodes()` here.

kchelstowski (approved):

awalkowiak (approved):

PR #4948, opened 2024-11-19
OBPIH-6821 Try next sequence number if generating an order number fails

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6821

**Description:** See ticket for details, but this PR fixes an issue where an organization's sequence number falls out of sync with the order numbers that we have generated for POs. Essentially the PR adds a retry mechanism where if the id we generate is not unique, check the next 10 sequence numbers if they're available.


EWaterman:
@awalkowiak feel free to have someone take over this PR if we want to close it before the sprint ends since I'm out for the rest of the week.

PR #4955, opened 2024-11-26
OBPIH-6839 Improve performance of order's derived status fetch and update

awalkowiak:

https://pihemr.atlassian.net/browse/OBPIH-6839

**Description:**
I added OrderSummaryService which basically copies the queries from the order summary helper views (with only one difference, each of these now pulls data for a specific order ID). This OrderSummaryService is a temporary (or to be expanded) solution for performance issues with order summary helper views during fetching and reloading data. In the next iteration, we should migrate from using views to using queries and/or we should move these to stored procedures (imho we should get rid of the views and leave only the materialized views plus queries / stored procedures)

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
@@ -14,15 +14,18 @@ import org.pih.warehouse.importer.CSVUtils
 import org.pih.warehouse.order.Order
 import org.pih.warehouse.order.OrderItem
 import org.pih.warehouse.order.OrderItemStatusCode
+import org.pih.warehouse.order.OrderService
 import org.pih.warehouse.order.OrderStatus
 import org.pih.warehouse.order.OrderSummary
+import org.pih.warehouse.order.OrderSummaryService
 import org.pih.warehouse.order.OrderSummaryStatus
 
 import java.math.RoundingMode
 
 class PurchaseOrderApiController {
 
-    def orderService
+    OrderService orderService
+    OrderSummaryService orderSummaryService
 
     def list() {
         List<OrderSummary> purchaseOrders = orderService.getPurchaseOrders(params)
@@ -234,7 +237,6 @@ class PurchaseOrderApiController {
     }
 
     def orderItemsDerivedStatus = {
-        Order order = Order.get(params.id)
-        render order?.getOrderItemsDerivedStatus() as JSON
+        render orderSummaryService.getOrderItemsDerivedStatus(params.id) as JSON
     }
 }
--- a/grails-app/jobs/org/pih/warehouse/jobs/RefreshOrderSummaryJob.groovy
+++ b/grails-app/jobs/org/pih/warehouse/jobs/RefreshOrderSummaryJob.groovy
@@ -1,12 +1,12 @@
 package org.pih.warehouse.jobs
 
 import grails.util.Holders
-import org.pih.warehouse.order.OrderService
+import org.pih.warehouse.order.OrderSummaryService
 import org.quartz.JobExecutionContext
 
 class RefreshOrderSummaryJob {
 
-    OrderService orderService
+    OrderSummaryService orderSummaryService
 
     static concurrent = false
 
@@ -21,7 +21,7 @@ class RefreshOrderSummaryJob {
         if (JobUtils.shouldExecute(RefreshOrderSummaryJob)) {
             def startTime = System.currentTimeMillis()
             log.info("Refreshing order summary: " + context.mergedJobDataMap)
-            orderService.refreshOrderSummary()
+            orderSummaryService.refreshOrderSummary()
             log.info "Finished refreshing order summary in " + (System.currentTimeMillis() - startTime) + " ms"
         }
     }
--- a/grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
+++ b/grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
@@ -1,0 +1,441 @@
+/**
+ * Copyright (c) 2012 Partners In Health.  All rights reserved.
+ * The use and distribution terms for this software are covered by the
+ * Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0.php)
+ * which can be found in the file epl-v10.html at the root of this distribution.
+ * By using this software in any fashion, you are agreeing to be bound by
+ * the terms of this license.
+ * You must not remove this notice, or any other, from this software.
+ **/
+package org.pih.warehouse.order
+
+import org.pih.warehouse.data.DataService
+import org.pih.warehouse.jobs.RefreshOrderSummaryJob
+
+
+/**
+ * TODO: This OrderSummaryService is temporary (or to be expanded) solution for performance issues with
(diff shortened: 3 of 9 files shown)


grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
```diff
@@ -0,0 +1,441 @@
...
+                GROUP BY id, order_status) AS order_summary
+            WHERE id = '${orderId}'
+        """
+    }
+
+    /**
+     * Refreshing entire Order Summary materialized view, should be only triggered from time to time
+     * (same statements as in the order-summary-materialized-view.sql)
+     * */
+    def refreshOrderSummary() {
```
alannadolny, line 384:
```suggestion
    void refreshOrderSummary() {
```

alannadolny: the same change in the functions below

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
```diff
@@ -0,0 +1,441 @@
...
+    def refreshOrderSummary(List<String> orderIds, Boolean isDelete) {
+        List statements = []
+        orderIds?.each { String orderId ->
+            if (isDelete) {
+                statements << "DELETE FROM order_summary_mv WHERE id = '${orderId}';"
+            } else {
+                String orderSummarySelect = getOrderSummarySelect(orderId)
+                statements << "REPLACE INTO order_summary_mv (${orderSummarySelect});"
+            }
+        }
```
alannadolny, line 414: you can write this function without pushing strings to the `statements`, you can use `inject`.

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
```diff
@@ -0,0 +1,441 @@
...
+
+        if (statements && checkIfOrderSummaryExists()) {
```
alannadolny, line 416: Wouldn't it be better to check the `statements.length`? just for clarity

alannadolny (commented):
GJ with that improvement :clap: If this approach is suggested over creating views we should add it to our conventions

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
awalkowiak: I just moved this one

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
awalkowiak: I just moved these few refreshes, from order service, but this makes sens

awalkowiak:
@alannadolny 
>  If this approach is suggested over creating views we should add it to our conventions

I am not sure yet if this is suggested overall, just looks like it is better than views in this particular case

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
```diff
@@ -0,0 +1,441 @@
...
+                        WHEN shipment.current_status IN ('SHIPPED', 'PARTIALLY_RECEIVED', 'RECEIVED') THEN SUM(shipment_item.quantity / order_item.quantity_per_uom)
+                        ELSE 0
+                    END AS quantity_shipped
+                FROM `order`
+                    LEFT OUTER JOIN order_item ON order.id = order_item.order_id
+                    LEFT OUTER JOIN order_shipment ON order_item.id = order_shipment.order_item_id
+                    LEFT OUTER JOIN shipment_item ON shipment_item.id = order_shipment.shipment_item_id
+                    LEFT OUTER JOIN shipment ON shipment.id = shipment_item.shipment_id
+                WHERE `order`.order_type_id = 'PURCHASE_ORDER'
+                    AND `order`.id = '${orderId}'
```
EWaterman, line 53: so are these queries faster because you're pre-filtering only for the order id that you need?

grails-app/jobs/org/pih/warehouse/jobs/RefreshOrderSummaryJob.groovy
```diff
@@ -21,7 +21,7 @@ class RefreshOrderSummaryJob {
         if (JobUtils.shouldExecute(RefreshOrderSummaryJob)) {
             def startTime = System.currentTimeMillis()
             log.info("Refreshing order summary: " + context.mergedJobDataMap)
-            orderService.refreshOrderSummary()
+            orderSummaryService.refreshOrderSummary()
```
EWaterman, line 24: do you know why this job is needed? It seems the refreshOrderSummary method just takes a copy of the view, deletes it, recreates it, and re-inserts the copy. I'm not sure what this is even achieving?

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
```diff
@@ -234,7 +237,6 @@ class PurchaseOrderApiController {
     }
 
     def orderItemsDerivedStatus = {
-        Order order = Order.get(params.id)
-        render order?.getOrderItemsDerivedStatus() as JSON
+        render orderSummaryService.getOrderItemsDerivedStatus(params.id) as JSON
```
EWaterman, line 240: is this where the time save comes in? I'm curious why it's faster since Order.getOrderItemsDerivedStatus() calls OrderIttemSummary.findAllByOrder(this) which I would think just does a select on order id on the pre-built view.

Is the view not fully materialized at this point or something? Or is the select not indexed properly (is that even a thing on materialized views??)

kchelstowski: @EWaterman if I'm not mistaken, the views are recreated whenever you access them, hence this might be why it is slower, but this needs a confirmation, as I didn't check it, I just remember a theoretical statement about views.

kchelstowski (approved):

grails-app/services/org/pih/warehouse/order/OrderSummaryService.groovy
awalkowiak: Yes, but on top of that, there is no need to build entire view

grails-app/jobs/org/pih/warehouse/jobs/RefreshOrderSummaryJob.groovy
awalkowiak: It is needed to be able to trigger rebuild of entire order summary **materialized** view (`order_summary_mv` table)

grails-app/controllers/org/pih/warehouse/api/PurchaseOrderApiController.groovy
EWaterman: thanks for the explanation. I definitely agree that these flows would benefit (at least visually) from being broken down a bit. Stacking views on top of views makes me uncomfortable

grails-app/jobs/org/pih/warehouse/jobs/RefreshOrderSummaryJob.groovy
EWaterman: I suppose I just don't understand how the materialized views work and why they need to be dropped entirely and recreated in this job. Does it get out of sync with order_summary somehow?

Anyways, this isn't relevant to the PR itself so we can move on.

PR #4959, opened 2024-11-27
OBPIH-6858 enable color logs for dev and testing only

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6858

**Description:** Disable ANSI Logback colors in prod so that we don't get gross log output like this when the terminal can't parse them properly:

image


grails-app/conf/application-development.yml
```diff
@@ -0,0 +1,6 @@
+
+# Only add richly formatted logs when developing locally. Some terminals can't handle ANSI colors, so
+# we don't want to break any log processors in production. https://logback.qos.ch/manual/layouts.html
+logging:
+    pattern:
+        console: "%cyan(%date{ISO8601}) %highlight(%-5level) %magenta([%thread]) %yellow(%logger{40}): %message%n%xException"
```
EWaterman, line 6: I originally tried doing this via Grails' environment properties:
```
environments:
    development:
        logging:
            pattern:
                console: <pattern with colors>
...
logging:
    pattern:
        console: <pattern without colors>
```
but unfortunately it didn't get picked up by Logback. I assume it's because the environments is a Grails thing.

grails-app/conf/logback.xml
```diff
@@ -1,5 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <configuration>
 
+    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
+    <include resource="org/springframework/boot/logging/logback/console-appender.xml"/>
```
EWaterman, line 5: This is what gets imported here: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/console-appender.xml

It saves us from having to define our own console appender since we're just using the default config anyways. Then we can override the format with the `logging.pattern.console:` property which maps to CONSOLE_LOG_PATTERN which is defined in https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/defaults.xml

src/integration-test/resources/logback-test.xml
EWaterman, line 1: Moved this file to test/resources so that it applies to unit tests as well

awalkowiak (approved):

PR #4961, opened 2024-11-28
OBPIH-6833 Original line not editable when delete split lines in stock transfer

alannadolny:


src/js/components/stock-transfer/StockTransferSecondPage.jsx
```diff
@@ -363,6 +369,18 @@ class StockTransferSecondPage extends Component {
...
+    const itemsGroupedByReferenceId = _.groupBy(filteredItems, 'referenceId');
+
+    // If an item doesn't have any subitems it should be enabled (status PENDING)
+    return filteredItems.map((item) => {
+      if (!itemsGroupedByReferenceId[item?.id]) {
+        return { ...item, status: StockTransferStatus.PENDING };
+      }
+
+      return item;
+    });
```
EWaterman, line 383: So right now you:
- loop through to filter the items
- create a map
- loop through the filtered items to set their status

I feel like this could be achieved in fewer iterations, possibly by utilizing the itemsGroupedByReferenceId map more, but  I'm not really worried about it. It's probably fast enough

EWaterman (approved):

awalkowiak (approved):

PR #4962, opened 2024-11-28
OBPIH-6835 Additional line counted in line items count on stock transfers list

alannadolny:


grails-app/domain/org/pih/warehouse/order/Order.groovy
```diff
@@ -602,7 +602,7 @@ class Order implements Serializable {
                     dateCreated     : dateCreated,
                     origin          : origin?.name,
                     destination     : destination?.name,
-                    orderItemsCount : orderItems?.size(),
+                    orderItemsCount : orderItems?.count { it.orderItemStatusCode != OrderItemStatusCode.CANCELED },
```
EWaterman, line 605: so when we split a line, the original line gets status CANCELED? Interesting I didn't know this.

awalkowiak (approved):

PR #4964, opened 2024-11-28
OBPIH-6860 set log level to INFO in prod

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6860

**Description:** Demoted some noisy logs to DEBUG and added the ability to set log level per deployment method (development is staying at DEBUG, production is changed to INFO).

Some logs we won't see anymore on prod:
```
2024-11-15 21:55:21,368 DEBUG [-exec-4514] org.pih.warehouse.SentryInterceptor     : updated Sentry context for /openboxes/api/purchaseOrders in 0 ms
2024-11-15 21:55:21,368  INFO [-exec-4514] org.pih.warehouse.RoleInterceptor       : No rule for purchaseOrderApi:list -> allow anonymous
2024-11-15 21:55:21,369  INFO [-exec-4514] org.pih.warehouse.core.UserService      : Is role ROLE_ADMIN in [ROLE_MANAGER, ROLE_BROWSER, ROLE_ASSISTANT, ROLE_AUTHENTICATED, ROLE_SUPERUSER, ROLE_ADMIN] = true
```

---

image


grails-app/controllers/org/pih/warehouse/LoggingInterceptor.groovy
```diff
@@ -31,19 +38,21 @@ class LoggingInterceptor {
...
             MDC.put('serverUrl', Holders.grailsApplication.config?.grails?.serverURL ?: "No server URL")
         } catch (Exception e) {
-            LoggingInterceptor.log.warn("Error occurred while adding attributes to Mapped Diagnostic Context: ${e.message}", e)
-
+            log.warn("Error occurred while adding attributes to Mapped Diagnostic Context: ${e.message}", e)
         }
+
+        // Now that we've populated the MDC, log the request itself, including query parameters.
+        String queryString = StringUtils.isBlank(request?.queryString) ? "" : "?${request?.queryString}"
+        log.info("${request.requestURI}${queryString} [user:${session?.user?.username}, location:${session?.warehouse?.name}]")
```
EWaterman, line 46: I moved what was in `AccessLogInterceptor` to here since it felt like it should be encompassed by the logging interceptor. I also added the query parameters to the end of the request.

awalkowiak (commented):
@EWaterman overall this looks good to me, I think I never looked at the `RoleInterceptor` logs and usually I was irritated by them. I additionally looked if there are any `log.debugs` that we might miss after the change and I don't see anything required there (plus we could also change the level through `openboxes-config.groovy` config when we need to debug some user roles). I am good with the change, but would like to get an approval from @jmiranda as well.

awalkowiak:
@EWaterman 
>No need to review this before the release. I was just needing a break from Cycle Count so I did this for fun.

And I read this after review :sweat_smile: but yeah my thought was that this does not have to go to the release yet.

EWaterman:
marking this change as do not merge as we're going to wait until 0.9.3 is cut

grails-app/controllers/org/pih/warehouse/LoggingInterceptor.groovy
jmiranda, line 46: I think this is the log message that @awalkowiak doesn't like. I find it helpful to demarcate logging for requests along with the message that reports response time. 

Aside: Now that I'm looking at the logs I'm noticing that the response time is no longer reported. @awalkowiak Do you know when that got removed? 

Anyway, I can be overruled if others would prefer to remove it or make it a DEBUG message.

jmiranda (approved):
Look good.

grails-app/controllers/org/pih/warehouse/LoggingInterceptor.groovy
awalkowiak: @jmiranda this one I like. I often look at this one. The one I dislike that is causing a lot of garbage is the one in the `UserService` -> `log.info "Is role ${role.roleType} in ${acceptedRoleTypes} = ${acceptedRoleType}"` (and I am happy that it was demoted to debug level)

PR #4965, opened 2024-11-29
OBPIH-6862 Buttons and columns in outbound return are not translated to Mexico language

alannadolny:


src/js/components/returns/outbound/SendOutboundReturn.jsx
```diff
@@ -481,7 +481,7 @@ class SendMovementPage extends Component {
                   disabled={values && values.status === 'COMPLETED'}
                   onClick={() => this.previousPage(values, invalid)}
                 >
-                  <Translate id="react.default.button.previous.label" defaultMessage="Previous" />
```
kchelstowski, line 484: this one should be working

alannadolny: yeah, but it's not consistent with the rest pages from this workflow

EWaterman (approved):

awalkowiak (approved):

PR #4967, opened 2024-12-02
OBPIH-6864 Fix assign identifier job

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6864

**Description:** The AssignIdentifierJob wasn't running properly because it was unable to auto-wire in the BlankIdentifierResolver beans. This should work according to Springboot, but it's not worth the hassle trying to debug why it's not working here so I just switched it to wire in the BlankIdentifierResolver classes explicitly.

---

Screenshot from 2024-12-02 10-26-31

### DIFF
--- a/grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
+++ b/grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
@@ -1,11 +1,22 @@
 package org.pih.warehouse.jobs
 
-import org.pih.warehouse.core.identification.BlankIdentifierResolver
+import org.pih.warehouse.inventory.TransactionIdentifierService
+import org.pih.warehouse.order.OrderIdentifierService
 import org.pih.warehouse.product.Product
+import org.pih.warehouse.product.ProductIdentifierService
+import org.pih.warehouse.receiving.ReceiptIdentifierService
+import org.pih.warehouse.requisition.RequisitionIdentifierService
+import org.pih.warehouse.shipping.ShipmentIdentifierService
 
 class AssignIdentifierJob {
 
-    List<BlankIdentifierResolver> blankIdentifierResolvers
+    // Every identifier service that implements BlankIdentifierResolver
+    ProductIdentifierService productIdentifierService
+    ShipmentIdentifierService shipmentIdentifierService
+    ReceiptIdentifierService receiptIdentifierService
+    OrderIdentifierService orderIdentifierService
+    RequisitionIdentifierService requisitionIdentifierService
+    TransactionIdentifierService transactionIdentifierService
 
     def sessionRequired = false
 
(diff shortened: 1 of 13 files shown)


EWaterman:
Currently this is throwing the following errors:
```
Exception occurred in job: Grails Job
org.quartz.JobExecutionException: javax.persistence.TransactionRequiredException: no transaction is in progress
        at grails.plugins.quartz.GrailsJobFactory$GrailsJob.execute(GrailsJobFactory.java:111)
        at org.quartz.Job$execute.call(Unknown Source)
        at grails.plugins.quartz.QuartzDisplayJob.execute(QuartzDisplayJob.groovy:34)
        at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
        at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: javax.persistence.TransactionRequiredException: no transaction is in progress
        at org.hibernate.internal.SessionImpl.checkTransactionNeeded(SessionImpl.java:3505)
        ....
```
Still need to figure out why

grails-app/jobs/org/pih/warehouse/jobs/AssignIdentifierJob.groovy
```diff
@@ -1,11 +1,24 @@
...
+import grails.gorm.transactions.Transactional
 
+import org.pih.warehouse.inventory.TransactionIdentifierService
+import org.pih.warehouse.order.OrderIdentifierService
+import org.pih.warehouse.product.ProductIdentifierService
+import org.pih.warehouse.receiving.ReceiptIdentifierService
+import org.pih.warehouse.requisition.RequisitionIdentifierService
+import org.pih.warehouse.shipping.ShipmentIdentifierService
+
+@Transactional
```
EWaterman, line 12: I had to add this to make the job work. Without it it was hitting "no transaction is in progress" errors. Maybe a better solution though is to wrap each individual generateForAllUnassignedIdentifiers() call in their own transaction based on the underlying table that they're working on.

@awalkowiak I see you had this PR relating to transactions in jobs so maybe you'll know better than me https://github.com/openboxes/openboxes/pull/4160

EWaterman: Possibly relevant since we are doing `sessionRequired=false` in all our jobs: https://grails.github.io/grails-quartz/guide/configuration.html

> Hibernate Sessions and Jobs
> Jobs are configured by default to have Hibernate Session bounded to thread each time job is executed. This is required if you are using Hibernate code which requires open session (such as lazy loading of collections) or working with domain objects with unique persistent constraint (it uses Hibernate Session behind the scene). If you want to override this behavior (rarely useful) you can use 'sessionRequired' property:
> 
> def sessionRequired = false

awalkowiak: @EWaterman here you might find a bit more details about the issue with `no transaction is in progress` https://pihemr.atlassian.net/browse/OBGM-584 (this is the OG ticket under which I introduced the wrapping service calls with the `withNewSession`), this was basically the same issue you run into here: https://github.com/openboxes/openboxes/pull/4967#issuecomment-2512574451, back then I remember `@Transactional` did not work for me. Does the `@Transactional` work now? Are the new identifiers persisted to the db correctly now?

src/main/groovy/org/pih/warehouse/core/identification/BlankIdentifierResolver.groovy
```diff
@@ -55,9 +55,7 @@ trait BlankIdentifierResolver<T extends GormEntity> {
                 log.info("Assigning identifier ${identifier} to ${entityType} with id: ${entity.id}")
                 setIdentifierOnEntity(identifier, entity)
 
-                // TODO: investigate why this merge with flush and no validation is required. Can we simplify this to
-                //       a regular save operation?
-                if (!entity.merge(flush: true, validate: false)) {
+                if (!entity.save()) {
```
EWaterman, line 58: from what I've gathered, the merge was required because the assign identifier job was wrapping all the generate id calls with the same `Product.withNewSession`, even though only one of them was actually saving to product. This meant that the entities might not have been in the session, and so we were forced to work with detached objects via merge. Now that we're leaving the transactions to the individual services themselves, which do the work to fetch the entities (and thus load them into the session) this is no longer required.

awalkowiak (approved):

PR #4968, opened 2024-12-02
OBPIH-6585 Make code a required field when creating/editing location types

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6585

**Description:** `LocationType.locationTypeCode` is currently a nullable field, but `LocationDimension.locationTypeCode` is a non-nullable field. If we ever have a location type with a null location type code, it causes column constraint failures when generating location dimensions for all locations and causes the whole action to fail.

350177317-0f0866ec-6f5b-4df8-b4a9-5864b35af124

This change is to make `LocationType.locationTypeCode` non-nullable.

Originally implemented in: https://github.com/openboxes/openboxes/pull/4755


EWaterman:
I verified that the following scenario works:

- created a location type with a null code
- apply my changes and run the migration
- location type changes to DEFAULT and can still load correctly
- edit the location type and changed the code to DEPOT
- change applied successfully

grails-app/views/locationType/create.gsp
```diff
@@ -41,7 +41,6 @@
                             </td>
                             <td valign="top" class="value ${hasErrors(bean: locationTypeInstance, field: 'description', 'errors')}">
                                 <g:select name="locationTypeCode" class="chzn-select-deselect" from="${org.pih.warehouse.core.LocationTypeCode.list()}"
-                                          noSelection="['':'']"
```
EWaterman, line 44: As a reminder, removing this line will default the selection to the first LocationTypeCode (which is DEPOT)

kchelstowski (commented):
I feel ok with the code, but I have a doubt if it's a good time to proceed with such change now. I'm worried about some existing data etc - I know you've added the default value, but I don't know if there are any edge cases where it would break something.

awalkowiak (commented):
I am not sure about these changes, on one hand these makes sense, but I am not sure about the DEFAULT location type code (Imho it could be just empty string). I'd like to get an opinion from @jmiranda. Plus this feels like something that does not to be released in 0.9.3? The immediate fix could be one of these:
1. Manual data fix in db (not sure if this makes sense, but we could take a look what was the fix before in the https://pihemr.atlassian.net/browse/OBPIH-5256
2. Make it nullable false, but blank true and allow empty strings there
3. On the `buildLocationDimension` insert, to do a fallback to empty string of something to not break due to null
4. Make nullable location type code on the location dimension for now.

EWaterman:
marking as do not merge for now since we decided to resolve this one after we cut the 0.9.3 release

grails-app/views/locationType/edit.gsp
```diff
@@ -46,7 +46,6 @@
                         class="value ${hasErrors(bean: locationTypeInstance, field: 'description', 'errors')}">
                         <g:select name="locationTypeCode" class="chzn-select-deselect"
                                   from="${org.pih.warehouse.core.LocationTypeCode.list()}"
-                                  noSelection="['': '']"
```
alannadolny, line 49: Removing this property means that the first one from the list is picked automatically or how is it working right now?

src/test/groovy/unit/org/pih/warehouse/core/LocationControllerSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+    PagedResultList buildStubbedPagedResultList(ArrayList expectedList) {
+        return new PagedResultList(null).with {
+            resultList = expectedList
+            totalCount = expectedList.size()
+            it
+        }
+    }
+
+    LocationGroup createLocationGroup(String name) {
+        return new LocationGroup(name: name).save(validate:false)
```
alannadolny, line 134:
```suggestion
        return new LocationGroup(name: name).save(validate: false)
```

src/test/groovy/unit/org/pih/warehouse/core/LocationControllerSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    LocationType createLocationType(String name) {
+        return new LocationType(name: name).save(validate:false)
```
alannadolny, line 138:
```suggestion
        return new LocationType(name: name).save(validate: false)
```

src/test/groovy/unit/org/pih/warehouse/core/LocationControllerSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    Organization createOrganization(String code, String name) {
+        return new Organization(code: code, name: name).save(validate:false)
```
alannadolny, line 142:
```suggestion
        return new Organization(code: code, name: name).save(validate: false)
```

src/test/groovy/unit/org/pih/warehouse/core/LocationControllerSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    Location createLocation(String name, LocationType locationType, LocationGroup locationGroup, Organization organization) {
+        return new Location(
+                name: name,
+                locationType: locationType,
+                locationGroup: locationGroup,
+                organization: organization)
+                .save(validate:false)
```
alannadolny, line 151:
```suggestion
                .save(validate: false)
```

grails-app/views/locationType/edit.gsp
EWaterman: yeah that's right. The first element of LocationTypeCode.list(), which is DEPOT

jmiranda: Thinking about this more, I would actually prefer the default to be empty and for the form submission to return and display a validation error for blank/null. 

Can you post a screenshot of the ugly page (I assume this is the exception stacktrace) so I can figure out the best way to handle it?

EWaterman: image

EWaterman: I agree it's probably better to default to empty and just handle the error better.

This is old gsp pages though so I don't know the best way to do that. Do we even have nice error handling at all for these pages? If it's not a simple fix, maybe it's fine to just leave it with the messy error given that the goal is to eventually migrate everything over to react... idk

kchelstowski: @EWaterman the reason for such error is that the data services throw validation exception when validation didn't pass.
It works the same way as the `.save(failOnError: true)` would work. 
Regular `.save()` returns `null` if the validation didn't pass.
On GSP pages we usually tend to handle such cases like that:
```
if (x.validate() && x.save()) {
  render(view: ...)
  return
}
// If error occured
render(view: ...)
```
and then inside view:
```
<g:hasErrors bean="${x}">
            <div class="errors" role="alert" aria-label="error-message">
                <g:renderErrors bean="${x}" as="list" />
            </div>
</g:hasErrors>
```
and the errors are listed in user-friendly way.

Since data service's `.save()` throws en error, you'd have to wrap that in `try/catch` clause instead, but I'm not a fan of try/catch clause, because there were some inconsistencies in transaction rollback, but I do not remember now if it was Grails 1 or Grails 3 thing (would have to find our discussions from the past).

EWaterman: I looked into it and it's actually a really simple fix. It does have the frontend error processing that you suggest Kacper. The thing that was missing was the x.validate() step. Changing the code to the following works:
```
if (locationTypeInstance.validate() && locationTypeDataService.save(locationTypeInstance)) {
    flash.message = "${warehouse.message(code: 'default.updated.message', args: [...])}"
    redirect(action: "list", id: locationTypeInstance.id)
}
else {
    render(view: "edit", model: [locationTypeInstance: locationTypeInstance])
}
```

Screenshot from 2025-01-10 10-13-07

EWaterman:
As a refresher for the review (since it has been a while), the main change is to make LocationType.locationTypeCode non-nullable.

As a safety measure, I also migrate any existing nulls to use DEFAULT. There shouldn't be any location types with a null code (we don't have any in our environments) since it doesn't really make sense for it to be null, but the migration is there just in case any other implementations have it.

Artur suggested we use an empty/blank value instead of DEFAULT. I'm okay with this, but in the frontend, that would allow users to not select a value for location type code from the dropdown. Because all location types *should* have a code, I'm not sure this is the behaviour we want. I think it's better to explicitly fail to the user and force them to pick a code. Location types with a null/DEFAULT/blank value should need to be fixed because they can cause unexpected behaviour in the system, so I think we shouldn't allow creating new types without a code.

I suppose we could create a Command object (right now it just uses the LocationType domain) that has `locationTypeCode(nullable: false)` and have the domain use `locationTypeCode(nullable: false, blank:true)` like Artur suggested. We'd still need the migration though.

jmiranda (approved):
This looks good to me. The only issue I found was that if somehow the location type code was empty string or some invalid value (i.e. INVALID), you'd get redirected to the exception page when attempting to manage location types.

image

Not sure how likely that scenario is so just something to consider if we encounter a support request.

PR #4971, opened 2024-12-04
OBPIH-6825 Prioritize active persons when fetching by name or email

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6825

**Description:** Because we don't enforce any uniqueness on name or email in our system for entries in the Person table, when fetching people by name and email, there's a chance we'll find more than one. When this happens, we simply pick the first one we get. This can cause inconsistent behaviour when we have two identical users, one active and one inactive.

This was uncovered in PO import but is relevant to all flows that fetch a single Person.

This change is to prioritize returning active Person entries in the case where we have duplicates. 

image-20241204-151550

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/PersonService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/PersonService.groovy
@@ -25,7 +25,7 @@ class PersonService {
      */
     private Person getPersonByEmail(String email) {
         List<Person> people = Person.findAllByEmail(email)
-        if (people?.empty) {
+        if (people.empty) {
             return null
         }
 
@@ -87,8 +87,7 @@ class PersonService {
         }
 
         InternetAddress internetAddress
-        try
-        {
+        try {
             internetAddress = new InternetAddress(recipient, false)
         }
         catch (AddressException ignored) {
--- a/grails-app/services/org/pih/warehouse/data/PersonService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/PersonService.groovy
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
 
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
+import javax.mail.internet.AddressException
+import javax.mail.internet.InternetAddress
+import org.apache.commons.lang.StringUtils
+
 import org.pih.warehouse.core.Person
 
 @Transactional
 class PersonService {
 
-    Person getPersonByNames(String[] names) {
-        return Person.findByFirstNameAndLastName(names[0], names[1])
+    /**
+     * @return The first person we can find who has the given email, giving preference to active people.
+     */
+    private Person getPersonByEmail(String email) {
+        List<Person> people = Person.findAllByEmail(email)
+        if (people?.empty) {
+            return null
+        }
+
+        // We give preference to active people, so if there is one found, return it, otherwise return anyone.
+        return people.find{ it.active } ?: people.first()
     }
 
-    Person getPersonByNames(String combinedNames) {
-        String[] names = extractNames(combinedNames)
-        return getPersonByNames(names)
+    /**
+     * @return The first person we can find who is active and has the given email
+     */
+    Person getActivePersonByEmail(String email) {
+        return Person.findByActiveAndEmail(true, email)
     }
 
-    Person getOrCreatePersonFromNames(String combinedNames) {
+    /**
+     * Fetch a user via their full name. We require the combined name to be provided (instead of first name
+     * and last name individually) because while internally we split the name so that it conforms to the firstName and
+     * lastName fields in the Person table, we want to eventually move away from forcing that name structure upon users.
+     *
+     * @return Returns the first person we can find who has the given email, giving preference to active people.
+     */
+    private Person getPersonByName(String combinedNames) {
         String[] names = extractNames(combinedNames)
-        Person person = getPersonByNames(names)
-        if (!person) {
-            person = new Person(firstName: names[0], lastName: names[1])
-            if (!person.save(flush: true)) {
-                throw new ValidationException("Cannot save recipient ${combinedNames} due to errors", person.errors)
-            }
+        List<Person> people = Person.findAllByFirstNameAndLastName(names[0], names[1])
+        if (people?.empty) {
+            return null
+        }
+
+        // We give preference to active people, so if there is one found, return it, otherwise return anyone.
+        return people.find{ it.active } ?: people.first()
+    }
+
+    /**
+     * Fetch an active user via their full name. We require the combined name to be provided (instead of first name
+     * and last name individually) because while internally we split the name so that it conforms to the firstName and
+     * lastName fields in the Person table, we want to eventually move away from forcing that name structure upon users.
+     *
+     * @param combinedNames The person's full name (first name + last name) combined into a single String.
+     * @return A Person who is active and has a firstName and lastName matching the combinedNames
+     */
+    Person getActivePersonByName(String combinedNames) {
+        if (StringUtils.isBlank(combinedNames)) {
+            return null
+        }
+
+        String[] names = extractNames(combinedNames)
+        return Person.findByActiveAndFirstNameAndLastName(true, names[0], names[1])
+    }
+
+    /**
+     * @param recipient Either a RFC822O internet/email address (ex: "Justin Miranda <justin@openboxes.com>")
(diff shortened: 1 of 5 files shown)


EWaterman:
Marking this as do not merge for now since it's not essential to the release as it's not a new bug. If we feel confident about it though, I'm happy to try fitting it in.

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
+        return Person.findByActiveAndFirstNameAndLastName(true, names[0], names[1])
+    }
+
+    /**
+     * @param recipient Either a RFC822O internet/email address (ex: "Justin Miranda <justin@openboxes.com>")
+     *        or simply the recipient's name (ex: "Justin Miranda"). We accept both formats since either
+     *        format can be provided during data imports.
+     * @return A Person who has the provided name and email. Will be a new Person if they didn't previously exist.
+     */
+    Person getOrCreatePersonByRecipient(String recipient) {
```
EWaterman, line 84: for the most part this method is copied from ShipmentService. It did refactor it though (and created tests for it)

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
+        if (StringUtils.isBlank(recipient)) {
+            return null
+        }
+
+        InternetAddress internetAddress
+        try
+        {
```
alannadolny, line 91:
```suggestion
        try {
```

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
+     *        or simply the recipient's name (ex: "Justin Miranda"). We accept both formats since either
+     *        format can be provided during data imports.
+     * @return A Person who has the provided name and email. Will be a new Person if they didn't previously exist.
+     */
+    Person getOrCreatePersonByRecipient(String recipient) {
+        if (StringUtils.isBlank(recipient)) {
+            return null
+        }
+
+        InternetAddress internetAddress
```
alannadolny, line 89: I know you are not the original author of this, but maybe you can explain how and why we use `InternetAdress`?

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
+            return person
+        }
+
+        // Otherwise create a new person.
+        if (!internetAddress.personal) {
+            throw new RuntimeException("Cannot save new recipient without a name: ${recipient}")
+        }
+        String[] names = extractNames(internetAddress.personal)
+        person = new Person(firstName: names[0], lastName: names[1], email: internetAddress.address)
+        if (!person.save(flush: true)) {
```
alannadolny, line 112: is this flush needed?

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
+
         String[] names = extractNames(combinedNames)
-        Person person = getPersonByNames(names)
-        if (!person) {
-            person = new Person(firstName: names[0], lastName: names[1])
-            if (!person.save(flush: true)) {
-                throw new ValidationException("Cannot save recipient ${combinedNames} due to errors", person.errors)
-            }
+        person = new Person(firstName: names[0], lastName: names[1])
+        if (!person.save(flush: true)) {
```
alannadolny, line 132: and this one flush?

grails-app/services/org/pih/warehouse/shipping/CombinedShipmentService.groovy
```diff
@@ -224,12 +226,10 @@ class CombinedShipmentService {
 
                 Person recipient = line.recipient ? Person.get(line.recipient) : null
                 if (!recipient && line.recipient) {
-                    String[] names = line.recipient.split(" ")
-                    if (names.length == 2) {
-                        String firstName = names[0], lastName = names[1]
-                        recipient = Person.findByFirstNameAndLastName(firstName, lastName)
-                    } else if (names.length == 1) {
-                        recipient = Person.findByEmail(names[0])
+                    if (line.recipient.contains(" ")) {
```
alannadolny, line 229: I don't get it, why do we check if `line.recipient` contains " "?

alannadolny: + You can just use the ternary operator for such an assignment

grails-app/services/org/pih/warehouse/data/PersonService.groovy
```diff
@@ -11,33 +11,136 @@ package org.pih.warehouse.data
...
 
 @Transactional
 class PersonService {
 
-    Person getPersonByNames(String[] names) {
-        return Person.findByFirstNameAndLastName(names[0], names[1])
+    /**
+     * @return The first person we can find who has the given email, giving preference to active people.
+     */
+    private Person getPersonByEmail(String email) {
```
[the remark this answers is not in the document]
EWaterman: the find only returns a single active person. To me, if we were going to return a list of person I'd name the method getAllPersonByEmail or getPersonsByEmail or getPeopleByEmail

grails-app/services/org/pih/warehouse/data/PersonService.groovy
EWaterman: Internet address is an email of the format: "Justin Miranda <justin@openboxes.com>". So this code parses that in and outputs a person with:
- email: justin@openboxes.com
- firstName: Justin
- lastName: Miranda

I don't know why that format was the original requirement. Maybe @jmiranda has the answer. This format is only used during the import packing list flow.

grails-app/services/org/pih/warehouse/data/PersonService.groovy
EWaterman: probably not but it was in the original code so I didn't want to change it without knowing why it was added

grails-app/services/org/pih/warehouse/data/PersonService.groovy
EWaterman: Same answer. It was in the original code so I didn't want to change it without knowing why it was added

src/test/groovy/unit/org/pih/warehouse/person/PersonServiceSpec.groovy
```diff
@@ -0,0 +1,104 @@
...
+        expect:
+        service.getOrCreatePersonByRecipient(recipient)?.id == expectedPerson
+
+        where:
+        recipient       || expectedPerson
+        null            || null
+        ''              || null
+        'a b'           || '2'
+        '<1@1.com>'     || '2'
+        'e f <2@2.com>' || '3'  // email takes priority, even if name doesn't match and user is inactive.
```
EWaterman, line 71: This is an interesting quirk of the system. This is how the logic was working originally so I didn't change it, but it could be worth validating that both the name and email match the user if we're given both

grails-app/services/org/pih/warehouse/data/PersonService.groovy
jmiranda: Theoretically it shouldn't be necessary. But practically, it depends on where this code is being invoked. For adding a single person to the system, probably not. But in the previous version of Grails, things were less predictable when we were persisting lots of data all at once. Like say in a data import. It might be possible to remove the flush now (and I would be happy with that refactoring) but I appreciate Evan's instinct to leave well enough alone for now. And if/when we do make the change we need to ensure we have some integration tests around it to make sure it works for all use cases:
* add a single person in an admin page 
* add a single person per line item and assign them as recipient on the shipment item

Again it should work without the flush, but we should perform a discovery before making changes. This might be a good scenario to add to the Cascading Behavior on Associations tech debt that we've discussed in the past.

grails-app/services/org/pih/warehouse/data/PersonService.groovy
jmiranda: Yeah, i think i was originally trying to impose a data standard here https://www.ietf.org/rfc/rfc822.txt but was quickly overruled (within 10 days) when met with reality.

image

awalkowiak (approved):
Approved on the tech huddle

PR #4972, opened 2024-12-05
OBPIH-6837 TD: Install second react table package

alannadolny:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6837

**Description:** Because of the changes in packages in a newer version of react table I didn't have to use a forked library as a replacement for react-table version 6. I was able to add the newest version of the library and I tested it and it worked fine. The example that I tried to run in the app: https://tanstack.com/table/v8/docs/framework/react/examples/basic

---

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@fortawesome/free-solid-svg-icons": "5.14.0",
     "@fortawesome/react-fontawesome": "0.1.8",
     "@hookform/resolvers": "3.3.4",
+    "@tanstack/react-table": "8.20.5",
     "@testing-library/react-hooks": "8.0.1",
     "axios": "1.6.8",
     "bootstrap": "4.6.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4000,6 +4000,19 @@
         "@sinonjs/commons": "^3.0.0"
       }
     },
+    "@tanstack/react-table": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -59,6 +59,7 @@
     "@fortawesome/free-solid-svg-icons": "5.14.0",
     "@fortawesome/react-fontawesome": "0.1.8",
     "@hookform/resolvers": "3.3.4",
+    "@tanstack/react-table": "8.20.5",
```
EWaterman, line 62: is the plan to replace line 104 `"react-table": "6.8.6",` with this? Or do they work independently?

alannadolny: Yeah, we want to replace it in the future, but it will probably be in the far-distant future. For now, we would like to have those libraries independently, the new library includes a new "hook" approach which will be needed in the cycle count implementation.

EWaterman: sounds good. It's convenient that we can bring in both versions at the same time and then gradually switch over to using this new dependency.

My main concern is around having to support different behaviours across different tables, which may lead to inconsistencies in our UI, but I don't know the react table library well enough to know if that's a real concern or not.

Either way we'd probably need this upgrade eventually, so good that we're starting now

EWaterman (approved):

PR #4973, opened 2024-12-06
OBPIH-6867 Add comments tab for PO shipments

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovement.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovement.groovy
@@ -90,7 +90,8 @@ class OutboundStockMovement implements Serializable, Validateable {
             "pending",
             "electronicType",
             "approvers",
-            "pendingApproval"
+            "pendingApproval",
+            "isReturn"
     ]
 
     static mapping = {
@@ -339,4 +340,8 @@ class OutboundStockMovement implements Serializable, Validateable {
         return requisition?.approvalRequired && origin?.approvalRequired && RequisitionStatus.compare(requisition.status, RequisitionStatus.PICKING) == -1
     }
 
+    // This has to be named with the get prefix to align with the StockMovement DTO
+    boolean getIsReturn() {
+        return shipment?.isFromReturnOrder
+    }
 }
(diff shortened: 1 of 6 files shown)


grails-app/views/stockMovement/_comments.gsp
```diff
@@ -1,6 +1,7 @@
 <section class="box" aria-label="Comments">
     <h2><warehouse:message code="comments.label"/></h2>
-    <g:if test="${requisition?.comments}">
+    <g:set var="comments" value="${stockMovement?.requisition?.comments ?: stockMovement?.shipment?.comments}" />
```
EWaterman, line 3: so it is displaying requisition comments if any exist, otherwise shipment comments? But then below in the editComment and deleteComment actions always takes in stockMovementId. Can we not also edit requisition comments here?

Also, is there a need to display both requisition and shipment comments at the same time?

grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovement.groovy
```diff
@@ -339,4 +340,8 @@ class OutboundStockMovement implements Serializable, Validateable {
         return requisition?.approvalRequired && origin?.approvalRequired && RequisitionStatus.compare(requisition.status, RequisitionStatus.PICKING) == -1
     }
 
+    // This has to be named with the get prefix to align with the StockMovement DTO
```
EWaterman, line 343: you can't name it `isReturn()` and just call the full method (with parentheses) in the gsp?

awalkowiak (approved):

grails-app/views/stockMovement/_comments.gsp
kchelstowski: at this moment we prioritize requisition over shipment, so if a stock movement has both, we would add a comment to the requisition.

kchelstowski: moreover - the requisition is always created before the shipment, so we shouldn't run into a situation where firstly we were adding comments to shipment (because requisition wouldn't exist), and then started adding comments to requisition.

alannadolny (approved):

PR #4974, opened 2024-12-07
OBPIH-6472 Reintroduce shipment events tab in stock movement UI (fixes after QA)

alannadolny:


src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -38,6 +38,7 @@ class Constants {
     static final String DISPLAY_DATE_FORMAT = "MMM DD, yyyy"
     static final String DISPLAY_DATE_DEFAULT_VALUE = "-"
     static final String GENERATE_NAME_DATE_FORMAT = "ddMMMyyyy"
+    static final String DATE_SHORT_TIME_FORMAT = "dd/MM/yyyy HH:mm"
```
alannadolny, line 41: I don't have any better idea for the name of that format, I am open to suggestions

kchelstowski: since it just adds time to the `EUROPEAN_DATE_FORMAT` maybe this could be just:
```groovy
static final String EUROPEAN_DATE_FORMAT_WITH_TIME = "${EUROPEAN_DATE_FORMAT} HH:mm"
```
we could even consider creating a constant for the `HH:mm`.

src/main/groovy/org/pih/warehouse/core/HistoryItem.groovy
```diff
@@ -9,7 +9,28 @@ class HistoryItem<T> {
...
     EventCode eventCode
 
+    String eventTypeName
+
     Comment comment
 
     User createdBy
+
+    @Override
+    int compareTo(Object o) {
```
kchelstowski, line 19: nitpicky, but the type could be `HistoryItem`, couldn't it?

src/main/groovy/org/pih/warehouse/core/Constants.groovy
awalkowiak: I am not sure now about the `EUROPEAN_` prefix there. I know it was introduced in the past, but it feels like this should be something different. I guess for now we can leave it as is, but we could think about adjusting this one later (if and how  we want to have it)

PR #4981, opened 2024-12-16
OBPIH-6870 Add endpoint to fetch cycle count candidates

kchelstowski:

**Description:**
To be added tomorrow

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -1,0 +1,17 @@
+package org.pih.warehouse.api
+
+import grails.converters.JSON
+import org.pih.warehouse.inventory.CycleCountCandidate
+import org.pih.warehouse.inventory.CycleCountCandidateFilterCommand
+import org.pih.warehouse.inventory.CycleCountService
+
+class CycleCountApiController {
+
+    CycleCountService cycleCountService
+
+    def getCandidates(CycleCountCandidateFilterCommand filterParams) {
+        List<CycleCountCandidate> candidates = cycleCountService.getCandidates(filterParams, params.facilityId)
+
+        render([data: candidates, totalCount: candidates.totalCount] as JSON)
+    }
+}
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
@@ -1,0 +1,65 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.Tag
+import org.pih.warehouse.product.Product
+
+class CycleCountCandidate {
+
+    String id
+
+    Inventory inventory
+
+    Product product
+
+    Location facility
+
+    String abcClass
+
+    Integer inventoryItemCount
+
+    String internalLocations
+
+    Integer quantityOnHand
+
+    Integer quantityAvailable
+
+    Integer negativeItemCount
+
+    Date dateLastCount
+
+    Date dateNextCount
+
+    Date dateLatestInventory
+
+    static constraints = {
+        version false
+        table "cycle_count_session"
+    }
+
+
+    Map toJson() {
+        return [
+                product: [
+                        id: product.id,
+                        name: product.name,
+                        productCode: product.productCode,
+                ],
+                dateLastCount: dateLastCount,
+                category: [
+                        id: product.category?.id,
+                        name: product.category?.name,
+                ],
+                internalLocations: internalLocations,
+                tags: product.tags?.collect { Tag tag ->
+                    [
+                        id: tag.id,
+                        tag: tag.tag,
+                    ]
+                },
+                productCatalogs: product.productCatalogs,
+                abcClass: abcClass,
+                quantityOnHand: quantityOnHand,
+        ]
+    }
+}
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
@@ -51,15 +51,19 @@ class CycleCountCandidate {
                         name: product.category?.name,
                 ],
                 internalLocations: internalLocations,
-                tags: product.tags?.collect { Tag tag ->
-                    [
-                        id: tag.id,
-                        tag: tag.tag,
-                    ]
-                },
+                tags: tagsToJson(),
                 productCatalogs: product.productCatalogs,
                 abcClass: abcClass,
                 quantityOnHand: quantityOnHand,
         ]
+    }
+
+    List<Map> tagsToJson() {
+        return product.tags?.collect { Tag tag ->
+            [
+                    id : tag.id,
+                    tag: tag.tag,
+            ]
+        }
     }
 }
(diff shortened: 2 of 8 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -0,0 +1,102 @@
...
+                usedAliases.add("tags")
+                "in"("tags.id", command.tags.collect { it.id })
+            }
+            if (command.sort) {
+                getCandidatesSortOrder(command.sort, command.order, delegate, usedAliases)
+            }
+        }
+    }
+
+    private static void getCandidatesSortOrder(String sort, String orderDirection, Criteria criteria, Set<String> usedAliases) {
```
alannadolny, line 63: I would name `sort` -> `sortBy`. Sort sounds like it should be `desc` or `asc`

kchelstowski: I'm fine with renaming it, but in my mind `sort` is what we sort by, and `desc`/`asc` is the direction.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCandidateFilterCommand.groovy
```diff
@@ -0,0 +1,40 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+import org.pih.warehouse.api.PaginationCommand
+import org.pih.warehouse.core.Tag
+import org.pih.warehouse.product.Category
+import org.pih.warehouse.product.ProductCatalog
+
+class CycleCountCandidateFilterCommand extends PaginationCommand implements Validateable {
```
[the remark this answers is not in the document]
kchelstowski: It's not achievable yet, because the cycle count table doesn't exist. Moreover I think we still need to determine how we would want to join the status data to this query. In my mind we would have to modify the view to join the cycle count table to return status. I would not recommend doing an additional query to cycle count table and then filtering it out statically, because we might lose the pagination.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -0,0 +1,102 @@
...
+import grails.validation.ValidationException
+import org.grails.datastore.mapping.query.api.Criteria
+import org.hibernate.criterion.Order
+import org.hibernate.sql.JoinType
+import org.pih.warehouse.core.Location
+
+@Transactional
+class CycleCountService {
+
+    List<CycleCountCandidate> getCandidates(CycleCountCandidateFilterCommand command, String facilityId) {
```
alannadolny, line 13: don't we need a filter for this field? Or has a ticket already been created for that?
image

kchelstowski: I skipped that part for now, because those fields were meant to "include later". For now negative item count is hardcoded as `null`, and I am not sure what this actually will be later.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCandidateFilterCommand.groovy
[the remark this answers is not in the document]
EWaterman: We discussed this in the meeting, but I agree that we should modify the view to join Cycle Counts, and also Cycle Count Requests. We likely need both because Request is the table that holds productId and so that's how we'll join with Candidate. And then we'll fetch the Cycle Count associated with that request to get the Cycle Count status (for the purpose of filtering out rows)

But yes that's for a future ticket

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -0,0 +1,17 @@
...
+import org.pih.warehouse.inventory.CycleCountService
+
+class CycleCountApiController {
+
+    CycleCountService cycleCountService
+
+    def getCandidates(CycleCountCandidateFilterCommand filterParams) {
+        List<CycleCountCandidate> candidates = cycleCountService.getCandidates(filterParams, params.facilityId)
+
+        render([data: candidates, totalCount: candidates.totalCount] as JSON)
```
EWaterman, line 15: is the "totalCount" what will be used to display the number next to the ALL PRODUCTS tab?

If so, I'm not sure this will work if we apply any filters since that will shrink the result set of the query (and presumably the totalCount) but we want the number to continue to show for the total number of un-filtered elements.

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
```diff
@@ -0,0 +1,69 @@
...
+
+    Date dateLatestInventory
+
+    static constraints = {
+        version false
+        table "cycle_count_session"
+    }
+
+
+    Map toJson() {
```
EWaterman, line 41: I'm surprised this is performant since we're fetching products, then their categories, catalogs, and tags. It seems like it would result in a lot of queries.

But if it's quick enough like you said, I'm fine with it for now.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -0,0 +1,102 @@
...
+        if (command.hasErrors()) {
```
EWaterman, line 14: :+1: for the way you've set up the command to be validated at the start

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -0,0 +1,102 @@
...
+            throw new ValidationException("Invalid params", command.errors)
+        }
+        Location facility = Location.read(facilityId)
```
EWaterman, line 17: I assume "read" is like find but for read-only. Seems like we should be doing this more often :+1:

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
kchelstowski: @EWaterman yeah, since the results are paginated, it's not that slow. The slowest part of any of those queries is the part when the view is re-built. 
As soon as the PR is approved I will prepare somewhat a description of the endpoint with avg times. Anyway, it was not more than 3s in any of the scenarios.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: `totalCount` is meant for the pagination, so as you mentioned, whenever any filters are applied, the `totalCount` changes as well. I haven't thought about the number next to the "ALL PRODUCTS", but if you say, this should not change regardless the filters, we could just set it on frontend once (during the initial fetch, when none filters would be applied), and not update it whenever we refetch the data.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: For now none exception is thrown. Since we need the location id in the URL, it's a very rare case that the location would not exist. Even if it doesn't, the output would just be: 
```json
{
  data: [],
  totalCount: 0,
}
```
I'm fine with addressing this edge case later, if we are not comfortable with the empty result for non-existing location, which is barely to happen.

awalkowiak (approved):

PR #4985, opened 2024-12-18
WMS-17 Add list putaways endpoint

jdolkowski:

Added additional filtering options in OrderService.getOrders()
Change the arguments definition of getOrders to be able using only orderTemplate. Existing bechavior stayed untouched.
Add new endpoint that list whole objects of Putaway

https://allofresh.atlassian.net/browse/WMS-17

Related to: https://github.com/openboxes/openboxes-mobile/pull/277

image


grails-app/services/org/pih/warehouse/order/OrderService.groovy
```diff
@@ -168,6 +168,39 @@ class OrderService {
...
+                    }
+
+                    orderItems {
+                        destinationBinLocation {
+                            or {
+                                ilike("name", "%" + params.q + "%")
+                                ilike("locationNumber", "%" + params.q + "%")
+                            }
+                        }
+                    }
```
jdolkowski, line 203: To the person who will be doing the review: Please make sure that this not affect existing filtering.
In my understainding it onlly adds new options.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -439,6 +439,14 @@ class UrlMappings {
             action = [DELETE: "removingItem"]
         }
 
+        // Putaway API
+
+        // MOBILE
+        "/api/listPutaways"(parseRequest: true) {
```
awalkowiak, line 445: If this interferes with the current `api/putaways` from develop, it is fine if it's split, but I would put it under a different endpoint (I'd avoid the `listPutaways` here)

grails-app/services/org/pih/warehouse/order/OrderService.groovy
awalkowiak: I think it is ok, but would need to be tested

PR #4990, opened 2024-12-20
WMS-25 Add global search endpoint

jdolkowski:

Adding logic for new endpoint globalSearch that is used by mobile to search the barcodes.

https://allofresh.atlassian.net/browse/WMS-25


jdolkowski:
DashboardController.globalSearch() seems to have similar logic but i am not sure if it is the same thing

PR #4991, opened 2025-01-02
OBPIH-6903 Implement cycle count request API (create)

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -937,6 +937,11 @@ class UrlMappings {
             action = [GET: "getCandidates"]
         }
 
+        "/api/facilities/$facilityId/cycle-counts/requests" {
+            controller = "cycleCountApi"
+            action = [POST: "createRequest"]
+        }
+
         // Error handling
 
         "401"(controller: "errors", action: "handleUnauthorized")
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -937,9 +937,9 @@ class UrlMappings {
             action = [GET: "getCandidates"]
         }
 
-        "/api/facilities/$facilityId/cycle-counts/requests" {
+        "/api/facilities/$facilityId/cycle-counts/requests/batch" {
             controller = "cycleCountApi"
-            action = [POST: "createRequest"]
+            action = [POST: "createRequests"]
         }
 
         // Error handling
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -5,8 +5,11 @@ import grails.validation.ValidationException
 import org.pih.warehouse.inventory.CycleCountCandidate
 import org.pih.warehouse.inventory.CycleCountCandidateFilterCommand
 import org.pih.warehouse.inventory.CycleCountRequest
+import org.pih.warehouse.inventory.CycleCountRequestBatchCommand
 import org.pih.warehouse.inventory.CycleCountRequestCommand
 import org.pih.warehouse.inventory.CycleCountService
+import org.springframework.validation.BeanPropertyBindingResult
+import org.springframework.validation.Errors
 
 class CycleCountApiController {
 
@@ -18,11 +21,27 @@ class CycleCountApiController {
         render([data: candidates, totalCount: candidates.totalCount] as JSON)
     }
 
-    def createRequest(CycleCountRequestCommand command) {
+    def createRequests(CycleCountRequestBatchCommand command) {
         if (command.hasErrors()) {
-            throw new ValidationException("Invalid cycle count", command.errors)
+            // Build errors manually to be able to include errors both for the batch command instance
+            // and errors for every element of the requests list
+            Errors errors = new BeanPropertyBindingResult(command, "requests")
+            // Iterate every element of requests and add its error to the errors instance.
+            command.requests.each { CycleCountRequestCommand request ->
+                request.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            // If there are not any errors in the elements, it means that there must be errors in the batch command instance
+            // e.g. missing required arguments
+            if (!errors.hasErrors()) {
+                command.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            throw new ValidationException("Invalid cycle count", errors)
         }
-        List<CycleCountRequest> cycleCountRequests = cycleCountService.createRequest(command)
+        List<CycleCountRequest> cycleCountRequests = cycleCountService.createRequests(command)
         render([data: cycleCountRequests] as JSON)
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -1,8 +1,11 @@
 package org.pih.warehouse.api
 
 import grails.converters.JSON
+import grails.validation.ValidationException
 import org.pih.warehouse.inventory.CycleCountCandidate
 import org.pih.warehouse.inventory.CycleCountCandidateFilterCommand
+import org.pih.warehouse.inventory.CycleCountRequest
+import org.pih.warehouse.inventory.CycleCountRequestCommand
 import org.pih.warehouse.inventory.CycleCountService
 
 class CycleCountApiController {
@@ -13,5 +16,13 @@ class CycleCountApiController {
         List<CycleCountCandidate> candidates = cycleCountService.getCandidates(filterParams, params.facilityId)
 
         render([data: candidates, totalCount: candidates.totalCount] as JSON)
+    }
+
+    def createRequest(CycleCountRequestCommand command) {
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid cycle count", command.errors)
+        }
+        List<CycleCountRequest> cycleCountRequests = cycleCountService.createRequest(command)
+        render([data: cycleCountRequests] as JSON)
     }
 }
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -39,7 +39,7 @@ class CycleCountApiController {
                     errors.addError(error)
                 }
             }
-            throw new ValidationException("Invalid cycle count", errors)
+            throw new ValidationException("Invalid cycle count request", errors)
         }
         List<CycleCountRequest> cycleCountRequests = cycleCountService.createRequests(command)
         render([data: cycleCountRequests] as JSON)
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
@@ -41,7 +41,7 @@ class CycleCountRequest {
 
     static constraints = {
         id generator: "uuid"
-        product(nullable: true, unique: "facility")
+        product(nullable: true)
         createdBy(nullable: true)
         updatedBy(nullable: true)
     }
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
@@ -1,0 +1,61 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.auth.AuthService
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.User
+import org.pih.warehouse.product.Product
+
+class CycleCountRequest {
+
+    String id
+
+    Location facility
+
+    Product product
+
+    CycleCountStatus status
+
+    // To be included when cycle count table is created
+    // CycleCount cycleCount
+
+    CycleCountRequestType requestType
+
+    Boolean blindCount
+
+    Date dateCreated
+
+    Date lastUpdated
+
+    User createdBy
+
+    User updatedBy
+
+    def beforeInsert() {
+        createdBy = AuthService.currentUser
+        updatedBy = AuthService.currentUser
+    }
+
+    def beforeUpdate() {
+        updatedBy = AuthService.currentUser
+    }
+
+    static constraints = {
+        id generator: "uuid"
+        product(nullable: true, unique: "facility")
+        createdBy(nullable: true)
+        updatedBy(nullable: true)
+    }
+
+    Map toJson() {
+        return [
+                id: id,
+                facility: facility.toBaseJson(),
+                product: product,
+                status: status.toString(),
+                requestType: requestType.toString(),
+                blindCount: blindCount,
+                dateCreated: dateCreated,
+                lastUpdated: lastUpdated,
+        ]
+    }
+}
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
@@ -13,7 +13,7 @@ class CycleCountRequest {
 
     Product product
 
-    CycleCountStatus status
+    CycleCountRequestStatus status
 
     // To be included when cycle count table is created
     // CycleCount cycleCount
--- a/grails-app/migrations/0.9.x/changelog-2024-12-18-1300-create-table-cycle-count-request.xml
+++ b/grails-app/migrations/0.9.x/changelog-2024-12-18-1300-create-table-cycle-count-request.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="kchelstowski" id="181220241300-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="cycle_count_request"/>
+            </not>
+        </preConditions>
+
+        <createTable tableName="cycle_count_request">
+            <column name="id" type="CHAR(38)">
+                <constraints nullable="false" primaryKey="true" />
+            </column>
+            <column name="version" type="BIGINT">
+                <constraints nullable="false" />
+            </column>
+            <column name="facility_id" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+            <column name="product_id" type="CHAR(38)">
+                <constraints nullable="true" />
+            </column>
+            <column name="status" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+            <column name="cycle_count_id" type="CHAR(38)">
+                <constraints nullable="true" />
+            </column>
+            <column name="blind_count" type="BIT">
+                <constraints nullable="false" />
+            </column>
+            <column name="request_type" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+            <column name="date_created" type="DATETIME">
+                <constraints nullable="false" />
+            </column>
+            <column name="last_updated" type="DATETIME">
+                <constraints nullable="false" />
+            </column>
+            <column name="created_by_id" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+            <column name="updated_by_id" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="kchelstowski" id="181220241300-1">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <foreignKeyConstraintExists foreignKeyName="fk_cycle_count_request_location"/>
+            </not>
+        </preConditions>
+        <addForeignKeyConstraint
+                baseColumnNames="facility_id"
+                baseTableName="cycle_count_request"
+                constraintName="fk_cycle_count_request_location"
+                deferrable="false"
+                initiallyDeferred="false"
+                referencedColumnNames="id"
+                referencedTableName="location"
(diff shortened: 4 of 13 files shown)


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -937,6 +937,11 @@ class UrlMappings {
             action = [GET: "getCandidates"]
         }
 
+        "/api/facilities/$facilityId/cycle-counts/requests" {
```
EWaterman, line 940: the [original design](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3332505612/Cycle+Count+Items+API#Create-Cycle-Count-Request) has: `/api/facilities/{facility_id}/cycle-counts/request/batch` (though now that I look at it, maybe `.../requests/batch` is better)

I did it that way to be more explicit about the fact that this is a batch API call. IMO `/cycle-counts/requests` would be used for simple CRUD APIs on a single CycleCountRequest. But maybe that's just personal preference.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -99,4 +101,30 @@ class CycleCountService {
             criteria.createAlias("product", "product", JoinType.INNER_JOIN)
         }
     }
+
+    List<CycleCountRequest> createRequest(CycleCountRequestCommand command) {
```
EWaterman, line 105: I'd do `createRequests` since this is creating multiple

src/main/groovy/org/pih/warehouse/inventory/CycleCountStatus.groovy
```diff
@@ -0,0 +1,14 @@
+package org.pih.warehouse.inventory
+
+enum CycleCountStatus {
```
EWaterman, line 3: I'll let Artur review these since he did the bulk of the work on statuses but do we need both CREATED and REQUESTED?

It's probably worthwhile adding docstrings to each of the statuses to describe what they mean. Some are obvious, but still probably helpful for future us to understand.

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+class CycleCountRequest {
+
+    String id
+
+    Location facility
+
+    Product product
+
+    CycleCountStatus status
```
EWaterman, line 16: I was assuming this would be a `CycleCountRequestStatus` enum. Won't a CycleCountRequest and CycleCount have separate statuses?

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -0,0 +1,33 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class CycleCountRequestCommand implements Validateable {
```
EWaterman, line 7: okay now I understand why you didn't make it a batch API call. I thought about doing it this way originally since it avoids needing to define the blindCount for each product, but I ended up [designing it with the batch](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3332505612/Cycle+Count+Items+API#Create-Cycle-Count-Request) since it was clearer and more flexible IMO.

With the batch it'd be making a POST on a List<CycleCountRequestCommand>:
```
/api/facilities/{facility_id}/cycle-counts/requests/batch

[
  {
    requestType: "MANUAL_REQUEST",
    blindCount: true,
    productId: "10001",
  },
  ...
],
```
the main advantage of this is that we can easily make non-batched calls by removing the `/batch` from the URI and using the same CycleCountRequestCommand object:
```
/api/facilities/{facility_id}/cycle-counts/requests

{
  requestType: "MANUAL_REQUEST",
  blindCount: true,
  productId: "10001",
}
```
which IMO makes the most intuitive sense structurally. Let me know your thoughts

grails-app/migrations/0.9.x/changelog-2024-12-18-1300-create-table-cycle-count-request.xml
```diff
@@ -0,0 +1,120 @@
...
+            <column name="version" type="BIGINT">
+                <constraints nullable="false" />
+            </column>
+            <column name="facility_id" type="CHAR(38)">
+                <constraints nullable="false" />
+            </column>
+            <column name="product_id" type="CHAR(38)">
+                <constraints nullable="true" />
+            </column>
+            <column name="status" type="CHAR(38)">
```
EWaterman, line 26: should this be VARCHAR since it's based on an enum?

grails-app/migrations/0.9.x/changelog-2024-12-18-1300-create-table-cycle-count-request.xml
```diff
@@ -0,0 +1,120 @@
...
+                <constraints nullable="false" />
+            </column>
+            <column name="cycle_count_id" type="CHAR(38)">
+                <constraints nullable="true" />
+            </column>
+            <column name="blind_count" type="BIT">
+                <constraints nullable="false" />
+            </column>
+            <column name="request_type" type="CHAR(38)">
```
EWaterman, line 35: same here. Should be VARCHAR?

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
kchelstowski: I'm fine with adding `/batch`.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: yeah, forgot to change it, I had the same feeling at some point, thanks for pointing that out

src/main/groovy/org/pih/warehouse/inventory/CycleCountStatus.groovy
kchelstowski: I consulted it with Artur before, and we agreed that for now we can go the easiest way possible and elaborate on it either in the code review or refactor it later, since this wouldn't be a difficult change.

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
kchelstowski: Yeah, if so, I can rename it to `CycleCountRequestStatus`.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
EWaterman: PS, in my original docs I had the batch payload as:
```
{
  requests: [
    {
      requestType: "MANUAL_REQUEST",
      blindCount: true,
      productId: "10001",
    },
    ...
  ],
}
```
so it'd have a separate `CycleCountRequestBatchCommand` object. This way, we can still have the a constraints block that you have for verifying things like not having the same product twice in the `List<CycleCountRequestCommand>`

kchelstowski: Actually since this is **product based**, and the `facility`, `requestType`, `blindCount` remains the same for all products, the way I built it is:
```
{
    "facility": "8a8a9e9665c4f59d0165c54ec6b10027",
    "products": [
        "10004",
        "10005",
    ],
    "blindCount": true
}
```
and then I loop through the `command.products`.
I thought it was redundant to build it like you've just shown, because the only "variable" here is the `productId`.

Aside:
This was also easier from the technical point of view, because if I had to do it in the way you described, we would need two command classes to make it work, so:
```
class Command1 {
  List<Command2> command
}
```
and
```
class Command2 {
  CycleCountRequestType requestType
  Boolean blindCount
  Product product
}
```
because we can't bind a command with list type, so something like this:
```
def createRequests(List<Command2> command)
```
wouldn't work

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+    List<Product> products
+
+    Boolean blindCount
+
+    static constraints = {
+        products(nullable: true, validator: { List<Product> products, CycleCountRequestCommand obj ->
+            String productCode = null
+            boolean duplicateExists = products.any { Product product ->
+                // FIXME: Most probably, this will have to be modified not to include completed cycle counts
```
EWaterman, line 19: yeah you'd want to change the findBy to something like a `countByProductAndFacilityAndStatus` (where status not in [COMPLETED, CANCELED])

grails-app/migrations/0.9.x/changelog-2024-12-18-1300-create-table-cycle-count-request.xml
kchelstowski: you are right, I've just copied the type from the `id`. Will fix it.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
kchelstowski: yeah, I left that `FIXME` to remember, as this will be doable after `CycleCount` domain is created and working.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -99,4 +101,30 @@ class CycleCountService {
...
+            )
+            cycleCountsRequests.add(cycleCountRequest)
+        }
+        cycleCountsRequests.each { CycleCountRequest cycleCountRequest ->
+            if (!cycleCountRequest.validate()) {
+                throw new ValidationException("Invalid cycle count request", cycleCountRequest.errors)
+            }
+            // Flush is needed to know what has already been persisted inside this loop. Without flush, we could save two CCR for the same product and facility.
+            // This won't be doable via UI, but could be possible via API.
+            cycleCountRequest.save(flush: true)
```
EWaterman, line 126: can't you prevent this in the constraints block of the CycleCountRequestCommand? Loop through all products and check for duplicates. That way we can fail immediately if there are two products provided.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -0,0 +1,33 @@
...
+                CycleCountRequest cycleCountRequest = CycleCountRequest.findByProductAndFacility(product, obj.facility)
+                if (cycleCountRequest) {
+                    productCode = product.productCode
```
EWaterman, line 22: using productCode causes this to do a lookup on product for each product in the list, correct? I was hoping we could do this validation entirely on productId (which I was assuming would avoid the lookup since we already have the ids in the request body).

But actually this is making me realize I don't know how Grails is binding domains to request/command objects. Is it smart enough to not pre-fetch the products until they're used?

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
EWaterman: yeah that's the tradeoff. I do still prefer having the XCommand and XBatchCommand (which wraps a List<XCommand>) objects since IMO that gives the clearest representation and isolation of the data and the operations we're performing, but I definitely get the appeal of wanting the simpler API that avoids redundant field duplication. Ultimately I'm fine with it the way you have it designed, it just gives us a bit less flexibility in the future if we need to change the API requirements.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: I can handle most of the cases in the validator, and will do it. This line was going a bit ahead of what the typical use cases would be - this line e.g. prevents us from persisting a cycle count request for a product if e.g. two people were performing it at one time, but I guess it's barely to happen so we can live with just validator validation to group products and see if there are any duplicates in the list.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
kchelstowski: `Product` is bound at the binding step, so at this point nothing is fetched. It also doesn't cause any N+1, because `productCode` is a string field. Potential N+1 would be caused, if I was accessing a collection here.

Aside: I only store and assign the `productCode` here for a user-friendly validation message, so that the user knows which product it is about (id is not always user-friendly for products, because it's not always id == productCode)

kchelstowski: And also I didn't see a point to have the property `String productId` in the command instead of `Product product`, because anyway we'd have to fetch the product at some point, and here, Grails does it automatically for us.

kchelstowski: this also gives us much flexibility while validating in the constraints and in the validator, because we can already operate on the product and its properties, instead of validating the `String` property, where the only possibilities would be to validate if it's null/blank.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: it's an interesting point. We don't want two users creating CCRs at the same time, causing two requests to be inserted for the same product, breaking the system.

As you mentioned it's not a huge issue for us in our current state since parallel save operations are uncommon, but as we continue to scale up we'll likely see it more often. I think this speaks to a larger problem with how Hibernate propagates the database session across requests. That said, our system as a whole isn't really set up to handle operating at the scale at which it would start being a real problem, and so there are likely many other challenges we'll have to fix first (horizontal scaling with sticky sessions and a load balancer, database scaling and replication...) before this one even becomes a problem for us.

So if we do want to deal with the problem, it'll probably require a deeper investigation as to the best approach to take, and that probably isn't needed for a while (at least I hope). For now I think we can assume it'll be fine.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
EWaterman: I was assuming we could bypass fetching the Product at the binding step entirely and just use the id since Product isn't actually used anywhere (except your error case but I'm fine with fetching it there). All we do is map it to the request.

But thinking about it, we do need to verify that the given id is pointing to a real product, so we'll need the lookup regardless.

Okay I'm convinced.

kchelstowski:
@EWaterman I rebuilt the logic and implemented the approach suggested by you, so that the payload would look like:
```
{
 "requests": [
     {
         "product": "10006",
         "facility": "8a8a9e9665c4f59d0165c54ec6b10027",
         "blindCount": true
     },
      {
         "product": "10006",
         "facility": "8a8a9e9665c4f59d0165c54ec6b10027",
         "blindCount": true
     },
      {
         "product": "10007",
         "facility": "8a8a9e9665c4f59d0165c54ec6b10027",
         "blindCount": true
     },
      {
         "product": "10007",
         "facility": "8a8a9e9665c4f59d0165c54ec6b10027",
         "blindCount": true
     }
 ]
}
```
I also addressed the rest of suggested improvements/change requests.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestBatchCommand.groovy
```diff
@@ -0,0 +1,28 @@
...
+
+    List<CycleCountRequestCommand> requests
+
+    static constraints = {
+        requests(validator: { List<CycleCountRequestCommand> requests ->
+            // First catch if any duplicates have been provided (e.g. 10006, 10006 in the payload)
+            // This doesn't check though if any duplicate is already persisted, this is checked in the CCR command below (.each + .validate() clause)
+            Map<String, List<Product>> duplicates = requests.product
+                    .groupBy { it.productCode }
+                    .findAll { it.value.size() > 1 }
```
EWaterman, line 16: you could probably do a countBy productCode (and then filter to > 1) so that you only need to maintain a List<String> (of duplicate productCodes) instead of a map but that's a tiny optimization so I'm not worried about it.

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestBatchCommand.groovy
```diff
@@ -0,0 +1,28 @@
...
+            if (!duplicates.isEmpty()) {
+                return ['duplicateExists', duplicates.keySet().toString()]
+            }
+            // Elements of a list are not validated by default, so proceed manual validation of every element in the list
+            requests.each { CycleCountRequestCommand command -> command.validate() }
+            // If any of elements have validation errors, throw an exception
+            if (requests.any { it.hasErrors() }) {
+                return false
```
EWaterman, line 24: what happens if instead of returning false you return the actual requests.errors? Would that save you from needing to build the errors yourself in the controller?

kchelstowski: it's not how the validator works. It expects you either to return true/false (validation passed/didn't pass) or to return a list with a key of an error message and optionally arguments of the message.
Looking for example at a few lines above
```
 if (!duplicates.isEmpty()) {
  return ['duplicateExists', duplicates.keySet().toString()]
}
```
this would point to the message:
```
cycleCountRequestBatchCommand.requests.duplicateExists
```
and the `duplicates.keySet().toString()` would be accesible via `{3}`.
Given that you would have:
```
return ['duplicateExists', duplicates.keySet().toString(), 'something']
```
the `something` would be accesible under `{4}` etc.
`{0}`, `{1}`, `{2}` are reserved for: 

> Parameters 0 to 2 are automatically mapped to 0: property name, 1: class name, 2: property value. Additional parameters are mapped starting at parameter 3.

The `return false` only indicates that there is an error in `requests` field, so that when you further access the `command.errors.requests`, it would not be empty.
More info here: https://docs.grails.org/latest/ref/Constraints/validator.html

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -0,0 +1,25 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class CycleCountRequestCommand implements Validateable {
+
+    Location facility
```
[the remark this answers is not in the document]
kchelstowski: Theoretically I could, but I would lose any validation for `facility`. Having it here, I don't have to check manually whether it's `null` or not, because it would be validated out of the box (remember that potentially users might use the api and paste into `{facility_id}` whatever they want).

alannadolny (approved):

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -0,0 +1,61 @@
...
+    String id
+
+    Location facility
+
+    Product product
+
+    CycleCountRequestStatus status
+
+    // To be included when cycle count table is created
+    // CycleCount cycleCount
```
awalkowiak, line 19: Was it agreed that it can be skipped now? To add it we will have to write a second migration. Would not it be better if we had the core cycle count domain here as well?

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -0,0 +1,61 @@
...
+        updatedBy = AuthService.currentUser
+    }
+
+    def beforeUpdate() {
+        updatedBy = AuthService.currentUser
+    }
+
+    static constraints = {
+        id generator: "uuid"
+        product(nullable: true, unique: "facility")
```
awalkowiak, line 44: Is the cycle count request meant to be deleted at the end once it is completed? Right now if now if it won't be deleted, then it will prevent to requesting it again in the future due to this constraint. Was there a case when the product is null on the CCR level?

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -14,4 +20,28 @@ class CycleCountApiController {
...
+                }
+            }
+            // If there are not any errors in the elements, it means that there must be errors in the batch command instance
+            // e.g. missing required arguments
+            if (!errors.hasErrors()) {
+                command.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            throw new ValidationException("Invalid cycle count", errors)
```
awalkowiak, line 42: "Invalid cycle count **request**"?

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -0,0 +1,25 @@
...
+
+    Product product
+
+    Boolean blindCount
+
+    static constraints = {
+        product(nullable: true, validator: { Product product, CycleCountRequestCommand obj ->
+            // TODO: Most probably, this will have to be modified not to include completed cycle counts
+            CycleCountRequest cycleCountRequest = CycleCountRequest.findByProductAndFacility(product, obj.facility)
```
awalkowiak, line 18: ~Again the same question, did we want to delete the requests after completing, or should we look here for only `PENDING` Requests to find duplicates~ NVM, I just read the TODO, but the other constraint in the CCR domain itself need either improvement or TODO as well.

awalkowiak: But I think you could already use the `findByProductAndFacilityAndStatusNotInList(product, obj.facility, [COMPLETED, CANCELED])`

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestBatchCommand.groovy
EWaterman: I see, so you can only return a single failure here but we need to return a list of them. That's annoying.

Okay yes then we'll need to parse the errors as you're doing. Looking through the code, it seems like that's what we've done historically as well (ProductController.batchSave or ProductSupplierAttributeApiController.updateAttributes for example)

(I found [a blog post](https://joelforjava.com/blog/2018/09/25/grails-nested-validation-in-command-objects.html) which describes a method of handling this at the validator level, but it feels hacky so I don't think we should do it.)

I've been thinking about how we can standardize our error messages across our system and so maybe there's a way to do this type of "compiling all errors together" work more generically, but for now I'm good with your solution.

Thanks for doing the command object refactoring by the way. I wasn't expecting it to make this validation so messy.

EWaterman (approved):

PR #4993, opened 2025-01-09
OBPIH-6844 Create filters component all products tab

alannadolny:


src/js/hooks/useQueryParamsListener.js
```diff
@@ -0,0 +1,22 @@
...
+      params.map((param) => query.get(param)).join('|'),
+    [query, params],
+  );
+
+  useEffect(() => {
+    callback();
+  }, [paramsDependencies]);
+};
+
+export default useQueryParamsListener;
```
kchelstowski, line 22: what's the flow behind it? what is it used for? and what is the `.join('|')` responsible for?

alannadolny: `join` is to concat all of the params to one simple object to detect changes in params.
I had to implement detecting the changes in params this way because those filters have a lot of asynchronous operations and this fact led to the issues with sending requests with wrong parameters, so now it's reacting to changes in params.
I joined it with "|" but it can be something else.

awalkowiak (approved):

PR #4996, opened 2025-01-14
OBPIH-6905 API to fetch product classifications

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6905

**Description:** Added new GET /api/products/classifications API for fetching all unique abcClass values from Product and InventoryLevel and hook it up in the all products tab for cycle counts.

---
Screenshot from 2025-01-16 13-55-30


grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
```diff
@@ -0,0 +1,55 @@
...
+        productClassifications += InventoryLevel.createCriteria().listDistinct() {
+            projections {
+                groupProperty("abcClass")
+            }
+        } as Set<String>
+
+        // Filter out null since it's not a valid classification. Note that abcClass is allowed to be blank, so blank
+        // strings are considered valid classifications here. Likely we'll want to sanitize the data at some point to
+        // prevent blanks from being allowed, but since this is just a list call, we return whatever we have in the db.
+        productClassifications.remove(null)
```
jmiranda, line 51: You could add a where clause in the criteria query to remove the nulls as well.

grails-app/controllers/org/pih/warehouse/api/ProductClassificationApiController.groovy
```diff
@@ -0,0 +1,14 @@
+package org.pih.warehouse.api
+
+import grails.converters.JSON
+import org.pih.warehouse.product.ProductClassificationService
+
+class ProductClassificationApiController {
```
jmiranda, line 6: There are multiple classification systems (taxonomies) related to products, so we should be thoughtful about what this might turn into and not pigeonhole product classification as ABC class. For example, product classifications represent categories, tags, ABC classes, UNSPSC, hazardous classes, narcotics classes. So I would either be very specific with the API or leave room for using this API for different classifications.

grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * It's important to note that the classifications themselves contain no inherent value. It's up to the implementers
+ * to decide what a classification really means, which is determined by the feature-specific rules that are configured
+ * for each classification.
+ *
+ * For example, if you're using the ABC Classification rules, the classes "A", "B", "C" are just strings.
+ * The meaning comes in when we configure the cycle count feature to require class "A" products to be counted weekly,
+ * class "B" products to be counted monthly, and class "C" products to be counted yearly.
+ */
+@Transactional
```
jmiranda, line 25: I don't remember our convention, but we might want to make the list() method @Transactional(readOnly=true)?

grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
jmiranda:
```
Location facility = Location.get(params.facility)
Set<String> productClassifications = Product.createCriteria().listDistinct() {
    projections {
        groupProperty("abcClass")
    }
    isNotNull("abcClass")
} as Set<String>
productClassifications += InventoryLevel.createCriteria().listDistinct() {
    projections {
        groupProperty("abcClass")
    }
    equals("inventory", facility.inventory)
    isNotNull("abcClass")
} as Set<String>
```

alannadolny (commented):
For me, it's not looking like it is done what it should be:
1. ABC classifications had to be an option for selects. We are using `SelectOptionsApiController` so it should be placed in this controller and mapped in the same way as others (to have an ID and label). Label is the thing that is displayed in the dropdown, ID is the property that identifies these options while filtering (in this case it's probably just the class)
2. In the ticket it is written: 
> Additionally, the cycle count filtering must be modified to allow filtering by those options.

It is meant to connect frontend with this endpoint (Adding a URL to the endpoint to Urls.js, adding a method for fetching those options, and then simply using it in useCycleCountFiltering - probably just calling the method in Promise.all) If you will have questions you can write to me on Slack, so I can paste you what exact places should be changed (Probably will be much faster because I already know those places :smile: )

grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
EWaterman: I thought about doing that but I'm not sure why I decided not to. I'll switch it since it's better to pre-filter

grails-app/controllers/org/pih/warehouse/api/ProductClassificationApiController.groovy
EWaterman: yeah I'm not sure how to narrow this down because it feels like we're running out of words to describe the "type" of a product. Like you said, we have categories, tags, classes... and they're all used in different ways. My original idea for this was that it'd only be used for the "ABC class" style classifications (I didn't call it ProductABCClassApiController because we can support whatever prioritization ruleset that we want to use, not just ABC classification) but it does raise an interesting point that the line between all the different "types" is blurry and there might be some overlap.

EWaterman:
thanks for the heads up @alannadolny. I didn't know that was how we were populating our filters. The SelectOptionsApiController methods need an ID like you said but the ABC classes are just plain strings (there's no ProductClassification domain, we search columns of multiple tables to build the list) and so they don't have an id. I suppose we could just use index as id?

jmiranda:
> ABC classifications had to be an option for selects. We are using SelectOptionsApiController so it should be placed in this controller and mapped in the same way as others (to have an ID and label). 

I would like to get rid of SelectOptionsApiController at some point. I argued against it when it was first introduced but was overruled at the time. But I still want to refactor it out of the API as there seems to be no reason for a consolidated API that applies a different rendering of the same data returned by existing APIs. And if the argument is "well we don't have an API for X yet", then we should add an API for X instead of another listing method to SelectOptions. 

The client should be responsible for mapping a response to the UI, not the API. In other words, if I need a list of category options I should just call the Category API from the client and let the client deal with mapping / filtering the values properly? If we absolutely need the API to return a different representation of the category response, then we should use content negotiation (i.e. Content-Type: )

I ran into the same issue when we introduced the JQuery data tables plugin. I needed to conform the API response to allow the plugin to render the data table properly. And it made me feel dirty. 

The SelectOptionsApi is slightly different in that we control the shape and content of the response, but we're still telling the server to do something it shouldn't need to do. And if we do need to tell the server to do something different re: rendering data it should be through a convention like content negotiation.

EWaterman:
I tend to agree with Justin's comment. Our APIs shouldn't contain frontend-specific logic.

Likely the "correct" way to achieve this is via a list API that allows the client to optionally specify what fields it needs via query parameter.


That said, in this case I can see how it'd be useful to have a standardized API response format for when we need to simply enumerate all of something and don't want to return the entire object. It feels like the path of least resistance here is sticking with the format used in SelectOptionsApiController because the client table is already expecting that format. It's generic enough for me to be okay with it. I don't think this API should be in the SelectOptionsApiController though. I think it should live in the feature-specific ProductClassificationApiController. From there I can standarize this type of "enumeration API" by using a common response dto of the format:
```
[
  {
    'id':  ??
    'label': "A"
    'value': ??
  },
  ...
]
```

EWaterman:
I've added the frontend logic. I decided to stick with the original API and construct the expected format on the client side so that the API can stay flexible and frontend agnostic.

The only thing I haven't figured out is how to get the filters to actually apply when searching. Right now on the client it sends:
`/openboxes/api/facilities/xxx/cycle-count-candidates?abcClass=%5Bobject%20Object%5D&...` (which is "[object Object]") so it's clearly not parsing the classification name correctly somewhere. I just need to figure out where (I'm not well versed in debugging the frontend...)

grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
```diff
@@ -0,0 +1,65 @@
...
+class ProductClassificationService {
+
+    /**
+     * Fetches all product classifications that we have configured.
+     *
+     * Currently, classifications exist as plain strings under the "abcClass" column of the Product and InventoryLevel
+     * tables. Until we move classifications to their own entity, we simply fetch all unique values from those columns.
+     */
+//    @Transactional(readOnly=true)
```
alannadolny, line 35: is this comment needed?

alannadolny (approved):

grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
EWaterman: I'm trying to get the tests to run with that enabled. Unfortunately the unit tests don't play nice with it

EWaterman: I'm just going to remove the line for now to get this merged. I suspect I'll have to convert these tests to integration tests because they're working with the db. I'll do that in a separate PR.

EWaterman:
I'm merging this since most of the new changes are to get integration tests working or addressing existing comments but feel free to leave further comments if you want.

PR #4997, opened 2025-01-15
OBPIH-6889 Implement get cycle counts API

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -932,9 +932,9 @@ class UrlMappings {
             action = [GET: "read", POST: "update", PUT: "update", DELETE: "delete"]
         }
 
-        "/api/facilities/$facilityId/cycle-count-candidates" {
+        "/api/facilities/$facilityId/cycle-counts" {
             controller = "cycleCountApi"
-            action = [GET: "getCandidates"]
+            action = [GET: "getCycleCounts"]
         }
 
         "/api/facilities/$facilityId/cycle-counts/requests/batch" {
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -932,9 +932,9 @@ class UrlMappings {
             action = [GET: "read", POST: "update", PUT: "update", DELETE: "delete"]
         }
 
-        "/api/facilities/$facilityId/cycle-counts" {
+        "/api/facilities/$facilityId/cycle-counts/candidates" {
             controller = "cycleCountApi"
-            action = [GET: "getCycleCounts"]
+            action = [GET: "getCandidates"]
         }
 
         "/api/facilities/$facilityId/cycle-counts/requests/batch" {
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -17,7 +17,7 @@ class CycleCountApiController {
 
     CycleCountService cycleCountService
 
-    def getCycleCounts(CycleCountCandidateFilterCommand filterParams) {
+    def getCandidates(CycleCountCandidateFilterCommand filterParams) {
         List<CycleCountCandidate> cycleCounts = cycleCountService.getCandidates(filterParams, params.facilityId)
 
         if (filterParams.format == "csv") {
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -5,6 +5,7 @@ import grails.validation.ValidationException
 import org.apache.commons.csv.CSVPrinter
 import org.pih.warehouse.inventory.CycleCountCandidate
 import org.pih.warehouse.inventory.CycleCountCandidateFilterCommand
+
 import org.pih.warehouse.inventory.CycleCountRequest
 import org.pih.warehouse.inventory.CycleCountRequestBatchCommand
 import org.pih.warehouse.inventory.CycleCountRequestCommand
@@ -16,17 +17,16 @@ class CycleCountApiController {
 
     CycleCountService cycleCountService
 
-    def getCandidates(CycleCountCandidateFilterCommand filterParams) {
-        List<CycleCountCandidate> candidates = cycleCountService.getCandidates(filterParams, params.facilityId)
+    def getCycleCounts(CycleCountCandidateFilterCommand filterParams) {
+        List<CycleCountCandidate> cycleCounts = cycleCountService.getCandidates(filterParams, params.facilityId)
 
         if (filterParams.format == "csv") {
-            CSVPrinter csv = cycleCountService.getCycleCountCsv(candidates)
+            CSVPrinter csv = cycleCountService.getCycleCountCsv(cycleCounts)
             response.setHeader("Content-disposition", "attachment; filename=\"CycleCountReport-${new Date().format("yyyyMMdd-hhmmss")}.csv\"")
             render(contentType: "text/csv", text: csv.out.toString())
             return
         }
-
-        render([data: candidates, totalCount: candidates.totalCount] as JSON)
+        render([data: cycleCounts, totalCount: cycleCounts.totalCount] as JSON)
     }
 
     def createRequests(CycleCountRequestBatchCommand command) {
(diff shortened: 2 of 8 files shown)


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -932,9 +932,9 @@ class UrlMappings {
             action = [GET: "read", POST: "update", PUT: "update", DELETE: "delete"]
         }
 
-        "/api/facilities/$facilityId/cycle-count-candidates" {
+        "/api/facilities/$facilityId/cycle-counts" {
```
kchelstowski, line 935: as explained quickly on the tech huddle - I didn't know whether to keep the `cycle-count-candidates` and create `cycle-counts` even though they would point to the same endpoint, or to have a common one. 
I'm up for suggestions.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -16,17 +17,16 @@ class CycleCountApiController {
 
     CycleCountService cycleCountService
 
-    def getCandidates(CycleCountCandidateFilterCommand filterParams) {
-        List<CycleCountCandidate> candidates = cycleCountService.getCandidates(filterParams, params.facilityId)
+    def getCycleCounts(CycleCountCandidateFilterCommand filterParams) {
```
kchelstowski, line 20: I also don't feel comfortable with both `getCandidates` and `getCycleCounts` - we base those on `CycleCountCandidate`, but we might use the endpoint for different usages - either for get candidates (`cycleCountRequest = null`) or to get cycle counts (to count tab, to resolve tab, to approve tab).

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -26,7 +27,9 @@ CREATE OR REPLACE VIEW cycle_count_session AS (
 	JOIN location on product_availability.location_id = location.id
 	JOIN product on product_availability.product_id = product.id
 	JOIN inventory on location.inventory_id = inventory.id
+    LEFT OUTER JOIN cycle_count_request ON product.id = cycle_count_request.product_id AND location.id = cycle_count_request.facility_id
     WHERE product_availability.quantity_on_hand > 0
-	GROUP BY location.id, product.id, abc_class
+       AND (cycle_count_request.id IS NULL OR (cycle_count_request.status <> 'COMPLETED' AND cycle_count_request.status <> 'CANCELED'))
```
kchelstowski, line 32: basically we have to join the cycle count request table, but since we won't delete the completed and canceled requests, we need to filter them out in order to have unique result per product/facility pair.
We have a constraint for product/facility pair to be unique, but only if a request is not completed or canceled.
Since the candidates don't need to look into completed and canceled requests, I filtered them out.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -64,6 +64,12 @@ class CycleCountService {
             if (command.sort) {
                 getCandidatesSortOrder(command.sort, command.order, delegate, usedAliases)
             }
+            if (command.status) {
+                createAlias("cycleCountRequest", "cycleCountRequest", JoinType.INNER_JOIN)
+                eq("cycleCountRequest.status", command.status)
+                return
+            }
+            isNull("cycleCountRequest")
```
kchelstowski, line 72: e.g. 
1. for "to count" tab we will filter by `CREATED` status
2. for all products tab we don't provide status, so we want to search for candidates that do not have the cycle count request created

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
EWaterman, line 935: I think this API should still be /cycle-count-candidates (though I'd make it /cycle-count/candidates) since ultimately it is candidates focused. (It's returning List<CycleCountCandidate> so fundamentally it's about candidates.)
 
I also wanted to use the GET /cycle-counts API for another purpose: https://pihemr.atlassian.net/wiki/spaces/OB/pages/3332505612/Cycle+Count+Items+API#Get-Cycle-Counts

`GET /api/facilities/{facility_id}/cycle-counts?ids=1,2,3`
```
[
  {
    request: { ... },
    cycleCount: {
      id: "1",
      ...other fields...
      cycleCountItems: [ ... ],
    },
  },
  ...
]
```
So it's an API for simply fetching cycle counts, their items, and associated requests, which we (might) need when loading the count page for an already in progress count. 

I like being able to have the CRUD style endpoints for cycle count using "cycle-counts" as a base resource. That way the /candidates API can stay as its own thing.

- GET /cycle-counts
- GET /cycle-counts/{cycle_count_id}
- GET /cycle-counts/candidates

EWaterman: (sorry I wrote this before I saw your messages. I didn't refresh the page.)

EWaterman: yeah it is a bit weird. My other comment suggested using just the one GET /cycle-counts/candidates API for all the tabs.

I get that it's a bit strange since for the other tabs like TO COUNT we could look at it like the primary resource we're fetching is cycle count (and we're joining candidate information to that) but since all tabs need essentially the same information, I think it's fine that they all use the GET candidates API

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
EWaterman: yeah this is where the most amount of weirdness happens. However since the response data is structured the same for both pages, I think it's fine to treat this as the original getCandidates (or maybe we can think of a new name that makes the most sense). We should remember that the backend has no knowledge of what the frontend is doing, so the API (and method names) should ideally be reflecting only what the data has (and not what it is being used for). Some other frontend could use the API in a different way.

I'll try to think of a method/API name that makes sense.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: We also need to filter out requests that are currently in progress. Those shouldn't show up on the all products tab.

grails-app/migrations/views/cycle-count-session.sql
EWaterman: Does the "cycle_count_request.id IS NULL" work if there isn't a cycle count request? I'd assume it'd fail if cycle_count_request is null.

kchelstowski: hmm, why would it fail in your opinion? 
We LEFT JOIN the `cycle_count_request` table, so essentially we take all the rows from the `cycle_count_session` table, and join the `cycle_count_request` even if it doesn't exist (so then all the rows from `cycle_count_request` would be `null`).
So this check basically does this: return all the records from the `cycle_count_session` where the cycle count request doesn't exist or its status is not neither COMPLETED nor CANCELED.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: on all products tab we don't provide the status property, hence we would return all candidates that DO NOT have the cycle count request.

grails-app/migrations/views/cycle-count-session.sql
EWaterman: I was thinking too Java-based. You're right about the left join

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: and they won't show, because the cycle count request would already exist for such candidate, so those would be filtered out.

EWaterman: is that true? Maybe I'm misunderstanding, but for all products tab we don't provide a status and so we do `isNull("cycleCountRequest")` and fetch only the rows that don't have a request.

I don't want to go in circles about this so if you tested it and it's working, I trust you and we can move on :laughing:

EWaterman: We discussed on slack. I'm being silly. This code works.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: Good point not to think about it in the frontend way. I always end up making that mistake.
I changed the url to `/cycle-counts/candidates` as suggested and kept the methods with the `candidates` sufix.

EWaterman (approved):

alannadolny (approved):

kchelstowski:
@awalkowiak to unblock the rest of the PRs and the testing, I'm merging this one as it is. If you have any suggestions/change requests, let me know and I will address them separately.

PR #4998, opened 2025-01-16
OBPIH 6924 Create a wrapper component for the inbound workflow

SebastianLib:

**Description:**

---


grails-app/i18n/messages.properties
```diff
@@ -4502,7 +4505,7 @@ react.productSupplier.form.bidName.tooltip=The bid during which the purchasing p
 react.productSupplier.form.productPackageQuantity.tooltip=The number of units per package
 react.productSupplier.form.minOrderQuantity.tooltip=Minimum Order Quantity - the smallest order the vendor will accept for this product
 react.productSupplier.form.packagePrice.tooltip=The most recent price paid per default package
-react.productSupplier.form.eachPrice.tooltip=The most recent price paid per smallest individual unit (package price�package size)
+react.productSupplier.form.eachPrice.tooltip=The most recent price paid per smallest individual unit (package price�package size)
```
alannadolny, line 4508: What happened here?

grails-app/i18n/messages.properties
```diff
@@ -3659,6 +3659,9 @@ react.stockMovement.enum.AvailableItemStatus.RECALLED=Recalled
 react.stockMovement.enum.AvailableItemStatus.HOLD=Hold
 react.stockMovement.enum.AvailableItemStatus.NOT_AVAILABLE=Not Available
 
+react.stockMovement.form.previous.label=Previous
+react.stockMovement.form.next.label=Next
```
alannadolny, line 3663: For the whole react app you should be able to use:
`react.default.button.next.label`
and
`react.default.button.previous.label`

src/js/components/Router.jsx
```diff
@@ -32,6 +32,11 @@ const AsyncStockMovementInbound = Loadable({
   loading: Loading,
 });
 
+const AsyncStockMovementInbound2 = Loadable({
+  loader: () => import('components/stock-movement-wizard/inboundV2/InboundV2'),
```
alannadolny, line 36: I don't know whether we should call it V2. If we can quickly change the naming after removing the old inbound I am fine with it. 
What do you think @awalkowiak?

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import useTranslate from 'hooks/useTranslate';
+import HeaderWrapper from 'wrappers/HeaderWrapper';
+
+import '../StockMovement.scss';
```
alannadolny, line 8: Shouldn't be relative path

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <HeaderWrapper className="align-items-center h-100 py-3">
+      <div className="create-page-title d-flex align-items-center justify-content-between w-100">
+        <h5 className="create-page-tile-main-content m-0 ">
+          <strong>{translate('react.stockMovement.label', 'Stock Movement')}</strong>
+          {title && (
+            <>
+              <span>{' | '}</span>
+              {title.map((item, index) => (
+                // eslint-disable-next-line react/no-array-index-key
```
alannadolny, line 22: Please, remove this comment

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import useTranslate from 'hooks/useTranslate';
+import HeaderWrapper from 'wrappers/HeaderWrapper';
+
+import '../StockMovement.scss';
+
+const InboundHeader = ({ title, additionalTitle }) => {
```
alannadolny, line 10: I would rather call it `subtitle` instead of `additional title`

alannadolny: I don't see where the `title` is used, do you forget to use it or am I missing something :smile:

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+                  <span style={{ color: item.color }}>{item.text}</span>
+                  {item.delimeter && <span>{item.delimeter}</span>}
+                </React.Fragment>
+              ))}
+            </>
+          )}
+        </h5>
+        {additionalTitle && (
+          <span
+            className={`${additionalTitle.className}`}
```
alannadolny, line 33: do we need a string interpolation here?

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
```diff
@@ -0,0 +1,97 @@
...
+      key: inboundV2Step.ADD_ITEMS,
+      title: translate('react.stockMovement.addItems.label', 'Add Items'),
+      Component: () => (<InboundV2AddItems />),
+    },
+    {
+      key: inboundV2Step.SEND,
+      title: translate('react.stockMovement.send.label', 'Send'),
+      Component: () => (<InboundV2Send />),
+    },
+  ], [translate]);
```
alannadolny, line 38: translate is a function, I think we should use simple data as a dependency (in this case you can use language)

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const stepsTitles = steps.map((step) => ({
+    title: step.title,
+    key: step.key,
+  }));
+
+  const [
+    Step,
+    {
+      // navigateToStep,
```
alannadolny, line 48: remove the comment if not needed

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
```diff
@@ -0,0 +1,97 @@
...
+  || is(inboundV2Step.SEND) ? mockInboundV2Title : undefined;
+
+  return (
+    <PageWrapper>
+      <WizardStepsV2 steps={stepsTitles} currentStepKey={Step.key} />
+      <InboundHeader
+        title={headerTitle}
+        additionalTitle={headerAdditionalTitle}
+      />
+      {/* <form> */}
```
alannadolny, line 70: is this needed in the future?

src/js/components/stock-movement-wizard/inboundV2/sections/InboundV2AddItems.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import Section from 'components/Layout/v2/Section';
+
+const InboundV2AddItems = () => {
+  console.log('InboundV2AddItems');
```
alannadolny, line 6: :point_up:

grails-app/i18n/messages.properties
SebastianLib: I was wondering about that because in the form I was inspired by, the `OutboundImport ` button has `react.outboundImport.form.next.label`:     
```
 <Button
        label="react.outboundImport.form.next.label"
        defaultLabel="Next"
        variant="primary"
        type="submit"
        className="fit-content align-self-end"
        disabled={!isValid}
        customRef={nextButtonRef}
      />
```

SebastianLib: but I will change it

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
SebastianLib: `title` is mapped on line 21. I call it addtitionalTitle because that's what it was called in the old inbound version, but I can change it

SebastianLib: `title` is mapped on line 21. I call it addtitionalTitle because that's what it was called in the old inbound version, but I can change it

src/js/components/Router.jsx
kchelstowski: It was my suggestion so you can blame me for that. The idea was to just have both of the versions for now, so that Sebastian can quickly look into the old workflow, but it is supposed to be changed at the end of the development, so when the old workflow will be removed.
This is also why I decided to create a feature branch for that.

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -0,0 +1,62 @@
...
+    <HeaderWrapper className="align-items-center h-100 py-3">
+      <div className="create-page-title d-flex align-items-center justify-content-between w-100">
+        <h5 className="create-page-tile-main-content m-0 ">
+          <strong>{translate('react.stockMovement.label', 'Stock Movement')}</strong>
+          {title && (
+            <>
+              <span>{' | '}</span>
+              {title.map((item, index) => (
+                // eslint-disable-next-line react/no-array-index-key
+                <React.Fragment key={index}>
```
kchelstowski, line 23: as explained in the other PR, indexes should not be keys, hence the eslint above should not be disabled.
Moreover, we can use fragments like `<> </>`.

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
```diff
@@ -0,0 +1,97 @@
...
+      {is(inboundV2Step.SEND) && (<Step.Component />)}
+
+      {/* these buttons are for testing purposes now */}
+      <div className="d-flex justify-content-between">
+        <Button
+          label="react.stockMovement.form.previous.label"
+          defaultLabel="Previous"
+          variant="primary"
+          className="fit-content align-self-end"
+          onClick={previous}
```
kchelstowski, line 82: this doesn't change anything here, but be aware that by calling it like that, if the `previous` was expecting to receive any argument, this would be run with the `event` argument.
What you did is equivalent to do:
```js
onClick={e => previous(e)}
```
so when you want to make sure you call a method without any params here, you should do:
```js
onClick={() => previous()}
```
but as I said, this doesn't change anything for this particular method. Treat it as mentoring comment.

kchelstowski (commented):
generally looks good, good job :clap:  I will do another round of review after you address Alan's suggestions

src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
SebastianLib: from what I remember it's as Alan says. I'll check it later because I'm finishing another ticket now

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
```diff
@@ -78,6 +78,10 @@ class StockMovementController {
         render(view: "/common/react", params: params)
     }
 
+    def createInboundV2() {
```
awalkowiak, line 81: Is this one required? Can we just use the regular `createInbound` action?

alannadolny: I agree with you, because this is a feature branch I don't see anything against replacing the old one.

kchelstowski: if we were to replace it, Sebastian would not be able to access the old workflow to compare those while working. We decided it would be easier for him to do it this way, and this will be removed at the end of the development (probably when the PR from feature branch to develop is created)

alannadolny (approved):

kchelstowski (approved):

grails-app/controllers/org/pih/warehouse/inventory/StockMovementController.groovy
awalkowiak: @kchelstowski That is not what I meant here. I rather wanted to move it to the urlmapping, to handle that part and point to proper (the old one) action, to avoid adding this `V2` in the controller.

kchelstowski: @awalkowiak ok now I get it, I don't know why I didn't realize we could actually use two URLs and map to the same action, and let it be handled further by the react router.

PR #4999, opened 2025-01-16
OBPIH-6846 Build checkbox logic

alannadolny:

**Description:** I modified the checkbox component to be able to use an indeterminate state. I also added a hook for managing checkboxes in the table.

---

Screenshot from 2025-01-16 15-27-10

### DIFF
--- a/src/js/hooks/cycleCount/useAllProductsTab.jsx
+++ b/src/js/hooks/cycleCount/useAllProductsTab.jsx
@@ -7,8 +7,10 @@ import { useSelector } from 'react-redux';
 import { CYCLE_COUNT_CANDIDATES } from 'api/urls';
 import { TableCell } from 'components/DataTable';
 import TableHeaderCell from 'components/DataTable/TableHeaderCell';
+import Checkbox from 'components/form-elements/v2/Checkbox';
 import { INVENTORY_ITEM_URL } from 'consts/applicationUrls';
 import useSpinner from 'hooks/useSpinner';
+import useTableCheckboxes from 'hooks/useTableCheckboxes';
 import useTableDataV2 from 'hooks/useTableDataV2';
 import useTableSorting from 'hooks/useTableSorting';
 import useTranslate from 'hooks/useTranslate';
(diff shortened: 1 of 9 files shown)


src/js/hooks/cycleCount/useAllProductsTab.jsx
```diff
@@ -87,6 +89,44 @@ const useAllProductsTab = ({ filterParams }) => {
...
+    selectHeaderCheckbox,
+    selectedRowsAmount,
+    headerCheckboxProps,
+  } = useTableCheckboxes();
+
+  const getProductIds = () => tableData.data.map((row) => row.product.id);
+
+  // Separated from columns to reduce the amount of rerenders of
+  // the rest columns (on checked checkboxes change)
+  const checkboxesColumn = columnHelper.accessor('selected', {
```
kchelstowski, line 104: do you think there would be any benefit from including the `useMemo` here not to re-render it every single time, since a lot of compoments are built here?

src/js/hooks/cycleCount/useAllProductsTab.jsx
```diff
@@ -87,6 +89,44 @@ const useAllProductsTab = ({ filterParams }) => {
...
 
+  const {
+    selectRow,
+    isChecked,
+    selectHeaderCheckbox,
+    selectedRowsAmount,
+    headerCheckboxProps,
+  } = useTableCheckboxes();
+
+  const getProductIds = () => tableData.data.map((row) => row.product.id);
```
kchelstowski, line 100: I guess this doesn't even have to be a method, but could be a const, like:
```js
const productIds = useMemo(() => tableData.data.map(...), [<any dependency indicating the tableData changed>])
```
I don't know though how the re-rendering of methods is executed for such cases, whether the mapping would be proceeded in the background while re-rendering or not.

src/js/hooks/useTableCheckboxes.js
```diff
@@ -0,0 +1,62 @@
...
+      setCheckedCheckboxes(
+        (checked) => checked.filter((checkedProductId) => identifier !== checkedProductId),
+      );
+      return;
+    }
+    setCheckedCheckboxes((checked) => ([...checked, identifier]));
+  }, [checkedCheckboxes]);
+
+  useEffect(() => {
+    const selectedRows = checkedCheckboxes.length;
```
kchelstowski, line 21: I think you mix two namings here - `selectedRows` vs `checkedCheckboxes`. Maybe we could align on one of them?
Moreover - for me it isn't a problem, but `selectedRows` sound like we store the whole rows there, not only the length, so how about adding any `amountOf` prefix?

src/js/hooks/cycleCount/useAllProductsTab.jsx
alannadolny: There are benefits, because I can't use `useMemo` on all columns, because there are problems with refreshing the checkboxes (even with correct dependencies)

kchelstowski (approved):

PR #5000, opened 2025-01-16
OBPIH-6922 Shorten the Name of Bin Location

SebastianLib:

**Description:**
I shortened the bin location name and put the number of bin locations involved in the tooltip

---

### DIFF
--- a/src/js/hooks/cycleCount/useAllProductsTab.jsx
+++ b/src/js/hooks/cycleCount/useAllProductsTab.jsx
@@ -135,15 +135,23 @@ const useAllProductsTab = ({ filterParams }) => {
           {translate('react.cycleCount.table.binLocation.label', 'Bin Location')}
         </TableHeaderCell>
       ),
-      cell: ({ getValue }) => (
-        <TableCell
-          className="rt-td"
-          tooltip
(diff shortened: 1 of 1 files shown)


src/js/hooks/cycleCount/useAllProductsTab.jsx
```diff
@@ -135,15 +135,23 @@ const useAllProductsTab = ({ filterParams }) => {
...
+            tooltip
+            tooltipLabel={`${getValue()} (${binLocationList.length})`}
+          >
+            {binLocationList.map((binLocationName) => (
+              <div>
+                {binLocationName.length > 16 ? `${binLocationName.substring(0, 16)}...` : binLocationName}
+              </div>
+            ))}
+          </TableCell>
+        );
```
alannadolny, line 153: I think we should do this by CSS, not just limiting the length, because when you make the screen even smaller (I know we are supporting certain screen sizes, but a lot of users have smaller screens than those supported) the bin location won't be visible. You can achieve that by packing every bin location to the div and then just hiding the overflow.

alannadolny (approved):
good job :clap:

kchelstowski (approved):

PR #5004, opened 2025-01-20
OBPIH-6929 Implement 'Show Negative Quantity Only' Filter

SebastianLib:

**Description:**
I don't know if it's done correctly

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -61,6 +61,9 @@ class CycleCountService {
                 usedAliases.add("tags")
                 "in"("tags.id", command.tags.collect { it.id })
             }
+            if (command.negativeQuantity) {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -61,6 +61,9 @@ class CycleCountService {
                 usedAliases.add("tags")
                 "in"("tags.id", command.tags.collect { it.id })
             }
+            if (command.negativeQuantity) {
+                lt("quantityOnHand", 0)
+            }
```
kchelstowski, line 66: this is not done in the way this should be done. Let's hold on with that one for a while to align on the requirements and if they are applicable for the v1.

TL:DR - the expectation is to filter out products for which exists at least one inventory item with negative quantity. There is `negativeItemCount` property prepared in the view, but it is hardcoded as `null` for now, and I assume this one would be used there, but this still needs to be implemented.

cc @EWaterman @jmiranda

PR #5007, opened 2025-01-20
OBPIH-6930 added page reload after changing location

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/components/location/LocationChooser/index.jsx
+++ b/src/js/components/location/LocationChooser/index.jsx
@@ -76,9 +76,8 @@ const LocationChooser = (props) => {
     props.changeCurrentLocation(location).then(() => {
       props.fetchMenuConfig();
       props.fetchSessionInfo();
-
-      window.location.reload();
     });
+    setIsOpen(false);
   };
 
   const toggleModalHandler = () => {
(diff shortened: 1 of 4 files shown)


src/js/components/location/LocationChooser/index.jsx
```diff
@@ -76,8 +76,9 @@ const LocationChooser = (props) => {
     props.changeCurrentLocation(location).then(() => {
       props.fetchMenuConfig();
       props.fetchSessionInfo();
+
+      window.location.reload();
     });
-    setIsOpen(false);
```
kchelstowski, line 80: this is not the solution we would want to have. This way, you would always reload the page, it's not what React's flow is about (SPA). 
What you need to do is to refetch the data when the location changes (using appropriate `useEffect`)

kchelstowski (changes requested):

src/js/components/location/LocationChooser/index.jsx
alannadolny: :point_up:

src/js/hooks/useTableDataV2.js
```diff
@@ -45,6 +45,7 @@ const useTableDataV2 = ({
   }));
 
   const fetchData = () => {
+    sourceRef.current = CancelToken.source();
```
kchelstowski, line 48: I think this looks good because this is what we also did for the old `useTableData`. The only request I have is to copy the comment from the `useTableData`, so that everyone understands why it is needed here.

alannadolny (approved):

kchelstowski (approved):

awalkowiak (approved):

PR #5015, opened 2025-01-23
OBPIH-6937 Create details form

SebastianLib:

**Description:**

---

image


src/js/hooks/inboundV2/useInboundV2Validation.js
```diff
@@ -0,0 +1,70 @@
+import { z } from 'zod';
+
+import useTranslate from 'hooks/useTranslate';
+import { validateFutureDate } from 'utils/form-utils';
+
+const useOutboundImportValidation = () => {
```
awalkowiak, line 6: This should be changed

grails-app/i18n/messages.properties
```diff
@@ -4642,3 +4642,15 @@ react.cycleCount.filter.productCatalogs.label=Product Catalogue
 react.cycleCount.filter.abcClass.label=ABC Class
 react.cycleCount.filter.negativeInventory.label=Show negative quantity only
 react.cycleCount.filter.search.label=Search...
+
+
+# InboundV2
+react.inboundV2.validation.requestedBy.required.label=Requested by is required
```
awalkowiak, line 4648: Can we just use existing inbound labels? IIRC theoretically we shouldn't have to change anything on the backend for this refactor, or the changes could be just minimal. Please refrain from using `V2` on a backend at all for now.

src/js/hooks/inboundV2/useInboundV2Form.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      setStockLists(newStockLists);
+    } finally {
+      spinner.hide();
+    }
+  };
+
+  const origin = watch('origin');
+  const destination = watch('destination');
```
kchelstowski, line 119: what's the difference between `watch` and `useWatch`? I can see you using `watch` here separately for those, and `useWatch` in one of the files above

src/js/hooks/inboundV2/useInboundV2Form.js
```diff
@@ -0,0 +1,140 @@
...
+
+  useEffect(() => {
+    if (origin?.id && destination?.id) {
+      fetchStockLists();
+    }
+  }, [origin, destination]);
```
kchelstowski, line 125: add `.id` to the dependency array

src/js/hooks/useWizard.js
```diff
@@ -59,11 +61,11 @@ const useWizard = ({ initialKey, steps }) => {
     navigateToStep(steps[lastIdx]?.key);
   };
 
-  const next = () => {
+  const next = (params = {}) => {
```
kchelstowski, line 64: add a comment above explaining why we could need the params (e.g. "params might be needed to join some query params to the URL while switching the step")

src/js/hooks/inboundV2/useInboundV2Validation.js
```diff
@@ -0,0 +1,70 @@
...
+        message: translate('react.inboundV2.validation.dateRequested.futureDate.label', 'The date cannot be in the future'),
+      }),
+  });
+
+  return {
+    validationSchema,
+    destinationSchema,
+    originSchema,
+    requestedBySchema,
+    stocklistSchema,
```
kchelstowski, line 66: return only `validationSchema` from this hook. This is the only object you actually use while executing this hook:
```js
  const { validationSchema } = useInboundV2Validation();
```

kchelstowski (changes requested):
nice job, those are rather minor things. It would be also good if you attached a screenshot of the UI.
When this is not a draft anymore, I will do another round and will test it locally, and compare it to the existing workflow (both from the UI side and the code side)

src/js/hooks/inboundV2/useInboundV2Form.js
SebastianLib: `watch` will trigger re-renders at the root level. `useWatch` will trigger re-renders at the hook or component level. I thought it would be a better approach to handle it this way

SebastianLib:
Right now, that looks like the first step of the UI

image

src/js/hooks/inboundV2/useInboundV2Validation.js
```diff
@@ -0,0 +1,66 @@
+import { z } from 'zod';
+
+import useTranslate from 'hooks/useTranslate';
+import { validateFutureDate } from 'utils/form-utils';
+
+const useInboundV2Validation = () => {
```
kchelstowski, line 6: let's rename this one not to include the `V2` anymore

kchelstowski (approved):

src/js/api/urls.js
```diff
@@ -17,6 +17,7 @@ export const STOCK_MOVEMENT_BY_ID = (id) => `${STOCK_MOVEMENT_API}/${id}`;
 export const STOCK_MOVEMENT_PENDING_SHIPMENT_ITEMS = `${STOCK_MOVEMENT_API}/pendingRequisitionItems`;
 export const STOCK_MOVEMENT_INCOMING_ITEMS = `${STOCK_MOVEMENT_API}/shippedItems`;
 export const STOCK_MOVEMENT_UPDATE_STATUS = (id) => `${STOCK_MOVEMENT_API}/${id}/status`;
+export const STOCK_MOVEMENT_UPDATE = (id) => `${STOCK_MOVEMENT_API}/${id}/updateRequisition`;
```
alannadolny, line 20: i would call it UPDATE_REQUISITION, we have also UPDATE_STATUS, UPDATE_ITEMS, etc. so I feel like something is missing here

src/js/hooks/inboundV2/useInboundForm.js
```diff
@@ -0,0 +1,143 @@
...
+      requestedBy: { id: values.requestedBy.id },
+    };
+    try {
+      let response;
+      if (queryParams.id) {
+        response = await stockMovementApi.updateStockMovement(queryParams.id,
+          formattedValues);
+      } else {
+        response = await stockMovementApi.createStockMovement(formattedValues);
+      }
```
alannadolny, line 72: can you shorten this to the ternary operator?

PR #5016, opened 2025-01-24
OBPIH-6933 User can manually enter a page number outside the valid page range

alannadolny:


src/js/hooks/useTablePagination.js
```diff
@@ -11,6 +11,21 @@ const useTablePagination = ({
   });
 
   const onPageChange = (page) => {
+    const maxPage = Math.floor(totalCount / pagination.pageSize);
```
awalkowiak, line 14: Does `maxPage` have to be recalculated on each page change?

alannadolny: do you suggest using `useCallback` with totalCount and pageSize as a dependency?

src/js/hooks/useTablePagination.js
```diff
@@ -11,6 +11,21 @@ const useTablePagination = ({
...
+    if (maxPage < page) {
```
kchelstowski, line 15: nitpicky, but since the comparators below start with the `page`, for me personally it would be more readable if it was:
```
if (page > maxPage)
```
but it's my own preference so I'm not requesting a change

src/js/hooks/useTablePagination.js
kchelstowski: yeah, we can use it, this can be cached until filters are applied (`totalCount` changes/page size changes)

awalkowiak (approved):

PR #5018, opened 2025-01-28
OBPIH-6893 Create the batch start count API

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
+++ b/grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
@@ -942,6 +942,11 @@ class UrlMappings {
             action = [POST: "createRequests"]
         }
 
+        "/api/facilities/$facility/cycle-counts/start/batch" {
+            controller = "cycleCountApi"
+            action = [POST: "startCycleCount"]
+        }
+
         // Error handling
 
         "401"(controller: "errors", action: "handleUnauthorized")
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -74,7 +74,7 @@ class CycleCountApiController {
                     errors.addError(error)
                 }
             }
-            throw new ValidationException("Invalid cycle count request", errors)
+            throw new ValidationException("Invalid cycle count", errors)
         }
         List<CycleCountDto> cycleCounts = cycleCountService.startCycleCount(command)
 
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -3,13 +3,16 @@ package org.pih.warehouse.api
 import grails.converters.JSON
 import grails.validation.ValidationException
 import org.apache.commons.csv.CSVPrinter
+import org.pih.warehouse.core.Location
 import org.pih.warehouse.inventory.CycleCountCandidate
 import org.pih.warehouse.inventory.CycleCountCandidateFilterCommand
-
+import org.pih.warehouse.inventory.CycleCountDto
 import org.pih.warehouse.inventory.CycleCountRequest
 import org.pih.warehouse.inventory.CycleCountRequestBatchCommand
 import org.pih.warehouse.inventory.CycleCountRequestCommand
 import org.pih.warehouse.inventory.CycleCountService
+import org.pih.warehouse.inventory.CycleCountStartBatchCommand
+import org.pih.warehouse.inventory.CycleCountStartCommand
 import org.springframework.validation.BeanPropertyBindingResult
 import org.springframework.validation.Errors
 
@@ -51,5 +54,30 @@ class CycleCountApiController {
         }
         List<CycleCountRequest> cycleCountRequests = cycleCountService.createRequests(command)
         render([data: cycleCountRequests] as JSON)
+    }
+
+    def startCycleCount(CycleCountStartBatchCommand command) {
+        if (command.hasErrors()) {
+            // Build errors manually to be able to include errors both for the batch command instance
+            // and errors for every element of the requests list
+            Errors errors = new BeanPropertyBindingResult(command, "requests")
+            // Iterate every element of requests and add its error to the errors instance.
+            command.requests.each { CycleCountStartCommand request ->
+                request.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            // If there are not any errors in the elements, it means that there must be errors in the batch command instance
+            // e.g. missing required arguments
+            if (!errors.hasErrors()) {
+                command.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            throw new ValidationException("Invalid cycle count request", errors)
+        }
+        List<CycleCountDto> cycleCounts = cycleCountService.startCycleCount(command)
+
+        render([data: cycleCounts] as JSON)
     }
 }
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
@@ -47,17 +47,17 @@ class CycleCount {
         if (items.every { it.status == CycleCountItemStatus.REVIEWED }) {
             return CycleCountStatus.REVIEWED
         }
-        if (items.every { it.status == CycleCountItemStatus.TO_REVIEW }) {
-            return CycleCountStatus.TO_REVIEW
+        if (items.every { it.status == CycleCountItemStatus.READY_TO_REVIEW }) {
+            return CycleCountStatus.READY_TO_REVIEW
         }
         if (items.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
             return CycleCountStatus.INVESTIGATING
         }
         if (items.any { it.status == CycleCountItemStatus.COUNTING }
-                && !items.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.TO_REVIEW] }) {
+                && !items.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.READY_TO_REVIEW] }) {
             return CycleCountStatus.COUNTING
         }
-        if (items.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.TO_REVIEW] }) {
+        if (items.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.READY_TO_REVIEW] }) {
             return CycleCountStatus.COUNTED
         }
         if (items.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
@@ -35,28 +35,32 @@ class CycleCount {
         updatedBy(nullable: true)
     }
 
+    List<CycleCountItem> getCycleCountItems() {
+        return CycleCountItem.findAllByCycleCount(this)
+    }
+
     CycleCountStatus getStatus() {
-        List<CycleCountItem> cycleCountItems = CycleCountItem.findAllByCycleCount(this)
-        if (cycleCountItems.every { it.status == CycleCountItemStatus.READY_TO_COUNT }) {
+        List<CycleCountItem> items = cycleCountItems
+        if (items.every { it.status == CycleCountItemStatus.READY_TO_COUNT }) {
             return CycleCountStatus.REQUESTED
         }
-        if (cycleCountItems.every { it.status == CycleCountItemStatus.REVIEWED }) {
+        if (items.every { it.status == CycleCountItemStatus.REVIEWED }) {
             return CycleCountStatus.REVIEWED
         }
-        if (cycleCountItems.every { it.status == CycleCountItemStatus.TO_REVIEW }) {
+        if (items.every { it.status == CycleCountItemStatus.TO_REVIEW }) {
             return CycleCountStatus.TO_REVIEW
         }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
+        if (items.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
             return CycleCountStatus.INVESTIGATING
         }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.COUNTING }
-                && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.TO_REVIEW] }) {
+        if (items.any { it.status == CycleCountItemStatus.COUNTING }
+                && !items.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.TO_REVIEW] }) {
             return CycleCountStatus.COUNTING
         }
-        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.TO_REVIEW] }) {
+        if (items.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.TO_REVIEW] }) {
             return CycleCountStatus.COUNTED
         }
-        if (cycleCountItems.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
+        if (items.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
             return CycleCountStatus.COMPLETED
         }
         return null
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
@@ -1,0 +1,64 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.auth.AuthService
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.User
+
+class CycleCount {
+
+    String id
+
+    Location facility
+
+    Date dateLastRefreshed
+
+    Date dateCreated
+
+    Date lastUpdated
+
+    User createdBy
+
+    User updatedBy
+
+    def beforeInsert() {
+        createdBy = AuthService.currentUser
+        updatedBy = AuthService.currentUser
+    }
+
+    def beforeUpdate() {
+        updatedBy = AuthService.currentUser
+    }
+
+    static constraints = {
+        id generator: "uuid"
+        createdBy(nullable: true)
+        updatedBy(nullable: true)
+    }
+
+    CycleCountStatus getStatus() {
+        List<CycleCountItem> cycleCountItems = CycleCountItem.findAllByCycleCount(this)
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.READY_TO_COUNT }) {
+            return CycleCountStatus.REQUESTED
+        }
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.REVIEWED }) {
+            return CycleCountStatus.REVIEWED
+        }
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.TO_REVIEW }) {
+            return CycleCountStatus.TO_REVIEW
+        }
+        if (cycleCountItems.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
+            return CycleCountStatus.INVESTIGATING
+        }
+        if (cycleCountItems.any { it.status == CycleCountItemStatus.COUNTING }
+                && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.TO_REVIEW] }) {
+            return CycleCountStatus.COUNTING
+        }
+        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.TO_REVIEW] }) {
+            return CycleCountStatus.COUNTED
+        }
+        if (cycleCountItems.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
+            return CycleCountStatus.COMPLETED
+        }
+        return null
+    }
+}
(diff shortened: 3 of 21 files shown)


kchelstowski:
A few comments from my side may come up to explain some decisions or to point out things I am not quite sure/would like to get a confirmation.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -942,6 +942,11 @@ class UrlMappings {
             action = [POST: "createRequests"]
         }
 
+        "/api/facilities/$facility/cycle-counts/start/batch" {
```
kchelstowski, line 945: @jmiranda I tried to follow your last suggestion that those params should be also bound to a command, but suddenly they aren't :thinking: they are only accessible via `params`.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -52,4 +55,29 @@ class CycleCountApiController {
...
+                }
+            }
+            // If there are not any errors in the elements, it means that there must be errors in the batch command instance
+            // e.g. missing required arguments
+            if (!errors.hasErrors()) {
+                command.errors.allErrors.each { error ->
+                    errors.addError(error)
+                }
+            }
+            throw new ValidationException("Invalid cycle count request", errors)
```
kchelstowski, line 77: (I see the typo)

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -0,0 +1,64 @@
...
+                && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.TO_REVIEW] }) {
+            return CycleCountStatus.COUNTING
+        }
+        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.TO_REVIEW] }) {
+            return CycleCountStatus.COUNTED
+        }
+        if (cycleCountItems.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
+            return CycleCountStatus.COMPLETED
+        }
+        return null
```
kchelstowski, line 62: I followed this:
Screenshot from 2025-01-28 10-08-02

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+                    quantityCounted: availableItem.quantityOnHand == 0 ? 0 : null,
+                    cycleCount: newCycleCount,
+                    facility: facility,
+                    inventoryItem: availableItem.inventoryItem,
+                    product: availableItem.inventoryItem.product,
+                    createdBy: AuthService.currentUser,
+                    updatedBy: AuthService.currentUser,
+                    // FIXME: I don't know what default values for custom and draft should be
+                    custom: false,
+                    draft: true,
```
kchelstowski, line 224: as mentioned in the FIXME - I am not sure what default values for those should be, as they weren't determined anywhere in the docs.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+            cycleCountItems.add(cycleCountItem)
+        }
+        return CycleCountDto.createFromCycleCountItems(cycleCountItems)
+    }
+
+    CycleCountDto updateCycleCount(CycleCount cycleCount, List<CycleCountItem> cycleCountItems, CycleCountStartCommand request) {
+        if (cycleCount.status in [CycleCountStatus.TO_REVIEW, CycleCountStatus.CANCELED]) {
+            throw new IllegalArgumentException("Cycle count is already in review or canceled")
+        }
+        if (cycleCount.status in [CycleCountStatus.COUNTING, CycleCountStatus.COUNTED]) {
```
kchelstowski, line 239: in the doc it was meant for "If the cycle count is “resolving” or “counted” - since we don't have `RESOLVING` status I assumed this is equivalent to the `COUNTING`? :thinking:

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+            throw new UnsupportedOperationException("Support will be added later")
+        }
+        if (cycleCount.status in [CycleCountStatus.COUNTING, CycleCountStatus.COUNTED, CycleCountStatus.INVESTIGATING, CycleCountStatus.TO_REVIEW]) {
```
kchelstowski, line 242: this was meant to work for cycle count "in progress", so I took this from:
Screenshot from 2025-01-28 10-11-05

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+            cycleCountItems.sort { it.countIndex }
+            if (cycleCountItems?.first()?.countIndex != request.countIndex) {
+                throw new IllegalArgumentException("Count index can't be higher than the highest count index of the items")
+            }
```
kchelstowski, line 246: > "If the given countIndex != max(count_index of the Items) , error"

(Will need to test it deeper though, as all my data contained only one item, hence I am yet sure if I should look for first or last item after sorting)

src/main/groovy/org/pih/warehouse/inventory/CycleCountDto.groovy
```diff
@@ -0,0 +1,20 @@
...
+
+class CycleCountDto {
+
+    String id
+
+    String status
+
+    List<CycleCountItemBasicDto> cycleCountItems
+
+    static CycleCountDto createFromCycleCountItems(List<CycleCountItem> items) {
```
kchelstowski, line 11: I gotta say it was quite tricky to work with the items without having the bidirectional relationship between Cycle count and Cycle count items, but I remember the reasons for promoting the unidirectional one-to-many relationship.

src/main/groovy/org/pih/warehouse/inventory/CycleCountDto.groovy
```diff
@@ -0,0 +1,20 @@
...
+        // Cycle count items have the same cycle count associated, so we can just look at first item in order to get the cycle count properties
+        CycleCountItem item = items?.first()
```
kchelstowski, line 13: this is one of the "tricky" parts I mentioned above. If I had an access to the list of the items from the cycle count domain, I wouldn't have to do that.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+
+    CycleCountDto createCycleCount(CycleCountStartCommand request, Location facility) {
+        CycleCount newCycleCount = new CycleCount(
+                facility: facility,
+                dateLastRefreshed: new Date()
+        )
+        List<AvailableItem> itemsToSave = determineCycleCountItemsToSave(facility, request.cycleCountRequest.product)
+        newCycleCount.save()
+        // 1:1 association between cycle count and cycle count request
+        request.cycleCountRequest.cycleCount = newCycleCount
```
kchelstowski, line 208: since the cycle count request's status is a persistent value, I'm wondering if we should change it right here, if so - should it be REQUESTED or IN_PROGRESS?

src/main/groovy/org/pih/warehouse/inventory/CycleCountDto.groovy
kchelstowski: eventually we can check if every item has the same cycle count (in case in the future this method would be used for "mixed" cycle count items (with different cycle count associated) and throw an exception.
Now, it can't happen, so let me know your thoughts.

src/main/groovy/org/pih/warehouse/inventory/CycleCountStartCommand.groovy
```diff
@@ -0,0 +1,20 @@
...
+
+    CycleCountRequest cycleCountRequest
+
+    Integer countIndex
+
+    static constraints = {
+        countIndex(min: 0, max: 1, validator: { Integer countIndex, CycleCountStartCommand obj ->
+            CycleCount cycleCount = obj.cycleCountRequest.cycleCount
+            if (!cycleCount && countIndex != 0) {
+                return ['mustEqualZero']
```
EWaterman, line 15: note that this will change when we introduce recount since starting a recount will use the same start count API. https://pihemr.atlassian.net/wiki/spaces/OB/pages/3363340321/TO+RESOLVE+page#Backend

src/main/groovy/org/pih/warehouse/inventory/CycleCountStartCommand.groovy
```diff
@@ -0,0 +1,20 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+
+class CycleCountStartCommand implements Validateable {
+
+    CycleCountRequest cycleCountRequest
```
EWaterman, line 7: does the client need to pass the whole request here or just request id and Grails binds the request object to it?

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -0,0 +1,64 @@
...
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.User
+
+class CycleCount {
+
+    String id
+
+    Location facility
+
+    Date dateLastRefreshed
```
EWaterman, line 13: now that I'm seeing this, I wonder if the dateLastRefreshed field would be better off in the CycleCountItem because that's where we're doing the actual QoH refresh. If we're in a recount and do a refresh, we only change QoH for the CycleCountItem with countIndex==1 (countIndex==0 stays unchanged). So it could be useful to have that date info at the item level to be able to clearly see why QoH changed between counts.

I don't know if you've had a chance to think about this yet. (For context, [this is the API that will use the field](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3332505612/Cycle+Count+Items+API#Refresh-Cycle-Count-Quantity-On-Hand))

src/main/groovy/org/pih/warehouse/inventory/CycleCountStartCommand.groovy
kchelstowski: we had that discussion recently regarding the `Product`.
Both options are possible, so to provide it like:
```
{
  cycleCountRequest: "UUID"
}
```
or 
```
{
  cycleCountRequest: {
    "id": "UUID"
  }
}
```

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -0,0 +1,65 @@
...
+                status: status,
+        )
+    }
+
+    static constraints = {
+        dateCounted(nullable: true)
+        comment(nullable: true)
+        assignee(nullable: true)
+        quantityCounted(nullable: true)
+        discrepancyCause(nullable: true)
```
EWaterman, line 63: Does Grails default to `nullable: false` for all fields? I always thought it was the other way around

src/main/groovy/org/pih/warehouse/inventory/DiscrepancyCause.groovy
```diff
@@ -0,0 +1,9 @@
+package org.pih.warehouse.inventory
+
+enum DiscrepancyCause {
```
EWaterman, line 3: I'm realizing that we never had a group discussion on what values should exist here (I wrote these ones myself). Lets discuss it with Manon.

We'll also need to localize each of these.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: we need to be able to distinguish between:
1) a cycle count exists and the current count has already been started (status in [COUNTING]) so just fetch it
2) a cycle count exists and the previous count was completed (status in [COUNTED]) so we're starting a recount

(I'm not totally sure where INVESTIGATING fits into those cases but it's in one of them. If INVESTIGATING means a recount is already started, then it's case 1. If it means a recount is required but hasn't started, then it's case 2.)

Case 2 can be left for TODO for now because it's recounting but we need to properly deal with 1.

There's two was to handle case 1 in this API:
A) Throw an error. You're not allowed to start a count that has already been started.
B) Simply return the cycle count and its items as they are because the count has already started / been initialized

If we pick A, then the frontend will need to do two things when starting a batch count:
- Call a GET list API to fetch all cycle counts in the batch that already started
- Call this POST start count API on all requests that haven't been started yet 

The design was done assuming solution B because it would be simpler for the frontend. This means this API needs to be able to gracefully handle both 1 and 2 cases.

Let me know if that makes sense. And if you feel like solution A makes more sense (which after seeing how complex this flow already is, I'm wishing was what I picked originally) we can talk about switching to that but unless you say otherwise I'll assume you want to stick with the original plan.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: I think it sorts low to high so it'd be last? I'm not sure. You could also just use `Collections.max()` if you don't want to bother sorting the whole list of items

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemBasicDto.groovy
```diff
@@ -0,0 +1,18 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.product.Product
+
+class CycleCountItemBasicDto {
```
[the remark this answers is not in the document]
kchelstowski: I followed the response pattern designed [here](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3332505612/Cycle+Count+Items+API#Batch-Start-Count). 
Let me know if this makes sense or we would want to return the whole cycle count item.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+                    createdBy: AuthService.currentUser,
+                    updatedBy: AuthService.currentUser,
+                    // FIXME: I don't know what default values for custom and draft should be
+                    custom: false,
+                    draft: true,
+            )
+            if (!cycleCountItem.validate()) {
+                throw new ValidationException("Invalid cycle count item", cycleCountItem.errors)
+            }
+            cycleCountItem.save()
```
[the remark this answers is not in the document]
kchelstowski: Yes, since it's being persisted in the same transaction, an error in any of the items would cause the rollback of whole operation

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+
+    /**
+     * Get product availability for a product and for facility.
+     * Filter out items with QOH = 0 (but keep those that were counted in the previous count)
+     * */
+    List<AvailableItem> determineCycleCountItemsToSave(Location facility, Product product) {
+        List<AvailableItem> availableItems =
+                productAvailabilityService.getAvailableItems(facility, [product.id])
+        CycleCount previousCycleCount =
+                CycleCountRequest.findAllByProductAndFacility(product, facility).cycleCount.sort { it.lastUpdated }.first()
```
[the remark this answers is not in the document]
kchelstowski: > Also, you're fetching all counts on the product then in the code throwing out all returned rows exept the most recent. Is there a way to move that into the query itself? Something like a SELECT * WHERE MAX(dateCompleted). That would save us from having to select more than we need so it'd be less work on the db with fewer rows locked

It's probably doable via the `createCriteria`, but for now I focused on deliver the easiest version possible, but I could try to improve the query

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -169,4 +173,94 @@ class CycleCountService {
...
+        List<CycleCountItem> previousCycleCountItems = CycleCountItem.findAllByCycleCount(previousCycleCount)
+        List<AvailableItem> filteredItems = availableItems.findAll { AvailableItem availableItem ->
+            (availableItem.quantityOnHand > 0) || (previousCycleCountItems.find { it.inventoryItem.id == availableItem.inventoryItem.id })}
```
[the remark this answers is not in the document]
kchelstowski: > filter out all bin locations + inventory items where QoH==0 AND (the previous cycle count did not include a count on that bin locations + inventory items OR the previous count also had QoH==0).

The quote above is taken from the tech design. Is it still applicable or the version you've just written is up to date?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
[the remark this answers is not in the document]
EWaterman: that's fine with me (though I encourage you to spend a bit of time trying to get it to work). If you're not going to fix it here though can you add a TODO comment for it? It could be a relatively important optimization because as we have more and more cycle counts on a product this will fetch more and more data, and so will only get worse over time.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: could you elaborate on that?

> a cycle count exists and the current count has already been started (status in [COUNTING]) so just fetch it

I kinda don't get what the current count means at this point. If we update a cycle count, it means that you've just fetched it, and it's our "existing cycle count" and "current count" at once.

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemBasicDto.groovy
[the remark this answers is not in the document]
EWaterman: hmm yeah I don't know if there was a good reason why I left out fields from the responses. Probably I just included the important ones. Sorry for not being clear about it. I'll leave it for you to decide but I think we'd be fine with just one CycleCountItemDto for all our responses here.

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
kchelstowski: yes, `nullable: false` is the default value.
source: https://docs.grails.org/latest/ref/Constraints/nullable.html

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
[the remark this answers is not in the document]
EWaterman: yes it's still applicable. In the second half of your OR, you're including items that were in a previous count, but you only want to include them if quantity counted != 0.

In my comment above I thought it'd be good to further refine this to also include them if quantity counted == 0 but QoH at the time was != 0. The idea being that even though there was zero last time, there was a discrepancy so we should double check out.

Lets talk about it with Manon in the tech huddle though to confirm the exact expected behaviour. Maybe since we're already auto-filling a 0 for empty bins it's fine to not filter them out.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: - You're on the TO COUNT tab and start a count on request_id == 1.
- Frontend calls the start count API and the backend creates and returns cycle_count_id == 1.
- You log out and come back to the TO COUNT tab and click start count again on the same request (1)
- Frontend calls the start count API (again). The fetches the count associated with that request, finds it, sees that it's active, and so just returns it

an alternative flow (that I wish I designed instead) would be:
- You're on the TO COUNT tab and start a count on request_id == 1.
- Frontend calls the start count API and the backend creates and returns cycle_count_id == 1.
- You log out and come back to the TO COUNT tab and click start count again on the same request (1)
- Frontend sees that the count is already in progress (based on the returned status) and so calls GET counts with id=1.

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
kchelstowski: tech huddle: this could be added to the cycle count item later

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -0,0 +1,64 @@
...
+    CycleCountStatus getStatus() {
+        List<CycleCountItem> cycleCountItems = CycleCountItem.findAllByCycleCount(this)
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.READY_TO_COUNT }) {
+            return CycleCountStatus.REQUESTED
+        }
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.REVIEWED }) {
+            return CycleCountStatus.REVIEWED
+        }
+        if (cycleCountItems.every { it.status == CycleCountItemStatus.TO_REVIEW }) {
+            return CycleCountStatus.TO_REVIEW
```
kchelstowski, line 47: change to "READY_TO_REVIEW"

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: tech huddle: remove the draft property

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: tech huddle: COUNTING -> INVESTIGATING

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
jmiranda: For what it's worth, it did work for me. I'll try to pull up the PR that I was working on to see if I did anything differently. Off the top of my head, the only difference I can remember is that the UrlMapping I used had braces around facility (I remember because Evan commented on it). I doubt that has any bearing since $facility and ${facility} should be equivalent, but worth a shot I guess.

```
"/api/facilities/${facility}/cycle-counts/start/batch"
```

jmiranda: Ok, I just tested this again and the command data binding works for GET requests, but not for POSTs. Interesting and annoying.

kchelstowski: @jmiranda thanks for testing it out. Yeah it's annoying. Btw - the link with the video doesn't work for me.

jmiranda: Sorry had to change the permissions on it. Should be accessible now.

awalkowiak (approved):

PR #5019, opened 2025-01-28
OBPIH-6936 remove duplicates from product classification list endpoint

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6936

**Description:** Remove duplicates from the product classification list endpoint by fetching the ABC classes as a Set

---

image

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
@@ -39,13 +39,13 @@ class ProductClassificationService {
             throw new IllegalArgumentException("Invalid facilityId: ${facilityId}")
         }
 
-        List<String> productClassifications = Product.createCriteria().listDistinct() {
+        Set<String> productClassifications = Product.createCriteria().listDistinct() {
             projections {
                 groupProperty("abcClass")
             }
             isNotNull("abcClass")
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/product/ProductClassificationService.groovy
```diff
@@ -55,15 +55,13 @@ class ProductClassificationService {
...
 
         // We convert to a DTO here instead of returning the raw string because it will make things easier
         // once product classification becomes a proper entity.
-        return productClassifications.collect {
-            new ProductClassificationDto(name: it)
-        }
+        return productClassifications.stream()
+                // Once ABC class becomes its own domain, this sort by name can be moved to the query.
+                .sorted()
+                .collect{ new ProductClassificationDto(name: it) }
```
alannadolny, line 65: do you have to use `stream` here? :thinking: 
Can't it be just:
```suggestion
        return productClassifications
                // Once ABC class becomes its own domain, this sort by name can be moved to the query.
                .sort()
                .collect{ new ProductClassificationDto(name: it) }
```

EWaterman: you're right. I'm still too Java-brained :sweat_smile:

awalkowiak (approved):

PR #5022, opened 2025-01-31
DO NOT MERGE! example implementation of binding java.time types

EWaterman:
DO NOT MERGE!

This PR is an example implementation of how we could use the new java.time classes in our domain objects.

PR that adds data binding support of Instant and LocalDate: https://github.com/openboxes/openboxes/pull/5023

The main piece of information here is TestingDatesApiSpec, which I've commented on explaining the drawbacks of java.util.Date, mainly around the timezone that it defaults to. I also demonstrate how Instant and LocalDate fix the issue.

Feel free to leave comments here as well.

Once my other PR gets merged, I'll close this one. I only created it so we'd have something to reference when reviewing the other PR.

PR #5023, opened 2025-01-31
OBPIH-6965 add data binding support for java.time classes

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6965

**Description:**

Add support for java.time.Instant and java.time.LocalDate data binding. With this change we can now use those data types in our Domain classes and DTO/Command objects. This is important because java.util.Date is functionally deprecated as of Java 8, so going forward we should stop using it if possible.

See my other PR for an example implementation that uses this change: https://github.com/openboxes/openboxes/pull/5022

This is based off [the discovery work I did on java.time in OBPIH-6916](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3382542337/OBPIH-6916+Standardizing+datetime+parsing). I encourage you to read that if you want some background information on this change.

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -11,6 +11,8 @@ package org.pih.warehouse.api
 
 import grails.converters.JSON
 import org.grails.web.json.JSONObject
+
+import org.pih.warehouse.DateUtil
 import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.DocumentService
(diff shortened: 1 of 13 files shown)


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -359,19 +361,12 @@ class StockMovementApiController {
...
-     * @param dateField
-     */
     private Date parseDateRequested(String date) {
-        return date ? Constants.EXPIRATION_DATE_FORMATTER.parse(date) : null
+        return DateUtil.asDate(date, Constants.EXPIRATION_DATE_FORMATTER)
     }
 
     private Date parseDateShipped(String date) {
-        return date ? Constants.DELIVERY_DATE_FORMATTER.parse(date) : null
+        return DateUtil.asDate(date, Constants.DELIVERY_DATE_FORMATTER)
```
EWaterman, line 369: This is a fix for https://pihemr.atlassian.net/browse/OBS-1831. I wanted something real to test my changes against so figured I'd fix the issue while I was at it.

src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -25,6 +25,10 @@ class Constants {
     static final adminControllers = ['createProduct', 'admin']
     static final adminActions = ['product': ['create'], 'person': ['list'], 'user': ['list'], 'location': ['edit'], 'shipper': ['create'], 'locationGroup': ['create'], 'locationType': ['create'], '*': ['delete']]
 
+    // TODO: Don't add more dates here! We should refactor all usages of these constants to instead use the
+    //       DateTimeFormatter constants in DateUtil. The backend should always return dates in the same format so that
+    //       the frontend can easily parse response objects. Let the frontend decide what the display format of each
+    //       individual field should be.
```
EWaterman, line 31: my idea here is that we should always use the same format when returning dates to the frontend, and then let the frontend format the dates for display.

As such, I hope we can eventually remove most of these from the backend and simply call toString() on the Instant/LocalDate datatypes in our toJson() method.

As for how to parse on the frontend, [Alan did a discovery that is quite relevant](https://pihemr.atlassian.net/wiki/spaces/OB/pages/3107422231)

PR #5025, opened 2025-02-03
OBPIH-6853 build table for TO RESOLVE tab

EWaterman:

- https://pihemr.atlassian.net/browse/OBPIH-6853
- https://pihemr.atlassian.net/browse/OBPIH-6854

**Description:**

The TO RESOLVE tab now contains a table like the TO COUNT tab with all the same filter abilities. For now the table contains a row for all cycle count requests where status == CREATED. There's a todo in the code to filter this down to only rows where the cycle count == COUNTED or INVESTIGATING, but that will require changes to the get cycle count candidates API to be able to filter by cycle count status (and not just cycle count request status) so I left that as a TODO for now.

---

Screenshot from 2025-02-03 15-23-14


src/js/hooks/cycleCount/useToResolveTab.jsx
```diff
@@ -0,0 +1,324 @@
...
+      header: () => (
+        <TableHeaderCell>
+          {translate('react.cycleCount.table.status.label', 'Status')}
+        </TableHeaderCell>
+      ),
+      cell: ({ getValue }) => (
+        <TableCell className="rt-td">
+          <StatusIndicator
+            variant="primary"
+            status={translate(`react.cycleCount.table.status.${getValue()}.label`, 'To Resolve')}
```
EWaterman, line 147: I wasn't sure how we normally translate values like this. The possible values are defined in an enum on the backend. In message.properties it's defined like:
```
react.cycleCount.status.COUNTED.label=To Resolve
react.cycleCount.status.INVESTIGATING.label=Resolution In Progress
...
```

alannadolny: https://pihemr.atlassian.net/browse/OBPIH-6931?atlOrigin=eyJpIjoiOGE1MTUwNWMxOTUyNDhhYzhhMTIzMzdiYzcxNTIyZDkiLCJwIjoiaiJ9

src/js/hooks/cycleCount/useToResolveTab.jsx
```diff
@@ -0,0 +1,324 @@
...
+    const payload = {
+      // TODO: replace this with a real call to startCount and pass the request id!
+      requests: checkedCheckboxes.map((productId) => ({
+        cycleCountRequest: productId,
+        countIndex: 1, // We only ever allow for a single recount, so index is always 1.
+      })),
+    };
+    spinner.show();
+    try {
+      await cycleCountApi.startCount(payload, currentLocation?.id);
```
alannadolny, line 306: you can also change the name of the method here to probably `startResolution`

EWaterman: it actually uses the same startCount API https://pihemr.atlassian.net/browse/OBPIH-6898

PR #5027, opened 2025-02-04
OBPIH-6966 Create an endpoint to get cycle counts

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -260,4 +260,13 @@ class CycleCountService {
 
         return availableItems
     }
+
+    List<CycleCountDto> getCycleCounts(List<String> ids) {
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -260,4 +260,13 @@ class CycleCountService {
 
         return availableItems
     }
+
+    List<CycleCountDto> getCycleCounts(List<String> ids) {
+        List<CycleCount> cycleCounts = CycleCount.createCriteria().list {
```
EWaterman, line 265: I think you can do `CycleCount.findByIdIn(ids)`

kchelstowski: I wanted to allow fetching all if you don't provide the ids, so this wouldn't work with version you proposed. I think those versions are anyway almost equal, it's just "syntatic sugar".

PR #5030, opened 2025-02-05
OBPIH-6931 filter cycle count candidates by cycle count status

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6931

**Description:**

We need to be able to filter the results in the TO COUNT and TO RESOLVE tabs by cycle count status (before it was being done on cycle count request status which doesn't have enough info for us to distinguish between the tabs). However it seems that the criteria doesn't support filtering by transient fields so the current code doesn't work. I'll try to think more about what to do here but maybe we can brainstorm together.


src/js/hooks/cycleCount/useToResolveTab.jsx
```diff
@@ -140,7 +137,6 @@ const useToResolveTab = ({
         </TableHeaderCell>
       ),
       cell: ({ getValue }) => (
-        // TODO: Use variant fetched from the API https://pihemr.atlassian.net/browse/OBPIH-6931
```
EWaterman, line 143: I removed the TODO here since I don't think this is something that the backend should be returning. Variant is only used for display purposes. The backend has no concept of what that field means (it's used as an html class), so it'd feel weird for the backend to decide on behalf of the frontend how it should be displaying rows of a certain status.

grails-app/migrations/0.9.x/changelog-2025-02-06-0000-add-column-status-to-cycle-count.xml
```diff
@@ -0,0 +1,19 @@
...
+    <changeSet author="ewaterman" id="060220250000-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="cycle_count" columnName="status"/>
+            </not>
+        </preConditions>
+
+        <addColumn tableName="cycle_count">
+            <column name="status" type="VARCHAR(100)">
+                <constraints nullable="false" />
```
EWaterman, line 15: TODO: set a default value for existing data! Or just truncate obdev3 and 5 because this isn't live yet??

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -23,10 +25,35 @@ class CycleCount {
     def beforeInsert() {
         createdBy = AuthService.currentUser
         updatedBy = AuthService.currentUser
+        status = recomputeStatus()
     }
 
     def beforeUpdate() {
         updatedBy = AuthService.currentUser
+        status = recomputeStatus()
```
awalkowiak, line 33: Is it updated after related items change as well? Have you considered using a formula for it (see point 8.2.11 here https://grails.github.io/legacy-gorm-doc/6.0.x/hibernate/manual/index.html#ormdsl)? Perhaps it would be a good/suiting approach as well?

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -23,10 +25,35 @@ class CycleCount {
...
+             *    not ever-growing because a cycle count doesn't stay open forever.
+             *
+             * 2) When we're working with a cycle count, we almost always want to fetch all of its items anyways for
+             *    validation purposes.
+             *
+             * Using hasMany greatly simplifies recomputing the cycle count status since it allows the cycle count to
+             * maintain an accurate list of items (which it uses to compute its status). Still, the disadvantage is
+             * important to note, and we should strive to replace this hasMany with a better option in the future.
+             */
+            cycleCountItems: CycleCountItem,
```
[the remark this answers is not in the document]
kchelstowski: @EWaterman that's weird, I might take a look if I find time, in my opinion if it's happening in the same session and transaction (I mean you update a few items, or add new), those should be immediately accessible via `CycleCountItem.findAllByCycleCount(this)`.

Anyway, I'm fine with having the bidirectional association, but I would lie if I didn't have in mind "why not a few tickets before??" :laughing:

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -23,10 +25,35 @@ class CycleCount {
...
+    ]
+
+    static mapping = {
+        cycleCountItems cascade: 'all-delete-orphan'
```
kchelstowski, line 56: I would like to have a convention around the cascading. There are pros and cons of having/not having cascade behavior. Due to the cascading we have sometimes run into unexpected issues difficult to spot. It's better to have some of the operations being done explicitly (in this case deleting).

grails-app/i18n/messages.properties
```diff
@@ -1534,17 +1534,17 @@ localization.invalid.custom.message=To enable database storage for custom transl
...
 # Locales
 locale.ach.label=Acholi
-locale.ar.label=عربى
+locale.ar.label=Ø¹Ø±Ø¨Ù
 locale.de.label=Deutsche
 locale.en.label=English
-locale.es.label=Español (España)
-locale.es_MX.label=Español (México)
+locale.es.label=EspanÌol (EspanÌa)
+locale.es_MX.label=EspaÃ±ol (MÃ©xico)
```
kchelstowski, line 1541: what happened here?

grails-app/migrations/0.9.x/changelog-2025-02-06-0000-add-column-status-to-cycle-count.xml
kchelstowski: We've run once recently into a similar issue. We've just cleared the data (via the script I've sent on Slack for you some time ago). I'm fine with doing the same again.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -249,7 +250,7 @@ class CycleCountService {
                 throw new IllegalArgumentException("Count index can't be higher than the highest count index of the items")
             }
         }
-        return CycleCountDto.createFromCycleCountItems(cycleCountItems)
+        return CycleCountDto.asDto(cycleCount)
```
kchelstowski, line 253: wouldn't `toDto` sound better?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -227,13 +227,14 @@ class CycleCountService {
...
-            if (!cycleCountItem.validate()) {
-                throw new ValidationException("Invalid cycle count item", cycleCountItem.errors)
-            }
-            cycleCountItem.save()
-            cycleCountItems.add(cycleCountItem)
+            newCycleCount.addToCycleCountItems(cycleCountItem)
         }
-        return CycleCountDto.createFromCycleCountItems(cycleCountItems)
+
+        if(!newCycleCount.save()) {
```
kchelstowski, line 233: nitpicky: missing space between if and `(`

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -201,16 +201,16 @@ class CycleCountService {
...
 
     CycleCountDto createCycleCount(CycleCountStartCommand request, Location facility) {
+        // We don't set the status on the cycle count here because it will be done automatically upon save.
         CycleCount newCycleCount = new CycleCount(
                 facility: facility,
                 dateLastRefreshed: new Date()
         )
+
         List<AvailableItem> itemsToSave = determineCycleCountItemsToSave(facility, request.cycleCountRequest.product)
-        newCycleCount.save()
```
kchelstowski, line 209: have you made sure it can be removed? I was running into error that "Not-null property references a transient value - transient instance must be saved before current operation", but I assume the cascading now makes it work, since you introduced the bidirectional association

kchelstowski (commented):
I would like to either do another round of review tomorrow

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: yup with the bidirectional it's not needed anymore

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
EWaterman: The problem with having it be derived was that it was inaccessible when building the session view in cycle-count-session.sql. I didn't try it the way you suggest but I feel like that would have the same issue. I can test though

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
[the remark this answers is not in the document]
EWaterman: if you have time to look at it, that'd be awesome. I was also assuming that they'd be accessible because it should be the same transaction, but I couldn't get it to work. You can see some of my older commits for references to what I was testing.

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
awalkowiak: Ok, I think this makes sense then.

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
EWaterman: yeah I agree. We don't ever delete cycle counts so I can change this to only cascade updates if we want, but I figured if we ever did need to delete a cycle count, we'd also want to delete the items (otherwise we'll get constraint violations)

PR #5042, opened 2025-02-12
User timezone not taken into account when displaying timestamps (fixes #5038)

### LINKED ISSUE #5038: User timezone not being honored when displaying some timestamps
User reported the following bug
> I can’t find how timezone is managed in 9.x version. When I change timezone in user profile. It doesn’t seem to make any effect for UI. For example it Product>Stock History tab.

Image

jmiranda:
This PR attempts to fix a bug (#5038) we noticed recently while performing a discovery on migrating our datetime dependencies from java.util to java.time.

Note: We'll end up consolidating the logic around displaying datetimes later. This is just to fix an annoying bug as unobtrusively as possible.

### DIFF
--- a/grails-app/taglib/org/pih/warehouse/FormatTagLib.groovy
+++ b/grails-app/taglib/org/pih/warehouse/FormatTagLib.groovy
@@ -26,6 +26,10 @@ class FormatTagLib {
     def date = { attrs, body ->
         if (attrs.obj != null) {
             DateFormat df = new SimpleDateFormat((attrs.format) ?: Constants.DEFAULT_DATE_FORMAT)
+            TimeZone tz = session.timezone
+            if (tz != null) {
+                df.setTimeZone(tz)
+            }
             out << df.format(attrs.obj)
         }
     }


grails-app/taglib/org/pih/warehouse/FormatTagLib.groovy
```diff
@@ -26,6 +26,10 @@ class FormatTagLib {
     def date = { attrs, body ->
         if (attrs.obj != null) {
             DateFormat df = new SimpleDateFormat((attrs.format) ?: Constants.DEFAULT_DATE_FORMAT)
+            TimeZone tz = session.timezone
+            if (tz != null) {
+                df.setTimeZone(tz)
+            }
```
EWaterman, line 32: This changes the default behaviour whenever we call `<format:date />` with a format that includes the date (not just time).

Did you test how this behaves for dates that don't have a time component such as expiration date?

For example, this one from outgoing/inbound transfers:
```
<format:date obj="${inventoryItem.expirationDate }"/>
```

My worry is that despite expiration date not being a precise moment in time (it's just a date), it is still stored as a full datetime object. If we have an expiration date of Jan 1st, 2025, it would be stored as 2025-01-01 00:00:00 UTC. This means if we convert the date to UTC-1 it'll show up as 2024-12-31 23:00:00 UTC-1, which will get formatted as 31/Dec/2024. So depending on where you are, expiration date is different, which I doubt is what we want.

This is a big reason why I wanted to switch to java.time instead. If the field was a LocalDate (instead of Date) it wouldn't be an issue. Date should only be used for precise exact moments in time.

You can test it out to see if my hunch is correct. If so, we'd probably need to do something silly here like check if the date is at midnight and if so, don't timezone convert, but that's not perfect either because we wouldn't be able to represent full datetimes that happen to occur at midnight.

ewaterman:
I'm going to close this PR since we've implemented a solution in https://github.com/openboxes/openboxes/pull/5474

The `format:date` taglib has since been deprecated in favour of `g:formatDate`

PR #5043, opened 2025-02-12
Default logo returns a 404 error (fixes #5041)

### LINKED ISSUE #5041: Default logo returns a 404 error on new installations
The default logo configuration in application.yml includes a relative path to an image within the application. 
```
# Allow users to customize logo image url as well as label
logo:
    url: "/assets/openboxes_logo_40x40.jpg"
    label: ""
```

Because the logo image does not exist under the configured path, Tomcat returns a 404 error and subsequently, the application does not render the logo image properly

Image

We, as developers, haven't noticed this very often because most of our existing production and test servers have configuration that overrides the default logo using a custom logo hosted on a separate server. 

Like this ...
```
openboxes.logo.url = "https://obdev3.pih-emr.org/images/pih_2020_logo.png"
```

Image

However, it becomes glaringly obvious when you install a new server with default configuration and see the broken image. 

Image

**Approach 1**
The easiest way to solve this is to hard-code the context path in the logo URL as highlighted in our configuration guide on this topic. 

```
logo:
    url: "/openboxes/assets/openboxes_logo_40x40.jpg"
```

**Approach 2**
A more ideal solution would be for the application to automatically add the context path in cases where it's needed either within the application.yml 

```
logo:
    url: "${self:custom.globalSchedule}/assets/openboxes_logo_40x40.jpg"
```

**Approach 3**
or in the API that gathers the configuration for the application. 
```
def logoUrl = location?.logo ? "${createLink(controller: 'location', action: 'viewLogo', id: location?.id)}" : grailsApplication.config.openboxes.logo.url
```
The second part of this is the one that's problematic. We need to wrap this configuration with something that adds the context path, i.e. using something like the createLink taglib. 
```
${createLink(url: grailsApplication.config.openboxes.logo.url}
```

For now, we are going to use the hard-coded fix to avoid a more time-consuming and potentially error-prone solution. But Approach 2 would probably be the best in the long-run.

jmiranda:
Fixed a long-standing bug with 0.9.x releases where the default configuration leads to a 404 error when rendering the default logo. See #5041 for more information.

This was tested locally using grails run-app, but I have not tested this as part of an actual deployment. If we encounter any issues with this approach, I'm going to hard-code the context path.

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -570,7 +570,7 @@ openboxes:
 
     # Allow users to customize logo image url as well as label
     logo:
-        url: "/assets/openboxes_logo_40x40.jpg"
+        url: "${server.contextPath}/assets/openboxes_logo_40x40.jpg"
         label: ""
     report:
         logo:


EWaterman (approved):
Since this is 404ing already, I'm fine with merging it in and seeing what happens

grails-app/conf/application.yml
```diff
@@ -570,7 +570,7 @@ openboxes:
 
     # Allow users to customize logo image url as well as label
     logo:
-        url: "/assets/openboxes_logo_40x40.jpg"
+        url: "${server.contextPath}/assets/openboxes_logo_40x40.jpg"
```
awalkowiak, line 573: And this does not affect the `displayLogo` tag? I mean, that this won't produce there `/openboxes/openboxes/assets/openboxes_logo_40x40.jpg`?

jmiranda: @awalkowiak Good point. I'll test it out, but I'm sure it does do that.

Can you and @alannadolny review ticket #5041 and see if you can think of a better approach?

jmiranda: Perhaps, we just need to make the change on the API side like Approach 3 suggests.

jmiranda: It looks like it works fine within displayLogo too.  Therefore I'm going to merge as-is

image

PR #5057, opened 2025-02-19
OBPIH-6901 Submit recount API

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6901

**Description:** The API for submitting a recount. Now that the recount has been submitted, we need to be able to submit not only the "snapshot" product inventory transaction, but also an adjustment transaction containing the changes in quantity that resulted from the count. I chose to split this logic out into its own component because the cycle count service was getting really big already.


EWaterman:
I haven't tested the code yet so I'm marking as do not merge, but I wanted to get some initial thoughts on the approach before I dug further into it.

src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -91,7 +91,6 @@ class Constants {
     static final String PRODUCT_INVENTORY_TRANSACTION_TYPE_ID = "11"
     static final String CYCLE_COUNT_PRODUCT_INVENTORY_TRANSACTION_TYPE_ID = "12"
     static final String CYCLE_COUNT_ADJUSTMENT_TRANSACTION_TYPE_ID = "13"
-    static final String CYCLE_COUNT_ADJUSTMENT_DEBIT_TRANSACTION_TYPE_ID = "14"
```
EWaterman, line 94: https://openboxes.slack.com/archives/C06JCHWR807/p1739986744402599

src/main/groovy/org/pih/warehouse/inventory/ProductInventorySnapshotTaker.groovy
```diff
@@ -0,0 +1,69 @@
...
+import org.pih.warehouse.product.Product
+
+/**
+ * Responsible for creating and persisting product inventory transactions. We call this a "snapshot taker" because
+ * product inventory transactions saved via this class are functionally a copy of what the quantity on hand
+ * (determined by product availability) was for each [bin location + lot number] combo of the product at the time.
+ * Importantly, transactions created via this class should NOT result in any quantity changes/adjustments.
+ */
+@Component
+class ProductInventorySnapshotTaker {
```
EWaterman, line 17: @jmiranda added you as a reviewer primarily to check out this file. I figure that as we move towards the "product inventory as snapshot" approach, we'll want this code to be in a class like this so it's easier to integrate into other places while keeping the logic the same.

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
```diff
@@ -0,0 +1,128 @@
...
+import org.springframework.stereotype.Component
+
+import org.pih.warehouse.core.Constants
+import org.pih.warehouse.product.Product
+
+/**
+ * Component responsible for creating and persisting the Transactions that result from a cycle count.
+ */
+@Component
+class CycleCountTransactionCreator {
```
kchelstowski, line 13: what's the reason it is not just a regular service? I can see it calling GORM operations, so imho it should be a regular service and should be marked as `@Transactional`

src/main/groovy/org/pih/warehouse/inventory/ProductInventorySnapshotTaker.groovy
```diff
@@ -0,0 +1,69 @@
...
+                inventory: facility.inventory,
+                transactionDate: transactionDate,
+                transactionType: transactionType,
+        )
+
+        transaction.transactionNumber = transactionIdentifierService.generate(transaction)
+
+        switch (sourceType) {
+            case ProductInventorySnapshotSource.CYCLE_COUNT:
+                transaction.cycleCount = source as CycleCount
```
kchelstowski, line 45: if you cast it to `CycleCount` here, why can't it become a type of the argument, instead of having `Object source`?

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
EWaterman: This is what convinced me: https://tedvinke.wordpress.com/2017/04/04/grails-anti-pattern-everything-is-a-service/

The flow I was imagining is controller -> service -> helper classes like this one, and since the service is already Transactional, not adding Transactional here would be an indicator that these classes shouldn't necessarily be called directly, it should always be through a service.

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
```diff
@@ -0,0 +1,128 @@
...
+    ProductInventorySnapshotTaker productInventorySnapshotTaker
+    TransactionIdentifierService transactionIdentifierService
+
+    /**
+     * Given a completed cycle count, will create and persist the resulting product inventory snapshot transaction,
+     * as well as the credit/debit adjustment transaction if there are discrepancies.
+     *
+     * @param itemQuantityOnHandIsUpToDate
+     */
+    List<Transaction> createTransactions(CycleCount cycleCount, boolean itemQuantityOnHandIsUpToDate=false) {
```
awalkowiak, line 25: When will `itemQuantityOnHandIsUpToDate` be passed as a `true`? Will it be a part of a request?

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
```diff
@@ -0,0 +1,128 @@
...
+            CycleCount cycleCount, Date transactionDate, boolean itemQuantityOnHandIsUpToDate) {
+        // We only care about the cycle count items from the most recent count. We create the entries first to avoid
+        // needlessly creating the transaction itself if there are no discrepancies.
+        List<TransactionEntry> entries = []
+        for (CycleCountItem cycleCountItem : cycleCount.itemsOfMostRecentCount) {
+            if (cycleCountItem.quantityCounted == null) {
+                continue
+            }
+
+            // We need to compare the quantity counted against the most up to date QoH in product availability.
```
awalkowiak, line 51: > We need to compare the quantity counted against the most up to date QoH in product availability.

A side question popped up in my head, and I do not remember if or when this was discussed. If something is already allocated, then should this be counted as well? So we are always counting QoH? Is there a chance that we miss in the count something that is allocated, but not yet sent?

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
```diff
@@ -0,0 +1,128 @@
...
+            // as was counted, if the QoH in the cycle count items is not up to date, and we have to re-compute it from
+            // product availability here, the amount of quantity adjustment/change might be different than what was
+            // displayed to the user during the count.
+            else {
+                // TODO: This doesn't account for any new bins/lots that have been created since the count started! We
+                //       need a full QoH fetch on the product, and to create new cycle count items for any new bins/lot
+                //       numbers.
+                Integer quantityFromProductAvailability = productAvailabilityService.getQuantityOnHandInBinLocation(
+                        cycleCountItem.inventoryItem, cycleCountItem.location) as Integer
+                actualQuantityOnHand = quantityFromProductAvailability ? quantityFromProductAvailability : 0
```
awalkowiak, line 68: You just can add `?: 0` at the end of the line above

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
EWaterman: sure we can discuss. And we can add Transactional here if we want. No need to make it a hard requirement that non-services always be called from services, I just felt it made the intended flow more explicit

src/main/groovy/org/pih/warehouse/inventory/CycleCountTransactionCreator.groovy
EWaterman: a good question. Lets discuss with Manon

EWaterman:
@jmiranda this is ready for review. Let me know if you have thoughts otherwise I'll merge it at the end of the day

src/main/groovy/org/pih/warehouse/core/Constants.groovy
jmiranda: Oh yeah we don't necessarily want to delete the DEBIT adjustment, just standardize on the use case where we use the CREDIT adjustments and allow the sign of the quantity data to dictate whether it's a credit or debit.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -967,6 +967,11 @@ class UrlMappings {
             action = [POST: "submitCount"]
         }
 
+        "/api/facilities/$facility/cycle-counts/$cycleCountId/recount" {
```
jmiranda, line 970: Let's pluralize the subresource to make it clear that it's a resource and not an action.

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -94,4 +94,19 @@ class CycleCount {
...
+    Integer getMaxCountIndex() {
+        return cycleCountItems.max{ it.countIndex }?.countIndex
+    }
+
+    /**
+     * @return All CycleCountItems with the highest countIndex, representing the newest (re)count.
+     */
+    Set<CycleCountItem> getItemsOfMostRecentCount() {
+        Integer countIndex = maxCountIndex
+        return cycleCountItems.findAll{ it.countIndex == countIndex}
```
jmiranda, line 110: I'd be okay with the following (slightly more succinct and sleek).
```
return cycleCountItems.findAll{ it.countIndex == maxCountIndex }
```

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -94,4 +94,19 @@ class CycleCount {
         }
         return null
     }
+
+    /**
+     * @return The largest count index of all the cycle count items. Helps determine what count we're on.
+     */
+    Integer getMaxCountIndex() {
+        return cycleCountItems.max{ it.countIndex }?.countIndex
```
jmiranda, line 102: I don't think this is possible (maybe if cycleCountItems is an empty list?), but if this could return a null, we might want to consider Elvis operator'ing a zero at the end.

grails-app/i18n/messages.properties
```diff
@@ -1122,6 +1122,7 @@ cycleCountStartRecountCommand.cycleCountRequest.noCycleCountFound=Cannot start r
 cycleCountStartRecountCommand.cycleCountRequest.invalidCycleCountStatus=Recount cannot be started when cycle count status is ({3})
 cycleCountStartRecountCommand.countIndex.invalid=Count index ({3}) is invalid for recount
 cycleCountSubmitCountCommand.cycleCount.invalidStatus=Invalid cycle count status to be able to submit the count
+cycleCountSubmitRecountCommand.cycleCount.invalidStatus=Invalid cycle count status to be able to submit the recount
```
jmiranda, line 1125: We've never adopted a standard naming convention for command classes before (and we can discuss this convention later). Throughout the application, we go back and forth between resource vs action so it doesn't really matter at this point. But this seems more like a combination of the action convention (SubmitRecount) with a namespace (CycleCount).

I like the action-oriented command class naming convention. 

So I think one of the following would 
* SubmitCycleCountRecountCommand 
* SubmitRecountCommand 

But would be fine with CycleCountRecountCommand as well.

grails-app/migrations/0.9.x/changelog-2025-02-21-0000-remove-cycle-count-debit-transaction-type.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
+    <changeSet author="ewaterman" id="210220250000-0">
```
jmiranda, line 6: Let's not do this here. Even if PIH has never created a debit adjustment it's possible there are existing transactions in other system so this is going to lead to some support headaches.

src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -91,7 +91,6 @@ class Constants {
     static final String PRODUCT_INVENTORY_TRANSACTION_TYPE_ID = "11"
     static final String CYCLE_COUNT_PRODUCT_INVENTORY_TRANSACTION_TYPE_ID = "12"
     static final String CYCLE_COUNT_ADJUSTMENT_TRANSACTION_TYPE_ID = "13"
```
jmiranda, line 93: Do we need the distinction between a product inventory transaction and cycle count product inventory transaction? Same with the adjustment?

src/main/groovy/org/pih/warehouse/core/Constants.groovy
jmiranda: Oh sorry, this was the cycle count adjustment debit, not the original one.

grails-app/migrations/0.9.x/changelog-2025-02-21-0000-remove-cycle-count-debit-transaction-type.xml
jmiranda: Nevermind, I just realized this was added erroneously. 

If it's possible, I would rather erase the earlier database migration (and manually remove the data from all "dev" databases). Since this never made it to production, that would be better than adding and removing.

EWaterman:
I'm going to submit this with the current changes but I'll start a doc for us where we can work on standardizing our naming scheme

PR #5060, opened 2025-02-21
OBPIH-7030 Reduce width of rows in count and recount workflow

SebastianLib:

**Description:**

---
image
image

### DIFF
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -618,11 +618,11 @@ tr.crossed-out td:before {
 
 .recalled-and-on-hold {
   background: repeating-linear-gradient(
-                  135deg,
-                  $orange-alpha,
-                  $orange-alpha 10px,
-                  $red-delta 10px,
-                  $red-delta 20px
+    135deg,
+    $orange-alpha,
+    $orange-alpha 10px,
+    $red-delta 10px,
+    $red-delta 20px
   );;
 }
 
@@ -905,9 +905,10 @@ div.crossed-out .table-inner-row:before {
 // Tooltip style
 .custom-tooltip {
   background-color: var(--blue-800);
+  border-radius: 2px;
   padding: 4px 8px;
   font-weight: 400;
-  font-size: 14px;
+  font-size: 12px;
   line-height: 15px;
   color: #FFFFFF;
 }
@@ -918,7 +919,7 @@ div.crossed-out .table-inner-row:before {
   font-size: 14px;
   line-height: 1.2;
   color: #FFFFFF;
-  max-width: 250px !important;
+  max-width: 250px
 }
 
 // Overriding built-in tooltip-style (react-tippy)
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -618,11 +618,11 @@ tr.crossed-out td:before {
 
 .recalled-and-on-hold {
   background: repeating-linear-gradient(
-    135deg,
-    $orange-alpha,
-    $orange-alpha 10px,
-    $red-delta 10px,
-    $red-delta 20px
+                  135deg,
+                  $orange-alpha,
+                  $orange-alpha 10px,
+                  $red-delta 10px,
+                  $red-delta 20px
   );;
 }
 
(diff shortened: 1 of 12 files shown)


src/css/main.scss
```diff
@@ -913,6 +913,15 @@ div.crossed-out .table-inner-row:before {
   color: #FFFFFF;
 }
 
+.custom-tooltip-v2 {
```
alannadolny, line 916: is the v2 needed?

src/css/main.scss
```diff
@@ -913,6 +913,15 @@ div.crossed-out .table-inner-row:before {
...
+  background-color: var(--blue-800);
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 20px;
+  color: #FFFFFF;
```
alannadolny, line 921: Do we have a defined variable for that color?

src/css/main.scss
```diff
@@ -1517,8 +1526,8 @@ div.search-input:focus-within {
   align-items: center;
 
   .info-bar-redirect > a {
-      text-decoration: underline;
-      color: #FFFFFF;
+    text-decoration: underline;
+    color: #FFFFFF;
```
alannadolny, line 1530: the same as above

src/js/components/form-elements/v2/DateField.jsx
```diff
@@ -74,6 +75,7 @@ const DateField = ({
       tooltip={tooltip}
       errorMessage={errorMessage}
       button={button}
+      hideErrorMessage={hideErrorMessage}
```
alannadolny, line 78: I think the error is hidden when the `errorMessage` is empty. Is this prop required?

src/js/components/form-elements/v2/SelectField.jsx
```diff
@@ -50,6 +51,7 @@ const SelectField = ({
       button={{ ...button, onClick: () => button.onClick(fieldProps?.value?.id ?? value) }}
       tooltip={tooltip}
       required={required}
+      hideErrorMessage={hideErrorMessage}
```
alannadolny, line 54: the same

src/js/components/form-elements/v2/TextInput.jsx
```diff
@@ -59,6 +60,7 @@ const TextInput = ({
       button={button}
       inputId={id || name}
       errorMessage={errorMessage}
+      hideErrorMessage={hideErrorMessage}
```
alannadolny, line 63: the same

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -158,10 +159,18 @@ const useCountStepTable = ({
...
             {...fieldProps}
           />
+          {error && (
+            <TooltipWrapper
+              content={error}
+              className="error-icon"
+            >
+              <RiErrorWarningLine />
+            </TooltipWrapper>
+          )}
```
alannadolny, line 173: 1. I would change the name of this component, because TooltipWrapper, suggests that the Tooltip should be passed as children of this because it is wrapping the passed children.
2. We should add a property to pass a custom icon instead of placing it as a wrapped component. The content here should be wrapped.

src/js/wrappers/TooltipWrapper.jsx
```diff
@@ -0,0 +1,34 @@
...
+    <Tooltip
+      delay="150"
+      duration="250"
+      hideDelay="50"
+      html={<div className="p-2 custom-tooltip-v2">{content}</div>}
+    >
+      {children}
+    </Tooltip>
+  </div>
+);
```
alannadolny, line 22: Can you elaborate more about the outside div? For me, it should look like this: (pseudocode)
```
    <Tooltip
      delay="150"
      duration="250"
      hideDelay="50"
      html={<div className="p-2 custom-tooltip-v2">{content}</div>}
    >
    // styling of the tooltip should be connected to the passed classname for that div (it can be just bootstrap class also)
    <div>
      {children}
      </div>
    </Tooltip>
);
```
I don't think we need the div that you used, because I don't recall whether we used this kind of wrapper for any of our tooltips.

alannadolny (commented):
Can you add a screenshot of the table rows and the new tooltips?

SebastianLib:
@alannadolny I've already added screenshots :D

src/js/wrappers/TooltipWrapper.jsx
SebastianLib: @alannadolny I wanted to do it this way, but then the position: absolute in tooltip doesn't work correctly

src/css/main.scss
SebastianLib: I added v2 because there is already class custom-tooltip

alannadolny: so I don't have a better idea of the name for this tooltip, but the current version doesn't seem right.

src/js/components/cycleCount/cycleCount.scss
```diff
@@ -97,8 +97,9 @@ limit-lines($lines-number) {
   }
 
   .static-cell-count-step {
-    padding: 10px;
-    padding-bottom: 18px;
+    display: flex;
+    align-items: center;
```
alannadolny, line 101: btw, that kind of style can be added using Bootstrap

alannadolny (approved):

PR #5061, opened 2025-02-22
OBPIH-6981 Create the update cycle count APIs 

kchelstowski:


grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -65,4 +68,28 @@ class CycleCountApiController {
...
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid cycle count item", command.errors)
+        }
+        CycleCountItemDto cycleCountItem = cycleCountService.updateCycleCountItem(command)
+
+        render([data: cycleCountItem] as JSON)
+    }
+
+    def deleteCycleCountItem() {
+        cycleCountService.deleteCycleCountItem(params.cycleCountItemId)
```
kchelstowski, line 82: TODO: add a check if it's a custom item

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -399,4 +399,46 @@ class CycleCountService {
...
+
+        return command.cycleCountItem.toDto()
+    }
+
+    CycleCountItemDto createCustomCycleCountItem(CycleCountCustomItemCommand command) {
+        CycleCountItem cycleCountItem = new CycleCountItem(
+                facility: command.facility,
+                status: CycleCountItemStatus.COUNTING,
+                countIndex: command.recount ? 1 : 0,
+                quantityOnHand: command.quantityCounted, // FIXME: should we assume that qoh = quantityCounted for a custom row?
```
EWaterman, line 417: a good question. I'd assume that because the system doesn't have that bin + lot for the product, that QoH would be 0

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -399,4 +399,46 @@ class CycleCountService {
             transaction.addToTransactionEntries(transactionEntry)
         }
     }
+
+    CycleCountItemDto updateCycleCountItem(CycleCountUpdateItemCommand command) {
+        command.cycleCountItem.quantityCounted = command.quantityCounted
+        command.cycleCountItem.comment = command.comment
+        command.cycleCountItem.countIndex = command.recount ? 1 : 0
+        command.cycleCountItem.status = CycleCountItemStatus.COUNTING
```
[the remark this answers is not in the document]
EWaterman: should we also update dateCounted here? I can't remember what the discussion was from that. Maybe only if quantityCounted changes?

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
```diff
@@ -0,0 +1,41 @@
...
+class CycleCountCustomItemCommand implements Validateable {
+
+    boolean recount
+
+    @BindUsing({ obj, source ->
+        Product product = Product.read(source['inventoryItem']['product'])
+        Date expirationDate = new Date(source['inventoryItem']['expirationDate'])
+        return InventoryItem.findByProductAndLotNumberAndExpirationDate(product, source['inventoryItem']['lotNumber'], expirationDate)
+    })
+    InventoryItem inventoryItem
```
EWaterman, line 18: I don't know why I didn't think of this before, but looking at inventory item:
```
    static constraints = {
        product(nullable: false)
        expirationDate(shared:"expirationDateConstraint")

grails.gorm.default.constraints = {
    expirationDateConstraint(nullable:true, min: new Date("01/01/2000"))
}
```
so it looks like product and lot number is enough for uniqueness on inventory item. Maybe you don't even need to provide expirationDate as a part of the findBy here

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+    def beforeValidate() {
+        String cycleCountId = RequestContextHolder.getRequestAttributes().params?.cycleCountId
+        String facilityId = RequestContextHolder.getRequestAttributes().params?.facility
+        cycleCount = CycleCount.read(cycleCountId)
+        facility = Location.read(facilityId)
+    }
+
+    static constraints = {
+        quantityCounted(nullable: true) // FIXME: should we allow for empty quantityCounted?
```
EWaterman, line 38: hmm I would say yes (we know we need to count this new bin but haven't done it yet) but maybe that's a question for Manon.

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -0,0 +1,25 @@
...
+
+    CycleCountItem cycleCountItem
+
+    def beforeValidate() {
+        String cycleCountItemId = RequestContextHolder.getRequestAttributes().params?.cycleCountItemId
+        cycleCountItem = CycleCountItem.get(cycleCountItemId)
+    }
+
+    static constraints = {
+        quantityCounted(nullable: true) // FIXME: should we allow clearing the quantityCounted?
```
EWaterman, line 22: yeah it makes sense to me that someone would want to do that

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -0,0 +1,25 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+import org.springframework.web.context.request.RequestContextHolder
+
+class CycleCountUpdateItemCommand implements Validateable {
+
+    boolean recount
```
EWaterman, line 8: same question. why not make this countIndex with a `min: 0, max: 1` range?

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -967,6 +967,16 @@ class UrlMappings {
             action = [POST: "submitCount"]
         }
 
+        "/api/facilities/$facility/cycle-counts/items/$cycleCountItemId" {
```
EWaterman, line 970: I was going to say that we should do `/api/facilities/$facility/cycle-counts/$cycleCountId/items/$cycleCountItemId` just to be extra clear and consistent with the other API but ultimately I think this is fine since item id is already enough to determine uniqueness.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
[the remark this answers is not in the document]
EWaterman: from Manon's docs https://pihemr.atlassian.net/wiki/spaces/OB/pages/3457777669/8.+Cycle+Count+-+MVP+Release

> Count date = Date of the count is initiated when creating the cycle count object on the product (click start count button or print count form button including the product). User should update it if some time has passed since that date until the physical count was actually done.

so yeah it looks like we want to suppor the user being able to change the dateCounted

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+import grails.databinding.BindUsing
+import grails.validation.Validateable
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+import org.springframework.web.context.request.RequestContextHolder
+
+class CycleCountCustomItemCommand implements Validateable {
+
+    boolean recount
```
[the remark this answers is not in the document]
kchelstowski: Let me know if after the short tech huddle's discussion it's still relevant

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -399,4 +399,46 @@ class CycleCountService {
...
+
+        return command.cycleCountItem.toDto()
+    }
+
+    CycleCountItemDto createCustomCycleCountItem(CycleCountCustomItemCommand command) {
+        CycleCountItem cycleCountItem = new CycleCountItem(
+                facility: command.facility,
+                status: CycleCountItemStatus.COUNTING,
```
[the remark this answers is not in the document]
kchelstowski: so in other words: if recount then INVESTIGATING, otherwise COUNTING?

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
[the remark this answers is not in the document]
EWaterman: no we can keep it as you have

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -65,4 +68,28 @@ class CycleCountApiController {
...
+    def updateCycleCountItem(CycleCountUpdateItemCommand command) {
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid cycle count item", command.errors)
+        }
+        CycleCountItemDto cycleCountItem = cycleCountService.updateCycleCountItem(command)
+
+        render([data: cycleCountItem] as JSON)
+    }
+
+    def deleteCycleCountItem() {
```
jmiranda, line 81: Would be good to standardize on a convention where we bind the ID as a method argument or the resource we are deleting. Both of these should work. The question is whether we need or want to do the lookup here vs within the service method.
```
def deleteCycleCountItem(String cycleCountItemId) 
def deleteCycleCountItem(CycleCountItem cycleCountItem)
```

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -65,4 +68,28 @@ class CycleCountApiController {
...
+        cycleCountService.deleteCycleCountItem(params.cycleCountItemId)
+
+        render(status: 204)
+    }
+
+    def createCustomCycleCountItem(CycleCountCustomItemCommand command) {
+        if (command.hasErrors()) {
+            throw new ValidationException("Invalid cycle count item", command.errors)
+        }
+        CycleCountItemDto cycleCountItem = cycleCountService.createCustomCycleCountItem(command)
```
jmiranda, line 91: Could we save the conversion toDto until the render? I think I want the service to be returning CycleCount and CycleCountItem. Why do we even need the DTO at this point?

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -56,6 +56,7 @@ class CycleCountItem {
                 status: status,
                 quantityOnHand: quantityOnHand,
                 quantityCounted: quantityCounted,
+                discrepancy: discrepancy,
```
jmiranda, line 59: Let's call this quantityVariance.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -399,4 +399,46 @@ class CycleCountService {
             transaction.addToTransactionEntries(transactionEntry)
         }
     }
+
+    CycleCountItemDto updateCycleCountItem(CycleCountUpdateItemCommand command) {
+        command.cycleCountItem.quantityCounted = command.quantityCounted
```
jmiranda, line 404: This method looks a bit weird. Could we do something like this
```
CycleCountItem cycleCountItem = CycleCountItem.get(command.id)

// Copy the standard parameters that are the same
cycleCountItem.properties = command.properties

// FIXME calculates the actual index based on the index of this cycle count item
command.cycleCountItem.countIndex = command.recount ? 1 : 0

// FIXME apply evan's requested changes
command.cycleCountItem.status = CycleCountItemStatus.COUNTING
cycleCountItem.date = new Date()

// return the cycle count item instead of the DTO.
return cycleCountItem
```

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
jmiranda: I think that's fair, but a validation would probably be a good idea if we were worried about a "race condition" i.e. someone added the new inventory item after we started the count.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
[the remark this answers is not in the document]
jmiranda: I'm ok with that for now, but just remember that we might eventually want to recount multiple times. But I'd be ok with explicitly preventing that now if it helped prevent us from harm.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+    @BindUsing({ obj, source ->
+        Product product = Product.read(source['inventoryItem']['product'])
+        Date expirationDate = new Date(source['inventoryItem']['expirationDate'])
+        return InventoryItem.findByProductAndLotNumberAndExpirationDate(product, source['inventoryItem']['lotNumber'], expirationDate)
```
jmiranda, line 16: Only Product and Lot Number are significant.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
jmiranda: I think the answer is yes from an API perspective. The fact that we currently create the cycle count items automatically from the cycle count service is good for the current UI, but there might be a need to create cycle count items from the client for use cases that aren't relatd to this "add missing inventory item to cycle count" use case.

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -0,0 +1,25 @@
+package org.pih.warehouse.inventory
+
+import grails.validation.Validateable
+import org.springframework.web.context.request.RequestContextHolder
+
+class CycleCountUpdateItemCommand implements Validateable {
```
jmiranda, line 6: Same point I made in Evan's PR. We should go with action or resource command naming convention. 

UpdateCycleCountItemCommand would be my preference. CycleCountItemCommand would be the other approach, which would make sense if we didn't need to distinguish between creating and updating cycle count item.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
```diff
@@ -0,0 +1,41 @@
+package org.pih.warehouse.inventory
+
+import grails.databinding.BindUsing
+import grails.validation.Validateable
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+import org.springframework.web.context.request.RequestContextHolder
+
+class CycleCountCustomItemCommand implements Validateable {
```
jmiranda, line 9: Same comment from Evan's PR (and below) about naming convention. Let's either go with a consolidated CycleCountItemCommand (that can be used for update, create, etc). Or an action-based command. In this case I don't think "custom" is an important distinction. It's just another cycle count item.

jmiranda (changes requested):

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -967,6 +967,16 @@ class UrlMappings {
...
+            controller = "cycleCountApi"
+            action = [PATCH: "updateCycleCountItem", DELETE: "deleteCycleCountItem"]
+        }
+
+        "/api/facilities/$facility/cycle-counts/$cycleCountId/items" {
+            controller = "cycleCountApi"
+            action = [POST: "createCustomCycleCountItem"]
```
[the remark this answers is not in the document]
kchelstowski: I can rename that, though in the domain we have a boolean flag `custom` that also determines whether it's a custom item or not, so I'd say it's not only UI specific.
Thoughts?

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: In one of the earliest API tickets I explained that in one of the tech huddles - the point was that back then it was easier for me to work on DTOs even on service levels to shorten the data as much as I could, so that the flow was more understandable for me, especially that back then we had to build the cycle count dto via an item, hence I wanted to avoid having such logic in the marshallers, and it just kept going like this throughout further APIs.
I can change that here, but we would somehow lose the "convention" we have throughout this API.

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
kchelstowski: It was expected to be named `discrepancy` in the docs. Have you guys had a conversation around that, or it's just your loose thought to rename that while giving the review?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: There is no need to fetch the cycle count item, because it's bound in the command and accessible via `command.cycleCountItem`.
I could try to play around with the `command.properties`, but I just wanted to be explicit here so that you can immediately see what fields are editable.

src/main/groovy/org/pih/warehouse/inventory/CycleCountCustomItemCommand.groovy
EWaterman: So I was wondering why we even have expirationDate as a field here if it's not being used anywhere. I can't remember what the original requirement was, but I wonder if the expectation is that if no inventoryItem exist for the given [product + lot number] pair, we should create it.

With your code currently, if the inventoryItem does not exist for the given [product + lot number] pair, it fails validation (because inventoryItem == null). This means expirationDate isn't used at all and we could remove it unless we're missing a requirement here to create the inventory item if it does not exist.

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemDto.groovy
```diff
@@ -22,6 +22,8 @@ class CycleCountItemDto {
 
     Integer quantityCounted
 
+    Integer discrepancy
+
     DiscrepancyReasonCode discrepancyReasonCode
```
jmiranda, line 27: There might be a time when we need to distinguish between reason code by activity (cycle count, allocation, picking, adjustments, etc) but when that happens we're probably going to build an extensible data model around it. 

For now, we need to remove the DiscrepancyReasonCode enum and replace it here with ReasonCode enum.

image

Source: https://openboxes.slack.com/archives/C06JCHWR807/p1738168383197519

jmiranda: @kchelstowski @EWaterman

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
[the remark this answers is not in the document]
jmiranda: Yes, that seems fine. What do we mean by "custom" though? Is there a more specific term for what you're trying to accomplish? Like, "add missing item"? And how does that item get treated different from other items being counted?

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemDto.groovy
EWaterman: it's currently an open question. Manon met with some of the Boston crew (Ali and Ginny) and they were unclear about the codes they want to display here. Likely it is going to be a subset of ReasonCode values but we've haven't gotten a conclusive answer yet.

But you're right we won't need a custom enum here.

For now we should probably use ReasonCode.listInventoryAdjustmentReasonCodes() and then have the frontend call ReasonCodeApiController.list() to fetch the list of possible values.

But we can do that in a separate PR once we know a bit more

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
EWaterman: I personally like the CycleCount prefix since it gives us a clear grouping of commands by feature. Feature + action. I think this is something we should standardize through discovery since we're definitely not consistent about it.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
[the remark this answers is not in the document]
EWaterman: We need the "custom" boolean for the sake of the frontend. It needs to know what items were manually added because they have special behaviour (only those items can be deleted / have their lot number + expiry date updated).

As far as the backend is concerned (and I think this is what you're getting at Justin) I don't necessarily think we need to distinguish between custom and non-custom items. We can just have the "create item" API and you decide if it's custom by setting the flag to true. We could even keep it as you have it, always setting custom to true until we need otherwise. So yeah we could just make this a createCycleCountItem method and a CycleCountItemCommand

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
EWaterman: yeah I like having the full CycleCountItem as a method arg. It's clearest to me. So if that's an option it gets my vote

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -399,4 +399,47 @@ class CycleCountService {
             transaction.addToTransactionEntries(transactionEntry)
         }
     }
+
+    CycleCountItemDto updateCycleCountItem(CycleCountUpdateItemCommand command) {
+        CycleCountItem cycleCountItem = command.cycleCountItem
+        cycleCountItem.properties = command.properties
+        cycleCountItem.countIndex = command.recount ? 1 : 0
+        cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
+        cycleCountItem.dateCounted = new Date()
```
EWaterman, line 408: I'm not sure if we want to auto-update the dateCounted here or if we just want to allow the user to (optionally) manually update it. Automatically makes sense to me but I think dateCounted is supposed to be a "when the physical count happened" and not "when the count was input into OB". A question for Manon.

Here's some new docs by Manon: https://pihemr.atlassian.net/wiki/spaces/OB/pages/3457777669/8.+Cycle+Count+-+MVP+Release#C.-Counted-by-%2F-Count-date

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: The:
```
def deleteCycleCountItem(CycleCountItem cycleCountItem)
```
doesn't work. It's probably due to the fact that the query params are not bound for `DELETE` action. 
As you @jmiranda recently discovered - this would only be bound if it was `POST` action.

The version with String works fine, so I will go with that.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: let's leave it like that for now until we are sure about the expected behavior. I wouldn't want this ticket to block the work anymore.

PR #5068, opened 2025-02-26
OBS-1831 Fix unparseable date: "01/01/2025 " exception during stock movements

EWaterman:

- https://pihemr.atlassian.net/browse/OBS-1831
- https://pihemr.atlassian.net/browse/OBS-1815

**Description:** This change switches the parsing of dateShipped and dateRequested fields during stock movements to use the new DateUtil.asDate method (with no SimpleDateFormat provided so that it uses the flexible format).

I also modified how the method works so that it uses the same DateTimeFormatter that we use when parsing a String to an Instant. That way it gets around the issue with parsing "MM/dd/yyyy" and defaulting to the local timezone (now it always uses UTC).

---

image-20241120-211537

### DIFF
--- a/grails-app/conf/application.yml
+++ b/grails-app/conf/application.yml
@@ -204,16 +204,17 @@ grails:
         # We now use the java.time classes instead (ex: Instant, LocalDate). Those fields define their own
         # ValueConverter classes and so don't use the formats defined here.
         dateFormats:
-            # Old date formats, don't use going forward! Kept first in the list to maintain backwards compatability.
+            # Date formats used by Excel and our old APIs. Kept first in the list to maintain backwards compatability.
             - 'MM/dd/yyyy HH:mm:ss XXX'
             - 'MM/dd/yyyy HH:mm:ss XX'
             - 'MM/dd/yyyy HH:mm:ss X'
             - 'MM/dd/yyyy HH:mm XXX'
             - 'MM/dd/yyyy HH:mm XX'
             - 'MM/dd/yyyy HH:mm X'
-            # Note that if this format is given, Grails will treat it as midnight in the local server timezone, which
-            # is ahead of UTC, can cause the date to actually be the next day. It can also cause inconsistencies across
-            # hosts/environments if they're configured for different local timezones. Avoid this format going forward!
+            # Note that Grails will treat this format as midnight in the local server timezone. For zones that are
+            # ahead of UTC, this can cause the date to actually become the next day. It can also cause
+            # inconsistencies across hosts/environments if they're configured for different local timezones.
+            # Avoid using this format going forward! Only kept to maintain backwards compatability.
             - 'MM/dd/yyyy'
             # ISO-8601 formats. These are better than the old formats, but new APIs should still use Instant instead
             # of Date whenever possible. Note that a format without time and timezone is NOT allowed here because that
--- a/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
@@ -14,7 +14,6 @@ import org.grails.web.json.JSONObject
 
 import org.pih.warehouse.DateUtil
 import org.pih.warehouse.core.ActivityCode
-import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.DocumentService
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.RoleType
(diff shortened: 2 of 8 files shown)


grails-app/controllers/org/pih/warehouse/api/StockMovementApiController.groovy
```diff
@@ -361,26 +360,18 @@ class StockMovementApiController {
         render([data: pendingRequisitionDetails] as JSON)
     }
 
-    private Date parseDateRequested(String date) {
-        return DateUtil.asDate(date, Constants.EXPIRATION_DATE_FORMATTER)
-    }
-
-    private Date parseDateShipped(String date) {
-        return DateUtil.asDate(date, Constants.DELIVERY_DATE_FORMATTER)
```
EWaterman, line 369: this is the real change. Notice that before we were providing the specific formatter to use but now we don't. Now this will use the flexible date formatter which allows for a much wider range of possible inputs

grails-app/utils/org/pih/warehouse/databinding/DataBindingConstants.groovy
```diff
@@ -22,13 +27,27 @@ class DataBindingConstants {
...
             "[HH:mm:ss][HHmmss]" +
             "[HH:mm][HHmm]"
     static final String FLEXIBLE_DATE_TIME_ZONE_PATTERN =
-            "${FLEXIBLE_DATE_PATTERN}[['T'][ ]${FLEXIBLE_TIME_PATTERN}][[XXX][XX][X]]"
+            "${FLEXIBLE_DATE_PATTERN}[['T'][ ]${FLEXIBLE_TIME_PATTERN}][[ ][XXX][XX][X]]"
+
+    /**
+     * DateTimeFormatter used to parse a String to a datetime class, such as java.time.Instant.
+     */
+    static final DateTimeFormatter FLEXIBLE_DATE_TIME_ZONE_FORMAT = new DateTimeFormatterBuilder()
```
EWaterman, line 43: not new code. Just moved from InstantValueConverter so that it can be shared

src/test/groovy/unit/org/pih/warehouse/utils/DateUtilSpec.groovy
```diff
@@ -51,27 +51,26 @@ class DateUtilSpec extends Specification {
...
+        // When not given a timezone, we default to UTC.
+        "01/01/2000 00:00"          || "2000-01-01T00:00:00Z" | "No timezone our format"
+        "2000-01-01 00:00"          || "2000-01-01T00:00:00Z" | "No timezone ISO format"
 
-        // It's noteworthy that the "yyyy" format for java.util.Date supports a two year format but it is year 0 based,
-        // not based on the year of this century (ie "25" is year 25, not 2025)!
-        "01/01/00 00:00 Z"          || "0001-01-01T00:00:00Z" | "No time or timezone our format two digit year"
+        // When not given a time or timezone, we default to midnight UTC.
+        "01/01/2000"                || "2000-01-01T00:00:00Z" | "No time or timezone our format"
+        "2000-01-01"                || "2000-01-01T00:00:00Z" | "No time or timezone ISO format"
```
EWaterman, line 60: we can parse to a Date without providing a timezone and it automatically uses UTC now. Yay!

grails-app/conf/application.yml
```diff
@@ -204,16 +204,17 @@ grails:
...
             - 'MM/dd/yyyy HH:mm XX'
             - 'MM/dd/yyyy HH:mm X'
-            # Note that if this format is given, Grails will treat it as midnight in the local server timezone, which
-            # is ahead of UTC, can cause the date to actually be the next day. It can also cause inconsistencies across
-            # hosts/environments if they're configured for different local timezones. Avoid this format going forward!
+            # Note that Grails will treat this format as midnight in the local server timezone. For zones that are
+            # ahead of UTC, this can cause the date to actually become the next day. It can also cause
+            # inconsistencies across hosts/environments if they're configured for different local timezones.
+            # Avoid using this format going forward! Only kept to maintain backwards compatability.
             - 'MM/dd/yyyy'
```
EWaterman, line 218: this problem will always be present so long as we use the Date class. Explicility calling
```
Date date = DateUtil.asDate("01/01/2000") 
```
will correctly default to UTC, but if we rely on Grails' databinding (ie have a controller method take in a Command object arg that has a Date field), Strings like "01/01/2000" will default to the local server time. This is why switching to Instant for new features is better

awalkowiak (approved):

alannadolny (approved):

PR #5073, opened 2025-02-27
OBPIH-6950 Call the update count APIs when editing/creating/deleting …

kchelstowski:
…cycle count items

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6950

**Description:**

---


src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -123,8 +172,46 @@ const useCountStep = () => {
...
-    console.log('save');
+  const save = async () => {
+    // This data should be combined to a single request
+    // eslint-disable-next-line no-restricted-syntax
+    for (const cycleCount of tableData.current) {
+      const cycleCountItemsToUpdate = cycleCount.cycleCountItems.filter((item) => ((item.updated || item.status.name !== CycleCountItemStatus.COUNTING) && !item.id.includes('newRow')));
+      // eslint-disable-next-line no-restricted-syntax
+      for (const cycleCountItem of cycleCountItemsToUpdate) {
+        try {
+          dispatch(showSpinner());
```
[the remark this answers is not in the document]
kchelstowski: I didn't know we have such hook, so thanks for the suggestion. I changed that although I don't like the names of the methods `show`, `hide`. I think they should be named `showSpinner` and `hideSpinner` but I don't want to implement such a refactoring now to the hook.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -380,7 +381,6 @@ class CycleCountService {
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
         cycleCountItem.dateCounted = new Date()
-        cycleCountItem.assignee = AuthService.currentUser
```
EWaterman, line 383: this should be in the command too, right? They're allowed to update the assignee in the UI

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -31,20 +35,34 @@ const useCountStep = () => {
...
       const countedDates = data?.data?.reduce((acc, cycleCount) => ({
         ...acc,
         [cycleCount?.id]: cycleCount?.cycleCountItems[0].dateCounted,
       }), {});
       setDateCounted(countedDates);
-    })();
+      const countedByMap = data?.data?.reduce((acc, cycleCount) => ({
+        ...acc,
+        [cycleCount?.id]: cycleCount?.cycleCountItems[0].assignee,
+      }), {});
```
alannadolny, line 56: btw, you could get the `countedDate` and `countedBy` in on reduce

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -123,8 +174,45 @@ const useCountStep = () => {
     setIsStepEditable(true);
   };
 
-  const save = () => {
-    console.log('save');
+  const save = async () => {
+    // eslint-disable-next-line no-restricted-syntax
```
alannadolny, line 178: why is this needed?

alannadolny: Looking at this code, I think you can disable those rules for the whole file instead of disabling the same rules x times

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: It is in the command

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: I did it initially, but ran into some problems with accessibility in the further state. I don't remember the exact context though. If you find time, you could try to do it.

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: It's needed because eslint suggested to use `.forEach` method instead of loop, but as you know, we can't use async actions inside `.forEach`

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -123,8 +174,45 @@ const useCountStep = () => {
...
+    for (const cycleCount of tableData.current) {
+      const cycleCountItemsToUpdate = cycleCount.cycleCountItems.filter((item) => ((item.updated || item.status.name !== CycleCountItemStatus.COUNTING) && !item.id.includes('newRow')));
```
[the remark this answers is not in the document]
kchelstowski: This could be removed now probably given that we disabled the validation on frontend. Initially I thought we needed to send all the items, so in order to know which items were not yet sent and are not edited, these must have been in READY_TO_COUNT status.

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
```diff
@@ -26,6 +27,8 @@ class CycleCountItemCommand implements Validateable {
 
     Location facility
 
+    User assignee
```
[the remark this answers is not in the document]
kchelstowski: I had a conversation with Manon about that yesterday. The details are specified in the OBPIH-7038. Let's leave it not nullable here, and handle that specific case in the OBPIH-7038.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: oof yeah I somehow missed your above comment

EWaterman (approved):

awalkowiak (approved):

PR #5074, opened 2025-02-27
OBPIH-7039 Populate Root Cause dropdown with proper values when recounting

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7039

**Description:** Previously we were using a custom DiscrepancyReasonCode enum to list the possible Root Cause values. This PR switches to use ReasonCode.listInventoryAdjustmentReasonCodes().

---

image


src/js/actions/index.js
```diff
@@ -102,13 +102,33 @@ export function hideUserActions() {
   };
 }
 
-export function fetchReasonCodes() {
-  const url = '/api/reasonCodes';
+export function fetchReasonCodes(activityCode = null, dispatchType = FETCH_REASONCODES) {
+  let url = '/api/reasonCodes';
+  if (activityCode) {
+    url += `?activityCode=${activityCode}`;
+  }
```
alannadolny, line 109: I am not a big fan of mutating values, so I prefer something like:
```
const url = `/api/reasonCodes${activityCode ? `?activityCode=${activityCode}` : ''}`
```

src/js/actions/index.js
```diff
@@ -102,13 +102,33 @@ export function hideUserActions() {
...
+          type: dispatchType,
+          payload: reasonCodes,
+        });
+        return;
+      }
       dispatch({
-        type: FETCH_REASONCODES,
-        payload: res.data,
+        type: dispatchType,
+        payload: [],
```
alannadolny, line 131: I think you should be able to do something like this:
```
        const reasonCodes = res?.data?.data?.map?.((reasonCode) => (
          {
            id: reasonCode.id,
            value: reasonCode.id,
            label: reasonCode.name,
          }
        ));
        dispatch({
          type: dispatchType,
          payload: reasonCodes || [],
        });
```
When something from the call chain is null/undefined, it will fall back to [].

Note:
`||` use the fallback when the checked value is falsy,
`??` use the fallback when the checked value is null.
In this case, I think `reasonCodes` will fall back to `null`.

src/js/reducers/reasonCodesReducer.jsx
```diff
@@ -10,17 +8,11 @@ const initialState = {
...
-          data: reasonCodes,
-          fetched: true,
-        };
-      }
-      return state;
+      return {
+        ...state,
+        data: action.payload,
+        fetched: true,
+      };
```
alannadolny, line 15: Won't that change break the use cases for reason codes in other places?

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
```diff
@@ -38,6 +41,11 @@ class CycleCountItemCommand implements Validateable {
 
     static constraints = {
         quantityCounted(nullable: true)
+        discrepancyReasonCode(nullable: true, validator: { ReasonCode discrepancyReasonCode ->
+            return ReasonCode.listInventoryAdjustmentReasonCodes().contains(discrepancyReasonCode) ?
+                    true :
+                    ['cycleCountItemCommand.discrepancyReasonCode.invalid']
+        })
```
alannadolny, line 48: This is just an idea, not a change request.
Maybe it will be worth having. a place where we can store our validationrs to keep the classes cleaner? I can see that this closure is duplicated.

alannadolny (approved):

PR #5075, opened 2025-02-28
OBPIH-6939 Show minimum of cycle count date and inventory in Last Counted date column

jmiranda:

Added views to calculate the last count date based on cycle count and product inventory transaction data. Last count date is being pulled into to the the cycle count session view and displayed on the All Product tab. 

The performance is still really bad for some locations, so I'm going to look into ways to make it faster.

https://pihemr.atlassian.net/browse/OBPIH-6939

TBD

> [!TIP]  
> For the best viewing experience while reviewing this code, I would recommend enabling **Unified view** and **Hide whitespace** 
> 
> image
> 
> Not doing this will make it very difficult to understand what's going on.

### DIFF
--- a/grails-app/migrations/views/cycle-count-session.sql
+++ b/grails-app/migrations/views/cycle-count-session.sql
@@ -11,10 +11,16 @@ SELECT
     cycle_count_request_summary.cycle_count_request_id                               as cycle_count_request_id,
 
     -- ABC Classification
-    MAX(COALESCE(inventory_level_summary.abc_class, product.abc_class))              as abc_class,
+    -- FIXME Using a grouping operator due to a GROUP BY issue. This isn't the best approach sicne it'll return an
+    --  empty string or NULL value ahead of valid ABC classes (A, B, C, etc). But I don't have a better solution and
+    --  this will likely work for 99% of cases.
+    MIN(COALESCE(inventory_level_summary.abc_class, product.abc_class))              as abc_class,
 
     -- Status of any pending cycle counts
     -- Consolidate the statuses down to a single field representing the status of the candidate itself.
+    -- FIXME Using a grouping operator because the SQL is invalid otherwise. There's probably a better way to handle
+    --  this (perhaps pulling this into a separate domain that can be joined to the CycleCountSession when needed for
+    --  certain queries).
     MAX(cycle_count_request_summary.status)                                          as status,
 
     # Inventory Item Count
--- a/grails-app/migrations/views/cycle-count-session.sql
+++ b/grails-app/migrations/views/cycle-count-session.sql
@@ -1,40 +1,56 @@
-CREATE OR REPLACE VIEW cycle_count_session AS (
-	SELECT
-		# ID hash of product.id and location.id
-		CRC32(CONCAT(location.id, product.id)) as id,
-        inventory.id as inventory_id,
-		product.id as product_id,
-		location.id as facility_id,
-        cycle_count_request.id as cycle_count_request_id,
-        COALESCE(inventory_level.abc_class, product.abc_class) as abc_class,
+CREATE OR REPLACE VIEW cycle_count_session AS
+(
+SELECT
+    # ID hash of product.id and location.id
+    CRC32(CONCAT(location.id, product.id))                                       as id,
+    inventory.id                                                                 as inventory_id,
+    product.id                                                                   as product_id,
+    location.id                                                                  as facility_id,
+    #cycle_count_request.id as cycle_count_request_id,
+    NULL                                                                         as cycle_count_request_id,
+    COALESCE(inventory_level_summary.abc_class, product.abc_class)                       as abc_class,
 
-        # Consolidate the statuses down to a single field representing the status of the candidate itself.
-        COALESCE(cycle_count.status, cycle_count_request.status) as status,
+    # Consolidate the statuses down to a single field representing the status of the candidate itself.
+    NULL                                                                         as status, # COALESCE(cycle_count.status, cycle_count_request.status) as status,
 
-		# Inventory Item Count
-		count(*) as inventory_item_count,
+    # Inventory Item Count
+    count(product_availability.id)                                               as inventory_item_count,
 
-		# Comma-separated list of internal locations
-		GROUP_CONCAT(DISTINCT product_availability.bin_location_name) as internal_locations,
+    # Comma-separated list of internal locations
+    GROUP_CONCAT(DISTINCT product_availability.bin_location_name)                as internal_locations,
 
-		# Quantities by SKU
-		sum(product_availability.quantity_on_hand) as quantity_on_hand,
-		sum(product_availability.quantity_available_to_promise) as quantity_available,
+    # Quantities by SKU
+    sum(product_availability.quantity_on_hand)                                   as quantity_on_hand,
+    sum(product_availability.quantity_available_to_promise)                      as quantity_available,
 
-        # Stuff we need to include now or wait for the materialized view
-		NULL as negative_item_count,
-		NULL as date_last_count,
-		NULL as date_next_count,
-		NULL as date_latest_inventory
-	FROM product_availability
-	JOIN location on product_availability.location_id = location.id
-	JOIN product on product_availability.product_id = product.id
-	JOIN inventory on location.inventory_id = inventory.id
-    LEFT OUTER JOIN inventory_level on product.id = inventory_level.product_id AND inventory.id = inventory_level.inventory_id
-    LEFT OUTER JOIN cycle_count_request ON product.id = cycle_count_request.product_id AND location.id = cycle_count_request.facility_id
-    LEFT OUTER JOIN cycle_count ON cycle_count_request.cycle_count_id = cycle_count.id
-    WHERE product_availability.quantity_on_hand > 0
-       AND (cycle_count_request.id IS NULL OR (cycle_count_request.status <> 'COMPLETED' AND cycle_count_request.status <> 'CANCELED'))
-	GROUP BY location.id, product.id, abc_class, cycle_count_request.id
-    ORDER BY abc_class, inventory_item_count desc
-);
+    # Stuff we need to include now or wait for the materialized view
+    -- Negative item count
+    (SELECT SUM(CASE WHEN pai.quantity_on_hand < 0 THEN 1 ELSE 0 END)
+     FROM product_availability pai
+     WHERE pai.product_id = product_availability.product_id
+       AND pai.location_id = product_availability.location_id)                   as negative_item_count,
+
+    (select date_counted
+     from product_count_history
+     where product_count_history.product_id = product_availability.product_id
+       and product_count_history.facility_id = product_availability.location_id) as date_last_count,
+
+    NULL                                                                         as date_next_count,
+    NULL                                                                         as date_latest_inventory
+FROM product_availability
+         JOIN location on product_availability.location_id = location.id
+         JOIN inventory on location.inventory_id = inventory.id
+         JOIN product on product_availability.product_id = product.id
+    # FIXME need to add a unique constraint to prevent duplicate inventory levels for a single facility / product
+    # Using min(abc_class) means we'll return the highest value if there are multiple inventory levels with different abc classes (A, B, C)
+         LEFT OUTER JOIN (select inventory_id, product_id, min(abc_class) as abc_class
+                          from inventory_level where inventory_level.internal_location_id IS NULL
+                          group by inventory_id, product_id) as inventory_level_summary
+                         ON product_availability.product_id = inventory_level_summary.product_id AND inventory.id = inventory_level_summary.inventory_id
+    # FIXME need to add the status as a subquery or a
+        #LEFT OUTER JOIN cycle_count_request ON product.id = cycle_count_request.product_id AND location.id = cycle_count_request.facility_id
+        #   AND (cycle_count_request.id IS NULL OR (cycle_count_request.status <> 'COMPLETED' AND cycle_count_request.status <> 'CANCELED'))
+GROUP BY location.id, product.id
+HAVING sum(product_availability.quantity_on_hand) > 0
+ORDER BY NULL
+    );
--- a/grails-app/migrations/views/cycle-count-session.sql
+++ b/grails-app/migrations/views/cycle-count-session.sql
@@ -7,7 +7,7 @@ SELECT
     product.id                                                                   as product_id,
     location.id                                                                  as facility_id,
 
-    -- Cycle count request ID (deprecated)
+    -- Cycle count request ID
     cycle_count_request_summary.cycle_count_request_id as cycle_count_request_id,
 
     -- ABC Classification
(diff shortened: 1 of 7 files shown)


grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,56 @@
...
+    product.id                                                                   as product_id,
+    location.id                                                                  as facility_id,
+    #cycle_count_request.id as cycle_count_request_id,
+    NULL                                                                         as cycle_count_request_id,
+    COALESCE(inventory_level_summary.abc_class, product.abc_class)                       as abc_class,
 
-        # Consolidate the statuses down to a single field representing the status of the candidate itself.
-        COALESCE(cycle_count.status, cycle_count_request.status) as status,
+    # Consolidate the statuses down to a single field representing the status of the candidate itself.
+    NULL                                                                         as status, # COALESCE(cycle_count.status, cycle_count_request.status) as status,
```
jmiranda, line 14: By the way, this is coming back. I'm just trying to figure out best way to pull it in.

jmiranda:
Note: For the best viewing experience while reviewing this code, I would recommend enabling **Unified view** and **Hide whitespace** 
image

Not doing this will make it very difficult to understand what's going on.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+                         where cycle_count_request.id IS NULL
+                            OR (cycle_count_request.status <> 'COMPLETED'
+                                    AND cycle_count_request.status <> 'CANCELED')) as cycle_count_request_summary
+            on cycle_count_request_summary.product_id = product_availability.product_id
+                   and cycle_count_request_summary.facility_id = product_availability.location_id
+
+
+GROUP BY product_availability.location_id, product_availability.product_id
+HAVING sum(product_availability.quantity_on_hand) > 0
+ORDER BY NULL
```
jmiranda, line 73: To be clear, there is a purpose to this line. Since I'm trying to tune performance as much as possible I decided to remove the default sorting on the view and leave it to the query to decide.  

I read about this in the following stackoverflow post (mysql reference included)
https://dba.stackexchange.com/questions/208166/group-by-needs-order-by-null-to-avoid-filesort

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+                                  left outer join cycle_count on cycle_count_request.cycle_count_id = cycle_count.id
+                         where cycle_count_request.id IS NULL
+                            OR (cycle_count_request.status <> 'COMPLETED'
+                                    AND cycle_count_request.status <> 'CANCELED')) as cycle_count_request_summary
+            on cycle_count_request_summary.product_id = product_availability.product_id
+                   and cycle_count_request_summary.facility_id = product_availability.location_id
+
+
+GROUP BY product_availability.location_id, product_availability.product_id
+HAVING sum(product_availability.quantity_on_hand) > 0
```
jmiranda, line 72: Someone please verify the requirements here. The original code in the view was looking at individual product availability rows (if I'm not mistaken). This HAVING clause just makes sure we are filtering out records with quantity on hand <= 0 at the product level, not for individual inventory items within a product.

grails-app/migrations/views/cycle-count-session.sql
jmiranda: Additional resources
* https://dba.stackexchange.com/questions/202386/mariadb-why-order-by-null-command-doesnt-remain-using-filesort-on-the-extr
* https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html

grails-app/migrations/views/cycle-count-session.sql
jmiranda: And it's back.

grails-app/migrations/views/product-receipt-history.sql
```diff
@@ -0,0 +1,11 @@
+create or replace view product_receipt_history as
+(
+select shipment.destination_id as facility_id,
+       receipt_item.product_id,
+       receipt_item.inventory_item_id,
+       actual_delivery_date    as date_received
+from receipt_item
+         join receipt on receipt_item.receipt_id = receipt.id
+         join shipment on receipt.shipment_id = shipment.id
+         join location on shipment.destination_id = location.id
```
awalkowiak, line 10: Is this join required?

grails-app/migrations/views/product-physical-count-history.sql
```diff
@@ -0,0 +1,14 @@
...
+create or replace view product_physical_count_history as
+(
+    select
+        transaction.inventory_id as inventory_id,
+        inventory_item.product_id,
+        max(transaction.transaction_date) as date_counted
+    from transaction_entry
+    join transaction on transaction.id = transaction_entry.transaction_id
+    join inventory_item on transaction_entry.inventory_item_id = inventory_item.id
+    where transaction.transaction_type_id = '11'
```
awalkowiak, line 12: Could you add a comment saying what is the `11` transaction type?

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+CREATE OR REPLACE VIEW cycle_count_session AS
+(
+SELECT
+    # ID hash of product.id and location.id
+    CRC32(CONCAT(location.id, product.id))                                       as id,
+    location.inventory_id                                                                 as inventory_id,
+    product.id                                                                   as product_id,
+    location.id                                                                  as facility_id,
+
+    -- Cycle count request ID (deprecated)
```
awalkowiak, line 10: Why is it deprecated? Looks like it is still used on the front end, should we pivot to something else, or this will be cleaned up in the next release?

grails-app/migrations/views/cycle-count-session.sql
awalkowiak: How this will work with what looks like the possibility that `quantity_on_hand` can have a negative value?

awalkowiak (commented):
I looked through it, and I have a few minor comments. On top of which, I'd really like to add +1 to a request with unifying SQL keywords (making them all upper case).

Btw... Now I know how it is to review the Order Summary views for someone that have not written them :sweat_smile:

grails-app/migrations/views/cycle-count-session.sql
awalkowiak: As I wrote in my review this `HAVING sum(product_availability.quantity_on_hand) > 0` seems risky, as looks like we can have negative quantities, so I think it might remove some records. Or if one row has a high negative value and the other one has a low positive value, then should it be removed here eitherway?

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+                              product_id,
+                              min(abc_class) as abc_class
+                          from inventory_level where inventory_level.internal_location_id IS NULL
+                          group by inventory_id, product_id) as inventory_level_summary
+                         ON product_availability.product_id = inventory_level_summary.product_id AND location.inventory_id = inventory_level_summary.inventory_id
+
+        LEFT OUTER JOIN (select cycle_count_request.facility_id                          as facility_id,
+                                cycle_count_request.product_id                           as product_id,
+                                cycle_count_request.id                                   as cycle_count_request_id,
+                                COALESCE(cycle_count.status, cycle_count_request.status) as status
```
EWaterman, line 61: I'm not sure this coalesced status should be a part of cycle_count_request_summary given that it will be cycle count status a vast majority of the time. It seems like this is the session status, not the request status

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+    # Quantities by SKU
+    sum(product_availability.quantity_on_hand)                                   as quantity_on_hand,
+    sum(product_availability.quantity_available_to_promise)                      as quantity_available,
+
+    # Stuff we need to include now or wait for the materialized view
+    -- Negative item count
+    (SELECT SUM(CASE WHEN pai.quantity_on_hand < 0 THEN 1 ELSE 0 END)
+     FROM product_availability pai
+     WHERE pai.product_id = product_availability.product_id
+       AND pai.location_id = product_availability.location_id)                   as negative_item_count,
```
EWaterman, line 35: has the potential to be its own view but this works

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -1,40 +1,74 @@
...
+
+    -- Date last counted
+    (select date_counted
+     from product_count_history
+     where product_count_history.product_id = product_availability.product_id
+       and product_count_history.inventory_id = location.inventory_id) as date_last_count,
```
EWaterman, line 41: See my comment in product_count_history. Does this not have the potentialy to return multiple rows for cycle count?

grails-app/migrations/views/cycle-count-session.sql
jmiranda: I forgot to update the comment. I was just trying to get rid of the JOIN to cycle count request in this query because it changed the granularity of the results from the intended (one row per facility+product) to something else (one row per facility+product+cycleCountRequest).

grails-app/migrations/views/cycle-count-session.sql
awalkowiak: Ok, I realized that the `negative_item_count` has a separate select, so it should not be an issue. You can disregard the previous comment (unless these lines are already filtered out here and we get only records with `negative_item_count` being 0 in the final results).

jmiranda: I used the same logic that was in the original view definition.  
```
WHERE product_availability.quantity_on_hand > 0
```
I just moved it down into the HAVING clause to deal with the aggregate case. The previous version would have only returned product availability records with quantity on hand > 0 which seemed like it would have skewed the value. This solution basically says, we only care about the product and facility if the sum (quantity on hand) > 0 which I think is what we want. 

To provide an example, if we had 
```
Product 1, ABC123, Bin 001, 10
Product 1, ABC123, Bin 002, -10
```
The original view would have returned a cycle count session record with quantity 10.
The current version would not return a cycle count session.

I think the latter is more accurate but doesn't allow for investigation through the cycle count feature. The former version is less accurate and allows for investigation.

grails-app/migrations/views/cycle-count-session.sql
jmiranda: I would be fine removing it but just note that there's a requirement to return products with quantity on hand > 0. My point was for someone to investigate that requirement to help us interpret its intent. 

To be honest, I actually think this clause needs to be in the query against cycle count session, not in the view itself. But if the requirement states that we should ignore inventory items (not products) with quantity on hand <= 0 then the original view definition is "correct". It would just produce inaccurate aggregate data so it didn't make sense to me.

grails-app/migrations/views/product-cycle-count-history.sql
```diff
@@ -0,0 +1,13 @@
+-- Using a generic view name in case we decide we need to return more data from cycle count
+-- tables. However, this can be renamed if necessary.
+create or replace view product_cycle_count_history as (
+    select
+        location.inventory_id as inventory_id,
+        cycle_count_item.product_id as product_id,
+        max(cycle_count_item.date_counted) as date_counted
+    from cycle_count
+        join cycle_count_request on cycle_count.id = cycle_count_request.cycle_count_id
```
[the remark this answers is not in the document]
jmiranda: Hard to interpret what I was doing. 

Two possible explanations: 1) I did that to make sure we were only counting cycle counts that were initiated by a request. 2) I might have been returning the request ID in a previous version of the view. 

I'm ok with removing it but think (1) is a possible reason to keep it. However, I'm not wedded to it.

grails-app/migrations/views/product-receipt-history.sql
jmiranda: I think there was a time when it was, but perhaps not now.

grails-app/migrations/views/product-count-history.sql
```diff
@@ -0,0 +1,19 @@
+create or replace view product_count_history as
```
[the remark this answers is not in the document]
jmiranda: > are we double counting cycle count here?

I don't think it really matters at the moment. We're querying for all counts over time and then using that to get the aggregate. If we included a "count" metric on this view then yes we'd probably want to make the view a little more discerning.

> If we're guaranteed creating a product inventory transaction after each count, we could technically rely solely on product_physical_count_history, right?

Eventually yes. At the moment, I don't know. I wanted a solution that would work whether we decided to use product availability + adjustments or just adjustments. I still think it makes more sense to just use adjustments for the first version, but either way I don't think this impacts this because we're only using this data to get the max(date_counted).

grails-app/migrations/views/cycle-count-session.sql
jmiranda: So just return the cycle_count_request.status? 

The COALESCE was part of the original view definition so I just moved it down a level to this JOIN subquery. I think the intent here is to find out whether something has been counted ever vs recently vs it's due for count.

grails-app/migrations/views/cycle-count-session.sql
jmiranda: Good call. I think I intended for there to be GROUP BYs either in the lower view definitions or this subquery. But the group by got lost somewhere along the way.

jmiranda: Added group by to product_count_history

grails-app/migrations/views/cycle-count-session.sql
jmiranda: I'm not going to worry about this at the moment, but we can come back to it if it becomes an issue.

grails-app/migrations/views/cycle-count-session.sql
jmiranda: I'm doing making a change at the moment. Let me know how you guys think we should deal with it.

jmiranda: Same with this. Not going to do anything at the moment as I think this approach is more correct. But I'm willing to change it if we determine the requirements state something different.

jmiranda:
@awalkowiak @EWaterman I made one change that I thought was necessary. Let me know if there's anything else that we need to deal with before merging. We'll probably find some more issues or improvements as we go, but I can work on them as part of the cycle count session ranking ticket https://pihemr.atlassian.net/browse/OBPIH-7020.

awalkowiak:
@jmiranda I tested it out locally, and I am getting this issue while entering any CC page (querying this view) 

> Internal server error
In aggregated query without GROUP BY, expression #&NoBreak;1 of SELECT list contains nonaggregated column 'openboxes.location.inventory_id'; this is incompatible with sql_mode=only_full_group_by

jmiranda:
@awalkowiak I don't see that error myself so could you suggest where it might be happening or fix the issue yourself?

awalkowiak (approved):

PR #5078, opened 2025-02-28
Add option to print count and recount forms in XLS and PDF

awalkowiak:
**Link to GitHub issue or Jira tickets:**
[OBPIH-7021 Export spreadsheet with count form](https://pihemr.atlassian.net/browse/OBPIH-7021)
[OBPIH-7015 Print blind count form pdf](https://pihemr.atlassian.net/browse/OBPIH-7015)
[OBPIH-7022 Export spreadsheet with recount form](https://pihemr.atlassian.net/browse/OBPIH-7022)
[OBPIH-7016 Recount form in PDF](https://pihemr.atlassian.net/browse/OBPIH-7016)

**Do not squash, use the "rebase and merge" option.**


src/js/consts/mimeType.js
```diff
@@ -0,0 +1,6 @@
+const MimeType = {
+  XLS: 'application/vnd.ms-excel',
+  PDF: 'application/pdf',
```
EWaterman, line 3: would it be worth merging this with fileFormat given that (based on your code) it's requirement that they have the same keys? Could just have a single enum with multiple fields. I have no idea if this is even possible with javascript :sweat_smile:

src/js/hooks/cycleCount/useToResolveTab.jsx
```diff
@@ -323,9 +326,18 @@ const useToResolveTab = ({
...
-    console.log('print resolve form pressed');
+  const printResolveForm = async (fileFormat) => {
+    spinner.show();
+    const payload = {
+      requests: checkedCheckboxes.map((cycleCountRequestId) => ({
+        cycleCountRequest: cycleCountRequestId,
+        countIndex: 1, // We only ever allow for a single recount, so index is always 1.
+      })),
+    };
+    const response = await cycleCountApi.startRecount(payload, currentLocation?.id, fileFormat, { responseType: 'blob' });
```
EWaterman, line 337: how does `responseType: 'blob'` work? Is that how we specify PDF?

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -45,21 +48,98 @@ class CycleCountApiController {
         BatchCommandUtils.validateBatch(command, "requests")
         List<CycleCountDto> cycleCounts = cycleCountService.startCycleCount(command)
 
-        render([data: cycleCounts] as JSON)
+        if (!params.format) {
+            params.format = "json"
+        }
+
+        withFormat {
```
EWaterman, line 55: I thought you needed to set the "accept" header for this to work? (ex: `'Accept': 'application/pdf'`). But clearly I was wrong and you just need to set the "format" query param.

How is this all interacting with setting `responseType: 'blob'` on the frontend? And then with the fileDownload method with the MIME setting? (I left a similar comment on the frontend side)

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -45,21 +48,98 @@ class CycleCountApiController {
...
+        }
+
+        boolean isRecount = cycleCounts?.any { it.status == CycleCountStatus.INVESTIGATING.name() }
+
+        withFormat {
+            xls {
+                isRecount ? exportRecountXls(cycleCounts) : exportCountXls(cycleCounts)
+            }
+            pdf {
+                String facilityName = cycleCounts?.first()?.cycleCountItems?.first()?.facility?.name  ?: ""
```
EWaterman, line 104: this is fine, but it might be easier to get the facility from the URI

grails-app/views/cycleCount/_printCount.gsp
```diff
@@ -0,0 +1,130 @@
+<%@ page contentType="text/html;charset=UTF-8" defaultCodec="html" %>
```
EWaterman, line 1: I'm surprised to see a GSP. I'm assuming this is how we create the PDF?

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -45,21 +48,98 @@ class CycleCountApiController {
...
+        response.contentType = "application/vnd.ms-excel"
+        documentService.generateExcel(response.outputStream, data)
+        response.outputStream.flush()
+    }
+
+    def renderCountPdf(List<CycleCountDto> cycleCounts, String facilityName) {
+        renderPdf(
+                template: "/cycleCount/printCount",
+                model: [cycleCounts: cycleCounts, facilityName: facilityName, datePrinted: new Date()],
+                filename: "Count form.pdf"
```
EWaterman, line 125: do we usually use spaces in the file name? I usually do something like count-form.pdf or count_form.pdf

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -179,6 +179,54 @@ class CycleCountService {
         return csv
     }
 
+    List<Map> getCountFormXls(List<CycleCountDto> cycleCounts) {
+        def data = []
+        SimpleDateFormat dateFormat = new SimpleDateFormat(Constants.EXPIRATION_DATE_FORMAT)
```
EWaterman, line 184: You can use Constants.EXPIRATION_DATE_FORMATTER

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -179,6 +179,54 @@ class CycleCountService {
...
+                ]
+            }
+        }
+
+        return data
+    }
+
+    List<Map> getRecountFormXls(List<CycleCountDto> cycleCounts) {
+        def data = []
+        SimpleDateFormat dateFormat = new SimpleDateFormat(Constants.EXPIRATION_DATE_FORMAT)
```
EWaterman, line 206: same. EXPIRATION_DATE_FORMATTER

grails-app/views/cycleCount/_printCount.gsp
awalkowiak: Yeah, I didn't want to spend time figuring out how we could do it differently, than we used to, so I went with the existing approach

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
awalkowiak: I just gave the naming of the files that were provided as templates in the tickets

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
awalkowiak: I did not realize it was there. I thought this was a simple cycle count get endpoint without facility mixed in because the facility id is not even used there. To be fair you, could provide a different facility id and this endpoint still would work, so I prefer pulling it out from the actual cycle count.

src/js/hooks/cycleCount/useToResolveTab.jsx
awalkowiak: This is `axios`'s property, this needs to be set to be able properly download the file.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -45,21 +48,98 @@ class CycleCountApiController {
...
+            params.format = "json"
+        }
+
+        withFormat {
+            xls {
+                exportCountXls(cycleCounts)
+            }
+            pdf {
+                renderCountPdf(cycleCounts, command.facility.name)
+            }
```
alannadolny, line 61: Shouldn't that kind of thing be in the service?

grails-app/views/cycleCount/_printRecount.gsp
```diff
@@ -0,0 +1,147 @@
...
+                        <th colspan="3" class="b-l0 b-t0" style="padding-bottom: 10px;"></th>
+                    </tr>
+                    <tr>
+                        <th>${warehouse.message(code: 'cycleCount.binLocation.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.lotNumber.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.expirationDate.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.quantityCounted.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.difference.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.quantityRecounted.label')}</th>
+                        <th>${warehouse.message(code: 'cycleCount.comments.label')}</th>
```
alannadolny, line 118: we should use g:message

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -179,6 +179,54 @@ class CycleCountService {
...
+                        "Date Counted": ""
+                ]
+            }
+        }
+
+        return data
+    }
+
+    List<Map> getRecountFormXls(List<CycleCountDto> cycleCounts) {
+        def data = []
```
alannadolny, line 205: def :smiling_face_with_tear:

src/js/api/services/CycleCountApi.js
```diff
@@ -11,13 +11,15 @@ import apiClient from 'utils/apiClient';
 
 export default {
   createRequest: (payload, locationId) => apiClient.post(CYCLE_COUNT_REQUESTS(locationId), payload),
-  startCount: (payload, locationId) => apiClient.post(CYCLE_COUNT_START(locationId), payload),
-  startRecount: (payload, locationId) => apiClient.post(
-    CYCLE_COUNT_RECOUNT_START(locationId), payload,
+  startCount: (payload, locationId, format = null, config = {}) => apiClient.post(
+    CYCLE_COUNT_START(locationId, format), payload, config,
```
alannadolny, line 15: the arguments list getting huge, we should use it like that in this case:
```suggestion
  startCount: ({ payload, locationId, format = null, config = {} }) => apiClient.post(
    CYCLE_COUNT_START(locationId, format), payload, config,
```

src/js/components/cycleCount/toCountTab/CountStepHeader.jsx
```diff
@@ -17,13 +19,29 @@ const CountStepHeader = ({ printCountForm, next, save }) => (
...
+          label="react.cycleCount.printCountForm.label"
+          variant="primary-outline"
+          StartIcon={<RiPrinterLine size={18} />}
+        />
+        <div className="dropdown-menu dropdown-menu-right nav-item padding-8" aria-labelledby="dropdownMenuButton">
+          <a href="#" className="dropdown-item" onClick={() => printCountForm(FileFormat.PDF)} role="button">
+            <Translate
+              id="react.cycleCount.printCountFormPdf.label"
+              defaultMessage="Print Count form PDF"
+            />
```
alannadolny, line 35: I would suggest using useTranslate, but that's not a change request

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -69,8 +72,17 @@ const useCountStep = () => {
...
+    show();
+    const response = await cycleCountApi.getCycleCounts(
+      currentLocation?.id,
+      cycleCountIds,
+      fileFormat,
+      { responseType: 'blob' },
+    );
+    const filename = extractFilenameFromHeader(response.headers['content-disposition']);
+    fileDownload(response.data, filename, MimeType[fileFormat]);
+    hide();
```
alannadolny, line 85: Can't we use here `exportFileFromApi` from `file-download-util.jsx`?

grails-app/views/cycleCount/_printRecount.gsp
awalkowiak: Right... I keep forgetting about this one.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
awalkowiak: No, this should be here, you can check it out in the docs here: https://grails.github.io/legacy-grails-doc/3.2.0/ref/Controllers/withFormat.html

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
awalkowiak: Added few more details there

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
EWaterman: we discussed this in Slack, but I'm happy with the solution you have here using the "format" param.

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
EWaterman: yeah that's fair. We haven't really discussed in depth the implications of prefixing all our APIs with `/api/facilities/$facilityId/`. To me if we're requiring it in the URI we should be using it in validation (and if we're not, we shouldn't have it in the uri). The idea is probably that we don't want people to be able to fetch data from other facilities that they shouldn't have access to.

But again we haven't really discussed this so I agree with you like this is fine for now.

alannadolny (approved):

PR #5079, opened 2025-03-01
OBPIH-7017 Import ABC Class in inventory levels

jmiranda:

https://pihemr.atlassian.net/browse/OBPIH-7017

**Description:**
This PR refactors the current inventory level importer (with Kelsey and Manon's blessing). The existing importer was written a long time ago and tried to do too much (update product, create product packages, create or update inventory levels, etc). This refactoring focuses on the facility-level inventory level rules. In fact, I've temporarily added logic to skip records that are pinned to internal locations (aka bin replenishment rules) to prevent us from unwittingly overriding these rules (i.e. these rules should not have ABC class assigned to them). 

Here's the current state of the PR

- [x] Allow user to import (update or create) an inventory level with an ABC class
- [x] Allow user to export inventory level data in a format compatible with the import feature
- [x] Test whether the other fields (particularly the fields that reference internal locations) are working properly
- [x] Allow deletion of duplicate inventory levels using InventoryLevelStatus INACTIVE (or a new status if that is currently used)
- [x] Allow user to clear replenishment bin and preferred bin (currently we’re going to require that the user to delete the rule altogether to reset)
- [x] Ensure inventory level lookup takes internal location into account (otherwise we might update the wrong inventory level)
- [x] Ensure we’re not overriding (removing) data that exists when cells are empty (we can add a string like NULL to signify that we want to delete)
- [x] Temporarily add logic to skip bin replenishment rules (any row with internal location set) so we can avoid overriding the wrong inventory level record
- [x] Allow export all inventory level records for a single facility
- [x] Test performance when uploading 100-1000ss of inventory level records at once
- [x] Detect and prevent import of duplicate inventory level records

Next steps 

- [ ] Suppress product availability refresh (this isn't a problem at the moment because it's not triggering a SQL statement)
- [ ] Fix DataServiceTests, add tests for new classes 
- [ ] Add subtask to allow users to export inventory levels for a single location 
---

Currently working [OBPIH-7017 Import ABC Class in inventory levels](https://app.screencastify.com/v3/watch/0t6LfxB7hw2d9xL5XHjG) 

Second video demonstrates [Untitled Video February 28, 2025 11:22 PM](https://app.screencastify.com/v3/watch/5ppblULVkG4WXz1oYliq)


grails-app/controllers/org/pih/warehouse/inventory/InventoryLevelController.groovy
```diff
@@ -12,13 +12,17 @@ package org.pih.warehouse.inventory
...
+import org.pih.warehouse.product.ProductService
 
 @Transactional
 class InventoryLevelController {
 
-    def productService
-    def dataService
+    InventoryLevelImportDataService inventoryLevelImportDataService
+    ProductService productService
+    //DataService dataService
```
alannadolny, line 25: if it not used, you can remove it :smile:

grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
```diff
@@ -120,4 +122,27 @@ class InventoryLevel {
...
+            abcClass             : "abcClass",
+            minQuantity          : "minQuantity",
+            reorderQuantity      : "reorderQuantity",
+            maxQuantity          : "maxQuantity",
+            forecastQuantity     : "forecastQuantity",
+            forecastPeriodDays   : "forecastPeriodDays",
+            "Unit of Measure"    : "product.unitOfMeasure"
+            //expectedLeadTimeDays   : "expectedLeadTimeDays",
+            //replenishmentPeriodDays: "replenishmentPeriodDays",
+            //preferredForReorder    : "preferredForReorder",
```
alannadolny, line 145: Why is it commented?

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -10,27 +10,217 @@
...
+     * Import a single row from the XLS file
+     *
+     * @param location
+     * @param fileName
+     * @return
+     */
+    def importInventoryLevel(Map params, Integer index) {
+        log.info "Import inventory levels(${index}): " + params
+
+        def product = Product.findByProductCode(params.productCode as String)
```
alannadolny, line 68: I think it's type of Product

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -10,27 +10,217 @@
...
+            product.save()
+        }
+    }
+
+    Location resolveInternalLocation(Location facility, String internalLocationName) {
+        Location internalLocation = facility.getInternalLocation(internalLocationName)
+        if (!internalLocation) {
+            throw new IllegalArgumentException("Unable to locate internal location ${internalLocationName} within facility ${facility?.name}")
+        }
+        return internalLocation
```
alannadolny, line 117: Can't we handle the lack of `params.internalLocation` here instead of checking it three times when assigning to:
```
params.internalLocation = params.internalLocation ? resolveInternalLocation(facility, params.internalLocation as String) : null
params.preferredBinLocation = params.preferredBinLocation ? resolveInternalLocation(facility, params.preferredBinLocation as String) : null
params.replenishmentLocation = params.replenishmentLocation ? resolveInternalLocation(facility, params.replenishmentLocation as String) : null
```

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -10,27 +10,217 @@
...
+     * @param inventory
+     * @param preferredBinLocation
+     * @param minQuantity
+     * @param reorderQuantity
+     * @param maxQuantity
+     * @return
+     */
+    def findOrCreateInventoryLevel(Location facility, Product product) {
+        log.info "Product ${product.productCode} facility ${facility}"
+        def inventoryLevel = InventoryLevel.findByInventoryAndProductAndBinLocationIsNull(facility.inventory, product)
```
alannadolny, line 169: that's type of InventoryLevel

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -10,27 +10,217 @@
...
+
+    /**
+     * Export the given inventory levels
+     *
+     * @param inventoryLevels
+     * @return
+     */
+    String exportInventoryLevels(Collection inventoryLevels) {
+        def sw = new StringWriter()
+        def csv = new CSVWriter(sw, {
```
alannadolny, line 186: those also have types

grails-app/services/org/pih/warehouse/requisition/RequisitionService.groovy
```diff
@@ -977,4 +981,146 @@ class RequisitionService {
...
     }
+
+    /**
+     * Export the given requisitions to CSV.
+     *
+     * @param requisitions
+     * @return
+     */
+    String exportRequisitions(requisitions) {
+        def formatDate = new SimpleDateFormat("dd/MMM/yyyy hh:mm:ss")
```
alannadolny, line 992: image

alannadolny (commented):
Is there any benefit of using defs everywhere instead of defining a type?

grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
jmiranda: Just deciding on whether we needed to keep it or not.

grails-app/controllers/org/pih/warehouse/inventory/InventoryLevelController.groovy
jmiranda: Yeah I think I commented it out because as part of the refactor and still needed to see whether it was needed. Thanks for the reminder.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
jmiranda: This is used to "cast" the value returned from the params map (that was populated by the excel importer). For some reason, the params map always returns an Object instead of the type we've defined in the property map (i.e. String, Integer, Date). So using "as" converts it to the expected type. I wish I didn't have to do it, but I suspect our beef is with the AbstractExcelImporter.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
jmiranda: Can you provide an example for what you mean? Something like this?
```
params.internalLocation = resolveInternalLocation(facility, params.internalLocation as String)
```

jmiranda:
> Is there any benefit of using defs everywhere instead of defining a type?

I'm not sure exactly which "def" you were referring to but I copied existing code (written under Grails 1.3.x) where "def" was more commonly used because using a type sometimes caused problems with hot-reloading and other times with classloading. Going forward, we should definitely try to use def so if you see specific examples definitely point them out.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
jmiranda: Sorry, just realized this was one of the places where "def" was used. Yeah this was existing code that I forgot to change. Will push a fix later.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
```diff
@@ -10,27 +10,217 @@
...
+        //  this check and ensure that the following code handles these rules properly (i.e. most of the time the
+        //  bin replenishment only needs internal location, min quantity, reorder quantity, max quantity). Therefore,
+        //  we should create separate command classes to validate the two types of inventory levels.
+        if (params.internalLocation) {
+            log.info "Skipping update to bin replenishment rule ${params}"
+            return
+        }
+
+        // Resolve any objects that require a database lookup
+        params.internalLocation = params.internalLocation ? resolveInternalLocation(facility, params.internalLocation as String) : null
```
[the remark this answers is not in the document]
jmiranda: Can you elaborate on the question a bit?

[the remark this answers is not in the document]
awalkowiak: Right above you have:
```
if (params.internalLocation) {
    log.info "Skipping update to bin replenishment rule ${params}"
    return
}
```
And wrote that we are skipping it for now, so this line here is not needed (unless you want to keep it for future but then perhaps is worth giving a comment here too, because someone else might be confused later).

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
alannadolny: > this was existing code

That's my bad, I didn't notice that

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
alannadolny: I meant adding 
```
if (internalLocationName == null) {
   return null
}
```
at the beginning of the resolveInternalLocation method, so you don't have to check it while assigning to params.XYZ

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
jmiranda: No worries, you can/should still call it out. It's probably best to "boy scout" these kinds of refactorings if time permits.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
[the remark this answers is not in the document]
jmiranda: Oh good call. I think it's fine to leave it for the future, but a comment would be helpful.

grails-app/services/org/pih/warehouse/importer/InventoryLevelImportDataService.groovy
jmiranda: Yeah that's what i was thinking you meant. It feels weird to deal with that in the resolve method since it seems to be the API client's responsibility to decide how it wants to deal with that case. A better approach might be to pass a "default" value to the resolve method, so I'll consider doing that.

awalkowiak (approved):

PR #5088, opened 2025-03-05
OBPIH-7003 Up/down arrow to navigate through cells in quantity column (fix after QA)

SebastianLib:

**Description:**

---


awalkowiak (approved):

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -157,7 +157,9 @@ const useCountStepTable = ({
       // on change function expects e.target.value for text fields,
       // in other cases it expects just the value
       const onChange = (e) => {
-        setValue(e?.target?.value ?? e);
+        const enteredValue = e?.target?.value ?? e;
+        table.options.meta?.updateData(cycleCountId, original.id, id, enteredValue);
+        setValue(enteredValue);
```
alannadolny, line 162: Isn't it causing performance issues? We are updating the table data for each letter.

alannadolny (commented):
I am a bit worried that the code got really complicated, so it does not look like an easy-to-apply hook at this moment. :/

awalkowiak (commented):
Could you reiterate why is there a need for this `focusCounter`? Additionally, some conflicts popped up

awalkowiak (commented):
Additionally, could you rename `setFocusId` and `setFocusIndex` to indicate that one is column related and other one row related?

awalkowiak (commented):
Overall I don't like the idea of having to reset focus when clicking on everything (this looks like an easy way to forget and introduce bugs). But if we don't have a better solution right now, I guess we can go with this one and think about a better approach without time pressure.

@alannadolny @kchelstowski do you have any thoughts about this?

PR #5092, opened 2025-03-05
OBPIH-7010 add reload button in recount flow to refresh QoH

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7010

**Description:** Added the "Reload" button to the recount page that does the following:

- if QoH for the [product + bin location + lot number] has changed, update it in the count item
- if QoH for the [product + bin location + lot number] is now 0, remove the count item (unless it's a custom row, in which case we set QoH for the item to 0)
- if a new [product + bin location + lot number] has been added, create a new count item for it

---

Screenshot from 2025-03-07 10-35-20

https://jam.dev/c/d85f7766-e873-471d-9aba-1539224eac2a

The video is a little confusing but prior to the recording I did the following:
- created the "TESTING" lot with QoH == 36
- added 500+ quantity to the existing "No lot OMronM2" lot

you can see that after the refresh, lot "TESTING" was added to the count and "No lot OMronM2" had a much larger count difference. (There are duplicate rows for some reason but this appears to be unrelated to my change so I didn't fix.)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -228,37 +228,47 @@ class CycleCountService {
...
     /**
-     * Batch method to start a cycle count.
-     * Fetch associated Cycle count with the CycleCountRequest and do the following:
-     * if a cycle count doesn't exist, create it, fetch all bin locations + inventory items for the product in a request,
-     * get the product availability for each item, filter out items that have qoh = 0 (but keep those that have qoh = 0 and were included in the previous count)
-     *
-     * */
+     * Batch start multiple cycle counts.
+     */
     List<CycleCountDto> startCycleCount(CycleCountStartBatchCommand command) {
```
EWaterman, line 233: I did a bit of cleanup on the start cycle count API while I was reworking the product availability flow. Mainly I removed some outdated validation (the updateCycleCount method is not needed anymore) and moved other validation logic to the command object instead like we do for recount.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -390,33 +374,31 @@ class CycleCountService {
...
+    /**
+     * Submits the (re)count as it is in its current state.
+     */
     CycleCountDto submitCount(CycleCountSubmitCountCommand command) {
-        command.cycleCount.cycleCountItems.each { CycleCountItem cycleCountItem ->
-            updateCycleCountItemForSubmit(cycleCountItem, command.refreshQuantityOnHand, command.failOnOutdatedQuantity)
+        CycleCount cycleCount = command.cycleCount
+
+        if (command.refreshQuantityOnHand) {
+            boolean hasQuantityChanged = cycleCountProductAvailabilityService.refreshProductAvailability(cycleCount)
```
EWaterman, line 384: we now do the QoH refresh outside of the cycleCountItems loop below. We can't just loop the items of the count because there might be new items created that the count is not aware of.

grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
```diff
@@ -42,42 +43,34 @@ class CycleCountTransactionService {
...
-        // needlessly creating the transaction itself if there are no discrepancies.
+
+        // We need to compare the quantity counted against the most up to date QoH in product availability.
+        // However, if the cycle count items already have an up to date QoH (which will be the case if
+        // refreshQuantityOnHand == true when submitting the count), we don't need to re-compute it.
+        if (!itemQuantityOnHandIsUpToDate) {
+            // While the final QoH of the product for each [bin location + lot number] will always be the same value
+            // as was counted, if the QoH in the cycle count items changes at this step, the amount of quantity
+            // adjustment/change might be different than what was displayed to the user during the count.
+            cycleCountProductAvailabilityService.refreshProductAvailability(cycleCount)
```
EWaterman, line 54: same here. We now refresh QoH outside of the loop on cycle count items because there may be new items that the count is not aware of

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -33,7 +34,7 @@ class CycleCount {
         status = recomputeStatus()
     }
 
-    SortedSet cycleCountItems
+    SortedSet<CycleCountItem> cycleCountItems
```
kchelstowski, line 37: Have you made sure the items are still sorted properly? When I was helping Sebastian with that, initially I had this Set typed, but if I remember correctly the sorting was not working. In the docs they do not set type for `SortedSet`, I know this should be working with the type, but maybe it's Grails' bug that it doesn't. 
Please make sure the sorting is still working when adding the type.

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -0,0 +1,139 @@
...
+     * Note that any items that are not yet added to the cycle count will be added to the count here, and any existing
+     * (non-custom) items whose QoH has become 0 will be removed from the count. As such, this method WILL mutate
+     * the cycle count items.
+     *
+     * @return true if a new cycle count item was added to the count or an existing one had their QoH modified.
+     */
+    private boolean updateItem(CycleCount cycleCount, Set<CycleCountItem> currentCountItems,
+                               AvailableItem availableItem, Location facility, int currentCountIndex,
+                               User assigneeForNewItems, Date dateCountedForNewItems,
+                               CycleCountItemStatus statusForNewItems) {
```
kchelstowski, line 85: I would suggest creating a class for this amount of arguments being passed to a method. I think we once agreed that 3-4 arguments should be the maximum.

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -0,0 +1,139 @@
...
+                CycleCountItemStatus.COUNTING : CycleCountItemStatus.INVESTIGATING
+
+        // The refresh is based on the current product availability in all [bin location + lot] combinations for all
+        // the products of the count. We do this because new items might have been created since the count started.
+        List<AvailableItem> availableItems = getAvailableItems(facility, cycleCount.products)
+        for (AvailableItem availableItem : availableItems) {
+            boolean itemHasChanged = updateItem(cycleCount, cycleCountItems, availableItem, facility, currentCountIndex,
+                    assigneeForNewItems, dateCountedForNewItems, statusForNewItems)
+
+            itemsHaveChanged = itemsHaveChanged || itemHasChanged
```
kchelstowski, line 66: I would add a comment explaining what is happening here. I had to think a longer time what it is doing.
I assume if the first item is changed, the flag is set to `true`, and then we don't even have to check further items in terms of `itemHasChanged`.

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -0,0 +1,139 @@
...
+        Date dateCountedForNewItems = cycleCountItems.find{ it.dateCounted }?.dateCounted ?: new Date()
+
+        // This logic is shared between count and recount flows so we need to check which we're in.
+        CycleCountItemStatus statusForNewItems = cycleCount.status.isCounting() ?
+                CycleCountItemStatus.COUNTING : CycleCountItemStatus.INVESTIGATING
+
+        // The refresh is based on the current product availability in all [bin location + lot] combinations for all
+        // the products of the count. We do this because new items might have been created since the count started.
+        List<AvailableItem> availableItems = getAvailableItems(facility, cycleCount.products)
+        for (AvailableItem availableItem : availableItems) {
```
kchelstowski, line 62: (a side thought): what if an available item was deleted for a particular inventoryItem + bin? Is it even possible? :thinking: 
if it is, then shouldn't we actually loop through `cycleCountItems` and find matching `availableItem` for an item?
I'm thinking about a situation where you have 5 cycleCountItems and suddenly only 4 matching availableItems - for this scenario, we wouldn't even "touch" one of the cycleCountItems for which the stock might not even exist.
Sounds complicated so let me know if you need a further clarification.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -354,33 +365,6 @@ class CycleCountService {
...
-            throw new UnsupportedOperationException("Support will be added later")
-        }
-        if (cycleCount.status in CycleCountStatus.listInProgress()) {
-            CycleCountItem itemWithHighestCountIndex = cycleCountItems.max { it.countIndex }
-            if (itemWithHighestCountIndex?.countIndex != request.countIndex) {
-                throw new IllegalArgumentException("Count index can't be higher than the highest count index of the items")
-            }
-        }
-        return CycleCountDto.toDto(cycleCount)
-    }
```
kchelstowski, line 371: what happened with that method? I can't find where it has been moved or why it has been removed

src/main/groovy/org/pih/warehouse/inventory/CycleCountStartCommand.groovy
```diff
@@ -5,4 +5,17 @@ import grails.validation.Validateable
...
 
     CycleCountRequest cycleCountRequest
+
+    static constraints = {
+        cycleCountRequest(validator: { CycleCountRequest cycleCountRequest ->
+            CycleCount cycleCount = cycleCountRequest.cycleCount
+            if (!cycleCount) {
+                // When first starting a count, the cycle count object won't exist yet, so this is valid.
+                return true
+            }
```
kchelstowski, line 15: you don't need to check for valid cases, by default this would return `true`, so this check is unnecessary
Not a change request but if you were to remove this check, remember to add null safe operators or handle it somehow else below

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
jmiranda: @kchelstowski Thanks for bringing that up.

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
jmiranda: Just create a command class something like UpdateOrCreateCycleCountItemCommand.

jmiranda: You could also potentially break this into multiple methods (find or create vs detect changes vs update).

jmiranda: The method that detect changes could also return data (or a command) to tell you what operation needs to be executed: UpdateCycleCountItemCommand, DeleteCycleCountItemCommand, etc.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
jmiranda: suggestion: It would be cool to separate the "detect changes" from the "apply changes" in case we want show the user what's about to happen. Detect changes could return a list of changes that could be displayed to the user.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -390,33 +374,31 @@ class CycleCountService {
...
+     * Submits the (re)count as it is in its current state.
+     */
     CycleCountDto submitCount(CycleCountSubmitCountCommand command) {
-        command.cycleCount.cycleCountItems.each { CycleCountItem cycleCountItem ->
-            updateCycleCountItemForSubmit(cycleCountItem, command.refreshQuantityOnHand, command.failOnOutdatedQuantity)
+        CycleCount cycleCount = command.cycleCount
+
+        if (command.refreshQuantityOnHand) {
+            boolean hasQuantityChanged = cycleCountProductAvailabilityService.refreshProductAvailability(cycleCount)
+            if (hasQuantityChanged && command.failOnOutdatedQuantity) {
```
jmiranda, line 385: Right now, the failOn is probably up to the frontend developer but it would be good to wire this into a feature flag that can be used to set this value on the frontend. i.e. instead of hard-coding this in React, we just pull from a the feature flag from the AppContext API.

grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
```diff
@@ -42,42 +43,34 @@ class CycleCountTransactionService {
 
     private Transaction createAdjustmentTransaction(
             CycleCount cycleCount, Date transactionDate, boolean itemQuantityOnHandIsUpToDate) {
-        // We only care about the cycle count items from the most recent count. We create the entries first to avoid
-        // needlessly creating the transaction itself if there are no discrepancies.
+
+        // We need to compare the quantity counted against the most up to date QoH in product availability.
+        // However, if the cycle count items already have an up to date QoH (which will be the case if
+        // refreshQuantityOnHand == true when submitting the count), we don't need to re-compute it.
+        if (!itemQuantityOnHandIsUpToDate) {
```
jmiranda, line 50: Shouldn't we recompute whether this is true instead of relying on an argument? Or is that what's happening within the caller when this is invoked?

grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
jmiranda: > While the final QoH of the product for each [bin location + lot number] will always be the same value as was counted

I'm struggling to understand this comment. Is this just saying the final quantity counted (i.e. the quantity entered into the cycle count item) will be the same as the value counted in the physical world? Or is "final QoH" referring to the QoH computed and stored in product availability?

I think I get the concept you're trying to lay out, though. Correct me if I'm wrong.

* The user start the count, QoH (product availability) is copied to cycle count item (X)
* The user counts the cycle count item in the physical world (puts the value on paper) (Y)
* The user enters the quantity counted (Y) into the cycle count item
* The user submits the count for review (X and Y are saved to cycle count item)
* The system computes the quantity variance (Z = X - Y) 
* The system displays the quantity variance (Z) to the user
* The user refreshes the cycle count which might update the quantity variance  (Z<sub>2</sub> = X<sub>2</sub> - Y)
* The system displays this new quantity variance to the user but the user doesn't not necessarily see how the value changed (before/after refresh)

jmiranda (commented):
Looks good. I would just try to be defensive about scenarios we aren't explicitly handling. For example, the one Kacper brought up but more generally when cycle count items and product availability don't match on more than just quantity). Once we determine how to handle these scenarios we can add them in, but until then failing fast would be recommended.

grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
EWaterman: yeah it's happening in the caller. This will be true if command.refreshQuantityOnHand == true when submitting a count. We trust that the caller already did the refresh (cycleCountService.submitCount calls `cycleCountProductAvailabilityService.refreshProductAvailability(cycleCount)`) so that we can avoid doing it again here. It's a performance saving thing.

There isn't really another way of determining if this is true since to do so would require fetching product availability again (which is what we're trying to avoid).

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: yeah I was actually considering doing something like this originally but it got kinda complex (I needed three lists, delete, create, update). I'll give it another think through again though

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -987,6 +987,11 @@ class UrlMappings {
             action = [POST: "createCycleCountItem"]
         }
 
+        "/api/facilities/$facility/cycle-counts/$cycleCountId/items/refresh" {
```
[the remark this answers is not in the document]
EWaterman: hmm an interesting idea. I don't know if I like having a generic operation handler API (how would we handle different request payloads? It seems like in the best case we'd end up with a huge switch statement) but I do think we should revisit our uri structure and the colon structure is intriguing.

I made it /items because it's an operation on the items, not the count (and we could have a `/items/$cycleCountItemId/refresh` API that refreshes just a single row/item if we really wanted) but I suppose we don't need to make that distinction when refreshing all the items of a count.

EWaterman:
I refactored the CycleCountProductAvailabilityService. The logic is the same but now it uses a CycleCountRefreshState context object as per suggestions in review comments. I also have the method returning CycleCountItemsForRefresh which contains lists of the items that were created, updated, and deleted, as a part of the QoH refresh. I think it looks cleaner now.

EWaterman:
@awalkowiak I definitely should have split the refactor into its own PR.

The main thing here is the CycleCountProductAvailabilityService, which I have unit tests for so I feel pretty comfortable that it is working. Before I merge this I'm going to retest the other flows that I refactored as thoroughly as I can to make sure nothing new is broken. I'll only merge end of day if I can confirm that

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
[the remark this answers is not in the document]
jmiranda: Sorry, I didn't make that more explicit. 

The batchOperation was meant to show that we could use a variable for the action handler as opposed to defining multiple URL mappings that looks the same.
```
"/api/facilities/$facility/cycle-counts/$cycleCountId:$batchOperation" {
    controller = "cycleCountApi"
    action = [POST: batchOperation]
}
```
Therefore, we would implement the actions we want to support as separate methods.
```
def refreshCycleCount(RefreshCycleCountCommand) { }
def doThatThing(DoThatThingCommand command) { }
def doThatOtherThing(DoThatOtherThingCommand) { }  
```

The main design element would just be changing from a RPC endpoint (which looks like a REST resource but isn't).
```
/api/facilities/$facility/cycle-counts/$cycleCountId/refresh
```
to an action-oriented with a different naming convention
```
/api/facilities/$facility/cycle-counts/$cycleCountId:refresh
```
Functionally equivalent, but I really want to avoid RPC endpoints. So we either need to come up with a subresource where a refresh could be implemented using REST or we support these non-REST actions using a non-REST naming convention.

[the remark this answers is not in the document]
EWaterman: hm interesting. So essentially we have a /resource:action structure where we're making a distinction between the resource we're acting on and the action we're performing. Yeah I do like that.

I'm going to leave it as /$cycleCountId/refresh for now until we can think it through more thoroughly, but I do think a structure like $cycleCountId:refresh that is more explicit about what is the resource and what is the action could be quite useful.

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
jmiranda: > That said, I do think having the ability to recover from transactions happening isn't a bad thing. 

I completely agree. I think this feature is going to be game-changing. 

Apologies, this got longer than I was hoping. We can discuss on a call if you see any merit to continuing this conversation. 

---

My main point is that while we might think we're covering our bases by implementing these three generic cases:

- same: there are the same number of cycle count items as product availability items
- more: there are more cycle count items than product availability items
- less: there are less cycle count items than product availability items

... it's possible that we might come to each of these three cases in more ways than we understand and it might also be possible that we might need to handle each case differently based on the specific use case that caused the state we're in. 

And I'm not saying that we're not covering all of these use cases properly at the moment. I'm saying that I don't know for sure because it's complex. There are a lot of in-betweens that make this more complicated than "there are transactions happening while we're cycle counting". 

For example, when a user is picking an order, the quantity picked might be anywhere in the warehouse (even if the system thinks it's in the bin location). It could be in the bin location, in-transit to the packing location, at the packing station, in a staging area, loaded onto a truck, etc. The system will always think the quantity being picked is still in the bin location so the inventory item will still be in product_availability (at the bin location) until it is shipped. If the quantity picked is anywhere else but the bin location then we might encounter a double debit (quantity picked, quantity adjusted) when the cycle count is completed. 

The opposite is true when someone is putting an inventory item away. They could physically move the stock to the putaway location but they haven't walked back to the computer to enter the putaway into the system yet. So the counter would likely have double count stock. 

I know we've thought a lot about these scenarios. So again, I'm not saying we're not handling these cases properly. I think the idea is to deal with these issues in the resolution process i.e. go off and investigate so that we understand where the stock might have gone. But that assumes that we can actually figure out where all of the inventory items went and explain discrepancies well enough to resolve them all without making a mistake. 

But mistakes happen. Someone could see that a count variance can be explained by an outbound pick (because the quantity variance matches the quantity to pick) so they resolve the issue as "item was picked and moved to staging area" when it hadn't been picked yet. After the count is done another user goes and picks the order and we find ourselves in a double debit situation. 

These are the easy cases so we'll probably handle them fine. But I'm concerned about scenarios that we haven't thought about or encountered yet. 

TL;DR: Thinking that we can homogenously handle each case (e.g "there are the same/less/more cycle count items than product availability items") feels dangerous to me.

EWaterman: that's a fair point. The system only matches the inventory items as they are in the app, so any user error that causes OB to not reflect reality will still result in inaccuracies. But that's essentially outside of our control. We only know what OB knows, and so all we can do is stay in sync with that. Using your example, if a warehouse has a process that moves stock before it has been picked in OB (and don't document anywhere physically that this has happened), then someone does a count, then the first person puts the pick in OB, well that's probably a bad process that needs fixing and retraining.

The way I see it, any solution will always be stuck having to work in tandem with training users in communication and in best practices of using the app (and processing stock movements in general).

I get that we want to be flexible to however they want to use the app, but some practices, like the example I said above, fundamentally introduce confusion into the system. And that could be fixed in a number of ways. Maybe they require anything that is picked to always immediately get put in OB, maybe they have a physical ledger of pending picks that they can reference when counting to make sure they don't double deduct the stock if they find it missing.

Another example could be someone input a stock transfer in OB but didn't move the physical stock yet. Another user refreshes the count, sees a new bin has been added, goes to count it, finds nothing so puts 0, then we've wiped out the whole bin. But that's the same indication to me of a process failure.

I do get your point, but I see these types of inaccuracies as an indication of a bad practice and failure of communication.

Either way, if someone misunderstands a situation and inputs inaccurate data, the hope is that they later realize what happened, and a new cycle count can be initialized to fix mistakes. Or if they don't realize, then at least it will get caught the next time a cycle count on the product is automatically scheduled.

jmiranda: > Maybe they require anything that is picked to always immediately get put in OB,

That is exactly what needs to happen. We built OB with a very low volume warehouse use case in mind. But now that we're doing cycle counts and allowing transactions to occur during the cycle count, that means we need to be more fine-grained with the transactions. Whether stock is being picked into a "tote" and taken to an outbound packing location, being loaded onto a truck, or is on a forklift heading from the receiving bay to a putaway location ... each of these micro transactions should be tracked in OB. 

This is why I've been advocating for handheld workflows. These activities (receipt, putaway, pick, pack, load, etc) are all internal transactions that should be recorded in the system to have precise details about where stock exists.

This may never happen with PIH, but for high volume warehouses this is going to be essential.

jmiranda: > maybe they have a physical ledger of pending picks that they can reference when counting to make sure they don't double deduct the stock if they find it missing.

Yes, a lot of these locations have physical bin cards that can be used to understand the flow. It just gets more complicated to unravel and understand the edge cases as more actors (picker, counter) and systems (OB, paper bin card) get introduced into a scenario.

PR #5096, opened 2025-03-06
OBPIH-7050 Replace mocked data with real data in Resolve workflow

alannadolny:


src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -55,7 +55,7 @@ const useCountStep = () => {
         currentLocation?.id,
         cycleCountIds,
       );
-      tableData.current = data?.data;
+      tableData.current = data?.data.filter((cycleCount) => cycleCount.countIndex !== 0);
```
awalkowiak, line 58: I think this should be `===`, I think on count step you want to see only count index 0 items.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -43,17 +43,41 @@ const useResolveStep = () => {
     currentLocation: state.session.currentLocation,
   }));
 
+  const mergeCycleCountItems = (items) => {
+    const duplicatedItems = _.groupBy(items,
+      (item) => `${item.binLocation?.id}-${item?.inventoryItem?.lotNumber}`);
+    return Object.values(duplicatedItems).map((itemsToMerge) => {
+      const maxCountIndex = _.maxBy(itemsToMerge, 'countIndex').countIndex;
```
awalkowiak, line 50: Perhaps this is worth moving to CycleCountDto? CycleCount already has `maxCountIndex` (this is not a change request, but a thought)

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -43,17 +43,41 @@ const useResolveStep = () => {
...
+      const itemFromCount = _.find(itemsToMerge, (item) => item.countIndex === maxCountIndex - 1);
+      const itemFromResolve = _.find(itemsToMerge, (item) => item.countIndex === maxCountIndex);
+      return itemFromCount ? {
```
awalkowiak, line 53: I think you always will have `itemFromCount`, see effectively you won't land in the `else` in this ternatry operator

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -43,17 +43,41 @@ const useResolveStep = () => {
...
+        ...itemFromResolve,
+        ...itemFromCount,
```
awalkowiak, line 55: If these `mergeCycleCountItems` will be used in the `save` action, then we might always get the `quantityCounted` from the count item instead of the recount item (unless later `quantityRecounted` is used for building request payload, then it is ok and you can disregard this comment).

src/js/hooks/cycleCount/useCountStep.jsx
alannadolny: I tried: 
```
console.log(data?.data.filter((cycleCount) => cycleCount.countIndex !== 0));
```
and the output is:
image

src/js/hooks/cycleCount/useResolveStep.js
alannadolny: yeah, there will be a need for building a new payload.

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: @alannadolny it's because you are checking `cycleCount` status not the status of items for each cycle count.

awalkowiak (approved):

PR #5105, opened 2025-03-07
OBPIH-7018 Draft unfinished count workflow

kchelstowski:

**Description:**

---


SebastianLib (approved):

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -327,6 +327,7 @@ const useCountStep = () => {
       }
       await submitCount();
     } finally {
+      dispatch(eraseDraft());
```
EWaterman, line 330: what if submit fails? We probably don't want to erase the draft since it's still in progress.

Not an essential change, but is there any easy way to only erase the draft if we actually succeed in submitting the count?

src/js/reducers/cycleCountReducer.js
```diff
@@ -34,6 +35,11 @@ export default function cycleCountReducer(state = initialState, action) {
         ...state,
         binLocations: action.payload,
       };
+    case ERASE_DRAFT:
+      return {
+        ...state,
+        requests: [],
+      };
```
EWaterman, line 42: I'm totally fine with this solution, but is using an action required here? All we're doing is clearing the requests prop. Seems like chaining that through the ERASE_DRAFT action is a bit redundant. I'm far from an expert in React though, so doing it as you have it might be more correct.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -8,7 +8,7 @@ SELECT
     product_availability.location_id                                                 as facility_id,
 
     -- Cycle count request ID
-    cycle_count_request_summary.cycle_count_request_id                               as cycle_count_request_id,
+    cycle_count_request_summary.cycle_count_request_id                              as cycle_count_request_id,
```
EWaterman, line 11: was this intended?

EWaterman (approved):

awalkowiak (approved):

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: yeah, you are right. Actually you would still have the access to the data if you don't refresh the page, but to be safe, I moved it to the `try` clause, so that the draft is only erased if we successfully submit.

src/js/reducers/cycleCountReducer.js
kchelstowski: It's how Redux's flow is working, so it's not redundant even though it might look like this.

grails-app/migrations/views/cycle-count-session.sql
kchelstowski: nope, I was adding a workaround to make the view working after Justin's changes and after removing it, one of the indents stayed there. I removed it.

PR #5106, opened 2025-03-07
OBPIH-7020 Implement cycle count session ranking using last counted date

jmiranda:

https://pihemr.atlassian.net/browse/OBPIH-7020

**Description:**
There's a thorough discussion in the ticket, along with links to the requirements, design, and a few videos showcasing the change.

---

The latest video demonstrating the current version included in this PR.
https://app.screencastify.com/v3/watch/JjfrdsZNsbRFU9eYz22g


grails-app/migrations/views/cycle-count-frequency.sql
```diff
@@ -0,0 +1,27 @@
...
+      union all
+      select 'B' as abc_class, NULL as facility_id, 180 as frequency
+      union all
+      select 'C' as abc_class, NULL as facility_id, 365 as frequency
+      union all
+      -- We cannot use NULL here because we cannot JOIN on a NULL value.
+      select 'DEFAULT' as abc_class, NULL as facility_id, 365 as frequency
+  ) cycle_count_frequency
+);
+# -- FIXME Move this to a proper database migration if it actually helps with performance.
```
jmiranda, line 14: This will be done before the PR is merged.

grails-app/migrations/views/cycle-count-metadata.sql
```diff
@@ -0,0 +1,25 @@
...
+             product_count_history.date_counted,
+             datediff(now(), date_counted) as days_since_last_count,
+             -- FIXME Used to clean data, this should not be used in the actual solution
+             CASE WHEN abc_class = 'A' THEN 'A' WHEN abc_class = 'B' THEN 'B' WHEN abc_class = 'C' THEN 'C' ELSE 'DEFAULT' END as abc_class
+         FROM product_count_history
+                  LEFT OUTER JOIN product_classification
+                                  ON product_count_history.inventory_id = product_classification.inventory_id
+                                      AND product_count_history.product_id = product_classification.product_id
+     ) as cycle_count_metadata_tmp
+         LEFT OUTER JOIN cycle_count_frequency ON cycle_count_frequency.abc_class = cycle_count_metadata_tmp.abc_class;
```
jmiranda, line 25: This would benefit from a concrete table for cycle count frequency but still need to test if it's required.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -47,36 +47,48 @@ SELECT
...
+                                 max(days_until_next_count) as days_until_next_count
+                          FROM cycle_count_metadata
+                          GROUP BY inventory_id, product_id) as cycle_count_metadata
+                         on cycle_count_metadata.inventory_id = location.inventory_id and
+                            cycle_count_metadata.product_id = product_availability.product_id
 
 GROUP BY product_availability.location_id, product_availability.product_id, location.inventory_id
-HAVING sum(product_availability.quantity_on_hand) > 0
-ORDER BY NULL
+HAVING sum(product_availability.quantity_on_hand) > 0 and cycle_count_metadata.days_until_next_count <= 0
```
jmiranda, line 91: These two conditions need to be moved to the candidate query since this will cause the records to disappear from tabs other than All Products.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -47,36 +47,48 @@ SELECT
...
+ORDER BY cycle_count_metadata.abc_class IS NULL asc, cycle_count_metadata.abc_class asc,
```
jmiranda, line 92: Need to move the ORDER BY to the candidate query.

grails-app/migrations/views/cycle-count-metadata.sql
jmiranda: I tested whether it helps and it does not so I'm going to leave cycle count frequency as a view for now.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -8,10 +8,10 @@ SELECT
     product_availability.location_id                                                 as facility_id,
 
     -- Cycle count request ID
-    cycle_count_request_summary.cycle_count_request_id                               as cycle_count_request_id,
+    MAX(cycle_count_request_summary.cycle_count_request_id)                               as cycle_count_request_id,
```
jmiranda, line 11: I'm hoping this resolves the issue that Evan brought up in Slack (https://openboxes.slack.com/archives/C06JCHWR807/p1741627530593889).

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -41,42 +41,62 @@ SELECT
...
+    -- NULL as date_last_count,
     (select max(date_counted)
      from product_count_history
      where product_count_history.product_id = product_availability.product_id
        and product_count_history.inventory_id = location.inventory_id
      group by inventory_id, product_id)                                              as date_last_count,
 
-    NULL                                                                             as date_next_count,
+#    NULL as date_next_count,
+#    NULL as days_until_next_count,
```
jmiranda, line 52: I'm still going back and forth with these NULL columns as I work through the performance issues, but will remove them before we merge.

jmiranda:
I've marked this as ready for review but I still want to do a final cleanup before merging. Also, performance is still really slow (3-5 seconds) so I'd really like to get that licked before we merge.

jmiranda:
@awalkowiak @EWaterman @kchelstowski @alannadolny Can you guys review this when you get a chance?

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -8,13 +8,13 @@ SELECT
...
-    cycle_count_request_summary.cycle_count_request_id                               as cycle_count_request_id,
+    MAX(cycle_count_request_summary.cycle_count_request_id)                          as cycle_count_request_id,
 
     -- ABC Classification
-    -- FIXME Using a grouping operator due to a GROUP BY issue. This isn't the best approach sicne it'll return an
+    -- FIXME Using a grouping operator due to a GROUP BY issue. This isn't the best approach since it'll return an
     --  empty string or NULL value ahead of valid ABC classes (A, B, C, etc). But I don't have a better solution and
     --  this will likely work for 99% of cases.
-    MIN(COALESCE(inventory_level_summary.abc_class, product.abc_class))              as abc_class,
+    MIN(inventory_level_summary.abc_class)              as abc_class,
```
jmiranda, line 17: Removed product.abc_class default to avoid JOIN (and because requirements aren't asking for this COALESCE). This has negligible impact on performance so we can add it back if necessary.

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -35,48 +35,60 @@ SELECT
...
     -- Date last counted
-    (select max(date_counted)
-     from product_count_history
-     where product_count_history.product_id = product_availability.product_id
-       and product_count_history.inventory_id = location.inventory_id
-     group by inventory_id, product_id)                                              as date_last_count,
-
-    NULL                                                                             as date_next_count,
+    -- NULL as date_last_count,
+    cycle_count_metadata.date_counted                                                as date_last_count,
```
jmiranda, line 47: This small change (removing subquery and pulling date counted from this metadata view) carved off about 3-4 seconds from the query.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -70,22 +72,40 @@ class CycleCountService {
...
+
             if (command.negativeQuantity) {
                 gt("negativeItemCount", 0)
             }
+
+            // FIXME It's possible we need this to be "ne" rather "gt" because we probably want to include
+            //  product/facility pairs where quantity on hand is negative.
+            // Moved this from the cycle count session view since it's a requirement of the candidate query,
+            // not the cycle count session.
+            gt("quantityOnHand", 0)
```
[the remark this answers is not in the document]
jmiranda: Yeah I like that. I thought about it, but figured this was a core requirement. But now that Katarzyna is asking about this in OBPIH-6929, it make sense to let the client control the behavior.

grails-app/migrations/views/cycle-count-candidate.sql
```diff
@@ -0,0 +1,19 @@
+CREATE OR REPLACE VIEW cycle_count_candidate AS
```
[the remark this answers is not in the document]
jmiranda: I created a topic to discuss this in the tech huddle channel. This view is now used for CycleCountCandidate so all queries go through this view which just wraps with session with the default sorting required for the ranking algorithm.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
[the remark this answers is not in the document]
jmiranda: Ok, so this is a little more complicated than "let the client decide". We now need to codify out of stock vs negative stock and use this new concept for both filtering and sorting. In other words, I want to show records with QoH <= 0 if they have ANY inventory items with negative quantity but I don't want to show records with QoH <=0 if they don't have inventory items with negative quantity. I'm try to wrap my brain around how to do this elegantly.

grails-app/migrations/views/product-count-history.sql
```diff
@@ -1,19 +1,16 @@
 create or replace view product_count_history as
```
EWaterman, line 1: I'm still confused why this view is needed if we have product_physical_count_history, which should already include the transactions from cycle counts. Can't we just pull from product_physical_count_history directly?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -107,6 +126,14 @@ class CycleCountService {
...
                 break
             default:
+                // FIXME We could potentially handle this in the future by making ABC class a first-class citizen
+                //  and allowing it to be configured with a sort value, the default sort value in cases of NULL would
+                //  be some large integer value (999).
+                // The default sort order for the cycle count session (i.e. at least for the All Products tab) should
+                // be by ABC class, then by days until next count. The first order guarantees that NULL abc classes
+                // are sorted last.
+                //criteria.order(Order.asc("abcClass").nulls(NullPrecedence.LAST))
+                //        .order("daysUntilNextCount", "asc")
```
EWaterman, line 136: can you clarify that this ordering is specified in the view declaration (cycle-count-candidate.sql) which is why we don't need to do anything here?


As to your comment, another way we could do it is have all this factor into the score calculation for a product. We'd be able to specify the importance of each factor based on the number of score it adds.
- class A product score = 10000 + 10 * number of days overdue
- class B product score = 1000 + 10 * number of days overdue
- etc

And then the default case could simply sort by score.

As I'm typing this though, I'm thinking maybe that's not the greatest idea. Lots of edge case behaviour. Ex: if we have a super overdue B class item, it could rank higher than only slightly overdue A class items. And that only gets more complex as we add more factors into the score. It would speed up sorting calculations quite a bit though if we could materialize the score.

Anyways it's a problem for later.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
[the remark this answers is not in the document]
EWaterman: there's two things here:
- filtering out products where QoH == 0 (ie out of stock, so we want to make this ne as your comment suggests)
- filtering down to only products with any [bin + lot] where QoH < 0 (isn't that what negativeItemCount is doing?)

grails-app/migrations/views/cycle-count-metadata.sql
```diff
@@ -0,0 +1,25 @@
...
+    date_add(cycle_count_metadata_tmp.date_counted, INTERVAL cycle_count_frequency.frequency DAY) as date_expected,
+    datediff(date_add(cycle_count_metadata_tmp.date_counted, INTERVAL cycle_count_frequency.frequency DAY), now()) as days_until_next_count
+FROM (
+         SELECT
+             product_count_history.inventory_id,
+             product_count_history.product_id,
+             product_count_history.date_counted,
+             datediff(now(), date_counted) as days_since_last_count,
+             -- FIXME Used to clean data, this should not be used in the actual solution
+             CASE WHEN abc_class = 'A' THEN 'A' WHEN abc_class = 'B' THEN 'B' WHEN abc_class = 'C' THEN 'C' ELSE 'DEFAULT' END as abc_class
```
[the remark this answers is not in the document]
EWaterman: yeah can you not just join on product_classification.abc_class and use that? That would let you get rid of this nested query entirely and do everything in a single query

grails-app/migrations/views/cycle-count-frequency.sql
```diff
@@ -0,0 +1,13 @@
+create or replace view cycle_count_frequency as (
+  SELECT abc_class, facility_id, frequency
+  FROM (
+      select 'A' as abc_class, NULL as facility_id, 90 as frequency
```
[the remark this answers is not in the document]
jmiranda: I wanted to build the view to suggest what this could look like if it was an actual table. I'm essentially using the view as an "interface" for what I eventually want to build.

grails-app/migrations/views/cycle-count-metadata.sql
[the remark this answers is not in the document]
jmiranda: ISNULL is not comprehensive enough, as far as I can tell. It's not just NULL but also empty string, whitespace, etc.

In addition, I thought I may need to map other ABC classes into valid ones until someone had a chance to clean the data. For example, if someone was using VEN we could just standardize on ABC.
```
CASE 
    WHEN abc_class = 'A' THEN 'A' 
    WHEN abc_class = 'B' THEN 'B' 
    WHEN abc_class = 'C' THEN 'C' 
    WHEN abc_class = 'V' THEN 'A'
    WHEN abc_class = 'E' THEN 'B'
    WHEN abc_class = 'N' THEN 'C'
    ELSE 'DEFAULT' 
END as abc_class
```
Here's a list of the ABC classes in my local database
```
MariaDB [openboxes]> select distinct(abc_class) from product union select distinct(abc_class) from inventory_level;
+-----------------------------------------------+
| abc_class                                     |
+-----------------------------------------------+
| NULL                                          |
| A                                             |
| UN2465                                        |
| UN2880                                        |
| UN3481                                        |
| B                                             |
| UN 1090                                       |
| UN3481 Li-ion batteries packed with equipment |
| C                                             |
| UN3481, II, PI967                             |
| UN1219                                        |
| UN1044                                        |
| Maintenance and Infrastructure                |
| B23-01-C3                                     |
| B18-01-D4                                     |
| 500                                           |
| 220                                           |
| B32-03-A1                                     |
|                                               |
| V                                             |
| BBB                                           |
+-----------------------------------------------+
21 rows in set (0.006 sec)
```

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -8,13 +8,13 @@ SELECT
     product_availability.location_id                                                 as facility_id,
 
     -- Cycle count request ID
-    cycle_count_request_summary.cycle_count_request_id                               as cycle_count_request_id,
+    MAX(cycle_count_request_summary.cycle_count_request_id)                          as cycle_count_request_id,
```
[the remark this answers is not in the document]
jmiranda: > but any consideration for "ANY_VALUE"? 

Yes, that's a great suggestion. I think I read about that recently but was having trouble getting it to work with a subquery and decided to abandon it. But this seems like a perfect place for it. 

One might even suggest it was introduced exactly for this purpose.

> This function is useful for GROUP BY queries when the [ONLY_FULL_GROUP_BY](https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_only_full_group_by) SQL mode is enabled, for cases when MySQL rejects a query that you know is valid for reasons that MySQL cannot determine.

Source: https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_any-value

grails-app/migrations/views/cycle-count-session.sql
jmiranda: Yeah, this was added before I joined product_classification. But I think you're right.

grails-app/migrations/views/product-classification.sql
```diff
@@ -0,0 +1,12 @@
...
+-- FIXME Consider cleaning the ABC classification here instead of in the session view.
+--  We could change the default case to DEFAULT here as well. The only
+CREATE OR REPLACE VIEW product_classification AS (
+    select
+        inventory_level.inventory_id as inventory_id,
+        inventory_level.product_id as product_id,
+        min(coalesce(inventory_level.abc_class, product.abc_class, NULL)) as abc_class
+    from inventory_level
+       join product on inventory_level.product_id = product.id
+    group by inventory_level.inventory_id, inventory_level.product_id
```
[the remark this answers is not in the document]
jmiranda: Because you can have multiple inventory levels for a product and inventory 
* facility-level rules: allows one facility level rule that can be used to set the min/max/reorder and ABC class for the entire facility
* bin replenishment rules: allows multiple inventory levels (one per internal location) that define the min/max/reorder for that bin. ABC class is generally ignored here but we might eventually want product_classification to become more of an inventory item classification. Our current workflow won't support this but I could imagine a time when we'd want an inventory level to set the ABC class for a zone or bin location (regardless of the product). However, we might need to completely redesign the inventory level feature at that point.

There was also a time (before we added the unique constraint) where you could import multiple facility-level records for a single product and inventory. So it's not impossible to imagine a scenario where some of this data has not been cleaned up.

grails-app/migrations/views/product-count-history.sql
jmiranda: Because when we do cycle counts by bin (0.9.5), I assume we probably won't have a product inventory (snapshot) transaction created for cycle counts (we'll just use adjustments). Using the snapshot transaction for these will be much more dangerous. 

At some point in the future, I would actually prefer to move in the opposite direction and use only cycle count records here (i.e. ignore the transactions altogether). If we didn't have to look at transaction data, then the candidates query would return in under a second.

jmiranda: These are great questions, by the way. Thanks for putting in the effort to understand the view.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
jmiranda: > can you clarify that this ordering is specified in the view declaration (cycle-count-candidate.sql) which is why we don't need to do anything here?

Yes, I was trying to get it to work in the query itself (rather than the session or candidate view). I'll end up removing this comment altogether. 

> As I'm typing this though, I'm thinking maybe that's not the greatest idea. 

The main reason I don't like a scoring algorithm right now is for the reason you brought up. If we get the weights wrong there will likely be items at the end of one subset that get sorted below the next subset. Or worse, items that are supposed to be sorted last, are sorted first. And I really don't want to be tinkering with weights for the next few weeks.

I have it working well in the candidates view right now (without the weights). I don't want the weights to be used at this point because those are more useful in helping to determine what the ABC classification should be. The ranking algorithm is just a straightup sort on a few properties.

grails-app/migrations/views/cycle-count-candidate.sql
```diff
@@ -0,0 +1,21 @@
...
+(
+SELECT *,
+       (negative_item_count > 0 OR quantity_on_hand > 0) as included,
```
jmiranda, line 4: I need some help with the naming of this property. This is essentially the requirement "by default, show me only records with stock on hand OR records with invalid stock"

* is_on_hand_or_in_error - awkward

Or two separate properties that the query client can enable. 

* is_on_hand
* is_in_error

cc @awalkowiak @EWaterman

grails-app/migrations/views/cycle-count-metadata.sql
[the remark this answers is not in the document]
EWaterman: I really would like to have product_classification be a real table and have products and inventory level map directly to objects. That'd make all this so much cleaner than dealing with raw strings

grails-app/migrations/views/cycle-count-session.sql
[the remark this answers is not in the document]
jmiranda: > EDIT: wait, do we even need the base level group by anymore? This query is so big it's hard to tell, but if all we're doing is grouping by product_availability.location_id, product_availability.product_id, location.inventory_id, and we know that the query as is will already be unique for all those, isn't the group by redundant?

I assume you're asking whether we need this. It's possible we don't I just like to be defensive. 
```
GROUP BY product_availability.location_id, product_availability.product_id, location.inventory_id
```

grails-app/migrations/views/cycle-count-candidate.sql
jmiranda: So it probably makes sense to break these out. I think we'll eventually want the frontend (API client) to be able to choose which groups to return

* Include Stock On Hand
* Include Stock with Negative Quantities
* Include Stock with Recent Variances
* Include Stock with Picking Errors

But I can't figure out how to do this generically in the query below. 
https://github.com/openboxes/openboxes/pull/5106/files#diff-a5e7f95312fa610204cf7b69a621d9756120d1063efce16f5815f91f3dbcaf91R90

First we did this 
```
gt("quantityOnHand", 0)
```
But @awalkowiak suggested we allow the client to pass this as a parameter in the command. 
```
gt("quantityOnHand", command.quantityOnHand)
```
That would be fine, but the API client doesn't really want to include a quantityOnHand value, they just want to know whether the stock is on hand or not. 
```
if (command.isOnHand) { 
    gt("quantityOnHand", 0)
}
```
But I needed to handle the case for both "on hand and in error" so I moved to this gem.
```
eq("included", true)
```
which should really be something like
```
eq("isOnHandOrInError", true)
```
But if we break out the logic, we could make this a little more useful and readable
```
// By default, these properties should return true
if (command.isOnHand || command.isInError) { 
    or { 
        eq("isOnHand", command.isOnHand)
        eq("isInError", command.isInError)
    }
}
```
Any other suggestions?

grails-app/migrations/views/cycle-count-candidate.sql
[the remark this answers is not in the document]
jmiranda: Fwiw, the reason to do this was to ensure that CycleCountCandidate was mapped to cycle_count_candidate (instead of cycle_count_session) but more importantly to move the ranking (order by) into the appropriate layer. 

Adding it to the session would impose an implementation detail on cycle count session that was unnecessary and unwanted. 

Secondly, it would have been way more complicated to add to the cycle count session view as I probably would have needed to use a nested query in the view definition to ensure that all of the columns required by the ranking algorithm were in scope. 

```
CREATE OR REPLACE VIEW cycle_count_session AS (
    select 
        ugly_inner_query.*, 
        CASE ... END as sort_order
    from (
        -- huge ugly query 
        select * from ... 
    ) as ugly_inner_query
    order by sort_order
)
```

src/main/groovy/org/pih/warehouse/inventory/CycleCountCandidateFilterCommand.groovy
```diff
@@ -30,7 +30,7 @@ class CycleCountCandidateFilterCommand extends PaginationCommand implements Vali
 
     String format
 
-    boolean negativeQuantity = false
+    Boolean negativeQuantity = false
```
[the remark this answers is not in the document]
jmiranda: So that means leave it like it is here?

grails-app/migrations/views/cycle-count-candidate.sql
EWaterman: These are very loose thoughts and I don't know if this is a good idea but it seems like we want both an include and exclude option on certain filters.

So we could have an enum representing the filters: [OUT_OF_STOCK, NEGATIVE_QUANTITY, IN_STOCK]
And the query params "include" and "exclude".

Ex: `GET /cycle-counts/candidates?include=NEGATIVE_QUANTITY&include=IN_STOCK&exclude=OUT_OF_STOCK`

and an implementation could be:
```
// validate that something isn't in both include and exclude

or {
  for (filter in command.includes) {
    switch (filter) {
      case OUT_OF_STOCK:
        eq("quantityOnHand", 0)
        break
      case NEGATIVE_QUANTITY:
        le("negativeItemCount", 0)
        break
      ....
    }
  }
}

and {
  for (filter in command.excludes) {
    switch (filter) {
      case OUT_OF_STOCK:
        ne("quantityOnHand", 0)
      ....
    }
  }
}
```
I kinda hate it? But maybe it works?

awalkowiak: Overall separating this one makes the most sense to me because these are two different things (negative qoh and positive qoh, but one on inventory item, and other total on inventory), that could be later combined like you suggested here: 
```
// By default, these properties should return true
if (command.isOnHand || command.isInError) { 
    or { 
        eq("isOnHand", command.isOnHand)
        eq("isInError", command.isInError)
    }
}
```
but if you won't be available for it today, we can split it later, I can add a fixme or something in the code.

jmiranda: @EWaterman I like the idea of generically handling includes and excludes but don't want to deal with that at the moment.

@awalkowiak If I'm not mistaken, separating these two different conditions means we'd have two WHERE conditions that are AND'd together. Which we don't want since that means we'd just see rows that are BOTH On Hand and In Error.
```
if (command.includeOnHand) { 
    eq("includeOnHand", command.includeOnHand)
}
if (command.includeInError) { 
    eq("includeInError", command.includeInError)
}
```
So we'd need to do the following from the start.
```
// By default, these properties should return true by default
if (command.includeStockOnHand || command.includeNegativeStock) { 
    or { 
        if (command.includeStockOnHand) { 
            eq("hasStockOnHand", command.includeStockOnHand)
        }
        if (command.includeNegativeStock) { 
            eq("hasNegativeStock", command.includeNegativeStock)
        }
    }
}
```
I'll give this a shot, but if it doesn't work on the first try I'm going to back to the simple version.
```
eq("hasStockOnHandOrNegativeStock", command.includeStockOnHandOrNegativeStock)
```
along with this column in cycle_count_candidate view
```
(negative_item_count > 0 OR quantity_on_hand > 0) as has_stock_on_hand_or_negative_stock
```

jmiranda: Actually nevermind, it's too late for me to be doing this so I'm just going to change the name of the property to make it more clear what we're doing. 

It just occurred to me that this might add too much confusion around the "Include negative quantities" filter checkbox (which should return only cycle count candidates with negative inventory items) and the implicit rule that we want to include cycle count candidate records with negative inventory items, by default.

grails-app/migrations/views/cycle-count-session.sql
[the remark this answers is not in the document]
jmiranda: FWIW, since the request ID is not needed for filtering / sorting I might eventually pull this out as a separate association which might require an N+1 query once the cycle count candidate query returns.

EWaterman (approved):

PR #5109, opened 2025-03-10
OBPIH-7032 Complete count after count with no discrepancy or after recount

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7032

**Description:** We're implementing the review and approval flows in 0.9.5 so this is a simple refactor to move the count straight to the COMPLETED state after a count is submitted with no discrepancies or after a recount is submitted (with or without discrepancies).

Please merge this after https://github.com/openboxes/openboxes/pull/5092. It fixes a bug in the product availability flow that this change uses.

---

https://jam.dev/c/1eb79d37-3ef4-4e77-bae7-9ebbbb9cb424

(There are two product inventories because I tested the same product twice.)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -421,12 +421,34 @@ class CycleCountService {
...
+            return
+        }
+
+        if (cycleCount.status != CycleCountStatus.COMPLETED) {
+            throw new IllegalStateException("Cannot complete cycle count when it's in status ${cycleCount.status}")
+        }
+
+        // The count completed successfully so update the request status to match and commit the adjustments.
+        cycleCount.cycleCountRequest.status = CycleCountRequestStatus.COMPLETED
+        cycleCountTransactionService.createTransactions(cycleCount, refreshQuantityOnHand)
```
awalkowiak, line 449: Would it not be better to create a transaction and close afterward?

awalkowiak (commented):
overall looks ok, but I have a minor question

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: sure I can flip these but this whole call is wrapped in a `@Transactional` so ultimately the order shouldn't matter much. If it fails at any point the whole thing should get rolled back

EWaterman:
I'm merging this as the only real change I made was addressing the one comment

PR #5114, opened 2025-03-11
OBPIH-7079 Change association from user to person for assignee in cyc…

kchelstowski:
…le count item

**Description:**

---

### DIFF
--- a/grails-app/migrations/0.9.x/changelog-2025-03-11-1400-add-person-foreign-key-for-assignee-in-cycle-count-item.xml
+++ b/grails-app/migrations/0.9.x/changelog-2025-03-11-1400-add-person-foreign-key-for-assignee-in-cycle-count-item.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9 http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+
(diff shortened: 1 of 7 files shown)


grails-app/migrations/0.9.x/changelog-2025-03-11-1400-add-person-foreign-key-for-assignee-in-cycle-count-item.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <changeSet author="kchelstowski" id="110320251400-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <foreignKeyConstraintExists foreignKeyName="fk_cycle_count_item_assignee" />
+            </not>
+            <columnExists tableName="cycle_count_item" columnName="assignee_id" />
+        </preConditions>
+
+        <addForeignKeyConstraint
```
EWaterman, line 14: can you put these two together in the same change set? They should always be run together in the same transaction otherwise if the first runs and the second fails, we could end up in a bad db state.

EWaterman:
Do we really need a person to be able to be marked as assignee? I don't know how our facilities are working but it seems strange that a non-openboxes user would be performing a count.

If this is true and we only want users to be able to be assignee, we could instead filter the dropdown list of assignees to only include users (not person).

Otherwise this is fine :+1:

grails-app/migrations/0.9.x/changelog-2025-03-11-1400-add-person-foreign-key-for-assignee-in-cycle-count-item.xml
kchelstowski: I'd say there is low risk we could run into such state, due to the `preConditions`. 
1. I don't remove the foreign key if it doesn't exist
2. I don't create a new one, if the foreign key somehow still exists.

alannadolny (approved):

awalkowiak (approved):

PR #5123, opened 2025-03-12
OBPIH-6931 fix status not updating to "in progress" when editing a count

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6931

**Description:** Fixes a bug where a cycle count status was staying at "To count" (instead of moving to "In progress", even after being edited.

---

https://jam.dev/c/665fdbb2-861f-4c2b-b0e7-a689a76df42d

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -397,17 +397,31 @@ class CycleCountService {
             determineCycleCountItemStatusForSubmit(cycleCountItem, command.requireRecountOnDiscrepancy)
         }
 
-        cycleCount.status = cycleCount.recomputeStatus()
+        // We've updated the status of the cycle count items so we need to also update the status of the count.
+        recomputeCycleCountStatus(cycleCount)
+
         // TODO: Investigate why status could be null here
         if (cycleCount.status?.isClosed()) {
             closeCycleCount(cycleCount, command.refreshQuantityOnHand)
         }
 
-        // TODO: The beforeUpdate() on CycleCount class is not triggered without
-        // the line below, so without it status is not correct in the DB.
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -397,19 +397,33 @@ class CycleCountService {
...
 
+    /**
+     * Recalculate the status of a cycle count based on the status of its items.
+     *
+     * This status will be automatically recomputed when a cycle count is saved, so this method only needs to be called
+     * when we want to trigger a status recalculation outside of that flow, such as when the status of a cycle
+     * count item changes.
+     */
+    private void recomputeCycleCountStatus(CycleCount cycleCount) {
+        cycleCount.status = cycleCount.recomputeStatus()
```
awalkowiak, line 419: Should we now get rid of the status update from `CycelCount.beforeUpdate()` (permanent or temporary)? Or it is fine there as well?

EWaterman: We'd need to also manually set the status when starting a recount because we call cycleCount.save() there as well. And potentially needs a fix to deal with that TODO in CycleCountSubmitCountCommand.

It's doable, but would require some additional investigation. I also have [this post-launch task](https://pihemr.atlassian.net/browse/OBPIH-7045) for us to reconsider if we even need recomputeStatus() at all (and instead simply set the status to a fixed value at each step, ex: submit count -> status is COUNTED).

So IMO think since it's working currently, I think this is fine for now.

awalkowiak: Ok

awalkowiak (approved):

PR #5126, opened 2025-03-13
OBPIH-7092 Fix issue related to null cycle count status

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
@@ -77,30 +77,25 @@ class CycleCount {
      * Determines what the CycleCountStatus should be for the cycle count.
      */
     CycleCountStatus recomputeStatus() {
-        if (!cycleCountItems || cycleCountItems.every { it.status == CycleCountItemStatus.READY_TO_COUNT }) {
-            return CycleCountStatus.REQUESTED
-        }
-        if (cycleCountItems.every { it.status == CycleCountItemStatus.REVIEWED }) {
-            return CycleCountStatus.REVIEWED
-        }
-        if (cycleCountItems.every { it.status == CycleCountItemStatus.READY_TO_REVIEW }) {
-            return CycleCountStatus.READY_TO_REVIEW
-        }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
-            return CycleCountStatus.INVESTIGATING
-        }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.COUNTING }
-                && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.READY_TO_REVIEW] }) {
-            return CycleCountStatus.COUNTING
-        }
-        // TODO: Once we add support for the "to review" tab, we should look for READY_TO_REVIEW instead of APPROVED
-        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.APPROVED] }) {
-            return CycleCountStatus.COUNTED
-        }
-        if (cycleCountItems.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED] }) {
+        Set<CycleCountItem> items = itemsOfMostRecentCount
+        if (items.every { it.status in CycleCountItemStatus.isCompleted()}) {
             return CycleCountStatus.COMPLETED
         }
-        return null
+        if (items.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
+            return CycleCountStatus.INVESTIGATING
+        }
+        if (items.any { it.status == CycleCountItemStatus.COUNTED }) {
+            return CycleCountStatus.COUNTED
+        }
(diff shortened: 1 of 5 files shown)


grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -93,7 +93,8 @@ class CycleCount {
                 && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.READY_TO_REVIEW] }) {
             return CycleCountStatus.COUNTING
         }
-        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.READY_TO_REVIEW] }) {
+        // TODO: Once we add support for the "to review" tab, we should look for READY_TO_REVIEW instead of APPROVED
+        if (cycleCountItems.every { it.status in [CycleCountItemStatus.COUNTED, CycleCountItemStatus.APPROVED] }) {
```
EWaterman, line 97: This looks like it makes the next "if" block unreacheable. If all the items are APPROVED, we wan the status to be COMPLETED. So I think we need to restructure this a bit.

something like:
```
if (cycleCountItems.every { it.status in [CycleCountItemStatus.REVIEWED, CycleCountItemStatus.APPROVED, CycleCountItemStatus.REJECTED, CycleCountItemStatus.CANCELED] }) {
    return CycleCountStatus.COMPLETED
}

if (cycleCountItems.any { it.status in [CycleCountItemStatus.COUNTED] }) {
    return CycleCountStatus.COUNTED
}
```

But personally I would completely flip these conditions to be:

- If all items are APPROVED or REVIEWED or REJECTED or CANCELED (or READY_TO_REVIEW for now), count status is COMPLETED
- Else if any items are INVESTIGATING, count status is INVESTIGATING
- Else if any items are COUNTED, count status is COUNTED
- Else if any items are COUNTING, count status is COUNTING
- Else if any items are READY_TO_COUNT, count status is REQUESTED

I think that's the cleaner solution but I'll leave it to you to decide

kchelstowski: @EWaterman I think I've raised that many times before that in my opinion we either duplicate a few statements or are missing few or few are not achievable. Let me know if in your opinion it's not a big risk refactoring the whole state machine now, as we are tight on time, or you feel like it's "must have" that we have not committed earlier?
I'm a bit frightened of proceeding with such change now, but if you make me confident about that, I'll proceed with that.

EWaterman: if you're feeling worried about it you can just do something like my first suggestion.

If it's helpful, I can write a bunch of unit tests on this method today to confirm what the behaviour is. That way you can verify that your changes still pass the tests.

awalkowiak: Yeah, I think the original state machine did not take into account count index at all. So it would makes sense to take items of the most recent count, but looks like this always takes items with max count index, does it work also for new inventory items that have been created during a count or recount? Overall, this does not feel like a huge deal with the refactoring of the state machine since it needs to support our **current** MVP case. And simply changing ordering and jumping to completed without approval should be fairly straightforward. Let me know if you have some issues with it, I will try to help you there.

kchelstowski: @EWaterman @awalkowiak I refactored the `recomputeStatus()` method. If you find time @EWaterman you could test it out to make sure it covers all edge cases.
One note is that this branch is a bit old (before merging a few fixes including yours), and I found one thing not working properly - e.g. cycle count status in response was as expected, but in the database it was not persisted, or when editing an item on resolving step, e.g. updating a root cause, it was not creating a new cycle count item, but updating an existing one.
So in order to test it, I suggest rebasing the branch, or just pulling my commits to your test branch, if you create one.

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -77,29 +77,25 @@ class CycleCount {
...
+        if (items.every { it.status.isCompleted() }) {
+            return CycleCountStatus.COMPLETED
         }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
+        if (items.any { it.status == CycleCountItemStatus.INVESTIGATING }) {
             return CycleCountStatus.INVESTIGATING
         }
-        if (cycleCountItems.any { it.status == CycleCountItemStatus.COUNTING }
-                && !cycleCountItems.any { it.status in [CycleCountItemStatus.INVESTIGATING, CycleCountItemStatus.READY_TO_REVIEW] }) {
+        if (items.any { it.status == CycleCountItemStatus.COUNTING }) {
```
EWaterman, line 87: @kchelstowski heads up that I put the COUNTING condition before the COUNTED one. I wasn't going to push any changes here but I needed to fix this to test other things locally so I figured I'd push straight to your branch and (hopefully) save you some headache.

This change made sense to me because if a count has one item that was counted but one that is not finished being counted yet, the count should still be in the COUNTING state because we're not done counting.

Let me know if you disagree though.

kchelstowski: That's fine @EWaterman. Thanks for the effort

PR #5130, opened 2025-03-14
OBPIH-7088 Input from quantity counted is shifted to recount columns when start resolution on products with discrepancy

alannadolny:


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -342,8 +342,9 @@ class CycleCountService {
 
         // If there are custom items, they won't have been discovered by the above loop (because they won't have
         // a product availability record) so make sure to create new recount items for them as well.
-        // "Most recent count" in this scenario is the previous count.
-        Set<CycleCountItem> customItemsOfLastCount = cycleCount.itemsOfMostRecentCount.findAll{ it.custom }
+        Set<CycleCountItem> customItemsOfLastCount = cycleCount.cycleCountItems.findAll {
+            (cycleCount.maxCountIndex - 1) == it.countIndex && it.custom
```
awalkowiak, line 346: `cycleCount.maxCountIndex - 1` - can we put a fallback to `0` if `maxCountIndex` is `0` to avoid accidentally falling into `-1`, or it is impossible at this stage?

awalkowiak (approved):

PR #5131, opened 2025-03-14
OBPIH-7088 Hide close button on resolve discrepancies modal

awalkowiak:


SebastianLib (approved):

src/js/utils/ConfirmModal.jsx
```diff
@@ -13,6 +13,7 @@ const ConfirmModal = ({
   onClose,
   buttons,
   className,
+  hideCloseButton = false,
```
kchelstowski, line 16: nitpicky, but we used to set default prop values in the `Component.defaultProps`

kchelstowski: I can also see this prop missing in the `propTypes`

awalkowiak: yup, forgot about this one

PR #5137, opened 2025-03-14
OBPIH-7015 Improve blind count form

awalkowiak:
- fix NoSuchElementException on cycle count list action
- print assignee and date counted if filled
- fix printing count and recount form
- fix row height if row with default (null) bin and lot and empty expiry date
- remove unnecessary format adding to the params

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -193,7 +193,7 @@ class CycleCountService {
                         "Lot Number": item.inventoryItem.lotNumber,
                         "Expiration Date": item.inventoryItem.expirationDate
                                 ? Constants.EXPIRATION_DATE_FORMATTER.format(item.inventoryItem.expirationDate) : "",
-                        "Bin Location": item.binLocation?.locationNumber,
+                        "Bin Location": item.binLocation?.name,
                         "Quantity Counted": "",
                         "Comment": "",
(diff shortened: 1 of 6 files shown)


EWaterman (approved):

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -193,7 +193,7 @@ class CycleCountService {
                         "Lot Number": item.inventoryItem.lotNumber,
                         "Expiration Date": item.inventoryItem.expirationDate
                                 ? Constants.EXPIRATION_DATE_FORMATTER.format(item.inventoryItem.expirationDate) : "",
-                        "Bin Location": item.binLocation?.locationNumber,
+                        "Bin Location": item.binLocation?.name,
                         "Quantity Counted": "",
                         "Comment": "",
                         "User Counted": "",
```
EWaterman, line 199: is it intentional that user counted and date counted are always left blank (even if we've set a value for them in the app)? This would be different behaviour than the PDF (not saying that's a bad thing, just observing).

EWaterman:
I tested this locally and it looks good :+1: 

Some notes (but I still approve this as is):
- a custom added row to a recount does not appear in the exported pdf
- unless you click save progress, the print count form pdf won't contain any changes. IMO when the print button is clicked, the frontend should call save before it calls print.
- if the location doesn't have bin locations enabled, we should probably not include that column in the csv or pdf

PR #5139, opened 2025-03-17
OBPIH-7087 Edited value of Date counted is not saved

SebastianLib:

**Description:**

---


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -514,7 +514,7 @@ class CycleCountService {
         cycleCountItem.properties = command.properties
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
-        cycleCountItem.dateCounted = new Date()
+        cycleCountItem.dateCounted = command.dateCounted
```
alannadolny, line 517: can be there a situation when the command.dateCounted will be null? Don't we need a fallback?

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -20,6 +20,8 @@ class CycleCountUpdateItemCommand implements Validateable {
 
     Person assignee
 
+    Date dateCounted;
```
kchelstowski, line 23: remove `;`

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -38,5 +40,6 @@ class CycleCountUpdateItemCommand implements Validateable {
         })
         comment(nullable: true, blank: true)
         assignee(nullable: true)
+        dateCounted(nullable: true);
```
kchelstowski, line 43: remove `;`

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
SebastianLib: these ; I added out of habit of working in react

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -514,7 +514,7 @@ class CycleCountService {
         cycleCountItem.properties = command.properties
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
-        cycleCountItem.dateCounted = new Date()
+        cycleCountItem.dateCounted = command.dateCounted ? command.dateCounted : new Date()
```
kchelstowski, line 517:
```suggestion
        cycleCountItem.dateCounted = command.dateCounted ?: new Date()
```

SebastianLib: Yes, but I think it's always safer to consider the possibility that the date might somehow be null

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -514,7 +514,7 @@ class CycleCountService {
         cycleCountItem.properties = command.properties
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
-        cycleCountItem.dateCounted = new Date()
+        cycleCountItem.dateCounted = command.dateCounted ?: new Date()
```
EWaterman, line 517: If command.dateCounted isn't provided but cycleCountItem.dateCounted already has a value, we probably don't want to override that with a new date.

Because the assigment is already happening at line 513:
```
cycleCountItem.properties = command.properties
```

I'd change this line to something like:
```
if (cycleCountItem.dateCounted = null) {
}
```

EWaterman: Also, I see you added the dateCounted field to the create item call too but aren't using that on the backend.

In `createCycleCountItem(CycleCountItemCommand)` you'll want to add dateCounted to CycleCountItemCommand and then do:
```
CycleCountItem cycleCountItem = new CycleCountItem(
    ...
    dateCounted: command.dateCounted ?: new Date()
)
```

src/js/components/cycleCount/toCountTab/CountStepTable.jsx
```diff
@@ -92,7 +92,7 @@ const CountStepTable = ({
           >
             <DateField
               className="date-counted-date-picker date-field-input"
-              onChange={setCountedDate}
+              onChangeRaw={setCountedDate}
```
EWaterman, line 95: why is this needed? Is this so you can type in the date field and it'll accept the raw input too (as opposed to just picking using the picker?) Or is it because it was formatting first before passing the string? Was it causing undesired behaviour?

SebastianLib: I added this because ResolveStepTable also has onChangeRaw. Without it, I couldn't use date.format(), which caused the wrong format to be sent to the backend. When onChange was used, the date format received was different from the expected one, and I couldn't use date.format(). As a result, I would have had to write an additional function to convert the format correctly for the backend.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -514,7 +514,9 @@ class CycleCountService {
         cycleCountItem.properties = command.properties
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
-        cycleCountItem.dateCounted = new Date()
+        if (cycleCountItem.dateCounted = null) {
```
EWaterman, line 517: should be `==` (sorry the code suggestion I gave you only had `=`)

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -528,7 +528,9 @@ class CycleCountService {
...
+            cycleCountItem.dateCounted = new Date()
```
kchelstowski, line 532: I don't really understand this part. This code now suggests that we always set the date for current date, no matter what we provide in the command.

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: We already have `cycleCountItem.properties = command.properties` so all this change does is only set the date counted as today if the client isn't specifying it themselves and the item doesn't already have a date counted set

kchelstowski: What if the `dateCounted` is provided as `null` in the command? Wouldn't the `cycleCountItem.properties = command.properties` override the date to be null? @SebastianLib have you checked such case?
I want to make sure this won't actually cause the problem mentioned by you @EWaterman here: https://github.com/openboxes/openboxes/pull/5139#discussion_r2001574078

By saying so, this is why I was against setting the properties via `.properties =`. If we set it expliciltly, we would immediately see which properties we mutate and which not.

EWaterman: good point. If the frontend is explicitly unsetting the dateCounted field (by providing dateCounted: null), we should respect that. Actually, since this is an update API, maybe we shouldn't be trying to provide a new Date() fallback at all. It's safe to assume that the default we provide when creating a new row is good enough.

In that case, we can simply remove this whole if block. Thoughts?

(And yeah I'm not sure how I feel about the `.properties =` solution. On the one hand, it's a good reminder that the API should be a simple update on count item fields with no complex logic, but on the other hand, it's less explicit about what it's doing and so can be somewhat obtuse.)

alannadolny (approved):

PR #5141, opened 2025-03-17
OBPIH-7061 Prevent fractional values in cycle count quantity field

SebastianLib:

**Description:**

---


grails-app/views/inventoryItem/_showRecordInventory.gsp
```diff
@@ -257,6 +258,12 @@
 		textField.val(parseInt(textField.val()) + delta);
 	}
 
+	function preventFractional(event) {
+		if (event.key === "." || event.key === ",") {
+			event.preventDefault();
+		}
+	}
```
alannadolny, line 265: how will it help for CYCLE COUNT?

alannadolny (changes requested):

kchelstowski (changes requested):
Why has it been done for record stock, not in cycle count? 🤔

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -141,12 +141,16 @@ const useCountStepTable = ({
       // When the input is blurred, we'll call the table meta's updateData function
       const onBlur = () => {
         if (isEdited) {
-          table.options.meta?.updateData(cycleCountId, original.id, id, value);
+          if (columnPath === cycleCountColumn.QUANTITY_COUNTED) {
+            setValue(parseInt(value, 10) || 0);
+            table.options.meta?.updateData(cycleCountId, original.id, id, parseInt(value, 10) || 0);
```
alannadolny, line 146: 1. You can assign the result of parseInt to a variable
2. I do not tolerate using else :imp:

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -140,13 +140,21 @@ const useCountStepTable = ({
       const isEdited = initialValue !== value;
       // When the input is blurred, we'll call the table meta's updateData function
       const onBlur = () => {
-        if (isEdited) {
-          table.options.meta?.updateData(cycleCountId, original.id, id, value);
+        if (!isEdited) return;
+
+        const parsedValue = parseInt(value, 10) || 0;
```
awalkowiak, line 145: I think you could move it into the if statement, because you are not using this later

awalkowiak (changes requested):

src/js/hooks/cycleCount/useCountStepTable.jsx
alannadolny: It should be moved to that if. What happens when you try to use `parseInt` on an object? Won't it throw an exception?

awalkowiak (approved):

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -145,11 +145,19 @@ const useCountStepTable = ({
       const isEdited = initialValue !== value;
       // When the input is blurred, we'll call the table meta's updateData function
       const onBlur = () => {
-        if (isEdited) {
-          table.options.meta?.updateData(cycleCountId, original.id, id, value);
+        if (!isEdited) return;
```
alannadolny, line 148: as you can see, in the previous version of code, we used brackets.

SebastianLib: sorry, it's just a habit

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -145,11 +145,20 @@ const useCountStepTable = ({
       const isEdited = initialValue !== value;
       // When the input is blurred, we'll call the table meta's updateData function
       const onBlur = () => {
-        if (isEdited) {
-          table.options.meta?.updateData(cycleCountId, original.id, id, value);
+        if (!isEdited) {
+          return;
+        }
+        if (columnPath === cycleCountColumn.QUANTITY_COUNTED) {
+          const parsedValue = parseInt(value, 10) || 0;
```
EWaterman, line 152: it feels unnecessary to me to specify the radix/base. It defaults to base 10 anyways and so just `parseInt(value)` is actually clearer to me (I had to look up parseInt to see what the 10 was doing). Same comment for useResolveStepTable

This is fine though if you prefer it this way.

EWaterman (approved):

EWaterman (changes requested):
You're still able to set fractional values if you use the arrow keys to navigate away from the field: https://jam.dev/c/269ff897-48e9-4c8e-b1e9-b4b4caedb5d3

Lot "Gac" I used the tab key and it correctly rounded down. Lot "Infra" I used the arrow key and it keeps the value.

SebastianLib:
@EWaterman oh no! then I'll also have to do parseInt when we click next

src/js/hooks/cycleCount/useCountStepTable.jsx
SebastianLib: @EWaterman from what I remember I wanted to do as you write here, but eslint recognized it as an error. I'll check tomorrow to be sure

EWaterman: I just tried it and you're right. Weird!

alannadolny (approved):

EWaterman (approved):
I'm less familiar with the useArrowsNavigation code, but from what I see, it looks fine. I tested the change and it looks good :+1:

PR #5143, opened 2025-03-17
OBPIH-7101 fix unique expiration date validation

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7101

**Description:** The specific issue brought up here is triggered when an existing (non-custom) row has an expiry and you try to add a new, custom row for the same lot and expiry. The error is caused by the fact that InventoryItems use a custom date format which the frontend doesn't use when adding new rows.

`InventoryItem.groovy`:
```
Map toJson() {
    [
        ...
        "expirationDate" : expirationDate?.format("MM/dd/yyyy"),
        ...
    ]
}
```
So when we’re validating in `useInventoryValidation.checkDifferentExpirationDatesForTheSameLot`, we get rows that have the same date but in different formats, so it thinks they’re different and throws a validation error:
```
"03/03/2027"                 -> existing row returned from the backend
"2027-03-03T17:00:00-08:00"  -> custom row not yet persisted to the backend
```

---

image

### DIFF
--- a/src/js/hooks/cycleCount/useInventoryValidation.js
+++ b/src/js/hooks/cycleCount/useInventoryValidation.js
@@ -45,15 +45,12 @@ const useInventoryValidation = ({ tableData }) => {
         cycleCountItems,
         (item) => item?.inventoryItem?.lotNumber,
       );
-      // The backend returns expiration date in the "MM/dd/yyyy" format, but we use the
-      // "yyyy-MM-dd'T'hh:mm:ssXXX" format when generating the dates on the frontend. We
-      // convert the dates to a moment for easy comparison.
+      // The backend returns expiration date in the "MM/dd/yyyy" format, but we use the ISO format
+      // "yyyy-MM-dd'T'hh:mm:ssXXX" when generating dates on the frontend. We convert the dates to
+      // a moment (defaulting to UTC if no timezone information is provided) for easy comparison.
       const expirationDates = dataGroupedByLotNumber[row?.inventoryItem?.lotNumber].map((item) => {
         const expirationDate = item?.inventoryItem?.expirationDate;
-        // TODO: isSame is returning false, even if the dates are the same! I suspect it's trying
-        //       to use local time when parsing "MM/dd/yyyy". Gotta find a way to make this default
-        //       to UTC. This kinda sucks though, maybe we need a more general solution to this...
-        return expirationDate == null ? null : moment(expirationDate);
+        return expirationDate == null ? null : moment.utc(expirationDate);
       });
       const uniqueDates = _.uniqWith(expirationDates, (arrVal, othVal) => arrVal?.isSame(othVal));
       if (uniqueDates.length > 1) {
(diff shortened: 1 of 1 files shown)


src/js/hooks/cycleCount/useInventoryValidation.js
```diff
@@ -44,11 +45,18 @@ const useInventoryValidation = ({ tableData }) => {
...
-        .map((item) => item?.binLocation?.id);
-      if (_.uniq(expirationDates).length !== 1 && _.uniq(binLocations).length !== 1) {
+      // The backend returns expiration date in the "MM/dd/yyyy" format, but we use the
+      // "yyyy-MM-dd'T'hh:mm:ssXXX" format when generating the dates on the frontend. We
+      // convert the dates to a moment for easy comparison.
+      const expirationDates = dataGroupedByLotNumber[row?.inventoryItem?.lotNumber].map((item) => {
+        const expirationDate = item?.inventoryItem?.expirationDate;
+        // TODO: isSame is returning false, even if the dates are the same! I suspect it's trying
+        //       to use local time when parsing "MM/dd/yyyy". Gotta find a way to make this default
+        //       to UTC. This kinda sucks though, maybe we need a more general solution to this...
```
EWaterman, line 55: would appreciate anyone's help getting this to work cleanly

EWaterman: Note that we can't just trim off the time portion of the date/moment because it could result in the date being different.

Ex: `2000-01-01T00:00:00Z == 1999-12-31T23:00:00-01:00` so we need to be careful about how we handle this. It's why just working with MM/dd/yyyy is probably simplest. No time component at all to work with...

src/js/hooks/cycleCount/useInventoryValidation.js
```diff
@@ -44,11 +45,18 @@ const useInventoryValidation = ({ tableData }) => {
...
+        return expirationDate == null ? null : moment(expirationDate);
+      });
+      const uniqueDates = _.uniqWith(expirationDates, (arrVal, othVal) => arrVal?.isSame(othVal));
```
alannadolny, line 58: you can try to use `.format` on the moment date, and then put those in `_.uniq`

EWaterman: I was doing something like that originally. It doesn't work either way because the moments are different. If I do something like this:

image

I end up with these expiration date strings being parsed in:
```
"03/03/2027"                -> "2027-03-03T08:00:00.000Z" (existing row, lot "LOT00016838")
"2027-03-03T16:00:00-08:00" -> "2027-03-04T00:00:00.000Z" (new row, lot "TEST")
```
So they're not the same day, despite appearing as if they are. I don't actually know what is happening to cause the dates to be different

Interestingly. the first time I pick a date using the datepicker, it always picks the day before I select. Then if I try again, it works. Maybe something to do with DateField.jsx? Not sure if that's related. https://jam.dev/c/7765ce91-6ee6-4c45-a5e0-978a6a54e38e

EWaterman: I fixed the first issue by defaulting to utc when parsing the string to a moment. The second issue (with the picker selecting the day before) is less obvious of a fix. I'm going to make a separate ticket to track that one though so this PR is ready for review as is.

PR #5145, opened 2025-03-18
OBPIH-7100 Fix and improve deleting new lines on recount form 

alannadolny:


src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -201,8 +201,20 @@ const useResolveStep = () => {
...
-    triggerValidation();
+  };
+
+  const removeRow = (cycleCountId, rowId) => {
+    try {
+      show();
+      if (!rowId.includes('newRow')) {
+        cycleCountApi.deleteCycleCountItem(currentLocation?.id, rowId);
+      }
+      removeRowFromState(cycleCountId, rowId);
```
kchelstowski, line 212: shouldn't we remove row from state only in the `if` above? I know that probably the bin icon is hidden for not new rows, but just for clarity?

kchelstowski: ok, sounds good, be careful though, because as far as I'm concerned, even if a cycle count item doesn't exist, the delete endpoint would still return 204, but I don't think we could run into such scenario.
Why I'm saying this is that probably in one of the steps I refetched the data after deleting instead of removing it from state, but I'm fine with current solution for now.

kchelstowski (approved):

PR #5146, opened 2025-03-18
OBPIH-7099 Prevent doing multiclik on Save on to count and recount steps

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/components/cycleCount/ConfirmStepHeader.jsx
+++ b/src/js/components/cycleCount/ConfirmStepHeader.jsx
@@ -1,14 +1,17 @@
-import React, { useState } from 'react';
+import React from 'react';
 
 import PropTypes from 'prop-types';
 
 import Button from 'components/form-elements/Button';
 
-const ConfirmStepHeader = ({ back, save }) => {
-  const [disabledSave, setDisabledSave] = useState(false);
-
+const ConfirmStepHeader = ({
+  back,
+  save,
+  isDisabledSave,
+  setIsDisabledSave,
+}) => {
   const handleSaveClick = () => {
-    setDisabledSave(true);
+    setIsDisabledSave(true);
     save();
   };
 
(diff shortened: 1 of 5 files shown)


src/js/utils/confirmationModalUtils.jsx
```diff
@@ -28,6 +29,7 @@ const confirmationModal = ({
         hideCloseButton={hideCloseButton}
       />
     ),
+    closeOnClickOutside,
```
EWaterman, line 32: just double checking: The behaviour before your change was to always close on click outside. Now it still defaults to that behaviour but you can specify for it to not by setting closeOnClickOutside = false.

EWaterman (approved):

kchelstowski (commented):
How does it prevent from clicking Save a few times?
Shouldn't we just add a proper spinner for such case?

SebastianLib:
@kchelstowski In this video, I try to click on 'Next Step' multiple times, but nothing happens https://jam.dev/c/ea4e6426-e461-4de4-a9cd-c49a645240c3

kchelstowski:
@SebastianLib ok, then why is this ticket title "Prevent doing multiclick"? I don't see anything related to preventing that in this PR. Could you then elaborate on what this PR is resolving?

SebastianLib:
@kchelstowski I think adding disabled to the button will be a good extra protection against multiclicks

src/js/components/cycleCount/ConfirmStepHeader.jsx
```diff
@@ -1,26 +1,35 @@
...
-      onClick={save}
-      label="react.default.button.save.label"
-      defaultLabel="Save"
-      variant="primary"
-    />
-  </div>
-);
+  const handleSaveClick = () => {
+    setDisabledSave(true);
+    save();
```
kchelstowski, line 12: and why is not save re-enabled after saving? if an exception is thrown, you would still have the save disabled and would not be able to save again

SebastianLib: I added corrections, let me know if it looks good now

kchelstowski (approved):

PR #5147, opened 2025-03-19
OBPIH-7102 Delete button on new row in count workflow doesn't remove row from the table

alannadolny:


src/js/hooks/cycleCount/useToResolveTab.jsx
```diff
@@ -306,7 +306,7 @@ const useToResolveTab = ({
       ),
       cell: ({ getValue }) => (
         <TableCell className="rt-td">
-          {getValue()}
+          {getValue().toString()}
```
kchelstowski, line 309: what's the purpose of that? Is it some additional change or something related to deleting, which I doubt? :thinking:

kchelstowski (approved):

src/js/hooks/cycleCount/useToResolveTab.jsx
alannadolny: It is an additional request - when the value was 0, it wasn't visible

PR #5148, opened 2025-03-19
OBPIH-7098 Fields should remain enabled after clicking "Save Progress" to trigger proper validation

SebastianLib:

**Description:**

---

### DIFF
--- a/grails-app/migrations/0.8.x/changelog-2019-08-15-1229-alter-table-inventory-snapshot-add-unique-index.xml
+++ b/grails-app/migrations/0.8.x/changelog-2019-08-15-1229-alter-table-inventory-snapshot-add-unique-index.xml
@@ -27,7 +27,7 @@
         </preConditions>
         <comment>Remove duplicates before creating unique constraint on inventory_snapshot.</comment>
         <sql>
-            SET sql_mode=(SELECT REPLACE(@@sql_mode, 'ONLY_FULL_GROUP_BY', ''));
+            SET PERSIST sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));
             CREATE TABLE IF NOT EXISTS inventory_snapshot_new
                 SELECT DISTINCT * FROM inventory_snapshot
                 GROUP BY  date, location_id, product_code, lot_number, bin_location_name;
--- a/grails-app/migrations/0.8.x/changelog-2019-08-15-1229-alter-table-inventory-snapshot-add-unique-index.xml
+++ b/grails-app/migrations/0.8.x/changelog-2019-08-15-1229-alter-table-inventory-snapshot-add-unique-index.xml
@@ -27,7 +27,7 @@
         </preConditions>
         <comment>Remove duplicates before creating unique constraint on inventory_snapshot.</comment>
         <sql>
-            SET PERSIST sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));
+            SET sql_mode=(SELECT REPLACE(@@sql_mode, 'ONLY_FULL_GROUP_BY', ''));
             CREATE TABLE IF NOT EXISTS inventory_snapshot_new
                 SELECT DISTINCT * FROM inventory_snapshot
                 GROUP BY  date, location_id, product_code, lot_number, bin_location_name;
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -529,6 +529,7 @@ class CycleCountService {
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
         cycleCountItem.dateCounted = new Date()
+        cycleCountItem.location = command.binLocation
 
         // We've updated the status of a cycle count item so we need to also update the status of the count.
         recomputeCycleCountStatus(cycleCountItem.cycleCount)
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -14,6 +14,7 @@ import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.importer.CSVUtils
 import org.pih.warehouse.product.Product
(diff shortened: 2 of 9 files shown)


grails-app/migrations/0.8.x/changelog-2019-08-15-1229-alter-table-inventory-snapshot-add-unique-index.xml
```diff
@@ -100,4 +100,4 @@
             SET FOREIGN_KEY_CHECKS = 1;
         </sql>
     </changeSet>
-</databaseChangeLog>
\ No newline at end of file
+</databaseChangeLog>
```
alannadolny, line 103: ?

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -10,6 +14,24 @@ class CycleCountUpdateItemCommand implements Validateable {
 
     boolean recount
 
+    @BindUsing({ obj, source ->
+        def productId = source['inventoryItem']['product'] instanceof Map ? source['inventoryItem']['product']['id'] : source['inventoryItem']['product']
+        def lotNumber = source['inventoryItem']['lotNumber']
+        def expirationDate = source['inventoryItem']['expirationDate'] ? DateUtil.asDate(source['inventoryItem']['expirationDate'].toString()) : null
```
alannadolny, line 20: Use appropriate types instead of `def`

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -10,6 +14,24 @@ class CycleCountUpdateItemCommand implements Validateable {
...
+
+        Product product = Product.read(productId)
+        InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber) ?: new InventoryItem(product: product, lotNumber: lotNumber)
+
+        if (inventoryItem.expirationDate != expirationDate) {
+            inventoryItem.expirationDate = expirationDate
+            inventoryItem.save()
+        }
+
+        inventoryItem
```
kchelstowski, line 30: this is definetely not a place where updating the date should take place.
Binder should only bind the `InventoryItem` and updating the date should take place in the service.

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -10,6 +14,24 @@ class CycleCountUpdateItemCommand implements Validateable {
 
     boolean recount
 
+    @BindUsing({ obj, source ->
+        def productId = source['inventoryItem']['product'] instanceof Map ? source['inventoryItem']['product']['id'] : source['inventoryItem']['product']
+        def lotNumber = source['inventoryItem']['lotNumber']
+        def expirationDate = source['inventoryItem']['expirationDate'] ? DateUtil.asDate(source['inventoryItem']['expirationDate'].toString()) : null
+
+        Product product = Product.read(productId)
+        InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber) ?: new InventoryItem(product: product, lotNumber: lotNumber)
```
kchelstowski, line 23: Let's follow the pattern done in the `CycleCountItemCommand`. Given that it's the same logic in this command and in `CycleCountItemCommand`, let's create a util binder to use them in both places.
You can check out for an example here: https://github.com/openboxes/openboxes/blob/3ed4a8492e6c049a1d2b98d45ffb16416e1dc4b1/src/main/groovy/org/pih/warehouse/outbound/ImportPackingListItem.groovy#L26-L27

kchelstowski (changes requested):

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -529,6 +529,7 @@ class CycleCountService {
         cycleCountItem.countIndex = command.recount ? 1 : 0
         cycleCountItem.status = command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING
         cycleCountItem.dateCounted = new Date()
+        cycleCountItem.location = command.binLocation
```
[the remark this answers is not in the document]
kchelstowski: @EWaterman if it was provided in command via `location`, it would be bound via the line you've attached, but since the names are different, it has to be done like that.
I don't mind it being like this, but if we want to get rid of that line, this would have to be named `location` in the command and sent like that from the client.

src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
EWaterman: Agreed. Also, I think we'd need to discuss with Manon what the desired behaviour is in this scenario. For example:
- you have a row with lot ABC and expiry Jan 1st 2026
- you add a custom row with lot ABC and expiry Feb 2nd 2027

In this scenario, it's unclear if we should actually update the expiry of lot ABC or if the custom row should just take the expiry from the non-custom row (ie use Jan 1st 2026). Updating the expiry and clicking save draft will change the expiry immediately so would impact anywhere else that uses that item.

and another scenario:
- you have a custom row with lot ABC and expiry Jan 1st 2026
- you edit the custom row to have expiry Feb 2nd 2027

in this scenario it might make sense to have the expiry be updated since the item isn't used anywhere else yet.

EWaterman: actually, the first scenario will trigger validation errors on the frontend so shouldn't be possible. You can't have a lot in multiple rows with two different expiry dates.
image

I asked Manon on slack how she wants to handle the second scenario

src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -207,7 +210,9 @@ const useCountStepTable = ({
 
       // Checks if the row is a new one (i.e., added by user and contains 'newRow' in id),
       // and if yes, allow navigation through `newRowFocusableCells`.
-      const isNewRow = (row) => row?.id?.includes('newRow');
+
+
+      const isNewRow = (row) => row?.id?.includes('newRow') || original.custom;
```
SebastianLib, line 215: Since the row order changes after saving progress, arrow navigation doesn't work correctly. Here's a small fix, but it will still need further adjustments. As far as I know, this order is supposed to be fixed anyway?

alannadolny (approved):

PR #5154, opened 2025-03-20
OBPIH-7096 force adjustment transactions to be after product inventory ones

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7096

**Description:** Force any adjustment transactions to always come after the product inventory transaction by setting the transaction date for the product inventory to be before the adjustment transaction date.

---

Before the change (adjustment comes first sometimes):
image

After (adjustment is always after):
image

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
@@ -2,7 +2,9 @@ package org.pih.warehouse.inventory
 
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
+import java.time.Instant
 
+import org.pih.warehouse.DateUtil
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.product.Product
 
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountTransactionService.groovy
```diff
@@ -27,13 +29,20 @@ class CycleCountTransactionService {
...
         List<Transaction> transactions = []
-        Date transactionDate = new Date()
 
-        List<Transaction> productInventoryTransactions = createProductInventoryTransactions(cycleCount, transactionDate)
+        // Set the product inventory transaction date to be one second before any adjustment transactions. This
+        // transaction date offset is necessary to guarantee that the transactions are applied in the correct order.
+        Instant now = Instant.now()
+        // -1 (instead of +1 to the adjustment transaction date) because transaction date can't be in the future.
+        Date productInventoryTransactionDate = DateUtil.asDate(now.minusSeconds(1))
+        Date adjustmentTransactionDate = DateUtil.asDate(now)
```
EWaterman, line 38: FYI I'm pretty sure this all comes down to the following method call in InventoryService when compuing product availability:
```
List getTransactionEntriesByInventoryAndProduct(Inventory inventory, List<Product> products) {
    def transactionEntries = criteria.list {
        transaction {
            eq("inventory", inventory)
            order("transactionDate", "asc")
            order("dateCreated", "asc")
        }
        if (products) {
            inventoryItem { inList("product", products) }
        }
    }
    return transactionEntries
}
```
so transactionDate is used to determine the order. It seems the order is unpredictable when there are two transactions with the same transactionDate and dateCreated. Sometimes it's correct, and other times the adjustment comes first. This change (hopefully) makes the adjustment always come second

kchelstowski (approved):

awalkowiak (commented):
Wouldn't just using a `new Date()` in each create method work here as well (instead of using the one assigned at the beginning)?

EWaterman:
@awalkowiak I wanted to be more explicit about it. If we just do `new Date()` then if it all happens within the same millisecond, we have the same problem (they'll be the same date).

But more importantly, we use DATETIME as the field type for transaction_date, which only has precision up to seconds, so we have to explicitly set them one second apart for them to actually be different.

awalkowiak (approved):

PR #5156, opened 2025-03-21
OBPIH-6899 Load the resolve discrepancies step page (Fix after QA)

alannadolny:


SebastianLib (approved):

src/js/hooks/cycleCount/useResolveStepTable.jsx
```diff
@@ -442,7 +442,7 @@ const useResolveStepTable = ({
         </TableHeaderCell>
       ), []),
       cell: useCallback(({ row: { original: { quantityVariance, quantityCounted } } }) => {
-        const variant = quantityCounted
+        const variant = (quantityCounted || quantityCounted === 0)
```
EWaterman, line 445: why not just `quanitityCounted != null`?

EWaterman (approved):
tested and confirmed the fix. Could implement my comment but also it's fine as is

image

PR #5157, opened 2025-03-21
OBPIH-7122 Recount workflow validations

SebastianLib:

**Description:**
This PR fix not display the same items. It combines the first count **countIndex = 0** with the first recount **countIndex = 1** into one record, keeping details from both. Any extra **countIndex = 1** records for the same item stay separate, not merged together. So, if someone adds lots of the same item, you get one merged record plus the rest as individual ones
Here is proof that it works correctly: https://jam.dev/c/e0e97537-fc7c-4fc6-80cb-080315e35094

---

### DIFF
--- a/src/js/hooks/cycleCount/useResolveStep.js
+++ b/src/js/hooks/cycleCount/useResolveStep.js
@@ -115,7 +115,7 @@ const useResolveStep = () => {
       return itemsFromResolve.map((item, index) => {
         // If this is the first item with maxCountIndex and itemFromCount exists,
         // merge data from both cycles. We do this because we wanted to combine data from
-        // countIndex 0 with the first countIndex 1 to retain info from both counts.
+        // countIndex 0 with the first countIndex 1 to avoid duplicating info from both counts
         if (index === 0 && itemFromCount) {
           return {
             ...itemFromCount,
@@ -132,9 +132,9 @@ const useResolveStep = () => {
           };
         }
 
-        // For example countIndex 0, 1, 1: the item with
+        // For example we have countIndex 0, 1, 1: the item with
         // countIndex 0 merges with the first item with countIndex 1,
-        // while the next item with countIndex 1 remains a separate record without merging.
+        // while the next item with countIndex 1 remains a separate record without merging
         return {
           ...item,
           quantityRecounted: item?.quantityCounted,
--- a/src/js/hooks/cycleCount/useResolveStep.js
+++ b/src/js/hooks/cycleCount/useResolveStep.js
@@ -89,10 +89,11 @@ const useResolveStep = () => {
   const mergeCycleCountItems = (items) => {
     const duplicatedItems = _.groupBy(items,
       (item) => `${item.binLocation?.id}-${item?.inventoryItem?.lotNumber}`);
-    return Object.values(duplicatedItems).map((itemsToMerge) => {
+
+    return Object.values(duplicatedItems).flatMap((itemsToMerge) => {
       if (itemsToMerge.length === 1) {
         const item = itemsToMerge[0];
-        return {
+        return [{
           ...item,
           quantityRecounted: item?.quantityCounted,
           dateRecounted: item?.dateCounted,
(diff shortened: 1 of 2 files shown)


src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -103,25 +104,50 @@ const useResolveStep = () => {
...
+            dateRecounted: item?.dateCounted,
+            countedBy: itemFromCount?.assignee,
+            recountedBy: item?.assignee,
+            rootCause: mapRootCauseToSelectedOption(item?.discrepancyReasonCode),
+          };
+        }
+
+        // For example we have countIndex 0, 1, 1: the item with
+        // countIndex 0 merges with the first item with countIndex 1,
+        // while the next item with countIndex 1 remains a separate record without merging
```
EWaterman, line 137: is this really the desired behaviour? Why don't we want to display the rows as identical? What happens if the user deletes the other one (the one with more information)?

I think we should confirm the desired behaviour with Manon.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -89,10 +89,11 @@ const useResolveStep = () => {
   const mergeCycleCountItems = (items) => {
     const duplicatedItems = _.groupBy(items,
       (item) => `${item.binLocation?.id}-${item?.inventoryItem?.lotNumber}`);
-    return Object.values(duplicatedItems).map((itemsToMerge) => {
+
+    return Object.values(duplicatedItems).flatMap((itemsToMerge) => {
```
awalkowiak, line 93: @SebastianLib can you explain the changes in this PR? What was the issue and how you are fixing that.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -103,25 +104,50 @@ const useResolveStep = () => {
...
-        quantityRecounted: itemFromResolve?.quantityCounted,
-        quantityCounted: itemFromCount?.quantityCounted,
-        quantityVariance: itemFromCount?.quantityVariance,
-        dateCounted: itemFromCount?.dateCounted,
-        dateRecounted: itemFromResolve?.dateCounted,
-        countedBy: itemFromCount?.assignee,
-        recountedBy: itemFromResolve?.assignee,
-        rootCause: mapRootCauseToSelectedOption(itemFromResolve?.discrepancyReasonCode),
-      };
+      const itemsFromResolve = groupedByCountIndex[maxCountIndex] || [];
```
EWaterman, line 113: why not just:
```
const itemsFromResolve = _.filter(itemsToMerge, (item) => item.countIndex === maxCountIndex);
```
or I suppose you could do:
```
const groupedByCountIndex = _.groupBy(itemsToMerge, 'countIndex');
const maxCountIndex = _.maxBy(itemsToMerge, 'countIndex').countIndex;
const itemFromCount = groupedByCountIndex[0][0];  // We  know there will always be exactly one matching count item
const itemsFromResolve = groupedByCountIndex[maxCountIndex] || [];
```
but it's a bit weird with the itemFromCount (you probably want some null checks)

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -103,25 +104,50 @@ const useResolveStep = () => {
...
+
+      return itemsFromResolve.map((item, index) => {
+        // If this is the first item with maxCountIndex and itemFromCount exists,
+        // merge data from both cycles. We do this because we wanted to combine data from
+        // countIndex 0 with the first countIndex 1 to avoid duplicating info from both counts
+        if (index === 0 && itemFromCount) {
```
alannadolny, line 119: why only the first item? I think I am not getting the idea of it. Why the previous version was wrong?

SebastianLib: The first countIndex 0 merges with the first countIndex 1 to avoid duplicating the initial count data, while the rest of the countIndex 1 items for the same product are added separately afterward. This fixes the issue where, if someone saves, say, 5 identical items, they’d only see 1 merged result instead of duplicates

src/js/hooks/cycleCount/useResolveStep.js
SebastianLib: unfortunately it doesn't work `const itemFromCount = groupedByCountIndex[0][0]`

EWaterman: yeah it was just pseudocode. Seems redundant to build the full groupedByCountIndex but only use it to get resolve items, which is why I suggested either using it for count items too, or just using a filter.

Anyways, it's just a suggestion. Not a requirement

src/js/hooks/cycleCount/useResolveStep.js
EWaterman: This is the behaviour I want to confirm with Manon. I personally don't see a need to distinguish between rows in this way. They should all display the same data. The fix is only to not merge the recount rows with each other but they can still all contain data from the original count.

You won't be able to proceed with rows that look like this anyways since it will fail validation on unique [bin + lot] so no need to over complicate the logic

So I'd simplify the logic down to just:
```
return itemsFromResolve.map((item, index) => {
    return {
          ...itemFromCount,
          ...item,
          commentFromCount: itemFromCount?.comment,
          quantityRecounted: item?.quantityCounted,
          quantityCounted: itemFromCount?.quantityCounted,
          quantityVariance: itemFromCount?.quantityVariance,
          dateCounted: itemFromCount?.dateCounted,
          dateRecounted: item?.dateCounted,
          countedBy: itemFromCount?.assignee,
          recountedBy: item?.assignee,
          rootCause: mapRootCauseToSelectedOption(item?.discrepancyReasonCode),
    };
});
```

SebastianLib: yes we need discuss about this behaviour with Manon and your simplified logic also works. Thanks!

EWaterman (approved):
Tested and it seems like there's still some inconsistent behaviour around recount difference (see the "Small gray tent" rows)

image

PR #5158, opened 2025-03-21
OBPIH-7107 Improve performance for candidates on to count and to reso…

kchelstowski:
…lve tabs

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -19,8 +19,8 @@ import org.pih.warehouse.inventory.CycleCountStartRecountBatchCommand
 import org.pih.warehouse.inventory.CycleCountStatus
 import org.pih.warehouse.inventory.CycleCountSubmitCountCommand
 import org.pih.warehouse.inventory.CycleCountSubmitRecountCommand
-import org.pih.warehouse.inventory.CycleCountCandidateInProgress
 import org.pih.warehouse.inventory.CycleCountUpdateItemCommand
+import org.pih.warehouse.inventory.PendingCycleCountRequest
 
 class CycleCountApiController {
 
@@ -39,10 +39,10 @@ class CycleCountApiController {
         render([data: cycleCounts, totalCount: cycleCounts.totalCount] as JSON)
     }
 
-    def getCandidatesInProgress(CycleCountCandidateFilterCommand filterParams) {
-        List<CycleCountCandidateInProgress> tabRecords = cycleCountService.getCandidatesInProgress(filterParams, params.facilityId)
+    def getPendingCycleCountRequests(CycleCountCandidateFilterCommand filterParams) {
+        List<PendingCycleCountRequest> pendingCycleCountRequests = cycleCountService.getPendingCycleCountRequests(filterParams, params.facilityId)
 
-        render([data: tabRecords, totalCount: tabRecords.totalCount] as JSON)
+        render([data: pendingCycleCountRequests, totalCount: pendingCycleCountRequests.totalCount] as JSON)
     }
 
     def createRequests(CycleCountRequestBatchCommand command) {
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -19,6 +19,7 @@ import org.pih.warehouse.inventory.CycleCountStartRecountBatchCommand
 import org.pih.warehouse.inventory.CycleCountStatus
 import org.pih.warehouse.inventory.CycleCountSubmitCountCommand
 import org.pih.warehouse.inventory.CycleCountSubmitRecountCommand
+import org.pih.warehouse.inventory.CycleCountCandidateInProgress
 import org.pih.warehouse.inventory.CycleCountUpdateItemCommand
 
 class CycleCountApiController {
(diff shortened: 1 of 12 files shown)


grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidateInProgress.groovy
```diff
@@ -0,0 +1,84 @@
...
+import org.pih.warehouse.core.Tag
+import org.pih.warehouse.core.User
+import org.pih.warehouse.product.Product
+
+
+class CycleCountCandidateInProgress {
+
+    String id
+
+    CycleCountRequest cycleCountRequest
```
kchelstowski, line 13: actually all the fields below come from the `cycleCountRequest`. This `cycleCountRequest` is quite redundant because e.g. `facility == cycleCountRequest.facility` etc., but I've included the `cycleCountRequest` separately here, not to have to refactor frontend, as the previous API returned the `cycleCountRequest` and we were looking for ids of those.
I felt like it was a bit risky to touch the frontend in that matter to push the draft, so I left it like that for now.
It doesn't hurt the performance much, so if you are fine with it for now, let's leave it, otherwise, I can do the appropriate refactor on frontend on Monday.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -100,6 +100,66 @@ class CycleCountService {
...
+                inList("status", command.statuses)
+            }
+
+            if (command.negativeQuantity) {
+                gt("negativeItemCount", 0)
+            }
+
+            applySortOrder(command.sort, command.order, delegate, usedAliases)
+
+        } as List<CycleCountCandidateInProgress>
```
kchelstowski, line 160: I know the code is pretty much the same (except a few filters) as in the candidates endpoint, but I felt like it's "form over substance" to make those two service methods reusable with some if conditions. I feel like sometimes it's better to copy the code even if two places have to be refactored in the future, than to spend days thinking about how to make it reusable. Those are two easy list methods so I assume it's better to live with such version, but let me know your thoughts.

grails-app/migrations/views/cycle-count-candidate-in-progress.sql
```diff
@@ -0,0 +1,49 @@
...
+        cycle_count_request ccr
+            LEFT OUTER JOIN product_availability pa
+                ON ccr.product_id = pa.product_id
+                AND ccr.facility_id = pa.location_id
+            LEFT OUTER JOIN cycle_count cc
+                ON ccr.cycle_count_id = cc.id
+            JOIN location ON pa.location_id = location.id
+            LEFT OUTER JOIN product_classification
+                         ON pa.product_id = product_classification.product_id AND
+                            location.inventory_id = product_classification.inventory_id
```
kchelstowski, line 36: I now realized I forgot to add `WHERE` clause for statuses (to filter out completed and canceled requests). Will add it on Monday

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -38,6 +39,12 @@ class CycleCountApiController {
         render([data: cycleCounts, totalCount: cycleCounts.totalCount] as JSON)
     }
 
+    def getCandidatesInProgress(CycleCountCandidateFilterCommand filterParams) {
+        List<CycleCountCandidateInProgress> tabRecords = cycleCountService.getCandidatesInProgress(filterParams, params.facilityId)
+
+        render([data: tabRecords, totalCount: tabRecords.totalCount] as JSON)
```
EWaterman, line 45: don't you also need to support exporting as csv like we have with the above method?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
EWaterman: I agree. Just because two things have essentially the same code, doesn't mean we should force them to use a common method. Sometimes they just do similar things and that's okay.

I think there's a way to look at it where CycleCountCandidate is an "enhanced" version of CycleCountCandidateInProgress, but yeah it feels a little forced.

If anything, I'd build a CycleCountXYZ interface that CycleCountCandidate and CycleCountCandidateInProgress can both extend from, which would contain all the common fields. But I do agree with you that it might be overkill for right now, especially since we want to make this change as simple as possible.

Maybe we can revisit it post-release.

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidateInProgress.groovy
```diff
@@ -0,0 +1,84 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.core.Tag
+import org.pih.warehouse.core.User
+import org.pih.warehouse.product.Product
+
+
+class CycleCountCandidateInProgress {
```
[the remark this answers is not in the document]
EWaterman: I'm not a huge fan of the name (though I like it better than CycleCountTab for the reason that you said) but I'm struggling to come up with anything better.

"Candidate in progress" to me implies a relationship to the candidate table, but there isn't one.

I thought maybe just "CycleCountProgress" but I don't know... all these different view names are quite confusing to me. It's hard to piece out what is what.

"CycleCountState" ??

Idk I'll think about it

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidateInProgress.groovy
EWaterman: hm interesting. Is the idea that we want this object to be distinct from CycleCountRequest? So if a field is added to CycleCountRequest, we aren't forced to also add it to CycleCountCandidateInProgress. I can see the positives and negatives of this. It seems simpler to just shove the request in there so that we don't need to remember to modify the class or the sql if we want to return an additional field, but then again I do like letting this be its own thing. But maybe all it's doing is hiding the fact that it's dependent on CycleCountRequest by moving that to the sql and so it'd be better to be more visible about the dependency.

These are just some loose thoughts. I'd be curious to hear what the others think

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: I forgot to mention that I left that for now, to push the initial draft. Moreover - the csv returns last counted date, which Justin said probably said caused the most issues, and as long as we don't need it, we are fine, so we can still use the old endpoint for the csv for now.

awalkowiak (approved):

PR #5159, opened 2025-03-22
OBPIH-7101 FIX. Handle null when comparing expiration dates

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7101

**Description:** Duplicate expiry date validation error was incorrectly triggering when two bins had the same lot and expiration date was null. https://jam.dev/c/1e1680e8-e3c6-44b9-994d-46d2b19af8e1

---

Fixed:
image

### DIFF
--- a/src/js/hooks/cycleCount/useInventoryValidation.js
+++ b/src/js/hooks/cycleCount/useInventoryValidation.js
@@ -58,7 +58,10 @@ const useInventoryValidation = ({ tableData }) => {
         const expirationDate = item?.inventoryItem?.expirationDate;
         return expirationDate == null ? null : moment.utc(expirationDate);
       });
-      const uniqueDates = _.uniqWith(expirationDates, (arrVal, othVal) => arrVal?.isSame(othVal));
+      const uniqueDates = _.uniqWith(
+        expirationDates,
+        (arrVal, othVal) => (arrVal == null ? othVal == null : arrVal.isSame(othVal)),
+      );
       if (uniqueDates.length > 1) {
         ctx.addIssue({
           code: z.ZodIssueCode.custom,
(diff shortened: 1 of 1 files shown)


src/js/hooks/cycleCount/useInventoryValidation.js
```diff
@@ -58,7 +58,10 @@ const useInventoryValidation = ({ tableData }) => {
         const expirationDate = item?.inventoryItem?.expirationDate;
         return expirationDate == null ? null : moment.utc(expirationDate);
       });
-      const uniqueDates = _.uniqWith(expirationDates, (arrVal, othVal) => arrVal?.isSame(othVal));
```
EWaterman, line 61: This was returning null when expiration date is null, which is falsy, which uniqWith treats as "not the same", hence the error.

src/js/hooks/cycleCount/useInventoryValidation.js
```diff
@@ -58,7 +58,10 @@ const useInventoryValidation = ({ tableData }) => {
...
+      const uniqueDates = _.uniqWith(
+        expirationDates,
+        (arrVal, othVal) => (arrVal == null ? othVal == null : arrVal.isSame(othVal)),
```
kchelstowski, line 63:
```suggestion
        (arrVal, othVal) => (arrVal === null ? othVal === null : arrVal.isSame(othVal)),
```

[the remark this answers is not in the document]
EWaterman: from what I read, `x == null` was the suggestion because also returns true if x is undefined. I made it '==' instead of '===' to cover a more broad range of values. But I trust your opnion so I'm okay with '=='.

PR #5160, opened 2025-03-24
OBPIH-7100 Fix and improve deleting new lines on recount form (Fix ordering)

alannadolny:


src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -125,13 +125,30 @@ const useResolveStep = () => {
     });
   };
 
+  const moveCustomItemsToTheBottom = (cycleCountItems) => {
```
awalkowiak, line 128: Can you add a simple comment on why this is here? (something about that we want to keep ordering of items that was before a save action, to display the validation properly, and that when we come back to it, then we don't care about this ordering)

awalkowiak (approved):

PR #5166, opened 2025-03-25
OBPIH-6989 ability to sort delivery note by product on stock movement details page

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-6989

**Description:** Change the delivery note download dropdown on the documents tab of the stock movement details page to support sorting by order index and by product name.

---

A recording I made for Kelsey that demos the behaviour: https://jam.dev/c/780f9f32-57af-4aa6-be16-63e5bb81147d

image

image


src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
```diff
@@ -789,6 +789,20 @@ class SendMovementPage extends Component {
...
                           }
+
+                          // Some documents have multiple download options so list them all out.
+                          if (document.downloadOptions) {
+                            return _.map(document.downloadOptions, (downloadOption) => (
+                              <DocumentButton
+                                link={downloadOption.uri}
+                                buttonTitle={downloadOption.name}
+                                {...document}
+                                key={idx}
```
EWaterman, line 800: I'm not sure if we even need this line. It works fine with it, but I'm not certain what it's actually doing (I just copied it from the below button). Maybe someone with more frontend experience can answer

alannadolny: https://legacy.reactjs.org/docs/lists-and-keys.html#keys

grails-app/views/stockMovement/_documents.gsp
```diff
@@ -52,9 +52,27 @@
                                         <warehouse:message code="default.button.delete.label"/>
                                     </g:link>
                                 </g:if>
-                                <g:link url="${document.uri}" target="_blank" class="button">
-                                    <warehouse:message code="default.button.download.label"/>
-                                </g:link>
+                                <g:if test="${document.downloadOptions}">
+                                    <span class="action-menu">
+                                        <button class="action-btn button">
+                                            <warehouse:message code="default.button.download.label"/>
```
alannadolny, line 58: I know that you are not the person who added this line, but we are using g:message
```suggestion
                                            <g:message code="default.button.download.label"/>
```

alannadolny (approved):

src/js/components/stock-movement-wizard/inbound/SendMovementPage.jsx
```diff
@@ -764,6 +764,20 @@ class SendMovementPage extends Component {
...
                         }
+
+                        // Some documents have multiple download options so list them all out.
+                        if (document.downloadOptions) {
+                          return _.map(document.downloadOptions, (downloadOption) => (
+                            <DocumentButton
+                              link={downloadOption.uri}
+                              buttonTitle={downloadOption.name}
+                              {...document}
+                              key={idx}
```
kchelstowski, line 775: index shouldn't be a key. Moreover - this map is nested, hence you will have the same key for those elements.

EWaterman: why is it bad to make it the index? I don't think this array is mutable (and so indexes don't change). It just uses data returned from the backend.

I suppose I could change the key to be the document/download name

kchelstowski: https://www.youtube.com/watch?v=xlPxnc5uUPQ

+ the key should be unique. since you have nested `.map`, the `idx` will be the same for each download option (you use` idx` from the nested map)

EWaterman: yes I understand that part and I'll fix that to make the index actually unique.

I watched the video and it makes sense (thanks for sharing). I was wondering if it's fine to use index in this case since the array shouldn't be changing, but actually I'm not certain about it anymore since there's the ability to add/remove documents from this page. I don't know what the behaviour is in that situation.

But that's a pre-existing concern since this was already using index as the id. And there isn't a better option right now in the payload IMO. The only other possible field to use is name (which is translated and not guaranteed to be unique). I think I'm going to make it work with ids as it was before and add a TODO to fix it to provide a proper unique id.

grails-app/views/stockMovement/_documents.gsp
EWaterman: I'm going to leave it as is since warehouse:message is used everywhere in the file and I want to refactor as little as possible right now, but that's good to know

kchelstowski (approved):

PR #5167, opened 2025-03-26
OBPIH-7125 Trim spaces for lot number while creating a cycle count item

kchelstowski:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useCountStep.jsx
+++ b/src/js/hooks/cycleCount/useCountStep.jsx
@@ -29,6 +29,7 @@ import { DateFormat } from 'consts/timeFormat';
 import useCountStepValidation from 'hooks/cycleCount/useCountStepValidation';
 import useSpinner from 'hooks/useSpinner';
 import confirmationModal from 'utils/confirmationModalUtils';
+import trimLotNumberSpaces from 'utils/cycleCountUtils';
 import exportFileFromApi from 'utils/file-download-util';
 import { checkBinLocationSupport } from 'utils/supportedActivitiesUtils';
 
@@ -302,7 +303,9 @@ const useCountStep = () => {
       show();
       resetValidationState();
       for (const cycleCount of tableData.current) {
-        const cycleCountItemsToUpdate = cycleCount.cycleCountItems.filter((item) => (item.updated && !item.id.includes('newRow')));
+        const cycleCountItemsToUpdate = cycleCount.cycleCountItems
+          .filter((item) => (item.updated && !item.id.includes('newRow')))
+          .map(trimLotNumberSpaces);
         for (const cycleCountItem of cycleCountItemsToUpdate) {
           await cycleCountApi.updateCycleCountItem(
             getPayload(cycleCountItem, cycleCount),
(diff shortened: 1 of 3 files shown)


alannadolny (approved):

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -302,15 +303,19 @@ const useCountStep = () => {
       show();
       resetValidationState();
       for (const cycleCount of tableData.current) {
-        const cycleCountItemsToUpdate = cycleCount.cycleCountItems.filter((item) => (item.updated && !item.id.includes('newRow')));
+        const cycleCountItemsToUpdate = cycleCount.cycleCountItems
+          .filter((item) => (item.updated && !item.id.includes('newRow')))
+          .map(trimLotNumberSpaces);
```
EWaterman, line 308: so you're also trimming the lot number when updating non-custom rows? What happens if we have a pre-existing item in our system with lot `"ABC   "` (some whitespace on the end). Won't updating the countItem now create a new inventory item for the lot `"ABC"` (no spaces)?

It might seem like an edge case but we actually have 1261 rows like that:
```
SELECT id, lot_number FROM inventory_item WHERE lot_number LIKE "% ";
```

Long term ideally we can migrate our existing data and stop allowing trailing whitespace for lot, but maybe the simplest short term solution is to only trim the lot number on items to update if the lot number has changed, otherwise leave it as is

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -302,15 +303,19 @@ const useCountStep = () => {
...
           await cycleCountApi.updateCycleCountItem(
             getPayload(cycleCountItem, cycleCount),
             currentLocation?.id,
             cycleCountItem?.id,
           );
         }
-        const cycleCountItemsToCreate = cycleCount.cycleCountItems.filter((item) => item.id.includes('newRow'));
+        const cycleCountItemsToCreate = cycleCount.cycleCountItems
+          .filter((item) => item.id.includes('newRow'))
+          .map(trimLotNumberSpaces);
```
EWaterman, line 318: similar to my other comment, trimming whitespace here will prevent you from being able to use an existing lot  `"ABC  "` since it will always force the lot to be `"ABC"`.

I think this is fine though because to me it's a data issue that should be fixed. We shouldn't have lots in our system with trailing whitespace. That's confusing for users.

I think it's worth doing an investigation (in a different ticket) to see if there's an easy way to strip whitespace from all string fields coming into the frontend as well as all string fields going to the db. I can't see a situation where we'd need to allow trailing whitespace so maybe there's an easy way to remove it from our system entirely

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: > What happens if we have a pre-existing item in our system with lot "ABC   " (some whitespace on the end). Won't updating the countItem now create a new inventory item for the lot "ABC" (no spaces)?

We discussed this case on our internal standup in the morning with Manon and we confirmed this would eventually create a new inventory item, but it would be expected, as the "dirty data" should eventually be removed by a user.

EWaterman: cool as long as it's known and was discussed :+1:

EWaterman (approved):

PR #5172, opened 2025-03-27
OBPIH-7114 Date picker selects day before if your local timezone is behind UTC

alannadolny:


src/js/utils/dateUtils.js
```diff
@@ -0,0 +1,16 @@
...
+const dateWithoutTimeZone = ({
+  date,
+  currentDateFormat,
+  outputDateFormat,
+}) => {
+  if (!date) {
+    return null;
+  }
+
+  const parsedDate = moment(date, currentDateFormat).utcOffset(0, true);
```
EWaterman, line 12: adding a comment here because this is relevant to the upcoming discovery on dates (and Alan if I'm wrong, please correct me): `.utcOffset(0, true)` will convert the given moment into the same exact date and time but in UTC.

So if given `"2000-01-01T00:00+01:00"` it will output `"2000-01-01T00:00Z"`, so functionally it strips out timezone info and replaces it with UTC.

This will be useful more generally, since we likely want to standardize the following:
- user provides input via date picker or data import in user local timezone
- we convert that to a utc moment for internal processing/validation
- we either convert that to utc string for sending to backend or to local timezone string for display

src/js/utils/dateUtils.js
```diff
@@ -0,0 +1,16 @@
+import moment from 'moment';
+
+const dateWithoutTimeZone = ({
```
EWaterman, line 3: I'm fine with this for now since likely the discovery will revisit this util file, but this isn't really a date without a timezone, it's a date in utc with the same time. So maybe a name similar to "dateInUtcPreserveTime" makes more sense

EWaterman (approved):

PR #5175, opened 2025-03-28
OBPIH-7144 Trigger validation on next when add lot with spaces

alannadolny:


EWaterman:
I talked myself out of this mid way through typing it out but leaving it in anyways for reference.

an alternative to doing this trimming during validation would be to do it:
- when processing any responses from the backend that return inventory item data (ie get, create, update, submit, refresh cycle count). That way, if the server gives us lots with spaces, they get trimmed out right away before getting any further into the frontend logic
- when taking in user input. That way if the user accidentally types a space, it gets immediately trimmed out, again before getting any further into the frontend. It'd be just like what we're doing in useCountStepTable to round decimals to integers: https://github.com/openboxes/openboxes/pull/5141

It probably works the way you have it, but if we do the trimming right when the lot enters the frontend, it would guarantee that our lot numbers are correct everywhere so we wouldn't need to remember to do this .trim() check any time we want to do a comparison of lots.

I know that's not a perfect solution either because we have to modify the responses to all our backend calls, but I see it as a temporary solution until we can fix the backend to also always trim lot number.

Ideally we do the following:
- backend strips whitespace for all user input string fields ALWAYS.
- we migrate old data to strip white space from varchar fields
- the frontend strips whitespace from user input fields ALWAYS

but until we get to that point I suppose any solution is going to be a temporary one... I suppose I'm fine with what you have for now.

EWaterman:
Tested:
- start a count on a product that has a lot with no space.
    - :heavy_check_mark: Add a custom row with the same lot but with a space -> fails validation as expected
    - :heavy_check_mark: Add a custom row with the same lot with no space -> allows you to proceed
- :heavy_check_mark: same for recount
- start a count on a product that has a lot with a space.
    - :heavy_check_mark: Add a custom row with the same lot with a space -> fails validation as expected
    - :x: Add a custom row with the same lot with no space -> fails validation!

Here's a recording of me trying to test that last case: https://jam.dev/c/902095e1-63dd-4b89-9d4b-e78e45bebc56

this is the query I ran on my local to find lots with a trailing space to test against:
```
SELECT location.name, product_availability.product_id, product_availability.lot_number, product_availability.bin_location_name
FROM product_availability
JOIN location ON location.id=product_availability.location_id
WHERE
    location.active
    AND product_availability.quantity_on_hand != 0
    AND product_availability.inventory_item_id in (SELECT inventory_item_id
                                                   FROM inventory_item
                                                   WHERE lot_number like "% ")
LIMIT 30;
```

awalkowiak (approved):

kchelstowski (approved):

PR #5178, opened 2025-03-31
OBPIH-7146 Add import template/export template buttons to add items page

SebastianLib:

**Description:**
I set this as a draft because I'm still waiting for confirmation from Manon regarding one behavior in the import templates. I also added changes from the OBPIH-6960-2 branch to this draft because I noticed that when I did the rebase, those changes weren't included. I probably did something wrong with the rebase since this is my first time doing it. I always used git merge before
---

### DIFF
--- a/src/js/components/DataTable/DataTable.scss
+++ b/src/js/components/DataTable/DataTable.scss
@@ -14,13 +14,16 @@
   border: none;
   border-radius: 4px;
 
+  .overflow-visible {
+    overflow: visible !important;
+  }
+
   &.hide-data .rt-tbody {
     display: none;
   }
 
   .rt-table {
     font-size: 14px;
-    overflow: visible;
 
     &::-webkit-scrollbar {
       width: 0.5em;
--- a/src/js/components/DataTable/v2/DataTable.jsx
+++ b/src/js/components/DataTable/v2/DataTable.jsx
@@ -41,7 +41,7 @@ const DataTable = ({
   return (
     <div className="app-react-table-wrapper table-v2">
       <div className="ReactTable app-react-table">
-        <div className="rt-table" role="grid">
+        <div className="rt-table overflow-visible" role="grid">
           <DataTableHeader
             headerGroups={table.getHeaderGroups()}
           />


src/js/api/urls.js
```diff
@@ -28,6 +28,8 @@ export const PICKLIST_ITEMS_EXPORT = (id) => `${STOCK_MOVEMENT_API}/exportPickLi
 export const PICKLIST_TEMPLATE_EXPORT = (id) => `${STOCK_MOVEMENT_API}/picklistTemplate/${id}`;
 export const PICKLIST_IMPORT = (id) => `${STOCK_MOVEMENT_API}/importPickListItems/${id}`;
 export const PACKING_LIST_TEMPLATE = `${STOCK_MOVEMENT_API}/packingList/template`;
+export const STOCK_MOVEMENT_IMPORT_CSV = (id) => `/stockMovement/importCsv/${id}`;
+export const STOCK_MOVEMENT_EXPORT_CSV = (id) => `/stockMovement/exportCsv/${id}`;
```
alannadolny, line 32: If the URL is connected to the app view or other frontend thing, we should use `applicationUrl.js`
This file + services should be used for API calls.

src/js/components/DataTable/DataTable.scss
```diff
@@ -20,6 +20,7 @@
 
   .rt-table {
     font-size: 14px;
+    overflow: visible;
```
alannadolny, line 23: Won't it affect other tables?

src/js/components/DataTable/TableHeaderCell.jsx
```diff
@@ -20,6 +24,27 @@ const TableHeaderCell = ({
...
+            <div className="sorting-arrows">
+              <RiArrowUpSFill className="arrow-up" />
+              <RiArrowDownSFill className="arrow-down" />
+            </div>
+          )}
+        </div>
+      </CustomTooltip>
+    );
+  }
+
```
alannadolny, line 47: What is this code doing? I am trying to figure it out, but i can't :smile:

src/js/components/stock-movement-wizard/inboundV2/sections/InboundV2AddItems.jsx
```diff
@@ -97,6 +107,38 @@ const InboundV2AddItems = ({
             />
 
             <div className="buttons-container">
+              <Button
+                onClick={() => {
+                  const fileInput = document.getElementById('csvInput');
+                  fileInput.value = null;
```
alannadolny, line 113: Why do we need to set the null here?

src/js/hooks/inboundV2/addItems/useInboundAddItemsForm.jsx
```diff
@@ -576,6 +588,60 @@ const useInboundAddItemsForm = ({
...
+      const { stockMovementId } = getValues().values;
+
+      formData.append('importFile', file.slice(0, file.size, 'text/csv'));
+      const config = {
+        headers: {
+          'content-type': 'multipart/form-data',
+        },
+      };
+
+      await apiClient.post(STOCK_MOVEMENT_IMPORT_CSV(stockMovementId), formData, config);
```
alannadolny, line 605: You can move this call to a service + Why do we need to use the `newFormData`? Can't it be done without it?

src/js/hooks/inboundV2/addItems/useInboundAddItemsForm.jsx
```diff
@@ -576,6 +588,60 @@ const useInboundAddItemsForm = ({
...
+      'Are you sure you want to refresh? Your progress since last save will be lost.',
+    );
+  };
+
+  const saveItemsAndExportTemplate = async (formValues, lineItems) => {
+    spinner.show();
+    const { movementNumber, stockMovementId } = formValues;
+    try {
+      await saveRequisitionItemsInCurrentStep(lineItems);
+      const response = await apiClient.get(STOCK_MOVEMENT_EXPORT_CSV(stockMovementId), { responseType: 'blob' });
```
alannadolny, line 633: this call also can be moved the the service

src/js/components/stock-movement-wizard/inboundV2/sections/InboundV2AddItems.jsx
SebastianLib: I set it to null here because without this, when we click the button the first time, everything works fine, but the second time nothing happens. So each time we click it, we need to reset fileInput to null

src/js/components/DataTable/DataTable.scss
SebastianLib: This is necessary for the tooltip in the header to display correctly. It will affect other tables as well, but it's not visible right now because, for example, in CycleCount we don't have a tooltip in the header. 

with overflow: visible 
image

without:
image

src/js/components/DataTable/TableHeaderCell.jsx
SebastianLib: Before this change, we didn’t have any tooltip for the header, and now I’ve wrapped it in CustomTooltip

src/js/hooks/inboundV2/addItems/useInboundAddItemsForm.jsx
SebastianLib: I will check this, but I copied this from old inbound 
image

SebastianLib: I tested and in theory, it could be done without **FormData**, but I’m not sure if it’s worth changing it in the inbound now since this way works. I’m worried that something could accidentally break if I will mess with it

src/js/components/DataTable/TableHeaderCell.jsx
```diff
@@ -20,6 +24,27 @@ const TableHeaderCell = ({
...
+        >
+          {children}
+          {required && <span className="ml-1 required">&#42;</span>}
+          {sortable && (
+            <div className="sorting-arrows">
+              <RiArrowUpSFill className="arrow-up" />
+              <RiArrowDownSFill className="arrow-down" />
+            </div>
+          )}
+        </div>
```
kchelstowski, line 43: let's wrap lines 30-43 in a separate component since it's used also below. By doing so we won't have to remember to refactor two places.

src/js/components/stock-movement-wizard/inboundV2/sections/InboundV2AddItems.jsx
kchelstowski: add a comment explaining that null

src/js/hooks/inboundV2/addItems/useInboundAddItemsForm.jsx
```diff
@@ -260,11 +265,15 @@ const useInboundAddItemsForm = ({
       try {
         spinner.show();
         const resp = await apiClient.post(STOCK_MOVEMENT_UPDATE_ITEMS(queryParams.id), payload);
+        const { data } = resp.data;
         const transformedData = {
-          ...resp.data.data,
-          lineItems: resp.data.data.lineItems?.map(transformLineItem),
+          ...data,
+          movementNumber: data.identifier,
```
kchelstowski, line 271: is it `movementNumber` anywhere? :thinking: why can't it be kept as `identifier`?

SebastianLib: We're using movementNumber in useInboundItemsForm. I'm saving it as movementNumber because in the old inbound it's also called movementNumber, so I wanted to keep the naming consistent. Do you think it's okay to stick with this name, or should I change it? 
image

kchelstowski: I can't see `movementNumber` being used on the backend though, so I assume it's just used on the frontend and mapped from `identifier` to `movementNumber`.
Generally I would prefer to keep naming backend<->frontend the same if possible, so if `movementNumber` is just a frontend thing, I would keep it as `identifier`, because this is what we use on the backend.

SebastianLib: Okay, I'll change it

src/js/components/DataTable/DataTable.scss
alannadolny: I meant tables in outbound, old inbound, etc. the `rt-table` classname comes from the react-table library so it's used in the whole app.

SebastianLib: when I tested different tables and everything was ok after this change, but to be 100% sure that nothing would happen I created a new className **.overflow-hidden**

alannadolny (approved):

PR #5179, opened 2025-03-31
OBPIH-7149 Fix default display date format

awalkowiak:
https://pihemr.atlassian.net/browse/OBPIH-7149

### DIFF
--- a/src/main/groovy/org/pih/warehouse/core/Constants.groovy
+++ b/src/main/groovy/org/pih/warehouse/core/Constants.groovy
@@ -43,9 +43,9 @@ class Constants {
     // Display date using day of year format. This is used by the `openboxes.display.date.format` property.
     // It is using `DD` instead `dd` because it is later used on the frontend by the moment, which interprets
     // `dd` and `DD` differently(`DD` is day of month, and `dd` is day of week, see: OBPIH-7149 and OBPIH-5397).
-    static final String DISPLAY_DATE_FORMAT = "MMM DD, yyyy"
+    static final String JS_DISPLAY_DATE_FORMAT = "MMM DD, yyyy"
     // Display date using "DOM" - day of month format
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -40,7 +40,12 @@ class Constants {
     static final String MONTH_DAY_YEAR_DATE_FORMAT = "MM/dd/yyyy"
     static final String DELIVERY_DATE_FORMAT = "MM/dd/yyyy HH:mm XXX"
     static final String EUROPEAN_DATE_FORMAT = "dd/MM/yyyy"
-    static final String DISPLAY_DATE_FORMAT = "MMM dd, yyyy"
+    // Display date using day of year format. This is used by the `openboxes.display.date.format` property.
+    // It is using `DD` instead `dd` because it is later used on the frontend by the moment, which interprets
+    // `dd` and `DD` differently(`DD` is day of month, and `dd` is day of week, see: OBPIH-7149 and OBPIH-5397).
+    static final String DISPLAY_DATE_FORMAT = "MMM DD, yyyy"
```
EWaterman, line 46: I'm going to approve the PR, but I'd personally rename this field to something like `DISPLAY_DATE_DOY_FORMAT` or `JS_DISPLAY_DATE_FORMAT` so that we don't accidentally use it anywhere on the backend gsps or exports. Having it be just `DISPLAY_DATE_FORMAT` implies this is what we should use by default whenever we want to display the date, which isn't actually true for the backend.

EWaterman (approved):

alannadolny (approved):

src/main/groovy/org/pih/warehouse/core/Constants.groovy
awalkowiak: Actually you are right, I'll do it now

PR #5182, opened 2025-04-01
OBPIH-7128 Fix bug with elements not being removed from a SortedSet

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
@@ -96,6 +96,9 @@ class CycleCountItem implements Comparable {
     int compareTo(Object that) {
         int diff = inventoryItem?.expirationDate <=> that.inventoryItem?.expirationDate
                 ?: location?.name <=> that.location?.name
+                    // This id comparison is required due to a quirk in SortedSet (CycleCount has a SortedSet of
+                    // CycleCountItem) where it uses compareTo as an equals method when adding/removing from the set,
+                    // and so we need to make sure this always resolves to a unique value for each item. (OBPIH-7128)
                     ?: id <=> that.id
         return diff
     }
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
@@ -96,9 +96,7 @@ class CycleCountItem implements Comparable {
     int compareTo(Object that) {
         int diff = inventoryItem?.expirationDate <=> that.inventoryItem?.expirationDate
                 ?: location?.name <=> that.location?.name
-        if (diff == 0) {
-            return -1
-        }
+                    ?: id <=> that.id
         return diff
     }
 }
--- a/src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountProductAvailabilityServiceSpec.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountProductAvailabilityServiceSpec.groovy
@@ -232,11 +232,7 @@ class CycleCountProductAvailabilityServiceSpec extends Specification implements 
 
         then: 'items should have changed'
         assert changedItems.itemsHaveChanged()
-
-        // TODO: We enter the correct flow but for some reason removeFromCycleCountItems doesn't actually remove
-        //       the item. This is almost certainly due to unit test weirdness. Possibly because the item hasn't been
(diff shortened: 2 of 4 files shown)


grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -96,9 +96,7 @@ class CycleCountItem implements Comparable {
     int compareTo(Object that) {
         int diff = inventoryItem?.expirationDate <=> that.inventoryItem?.expirationDate
                 ?: location?.name <=> that.location?.name
-        if (diff == 0) {
-            return -1
-        }
+                    ?: id <=> that.id
```
kchelstowski, line 99: This is actually a good one. Let me explain:
The problem with "deleted object would be re-saved by cascade" was due to the fact, that this line:
https://github.com/openboxes/openboxes/blob/605805ae63d5482a67588570ce198ff8cdb22b03/grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy#L180

was not removing an item from the `cycleCountItems` - it was eventually setting the `cycleCountItem.cycleCount` to `null`, but it didn't remove the element from the Set. I thought it's Grails' bug that it has some problems with removing elements from `SortedSet`, tried removing it manually from the Set via `.remove`, but it also didn't work.
It turned out that `SortedSet` uses the `compareTo` even if you remove an element and even if the set contains only one element - it then compares it with itself.
It turned out that if the elements are equal, we mustn't hardcode -1/1 for such case (we did it to keep the order of the items in the table, as they were jumping back and forth). I believe the easiest solution for that is to just do the fallback for id, as it would always be different and the order would stay the same. I tested it, and feels like it is working fine.

alannadolny (approved):

SebastianLib (approved):

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
EWaterman: wait, so it calls compareTo when inserting/removing from the SortedSet not just for sorting the list but also for equalilty (to know which entry to remove)??

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountProductAvailabilityServiceSpec.groovy
```diff
@@ -361,27 +353,28 @@ class CycleCountProductAvailabilityServiceSpec extends Specification implements
...
         Location binLocation = new Location(name: 'binLocation')
+        CycleCountItem countItem = new CycleCountItem(
+                inventoryItem: inventoryItem,
+                location: binLocation,
+                product: product,
+                countIndex: 0,  // count
+                quantityOnHand: 20,
+                custom: false,
+        )
+        countItem.id = '0'
```
EWaterman, line 364: it's really frustrating that I had to do this but id still ends up as null if we put it in the constructor. It's some grails DataTest weirdness...
```
CycleCountItem countItem = new CycleCountItem(id: '1')
assert countItem.id == null
```

EWaterman (approved):

awalkowiak (approved):

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountProductAvailabilityServiceSpec.groovy
jmiranda: AFAIR, in Grails, you cannot set the ID unless you use "assigned" as the generator.

jmiranda: Actually I always thought it was due to the "generator", but it might just be that the id property is not bindable. https://stackoverflow.com/a/43546046/136597. 

I actually would like to be able to bind the ID for testing purposes so maybe we could look into making it bindable for all domains using the following in application.groovy. 
```
grails.gorm.default.mapping = {
    '*'(cascadeValidate: 'none')
    id generator: 'uuid', bindable: true
}
```

EWaterman: oh interesting. Maybe best to do this only in tests at first (surely there's a way to define a test only config file) since we do a lot of param binding to domains:
```
new Domain(params)
```
which comes directly from the API request body so there might be some weird behaviour if the client is passing id in the request body and we suddenly start setting/saving it on the domain.

PR #5190, opened 2025-04-08
OBPIH-7175 Header improvements

SebastianLib:

**Description:**
image

---


src/js/components/stock-movement-wizard/inboundV2/InboundHeader.jsx
```diff
@@ -1,22 +1,26 @@
...
 
 import useTranslate from 'hooks/useTranslate';
 import HeaderWrapper from 'wrappers/HeaderWrapper';
 
-const InboundHeader = ({ title, status }) => {
+const InboundHeader = () => {
+  const { headerInfo, headerStatus } = useSelector((state) => ({
+    headerInfo: state.inbound.headerInfo,
+    headerStatus: state.inbound.headerStatus,
+  }));
```
kchelstowski, line 12: have you made sure this would eventually get cleared for different scenarios, like refreshing the page, changing the page coming to inbound again etc?

SebastianLib: I've checked, and the only issue I see is when someone intentionally enters an incorrect ID in the URL dispatch doesn't clear the data. Previously, I didn't implement a function to catch that and redirect the user to step 1. I probably should have done that a while ago, but I must have overlooked it, and I assume that this should be in other ticket. However, it might be worth adding an option where, if the ID in the URL is wrong, the user gets redirected to step 1 in this ticket

src/js/hooks/inboundV2/addItems/useInboundAddItemsForm.jsx
```diff
@@ -574,19 +579,36 @@ const useInboundAddItemsForm = ({
...
+      dispatch(updateInboundHeader([], {}));
+      // In case of an error, redirect to the "create" step without the id parameter
+      history.push({
+        pathname: location.pathname,
+        search: queryString.stringify({
+          ...queryParams,
+          id: undefined,
+          step: InboundV2Step.CREATE,
+        }, { skipNull: true }),
+      });
```
SebastianLib, line 606: It seems to me that I shouldn't add this to the wizard logic, because it could become very complicated to use it later

awalkowiak:
It was merged in other PR

PR #5191, opened 2025-04-08
OBPIH-7170 Fix bug with cycle count missing cycle count items with th…

kchelstowski:
…e same bin

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
@@ -100,6 +100,10 @@ class CycleCountItem implements Comparable {
                     // CycleCountItem) where it uses compareTo as an equals method when adding/removing from the set,
                     // and so we need to make sure this always resolves to a unique value for each item. (OBPIH-7128)
                     ?: id <=> that.id
+                        // The identityHashCode is needed in case of adding a new item that doesn't have the id yet.
+                        // There is a case where you put the same expirationDate and location, and due to the SortedSet behavior
+                        // this would assume the items are the same and it wouldn't add more than one with the same expirationDate and location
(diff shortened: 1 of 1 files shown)


grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -100,6 +100,10 @@ class CycleCountItem implements Comparable {
                     // CycleCountItem) where it uses compareTo as an equals method when adding/removing from the set,
                     // and so we need to make sure this always resolves to a unique value for each item. (OBPIH-7128)
                     ?: id <=> that.id
+                        // The identityHashCode is needed in case of adding a new item that doesn't have the id yet.
+                        // There is a case where you put the same expirationDate and location, and due to the SortedSet behavior
+                        // this would assume the items are the same and it wouldn't add more than one with the same expirationDate and location
+                        ?: System.identityHashCode(this) <=> System.identityHashCode(that)
```
kchelstowski, line 106: This is another bug with the SortedSet and comparison thing. The case was that when creating a cycle count with items in the same bin but e.g. different lot, those would be treated as the same, because we do not have the id at this point, so the comparison `id <=> that.id` would compare `null` to `null`, so we need one more comparison to be able to compare not persisted objects.
I added the `identityHashCode`, as it returns different hashes for "the same" items.

awalkowiak (commented):
I wonder if we should use lot number in here instead, but I remember we wanted to keep it as close to the stock card ordering. So perhaps it would be worth to add sort by lot in both places (not necessarily right now)? :thinking: 
So for now I think this is ok to me, **unless** this might cause some duplicated items on save with same exp, lot and bin.

awalkowiak (approved):
As we don't have a batch item create, I think we can merge as is now, and revisit while working on batch

PR #5195, opened 2025-04-09
OBPIH-7173 Rows duplicated when add the same inventory with no lot as added on recount 

alannadolny:


SebastianLib (approved):

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -254,7 +254,7 @@ const useResolveStep = () => {
         id: productId,
       },
       inventoryItem: {
-        lotNumber: null,
+        lotNumber: '',
```
awalkowiak, line 257: Why empty string? Does it create an inventory item with an empty string for the default lot which should be `null`? Is it only required here? Does it work correctly on the `useCountStep` with `null`?

kchelstowski (approved):

jmiranda:
@EWaterman @awalkowiak please don't merge this until i have a chance to review

src/js/hooks/cycleCount/useResolveStep.js
EWaterman: I did some very quick digging. Reload returns a list of CycleCountItemDto in its response body, which is constructed via this method:
```
CycleCountItemDto toDto() {
    return new CycleCountItemDto(
            ...
            inventoryItem: inventoryItem,
            ...
    )
}
```

I ran a quick test by adding a new row with an empty lot during a recount then clicked reload and the backend returned `lot: null` in the response payload of the reload call so we should be careful here as per Justin's comment. We should make sure we're being consistent about the behaviour.

I'd say if you're changing it here, you should also change it for count. It feels weird for it to not be consistent across count and recount and I fear if we do something like add an empty item during a count, then add another one during a recount, it'd cause issues.

I'm assuming the backend returns `lot: ''` when an item already exists for the product with an empty lot? Maybe you can trigger that scenario. Start a recount on a proudct with an item that has a blank lot then add a new row, with a blank lot (and a different bin) to see if they're treated the same.

Also, can you double check that in the database it stores the lot the same if '' is provided vs null? If not, we'll need to investigate further to make sure we're being thoroughly consistent

alannadolny: @EWaterman @jmiranda I added a new entry on record stock (empty bin, empty lot, empty expiration date)
image
As you can see in the DBeaver, I have `[NULL]` in expiration date, comment, and lot_status. The lot_number is blank, without the information about the null, so it looks like it's currently stored as an empty string.
Now, after my change in the DBeaver, creating a new entry with an empty bin, lot, and exp date looks the same.
image

awalkowiak (approved):
@jmiranda @EWaterman I confirmed the behavior with @alannadolny now we always save and get the empty string (while saving or reading the lot), and the `null` that @EWaterman mentioned probably was caused by the previous bug. However if we'd ever change `convertEmptyStringsToNull` setting to `true` we have to remember about this one.

awalkowiak:
Let's hold with merging because @kkrawczyk123 is doing some additional investigation

PR #5198, opened 2025-04-10
OBPIH-7169 Detect discrepancy caused by removed inventories during co…

kchelstowski:
…unt workflow

**Description:**

---
Screenshot from 2025-04-10 15-33-12

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
@@ -124,7 +124,7 @@ class CycleCountProductAvailabilityService {
                     it.inventoryItem == cycleCountItem.inventoryItem &&
                     it.custom
         }
-        if (!wasItemEverCustomAdded) {
+        if (!wasItemEverCustomAdded && cycleCountItem.quantityCounted <= 0) {
             refreshState.addItemToDelete(cycleCountItem)
             return
         }
--- a/src/js/components/cycleCount/CycleCountDiscrepancyInfoBar.jsx
+++ b/src/js/components/cycleCount/CycleCountDiscrepancyInfoBar.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+const CycleCountDiscrepancyInfoBar = ({ outOfStockItems }) => (
+  <div className="d-flex justify-content-between flex-column draft-modal">
+    {outOfStockItems.map((item) => (
+      <div key={item.id}>
+        <Translate
+          id="react.cycleCount.modal.outOfStockDiscrepancyFound.label"
+          defaultMessage=""
+          data={{
+            binLocation: item.binLocation?.name,
+            lotNumber: item.inventoryItem?.lotNumber,
+            quantityOnHand: item.quantityOnHand,
+            quantityCounted: item.quantityCounted,
+          }}
+        />
+      </div>
+    ))}
+  </div>
+);
+
+export default CycleCountDiscrepancyInfoBar;
+
+CycleCountDiscrepancyInfoBar.propTypes = {
+  outOfStockItems: PropTypes.arrayOf(PropTypes.shape({}).isRequired).isRequired,
+};
--- a/src/js/hooks/cycleCount/useResolveStep.js
+++ b/src/js/hooks/cycleCount/useResolveStep.js
@@ -32,6 +32,7 @@ const useResolveStep = () => {
   // Table data is stored using useRef to avoid re-renders onBlur
   // (it removes focus while selecting new fields)
   const tableData = useRef([]);
+  const cycleCountsWithItemsWithoutRecount = useRef([]);
   // Saving selected "recounted by" option
   const [recountedBy, setRecountedBy] = useState({});
   // Saving selected "date recounted" option, initially it's the date fetched from API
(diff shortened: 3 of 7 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -124,7 +124,7 @@ class CycleCountProductAvailabilityService {
                     it.inventoryItem == cycleCountItem.inventoryItem &&
                     it.custom
         }
-        if (!wasItemEverCustomAdded) {
+        if (!wasItemEverCustomAdded && cycleCountItem.quantityCounted <= 0) {
```
EWaterman, line 127: Why <= 0 and not == 0? We can't count a negative number right? I suppose it's fine this way though.

kchelstowski: yeah, just to be extra sure I made it `<= 0`. Feels like we don't have any `min/max` validation neither in command nor in the domain on the `quantityCounted` field so potentially a user might send a negative value through the API.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -146,6 +147,15 @@ const useResolveStep = () => {
     }).filter(Boolean);
   };
 
+  const getItemsWithoutRecountIndexes = (items) => {
+    const duplicatedItems = _.groupBy(items,
+      (item) => `${item.binLocation?.id}-${item?.inventoryItem?.lotNumber}`);
+
+    const duplicatedItemsValues = Object.values(duplicatedItems);
+
+    return duplicatedItemsValues.filter((group) => group.length < 2).flat();
```
EWaterman, line 156: so this is checking the number of items that exist for a bin and lot? What if you create a new custom item during the recount? Then if that bin + lot are new for the count, this will return those rows as well, which given the method name I assume you don't want.

Why not check that "group" doesn't have any elements with countIndex > 0/

Or maybe simpler: filter where group.length == 1 and group[0].countIndex == 0

src/js/components/cycleCount/CycleCountDiscrepancyInfoBar.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+import Translate from 'utils/Translate';
+
+const CycleCountDiscrepancyInfoBar = ({ outOfStockItems }) => (
```
EWaterman, line 7: isn't this popup meant to show ANY stock changes (not just ones where QoH==0) that happen when re-fetching/reloading the data? So not just outOfStockItems but itemsWithStockChange (or something).

Also, I think "CycleCountDiscrepancyInfoBar" could be a confusing name because we already use "discrepancy" to mean when QoH != quantity counted, but that's not what we're talking about here. Here it's when QoH has changed after a reload.

CycleCountStockChangeInfoBar or something like that is clearer to me

src/js/hooks/cycleCount/useResolveStep.js
kchelstowski: I think that if you create a custom row like that, you don't want to display the discrepancy anymore, because you intentionally add such an item. It's actually copied from Alan's PR where he was supposed to hide such a row, since we actually need the same logic to handle two different cases, I copied it from his PR: https://github.com/openboxes/openboxes/pull/5193

src/js/components/cycleCount/CycleCountDiscrepancyInfoBar.jsx
kchelstowski: I don't think so, the ticket description says clearly that we want the popup/info bar only if QOH is equal to 0 (removed inventory item during count).
It's not really related to cycle count reload, it's the case when you want to submit a count (go to resolve page) and in the meantime the stock changed to 0.

src/js/hooks/cycleCount/useResolveStep.js
alannadolny: If you are OK with checking the count indexes of that group, it should be done like that. We can also check if the group doesn't contain a customs row. But it is something that needs to be adjusted in my PR. At this point, you can skip that code.

src/js/components/cycleCount/CycleCountDiscrepancyInfoBar.jsx
kchelstowski: Moreover I think that this popup is supposed to show for QOH==0, because we are supposed to hide such a row in the table, so this popup is not to have users confused that an item suddenly disappeared from the table.

EWaterman: > the ticket description says clearly that we want the popup/info bar only if QOH is equal to 0

I read it and yeah you're right. This is about rows disappearing and us wanting to notify the user about it. I still think the naming is confusing though and CycleCountStockChangeInfoBar or even CycleCountStockDiscrepancyInfoBar would be clearer IMO.

> It's not really related to cycle count reload

I'm still not super familiar with how the frontend but won't it also display if you reload and a product is removed?

src/js/hooks/cycleCount/useResolveStep.js
EWaterman: > I think that if you create a custom row like that, you don't want to display the discrepancy anymore, because you intentionally add such an item.

Right, so we shouldn't return that row here. Essentially we're checking "is there a non-custom item that was in the count but is now not in the recount?" and you'd do that by checking if there's an element in "group" with index 0 but not one with index > 0 (or if you want to do like Alan says, which I agree is more correct, you could try to determine what the current index is, and filter down to rows where there's an item with currentIndex -1 but not one with currentIndex).

EWaterman: Also, how are you preventing this from displaying multiple times? If you move from count to recount, then see the popup, then add some counts, save draft, leave the page and return, does it show the popup again? It probably shoudn't.

It looks like it's building the cycleCountsWithItemsWithoutRecount list of items on refetchData, but should that only happen when parsing the result of reload or start recount?

kchelstowski: It was not determined whether it should display multiple times or not. If it would not be meant to display multiple times, the complexity changes a lot - we've had a discussion with Alan and initially I wanted to handle that getting a response from submitting the count (adding some additional property to return out of stock items), but it turned out almost impossible or very difficult to handle all the cases to pass such a response to the resolve form (those are separate URLs, components), and using Redux for that would also be problematic due to the fact, that you might just go out and then come back to such a form from "to resolve tab" - it was also not described whether we should show a user such discrepancy indicator or not.
Since this is needed for the release I would not overthink this too much, as those are things that were not described, so in my opinion we could improve this after first round of testing

awalkowiak (approved):

PR #5199, opened 2025-04-10
OBPIH-6780 Add Page numbers on Picklist and Delivery Note

SebastianLib:

**Description:**

---

### DIFF
--- a/grails-app/views/deliveryNote/print.gsp
+++ b/grails-app/views/deliveryNote/print.gsp
@@ -34,11 +34,6 @@
             text-align: center;
             position: running(header);
         }
-        div.footer {
-            display: block;
-            text-align: center;
-            position: running(footer);
-        }
 
         @page {
             size: letter;
--- a/grails-app/views/deliveryNote/print.gsp
+++ b/grails-app/views/deliveryNote/print.gsp
@@ -47,7 +47,6 @@
             @bottom-center {
                 content: "Page " counter(page) " of " counter(pages);
                 font-size: 12px;
-                color: #333;
                 font-family: sans-serif;
             }
         }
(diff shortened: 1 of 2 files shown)


grails-app/views/deliveryNote/print.gsp
```diff
@@ -44,7 +44,11 @@
             size: letter;
             background: white;
             @top-center { content: element(header) }
-            @bottom-center { content: element(footer) }
```
SebastianLib, line 47: Actually, this element doesn’t even exist in the file, so I could remove the style for div.footer. But since a lot of the other unused styles haven’t been removed either, I’m not going to remove this one for now. Or maybe should I?

awalkowiak: Remind me, why you could not re-add the:
```
<div class="footer">
    <div id="page-footer" class="small">
        <div>
            Page <span class="page" /> of <span class="pagecount" />
        </div>
    </div>
</div>
```

We have all the styling there, so perhaps this could work as well?

SebastianLib: if we use:         
```
@page {
            size: letter;
            background: white;
            @top-center { content: element(header) }
            @bottom-center { content: element(footer) }
        }
```

```
<div class="footer">
    <div id="page-footer" class="small">
        <div>
            Page <span class="page" /> of <span class="pagecount" />
        </div>
    </div>
</div>
``` 
the page numbers is not working properly and we see something like that:
image
image

if we leave it as it is, only when we click print we see the page numbers: 
image
image

awalkowiak: Ok then i think you can get rid of the unused styling

awalkowiak (approved):

PR #5200, opened 2025-04-10
OBPIH-7173 Rows duplicated when add the same inventory with no lot as added on recount (fixes after QA)

alannadolny:
I decided to change the default row values (lot number) to null instead of an empty string, because it looks more appropriate in that case. The main issue here is that we can have *TWO* different values for the default lot number. Sometimes I got null, and the other time I got an empty string. The empty strings come from record stock, but I noticed that I was able to create an inventory item with null (I don't know how; it happened just once). So to unify the behavior, I added null as a fallback to always get the same default lot.
The issue with grouping rows that caused the duplicated rows behaviour again existed because, during finding an inventory item, we searched for null, but the inventory item had an empty string (it was also achievable to reproduce this the other way.)

### DIFF
--- a/grails-app/init/org/pih/warehouse/BootStrap.groovy
+++ b/grails-app/init/org/pih/warehouse/BootStrap.groovy
@@ -212,7 +212,7 @@ class BootStrap {
                     name       : inventoryItem?.product?.name,
                     productCode: inventoryItem?.product?.productCode
                 ],
-                lotNumber     : inventoryItem.lotNumber,
+                lotNumber     : inventoryItem.lotNumber ?: null,
                 expirationDate: inventoryItem.expirationDate?.format("MM/dd/yyyy")
             ]
         }
--- a/src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
+++ b/src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
@@ -15,15 +15,16 @@ class CycleCountItemCommand implements Validateable {
 
     @BindUsing({ obj, source ->
         Product product = Product.read(source['inventoryItem']['product'])
-        InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, source['inventoryItem']['lotNumber'])
+        String lotNumber = source['inventoryItem']['lotNumber']
+        InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber)
 
         // Currently default lotNumbers can be null or empty string,
         // to avoid the situation of not finding appropriate one
         // we are trying to find another inventoryItem with
         // empty lot
-        if (!inventoryItem && !source['inventoryItem']['lotNumber']) {
-            String lotNumber = source['inventoryItem']['lotNumber'] == null ? '' : null
-            inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber)
+        if (!inventoryItem && !lotNumber) {
+            String otherBlankLotNumber = lotNumber == null ? '' : null
+            inventoryItem = InventoryItem.findByProductAndLotNumber(product, otherBlankLotNumber)
         }
 
         return inventoryItem ?: new InventoryItem(
--- a/src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
+++ b/src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
@@ -16,6 +16,16 @@ class CycleCountItemCommand implements Validateable {
     @BindUsing({ obj, source ->
         Product product = Product.read(source['inventoryItem']['product'])
         InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, source['inventoryItem']['lotNumber'])
+
+        // Currently default lotNumbers can be null or empty string,
+        // to avoid the situation of not finding appropriate one
+        // we are trying to find another inventoryItem with
+        // empty lot
+        if (!inventoryItem && !source['inventoryItem']['lotNumber']) {
+            String lotNumber = source['inventoryItem']['lotNumber'] == null ? '' : null
+            inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber)
+        }
+
         return inventoryItem ?: new InventoryItem(
                 product: product,
                 lotNumber: source['inventoryItem']['lotNumber'],
--- a/grails-app/domain/org/pih/warehouse/inventory/InventoryItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/InventoryItem.groovy
@@ -82,7 +82,7 @@ class InventoryItem implements Serializable {
                 "inventoryItemId": id,
                 "productId"      : product?.id,
                 "productName"    : product?.name,
-                "lotNumber"      : lotNumber,
+                "lotNumber"      : lotNumber ?: null,
                 "expirationDate" : expirationDate?.format("MM/dd/yyyy"),
                 "quantityOnHand" : quantity ?: 0,
                 "quantityATP"    : quantity ?: 0,       //todo: quantity available to promise will coming soon
--- a/src/js/hooks/cycleCount/useCountStep.jsx
+++ b/src/js/hooks/cycleCount/useCountStep.jsx
@@ -234,7 +234,7 @@ const useCountStep = () => {
         id: productId,
       },
       inventoryItem: {
-        lotNumber: '',
+        lotNumber: null,
         expirationDate: null,
       },
       binLocation: null,
--- a/src/js/hooks/cycleCount/useResolveStep.js
+++ b/src/js/hooks/cycleCount/useResolveStep.js
@@ -254,7 +254,7 @@ const useResolveStep = () => {
         id: productId,
       },
       inventoryItem: {
-        lotNumber: '',
+        lotNumber: null,
         expirationDate: null,
       },
       binLocation: null,


grails-app/init/org/pih/warehouse/BootStrap.groovy
```diff
@@ -212,7 +212,7 @@ class BootStrap {
                     name       : inventoryItem?.product?.name,
                     productCode: inventoryItem?.product?.productCode
                 ],
-                lotNumber     : inventoryItem.lotNumber,
+                lotNumber     : inventoryItem.lotNumber ?: null,
```
EWaterman, line 215: I know this isn't new from you, but I wonder why there is both InventoryItem.toJson() and this? Surely this should be:

```
JSON.registerObjectMarshaller(InventoryItem) { InventoryItem inventoryItem ->
    return inventoryItem.toJson()
}
```

EWaterman: but more importantly, this is going to have implications anywhere inventory items are used. I wouldn't be surprised if this breaks the frontend elsewhere because it was expecting '' but is now getting null. This will need testing to make sure old flows still work.

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
```diff
@@ -16,6 +16,16 @@ class CycleCountItemCommand implements Validateable {
...
         InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, source['inventoryItem']['lotNumber'])
+
+        // Currently default lotNumbers can be null or empty string,
+        // to avoid the situation of not finding appropriate one
+        // we are trying to find another inventoryItem with
+        // empty lot
+        if (!inventoryItem && !source['inventoryItem']['lotNumber']) {
+            String lotNumber = source['inventoryItem']['lotNumber'] == null ? '' : null
+            inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber)
+        }
```
EWaterman, line 27: minor suggestion to make it a bit more readable IMO:
```
String lotNumber = source['inventoryItem']['lotNumber']
InventoryItem inventoryItem = InventoryItem.findByProductAndLotNumber(product, lotNumber)

// Your comment
if (!inventoryItem && StringUtils.isEmpty(lotNumber)) {
    String otherBlankLotNumber = lotNumber == null ? '' : null
    inventoryItem = InventoryItem.findByProductAndLotNumber(product, otherBlankLotNumber)
}
```

jmiranda:
TL;DR 

In the short-term, I think the following is probably ok. 
* Continue to allow lot number to be NULL or empty string 
* Coaleasce towards making all default inventory_item records consistent i.e lot number = NULL 
* Identify any products that have multiple inventory items with lot number = NULL or empty string. 

Long-term, here's where I think we need to end up. 
* Consider choosing a default lot number to use to allow us to avoid 
* Change all 
* We should probably make lotNumber `nullable:false`.

---

My commentary on the PR and discussion above.

> The empty strings come from record stock, but I noticed that I was able to create an inventory item with null (I don't know how; it happened just once).

In response to the  "I don't know how" comment, if we look at the InventoryItem constraints ...
```
    // Notice the unique constraint on lotNumber/product
    static constraints = {
        product(nullable: false)
        ...
    }
```
... the lotNumber is nullable:true and has a unique constraint (Grails) so in theory you should only be able to add one record where lotNumber = null (as you observed). That should explain the how. The unique constraint (Grails) should prevent more records where lotNUmber = null, but it allows you to create an additional record where lotNumber = "" since we don't have a `blank:false` constraint.

NOTE: We've discussed this a bunch in the past, but worth noting again that if this unique constraint only existed on the MySQL side, you would be able to insert multiple records with lotNumber = null because MySQL chooses to ignore NULL values in this case (https://dba-presents.com/databases/mysql/274-mysql-unique-index-with-null-values). 

Additional Resource:
* https://bugs.mysql.com/bug.php?id=17825
* https://bugs.mysql.com/bug.php?id=8173
* https://forums.mysql.com/read.php?22,53591,53975#msg-53975

For the reasons discussed in those articles, I think we should eventually make lotNumber nullable: false, but that puts us into a situation where we need to find a suitable default constant for the default lot number. We've done this on the reporting side with the use of a DEFAULT for lot number and bin location because it was necessary for the product availability unique constraint. Doing that with our transaction data is a little different, as explained below.

> So to unify the behavior, I added null as a fallback to always get the same default lot.

I think this is the best approach but no matter which one we choose (null or empty string or allow either) we need to make sure we're always checking for both and not creating a record with null and another with empty string for the same product. There's code all over the place that will lookup the default inventory item by `isNull("lotNumber" OR eq("lotNumber", "")` and we need to make sure this is always how we lookup default inventory items. 

See InventoryService.findInventoryItemByProductAndLotNumber() for example.
```
        def inventoryItems = InventoryItem.createCriteria().list() {
            and {
                eq("product.id", product?.id)
                if (lotNumber) {
                    eq("lotNumber", lotNumber)
                } else {
                    // Lookup by default lot number (either null or empty string)  
                    or {
                        isNull("lotNumber")
                        eq("lotNumber", "")
                    }
                }
            }
        }
        // Ensure we limit the results to 1
        if (inventoryItems) {
            return inventoryItems.get(0)
        }
        return null
```
NOTE: This does not take into account strings with whitespace, so that should be something we look into as well. 
---

We encountered a similar issue in a PR while we were working on the outbound import and I recommended making changes to ensure that we always perform lookups for the default InventoryItem in the same way. I think I tried to 

If that doesn't exist, we're getting to a place where we might want to abandon the "codify default lot number through logic" and choose an explicit constant for the default lot number. 
```
    static final String DEFAULT_LOT_NUMBER = "DEFAULT" // default product availability lot number
    static final String DEFAULT_BIN_LOCATION_NAME = "DEFAULT" // default product availability bin location name
```
NOTE: Default bin location has a similar problem but I'm not sure we want to go in the same direction for that at the moment since the bin location is a foreign key (and can be null) so we might not have the same problem there i.e. we don't need to do a lookup on location.name like we do with inventory_item.lot_number.
```
MariaDB [openboxes]> select count(*) from location where name = '' or name is null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.000 sec)
```
whereas the 
```
MariaDB [openboxes]> select count(*), count(lot_number = ''), count(lot_number is null) from inventory_item where lot_number = '' or lot_number is null;
+----------+------------------------+---------------------------+
| count(*) | count(lot_number = '') | count(lot_number is null) |
+----------+------------------------+---------------------------+
|    16757 |                   6346 |                     16757 |
+----------+------------------------+---------------------------+
1 row in set (0.016 sec)
```
NOTE: I am actually floored that count(lot_number is null) returned 16757. If that's true and I didn't mess up somewhere, than that means lot_number is null includes lot_number = "" in its set, which means we don't need to use the OR logic discussed above when querying for the default lot number.


The reason we haven't done this yet is that any constant string that we decide to use as the "default" could actually be a valid lot number, so we'd find ourselves in a situation where we'd have to disambiguate "default" from our default value. 

```
MariaDB [openboxes]> select id, product_id, lot_number, expiration_date from inventory_item where lower(lot_number) like '%default%';
+----------------------------------+----------------------------------+------------+---------------------+
| id                               | product_id                       | lot_number | expiration_date     |
+----------------------------------+----------------------------------+------------+---------------------+
| 09f5e11678d5cf2f0179143d287a6222 | 21099                            | Default    | NULL                |
| 09f5e1168222682601823aadd5030ac8 | 10033                            | Default    | NULL                |
| 09f5e1168290e3c10182c58b84b73104 | 17640                            | DEFAULT    | 2022-09-30 00:00:00 |
| 09f5e116860b6c20018620d5188d5b5e | 09f5e1167c3c9801017c4d406ec42732 | Default    | NULL                |
+----------------------------------+----------------------------------+------------+---------------------+
4 rows in set (0.017 sec)
```

With that said, I think it's probably safe (or at least safer) to use a constant ("DEFAULT", "default", or "somestringthatwillneverexisteverinthehistoryoflotnumbers") and live with the consequences, than to continue to try to treat NULL and/or empty string as the indicator of a default value.

jmiranda:
To answer my own question, are there any products with more than 1 inventory item where lot number is NULL or empty?

Yes. And that's a bummer. 
```
MariaDB [openboxes]> select product_id, count(*) from inventory_item where lot_number = '' or lot_number is null group by product_id having count(*) > 1;
+----------------------------------+----------+
| product_id                       | count(*) |
+----------------------------------+----------+
| 09f5e116828453b601828e3dd9e45b11 |        2 |
| 09f5e116860b6c2001864c8ae5ab18e0 |        2 |
| 10023                            |        2 |
| 10042                            |        2 |
| 10281                            |        2 |
| 10319                            |        2 |
| 10477                            |        2 |
| 10667                            |        2 |
| 10886                            |        2 |
| 10923                            |        2 |
| 11045                            |        2 |
| 11049                            |        2 |
| 11092                            |        2 |
| 11137                            |        2 |
| 11329                            |        2 |
| 11382                            |        2 |
| 11725                            |        2 |
| 11737                            |        2 |
| 12071                            |        2 |
| 12689                            |        2 |
| 14183                            |        2 |
| 14453                            |        2 |
| 16043                            |        2 |
| 21034                            |        2 |
| cab2b1078a50e9fc018abe350ec93163 |        2 |
| ff808081732986c20173389c0e6d4055 |        3 |
+----------------------------------+----------+
26 rows in set (0.026 sec)
```

grails-app/init/org/pih/warehouse/BootStrap.groovy
jmiranda: Interesting point. This might actually have an implicit coersion from "" to null if the "" evaluates to false within the elvis operator. 
```
inventoryItem.lotNumber ?: null
```

> I know this isn't new from you, but I wonder why there is both InventoryItem.toJson() and this? Surely this should be:

We probably missed registering some object marshallers. I would be happy to see this in 0.9.5 but adding it to the hotfix would be dangerous since it would change the JSON response anywhere we're returning an inventory item object.
```
JSON.registerObjectMarshaller(InventoryItem) { InventoryItem inventoryItem ->
    return inventoryItem.toJson()
}
```

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
jmiranda: We should not ever use findByProductAndLotNumber(). In fact, I think we might want to explicit define that method in InventoryItem and throw an UnsupportedOperationException, if that's possible. 

I'll try to find the correct method, but I assume this is either a getter/finder on Product or InventoryItem or a method on InventoryService.

jmiranda: Ok, so I mentioned that we were trying to standardize the logic around finding an inventory item by lot number and returning the default inventory item when we were working on the outbound export. I found the code that was written to handle this. 

As much as possible, let's try to use this method to find an inventory item by lot number related to a product since it standardizing the logic. 
https://github.com/openboxes/openboxes/blob/3327c0eb8b76c54b4c5f0a1b8628105e7fbad047/grails-app/domain/org/pih/warehouse/product/Product.groovy#L472

This might break if we have a product with multiple inventory item records with lotNumber = NULL and another with lotNumber = "". But this is a data cleaning problem, not a problem with the finder. 

This method handles that problem by allowing the query to return a list of objects, but the method itself returns the first inventory item.
https://github.com/openboxes/openboxes/blob/3327c0eb8b76c54b4c5f0a1b8628105e7fbad047/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy#L1587

Which is admittedly dangerous, but it's consistent.

awalkowiak (approved):
@alannadolny If there are any points from the discussion in this PR, that should be ticketed, please do so.

PR #5201, opened 2025-04-11
OBPIH-7202 Fix sending a refresh request for cycle count that was COM…

kchelstowski:
…PLETED

**Description:**

---


src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -491,11 +493,16 @@ const useResolveStep = () => {
           currentLocation?.id,
           cycleCount?.id);
         } catch {
+          cycleCountIdsForOutdatedProducts.push(cycleCount.id);
```
kchelstowski, line 496: this is the most important change - the problem in this ticket was that: assuming we proceed with a cycle count for two products (so we actually proceed with two cycle counts) and try to submit the recount, so we send two recounts requests - first of them was successful (expected), second one was not (also expected, as qoh changed), we were then sending refresh request also for the CC that was already completed.
In order to fix this, I store cycle count ids for the cycle counts that the recount fails and send the refresh only for those, and hide the CCs that were completed.

awalkowiak (approved):

PR #5202, opened 2025-04-14
OBPIH-6962 Allow Export of only Results on Product Source List Page

SebastianLib:

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
@@ -60,8 +60,8 @@ class ProductSupplierService {
         Set<String> usedAliases = new HashSet<>()
         return ProductSupplier.createCriteria().list(command.paginationParams) {
             if (forExport) {
-                createAlias("p.synonyms", "ps", JoinType.LEFT_OUTER_JOIN)
                 createAlias("product", "p", JoinType.INNER_JOIN)
+                createAlias("p.synonyms", "ps", JoinType.LEFT_OUTER_JOIN)
                 createAlias("supplier", "s", JoinType.LEFT_OUTER_JOIN)
                 createAlias("manufacturer", "m", JoinType.LEFT_OUTER_JOIN)
                 createAlias("contractPrice", "cp", JoinType.LEFT_OUTER_JOIN)


grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -283,47 +283,88 @@ class ProductSupplierController {
...
+                    supplier: params.supplier ?: null,
+                    defaultPreferenceTypes: params.defaultPreferenceTypes ? params.list('defaultPreferenceTypes') : null,
+                    createdFrom: params.createdFrom ? new SimpleDateFormat("MM/dd/yyyy").parse(params.createdFrom) : null,
+                    createdTo: params.createdTo ? new SimpleDateFormat("MM/dd/yyyy").parse(params.createdTo) : null,
+                    sort: "dateCreated",
+                    order: "desc",
+                    unlimitedMax: true,
+            )
+
+            productSuppliers = productSupplierService.getProductSuppliers(command)
```
SebastianLib, line 300: I'm not sure if I did the right thing here by using productSupplierService. I used it because it saved me from writing the same code again, but I had to add command.unlimitedMax to be able to return more than 100 records. Maybe I should’ve done it using createCriteria instead?

src/main/groovy/org/pih/warehouse/product/ProductSupplierFilterCommand.groovy
```diff
@@ -23,6 +23,8 @@ class ProductSupplierFilterCommand extends PaginationCommand implements Validate
 
     String order
 
+    Boolean unlimitedMax = false
```
kchelstowski, line 26: maybe you could just override the getMax method from the `PaginationCommand` and that would be cleaner? I don't have hard opinion though, so I'll wait for guys' opinion too.

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -283,47 +283,88 @@ class ProductSupplierController {
...
+                    product: params.product ?: null,
+                    includeInactive: params.includeInactive ?: null,
+                    supplier: params.supplier ?: null,
+                    defaultPreferenceTypes: params.defaultPreferenceTypes ? params.list('defaultPreferenceTypes') : null,
+                    createdFrom: params.createdFrom ? new SimpleDateFormat("MM/dd/yyyy").parse(params.createdFrom) : null,
+                    createdTo: params.createdTo ? new SimpleDateFormat("MM/dd/yyyy").parse(params.createdTo) : null,
+                    sort: "dateCreated",
+                    order: "desc",
+                    unlimitedMax: true,
+            )
```
kchelstowski, line 298: this feels odd - command should be bound automatically (use it as an argument of the method, like we used to do), but again - in my opinion this should be a separate Api endpoint instead of refactoring this old one.

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
kchelstowski: I would say it's definitely good you rely on the method from the `productSupplierService` and this is what we discussed during the planning, but again - in my opinion this should be handled as a separate endpoint in the `ProductSupplierApiController`.

grails-app/controllers/org/pih/warehouse/product/ProductSupplierController.groovy
```diff
@@ -283,47 +283,88 @@ class ProductSupplierController {
 
         def productSuppliers = []
 
-        if (params.hasProperty("productSupplier.id")) {
```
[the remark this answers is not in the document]
alannadolny: I am agree with you

grails-app/controllers/org/pih/warehouse/api/ProductSupplierApiController.groovy
```diff
@@ -12,6 +12,8 @@ import org.springframework.http.HttpStatus
 class ProductSupplierApiController {
 
     ProductSupplierService productSupplierService
+    def documentService
+    def dataService
```
kchelstowski, line 16: let's use types, not def, when it's not needed

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -48,7 +50,7 @@ class ProductSupplierService {
     ProductSupplierIdentifierService productSupplierIdentifierService
     def dataSource
     ProductSupplierDataService productSupplierGormService
-
+    def dataService
```
kchelstowski, line 53: use type

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -446,4 +448,126 @@ class ProductSupplierService {
         }
         return productSupplier
     }
+
+    List<Map> getExportData(ProductSupplierFilterCommand filterParams) {
+
+        def productSuppliers = []
+
+        if(filterParams.exportResults) {
```
kchelstowski, line 456:
```suggestion
        if (filterParams.exportResults) {
```

kchelstowski: additionally - what is that flag responsible for?

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -446,4 +448,126 @@ class ProductSupplierService {
...
+
+            if (useDerivedPackage) {
+                entry["defaultProductPackage.uom.code"] = productPackage?.uom?.code
+                entry["defaultProductPackage.quantity"] = productPackage?.quantity
+                entry["defaultProductPackage.productPrice.price"] = productPackage?.productPrice?.price
+            }
+
+            entry["globalProductSupplierPreference"] = globalPreferencesByProductSupplier[productSupplier?.id]
+            return entry
+        }
```
kchelstowski, line 569: actually those lines give me a headache. I was actually expecting that when we have a change to refactor an existing endpoint, we would refactor it, not copy-paste it.
Can't those joins be handled in the main `createCriteria`?

src/main/groovy/org/pih/warehouse/product/ProductSupplierFilterCommand.groovy
```diff
@@ -23,6 +23,18 @@ class ProductSupplierFilterCommand extends PaginationCommand implements Validate
...
     String order
 
+    String format
+
+    Boolean exportResults
+
+    @Override
+    Integer getMax() {
+        if (exportResults) {
+            return Integer.MAX_VALUE
```
kchelstowski, line 33: I have mixed feelings about that. What would happen if the offset was `null`? wouldn't it return all?

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
SebastianLib: exportResults flag tells us whether to export only the filtered results or all the data

awalkowiak: I would say, this should depend on the filter params. If client wants to export filtered data, then he provides the filter params, and if he does not want to filtered out data, then he does add the filter params to the url.

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
SebastianLib: I simplified the code by removing the if-else statement and using only getProductSuppliers. I also tried moving the joins into the main createCriteria inside getProductSuppliers, but the exported data didn't contain all the correct information. I could try to improve it further and handle everything in one place, but it might take me a bit more time due to small backend experience

grails-app/services/org/pih/warehouse/data/ProductSupplierService.groovy
```diff
@@ -446,4 +448,82 @@ class ProductSupplierService {
         }
         return productSupplier
     }
+
+    List<Map> getExportData(ProductSupplierFilterCommand filterParams) {
```
kchelstowski, line 452: I still don't feel comfortable with that method. As I said in my previous comment, I would like to get rid of unnecessary :spaghetti: and make it more readable if we have a chance.
@SebastianLib please raise any issues you face, so we could figure them out together.

kchelstowski (changes requested):

PR #5204, opened 2025-04-14
OBPIH-7168 Delete all inventories when recount in progress (without status changes)

alannadolny:
I am creating a second pr with the rest of the changes from the https://github.com/openboxes/openboxes/pull/5193 without the changes in statuses

### DIFF
--- a/src/js/hooks/cycleCount/useResolveStep.js
+++ b/src/js/hooks/cycleCount/useResolveStep.js
@@ -10,6 +10,7 @@ import {
 } from 'react';
 
 import _ from 'lodash';
+import moment from 'moment';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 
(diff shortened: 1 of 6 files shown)


src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -98,16 +100,24 @@ const useResolveStep = () => {
     const duplicatedItems = _.groupBy(items,
       (item) => `${item.binLocation?.id}-${item?.inventoryItem?.lotNumber}`);
 
-    const maxCountIndex = _.maxBy(items, 'countIndex').countIndex;
+    const maxCountIndex = _.maxBy(items, 'countIndex')?.countIndex;
 
-    return Object.values(duplicatedItems).flatMap((itemsToMerge) => {
+    const duplicatedItemsValues = Object.values(duplicatedItems);
+
+    const areRecountItemsExist = maxCountIndex > 0;
```
awalkowiak, line 107: Can you change the ordering in here (plus `areRecountItemsExist` seems not to be required, I'd add a comment instead of the const, unless you wanted to use it in the flat map, then rename to `recountItemsExist`)? Currently, it's:
```
duplicated items (grouping)
maxcount (find max)
duplicated items (get values)
maxcount (recount existis?)
duplicated items (flatmap)
```
Looks like you could check the maxCountIndex and then do a grouping, etc:
```javascript
const mergeCycleCountItems = (items) => {
    const maxCountIndex = _.maxBy(items, 'countIndex')?.countIndex;
    
    // Please describe when no recount items are present
    if (maxCountIndex === 0) {
      return [];
    }
    
    anything with duplicated items... 
}
```

awalkowiak (changes requested):

awalkowiak (approved):

PR #5207, opened 2025-04-15
OBPIH-7169 Fix items not being removed when stock is removed during r…

kchelstowski:
…ecount

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
@@ -182,7 +182,8 @@ class CycleCountApiController {
     }
 
     def refreshCycleCount(String cycleCountId) {
-        CycleCountDto cycleCount = cycleCountService.refreshCycleCount(cycleCountId)
+        boolean removeOutOfStockItemsImplicitly = params.boolean("removeOutOfStockItemsImplicitly", false)
+        CycleCountDto cycleCount = cycleCountService.refreshCycleCount(cycleCountId, removeOutOfStockItemsImplicitly)
 
         render([data: cycleCount] as JSON)
     }
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
@@ -38,7 +38,7 @@ class CycleCountProductAvailabilityService {
      *
      * @return CycleCountItemsForRefresh the items that were created, updated, and deleted as a result of the refresh.
      */
-    CycleCountItemsForRefresh refreshProductAvailability(CycleCount cycleCount) {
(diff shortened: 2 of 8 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -124,7 +126,7 @@ class CycleCountProductAvailabilityService {
                     it.inventoryItem == cycleCountItem.inventoryItem &&
                     it.custom
         }
-        if (!wasItemEverCustomAdded && cycleCountItem.quantityCounted <= 0) {
+        if (!wasItemEverCustomAdded && (cycleCountItem.quantityCounted <= 0 || removeOutOfStockItemsImplicitly)) {
```
kchelstowski, line 129: since we have two collapsing behaviors - we want NOT to remove such an item when "zeroing" stock happens while count, but we want to remove the item when it happens while recount (reload, submitting the recount), my suggestion is to just add a flag determining whether we want to remove such an item implicitly even it quantityCounted is <= 0.
Behavior for custom items stays the same.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -161,7 +161,8 @@ const useResolveStep = () => {
 
     const duplicatedItemsValues = Object.values(duplicatedItems);
 
-    return duplicatedItemsValues.filter((group) => group.length < 2).flat();
+    return duplicatedItemsValues
+      .filter((group) => group.length < 2 && group.every((item) => item.countIndex === 0)).flat();
```
kchelstowski, line 165: there was a bug that a custom item added during a recount was visible in the indicator, so unfortunately we have to look for countIndex === 0 here, even though we wanted to avoid relying on them. cc @alannadolny

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
```diff
@@ -182,7 +182,8 @@ class CycleCountApiController {
     }
 
     def refreshCycleCount(String cycleCountId) {
-        CycleCountDto cycleCount = cycleCountService.refreshCycleCount(cycleCountId)
+        boolean removeOutOfStockItemsImplicitly = params.boolean("removeOutOfStockItemsImplicitly", false)
```
awalkowiak, line 185: Nitpicky: What does "implicitly" mean in this context? Do we need it in this param?

src/js/hooks/cycleCount/useResolveStep.js
alannadolny: It can be simplified to the code that is included in my recently merged pr

alannadolny (approved):

grails-app/controllers/org/pih/warehouse/api/CycleCountApiController.groovy
kchelstowski: I wanted to somehow indicate that we „force” remove to happen - currently we might remove an item without this param, if the quantityCounted is <= 0, so I wanted it to sound like „remove an item, no matter how many other conditions we have”

PR #5210, opened 2025-04-15
OBPIH-7177 Inventory has changed not detected when add line on recount and the same line on record stock 

alannadolny:
The issue was caused by the fact that we are calling the save action for every item before submitting a recount. During save, we were looking for existing inventory to assign the already existing QoH for the newly created inventory on recount. So, it led to the inability to detect discrepancies because the QoH was the same on the record stock and in the cycle count item. As a fix, I am setting QoH for the newly created row to 0, so that the discrepancy can be found during submitting a recount, and after refreshing, the existing inventory data will be fetched and correctly displayed. In that case, the information about the discrepancy will be visible for every cycle count that includes newly created inventory.

- in case when we added a new inventory on record stock and the recount stays the same: discrepancy found
- in case when we added a new inventory on recount and not on the record stock: discrepancy not found

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -589,14 +589,13 @@ class CycleCountService {
             }
             command.inventoryItem.save()
         }
-        Integer currentQuantityOnHand = productAvailabilityService.getQuantityOnHandInBinLocation(command.inventoryItem, command.binLocation) ?: 0
         CycleCount cycleCount = command.cycleCount
 
         CycleCountItem cycleCountItem = new CycleCountItem(
                 facility: command.facility,
                 status: command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING,
                 countIndex: command.recount ? 1 : 0,
-                quantityOnHand: currentQuantityOnHand,
+                quantityOnHand: 0,
                 quantityCounted: command.quantityCounted,
                 cycleCount: cycleCount,
                 location: command.binLocation,


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -589,14 +589,13 @@ class CycleCountService {
...
         }
-        Integer currentQuantityOnHand = productAvailabilityService.getQuantityOnHandInBinLocation(command.inventoryItem, command.binLocation) ?: 0
         CycleCount cycleCount = command.cycleCount
 
         CycleCountItem cycleCountItem = new CycleCountItem(
                 facility: command.facility,
                 status: command.recount ? CycleCountItemStatus.INVESTIGATING : CycleCountItemStatus.COUNTING,
                 countIndex: command.recount ? 1 : 0,
-                quantityOnHand: currentQuantityOnHand,
+                quantityOnHand: 0,
```
EWaterman, line 598: Doing this loses some functionality.

What `currentQuantityOnHand` was giving us: if we start a recount, then do a record stock to add an item X for the product with quantity 10, then we go back to the recount and add a custom row for item X, it'll correctly pull the QoH of 10 and the user will see that in the UI as expected. No discrepancy error popup required.

With your change, that custom row will always show 0 QoH for item X until we click reload or try (and fail) to submit the count.

The scenario you're trying to fix is the opposite one: where you add a custom row (which has QoH 0) to the recount and then do the record stock for that same item. In that case, yes you want the warning popup on submit. But I would have assumed that scenario would be covered by this code:
```
CycleCountDto submitCount(CycleCountSubmitCountCommand command) {
    CycleCount cycleCount = command.cycleCount

    if (command.refreshQuantityOnHand) {
        CycleCountProductAvailabilityService.CycleCountItemsForRefresh refreshedItems =
                cycleCountProductAvailabilityService.refreshProductAvailability(cycleCount)

        if (refreshedItems.itemsHaveChanged() && command.failOnOutdatedQuantity) {
            throw new IllegalArgumentException("Quantity on hand for a cycle count item is no longer up to date")
        }
    }
```
I would assume that `refreshedItems.itemsHaveChanged()` would be true here in your scenario but it must not be. Did you debug through the refreshProductAvailability logic to see what's happening? Maybe it's a bug in the refresh logic?

alannadolny: No, this code does not cover this scenario. And it's not a bug in the refreshProductAvailability logic.
It's caused by the fact that the new items are already created using appropriate QoH, pulled from the record stock, so it's not even a bug. But this is something that was required in the ticket. And I know that the pop-up will be visible every time for new items, and I also explained it to Manon.

EWaterman (approved):

PR #5215, opened 2025-04-17
OBPIH-7231 Fix storing internal locations filter when refreshing a page

kchelstowski:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useCycleCountFilters.js
+++ b/src/js/hooks/cycleCount/useCycleCountFilters.js
@@ -65,8 +65,8 @@ const useCycleCountFilters = () => {
       const idList = getParamList(queryPropsParam);
       return elementsList
         .filter(({ id }) => idList.includes(id))
-        .map(({ id, label }) => ({
-          id, label, name: label, value: id,
+        .map(({ id, label, name }) => ({
+          id, label: label ?? name, name: name ?? label, value: id,
         }));
     }
     return null;


src/js/hooks/cycleCount/useCycleCountFilters.js
```diff
@@ -65,8 +65,8 @@ const useCycleCountFilters = () => {
       const idList = getParamList(queryPropsParam);
       return elementsList
         .filter(({ id }) => idList.includes(id))
-        .map(({ id, label }) => ({
-          id, label, name: label, value: id,
+        .map(({ id, label, name }) => ({
+          id, label: label ?? name, name: name ?? label, value: id,
```
kchelstowski, line 69: there was a problem that the `internaLocations` endpoint doesn't return `label` property, whereas every `console.log`, even `console.log` written directly under the fetching method, was eventually returning objects with `label` property, whereas while mapping them, the `label` was `undefined` :thinking: 
It was misleading in the debugging, hence I even asked GPT - it turned out that JS does some quirk with console.log, when we mutate an object later on:
Screenshot from 2025-04-17 14-13-31

This is why I prefer immutable approach in programming.
FYI: `console.table` was returning the real value, not the "If the object is later mutated (like by a component or somewhere else in your code), it reflects that mutation even in earlier logs."

kchelstowski: The real mutating happens in the Select compoment in the `mapOptions` method:
```js
mapOptions(values) {
    return (_.map(values, (value) => {
      const newValue = value;
      }
      if (value.options) {
        newValue.options = this.mapOptions(value.options);
      }
      if (value && this.props.labelKey && !value.label) {
        newValue.label = value[this.props.labelKey];
      }
      if (value && this.props.valueKey && !value.value) {
        newValue.value = value[this.props.valueKey];
      }
      return newValue;
    }));
  }
```
so even if you added here something like:
```js
newValue.test = '123';
```
the `console.log` that would be put directly under fetching method, would already log the objects with `test` property :exploding_head:

EWaterman (approved):

PR #5217, opened 2025-04-18
OBPIH-7158 Bump up unit tests coverage (0.9.4 test cycle)

alannadolny:
I am pushing my current version of unit tests because I don't have more available time to add a few more

### DIFF
--- a/src/test/groovy/unit/org/pih/warehouse/api/CycleCountApiControllerSpec.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/api/CycleCountApiControllerSpec.groovy
@@ -159,9 +159,7 @@ class CycleCountApiControllerSpec extends Specification implements DataTest, Con
     void 'createRequests should fail when batch has duplicated product'() {
         given: "the following commands"
         CycleCountRequestCommand cycleCountRequestCommand = new CycleCountRequestCommand(
-                product: Spy(Product) {
-                    getProductCode() >> PRODUCT_CODE
-                },
+                product: new Product(productCode: PRODUCT_CODE),
                 blindCount: true,
         )
         CycleCountRequestBatchCommand cycleCountRequestBatchCommand = new CycleCountRequestBatchCommand(requests: [cycleCountRequestCommand])
--- a/src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountItemSpec.groovy
+++ b/src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountItemSpec.groovy
@@ -27,50 +27,60 @@ class CycleCountItemSpec extends Specification implements DomainUnitTest<CycleCo
 
     void 'compareTo should return items in order starting with expiration date and location name'() {
         given:
-        mockDomain(InventoryItem)
-        mockDomain(Location)
         CycleCountItem cycleCountItem1 = new CycleCountItem(
-                inventoryItem: new InventoryItem(),
+                inventoryItem: new InventoryItem(
+                        expirationDate: new Date(2025, 03, 02)
+                ),
                 location: new Location()
         )
         CycleCountItem cycleCountItem2 = new CycleCountItem(
-                inventoryItem: new InventoryItem(),
-                location: new Location()
+                inventoryItem: new InventoryItem(
+                        expirationDate: new Date(2025, 02, 02),
+                ),
+                location: new Location(
+                        name: 'A',
+                )
         )
         CycleCountItem cycleCountItem3 = new CycleCountItem(
-                inventoryItem: new InventoryItem(),
-                location: new Location()
+                inventoryItem: new InventoryItem(
+                        expirationDate: new Date(2025, 02, 02),
+                ),
+                location: new Location(
+                        name: 'B',
+                )
         )
         CycleCountItem cycleCountItem4 = new CycleCountItem(
-                inventoryItem: new InventoryItem(),
-                location: new Location()
+                inventoryItem: new InventoryItem(
+                        expirationDate: new Date(2025, 01, 02),
+                ),
+                location: new Location(
+                        name: 'C',
+                )
         )
         CycleCountItem cycleCountItem5 = new CycleCountItem(
-                inventoryItem: new InventoryItem(),
-                location: new Location()
+                inventoryItem: new InventoryItem(
+                        expirationDate: new Date(2025, 01, 02),
+                ),
+                location: new Location(
+                        name: 'C',
+                )
         )
 
         and:
-        cycleCountItem1.inventoryItem.expirationDate = new Date(2025, 03, 02)
-        cycleCountItem2.inventoryItem.expirationDate = new Date(2025, 02, 02)
-        cycleCountItem3.inventoryItem.expirationDate = new Date(2025, 02, 02)
-        cycleCountItem4.inventoryItem.expirationDate = new Date(2025, 01, 02)
-        cycleCountItem5.inventoryItem.expirationDate = new Date(2025, 01, 02)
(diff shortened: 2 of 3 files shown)


src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountSpec.groovy
```diff
@@ -265,4 +265,49 @@ class CycleCountSpec extends Specification implements DomainUnitTest<CycleCount>
         expect:
         assert cycleCount.recomputeStatus() == CycleCountStatus.COMPLETED
     }
+
+    void 'getMaxCountIndex should return #maxCountIndex when there are items with #countIndexes count indexes'() {
+        given:
+        Set<CycleCountItem> cycleCountItems = new TreeSet<>()
+        for (int countIndex : countIndexes) {
+            CycleCountItem cycleCountItem = Spy(CycleCountItem)
```
EWaterman, line 273: Is the Spy and the TreeSet needed because otherwise the items are non-unique because they all have id==null?

If you wanted to avoid needing to mock/stub out the CycleCountItem so that you can test with real methods, you can do:
```
CycleCountItem item = new CycleCountItem(
        countIndex: 1,
        location: bin1,
        status: CycleCountItemStatus.APPROVED
)
item.id = 1
CycleCount cycleCount = CycleCount cycleCount = new CycleCount(
        cycleCountItems: [item]
)
```
Kind of weird to have to set the id manually, but that's the best solution that I've found

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountItemSpec.groovy
```diff
@@ -0,0 +1,78 @@
...
+        10              | 3              || 7
+        10              | 20             || -10
+        null            | 1              || null
+        1               | null           || null
+    }
+
+    void 'compareTo should return items in order starting with expiration date and location name'() {
+        given:
+        mockDomain(InventoryItem)
+        mockDomain(Location)
```
EWaterman, line 31: you can do `mockDomains(InventoryItem, Location)`. But do you even need to mock the domains? I usually don't need to unless I call `.save()` on the domain

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountItemSpec.groovy
```diff
@@ -0,0 +1,78 @@
...
+        and:
+        cycleCountItem1.inventoryItem.expirationDate = new Date(2025, 03, 02)
+        cycleCountItem2.inventoryItem.expirationDate = new Date(2025, 02, 02)
+        cycleCountItem3.inventoryItem.expirationDate = new Date(2025, 02, 02)
+        cycleCountItem4.inventoryItem.expirationDate = new Date(2025, 01, 02)
+        cycleCountItem5.inventoryItem.expirationDate = new Date(2025, 01, 02)
+        cycleCountItem2.location.name = 'A'
+        cycleCountItem3.location.name = 'B'
+        cycleCountItem4.location.name = 'C'
+        cycleCountItem5.location.name = 'C'
```
EWaterman, line 62: This is fine, but I find it clearer to declare the fields in the constructor:
```
CycleCountItem cycleCountItem1 = new CycleCountItem(
        inventoryItem: new InventoryItem(
            expirationDate: new Date(2025, 03, 02),
        ),
        location: new Location(
            name: 'A',
        ),
)
...
```

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountItemSpec.groovy
```diff
@@ -0,0 +1,78 @@
...
+        List<CycleCountItem> cycleCountItems = new ArrayList<>()
+        cycleCountItems.addAll(Arrays.asList(cycleCountItem5, cycleCountItem4, cycleCountItem3, cycleCountItem2, cycleCountItem1));
+        when:
+        cycleCountItems.forEach { println(it.hashCode()) }
+        println("====")
+        cycleCountItems.sort()
+        cycleCountItems.forEach { println(it.hashCode()) }
```
EWaterman, line 69: old print statements?

EWaterman: also can't you do?
```
List<CycleCountItem> cycleCountItems = [cycleCountItem5, cycleCountItem4, cycleCountItem3, cycleCountItem2, cycleCountItem1];
```

src/test/groovy/unit/org/pih/warehouse/api/CycleCountApiControllerSpec.groovy
```diff
@@ -0,0 +1,281 @@
...
+        response.json.data.size() == 2
+        response.json.totalCount == 2
+    }
+
+    void 'createRequests should fail when batch has duplicated product'() {
+        given: "the following commands"
+        CycleCountRequestCommand cycleCountRequestCommand = new CycleCountRequestCommand(
+                product: Spy(Product) {
+                    getProductCode() >> PRODUCT_CODE
+                },
```
EWaterman, line 164: why Spy instead of `new Product(productCode: PRODUCT_CODE)`?

src/test/groovy/unit/org/pih/warehouse/api/CycleCountApiControllerSpec.groovy
```diff
@@ -0,0 +1,281 @@
...
+        CycleCountRequest.metaClass.static.findByProductAndFacilityAndStatusNotInList = {
+            Product product, Location location, List<CycleCountRequestStatus> statuses ->
+            return Mock(CycleCountRequest)
+        }
+        Location.metaClass.static.findById = {
+            String id -> Mock(Location)
+        }
+
+        when: "the command is invalid"
+        cycleCountRequestBatchCommand.validate()
```
EWaterman, line 182: I don't think asserts like this register under a "when" block. Also shouldn't it be false since it fails validation? You'd want:
```
...

then:
assert !cycleCountRequestBatchCommand.validate()

when:
controller.createRequests(cycleCountRequestBatchCommand)

then:
thrown(ValidationException)
```
(I like to explicitly declare my asserts with "assert" in tests.)

EWaterman (commented):
nice work writing tests! :+1:

src/test/groovy/unit/org/pih/warehouse/api/CycleCountApiControllerSpec.groovy
alannadolny: I didn't treat this as an assertion - after calling the `validate`, the errors are assigned to `CycleCountRequestCommand`, so that the errors are thrown in `createRequests`

src/test/groovy/unit/org/pih/warehouse/inventory/CycleCountSpec.groovy
```diff
@@ -273,4 +273,52 @@ class CycleCountSpec extends Specification implements DomainUnitTest<CycleCount>
...
+
+    void 'getItemsOfMostRecountCount should return #numberOfItems items with the highest count index: #highestCountIndex when there #countIndexes count indexes'() {
+        given:
+        Set<CycleCountItem> cycleCountItems = new TreeSet<>()
+        for (int countIndex : countIndexes) {
+            CycleCountItem cycleCountItem = Spy(CycleCountItem)
+            cycleCountItem.countIndex = countIndex
+            cycleCountItems.add(cycleCountItem)
+        }
+        CycleCount cycleCount = Spy(CycleCount)
```
EWaterman, line 310: if you're not using Spy anymore in the above test, probably should be consistent and not use it here either

PR #5218, opened 2025-04-18
OBPIH-7237 Fix reload issue for empty cycle count items

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
@@ -113,6 +113,10 @@ class CycleCount {
         return cycleCountItems.findAll{ it.countIndex == countIndex}
     }
 
+    Set<CycleCountItem> getItemsOfSpecificCount(Integer countIndex) {
+        return cycleCountItems.findAll { it.countIndex == countIndex }
+    }
+
     /**
      * @return a list of all the products being counted by the cycle count.
      */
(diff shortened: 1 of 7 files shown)


grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -113,6 +113,10 @@ class CycleCount {
         return cycleCountItems.findAll{ it.countIndex == countIndex}
```
EWaterman, line 113: if you wanted, you could refactor this method to do `return getItemsOfSpecificCount(maxCountIndex)`

grails-app/services/org/pih/warehouse/inventory/CycleCountProductAvailabilityService.groovy
```diff
@@ -202,18 +204,18 @@ class CycleCountProductAvailabilityService {
         // Output fields
         CycleCountItemsForRefresh cycleCountItemsForRefresh
 
-        CycleCountRefreshState(CycleCount cycleCount) {
+        CycleCountRefreshState(CycleCount cycleCount, Integer countIndex) {
             this.cycleCount = cycleCount
 
             // We only want to refresh the items of the most recent count. We do this because previous counts are
             // already completed, so their data should be considered static.
```
EWaterman, line 211: can you update the comment to match your change? Something like:
```
// We only want to refresh the items of the most recent count. We do this because previous counts are
// already completed, so their data should be considered static. We allow countIndex to be specified
// for the case where the current count index has no items (QoH has been zeroed out for all items)
// and so we can't assume currentCountIndex == maxCountIndex.
```

EWaterman (commented):
You might need to update the tests in CycleCountProductAvailabilityServiceSpec. It shouldn't be too hard hopefully but let me know if you have questions.

awalkowiak (approved):

PR #5221, opened 2025-04-22
OBPIH-7042 Add Zone to Bin location filter in cycle count tabs

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useCycleCountFilters.js
+++ b/src/js/hooks/cycleCount/useCycleCountFilters.js
@@ -105,13 +105,11 @@ const useCycleCountFilters = () => {
         fetchBins(currentLocation?.id),
       ]);
 
-      const groupedBins = groupBinLocationsByZone(binList, translate);
-
       setSelectOptions({
         categories: categoryList,
         catalogs: catalogList,
         tags: tagList,
-        internalLocations: groupedBins,
+        internalLocations: groupBinLocationsByZone(binList, translate),
         abcClasses: classificationList,
       });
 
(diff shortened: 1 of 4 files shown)


src/js/hooks/cycleCount/useCycleCountFilters.js
```diff
@@ -106,7 +109,7 @@ const useCycleCountFilters = () => {
         categories: categoryList,
         catalogs: catalogList,
         tags: tagList,
-        internalLocations: binList,
+        internalLocations: groupBinLocationsByZone(binList, translate),
```
kchelstowski, line 112: could you describe the reason for translate needing to be passed to the method?

SebastianLib: I had to pass translate to groupBinLocationsByZone because I was getting an "Invalid Hook Call" error in useCycleCountFilters when I tried using it without passing translate. That’s how I worked around it
image

awalkowiak: And why was it working in the `useCountStepTable` before, without needing to pass it with params?

SebastianLib: I think we didn’t have to pass translate to groupBinLocationsByZone in useCountStepTable because the function is called synchronously during render. But in useCycleCountFilters, it’s used inside an async function, so useTranslate doesn’t work there

kchelstowski (approved):

PR #5233, opened 2025-04-24
OBPIH-7178 Batch endpoint for insert/update cycle count items

kchelstowski:

**Description:**

---


src/main/groovy/org/pih/warehouse/inventory/CycleCountUpdateItemCommand.groovy
```diff
@@ -23,7 +25,7 @@ class CycleCountUpdateItemCommand implements Validateable {
     Date dateCounted
 
     def beforeValidate() {
-        String cycleCountItemId = RequestContextHolder.getRequestAttributes().params?.cycleCountItemId
+        String cycleCountItemId = RequestContextHolder.getRequestAttributes().params?.cycleCountItemId ?: id
```
EWaterman, line 28: just describing this for myself because I was confused initially.

So if the cc item id is in the URI, use that, else assume it's in the request payload (and if its in both, use the one in the URI). And I'm assuming we support both cases because we still have the non-batched API, which puts it in the URI.

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -712,10 +718,13 @@ const useResolveStep = () => {
     // Nested path in colum names contains "_" instead of "."
     const nestedPath = columnId.replaceAll('_', '.');
     // Update data for: cycleCount (table) -> cycleCountItem (row) -> column (nestedPath)
+    const valueChanged = _.get(tableData.current, `[${tableIndex}].cycleCountItems[${rowIndex}].${nestedPath}`) !== value;
```
kchelstowski, line 721: there was another bug with the `console.log` not displaying the live data (ref: https://github.com/openboxes/openboxes/pull/5215#discussion_r2048905176)
it was difficult to determine what the issue is, but it turned out that `updateData` method is called a lot of times after each rerender, not only when we actually update the data, so it always set the `updated` flag to `true` even if something a while ago set it to `false` (`markAllItemsAsNotUpdated` method).
So this `valueChanged` flag is needed to determine if a value really changed and only then update the `updated` flag to `true`.

EWaterman (approved):

PR #5236, opened 2025-04-26
OBPIH-7041: prepend zone to bin location in CC tables

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7041

**Description:** Prepend zone to bin location name in all cycle count table:
- All products, to count, and to recount tabs
- count and recount steps
- count and recount confirmation pages

CURRENTLY LEFT TO DO:
- add zone for bins on all products tab
- add zone for non-custom rows of count and recount steps
- recount step is including zone twice in name for custom rows

---

Bins now have zone (ex: "Container 1") prepended to their display name:
Screenshot from 2025-04-25 09-32-54

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
@@ -53,7 +53,7 @@ class CycleCountItem implements Comparable {
                 facility: facility.toBaseJson(),
                 product: product,
                 inventoryItem: inventoryItem,
-                binLocation: location?.toBaseJson(),
+                binLocation: location?.toJson(),
                 countIndex: countIndex,
                 status: status,
                 quantityOnHand: quantityOnHand,
--- a/grails-app/migrations/views/cycle-count-session.sql
+++ b/grails-app/migrations/views/cycle-count-session.sql
@@ -5,7 +5,7 @@ SELECT
     CRC32(CONCAT(product_availability.location_id, product_availability.product_id)) as id,
 
     # Core properties
-    location.inventory_id                                                            as inventory_id,
+    facility.inventory_id                                                            as inventory_id,
     product_availability.product_id                                                  as product_id,
     product_availability.location_id                                                 as facility_id,
 
@@ -37,14 +37,14 @@ SELECT
     # Inventory Item Count
     count(product_availability.id)                                                   as inventory_item_count,
 
-    # Comma-separated list of internal locations
+    -- A comma-separated list of internal locations in the format <zone>:<bin location name>
     GROUP_CONCAT(DISTINCT (
-        CASE
-            WHEN product_availability.quantity_on_hand != 0 THEN product_availability.bin_location_name
-            ELSE NULL
-            END
+        IF(
+            product_availability.quantity_on_hand != 0,
+            CONCAT_WS(':', zone.name, product_availability.bin_location_name),
+            NULL
         )
-    )                                                                                as internal_locations,
+    ))                                                                                as internal_locations,
 
     # Quantities by SKU
     -- FIXME The sum could include both positive and negative quantities since we no longer exclude negative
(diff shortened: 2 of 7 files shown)


src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -269,7 +269,8 @@ const useCountStepTable = ({
 
   const columns = [
     columnHelper.accessor(
-      (row) => (row?.binLocation?.label ? row?.binLocation : row.binLocation?.name), {
+      // TODO: This doesn't seem to do anything?????? How to change the value???
+      (row) => getBinLocationToDisplay(row?.binLocation), {
```
EWaterman, line 273: :thinking:

grails-app/migrations/views/cycle-count-session.sql
```diff
@@ -37,14 +37,14 @@ SELECT
     # Inventory Item Count
     count(product_availability.id)                                                   as inventory_item_count,
 
-    # Comma-separated list of internal locations
+    -- A comma-separated list of internal locations in the format <zone>:<bin location name>
```
alannadolny, line 40: Don't we need a space after `:`?

grails-app/migrations/views/pending_cycle_count_request.sql
```diff
@@ -15,24 +15,23 @@ CREATE OR REPLACE VIEW pending_cycle_count_request AS
...
         cycle_count_request ccr
             LEFT OUTER JOIN product_availability pa
                 ON ccr.product_id = pa.product_id
                 AND ccr.facility_id = pa.location_id
             LEFT OUTER JOIN cycle_count cc
                 ON ccr.cycle_count_id = cc.id
-            JOIN location ON pa.location_id = location.id
+            JOIN location facility ON pa.location_id = facility.id
+            LEFT OUTER JOIN location bin_location ON pa.bin_location_id = bin_location.id
+            LEFT OUTER JOIN location zone ON bin_location.zone_id = zone.id
```
alannadolny, line 31: Have you checked the performance with/without your changes? There are a lot of new joins, so I am curious how it looks right now :thinking:

src/js/hooks/cycleCount/useCountStepTable.jsx
alannadolny: Those cells that you can see in the table come from the "cell" component created in our "defaultColumn" definition (that's the code that I wanted to split into a separate component, currently it's :spaghetti:). Try to use your function  within the code block starts with 
image

grails-app/migrations/views/pending_cycle_count_request.sql
EWaterman: it's hard for me to test on my local since the performace is all over the place. It doesn't seem to be significantly slower than it was before but it was already really quite slow so I hope the task for materializing the views will help with that

src/js/hooks/cycleCount/useCountStepTable.jsx
EWaterman: I poked around with it a bit more and still wasn't able to figure it out. I can't simply set the value of the TableCell Component to be getBinLocationToDisplay(value) because we only want to do that if a value is selected(?) It works fine currently when selecting new values, but existing, non-custom rows don't get the prefix. I wonder if it needs to be set differently in the getFieldProps for the SelectField, but I can't figure out how.

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -53,7 +53,7 @@ class CycleCountItem implements Comparable {
                 facility: facility.toBaseJson(),
                 product: product,
                 inventoryItem: inventoryItem,
-                binLocation: location?.toBaseJson(),
+                binLocation: location?.toJson(),
```
awalkowiak, line 56: I think it would be better to add zone name and zone id to the `toBaseJson()` instead using `toJson`, because it checks some activity codes, etc, which might degrade performance in huge cycle counts.

awalkowiak (changes requested):

src/js/utils/groupBinLocationsByZone.jsx
```diff
@@ -2,6 +2,15 @@ import React from 'react';
 
 import _ from 'lodash';
 
+/**
+ * Stringify a bin location for display, prepending the bin's zone if it has one and the zone
+ * isn't already prepended.
+ */
+const getBinLocationToDisplay = (bin) => (
+  (bin?.zoneName == null || bin?.name.startsWith(`${bin?.zoneName}: `))
```
kchelstowski, line 10:
```suggestion
  (bin?.zoneName === null || bin?.name.startsWith(`${bin?.zoneName}: `))
```

kchelstowski (commented):
Performance would be my only concern, so if it's fine, then I'm fine with the PR. 
Remember though to use `===` comparison in JS.

PR #5240, opened 2025-05-05
OBS-1866 prevent picking during outbounds with no product availability

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBS-1866

**Description:** There was a pending outbound that had a pick for product 10665 on the default bin, which I assume had non-zero quantity at the time it was picked. For some unknown reason, after an internal stock transfer that moved the stock to a different bin, the default bin no longer had a product inventory record. This allowed a user to submit the pending outbound which had a pick on the default bin, even though the bin had zero quantity. This caused the default bin to have a negative quantity.

What caused the product inventory record to not exist is unknown, but this code change handles the case where we're trying to pick from a bin that does not have a product inventory. We'll now throw an invalid pick error in that scenario.

### DIFF
--- a/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
+++ b/grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
@@ -661,7 +661,7 @@ class ShipmentService {
         shipment.shipmentItems.each { shipmentItem ->
             // FIXME this seems a little expensive so we should rework validateShipmentItem to not throw an exception
             try {
-                validateShipmentItem(shipmentItem, true)
+                validateShipmentItem(shipmentItem)
             } catch (ValidationException e) {
                 log.warn("Validation error " + e.message)
                 if (!errors) {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/shipping/ShipmentService.groovy
```diff
@@ -699,26 +699,15 @@ class ShipmentService {
...
-    }
-
     /**
      * Validate the shipment item when it's being added to the shipment.
      *
      * @param shipmentItem shipment item to validate
-     * @param binLocationRequired if shipment item has a bin location we validate quantity only against that bin
      * @return
      */
-    boolean validateShipmentItem(ShipmentItem shipmentItem, boolean binLocationRequired) {
```
EWaterman, line 719: binLocationRequired wasn't actually being used (looks like it was [leftover after an old refactor](https://github.com/openboxes/openboxes/commit/74b5206d991bbde9276fce60c2d732704f56aa91#diff-a701539d6f9ea99ef6718917e53ab215d7b15ac68ecd7a34443ff2a689f8e823)) so I merged the above method with this one.

ewaterman:
Not to be merged until after the release

PR #5242, opened 2025-05-07
OBPIH-7252 Create initial API endpoint for cycle count transaction report

jmiranda:


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -997,6 +997,11 @@ class UrlMappings {
             action = [POST: "refreshCycleCount"]
         }
 
+        "/api/reports/cycle-count-details" {
+            controller = "cycleCountApi"
+            action = [POST: "getCycleCountTransactionReport", GET: "getCycleCountTransactionReport"]
```
alannadolny, line 1002: Why do we need POST and GET for the same method? I think it should be only GET to be restful

alannadolny (commented):
I think the `totalCount` property is missing. It's necessary for the backend pagination.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
jmiranda: @alannadolny I was thinking I'd future-proof the endpoint in case we ever need to send requests with more complex search filters. The initial filters are probably ok as GET request parameters, but I'm imagining cases where reporting resources will need to handle complex filter expressions that are only really possible with POST requests.
```
"filterExpression": {
    "operator": "OR",
    "conditions": [
     {
       "field": "variancePercentage",
       "operator": "GT",
       "value": 10
      },
      {
        "field": "varianceAmount",
        "operator": "GT",
        "value": 1000.00
      }
    ]
}
```
But even in the simplest cases, it's more straightforward to provide the client with a way to build a POST body that includes all of the filters, groupBys, pagination, etc.
```
"startDate": "2025-04-01",
"endDate": "2025-04-30",
"filters": [
    "facilities": [1],
    "products": ["10001", "10002"],
    ...
],
"groupBy": [],
"orderBy": [],
"pagination": { 
    "max": "10", 
    "offset": 0 
}
```

PR #5253, opened 2025-05-14
OBPIH-7269 Add import count widget (dropzone on UI, upload endpoint, …

kchelstowski:
…importer class)

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -262,6 +262,7 @@ class CycleCountService {
         cycleCounts.each { CycleCountDto cycleCount ->
             cycleCount.cycleCountItems.each { CycleCountItemDto item ->
                 data << [
+                        "Cycle count id": cycleCount.id,
                         "Product Code": item.product.productCode,
                         "Product Name": item.product.name,
                         "Lot Number": item.inventoryItem.lotNumber,
--- a/src/js/api/services/CycleCountApi.js
+++ b/src/js/api/services/CycleCountApi.js
@@ -8,7 +8,7 @@ import {
   CYCLE_COUNT_REQUESTS,
   CYCLE_COUNT_START,
   CYCLE_COUNT_SUBMIT_COUNT,
-  CYCLE_COUNT_SUBMIT_RECOUNT,
+  CYCLE_COUNT_SUBMIT_RECOUNT, IMPORT_CYCLE_COUNT_ITEMS, IMPORT_PACKING_LIST,
 } from 'api/urls';
 import apiClient, { apiClientCustomResponseHandler } from 'utils/apiClient';
 
@@ -48,4 +48,14 @@ export default {
     apiClient.post(CYCLE_COUNT_ITEMS_BATCH(locationId, cycleCountId), payload),
   updateCycleCountItems: (payload, locationId, cycleCountId) =>
     apiClient.patch(CYCLE_COUNT_ITEMS_BATCH(locationId, cycleCountId), payload),
+  importCycleCountItems: (file, locationId) => {
+    const formData = new FormData();
+    formData.append('importFile', file);
+    const config = {
+      headers: {
+        'content-type': 'multipart/form-data',
+      },
+    };
+    return apiClient.post(IMPORT_CYCLE_COUNT_ITEMS(locationId), formData, config);
+  },
 };
--- a/src/js/api/services/CycleCountApi.js
+++ b/src/js/api/services/CycleCountApi.js
@@ -8,7 +8,8 @@ import {
   CYCLE_COUNT_REQUESTS,
   CYCLE_COUNT_START,
   CYCLE_COUNT_SUBMIT_COUNT,
-  CYCLE_COUNT_SUBMIT_RECOUNT, IMPORT_CYCLE_COUNT_ITEMS, IMPORT_PACKING_LIST,
+  CYCLE_COUNT_SUBMIT_RECOUNT,
+  IMPORT_CYCLE_COUNT_ITEMS,
 } from 'api/urls';
 import apiClient, { apiClientCustomResponseHandler } from 'utils/apiClient';
 
--- a/src/js/api/services/CycleCountApi.js
+++ b/src/js/api/services/CycleCountApi.js
@@ -3,13 +3,13 @@ import queryString from 'query-string';
 import {
   CYCLE_COUNT, CYCLE_COUNT_ITEM,
   CYCLE_COUNT_ITEMS_BATCH,
+  CYCLE_COUNT_ITEMS_IMPORT,
   CYCLE_COUNT_RECOUNT_START,
   CYCLE_COUNT_REFRESH_ITEMS,
   CYCLE_COUNT_REQUESTS,
   CYCLE_COUNT_START,
   CYCLE_COUNT_SUBMIT_COUNT,
   CYCLE_COUNT_SUBMIT_RECOUNT,
-  IMPORT_CYCLE_COUNT_ITEMS,
 } from 'api/urls';
 import apiClient, { apiClientCustomResponseHandler } from 'utils/apiClient';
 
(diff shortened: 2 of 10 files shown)


src/js/api/services/CycleCountApi.js
```diff
@@ -8,7 +8,7 @@ import {
   CYCLE_COUNT_REQUESTS,
   CYCLE_COUNT_START,
   CYCLE_COUNT_SUBMIT_COUNT,
-  CYCLE_COUNT_SUBMIT_RECOUNT,
+  CYCLE_COUNT_SUBMIT_RECOUNT, IMPORT_CYCLE_COUNT_ITEMS, IMPORT_PACKING_LIST,
```
alannadolny, line 11:
```suggestion
  CYCLE_COUNT_SUBMIT_RECOUNT,
  IMPORT_CYCLE_COUNT_ITEMS,
  IMPORT_PACKING_LIST,
```

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -564,6 +565,23 @@ const useCountStep = () => {
...
+    if (e instanceof File) {
+      setImportFile(e);
+    }
+  };
+
+  const importItems = async () => {
+    try {
+      show();
+      const response = await cycleCountApi.importCycleCountItems(importFile, currentLocation?.id);
+      console.log(response);
```
alannadolny, line 578: Is it by accident?

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -564,6 +565,23 @@ const useCountStep = () => {
...
+    } finally {
+      hide();
+    }
+    // TODO: Map items to the table
```
alannadolny, line 582: I assume it's for another ticket?

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
```diff
@@ -0,0 +1,68 @@
...
+    static Map PROPERTY_MAP = [
+            cycleCountId:             ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            productCode:                ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "product.name":         ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "lotNumber":            ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "expirationDate":       ([expectedType: ExpectedPropertyType.DateJavaType, defaultValue: null]),
+            "binLocation":          ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "quantityCounted":       ([expectedType: ExpectedPropertyType.IntType, defaultValue: null]),
+            "comment":            ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+    ]
```
alannadolny, line 39: I don't know whether it's just GitHub formatting, but the indentations are not aligned

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: No, I left it here so that I don't have an eslint error that `response` is not used. Anyway this will be adjusted in the next ticket, so I guess nothing wrong with leaving it for now.

src/js/hooks/cycleCount/useCountStep.jsx
kchelstowski: Yeah

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -262,6 +262,7 @@ class CycleCountService {
         cycleCounts.each { CycleCountDto cycleCount ->
             cycleCount.cycleCountItems.each { CycleCountItemDto item ->
                 data << [
+                        "Cycle count id": cycleCount.id,
```
EWaterman, line 265: I'm not a huge fan of needing to put cycle count id in the xls file since I'd rather not expose our internal ids to users. Can the frontend not look up the count by product? If this is the only/best solution, we should discuss with Manon to confirm that it's okay to include it.

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
```diff
@@ -0,0 +1,68 @@
...
+                    H: "comment",
+            ]
+    ]
+
+    static Map PROPERTY_MAP = [
+            cycleCountId:             ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            productCode:              ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "product.name":           ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "lotNumber":              ([expectedType: ExpectedPropertyType.StringType, defaultValue: null]),
+            "expirationDate":         ([expectedType: ExpectedPropertyType.DateJavaType, defaultValue: null]),
```
EWaterman, line 35: What happens if the date is poorly formatted at this point? Does the import fail? If so we should probably think of a different solution since we eventually want users to be able to resolve bad data in the UI, In a previous tech huddle we discussed potentially making all these fields StringType and letting the frontend handle all the validation. But if the import doesn't fail this is fine.

EWaterman:
Can you add a screenshot or recording of what the widget looks like? It would be good to be able to review the actual design of it.

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
kchelstowski: I checked that, and if you e.g. provide a random string instead of a date, it is parsed to `null`, so I believe it's expected.

kchelstowski:
@EWaterman attaching a screenshot:
Screenshot from 2025-05-15 11-18-12

I have not changed anything in terms of the design, it's the same as in the outbound import.

src/js/api/urls.js
```diff
@@ -194,3 +194,4 @@ export const CYCLE_COUNT_ITEMS_BATCH = (locationId, cycleCountId) => `${CYCLE_CO
 export const CYCLE_COUNT_SUBMIT_COUNT = (locationId, cycleCountId) => `${CYCLE_COUNT(locationId)}/${cycleCountId}/count`;
 export const CYCLE_COUNT_SUBMIT_RECOUNT = (locationId, cycleCountId) => `${CYCLE_COUNT(locationId)}/${cycleCountId}/recount`;
 export const CYCLE_COUNT_REFRESH_ITEMS = (locationId, cycleCountId, removeOutOfStockItemsImplicitly) => `${CYCLE_COUNT(locationId)}/${cycleCountId}/refresh${removeOutOfStockItemsImplicitly ? '?removeOutOfStockItemsImplicitly=true' : ''}`;
+export const IMPORT_CYCLE_COUNT_ITEMS = (locationId) => `${CYCLE_COUNT(locationId)}/items/upload/count`;
```
alannadolny, line 197: To match this constant with our previous naming for cycle count actions, I would call it `CYCLE_COUNT_IMPORT_ITEMS`

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
EWaterman: This is fine for now but we might want to refactor this later to return it to the client as a string so that we can notify the user that they provided an invalid date. Otherwise if we just return null when we have an invalid format we wont be able to distinguish between the case where the lot just doesnt't have an expiry date (because it also returns null)

EWaterman (approved):
I'm approving but I think we should be ready discuss the design with Manon. I was originally expecting the importer to be a popup window (or an expandable section if that's easier).

Having it always be there is a bit awkward since some users will never use it and it takes up so much of the screen so they have to scroll just to see the table fully.

PR #5260, opened 2025-05-19
OBPIH-7270 Apply import cycle count items to count step

kchelstowski:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useCountStep.jsx
+++ b/src/js/hooks/cycleCount/useCountStep.jsx
@@ -12,7 +12,7 @@ import _ from 'lodash';
 import queryString from 'query-string';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
-import { getCurrentLocation, getCycleCountRequestIds } from 'selectors';
+import { getBinLocations, getCurrentLocation, getCycleCountRequestIds } from 'selectors';
 
 import {
   eraseDraft,
@@ -65,9 +65,11 @@ const useCountStep = () => {
   const {
     cycleCountIds,
     currentLocation,
+    binLocations,
   } = useSelector((state) => ({
     cycleCountIds: getCycleCountRequestIds(state),
     currentLocation: getCurrentLocation(state),
+    binLocations: getBinLocations(state),
   }));
 
   const resetFocus = () => {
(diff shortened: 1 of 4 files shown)


src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
```diff
@@ -42,17 +46,43 @@ class CycleCountItemsExcelImporter extends AbstractExcelImporter implements Data
...
+                    name: locationMap[row.binLocation]?.name
+            ]
+            row.product = [
+                    id: productMap[row.productCode]?.id,
+                    name: productMap[row.productCode]?.name,
+                    productCode: productMap[row.productCode]?.productCode
+            ]
+        }
+
+        return data
```
alannadolny, line 85: I added fetching locations and products in one database call to avoid performance issues. At this moment, I tried importing 200 rows. Getting the response took < 500ms, rendering lines on the frontend side took 1s

EWaterman: nice good idea :+1: :+1: :+1:

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
```diff
@@ -42,17 +46,43 @@ class CycleCountItemsExcelImporter extends AbstractExcelImporter implements Data
...
+        // 3. Replace data from importer with the fetched data
+        data.each { row ->
+            row.binLocation = [
+                    id: locationMap[row.binLocation]?.id,
+                    name: locationMap[row.binLocation]?.name
+            ]
+            row.product = [
+                    id: productMap[row.productCode]?.id,
+                    name: productMap[row.productCode]?.name,
+                    productCode: productMap[row.productCode]?.productCode
```
EWaterman, line 81: What happens when we've been given invalid input (ie a product code that doesn't map to a product or a name that doesn't map to a bin)? We still need to be able to return that code/name back to the client so that it can error properly. So I think it's safer to do:
```
productCode: productMap[row.productCode]?.productCode ?: row.productCode
```
and same for location:
```
name: locationMap[row.binLocation]?.name ?: row.binLocation
```
That way, if we can't resolve it to an entity, id is null (so the client will know it was invalid) but we preserve the name/code that was input by the user so the client can still use it for display.

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -575,15 +576,77 @@ const useCountStep = () => {
...
-      console.log(response);
+      let cycleCounts = _.groupBy(response.data.data, 'cycleCountId');
+      tableData.current = tableData.current.map((cycleCount) => ({
+        ...cycleCount,
+        cycleCountItems: [
+          ...cycleCount.cycleCountItems
+            .map((item) => {
+              const correspondingImportItem = cycleCounts[cycleCount.id]?.find(
+                (cycleCountItem) => cycleCountItem.cycleCountItemId === item.id,
+              );
```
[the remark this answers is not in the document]
kchelstowski: As I wrote in one of the comments and also on the tech huddle channel - since we allow to import invalid lines (and to add invalid custom lines, that are validated when clicking "Next", we can't make it (productCode + binLocation + lotNumber) a "unique key", because if we did so, we would not be able to add a new custom line with the same productCode, binLocation, lotNumber, and we would eventually update an existing line. 
What if a user gave three lines with the same productCode, binLocation, lotNumber? I don't feel comfortable with dealing with such case to be honest, unless we convince Manon that both "save progress" and "next" should behave in the same way in terms of the validation.

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -575,15 +576,77 @@ const useCountStep = () => {
...
+
+              if (correspondingImportItem) {
+                // Remove items from the import that have a corresponding item
+                // in the current cycle count. It allows us to treat items with
+                // the wrong ID as new rows that do not already exist.
```
kchelstowski, line 634: could you explain with an example what's the purpose of that? If a corresponding item exists, we should update its values instead of removing it, but maybe I don't understand the point of those lines?

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
kchelstowski: This is what I was afraid of when it was agreed that the import should only take care of parsing the data to JSON. It's getting hard to handle all those edge cases on the frontend, especially I've no idea how we would wanna display (in the table) an error that the product doesn't exist if at this point we wouldn't know which cycle count (which table) we are referring to (let's say a user didn't also provide the cycle count id).

src/js/hooks/cycleCount/useCountStep.jsx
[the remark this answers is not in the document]
EWaterman: That's why in my pseudocode above it maps that key to a list of rows. We wouldn't want to assume it's unique. It then loops through the list of rows matching that key, merging it with existing rows that also have that key (that's what the lines after `While int i < matchingCountTableRows.size && i < matchingXlsRows.size` is doing).

Lets discuss it in tech huddle tomorrow if you think it won't work

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -575,15 +576,77 @@ const useCountStep = () => {
...
+    [cycleCountId]: cycleCounts[cycleCountId]
+      .filter((item) => item.cycleCountItemId !== itemId),
+  });
+
+  const mergeImportItems = (originalItem, importedItem) => ({
+    ...originalItem,
+    ...importedItem,
+    inventoryItem: {
+      product: {
+        id: importedItem?.product?.id || originalItem?.inventoryItem?.product?.id,
```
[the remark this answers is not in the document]
EWaterman: So this case can happen if the row in the xls has the same cycle count item id, but the product id/code was changed in the xls for some reason?

I still think we can implement this without needing cycle count item id, so this wouldn't be a problem here but we can discuss.

But in case we keep count item id, I think if the count item id is the same but the product id is not the same, then we should probably error. I'm worried that it's not safe to assume that the user meant to put a different product in that row. They don't know what cycle count item id means (which is another reason why I don't like having it in the xls) so I don't think we can use cycle count item id alone as the "source of truth" for identifying matching rows. Product id must also match otherwise something strange has happened in the xls.

But again, I'd much rather not have cycle count item id in the xls at all

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
EWaterman: yeah I was imagining it'd be awkward, but in the case of invalid bin, I was hoping it'd be possible to alter the display value for bin to be the user input but then have the actual value of the field be something that represents invalid input. But I haven't actually tried to implement that so it could be impossible. I personally think it's fine for the backed to simply return null if a bin is invalid, as long as we have a good way of notifying to users that their input is lost/modified. The main thing I want to avoid is: the user does an import on bin "ABC" which doesn't exist (they meant to type "ABD", we set the bin to null, they don't notice, then submit the count and suddenly we have stock in the default bin instead of the bin they wanted ("ABD").

Lets talk with Manon about alternative solutions tomorrow.

awalkowiak (commented):
Discussed on the tech huddle, merging

PR #5262, opened 2025-05-20
OBPIH-7194 Refactor transactions after Record Stock

alannadolny:


grails-app/services/org/pih/warehouse/inventory/RecordStockProductInventoryTransactionService.groovy
```diff
@@ -0,0 +1,105 @@
...
+                    inventoryItem: binLocationItem.inventoryItem,
+            )
+            availableItems.add(availableItem)
+        }
+
+        return availableItems
+    }
+
+    Integer getTransactionEntryQuantity(List<AvailableItem> availableItems, Integer newQuantity, InventoryItem inventoryItem) {
+        AvailableItem availableItem = availableItems.find {
```
EWaterman, line 71: I implemented a quite similar solution for inventory import except I made this a method in the ProductAvailabilityService and instead of returning a list of AvailableItem, I return a Map<String, AvailableItem> keyed on [product + bin + lot] for easy lookups. I'm going to make a PR for that soon, but hopefully we can unify the behaviour

grails-app/services/org/pih/warehouse/inventory/RecordStockProductInventoryTransactionService.groovy
```diff
@@ -0,0 +1,49 @@
...
+        Transaction transaction = new Transaction(recordInventoryCommand.properties)
+        transaction.transactionType = TransactionType.get(Constants.ADJUSTMENT_CREDIT_TRANSACTION_TYPE_ID)
+        transaction.inventory = recordInventoryCommand.inventory
+        transaction.comment = recordInventoryCommand.comment
+        transaction.transactionNumber = inventoryService.generateTransactionNumber(transaction)
+        transaction.transactionDate = transactionDate
+        return transaction
+    }
+
+    TransactionEntry createTransactionEntry(
```
EWaterman, line 26: can you name this createAdjustmentTransactionEntry just for clarity? I originally made the ProductInventoryTransactionService class for the baseline transaction only, so it'd be good to clarify that this method is for adjustments only

awalkowiak (approved):

PR #5271, opened 2025-05-22
Disable MySQL 5.7 test runner

EWaterman:

**Link to GitHub issue or Jira ticket:** N/A

**Description:** We're not supporting MySQL 5.7 for 0.9.x releases any longer, and since we've started using CTEs (which are not supported in 5.7), this marks the final nail in the coffin for abandoning 5.7 support.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         # If we start actively supporting other databases, they should be added here so that we can test against them.
-        testdatabase: ["mysql:5.7.44", "mysql:8.0.36", "mariadb:10.3.39"]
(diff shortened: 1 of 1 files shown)


.github/workflows/backend-tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         # If we start actively supporting other databases, they should be added here so that we can test against them.
-        testdatabase: ["mysql:5.7.44", "mysql:8.0.36", "mariadb:10.3.39"]
+        testdatabase: ["mysql:8.0.36", "mariadb:10.3.39"]
```
EWaterman, line 20: I figured I'd make this a PR as a reminder for where this config lives. It's really easy to add/remove db versions to test against so if we upgrade in the future, all we need to do is remember to update this list

PR #5272, opened 2025-05-22
Arm64 support

mdpearson:
### :sparkles: Support for ARM64-based platforms

This PR makes OpenBoxes run correctly on Macs running Apple Silicon (or, at least, _my_ Mac running Apple Silicon ;-) )

**Link to GitHub issue or Jira ticket:** _What should I use for tickets?_

#### Description

I ran into a few gotchas with OpenBoxes on an ARM-based Mac, namely:
1. Node 14 only has x86-based builds.
2. MySQL 5 only has x86-based builds.
3. Different parts of the project try to pull in different JNA builds (4.2.2 and 5.8.0). Only the latter one works on ARM.
4. There's some mysterious interaction between Spring Boot, Gradle and possibly JNA on ARM that causes the app to look in the gradle daemon's working directory for project static assets.

This PR fixes all of them with a mix of documentation and code. Unfortunately it doesn't pass the tests cleanly; I thought it would be easier to discuss these problems here rather than in Slack.

#### Issues for discussion

I'm still seeing 28 test failures of the 0.9.4 build.

###### 17 test failures

- `org.pih.warehouse.api.spec.auth.AuthApiSpec`
- `org.pih.warehouse.api.spec.product.CategoryApiCRUDSpec`
- `org.pih.warehouse.api.spec.product.ProductApiCRUDSpec`
- `org.pih.warehouse.api.spec.product.ProductApiDemandSpec`
- `org.pih.warehouse.api.spec.product.ProductClassificationApiCRUDSpec`
- `org.pih.warehouse.api.spec.product.ProductClassificationApiListFiltersOutEmptySpec`

```stacktrace
java.lang.IllegalStateException: No GORM implementations configured. Ensure GORM has been initialized correctly
	at org.grails.datastore.gorm.GormEnhancer.findSingleDatastore(GormEnhancer.groovy:378)
	at org.grails.datastore.gorm.GormEnhancer.findSingleTransactionManager(GormEnhancer.groovy:397)
	at org.pih.warehouse.api.spec.base.ApiSpec.setup(ApiSpec.groovy)
```

###### 10 test failures

- `org.pih.warehouse.smoke.spec.IdentifierSpec`

```stacktrace
java.lang.NullPointerException: Cannot invoke method generate() on null object
	at org.pih.warehouse.smoke.spec.IdentifierSpec.invoiceIdentifierService can generate identifiers with the current configuration(IdentifierSpec.groovy:79)
```

###### 1 test failure

- `org.pih.warehouse.smoke.spec.LiquibaseUtilSpec`

```stacktrace
java.lang.IllegalArgumentException: GrailsApplication not found
	at org.springframework.util.Assert.notNull(Assert.java:134)
	at grails.util.Holders.getGrailsApplication(Holders.java:129)
	at util.LiquibaseUtil.getCurrentVersionsByFolderName(LiquibaseUtil.groovy:106)
	at org.pih.warehouse.smoke.spec.LiquibaseUtilSpec.getCurrentVersionsByFolderName should only the newest version for new installs(LiquibaseUtilSpec.groovy:15)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,21 @@ https://openboxes.atlassian.net/wiki/spaces/OBW/pages/1719435265/Push-button+dep
 * The database we use in production won't run natively on modern Macs.
   * Easiest way to install a compatible database is by using [`brew install mariadb`](https://mariadb.com/kb/en/installing-mariadb-on-macos-using-homebrew/).
   * Alternatively, ARM builds of MySQL 8 are available. Consider downloading [8.x LTS here](https://dev.mysql.com/downloads/mysql/).
+* No ARM build of Node 14 exists either.
+  * But you can run it via x86 emulation as follows:
+    1. `/usr/sbin/softwareupdate --install-rosetta --agree-to-license`
+    2. Install `nvm` (_only_ – do _not_ install `node` itself) using [these instructions](https://nodejs.org/en/download)
+    3. Launch a terminal that uses Rosetta to pretend to be an x86 host: `arch -x86_64 bash`
+    4. Within that terminal, run `nvm install 14`
+    5. Open up a new terminal and run the following command; you should see an `x86_64` binary, which should be able to run successfully on an ARM host.
+      ```bash
+      $ hash -r && type node | awk '{print $3}' | xargs file
+      \*\*/.nvm/versions/node/v14.21.3/bin/node: Mach-O 64-bit executable x86_64
+      $ arch; node --version
+      arm64
+      v14.21.3
+      ```
+
 #### Optional
 * [IntelliJ IDEA](https://www.jetbrains.com/idea/download/)
 * Chrome
--- a/README.md
+++ b/README.md
@@ -195,13 +195,8 @@ npm run watch
 ```
 
 #### 9. Upgrade the project to the currently installed grails version 
-Either of the following actions (upgrade, compile, run-app) should generate the all important Spring configuration 
+Either of the following actions (compile, run-app) should generate the all important Spring configuration 
 (`/WEB-INF/applicationContext.xml`) and start the dependency resolution process.  
-
-```    
-grails upgrade
-```
-OR
 
 ```    
 grails compile
--- a/README.md
+++ b/README.md
@@ -91,13 +91,18 @@ https://openboxes.atlassian.net/wiki/spaces/OBW/pages/1719435265/Push-button+dep
 #### Required
 * [Java 8 (must install Java 8)](https://www.oracle.com/pl/java/technologies/javase/javase8-archive-downloads.html) or via SDK
 * [MySQL 5.7 or MySQL 8.0](https://downloads.mysql.com/archives/community/) or [MariaDB 10.11.4](https://mariadb.com/kb/en/mariadb-10-11-4-release-notes/)
-  * Mac users: 5.7.31 is the latest 5.7.x with a pre-built installer and works fine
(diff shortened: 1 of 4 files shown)


EWaterman:
The tests seem to be passing in this PR. Do you mean they're failing for you locally when you run them?

awalkowiak:
@EWaterman fyi, I am copy pasting slack repsonse from Matt
> Hey [@Artur](https://openboxes.slack.com/team/UABF02Z18) Tests seem to pass OK in GitHub for both MySQL 8 and MariaDB. The 5.7 tests wedged, but it sounds like that’s not so important. The tests I mentioned in the PR, so far as I can tell, are only failing locally. It looks like my changes didn’t break anything visible to GitHub so the failing tests may just be a sign that something in ARM land is off, or with my DB setup, since they all seem to be DB related. If you are dumping mysql5.7 support “soon” — what’s the plan going forward? MySQL 8? Some flavor of MariaDB? Both? Neither?

and my response there was:

> Plan forward is both - MySQL 8 and MariaDB

ewaterman:
@jmiranda @awalkowiak are you guys still wanting to merge this in?

I did a quick check on regular and dockerized setups and from what I can see the static files are loading properly. I'm not sure how it'll behave on a non-embedded tomcat env but I suppose we'll find out pretty quick if something is funky so I don't mind merging this in.

I'll leave it to you guys

jmiranda:
@ewaterman Yes, eventually. We had to take a pause this work because the VVG project decided not to target iOS for the mobile application. 

I think it would be a good idea to test this a bit more thoroughly and even add some logging to make sure the customizer thingy is being used in the non-embedded environment. 

What are your thoughts on the test failures? Are those failures caused by the change or were those happening already? In either case, we should probably keep working on this one until we can get some clarity.

jmiranda:
Note: In case the "target iOS for the mobile application" comment was confusing (because this PR is targeting core) ... the missing thread is that Matt was doing some iOS build work for the mobile application and was trying to get the server running locally as well. So that's why this PR is necessary. I'd prefer to find a non-coding solution, but we can implement this change if necessary. 

@mdpearson What I'd like to know from you is whether this is required to get the application running via one or all of the following deployment methods? 

* run-app (grails run-app)
* docker (docker compose up)
* deployed war (openboxes.war -> webapps)
* standalone war (java -jar)

ewaterman:
I'm not sure about the tests. I've done some changes to the tests since this PR was created but I suspect that wouldn't resolve these issues as they appear to be specific to running them on arm platforms. That said, the tests run via testcontainers which initialize their own test database so I'm not certain why they'd fail. Maybe Matt was overriding the datasource to use his local db instead of testcontainers and there was some config weirdness causing the GORM errors. I also don't know what would cause spring bean errors in the other tests.

Definitely needs some more debugging.

Once this gets stable, we could consider adding a dedicated arm64 runner for github actions to verify the app can be built and run on arm architecture, but that's a down the line problem.

jmiranda:
@mdpearson Another question would be, what were the errors that lead us to conclude that a Grails application does not run in an `arm64` environment? 

> Link to GitHub issue or Jira ticket: What should I use for tickets?

~Create a GitHub issue for anything related to the community.~

I created the following ticket. Please add relevant error logs for the issues encountered.
https://github.com/openboxes/openboxes/issues/5642

mdpearson:
I'm going to close this PR for now and will push more focused changes as I come across them. Thanks for the commentary -- I'll definitely incorporate it in future work. Cheers

PR #5276, opened 2025-05-23
OBPIH-7143 return quantity allocated in cycle count pages

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7143

**Description:**
1) On the backend, return quantityAllocated to the client for all rows in all cycle count tables (via PendingCycleCountRequest and CycleCountCandidate views) so that the client can grey out rows with quantityAllocated > 0
2) On the backend, prevent a count from being requested on a product with quantityAllocated > 0

### DIFF
--- a/src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
+++ b/src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
@@ -20,11 +20,18 @@ class CycleCountRequestCommand implements Validateable {
 
     static constraints = {
         product(nullable: true, validator: { Product product, CycleCountRequestCommand obj ->
-            // TODO: Most probably, this will have to be modified not to include completed cycle counts
             CycleCountRequest cycleCountRequest = CycleCountRequest.findByProductAndFacilityAndStatusNotInList(product, obj.facility, [CycleCountRequestStatus.COMPLETED, CycleCountRequestStatus.CANCELED])
             if (cycleCountRequest) {
                 return ['duplicateExists', product.productCode]
             }
+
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestCommand.groovy
```diff
@@ -20,11 +20,18 @@ class CycleCountRequestCommand implements Validateable {
...
             if (cycleCountRequest) {
                 return ['duplicateExists', product.productCode]
             }
+
+            // If a product is in a pending outbound, we don't allow you to proceed with the count. There is too much
+            // stock uncertainty in that scenario so to avoid confusion we require the outbound to be resolved first.
+            CycleCountCandidate candidate = CycleCountCandidate.findByFacilityAndProduct(obj.facility, product)
+            if (candidate != null && candidate.quantityAllocated > 0) {
+                return ['hasQuantityAllocated', product.productCode]
+            }
```
EWaterman, line 33: This is slow. We have a task to materialize the CycleCountCandidate views so that would speed this up, but I'm  wondering if we even need to validate this on the backend.

We have https://pihemr.atlassian.net/browse/OBPIH-7046 which will have the frontend greying out rows where quantityAllocated > 0 so that would already be covering this case. The only reason to also have this logic on the backend would be if we want to protect against people calling the API directly (and so bypassing the frontend validation). But do we care about that? The backend logic still works if you do a count where quantityAllocated > 0 so there's nothing malicious that can be done.

EWaterman: Good idea. I don't know why I was determined to use CycleCountCandidate for this

PR #5278, opened 2025-05-26
OBPIH-7307 To Count and To Resolve products visible and disabled on All Products tab

SebastianLib:

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -94,8 +94,23 @@ class CycleCountService {
             //  product/facility pairs where quantity on hand is negative.
             // Moved this from the cycle count session view since it's a requirement of the candidate query,
             // not the cycle count session.
-            if (command.includeStockOnHandOrNegativeStock) {
+            if (command.includeStockOnHandOrNegativeStock && !command.includeNullStatus) {
                 eq("hasStockOnHandOrNegativeStock", command.includeStockOnHandOrNegativeStock)
+            }
+
+            // Products from the "To Count" and "To Resolve" tabs can have negative quantity,
+            // so we want to include them in the results.
+            if (command.includeNullStatus) {
+                or {
+                    inList("status", command.statuses)
+                    and {
+                        or {
+                            isNull("status")
+                            not { inList("status", command.statuses) }
+                        }
+                        eq("hasStockOnHandOrNegativeStock", command.includeStockOnHandOrNegativeStock)
+                    }
+                }
             }
 
             // FIXME Sort order should allow multiple sort order rules ("columna, -columnb"). We should consider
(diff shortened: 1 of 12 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -89,10 +94,23 @@ class CycleCountService {
...
                 eq("hasStockOnHandOrNegativeStock", command.includeStockOnHandOrNegativeStock)
             }
 
+            // Products from the "To Count" and "To Resolve" tabs can have negative stock,
+            // but we want to include them in the results
+            if (command.showCountAndResolveProducts) {
+                or {
+                    if (command.statuses) {
+                        inList("status", command.statuses)
+                    }
```
kchelstowski, line 107: in the code above when checking statuses, you say that if showCountAndResolveProducts, then show cycle counts with null status. Here you do something opposite. Could you explain the reason for double checking statuses here?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -89,10 +94,23 @@ class CycleCountService {
...
+                    and {
```
kchelstowski, line 108: what's the purpose of that `and` if you have only one statement inside?

src/js/hooks/cycleCount/useAllProductsTab.jsx
```diff
@@ -81,6 +83,14 @@ const useAllProductsTab = ({
...
     sortingParams,
   }) => _.omitBy({
+    statuses: [
+      CycleCountCandidateStatus.CREATED,
+      CycleCountCandidateStatus.REQUESTED,
+      CycleCountCandidateStatus.COUNTING,
+      CycleCountCandidateStatus.COUNTED,
+      CycleCountCandidateStatus.INVESTIGATING,
+    ],
+    showCountAndResolveProducts: true,
```
kchelstowski, line 93: btw I think something like `showCycleCountsInProgress` would sound better than "countAndResolve", since we might have also the approve step in the future and this would have to be changed if the cycle counts at the approval step should also be a part of that statement.

src/js/hooks/cycleCount/useCycleCountProductAvailability.jsx
```diff
@@ -0,0 +1,34 @@
+import CycleCountCandidateStatus from 'consts/cycleCountCandidateStatus';
+
+const useCycleCountProductAvailability = (status) => {
+  if (
+    status === CycleCountCandidateStatus.CREATED
+    || status === CycleCountCandidateStatus.REQUESTED
+    || status === CycleCountCandidateStatus.COUNTING
```
kchelstowski, line 7: maybe it'd be a good idea to create some simple util methods to say "isInProgress", "isCounting" etc.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
SebastianLib: I want to include all products that have statuses listed in command.statuses regardless of includeStockOnHandOrNegativeStock. For the rest, we filter (with statuses null) by hasStockOnHandOrNegativeStock, thanks to this the backend returns good results to me. am I right? Can i do it better?

PR #5279, opened 2025-05-26
:whale: Fixing deployment using Docker

ricardoapaes:

During my deployment attempts, I noticed several problems related to specifying files that no longer exist (likely due to renaming). With the applied changes, I can now successfully run the project as described in the documentation.


jmiranda:
Thank you, @ricardoapaes!

jmiranda:
@awalkowiak @EWaterman Not a priority, but it would be nice to figure out why the tests are failing.

docker/docker-compose.yml
```diff
@@ -1,10 +1,9 @@
-version: "3.8"
 services:
     app:
       extends: 
-        file: docker-compose-generic.yml
+        file: docker-compose-base.yml
         service: app
-      image: ghcr.io/openboxes/openboxes:feature-upgrade-to-grails-3.3.10
+      image: ghcr.io/openboxes/openboxes:v0.9.3
```
ewaterman, line 6: would we want this to be "latest"? Otherwise we'd need to make sure we update this each release

EWaterman (approved):

docker/docker-compose.yml
awalkowiak: @EWaterman After the initial approval on my end, I started wondering which one should actually be here. `Develop` or `Main`. I was leaning to develop, but thought we could change it later on our own, once it is merged.

PR #5282, opened 2025-05-27
OBPIH-7113 dynamic sorting on receiving locations

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7113

**Description:**

Video recording describing the PR: https://jam.dev/c/b2a026dc-56f9-4abc-9b9d-932cf401081f

Adds the ability to provide the "sort" query param when fetching internal locations. The param is a comma separated list of fields and you can put a '-' in front of a field to have it sort descending. For example: "?sort=x,-y" means sort ascending by x then descending by y.

I achieved this by adding a new SortParam class. We can bind the "sort" query param to a list of SortParam and then use SortUtil to easily sort a list of objects.

So for example, if we want to sort Locations by order then by name, both ascending:
```
List<Location> locations = [
    new Location(id: 0, order: 0, name: "Z"),
    new Location(id: 1, order: 1, name: "A"),
    new Location(id: 2, order: 0, name: "A")
]
SortParamList sortParams = [
    new SortParam(fieldName: "name", ascending: true)
    new SortParam(fieldName: "order", ascending: true)
]
List<Location> locationsSorted = SortUtil.sort(locations, sortParams)
assert locationsSorted[0].id == 2
assert locationsSorted[1].id == 0
assert locationsSorted[1].id == 1
```

And to use it in an API:
- client does a `GET /api/locations?sort=order,name`
- we bind that to a command object:
  ```
  class LocationCommand {
      SortParamList sort
  }
  ```
- then in the service:
  ```
  List<Location> getLocations(LocationCommand command) {
      List<Location> locations = Location.find(...)
      return SortUtil.sort(locations, command.sort)
  }
  ```

Note that this should only be used when we can't sort in the query itself since doing it in the query is typically faster.

I was originally going to implement this as the ticket suggested and just have the default sort behaviour be to sort by location type but I was worried about the changes it would have on other places in the code so I ended up implementing the full solution.

---

https://jam.dev/c/bd9c2d59-1a28-4830-aede-4336923bf636

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
@@ -16,7 +16,7 @@ import org.pih.warehouse.core.ActivityCode
 import org.pih.warehouse.core.InternalLocationSearchCommand
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.LocationTypeCode
-import org.pih.warehouse.core.ReceivingBinSearchCommand
+import org.pih.warehouse.core.ReceivingLocationSearchCommand
 
 @Transactional
 class InternalLocationApiController {
@@ -36,7 +36,7 @@ class InternalLocationApiController {
         render([data: locations, totalCount: locations?.totalCount] as JSON)
     }
 
-    def listReceiving(ReceivingBinSearchCommand command) {
+    def listReceiving(ReceivingLocationSearchCommand command) {
         command.location = getFacility(command)
 
         // When searching receiving bins, we need different default search parameters. Mainly we want to include
--- a/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
@@ -11,30 +11,22 @@ package org.pih.warehouse.api
 
 import grails.converters.JSON
 import grails.gorm.transactions.Transactional
+
 import org.pih.warehouse.core.ActivityCode
+import org.pih.warehouse.core.InternalLocationSearchCommand
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.LocationTypeCode
+import org.pih.warehouse.core.ReceivingBinSearchCommand
 
 @Transactional
 class InternalLocationApiController {
 
     def locationService
 
-    def list() {
-        String locationId = params?.location?.id ?: session?.warehouse?.id
-        Location parentLocation = locationId ? Location.get(locationId) : null
-        if (!parentLocation) {
-            throw new IllegalArgumentException("Must provide location.id as a request parameter")
-        }
+    def list(InternalLocationSearchCommand command) {
+        command.location = getFacility(command)
 
-        LocationTypeCode[] locationTypeCodes = params.locationTypeCode ? params.list("locationTypeCode") : [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION]
-
-        Map activityCodeParams = [:]
-        activityCodeParams.allActivityCodes = params.allActivityCodes ? params.list("allActivityCodes") as ActivityCode[]: []
-        activityCodeParams.anyActivityCodes = params.anyActivityCodes ? params.list("anyActivityCodes") as ActivityCode[]: []
-        activityCodeParams.ignoreActivityCodes = params.ignoreActivityCodes ? params.list("ignoreActivityCodes") as ActivityCode[] : []
-
-        List<Location> locations = locationService.getInternalLocations(parentLocation, locationTypeCodes, activityCodeParams)
+        List<Location> locations = locationService.getInternalLocations(command)
         render([data: locations?.collect { [id: it.id, name: it.name, zoneId: it.zone?.id, zoneName: it.zone?.name] }] as JSON)
     }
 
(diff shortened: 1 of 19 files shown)


grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
```diff
@@ -44,27 +36,19 @@ class InternalLocationApiController {
...
-        Map activityCodeParams = [:]
-        activityCodeParams.allActivityCodes = params.allActivityCodes ? params.list("allActivityCodes") as ActivityCode[]: []
-        activityCodeParams.anyActivityCodes = params.anyActivityCodes ? params.list("anyActivityCodes") as ActivityCode[]: []
-        activityCodeParams.ignoreActivityCodes = params.ignoreActivityCodes ? params.list("ignoreActivityCodes") as ActivityCode[] : [ActivityCode.RECEIVE_STOCK]
-
-        String[] receivingLocationNames = [locationService.getReceivingLocationName(shipmentNumber), "Receiving ${shipmentNumber}"]
-        List<Location> locations = locationService.getInternalLocations(parentLocation, locationTypeCodes, activityCodeParams, receivingLocationNames)
+        // When searching receiving bins, we need different default search parameters. Mainly we want to include
+        // the normal bin locations as well as only the receiving bins that match the given shipment number.
+        command.locationTypeCode = params.locationTypeCode ? params.list("locationTypeCode") : [LocationTypeCode.BIN_LOCATION]
```
EWaterman, line 44: Something weird is going on here. It's throwing the following:
```
Java.lang.String cannot be cast to java.lang.Enum
```
when the query in LocationService runs:
```
List<Location> internalLocations = Location.createCriteria().list() {
    locationType {
        'in'("locationTypeCode", command.locationTypeCode)
    }
}
```
For some reason it works totally fine for cycle count and command.locationTypeCode is exactly the same in both those scenarios so I don't know what's going on. I'm still looking into it

grails-app/controllers/org/pih/warehouse/api/InternalLocationApiController.groovy
```diff
@@ -11,30 +11,22 @@ package org.pih.warehouse.api
...
-            throw new IllegalArgumentException("Must provide location.id as a request parameter")
-        }
-
-        LocationTypeCode[] locationTypeCodes = params.locationTypeCode ? params.list("locationTypeCode") : [LocationTypeCode.INTERNAL, LocationTypeCode.BIN_LOCATION]
-
-        Map activityCodeParams = [:]
-        activityCodeParams.allActivityCodes = params.allActivityCodes ? params.list("allActivityCodes") as ActivityCode[]: []
-        activityCodeParams.anyActivityCodes = params.anyActivityCodes ? params.list("anyActivityCodes") as ActivityCode[]: []
-        activityCodeParams.ignoreActivityCodes = params.ignoreActivityCodes ? params.list("ignoreActivityCodes") as ActivityCode[] : []
+    def list(InternalLocationSearchCommand command) {
```
EWaterman, line 26: I didn't originally implement this with a command but `locationService.getInternalLocations` was so annoying to work with when activityCodeParams was a Map because I needed optional params and the compiler was complaining that it couldn't distinguish between methods so I got fed up enough to refactor it.

grails-app/utils/org/pih/warehouse/sort/SortParam.groovy
```diff
@@ -0,0 +1,22 @@
+package org.pih.warehouse.sort
+
+/**
+ * A POJO for holding a sorting condition on an entity.
+ */
+class SortParam {
```
awalkowiak, line 6: Why POJOs instead `Command` objects?

grails-app/utils/org/pih/warehouse/sort/SortParamList.groovy
```diff
@@ -0,0 +1,29 @@
+package org.pih.warehouse.sort
+
+/**
+ * A POJO for holding a list of sorting conditions on an entity.
+ */
+class SortParamList {
```
awalkowiak, line 6: Same quesition here with the command as above

grails-app/utils/org/pih/warehouse/sort/SortUtil.groovy
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Parses a string matching our sort param pattern into a list of sort params. For use by APIs that accept
+     * sort criteria, such as list endpoints. Typically these APIs will accept a "sort" query param like: ?sort=x,-y
+     *
+     * For example: given a string "x,-y", this signifies the entity should first be sorted ascending by "x",
+     * then descending by "y".
+     *
+     * @param sortString A string of the form "x,-y,z"
+     */
+    static SortParamList bindSortParams(String sortString) {
```
awalkowiak, line 22: Could this be achieved by auto-binding with a command? Do we need to add a util for it?

grails-app/utils/org/pih/warehouse/sort/SortUtil.groovy
```diff
@@ -0,0 +1,102 @@
...
+
+            sortParams.add(new SortParam(fieldName, ascending))
+        }
+        return sortParams
+    }
+
+    /**
+     * Sorts a given list of entities using the provided sort parameters.
+     */
+    static <T> List<T> sort(List<T> toSort, SortParamList sortParams) {
```
awalkowiak, line 46: Is there a case that will utilize it? Feels like this is basically achieving exactly the same what command + criteria builder would do or command + `.sort {  }`

grails-app/utils/org/pih/warehouse/sort/SortUtil.groovy
ewaterman: It actually is auto-binding. SortParamListValueConverter is where the logic is. All it does is call this method so we could probably just move this method there if we don't want to expose the behaviour to the rest of the app

grails-app/utils/org/pih/warehouse/sort/SortParam.groovy
ewaterman: it's used by command objects (See the new ReceivingBinSearchCommand). Idk if that makes it a command itself but probably it's just a semantics thing. If you prefer, I'm fine with naming these SortParamCommand and SortParamListCommand if that is clearer

grails-app/utils/org/pih/warehouse/sort/SortUtil.groovy
ewaterman: it's used in LocationService. In that code, we've already fetched from the db and so have a List<Location> that we need to sort. So yes we're essentially doing the equivalient of List.sort(), it's just for convenience so that we can simply provide a SortParamList to sort by (which is provided by the user via the `?sort=x,y,z` query param.

The big thing is that it allows us to dynamically sort by a user-given list of params, instead of a fixed set of params defined in the command object that List.sort() would do.

We could also make a new util method that would take in a SortParamList and output criteria that we can use to query the db with.

src/main/groovy/org/pih/warehouse/core/ReceivingBinSearchCommand.groovy
```diff
@@ -0,0 +1,17 @@
+package org.pih.warehouse.core
+
+/**
+ * Used to filter the returned results when listing/searching internal receiving bin locations.
+ */
+class ReceivingBinSearchCommand extends InternalLocationSearchCommand {
```
jmiranda: I know the term "receiving bin" has been used a lot throughout the system and maybe in requirements, but I would love to get rid of the notion of a "receiving bin" once and for all. It's more appropriate to call these locations "receiving locations". It's usually not a bin in the typical sense of a "storage bin" or "bin location". It's a staging location on the floor of the warehouse near the dock door.

jmiranda: Location 02 is the receiving area.
image

PR #5289, opened 2025-05-29
OBPIH-7146 Display correct values after file import

SebastianLib:

**Description:**
The problem was that when we fetched data from the import, it didn’t display correctly because the columns weren’t re-rendering due to the use of useMemo. By adding await trigger() in fetchLineItems, the columns now re-render properly, and we can see the correct data.

---


alannadolny (commented):
Can you add a description to the PR explaining what's going on here?

PR #5291, opened 2025-05-29
OBPIH-7254 Refactor transaction during Load demo data

alannadolny:


awalkowiak (approved):

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -200,52 +209,120 @@ class LoadDataService {
         csvReader.close();
     }
 
+    private Transaction createAdjustmentTransaction(Date transactionDate, Inventory inventory) {
+        return new Transaction(
+                transactionDate: transactionDate,
+                transactionType: TransactionType.get(Constants.ADJUSTMENT_CREDIT_TRANSACTION_TYPE_ID),
+                inventory: inventory
+        )
```
EWaterman, line 217: how is this working without setting a transactionNumber?

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -200,52 +209,120 @@ class LoadDataService {
...
+    }
+
+    private TransactionEntry createAdjustmentTransactionEntry(
+            Map<String, String> csvReaderRow,
+            Product product,
+            Location binLocation,
+            Map<String, AvailableItem> availableItems
+    ) {
+        DateFormat dateFormat = new SimpleDateFormat('DD/mm/yyyy')
```
EWaterman, line 226: Shouldn't it be `'dd/MM/yyyy'`?

"D" is day in year and "m" is month in hour: https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html

Actually, why isn't it our normal expiration date format? From Constants.groovy:
```
EXPIRATION_DATE_FORMAT = "MM/dd/yyyy"
```

I know this isn't new code but I don't understand how it was ever working correctly...

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
```diff
@@ -200,52 +209,120 @@ class LoadDataService {
...
-        };
+            if (transactionEntry) {
+                adjustmentTransaction.addToTransactionEntries(transactionEntry)
+            }
+        }
 
-        transaction.forceRefresh = Boolean.TRUE
-        transaction.save(flush: true, failOnError: true)
+        adjustmentTransaction.forceRefresh = Boolean.TRUE
+        adjustmentTransaction.save(flush: true, failOnError: true)
```
EWaterman, line 323: We briefly discussed this in my PR, but I switched the logic in inventory import to be:
```
if (!adjustmentTransaction.save()) {
    throw new ValidationException("Invalid transaction", adjustmentTransaction.errors)
}
```
Any reason we can't do the same here? Why is the flush and force refresh need?

EWaterman: Also, you'll want to only create the transaction if adjustmentTransaction.transactionEntries == true. There's no need to create the transaction if it'd have no entries

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
EWaterman: yeah it'd be good to take a look while you're already modifying this code. If it's importing expiry dates totally wrong, it'd probably be an easy fix

grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
alannadolny: I checked the transactionNumber and it seems to be autogenerated :thinking: 
image

EWaterman (approved):

PR #5299, opened 2025-06-02
OB-3230: Added receiving, packing, loading locations to be returned f…

druchniewicz:


grails-app/domain/org/pih/warehouse/inventory/OutboundStockMovementListItem.groovy
```diff
@@ -179,10 +179,9 @@ class OutboundStockMovementListItem implements Serializable, Validateable {
                 expectedShippingDate : shipment?.expectedShippingDate?.format("MM/dd/yyyy HH:mm XXX"),
                 requestedDeliveryDate: requisition?.requestedDeliveryDate?.format("MM/dd/yyyy HH:mm XXX"),
                 picklist             : requisition?.picklist,
-                // Needs to be added to OutboundStockMovementListItem (or moved to the StockMovement :thinking:)
-                //receivingLocation: receivingLocation?.toJson(LocationTypeCode.INTERNAL),
-                //packingLocation  : packingLocation?.toJson(LocationTypeCode.INTERNAL),
-                //loadingLocation  : loadingLocation?.toJson(LocationTypeCode.INTERNAL),
+                receivingLocation: shipment?.receivingScheduled?.eventLocation?.name ?: "Unassigned",
```
awalkowiak: I think it would be better to do `shipment?.receivingScheduled?.eventLocation?.toJson(LocationTypeCode.INTERNAL)`, so we have a few other information, not only the name. (same request on `packingLocation` and `loadingLocation`)

awalkowiak: Other point is, if this should be added to `OutboundStockMovement` as well? Or it's required only here on the "`ListItem`"

druchniewicz: hmm, at this moment seems that is not needed as the list of `OutboundStockMovementListItem` is being fetched and used

PR #5302, opened 2025-06-03
OBPIH-7258 Fix display root causes and comments + improve disabling "Import Count" button

SebastianLib:
…le import button

**Description:**

---

### DIFF
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -1729,3 +1729,7 @@ div.search-input:focus-within {
   cursor: not-allowed;
   pointer-events: none;
 }
+
+.no-transition {
+  transition: none;
+}
--- a/src/js/components/form-elements/v2/ButtonFileSelect.jsx
+++ b/src/js/components/form-elements/v2/ButtonFileSelect.jsx
@@ -15,7 +15,7 @@ const ButtonFileSelect = ({
   defaultLabel,
   variant,
   disabled,
-  removeBtnClassName,
+  className,
 }) => {
   const translate = useTranslate();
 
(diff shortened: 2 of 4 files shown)


src/js/components/form-elements/v2/ButtonFileSelect.jsx
```diff
@@ -44,7 +45,7 @@ const ButtonFileSelect = ({
   return (
     <label
       htmlFor="csvInput"
-      className={`btn ${variant}-button ${disabled ? `${variant}-disabled` : ''} align-self-end btn-xs`}
+      className={`${!removeBtnClassName && 'btn'} ${variant}-button ${disabled ? `${variant}-disabled` : ''} align-self-end btn-xs`}
```
SebastianLib: By removing this bootstrap classname, when we're on the count workflow and change the location, this button turns grey at the same time as the other buttons. Without it, this button would become grey with a delay

src/css/main.scss
```diff
@@ -1729,3 +1729,7 @@ div.search-input:focus-within {
   cursor: not-allowed;
   pointer-events: none;
 }
+
+.no-transition {
+  transition: none;
```
kchelstowski, line 1734: have you checked if such class is not built it in the bootstrap we are using?

SebastianLib: yes, I checked and there is no class like "no-transition" in bootstrap. You can check here: https://www.w3schools.com/bootstrap4/bootstrap_ref_all_classes.asp

kchelstowski (approved):

PR #5303, opened 2025-06-03
OBPIH-7337 export recount enhancements

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7337

**Description:** Modified CycleCountService.getRecountFormXls to also export the value of “Quantity Recounted“, "Root Cause", “Comment“, “User Recounted“, and “Date Recounted“.

---

Screenshot from 2025-06-03 09-59-13

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
@@ -3,6 +3,8 @@ package org.pih.warehouse.inventory
 import grails.gorm.PagedResultList
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
+import org.apache.commons.collections4.keyvalue.MultiKey
+import org.apache.commons.collections4.map.MultiKeyMap
 import org.apache.commons.csv.CSVPrinter
 import org.apache.commons.lang.StringEscapeUtils
 import org.grails.datastore.mapping.query.api.Criteria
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -301,29 +307,82 @@ class CycleCountService {
...
+
+    /**
+     * Converts a CycleCountDto to a list of rows/maps. For use when exporting to XLS.
+     */
+    private List<Map> getRecountAsXlsMap(CycleCountDto cycleCount) {
+        int currentCountIndex = cycleCount.maxCountIndex
+
+        // Build a map to make it easier to group the count and recount items into one row in the XLS.
+        // The outer map is keyed on [product code + lot + bin]. The inner map is keyed on count index.
+        MultiKeyMap<String, Map<Integer, CycleCountItemDto>> countItemsMap = [:]
```
ewaterman, line 324: I quite like MultiKeyMap and MultiKey. I find them very useful as a way to avoid needing to make nested maps (though in this rare case I *also* have a nested map for convenience).

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/MultiKeyMap.html

kchelstowski: nitpicky, but I think an interface should be the type, so in this case it should probably be:
```groovy
IterableMap<String, Map<Integer, CycleCountItemDto>> countItemsMap = new MultiKeyMap()
```

in the same way that we declare a list - we don't do:
```groovy
ArrayList<String> = new ArrayList<>()
```
but
```groovy
List<String> = new ArrayList<>()
```

PR #5308, opened 2025-06-04
OBPIH-7288: add bulk update cycle count request API

EWaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7288

**Description:** As a part of the assign count feature, we need to be able to update a cycle count request so that we can set the assignee and count date.

`PATCH /api/facilities/$facilityId/cycle-counts/requests/batch`
```
{
    "commands": [
        {
            "cycleCountRequest": "cab2b1059663f58b0196645b37630003",
            "countAssignee": "cab2b1078d3fe94c018e7b7c69c109f3",
            "countDeadline": "2025-06-01",
            "recountAssignee": "cab2b1078d3fe94c018e7b7c69c109f3",
            "recountDeadline": "2025-06-02"
        }
    ]
}
```


grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -21,6 +24,28 @@ class CycleCountRequest {
...
+    /**
+     * The person who is responsible for performing the recount. Note that this might be different from
+     * the person who *actually* performs the recount.
+     */
+    Person recountAssignedTo
+
+    /**
+     * The date that the recount should be performed by.
+     */
+    LocalDate recountDeadline
```
ewaterman, line 47: I know these are different from what was in the ticket so we can discuss column names if you want

EWaterman:
I haven't tested this yet since it's just the API. I'll try to find time soon to set it up in postman to test it out

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -21,6 +24,28 @@ class CycleCountRequest {
 
     Boolean blindCount
 
+    /**
+     * The person who is responsible for performing the count. Note that this might be different from
+     * the person who *actually* performs the count.
+     */
+    Person countAssignedTo
```
kchelstowski: wouldn't it sound more convenient if we make it `countAssignee`, in the same way we have the `assignee`? unless the context is much different?

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestUpdateCommand.groovy
```diff
@@ -0,0 +1,25 @@
...
+
+import grails.databinding.BindUsing
+import grails.validation.Validateable
+import java.time.LocalDate
+
+import org.pih.warehouse.core.Person
+
+class CycleCountRequestUpdateCommand implements Validateable {
+
+    @BindUsing({ obj, source -> CycleCountRequest.get(source['id'] as Serializable) })
```
kchelstowski: we can get rid of it, and just pass the id in the payload via `cycleCountRequest` key, so it would get bound automatically.

src/main/groovy/org/pih/warehouse/core/dtos/BulkCommand.groovy
```diff
@@ -0,0 +1,33 @@
...
+    static constraints = {
+        commands(validator: { Collection<T> commands ->
+            // The individual commands are not automatically validated so we have to do it manually.
+            commands.each { it.validate() }
+            if (commands.any { it.hasErrors() }) {
+                return false
+            }
+        })
+    }
+}
```
kchelstowski: nice idea, but let's remember this could not unfortunately replace all the `XBatchCommand` we have due to the fact that some of the validators are customized (e.g. `CycleCountRequestBatchCommand` and the statement with duplicates)

grails-app/migrations/0.9.x/changelog-2025-06-05-0000-add-assignment-columns-to-cycle-count-request.xml
```diff
@@ -0,0 +1,82 @@
...
+            </column>
+        </addColumn>
+        <addForeignKeyConstraint
+                baseColumnNames="count_assigned_to"
+                baseTableName="cycle_count_request"
+                constraintName="fk_cycle_count_request_count_assigned_to"
+                deferrable="false"
+                initiallyDeferred="false"
+                referencedColumnNames="id"
+                referencedTableName="user"
```
kchelstowski: shouldn't it be `person` table?

grails-app/migrations/0.9.x/changelog-2025-06-05-0000-add-assignment-columns-to-cycle-count-request.xml
```diff
@@ -0,0 +1,82 @@
...
+            </column>
+        </addColumn>
+        <addForeignKeyConstraint
+                baseColumnNames="recount_assigned_to"
+                baseTableName="cycle_count_request"
+                constraintName="fk_cycle_count_request_recount_assigned_to"
+                deferrable="false"
+                initiallyDeferred="false"
+                referencedColumnNames="id"
+                referencedTableName="user"
```
kchelstowski: same as above

src/main/groovy/org/pih/warehouse/core/dtos/BulkCommand.groovy
ewaterman: good point. If any bulk/batch command needs custom validation, it should implement it itself in its own class. Maybe it'd still be able to extend this class though.

Or at least import the validation from it:
```
class CustomBulkCommand {
    ...

    static constraints = {
        importFrom BulkCommand
        ...
    }
}
```

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
ewaterman: I'm fine either way. I'll make it countAssignee so that it's consistent with assignee like you said.

src/main/groovy/org/pih/warehouse/core/dtos/BulkCommand.groovy
kchelstowski: I think actually the `importFrom` wouldn't help here, because those bulk commands would usually have only one property (`commands`) that would be validateable, but I'm fine with just using this class whenever it's possible, and if not, just create a custom one :+1:

src/main/groovy/org/pih/warehouse/inventory/CycleCountRequestUpdateCommand.groovy
ewaterman: it'd work like this?
```
{
  'cycleCountRequest': 123,
}
```

I was worried doing it that way would require the payload to be:
```
{
  'cycleCountRequest': { 'id': 123 }
}
```

but maybe I was misremembering how binding works

EDIT: Yeah it'd work as Kacper suggested.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -3,16 +3,20 @@ package org.pih.warehouse.inventory
...
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
+import java.time.LocalDate
 import org.apache.commons.csv.CSVPrinter
 import org.apache.commons.lang.StringEscapeUtils
 import org.grails.datastore.mapping.query.api.Criteria
 import org.hibernate.ObjectNotFoundException
 import org.hibernate.criterion.Order
 import org.hibernate.sql.JoinType
+import org.springframework.validation.BindingResult
```
kchelstowski: seems not to be used anymore (6th and 19th line also)

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -260,6 +261,31 @@ class CycleCountService {
...
+        }
+        return updatedRequests
+    }
+
+    private CycleCountRequest updateRequest(CycleCountRequestUpdateCommand command) {
+        CycleCountRequest cycleCountRequest = command.cycleCountRequest
+        cycleCountRequest.countAssignee = command.countAssignee
+        cycleCountRequest.countDeadline = command.countDeadline
+        cycleCountRequest.recountAssignee = command.recountAssignee
+        cycleCountRequest.recountDeadline = command.recountDeadline
```
alannadolny, line 281: What if we want to:
1. Update fields related to the initial count - before counting
2. Update fields related to verification count - before recounting
Will the fields for the initial count be null after that? I think we shouldn't clear that while assigning recount data.

ewaterman: you're right. It should only update if the value in the command is not null

ewaterman: Actually, is there a way to unassign someone in the widget? If I update it to:
```
cycleCountRequest.countAssignee = command.countAssignee ?: cycleCountRequest.countAssignee
cycleCountRequest.countDeadline = command.countDeadline ?: cycleCountRequest.countDeadline
cycleCountRequest.recountAssignee = command.recountAssignee ?: cycleCountRequest.recountAssignee
cycleCountRequest.recountDeadline = command.recountDeadline ?: cycleCountRequest.recountDeadline
```
then there is no longer a way to set the assignee or deadline back to null.

If that's not a problem, then the above is fine. Otherwise we could do it similarly to how you had in your PR with something like:
```
class CycleCountRequestUpdateCommand {
    Map<String, CycleCountAssignmentCommand> assignments

    CycleCountAssignmentCommand getAssignmentByCountIndex(String countIndex) {
        return assignments.get(countIndex)
    }
}

class CycleCountAssignmentCommand {
    Person assignee
    LocalDate deadline
}
```
and then in the service do:
```
private CycleCountRequest updateRequest(CycleCountRequestUpdateCommand command) {
    CycleCountRequest cycleCountRequest = command.cycleCountRequest
    CycleCountAssignmentCommand countAssignment = command.getAssignmentByCountIndex(0)
    if (countAssignment) {
        cycleCountRequest.countAssignee = countAssignment.assignee
        cycleCountRequest.countDeadline = countAssignment.deadline
    }
    CycleCountAssignmentCommand recountAssignment = command.getAssignmentByCountIndex(1)
    if (recountAssignment) {
        cycleCountRequest.recountAssignee = recountAssignment.assignee
        cycleCountRequest.recountDeadline = recountAssignment.deadline
    }
    ...
}
```
so the request payload would look more like:
```
{
    'assignments': {
        '0': {
            'assignee': 'Evan',
            'deadline': '2025-01-01'
        },
        '1': {
            'assignee': 'Alan',
            'deadline': '2025-01-06'
        }
    }
}
```
that way the API is friendlier to supporting multiple count indexes (even though the db table still stores them flat)

@alannadolny I'd probably do it the second way, but since I'm away, I leave it up to you.

alannadolny: @EWaterman There is an option to unassign a person, but I think we need a confirmation from Manon about one of the possible behaviours. On the "to resolve" tab, we should see cycle counts in status "in progress" and "to resolve" and I am not sure whether we should be able to unassign someone from that cycle count when it is pending. I think we can get back to that topic when she is back. At this moment, I am thinking about blocking the unassigning option when the counting/recounting is started.

PR #5309, opened 2025-06-05
OBPIH-7046 Prevent selecting to count when in pending outbound

SebastianLib:

**Description:**

---


src/js/components/DataTable/v2/DataTableBody.jsx
```diff
@@ -44,11 +43,10 @@ const DataTableBody = ({
       />
       {dataLength > 0 &&
         !loading &&
-        rowModel.rows.map((row, index) => {
+        rowModel.rows.map((row) => {
           const { isProductDisabled, label, defaultMessage } = showDisabledProducts
-            ? useCycleCountProductAvailability(data[index].status)
+            ? useCycleCountProductAvailability(row.original)
```
kchelstowski: could you explain the reason for that change? Maybe it was intentionally done like this before? 
By the way - it feels weird to me that those table components are probably supposed to be reusable for other features and inside it, we have logic connected to cycle count (`useCycleCountProductAvailability` hook).
@alannadolny do you know more context?

SebastianLib: I changed it because useCycleCountProductAvailability has been extended. It now checks not only by status but also by quantityAllocated. I think it might actually be a better idea to remove this logic from DataTableBody and try to pass the data via meta or props instead.

src/js/hooks/cycleCount/useAllProductsTab.jsx
```diff
@@ -137,8 +137,22 @@ const useAllProductsTab = ({
...
+    }
+    return {
+      ...tableData,
+      data: tableData.data.map((row) => ({
+        ...row,
+        // Added meta to rows because TanStack Table doesn't support meta at the row level
+        meta: useCycleCountProductAvailability(row),
+      })),
+    };
+  }, [tableData]);
```
SebastianLib, line 152: I added meta to rows so that useCycleCountProductAvailability is called only once per row instead of multiple times per cell. I think this is a better solution

src/js/components/DataTable/v2/DataTableBody.jsx
```diff
@@ -44,18 +41,19 @@ const DataTableBody = ({
...
       {dataLength > 0 &&
         !loading &&
-        rowModel.rows.map((row, index) => {
-          const { isProductDisabled, label, defaultMessage } = showDisabledProducts
-            ? useCycleCountProductAvailability(data[index].status)
-            : { isProductDisabled: false, label: '', defaultMessage: '' };
-
+        rowModel.rows.map((row) => {
+          const { isProductDisabled, label, defaultMessage } = row.original?.meta || {
+            isProductDisabled: false,
```
awalkowiak: `isRowDisabled`

awalkowiak (changes requested):

awalkowiak (approved):

PR #5313, opened 2025-06-06
OBPIH-7271 Handle invalid count import

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
@@ -1,0 +1,73 @@
+package org.pih.warehouse.inventory
+
+import grails.util.Holders
+import org.grails.plugins.web.taglib.ApplicationTagLib
+import org.pih.warehouse.core.LocalizationService
+import org.pih.warehouse.importer.ImportDataCommand
+
+class CycleCountImportService {
+
+    LocalizationService localizationService
+
+    void validateCountImport(ImportDataCommand command) {
+        // Firstly clear the command's errors, as it might contain some additional errors related to the file itself
+        // that we do not care about while validating the data, and we don't them to be visible for a user
+        command.clearErrors()
+        // Store items that should be ignored for an import - for some cases we want not only to display an error, but also to ignore such row
+        List<Map> itemsToRemove = []
+        command.data.eachWithIndex { row, index ->
+            CycleCountRequest cycleCountRequest = CycleCount.read(row.cycleCountId)?.cycleCountRequest
+            // If a user provides an invalid cycle count id, skip that row and do not look for other errors, as they do not matter at this point
+            if (!cycleCountRequest) {
+                command.errors.reject("Row ${index + 1}: This row is skipped. Cycle count with given cycle count id ${row.cycleCountId} does not exist")
+                itemsToRemove.add(row)
+                // as explained above - since a cycle count is not found, there is no point to validate the row itself
+                return
+            }
+            CycleCountItem cycleCountItem = CycleCountItem.read(row.cycleCountItemId)
+            // If a cycle count item is not found, but user provided an id, do not treat such row as a custom row, but throw an error and ignore the row.
+            if (!cycleCountItem && row.cycleCountItemId) {
+                command.errors.reject("Row ${index + 1}: This row is skipped. Cycle count item with given cycle count item id ${row.cycleCountItemId} does not exist")
+                itemsToRemove.add(row)
+            }
+            // Ignore the row if product code is blank
+            if (!row.productCode) {
+                command.errors.reject("Row ${index + 1}: This row was ignored in the import because product code was blank - you must use a product associated with this cycle count")
+                itemsToRemove.add(row)
+            }
+            // Ignore the row is product code is different than the product code that cycle count is associated with
+            if (row.productCode != cycleCountRequest.product.productCode) {
+                command.errors.reject("Row ${index + 1}: This row is skipped, because you used an invalid product - you must use a product associated with this cycle count")
+                itemsToRemove.add(row)
+            }
+            // If bin location is not null, but an object with id and name properties, it means it is not a default bin
+            // but user provided an invalid location id
+            if (row.binLocation != null && !row.binLocation.id) {
+                command.errors.reject("Row ${index + 1}: Provided bin location does not exist in the system")
+            }
+            // Validation cases for an existing item (update)
+            // We mustn't update bin location, lot number and expiration date via import for an existing item
+            if (cycleCountItem) {
+                if (row.binLocation?.id != cycleCountItem.location?.id) {
+                    command.errors.reject("Row ${index + 1}: You cannot update bin location for an existing item")
+                }
+                if (row.lotNumber != cycleCountItem.inventoryItem?.lotNumber) {
+                    command.errors.reject("Row ${index + 1}: You cannot update lot number for an existing item")
+                }
+                if (row.expirationDate != cycleCountItem.inventoryItem.expirationDate) {
+                    command.errors.reject("Row ${index + 1}: You cannot update expiration date for an existing item")
+                }
+            }
+
+        }
+        // In the end remove from the list all rows that are supposed to be ignored
+        command.data.removeAll(itemsToRemove)
+    }
+
+    List<String> buildErrors(ImportDataCommand command) {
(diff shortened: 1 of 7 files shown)


src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
```diff
@@ -71,10 +74,10 @@ class CycleCountItemsExcelImporter extends AbstractExcelImporter implements Data
 
         // 3. Replace data from importer with the fetched data
         data.each { row ->
-            row.binLocation = [
+            row.binLocation = (locationMap[row.binLocation] && row.binLocation) ? [
```
ewaterman, line 77: nitpick: I don't know if it actually matters, but I think it's a better practice to flip the condition to: `row.binLocation && locationMap[row.binLocation]` since if `row.binLocation` is nulll/falsey, it won't attempt to check `locationMap[null]`

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -0,0 +1,73 @@
...
+import org.pih.warehouse.importer.ImportDataCommand
+
+class CycleCountImportService {
+
+    LocalizationService localizationService
+
+    void validateCountImport(ImportDataCommand command) {
+        // Firstly clear the command's errors, as it might contain some additional errors related to the file itself
+        // that we do not care about while validating the data, and we don't them to be visible for a user
+        command.clearErrors()
```
ewaterman, line 15: what file errors were there? Do we really not care about them?

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -0,0 +1,73 @@
...
+            if (!cycleCountRequest) {
+                command.errors.reject("Row ${index + 1}: This row is skipped. Cycle count with given cycle count id ${row.cycleCountId} does not exist")
+                itemsToRemove.add(row)
+                // as explained above - since a cycle count is not found, there is no point to validate the row itself
+                return
+            }
+            CycleCountItem cycleCountItem = CycleCountItem.read(row.cycleCountItemId)
+            // If a cycle count item is not found, but user provided an id, do not treat such row as a custom row, but throw an error and ignore the row.
+            if (!cycleCountItem && row.cycleCountItemId) {
+                command.errors.reject("Row ${index + 1}: This row is skipped. Cycle count item with given cycle count item id ${row.cycleCountItemId} does not exist")
```
ewaterman, line 30: would it be index + 2 because xls rows start at 1 (not 0) and we have the header row?

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -0,0 +1,73 @@
...
+
+        }
+        // In the end remove from the list all rows that are supposed to be ignored
+        command.data.removeAll(itemsToRemove)
+    }
+
+    List<String> buildErrors(ImportDataCommand command) {
+        ApplicationTagLib g = Holders.grailsApplication.mainContext.getBean(ApplicationTagLib)
+        return command.errors.allErrors.collect {
+            g.message(error: it, locale: localizationService.currentLocale)
```
ewaterman, line 70: Should we be localizing these errors? I know we normally don't, but since this is an import it feels like the errors would be more commonly encountered by users.

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -0,0 +1,73 @@
...
+            // We mustn't update bin location, lot number and expiration date via import for an existing item
+            if (cycleCountItem) {
+                if (row.binLocation?.id != cycleCountItem.location?.id) {
+                    command.errors.reject("Row ${index + 1}: You cannot update bin location for an existing item")
+                }
+                if (row.lotNumber != cycleCountItem.inventoryItem?.lotNumber) {
+                    command.errors.reject("Row ${index + 1}: You cannot update lot number for an existing item")
+                }
+                if (row.expirationDate != cycleCountItem.inventoryItem.expirationDate) {
+                    command.errors.reject("Row ${index + 1}: You cannot update expiration date for an existing item")
```
ewaterman, line 58: when saving a count item normally (not via import), we only use expiration date when creating a new inventory item that doesn't already exist, otherwise the expiration date that the user provides is ignored. Can we do the same here instead of failing?

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -0,0 +1,73 @@
...
+                command.errors.reject("Row ${index + 1}: This row is skipped, because you used an invalid product - you must use a product associated with this cycle count")
+                itemsToRemove.add(row)
+            }
+            // If bin location is not null, but an object with id and name properties, it means it is not a default bin
+            // but user provided an invalid location id
+            if (row.binLocation != null && !row.binLocation.id) {
+                command.errors.reject("Row ${index + 1}: Provided bin location does not exist in the system")
+            }
+            // Validation cases for an existing item (update)
+            // We mustn't update bin location, lot number and expiration date via import for an existing item
```
ewaterman, line 49: it's a pretty rigid solution. I bet there will be users who will try to fix stock being in the incorrect bin or having the wrong lot by simply changing the bin or lot number for a row (that's how inventory import works). Since we have CCI id in the row, it does make sense that we need to fail here, it's just unfortunate because I know users will be trying to do this

src/main/groovy/org/pih/warehouse/importer/CycleCountItemsExcelImporter.groovy
kchelstowski: actually the case here is that we might have a default bin location (null), so I want to check explicitly two things together:
1. if a user provided the bin location
2. if a user provided the bin location, check if it was found

Further representation of binLocation is:
1. if a user uses the default bin: `null`
2. if a user provides an invalid (not existing) bin location:
```
{
  id: null,
  name: null
}
```
by doing so I can further (in the validator) know, whether it's `null`, because we use the default bin, or if the given bin location doesn't exist.

P.S. I should've probably commented that out

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
kchelstowski: yeah, we don't, those are e.g. `importType` which is handled on the frontend and we don't need to validate it on the  backend, or `location` which we also don't care about here.

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
kchelstowski: you are right, but in other imports we also do `index + 1`, so I assumed it's intentional and users know which row we are referring to (even though it's `-1`)

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
kchelstowski: I know we've had many discussions around this column, that it probably even doesn't make sense, but I wanted to be explicit here and also fail on the expiration date, so that it is more readable in the future, especially if we add an ability to edit an existing item via UI.
Let me know if you are fine with it

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
kchelstowski: I think if it passes the QA, we might think about localizing them. We would probably have to add a code argument to the `.reject` in the validator. I wanted to get this to QA asap, so I skipped it for now.

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -299,7 +299,7 @@ class CycleCountService {
         cycleCounts.each { CycleCountDto cycleCount ->
             cycleCount.cycleCountItems.each { CycleCountItemDto item ->
                 data << [
-                        "Cycle count id": cycleCount.id,
+                        "Product cycle count id": cycleCount.id,
```
kchelstowski, line 302: it was Manon's request to make this column "Product cycle count id" so that it is more understandable by the users.

awalkowiak (approved):
LGTM, a few minor things will be revisited later (row indexes, l18n)

PR #5314, opened 2025-06-06
OBPIH-7311 Improve Transaction report after Product Inventory transactions refactor

alannadolny:
Diagram explaining calculations on a timeline:
image

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
@@ -637,7 +637,12 @@ class InventorySnapshotService {
             Integer openingBalance = calculateBalance(value, closingBalance)
             Integer credits = getCreditTransactionEntries(value).sum { it.quantity } as Integer ?: 0
             Integer debits = getDebitTransactionEntries(value).sum { Math.abs(it.quantity) } as Integer ?: 0
-            Integer adjustments = closingBalance - openingBalance - credits + debits
+            // In the new version of the report, it's not based on the inventory snapshot.
+            // So we don't have to calculate it in the following way:
+            // closingBalance - openingBalance - credits + debits,
+            // because the data is accurate, so we can just compare
+            // the closing and opening balance
+            Integer adjustments = closingBalance - openingBalance
 
             return [:].with {
                 it["Code"] = key.productCode


grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
+                        [(transactionTypeName): total]
+                    }
+
+            [(product): totalsByType]
         }
-        return data.sort { it."Code" }
     }
 
-    def getTransactionReportSummary(Location location, List<Category> categories, List<Tag> tags, List<ProductCatalog> catalogs, Date startDate, Date endDate) {
+    List<Object> getTransactionReport(Location location, List<Category> categories, List<Tag> tagsList, List<ProductCatalog> catalogsList, Date startDate, Date endDate, Boolean includeDetails) {
```
ewaterman, line 566: it's a bit hard to tell with all the changes, but this flow is all newly added by you, right? If so, I think this would be better off in a different service since it doesn't actually have anything to do with inventory snapshots anymore and we're starting to deprecate/remove all inventory snapshot stuff. You could even put it in a new TransactionReportService.

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
+        List<TransactionEntry> debits = getDebitTransactionEntries(transactionEntries)
+        Integer quantityFromCredits = credits.sum { Math.abs(it.quantity) } as Integer ?: 0
+        Integer quantityFromDebits = debits.sum { Math.abs(it.quantity) } as Integer ?: 0
 
-        def transactionData = getTransactionReportData(location, startDate, endDate)
-
-        def transactionTypes = TransactionType.createCriteria().list {
-            'in'("transactionCode", [TransactionCode.DEBIT, TransactionCode.CREDIT])
-        }
+        return balance - quantityFromCredits + quantityFromDebits
```
ewaterman, line 473: I wonder if it's possible to have a DEBIT transaction with a negative quantity (which would turn into a credit). Probably not, but maybe good to handle it just in case.

I'll make a separate comment about this, but what about product inventory transactions?

Anyways, you could do:
```
Integer balance = 0
for (TransactionEntry entry in transactionEntries) {
    switch(entry.transaction.transactionType.transactionCode) {
        case TransactionCode.CREDIT:
            balance += entry.quantity
        case TransactionCode.DEBIT:
            balance -= entry.quantity
        case TransactionCode.PRODUCT_INVENTORY:
            balance = entry.quantity
    }
}
return balance
```

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
+            Integer initialQuantity = initialQuantityForBalanceCalculations[key] ?: 0
+            Integer closingBalance = entriesAfterEndDate.size()
+                    ? calculateBalance(entriesAfterEndDate, initialQuantity)
+                    : initialQuantity
+            Integer openingBalance = calculateBalance(value, closingBalance)
+            Integer credits = getCreditTransactionEntries(value).sum { it.quantity } as Integer ?: 0
+            Integer debits = getDebitTransactionEntries(value).sum { Math.abs(it.quantity) } as Integer ?: 0
+            Integer adjustments = closingBalance - openingBalance - credits + debits
+
+            return [:].with {
```
ewaterman, line 647: since this is a new map, why not just:
```
return [
    'code': key.productCode,
    ...
]
```
EDIT: I guess this is so you can have your complex conditionals and init the map at the same time. Kinda neat.

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
         }
-        return data.sort { it."Code" }
     }
 
-    def getTransactionReportSummary(Location location, List<Category> categories, List<Tag> tags, List<ProductCatalog> catalogs, Date startDate, Date endDate) {
+    List<Object> getTransactionReport(Location location, List<Category> categories, List<Tag> tagsList, List<ProductCatalog> catalogsList, Date startDate, Date endDate, Boolean includeDetails) {
+        List<TransactionCode> adjustmentTransactionCodes = [
+                TransactionCode.CREDIT,
+                TransactionCode.DEBIT
+        ]
```
ewaterman, line 570: What about PRODUCT_INVENTORY transactions? Baseline transactions won't have any adjustment initially, but if we have any backdated transactions, they could. For example (the table is ordered by transactionDate):
```
dateCreated | transactionDate | transactionType   | quantity
Jan 1       | Jan 1           | CREDIT            | 10
Jan 4       | Jan 2           | CREDIT            | 20 -> Was backdated
Jan 3       | Jan 3           | PRODUCT_INVENTORY | 10
```
so in that scenario, we can't simply sum up the adjustment transactions because the product inventory now has an implied adjustment of -20. The actual QoH is 10, but if we simply sum up the adjustments, we'd get QoH == 30.

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
+        // with appropriate filter applied and with transaction type code that is credit or debit
+        List<TransactionEntry> transactionEntriesWithinDateRange = getFilteredTransactionEntries(
+                adjustmentTransactionCodes,
+                startDate,
+                endDate,
+                categories,
+                tagsList,
+                catalogsList,
+                location,
+                null
```
ewaterman, line 582: why don't these also sort by transaction date?

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
+        // Final calculations of data:
+        // 1. Get the current QoH
+        // 2. Calculate closing balance using transaction entries between endDate <-> today
+        // 3. Calculate opening balance using transaction entries between startDate <-> endDate
+        // Additional calculation info:
+        // 1. CREDITS = transaction entries in relation with transactions that are CREDIT type
+        // and the quantity of that transaction entry is greater than 0
+        // 2. DEBITS = transaction entries in relation with transaction that are DEBIT type
+        // and transaction that are CREDIT type, but with quantity lower than 0
+        return productsMap.collect { key, value ->
```
awalkowiak, line 631: I have a question about this calculation. How the opening and closing balance is (if) affected by `PRODUCT_INVENTORY` transactions here - in other words, is it "backwards" compatible? Did you check out reports in some time in the past where there are no baselines yet (but you have locally the old `PRODUCT_INVENTORY` transactions in that time frame)? Can you share two CSVs or table screenshots of the report for the same date range before and after changes?

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
-        // We need all products that have either an opening balance or closing balance
-        def products = new HashSet()
-        products.addAll(balanceOpeningMap.keySet())
-        products.addAll(balanceClosingMap.keySet())
-
-        def data = []
-        products.findAll { (categories.contains(it.category) && ((tags && it.hasOneOfTags(tags)) || (catalogs && it.hasOneOfCatalogs(catalogs)))) ||
-                (!tags && !catalogs && categories.contains(it.category)) }.each { Product product ->
+                if (startDate) {
+                    gt('transactionDate', startDate)
```
awalkowiak, line 529: Should one of the dates (start or end) be also equal? `gte` for start date or `lte` for end date? Looking at your diagram in the description, you could miss a transaction that is at the exact date as your opening balance date or closing balance date (if both are excluded). How was this calculated before? Or am I overthinking this part? :thinking:

grails-app/services/org/pih/warehouse/inventory/InventorySnapshotService.groovy
```diff
@@ -464,155 +464,207 @@ class InventorySnapshotService {
...
-            }
+    Map<Product, Map<String, Integer>> getDetailedTransactionReportData(Map<Product, List<TransactionEntry>> transactionEntries) {
+        return transactionEntries.collectEntries { product, entriesForProduct ->
+            Map<String, Integer> totalsByType = entriesForProduct
+                    .groupBy { entry ->
+                        entry.transaction.transactionType.name
+                    }
+                    .collectEntries { transactionTypeName, entriesByType ->
+                        Integer total = entriesByType.sum { entry ->
+                            Math.abs(entry.quantity as Integer)
```
awalkowiak, line 557: This will be erroneous for the new credit adjustment that has a negative quantity - these entries should be counted as debits

PR #5317, opened 2025-06-10
OBPIH 7260 Create table for products inventory changes report

SebastianLib:

**Description:**

---


src/js/consts/valueIndicatorVariant.js
```diff
@@ -10,8 +12,8 @@ const valueIndicatorVariant = {
   TRANSACTION: 'TRANSACTION',
 };
 
-export const getCycleCountDifferencesVariant = (number, value) => {
-  if (Number.isNaN(number) || value === null) {
+export const getCycleCountDifferencesVariant = (number) => {
+  if (_.isNaN(number) || _.isNil(number)) {
```
alannadolny, line 16: why is this change needed?

src/js/hooks/cycleCount/useProductsTab.jsx
```diff
@@ -0,0 +1,412 @@
...
+  const emptyTableMessage = !filterParams.startDate && !filterParams.endDate
+    ? {
+      id: 'react.cycleCount.reporting.emptyTable.label',
+      defaultMessage: 'Select a time range from above filters to load the table.',
+    }
+    : {
+      id: 'react.cycleCount.table.noResultFound.label',
+      defaultMessage: 'No result found.',
+    };
+  const exportData = () => {
```
alannadolny:
```suggestion
    };
    
  const exportData = () => {
```

src/js/consts/valueIndicatorVariant.js
SebastianLib: I noticed that checking if the value is null and returning the null variant was only useful in useResolveStepTable. In other cases, the value check wasn't necessary and might even have been misleading. Now in useResolveStepTable, I’m doing something like this: `const variant = value && getCycleCountDifferencesVariant(recountDifference)`;  and it works the same way even after the changes.

PR #5321, opened 2025-06-11
OBPIH-7046 Don't block count for 'In Progress' products (fix)

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useCycleCountProductAvailability.jsx
+++ b/src/js/hooks/cycleCount/useCycleCountProductAvailability.jsx
@@ -1,10 +1,10 @@
 import { ALL_PRODUCTS_TAB, TO_RESOLVE_TAB } from 'consts/cycleCount';
+import cycleCountCandidateStatus from 'consts/cycleCountCandidateStatus';
 import useQueryParams from 'hooks/useQueryParams';
 import { isCounting, isResolving } from 'utils/checkCycleCountStatus';
 
 const useCycleCountProductAvailability = (row) => {
   const { tab } = useQueryParams();
-
   if (isCounting(row.status) && tab === ALL_PRODUCTS_TAB) {
     return {
       isRowDisabled: true,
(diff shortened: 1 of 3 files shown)


alannadolny (approved):

src/js/hooks/cycleCount/useCycleCountProductAvailability.jsx
```diff
@@ -32,7 +32,8 @@ const useCycleCountProductAvailability = (row) => {
     };
   }
 
-  if (row.quantityAllocated > 0) {
+  if (row.quantityAllocated > 0
+    && row?.cycleCountRequest?.status !== cycleCountCandidateStatus.IN_PROGRESS) {
```
ewaterman: Is this the right status? We want to block users from counting or requesting a count on a product that has quantity allocated in two situations:
1) Always when we're on the all products tab
2) When on the to count tab, when a count hasn't been started OR the count has been started but doesn't have any progress saved, ie the row has the "in progress" status, ie the update items API hasn't been called, ie the cycle count candidate status is still REQUESTED.

The statuses are quite confusing but the important thing to remember is:

The "In progress" status in the to count tab == cycleCountCandidateStatus.COUNTING
The "To count" status in the to count tab == cycleCountCandidateStatus.REQUESTED

I think cycleCountRequest.status === IN_PROGRESS when we start the count, regardless of if any data has been saved yet.

So actually we probably want:
```
if (row.quantityAllocated > 0 && tab === TO_COUNT_TAB && row.status === cycleCountCandidateStatus.REQUESTED) {
    return {
      defaultMessage: 'Cannot start count on this product with pending stock movement',
      ...
    };
}

if (row.quantityAllocated > 0 && tab === ALL_PRODUCTS_TAB) {
    return {
      defaultMessage: 'Cannot request count on this product with pending stock movement',
      ...
    };
}
```

SebastianLib: It's true that after starting the count, cycleCountRequest.status becomes IN_PROGRESS, while my row.status hasn't yet been updated to "COUNTING" until we don't save any progress, and this is confusing. Thanks for noticing this!

PR #5322, opened 2025-06-11
OBPIH-7289 Add assignee and deadline data to the table

alannadolny:
This PR should be merged after: https://github.com/openboxes/openboxes/pull/5308

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
@@ -53,6 +53,15 @@ class CycleCountRequest {
                 product: product,
                 status: status.toString(),
                 requestType: requestType.toString(),
+                initialCount: [
+                        deadline: countDeadline,
+                        assignee: countAssignee
+                ],
+                verificationCount: [
+                        deadline: recountDeadline,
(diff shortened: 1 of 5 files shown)


SebastianLib (approved):

grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -53,6 +53,15 @@ class CycleCountRequest {
...
                 status: status.toString(),
                 requestType: requestType.toString(),
+                initialCount: [
+                        deadline: countDeadline,
+                        assignee: countAssignee
+                ],
+                verificationCount: [
+                        deadline: recountDeadline,
+                        assignee: recountAssignee
+                ],
```
ewaterman, line 63: any particular reason to structure it this way vs simply having the fields flat? I'm fine with this but I don't know if there's much to gain over simply doing:
```
return [
    countDeadline: countDeadline,
    countAssignee: countAssignee
    ...
]
```

awalkowiak: Looks like this is the same convention as on the inventroy transaction report

PR #5327, opened 2025-06-12
OB-3248: Added auto transition status from PICKING to PICKED

druchniewicz:


druchniewicz:
@awalkowiak please take a look on it and the best would be if you could finish it and test it

PR #5332, opened 2025-06-13
OBPIH-7195 fix5: merge rows with same product, bin, lot into same transaction entry

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7195

**Description:** This is an alternate solution to the original implementation. I originally was trying to keep it so that if we had multiple rows in the import with the same [product, lot, bin] key, they'd stay as separate transaction entries of the adjustment transaction. But this has led to edge case scenario weirdness which caused multiple re-opens of the ticket.

Kasia's video hilights the current state of things well: https://jam.dev/c/6369f1cb-9376-48b0-8459-3fa50139668b

Because of all these edge cases, I opted for a (hopefully) more straight forward solution where we simply merge all rows with the same key together and create a single adjustment transaction entry from them (if one is required). I don't know why I didn't think of doing it this way from the beginning but :shrug: 

This code is currently untested so it'd be great if someone can pick this up while I'm gone.

### DIFF
--- a/grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
@@ -185,35 +185,32 @@ class InventoryImportDataService implements ImportDataService {
         Transaction transaction = new Transaction()
 
         for (entry in inventoryImportData.rows) {
+            // Merge all rows with the same key (ie the same product + bin + lot) into a single transaction entry.
             List<InventoryImportDataRow> rowsForKey = entry.value
+            if (!rowsForKey) {
+                continue
+            }
 
             // If we have no product availability record for the item, we assume that means we have no quantity.
-            int quantityOnHandToUse = availableItems.get(entry.key)?.quantityOnHand ?: 0
-
-            for (InventoryImportDataRow row in rowsForKey) {
-                int adjustmentQuantity = row.quantity - quantityOnHandToUse
-
-                // Update QoH so that we don't double count it while computing adjustmentQuantity when we have multiple
-                // rows with the same key (which results in rowsForKey having more than one entry). For example, if a
-                // [product + bin + lot] has a QoH of 5, the following rows should create the following adjustments:
-                // - row with the same [product + bin + lot] and quantity == 6  -> should cause a +1 adjustment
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -185,35 +185,32 @@ class InventoryImportDataService implements ImportDataService {
...
-                )
-                transaction.addToTransactionEntries(transactionEntry)
-            }
+            TransactionEntry transactionEntry = new TransactionEntry(
+                    transaction: transaction,
+                    quantity: adjustmentQuantity,
+                    product: row.product,
+                    binLocation: row.binLocation,
+                    inventoryItem: row.inventoryItem,
+                    comments: rowsForKey.comments.join(", "),
```
ewaterman, line 211: Note here that import rows with the same key will have their comments merged in the resulting adjustment transaction entry. This behaviour would be good to confirm with Manon / Kasia to see if that's acceptable.

EWaterman:
whoever picks this up, please test the code out. I didn't have time to test it myself thoroughly before leaving

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -185,35 +185,32 @@ class InventoryImportDataService implements ImportDataService {
...
+            int adjustmentQuantity = totalQuantity - quantityOnHand
+            if (adjustmentQuantity == 0) {
+                continue
+            }
 
-                if (adjustmentQuantity == 0) {
-                    continue
-                }
+            // We know we have at least one row for the key at this point so this is safe to do.
+            InventoryImportDataRow row = rowsForKey[0]
```
awalkowiak, line 203: What if we are importing with different products, will it still work?

ewaterman: Yes. `inventoryImportData.rows` is a `Map<String, List<InventoryImportDataRow>>` keyed on [product code + bin location name + lot number] so every value in the rowsForKey List have the same product.

(PS I'm only replying because I'm at the train station and bored. I'm going back on vacation now :sweat_smile:)

PR #5334, opened 2025-06-15
Brazilian portuguese translation

ricardoapaes:

[//]: <> Adding more terms translated into Brazilian Portuguese in the project.

**Description:**

I'm Brazilian and it would be very useful to use the project in my language, the idea would be to gradually create a translation into Brazilian Portuguese.


EWaterman:
Hi Ricardo! We're thrilled to have new translations for the app! This has been a huge gap for us so it's amazing to have your help.

Have you checked out our [Crowdin page](https://crowdin.com/project/openboxes)? That's how we input our translations (and then they get pulled in automatically to the app via pull requests [like this one](https://github.com/openboxes/openboxes/pull/5189)).

@jmiranda @awalkowiak do you think this would work as is or do the translations need to be done through crowdin?

awalkowiak (changes requested):
@EWaterman thanks for reminding me about that PR. 

@ricardoapaes thank you for the translations! I have one suggestion - this should follow the same route as Spanish (Mexico). You'd need to do:
1. All common translations for `pt` and `pt_BR`, and this should be done through the Crowdin (you can request to join as translator this project https://crowdin.com/project/openboxes/, and you should be able to import your file there)
2. All differences, custom translations relevant for Brazil should be added in the `pt_BR` (see this file for Spanish (Mexico) https://github.com/openboxes/openboxes/blob/develop/grails-app/i18n/messages_es_MX.properties), in your case it would be `messages_pt_BR.properties` <- this file needs to be added to the repository directly, does not have to go through crowdin
3. Add `pt_BR` here https://github.com/openboxes/openboxes/blob/develop/grails-app/conf/application.yml#L552

PR #5335, opened 2025-06-16
:whale: Setting base url using Docker and removing .env

ricardoapaes:

[//]: <> Add custom domain support and .env.example

**Description:**

This pull request addresses an issue encountered when deploying the application to a production environment with a custom domain and a reverse proxy.

Previously, after making local adaptations, the application functioned as expected on localhost:80. However, upon deployment to production, the login functionality failed due to redirects to http://app:8080/.

After investigating, I discovered that an environment variable could be used to specify the application's domain. I tested this solution, and it successfully resolved the redirection issue. Therefore, I am proposing an update to the docker-compose.yml file to include this necessary environment variable.

Additionally, this pull request proposes replacing the .env file with a .env.example file. This change ensures that local environment configurations are not inadvertently committed to the repository, as .env will be added to .gitignore. This promotes better security practices and simplifies environment management for developers.

***Why these changes?***
- Custom Domain Support: Enables seamless deployment and proper login redirection when using a custom domain with a reverse proxy in production.
- Improved Environment Management: Prevents sensitive environment variables from being committed to the repository, enhancing security and streamlining development workflows.

### DIFF
--- a/docker/docker-compose-base.yml
+++ b/docker/docker-compose-base.yml
@@ -8,6 +8,7 @@ services:
         DATASOURCE_USERNAME: ${DATASOURCE_USERNAME:-openboxes}
         DATASOURCE_PASSWORD: ${DATASOURCE_PASSWORD:-openboxes}
         DATASOURCE_URL: ${DATASOURCE_URL:-jdbc:mysql://db/openboxes?useSSL=false}
+        GRAILS_SERVERURL: ${OPENBOXES_SERVER_URL}
         JAVA_TOOL_OPTIONS: ${JAVA_TOOL_OPTIONS:--Xms1024m -Xmx1024m -XX:+UseParallelGC -Djava.awt.headless=true}
       healthcheck:
         test: "curl --fail --silent localhost:8080/openboxes/health | grep UP || exit 1"
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -2,3 +2,4 @@ DATASOURCE_URL=jdbc:mysql://host.docker.internal:3306/openboxes?useSSL=false
 DATASOURCE_USERNAME=openboxes
 DATASOURCE_PASSWORD=openboxes
 JAVA_TOOL_OPTIONS=-Xms1024m -Xmx1024m -XX:+UseParallelGC -Djava.awt.headless=true
+OPENBOXES_SERVER_URL=https://youdomain.com/openboxes/
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,1 +1,2 @@
 /mysql
+.env


ewaterman (approved):

docker/docker-compose-base.yml
```diff
@@ -8,6 +8,7 @@ services:
         DATASOURCE_USERNAME: ${DATASOURCE_USERNAME:-openboxes}
         DATASOURCE_PASSWORD: ${DATASOURCE_PASSWORD:-openboxes}
         DATASOURCE_URL: ${DATASOURCE_URL:-jdbc:mysql://db/openboxes?useSSL=false}
+        GRAILS_SERVERURL: ${OPENBOXES_SERVER_URL}
```
ewaterman: to look into: What happens if we don't define a value for `OPENBOXES_SERVER_URL`? Does it fail or does it fall back to the old behaviour?

jmiranda: Yeah that's a good question. Does yml support conditional logic? My suspicion is no.

jmiranda: To clarify my point since there's a bit baked into "conditional logic" ...

By that I mean you want yml to support something like this so we don't set the GRAILS_SERVERURL unless it's provided. 
```
if (env.OPENBOXES_SERVER_URL) 
   GRAILS_SERVERURL: ${OPENBOXES_SERVER_URL}
```
But I don't really think that's possible without making the yml file a jinja template file (yml.j2) that gets rendered. 

The next best thing would be to verify the default value set by Grails and set it ourselves.
```
GRAILS_SERVERURL: ${OPENBOXES_SERVER_URL:default}
```
where default is most likely something like 
```
http://localhost:8080
```

jmiranda: With all of that said, the docker-compose files are under our control so requiring a develop to have to set a variable of our choosing (i.e. OPENBOXES_SERVER_URL) seems appropriate. The only thing we need to worry about is making sure it's documented (additionally it would be nice if we could find a way to handle the case of the value not being set in the environment and report the error to the developer in a clear manner).

I'm wondering why we're switching between GRAILS_SERVERURL and OPENBOXES_SERVER_URL. I think I would prefer to stick with requiring the environment to include GRAILS_SERVERURL because it's more self-documenting (at least to a Grails developer) than OPENBOXES_SERVER_URL in the case where we don't document all of the available env properties.

Lastly, GRAILS_SERVERURL looked weird from the first time I saw it and now that I'm thinking about it, I think it might actually be wrong. The config property is `grails.serverURL` which I believe means that Grails will be looking for a variable like `GRAILS_SERVER_URL` (both the period and case change get hyphentated).

ewaterman: yeah I agree that I don't personally see a need for `OPENBOXES_SERVER_URL` if we have the known `GRAILS_SERVER_URL` property available (I think you're right and we do need the underscore).

I'm personally comfortable with having a default like we do the other env vars. I looked it up and the default would be `http://localhost:8080/openboxes`

PR #5336, opened 2025-06-16
OBPIH-7290 Assign counted by and recounted by while creating CC

alannadolny:


grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -472,13 +475,19 @@ class CycleCountService {
         // 1:1 association between cycle count and cycle count request
         request.cycleCountRequest.cycleCount = newCycleCount
         request.cycleCountRequest.status = CycleCountRequestStatus.IN_PROGRESS
+        Person assignee = request.cycleCountRequest.countAssignee
+        LocalDate countDeadline = request.cycleCountRequest.countDeadline
+        Date deadline = countDeadline ? DateUtil.asDate(countDeadline) : new Date()
```
kchelstowski, line 480: I know it's not your convention, but it feels weird to me that we use two different names for the same thing. I'm just thinking why can't we call it `dateCounted` also on `cycleCountRequest` instead of `deadline` :thinking:

kchelstowski (approved):

PR #5338, opened 2025-06-17
OBPIH-7321 Generate dashboard tiles

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7321

**Description:**
image

---


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -1033,6 +1033,26 @@ class UrlMappings {
             action = [POST: "getInventoryAuditSummary", GET: "getInventoryAuditSummary"]
         }
 
+        "/api/reports/indicators/totalCount" {
+            controller = "indicator"
```
SebastianLib: Do we want to rename this to indicatorApi? I used the old IndicatorController.

kchelstowski: let's rename it to include `api`

grails-app/controllers/org/pih/warehouse/reporting/IndicatorController.groovy
```diff
@@ -26,7 +32,37 @@ class IndicatorController {
...
+        render([data: data] as JSON)
+    }
+
+    def getNotFinishedItems(IndicatorCommand command) {
+        def data = reportService.getNotFinishedItems(command)
+        render([data: data] as JSON)
+    }
 }
 
+class IndicatorCommand implements Validateable {
```
kchelstowski: let's separate it from the controller class

grails-app/controllers/org/pih/warehouse/reporting/IndicatorController.groovy
```diff
@@ -26,7 +32,37 @@ class IndicatorController {
...
+    Location facility
+    Date startDate
+    Date endDate
 
+    static constraints = {
+        facility(nullable: false)
```
kchelstowski: `nullable: false` is the default setting. You don't have to write it explicitly.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1133,4 +1134,33 @@ class ReportService implements ApplicationContextAware {
 
         return new PaginatedList<InventoryAuditSummary>(data, results.totalCount);
     }
-}
\ No newline at end of file
+
+    def getTotalCount(IndicatorCommand command) {
```
kchelstowski: I know it's only mocked result, but anyway you know the result type - it's `Map`, it doesn't have to be `def`.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1133,4 +1134,33 @@ class ReportService implements ApplicationContextAware {
...
+        return [
+                type: 'totalCount',
```
kchelstowski: we used to use double `"` for strings on the backend.

src/js/components/cycleCountReporting/IndicatorsTab/IndicatorNumberCards.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+
+import PropTypes from 'prop-types';
+import { RiInformationLine } from 'react-icons/ri';
+import { SortableContainer } from 'react-sortable-hoc';
+
+import LoadingNumbers from 'components/dashboard/LoadingNumbers';
+import NumberCard from 'components/dashboard/NumberCard';
+
+const SortableIndicatorCards = SortableContainer(({ data }) => (
```
kchelstowski: Let's stick to one component per file.
In other words: let's make the `SortableIndicatorCards` a separate file.

src/js/components/cycleCountReporting/IndicatorsTab/IndicatorNumberCards.jsx
```diff
@@ -0,0 +1,64 @@
...
+  <div className="card-component">
+    {data.map((card, index) => (
+      <NumberCard
```
kchelstowski: isn't `key` missing?

src/js/components/cycleCountReporting/IndicatorsTab/IndicatorNumberCards.jsx
```diff
@@ -0,0 +1,64 @@
...
+        disabled
+        hideDraghandle
+      />
+    ))}
+  </div>
+));
+
+const IndicatorNumberCards = ({
+  loading,
+  numberCards,
```
kchelstowski: what is `numberCards`? It might be difficult to understand the context without digging deeper into the code. Maybe a different name fits better?

src/js/utils/option-utils.jsx
```diff
@@ -311,6 +311,26 @@ export const fetchOrganization = async (id) => {
...
+  return data.data;
+};
+
+export const fetchIndicatorTargetProgress = async (params = {}) => {
+  const { data } = await apiClient.get('/api/reports/indicators/targetProgress', { params });
+  return data.data;
+};
+
+export const fetchIndicatorNotFinishedItems = async (params = {}) => {
+  const { data } = await apiClient.get('/api/reports/indicators/notFinishedItems', { params });
```
kchelstowski: not a must have, but if we know such endpoints will exists, we can already prepare api/url frontend factory for that.

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1133,4 +1134,33 @@ class ReportService implements ApplicationContextAware {
...
+                number: 42,
+                subtitleValue: "1281/26182",
+        ]
+    }
+
+    Map getNotFinishedItems(IndicatorCommand command) {
+        return [
+                type: "notFinishedItems",
+                number: 300
+        ]
```
alannadolny: I think we can use more appropriate naming of the properties, and a little bit more "generic"
type: sounds like a type of the indicator, but it looks to be the name of that indicator, so I would call it `name`
number: I think it can be `value`, so in every indicator it can be value, not number, string, etc.
subtitleValue: Because we have value, this one can be named `subvalue`

src/js/components/cycleCountReporting/IndicatorsTab/SortableIndicatorCards.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import { RiInformationLine } from 'react-icons/ri';
+import { SortableContainer } from 'react-sortable-hoc';
+
+import NumberCard from 'components/dashboard/NumberCard';
+
+const SortableIndicatorCards = SortableContainer(({ data }) => (
```
alannadolny, line 8: Should this container for indicators be sortable? I would say that, for reporting, it's not needed, but I am not sure what the requirements are.

src/js/consts/cycleCountReportingIndicators.js
alannadolny, line 1: Because we are storing the config on the frontend side, we can add a type property which will be the component that should be used while generating tiles. Adding that, I think you can remove the "numberType"? But I am not sure.
The benefit of that will be something like:
```
  <div className="card-component">
    {data.map((card, index) => (
      <Tile
        key={card.type}
        index={index}
        cardTitle={card.title}
        cardTitleDefaultValue={card.titleDefaultValue}
        cardNumberType={card.numberType}
        cardNumber={card.number}
        cardType={card.type}
        cardSubtitle={card.subtitle}
        cardSubtitleDefaultValue={card.subtitleDefaultValue}
        cardSubtitleValue={card.subtitleValue}
        cardInfo={card.info}
        cardInfoDefaultValue={card.infoDefaultValue}
        showPercentSign={card.showPercentSign}
        infoIcon={<RiInformationLine size={20} />}
        disabled
        hideDraghandle
      />
    ))}
  </div>
```
  Where the Tile will be a property from the config.

src/js/consts/cycleCountReportingIndicators.js
```diff
@@ -0,0 +1,41 @@
+const cycleCountReportingIndicators = {
+  totalCount: {
+    title: 'react.cycleCount.indicators.cardTitle.totalCount.label',
+    titleDefaultValue: 'Total count physical inventory',
+    numberType: 'number',
+    subtitle: 'react.cycleCount.indicators.cardSubtitle.inMyInventory.label',
+    subtitleDefaultValue: 'in my inventory',
+    info: 'react.cycleCount.indicators.cardInfo.totalCount.label',
+    infoDefaultValue: 'Total physical inventory count',
```
alannadolny: title -> titleLabel
titleDefaultValue -> defaultTitle
numberType -> not sure what it means
subtitle -> subtitleLabel
subtitleDefaultValue -> defaultSubtitle
info -> infoLabel
infoDefaultValue -> defaultInfo

src/js/components/cycleCountReporting/IndicatorsTab/SortableIndicatorCards.jsx
SebastianLib: I’m adding `SortableContainer` here because in NumberCard there is this line: c`onst NumberCard = SortableElement({...`, so if I remove `SortableContainer`, an error occurs. I could remove it, but then I would have to create a separate file without it, so I think it’s fine to leave it as is.

src/js/consts/cycleCountReportingIndicators.js
SebastianLib: numberType was already present before my changes in this lines           
```
<span className="result-card">
            {' '}
            {cardNumberType === 'number' ? `${cardNumberLocale}${showPercentSign ? '%' : ''}` : `${cardNumberLocale} ${currencyCode}`}
            {' '}
`</span>`

alannadolny (approved):

PR #5339, opened 2025-06-17
OBPIH-7271 Fix validations on lot number, bin location, date counted …

kchelstowski:
…in cycle count import

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
@@ -9,6 +9,17 @@ class CycleCountImportService {
 
     LocalizationService localizationService
 
+    /**
+     * Util function to normalize the default lot number - default lot number can be represented
+     * by either empty string or null. If we compare lot numbers where one of them is empty string
+     * and the other is null, we want them to be treated as equal.
+     * @param lotNumber
+     * @return
+     */
+    private static normalizeLotNumber(String lotNumber) {
+        return lotNumber ?: null
+    }
+
     void validateCountImport(ImportDataCommand command) {
         // Firstly clear the command's errors, as it might contain some additional errors related to the file itself
         // that we do not care about while validating the data, and we don't them to be visible for a user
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
```diff
@@ -9,6 +9,17 @@ class CycleCountImportService {
...
     LocalizationService localizationService
 
+    /**
+     * Util function to normalize the default lot number - default lot number can be represented
+     * by either empty string or null. If we compare lot numbers where one of them is empty string
+     * and the other is null, we want them to be treated as equal.
+     * @param lotNumber
+     * @return
+     */
+    private static normalizeLotNumber(String lotNumber) {
```
alannadolny, line 19: Is the word "normalize" proper here? Won't "standardize" be better?

src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -632,8 +632,12 @@ const useCountStep = () => {
...
                 if (correspondingImportItem && !assigneeImported.current[cycleCount.id]) {
                   assignCountedBy(cycleCount.id)(correspondingImportItem.assignee);
-                  setDateCounted((prevState) =>
-                    ({ ...prevState, [cycleCount.id]: correspondingImportItem.dateCounted }));
+                  // Do not allow to clear the date counted dropdown
+                  // if dateCounted was not set in the sheet
+                  if (correspondingImportItem.dateCounted) {
+                    setDateCounted((prevState) =>
+                      ({ ...prevState, [cycleCount.id]: correspondingImportItem.dateCounted }));
+                  }
```
alannadolny, line 640: Can you explain that a little bit more? I don't understand the exact case that you described in the comment above the code.

kchelstowski: From the backend perspective, if the date counted is invalid, it is returned as `null` from the import.
If it's invalid, we don't want to update it on the frontend though (it was Kuba's request), so if `correspondingImportItem.dateCounted` exists, we know, it must be valid at this point, so update the state.

grails-app/services/org/pih/warehouse/inventory/CycleCountImportService.groovy
kchelstowski: I find them as synonyms, so let me know, and let's here from others what you prefer, so I can eventually adjust it.

PR #5350, opened 2025-06-26
OBPIH-7287 Cycle count assignment fixes

alannadolny:


grails-app/domain/org/pih/warehouse/inventory/CycleCountRequest.groovy
```diff
@@ -75,6 +78,24 @@ class CycleCountRequest {
         recountDeadline(nullable: true)
     }
 
+    // Access to the information about available cycle count items before creating the cycle count
+    List<AvailableItem> getItemsAvailableForCycleCount() {
+        CycleCountProductAvailabilityService cycleCountProductAvailabilityService = Holders.grailsApplication.mainContext.getBean(CycleCountProductAvailabilityService)
```
awalkowiak: I don't like the idea of loading service in the domain. Is there an option to do this inside the service instead?

alannadolny: I also don't like that, but nothing better comes to my mind.

awalkowiak: I would just calculate it and concatenate with toJson result, or perhaps ou could just add a transient with setter for this one :thinking:

PR #5351, opened 2025-06-26
OB-3254: Implemented logic for auto receiving

druchniewicz:


grails-app/controllers/org/pih/warehouse/RoleInterceptor.groovy
```diff
@@ -112,7 +112,7 @@ class RoleInterceptor {
         }
 
         if (!rule) {
-            log.debug "No rule for ${controllerName}:${actionName} -> allow anonymous"
+//            log.debug "No rule for ${controllerName}:${actionName} -> allow anonymous"
```
druchniewicz: temp change

grails-app/services/org/pih/warehouse/core/UserService.groovy
```diff
@@ -291,7 +291,7 @@ class UserService {
         User user = getUser(userId)
         return getEffectiveRoles(user).any { Role role ->
             boolean acceptedRoleType = acceptedRoleTypes.contains(role.roleType)
-            log.debug "Is role ${role.roleType} in ${acceptedRoleTypes} = ${acceptedRoleType}"
+//            log.debug "Is role ${role.roleType} in ${acceptedRoleTypes} = ${acceptedRoleType}"
```
druchniewicz: temp change

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -489,4 +489,38 @@ class ReceiptService {
...
+            partialReceiptItem.quantityReceiving = item.quantity
+            partialReceiptItem.binLocation = createTemporaryReceivingBin(shipment)
+            partialReceiptItem.product = item.product
+            partialReceiptItem.shouldSave = item.quantity != null
+            partialReceiptContainer.partialReceiptItems.add(partialReceiptItem)
+        }
+
+        savePartialReceipt(partialReceipt, true)
+
+        //does not change status
```
druchniewicz: does not work yet

jmiranda:
I might not have time this week, but let me review before this gets merged.

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
awalkowiak: @druchniewicz you have to create a receipt event. See this part of code: 
https://github.com/openboxes/openboxes/blob/develop/grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy#L278 ->
https://github.com/openboxes/openboxes/blob/develop/grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy#L309-L331

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptCreationJob.groovy
```diff
@@ -0,0 +1,33 @@
...
+        if (!Holders.config.openboxes.jobs.automaticReceiptCreationJob.enabled) {
+            log.info"Automatic receipt creation job is disabled"
+            return
+        }
+
+        log.info "Running automatic receipt creation job ... "
+        List<Shipment> shippedShipments = shipmentService.getAllShippedShipments()
+        shippedShipments.each {
+            if (!it.isFullyReceived()) {
+//                receiptService.createAutomaticReceipt(it)
```
druchniewicz: temp comment

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptCreationJob.groovy
```diff
@@ -0,0 +1,39 @@
+package org.pih.warehouse.jobs
+
+import grails.util.Holders
+import org.pih.warehouse.core.ActivityCode
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.shipping.Shipment
+import org.quartz.JobExecutionContext
+
+class AutomaticReceiptCreationJob {
```
jmiranda: Just call it AutomaticReceiptJob

jmiranda: and make that change everywhere there's AutomaticReceiptCreation.

grails-app/domain/org/pih/warehouse/shipping/Shipment.groovy
```diff
@@ -44,10 +49,12 @@ class Shipment implements Comparable, Serializable, Historizable {
 
     def afterInsert() {
         publishRefreshEvent()
+        publishAutomaticReceiptEvent()
```
jmiranda: Do we need this? If this is triggered, it should be triggered for the specific shipment, not for all shipments. Otherwise, we're potentially going to invoke this once for every shipment created. I would rather the automatic receipt trigger come from somewhere else i.e. integration layer (or if we keep the quartz job ... just let it do its thing when the cron triggers the job.

grails-app/services/org/pih/warehouse/shipping/ShipmentStatusTransitionEventService.groovy
```diff
@@ -9,18 +9,26 @@
...
 @Transactional
 class ShipmentStatusTransitionEventService implements ApplicationListener<ShipmentStatusTransitionEvent> {
 
+    GrailsApplication grailsApplication
     def notificationService
     def webhookPublisherService
 
     void onApplicationEvent(ShipmentStatusTransitionEvent event) {
+
+        def areNotificationsEnabled = grailsApplication.config.openboxes.shipping.notifications.enabled ?: false
```
jmiranda: notificationsEnabled

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptCreationJob.groovy
```diff
@@ -0,0 +1,39 @@
...
+            return
+        }
+
+        log.info "Running automatic receipt creation job... "
+        List<Location> autoReceivingLocations = locationService.getLocationsSupportingActivities([ActivityCode.AUTO_RECEIVING])
+        autoReceivingLocations.each {
+            List<Shipment> shippedShipments = shipmentService.getShippedShipmentsByDestination(it)
+            shippedShipments.each {
+                if (!it.isFullyReceived()) {
+                    receiptService.createAutomaticReceipt(it)
```
jmiranda: We need to be careful that there hasn't been a partial receipt already. Since the time window is 5 minutes, it's possible that a user could sneak in a few receipt items on the mobile. And if the createAutomaticReceipt() method just copies data from the shipment, then we're hosed (could lead to an over-receipt).

jmiranda: Also I feel like if we keep the quartz job, we'll want a way to pass a shipmentId to the job and only process that one. And/or we want an API that allows for an external client to trigger the auto-receipt (note: this won't trigger the quartz job).

grails-app/conf/application.yml
```diff
@@ -603,6 +608,8 @@ openboxes:
     shipping:
         search:
             maxResults: 1000
+        notifications:
+            enabled: false
```
jmiranda: Do we have a system-level notifications flag? I think there's a supported activity if I'm not wrong.

jmiranda (commented):
The most pressing issue is the automatic receipt event being emitted on saving a shipment. If anything, it should trigger the job just for that shipment if the location has auto receiving enabled.

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptCreationJob.groovy
olewandowski1: Done

grails-app/conf/application.yml
olewandowski1: Changed

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptJob.groovy
```diff
@@ -0,0 +1,68 @@
...
+
+                log.info("Creating automatic receipt for shipment ${shipmentId}")
+                receiptService.createAutomaticReceipt(shipment)
+            } catch (Exception e) {
+                log.error("Error processing shipment ${shipmentId}", e)
+            }
+            return
+        }
+
+        if (Holders.config.openboxes.jobs.automaticReceiptJob.shipmentStatusCheckerEnabled) {
```
awalkowiak: You forgot to replace `shipmentStatusCheckerEnabled` here with `bulkShipmentAutoReceipt`

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptJob.groovy
```diff
@@ -0,0 +1,68 @@
...
+            log.info "Running automatic receipt job for all shipped shipments... "
+            List<Location> autoReceivingLocations = locationService.getLocationsSupportingActivities([ActivityCode.AUTO_RECEIVING])
+            autoReceivingLocations.each {
+                List<Shipment> shippedShipments = shipmentService.getShippedShipmentsByDestination(it)
+                shippedShipments.each {
+                    if (!it.isFullyReceived()) {
```
awalkowiak: I wonder if this should not be expanded for a case when you might have not yet received receipt (or partial receipt) at that moment. In this version  it will always try to create a new receipt and we might get too much receipts in the end. It would be safer to check also if the shipment does not have receipt created already that is not yet received (so if it is received, but we have some quantity remaining, then we create a new one).

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -489,4 +491,35 @@ class ReceiptService {
...
+
+            PartialReceiptItem partialReceiptItem = partialReceiptContainer.partialReceiptItems.find {
+                it?.shipmentItem?.id == item.id
+            }
+
+            if (!partialReceiptItem) {
+                partialReceiptItem = new PartialReceiptItem()
+            }
+
+            partialReceiptItem.quantityReceiving = item.quantity
```
awalkowiak: I think it should take `item.quantityRemaining` instead full `quantity`

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptJob.groovy
```diff
@@ -0,0 +1,68 @@
...
+                    log.warn("Shipment ${shipmentId} not found, skipping")
+                    return
+                }
+
+                if (!shipment.destination?.supports(ActivityCode.AUTO_RECEIVING)) {
+                    log.debug("Shipment ${shipmentId}: origin does not support AUTO_RECEIVING")
+                    return
+                }
+
+                if (shipment.isFullyReceived()) {
```
awalkowiak, line 43: Same case as below, we might have a case when there is a partial receipt (pending or received), so we'd need to makge sure that we are not creating redundant receipts

awalkowiak (changes requested):

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptJob.groovy
olewandowski1: Added a check for pending shipment receipts - if none are found, we can be confident the remaining quantity is fully up to date.

grails-app/jobs/org/pih/warehouse/jobs/AutomaticReceiptJob.groovy
olewandowski1: Added a check for pending shipment receipts - if none are found, you can be confident the remaining quantity is fully up to date.

awalkowiak (approved):

PR #5352, opened 2025-06-27
OBPIH-7306 Refactored inventory audit API to include actual data  

jmiranda:


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1122,18 +1120,73 @@ class ReportService implements ApplicationContextAware {
                 lte("transactionDate", command.endDate)
             }
         }
+    }
+
+    def getInventoryAuditReportSummary(GetInventoryAuditReportCommand command) {
+
+        def inventoryAuditFilters = buildInventoryAuditSummaryFilters(command)
```
SebastianLib, line 1128: Could you add an option to filter by multiple products?

jmiranda: Ah right, thanks for the reminder.

jmiranda: Fixed in the most recent commit

awalkowiak (approved):

PR #5355, opened 2025-06-30
OBPIH-7323 Create indicator: inventory accuracy

SebastianLib:

**Description:**

---


grails-app/migrations/views/cycle-count-product-summary.sql
```diff
@@ -0,0 +1,48 @@
...
+        SELECT MAX(count_index)
+        FROM cycle_count_item
+                 LEFT OUTER JOIN cycle_count ON cycle_count_item.cycle_count_id = cycle_count.id
+        WHERE cycle_count_item.cycle_count_id = cycle_count_item_variance.cycle_count_id
+          AND cycle_count_item.product_id = cycle_count_item_variance.product_id
+          AND cycle_count_item.inventory_item_id = cycle_count_item_variance.inventory_item_id
+          AND cycle_count.facility_id = cycle_count_item_variance.facility_id
+    )
+)
+
```
SebastianLib, line 40: I removed the cycle_count_item_statistics from the view because, at this point, it doesn't seem necessary and was causing errors 
```
cycle_count_item_statistics AS
(
SELECT sum(case when has_variance = FALSE then 1 else 0 end) as accurate_item_count,
       count(cycle_count_item_id)                            as total_cycle_count_item_count,
       count(distinct product_id) as total_product_count,
       count(cycle_count_id)                                 as total_cycle_count_count

FROM cycle_count_item_final_count
    )
```

grails-app/migrations/views/cycle-count-product-summary.sql
```diff
@@ -0,0 +1,48 @@
...
+    SELECT *
+    FROM cycle_count_item_variance
+    WHERE cycle_count_item_variance.count_index = (
+        SELECT MAX(count_index)
+        FROM cycle_count_item
+                 LEFT OUTER JOIN cycle_count ON cycle_count_item.cycle_count_id = cycle_count.id
+        WHERE cycle_count_item.cycle_count_id = cycle_count_item_variance.cycle_count_id
+          AND cycle_count_item.product_id = cycle_count_item_variance.product_id
+          AND cycle_count_item.inventory_item_id = cycle_count_item_variance.inventory_item_id
+          AND cycle_count.facility_id = cycle_count_item_variance.facility_id
```
SebastianLib, line 37: Initially, there was AND `cycle_count_item.location_id = cycle_count_item_variance.location_id` here, but I changed it to `AND cycle_count.facility_id = cycle_count_item_variance.facility_id` because cycle_count_item.location_id referred to the bin location, not our current location. As a result, locations without a bin location were not returned by this query.

grails-app/domain/org/pih/warehouse/reporting/CycleCountProductSummary.groovy
```diff
@@ -0,0 +1,31 @@
...
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class CycleCountProductSummary implements Serializable {
+
+    Long cycleCountId
+    Date dateCounted
+    Location facility
+    Product product
+    Boolean hasVariance
```
alannadolny, line 12: How is the variance working here? Shouldn't it be a number? I assume when the variance is 0, the `hasVariance` is false, but that looks a little bit odd to me.

grails-app/domain/org/pih/warehouse/reporting/InventoryAccuracyResult.groovy
```diff
@@ -0,0 +1,13 @@
+package org.pih.warehouse.reporting
+
+import java.math.RoundingMode
+
+class InventoryAccuracyResult implements Serializable {
+    Integer accurateCount
+    Integer totalCount
+
+    BigDecimal getAccuracyPercentage() {
+        if (!totalCount) return 0
```
alannadolny: add curly brackets

grails-app/domain/org/pih/warehouse/reporting/CycleCountProductSummary.groovy
SebastianLib: hasVariance is a Boolean, so I have to set it as a Boolean value. In the view, we do something like this:
`(CASE WHEN quantity_variance != 0 THEN 1 ELSE 0 END) AS has_variance`

grails-app/domain/org/pih/warehouse/reporting/CycleCountProductSummary.groovy
```diff
@@ -0,0 +1,31 @@
+package org.pih.warehouse.reporting
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class CycleCountProductSummary implements Serializable {
+
+    Long cycleCountId
```
kchelstowski: actually, why `Long`? is it the CRC id generated in the view? I'm asking because typical cycle count id has uuid type (String)

grails-app/migrations/views/cycle-count-product-summary.sql
```diff
@@ -0,0 +1,48 @@
...
+    )
+)
+
+SELECT
+    cycle_count_id,
+    date_counted,
+    facility_id,
+    product_id,
+    COALESCE(has_variance) AS has_variance
```
kchelstowski: is the `COALESCE` needed if you have only one value inside?

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1165,6 +1166,36 @@ class ReportService implements ApplicationContextAware {
         ]
     }
 
+    Map getInventoryAccuracy(IndicatorApiCommand command) {
+        String query = """
+        SELECT 
+            SUM(CASE WHEN cps.hasVariance = false THEN 1 ELSE 0 END),
```
kchelstowski, line 1240: I guess we are not able to make that `CASE` in a typical `createCriteria` so I guess I'm fine with the HQL query here.

PR #5370, opened 2025-07-07
OBPIH-7322 Create indicator: products inventoried (fix)

SebastianLib:

**Description:**

---


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1206,18 +1206,21 @@ class ReportService implements ApplicationContextAware {
     }
 
     Map getProductsInventoried(IndicatorApiCommand command) {
```
SebastianLib, line 1208: I fixed it because the previous logic only looked at the last counted date, which caused some products to be missing from the indicator when they were actually counted within the selected date range. Now we checks if the product was counted at least once within the range, regardless of when the last count was

awalkowiak (approved):

PR #5371, opened 2025-07-07
OBPIH-7361 Fix single product dialog in transaction report (Fix after QA)

alannadolny:


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1349,15 +1354,15 @@ class ReportService implements ApplicationContextAware {
...
+                    order(orderBy, sortOrder)
                 }
             }
         } as List<TransactionEntry>
     }
 
-    List<TransactionEntry> getFilteredTransactionEntries(List<TransactionCode> transactionCodes, Date startDate, Date endDate, Location location, Product product, String orderBy) {
-        return getFilteredTransactionEntries(transactionCodes, startDate, endDate, null, null, null, location, product, orderBy)
+    List<TransactionEntry> getFilteredTransactionEntries(List<TransactionCode> transactionCodes, Date startDate, Date endDate, Location location, Product product, String orderBy, String sortOrder = "desc") {
+        return getFilteredTransactionEntries(transactionCodes, startDate, endDate, null, null, null, location, product, orderBy, sortOrder)
```
kchelstowski, line 1365: those arguments are getting strange. If another refactor is needed here, I'd consider creating a command class for those arguments as we some time ago agreed that we shouldn't pass more than 3/4 arguments.

PR #5372, opened 2025-07-08
OBPIH-7198 Add possibility to migrate old product inventory transactions

awalkowiak:
This is the first iteration of the old product inventory transactions migration to the Inventory Baseline and Adjustment pair. It is not the best performance-wise, but I am not sure if I can improve it by a lot (or at all?). There are some questionable parts of the code, like the usage of pure sql update, so if you have any questions, let me know. Some further improvements that I am considering:
- Option to do migration in batches of 100-200 transactions at once
- Showing progress :thinking: (this is really nice to have and would require a new domain)

Current data migration page example:
Screenshot from 2025-07-08 14-37-20

### DIFF
--- a/grails-app/conf/runtime.groovy
+++ b/grails-app/conf/runtime.groovy
@@ -90,16 +90,6 @@ openboxes.products.merge.enabled = false
 // Cycle Count configuration (OBPIH-7033)
 openboxes.cycleCount.products.maxAmount = 50
 
-// Inventory baseline transaction configuration (OBPIH-7194, OBPIH-7195)
-openboxes.transactions.inventoryBaseline.recordStock.enabled = true
-openboxes.transactions.inventoryBaseline.inventoryImport.enabled = true
-
-// Inventory snapshot configuration (OBPIH-7254)
-openboxes.transactions.inventoryBaseline.loadDemoData.enabled = true
-
-// Inventory snapshot configuration (OBPIH-7198)
-openboxes.transactions.inventoryBaseline.migration.enabled = true
-
 openboxes.security.rbac.rules = [
     [controller: '*', actions: ['delete'], accessRules: [ minimumRequiredRole: RoleType.ROLE_SUPERUSER ]],
     [controller: '*', actions: ['remove'], accessRules: [ minimumRequiredRole: RoleType.ROLE_SUPERUSER ]],
--- a/grails-app/controllers/org/pih/warehouse/data/MigrationController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/data/MigrationController.groovy
@@ -379,7 +379,8 @@ class MigrationController {
         Map results = migrationService.migrateProductInventoryTransactions(location, performMigration)
 
         long responseTime = System.currentTimeMillis() - startTime
-        String responseTimeMessage = "${performMigration ? 'Migration' : 'Generating preview'} took ${(responseTime)} ms"
+        String responseTimeMessage = "${performMigration ? 'Migration' : 'Generating preview'} of product inventory " +
+                "transactions at location ${location.name} took ${(responseTime)} ms"
         log.info "$responseTimeMessage"
 
         render([responseTime: "$responseTimeMessage", results: results] as JSON)
--- a/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
@@ -9,6 +9,7 @@
  **/
 package org.pih.warehouse.data
 
+import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
 import groovy.sql.Sql
@@ -49,6 +50,7 @@ class MigrationService {
     ProductInventoryTransactionMigrationService productInventoryTransactionMigrationService
     def persistenceInterceptor
     def dataSource
+    GrailsApplication grailsApplication
 
     def getStockMovementsWithoutShipmentItems() {
         String query = """
@@ -404,7 +406,7 @@ class MigrationService {
 
         // 1. Find all transactions with the old Product Inventory type (by PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
         TransactionType oldProductInventoryType = TransactionType.get(Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
-        List transactions = Transaction.findAllByInventoryAndTransactionType(location.inventory, oldProductInventoryType)?.sort { it.transactionDate}
+        List<Transaction> transactions = Transaction.findAllByInventoryAndTransactionType(location.inventory, oldProductInventoryType)?.sort { it.transactionDate}
 
 
         Map<String, List<String>> results = [:]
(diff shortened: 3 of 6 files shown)


grails-app/views/migration/dataMigration.gsp
alannadolny, line 1: Shouldn't those sentences be translatable?

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
         return results
     }
 
+    Map migrateProductInventoryTransactions(Location location, boolean performMigration) {
+        log.info("Migrating Product inventory transactions at location ${location.name} ${performMigration}")
+
+        // 1. Find all transactions with the old Product Inventory type (by PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
+        TransactionType oldProductInventoryType = TransactionType.get(Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
+        List transactions = Transaction.findAllByInventoryAndTransactionType(location.inventory, oldProductInventoryType)?.sort { it.transactionDate}
```
alannadolny:
```suggestion
        List<Transaction> transactions = Transaction.findAllByInventoryAndTransactionType(location.inventory, oldProductInventoryType)?.sort { it.transactionDate }
```

grails-app/conf/runtime.groovy
```diff
@@ -97,6 +97,9 @@ openboxes.transactions.inventoryBaseline.inventoryImport.enabled = true
 // Inventory snapshot configuration (OBPIH-7254)
 openboxes.transactions.inventoryBaseline.loadDemoData.enabled = true
 
+// Inventory snapshot configuration (OBPIH-7198)
+openboxes.transactions.inventoryBaseline.migration.enabled = true
```
ewaterman: I realize all the other config options are here so this is fine but I'm curious why have this in runtime.groovy instead of application.yml?

src/main/groovy/org/pih/warehouse/core/Constants.groovy
```diff
@@ -88,13 +88,15 @@ class Constants {
 
     // these are direct references to transaction types by primary key
     static final String CONSUMPTION_TRANSACTION_TYPE_ID = "2"
+    // Deprecated - This transaction type should not be used to create new transactions
```
ewaterman: aren't we only deprecating the product inventory?

grails-app/controllers/org/pih/warehouse/data/MigrationController.groovy
```diff
@@ -359,6 +371,20 @@ class MigrationController {
...
+    def migrateProductInventoryTransactions() {
+        long startTime = System.currentTimeMillis()
+        Location location = Location.get(session.warehouse.id)
+
+        boolean performMigration = params.boolean("performMigration", false)
+        Map results = migrationService.migrateProductInventoryTransactions(location, performMigration)
+
+        long responseTime = System.currentTimeMillis() - startTime
+        String responseTimeMessage = "${performMigration ? 'Migration' : 'Generating preview'} took ${(responseTime)} ms"
+        log.info "$responseTimeMessage"
```
ewaterman, line 384: I'd add a bit more context here to clarify exactly what is getting migrated so that future us aren't confused:
```
log.info("Migration ${performMigration ? '' : '(preview only)'} of product inventory transactions at location ${location.name}. took ${(responseTime)} ms")
```

grails-app/services/org/pih/warehouse/inventory/ProductInventoryTransactionMigrationService.groovy
```diff
@@ -0,0 +1,25 @@
...
+ * to the new Inventory Baseline and Adjustment pair
+ */
+@Transactional
+class ProductInventoryTransactionMigrationService extends ProductInventoryTransactionService<ImportDataCommand> {
+
+    ConfigService configService
+
+    @Override
+    void setSourceObject(Transaction transaction, ImportDataCommand sourceObject) {
+        // Inventory Import has no source object.
```
ewaterman: update comment. Also I don't know if there's a way to set a null source object but you probably don't want it to be ImportDataCommand

EWaterman (commented):
I haven't reviewed the migration itself yet. Will do that tomorrow. Can you please post a screenshot of what the migration page looks like for reference?

src/main/groovy/org/pih/warehouse/core/Constants.groovy
awalkowiak: shoot... wrong line, wanted to put it next to the `INVENTORY_TRANSACTION_TYPE_ID`

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                        it.transactionDate,
+                        "Migrated from single Product Inventory transaction to Baseline + Adjustment pair",
+                        null,
+                        // don't validate transaction date, there is old transaction at the same time, that will be removed
+                        false,
+                        true
+                )
+                if (baselineTransaction) {
+                    baselineTransaction.disableRefresh = true
+                    // Ugly workaround to omit auto-timestamping dateCreated
```
awalkowiak: @awalkowiak move it up and change to fixme, and link gh issue related / documentation

awalkowiak: and move it to a method

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                }
+
+                // 5. Delete the old transaction (with flush, to get rid of that transaction for next calculations)
+                it.disableRefresh = true
+                it.delete(flush: true, failOnError: true)
+            }
+
+            // Zero out stock for products that were migrated but had no stock initially (hence hand no baseline created)
+            List<Product> shouldBeZeroedOut = products - currentInventoryBaseline?.transactionEntries?.collect { it.inventoryItem.product }?.unique()
+            if (shouldBeZeroedOut) {
```
awalkowiak: @awalkowiak Make it configurable for testing

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                                "this transaction is created to ensure that products that had no stock before, " +
+                                "are currently zeroed out too"
+                )
+            }
+
+            // Refresh PA for all products that were migrated
+            productAvailabilityService.triggerRefreshProductAvailability(location.id, products?.id, true)
+        }
+
+        return [
```
awalkowiak, line 513: @awalkowiak consider to store this somewhere (additional log or csv)

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                Transaction adjustmentTransaction = createAdjustmentTransaction(
+                        location,
+                        it.transactionEntries,
+                        availableItems,
+                        adjustmentTransactionDate,
+                        "Migrated from single Product Inventory transaction to Baseline + Adjustment pair"
+                )
+                // Ugly workaround to omit auto-timestamping dateCreated
+                if (adjustmentTransaction?.id) {
+                    Date adjustmentCreatedDate = DateUtil.asDate(DateUtil.asInstant(it.dateCreated).plusSeconds(1))
```
awalkowiak: @awalkowiak add a fixme to standardize the "+1" second

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                // 3. Create inventory baseline for old transaction that is being migrated (if enabled)
+                Map<String, AvailableItem> availableItems = productAvailabilityService.getAvailableItemsAtDateAsMap(
+                        location, currentTransactionProducts, it.transactionDate)
+
+                Transaction baselineTransaction = productInventoryTransactionMigrationService.createInventoryBaselineTransactionForGivenStock(
+                        location,
+                        null,
+                        availableItems.values(),
+                        it.transactionDate,
+                        "Migrated from single Product Inventory transaction to Baseline + Adjustment pair",
```
awalkowiak: Add a old transaction number as a part of comment

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+
+                // 4. Create adjustment transactions for the old transaction that is being migrated
+                // Date objects are mutable, so we use Instant to clone the date in the command and avoid directly modifying it.
+                Date adjustmentTransactionDate = DateUtil.asDate(DateUtil.asInstant(it.transactionDate).plusSeconds(1))
+                Transaction adjustmentTransaction = createAdjustmentTransaction(
+                        location,
+                        it.transactionEntries,
+                        availableItems,
+                        adjustmentTransactionDate,
+                        "Migrated from single Product Inventory transaction to Baseline + Adjustment pair"
```
awalkowiak: Add a old transaction number as a part of comment

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                        false,
+                        true
+                )
+                if (baselineTransaction) {
+                    baselineTransaction.disableRefresh = true
+                    // Ugly workaround to omit auto-timestamping dateCreated
+                    // Needs to flush it first to be able to overwrite dateCreated
+                    baselineTransaction.save(flush: true)
+                    // Unfortunately, we need to use raw SQL to update the dateCreated field (otherwise it will be
+                    // auto-timestamped)
```
ewaterman: I mean, it is true that date_created is changing to be the time of the migration. It represents the creation of the row in the database. Are there any scenarios where we can't simply rely on transaction_date? I know creation date will be used if two transactions have the same transaction date, but does that actually happen in practice? We really shouldn't allow that to happen imo since it would likely lead to confusion

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -359,6 +399,128 @@ class MigrationService {
...
+                            """UPDATE transaction set date_created = ? WHERE id = ?""",
+                            [adjustmentCreatedDate, adjustmentTransaction.id]
+                    )
+
+                    recordMigrationResult(results, [adjustmentTransaction])
+                }
+
+                // 5. Delete the old transaction (with flush, to get rid of that transaction for next calculations)
+                it.disableRefresh = true
+                it.delete(flush: true, failOnError: true)
```
ewaterman, line 486: I imagine the slow part of this is all the saves/deletes with flushing but it'd be good to test it out somehow. If we could zero in on a specific piece of code that's really slow it'd probably be easier to refactor. You could even just run it locally after adding a bunch of log statements that time each block of code.

One option for speeding this process up is to simply collect a list of transaction ids to delete and then run a single, `DELETE FROM transaction WHERE id IN (${transactionIdsToDelete});` query at the end.

Another thought would be to take a totally different strategy for this migration and do it kinda like we do for building stock history.
- First, select all products at a location that have a product inventory transaction (like you already do).
- Then select ALL transactions for the products at the facility ordered by transaction date and creation date
- Then loop through the transactions, summing the quantity for the product until you hit a product inventory, in which case you'd:
  - add the transaction id to a list to delete
  - init (but don't save yet) a new adjustment and baseline transaction based on the product inventory transaction (like you already do) and add those to a list to be saved.
- Then once you finish looping, you have all the transaction ids that you need to delete, and all the new transactions that you need to save. You can do the deletes in a single query with a flush, then do the inserts (with no flush required).

grails-app/views/migration/dataMigration.gsp
awalkowiak: I'd say no need for it

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: Actually, @kchelstowski made some additional investigation locally, and the worst part is validation and transaction number generation. Even saving/deleting in batch does not really help here. We also discussed it a bit on yesterday's tech huddle, and for now, something that might help here would be running products in parallel (but there is an issue with transactions that have multiple products).

I am not sure about the looping over all transactions would have way better performance. Initially, I thought the issue was available items calculation, for which it could help, but it turned out it is not that bad (since the validation and transaction number are bigger issues and I think we cannot omit that part in this approach) :thinking:

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: We have to update the created date too to the original transaction's created date, if we don't want to treat all these transactions as backdated, since currently i think we have only one option to find backdated - compare these two dates (we would end up with transaction date being from the past and date created being the day that migration was run)

grails-app/conf/runtime.groovy
awalkowiak: Yeah I'm not sure either, I moved these to applicaiton.yml

PR #5373, opened 2025-07-08
OBPIH-7359 Items inventoried this month indicator issue 

alannadolny:


grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
```diff
@@ -102,19 +102,24 @@ class NumberDataService {
 
     @Cacheable(value = "dashboardCache", key = { "getItemsInventoried-${location?.id}" })
     NumberData getItemsInventoried(Location location) {
-        Date firstOfMonth = LocalDate.now().withDayOfMonth(1).toDate();
+        Date firstOfMonth = LocalDate.now().withDayOfMonth(1).toDate()
+        List<String> transactionTypeIds = [
```
awalkowiak, line 112:
```suggestion
        /**
        * After migrating from single product inventory transaction to baseline + adjustment transactions
        * there is a case when we might have only an adjustment if it's first inventory record
        * and we can't rely on CREDIT type code for this case and ADJUSTMENT_CREDIT_TRANSACTION_TYPE_ID
        * is only used for baseline + adjustment and for manual stock adjustment
        */
        List<String> transactionTypeIds = [
```

PR #5378, opened 2025-07-13
OBPIH-7313 Build view for Inventory Transactions Report to contain da…

kchelstowski:
…ta for all inventory transactions

**Description:**

---


grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,110 @@
...
+            return VarianceTypeCode.LESS
+        }
+        return VarianceTypeCode.EQUAL
+    }
+
+    /**
+     * Transaction type display name used in the "Inventory Transactions" table
+     * possible options are: CC (Cycle count transaction), IA (Inventory adjustment), RS (Record stock), II (Inventory import)
+     */
+    String getTransactionTypeDisplayName() {
```
kchelstowski: haven't figured out a better name for that one, do you guys have an idea?

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,110 @@
...
+        if (transaction.cycleCount) {
+            return "CC"
```
kchelstowski: I considered creating a constant for that but felt like it's not that necessary as there is a small change for this one to be reused. Do you mind it being hardcoded?

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (transaction.comment?.contains("Imported from")) {
```
kchelstowski, line 76: this might feel tricky, but I consulted this with Manon. Since we don't actually have any way to distinguish the record stock and II, the only thing that currently is different while importing is the comment that is added. 
We are aware that if someone adds such comment manually while recording stock, this would be treated as II, but Manon said it was not a big deal if it happened.
One more possible solution would be to look at distinct products in the entries (if > 1, then import), but again - someone might import only one item and this would be treated as record stock.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -1033,6 +1033,11 @@ class UrlMappings {
             action = [POST: "getInventoryAuditSummary", GET: "getInventoryAuditSummary"]
         }
 
+        "/api/reports/inventory-transactions-summary" {
+            controller = "cycleCountApi"
```
ewaterman: Why in cycle count? Won't this eventually encompass record stock and inventory import transactions as well? Seems to me like the logic should be in more general InventoryTransactionSummary (or whatever name) controller and service classes

grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -0,0 +1,68 @@
...
+              facility.id as facility_id,
+              transaction.id as transaction_id,
+              transaction.transaction_date as baseline_transaction_date,
+              SUM(transaction_entry.quantity) as quantity_balance
+            FROM transaction_entry
+            JOIN transaction ON transaction.id = transaction_entry.transaction_id
+            JOIN transaction_type ON transaction.transaction_type_id = transaction_type.id
+            JOIN location facility ON transaction.inventory_id = facility.inventory_id
+            JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
+            WHERE transaction_type.id = '12' -- baseline inventory transaction
```
ewaterman, line 14: I don't like such a heavy reliance on baseline transactions because we can configure cycle count, inventory import, and record stock to not create a baseline transaction and only do adjustments.

However, like we discussed in the tech huddle, this is fine for now until we can properly create a way for the adjustment transactions of record stock and inventory import to reference back to the feature that created them (possibly via a 'document' like Justin has been mentioning).

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,110 @@
...
+     * Root causes are only captured for the cycle count
+     */
+    List<String> getRootCauses() {
+        if (transaction.cycleCount) {
+            return transaction
+                    .cycleCount
+                    .cycleCountItems
+                    .discrepancyReasonCode
+                    // filter out null/empty
+                    .findAll { it }
```
ewaterman, line 46: do we need to filter out duplicates?

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
ewaterman: I think these should be localizable since other languages will translate "cycle count" to something else

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
ewaterman: I prefer leaving it to the frontend to handle all display related fields. The fact that we want to display an abbreviation is a frontend decision. It's not a property of the transaction summary itself. A different frontend might want to display the full name.

Ultimately I'm fine with the solution you have but I'd prefer to shift it to the frontend.

You could make an enum like:
```
enum TransactionAction {
    CYCLE_COUNT,
    RECORD_STOCK,
    INVENTORY_IMPORT,
}
```
representing the action that was taken to create the transaction. Then instead of returning `transactionTypeDisplayName`, return an `action` field and let the frontend decide what to do with that (ex: if action==='CYCLE_COUNT' then display 'CC').

This hilights to me that we need an easier way in general to determine what feature was used to create the transaction. So you could take this even further if you wanted and put this logic into an `action` transient field in Transaction:
```
class Transaction {
   static transients = ['action']
    TransactionAction action
    ...
    TransactionAction getAction() {
        if (transaction.cycleCount) {
            return TransactionAction.CYCLE_COUNT
        }
        ...
    }
}
```
and then this method can be removed entirely (and in toJson simply return the 'action' of the transaction).

No need to do that here if that feels like too much for one PR, but we can discuss.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
ewaterman: Yeah this is really gross but I suppose its okay for now. Can you at least add a TODO or FIXME comment indicating we need to find a better solution.

I'm going to make a discovery task for us to add a way to identify the feature/action that created the transaction. Either a 'document' like Justin has been talking about or making the 'action' field like I mention in my other comment an actual field in the domain that gets set when the transaction is created.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
kchelstowski: I'll give it a shot tomorrow, thanks for the suggestion

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
kchelstowski: Yeah, I left it in the cycle count, because this is where the report is accessible from, from the UI, but I also felt it was not quite right anymore, but I didn't really have an idea which controller and service I should use here. 
I'll probably create a separate one for the InventoryTransactionSummary, but I'm open for other suggeestions

grails-app/migrations/views/inventory-transactions-summary.sql
kchelstowski: Actually we left join the "helper view" with the baseline transactions, so it's not that heavily relying on the baselines.
The problem I described today was rather for the adjustments without baselines, but with some baselines existing for another adjustment. Since we were searching for the latest baseline, it was joining it anyway, so what I did is that I join baseline only if no other adjustment was created between our adjustment and the baseline we are joining to.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
kchelstowski: good point, I need to check it out how it currently handles those. If so, then I'll just make it a set probably.

grails-app/controllers/org/pih/warehouse/api/InventoryTransactionSummaryApiController.groovy
```diff
@@ -0,0 +1,21 @@
...
+import grails.validation.ValidationException
+import org.pih.warehouse.inventory.InventoryTransactionSummaryService
+import org.pih.warehouse.inventory.InventoryTransactionsSummary
+import org.pih.warehouse.report.CycleCountReportCommand
+
+class InventoryTransactionSummaryApiController {
+
+    InventoryTransactionSummaryService inventoryTransactionSummaryService
+
+    def getInventoryTransactionsSummary(CycleCountReportCommand command) {
```
ewaterman, line 13: can you rename CycleCountReportCommand to something else now that it's not just cycle count related?

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,119 @@
...
+                quantityAfter: quantityAfter,
+                quantityDifference: quantityDifference,
+                dateRecorded: dateRecorded,
+                recordedBy: [
+                    name: recordedBy.name,
+                ],
+                rootCauses: rootCauses,
+                comments: comments,
+                varianceTypeCode: varianceTypeCode?.name(),
+                transactionAction: getTransactionActionDisplayName(transactionAction),
```
ewaterman: I was thinking that this would be just:
```
transactionAction: transactionAction
```
and then do the translation to display name in useInventoryTransactionsTab.jsx:
```
header: () => (
  ...
),
cell: ({ getValue }) => {
  const value = {translate('"enum.TransactionAction.${getValue()}"', '?')};
  return (
    <TableCell
      className="rt-td pb-0 d-flex align-items-start"
    >
      <ValueIndicator
        variant={valueIndicatorVariant.TRANSACTION}
        value={value}
      />
    </TableCell>
  );
```
That way the backend isn't telling the frontend what value to display. The frontend gets to decide for itself.

kchelstowski: This is why I didn't make the `transactionAction` translated, but added an additional `transactionActionDisplayName`. I know there was a reason to make translations on backend more, than on the frontend, it was probably crowdin related thing, but I can't remind myself what it was. 
I would have to try to test your version to see if it works fine.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
jmiranda: Let's move this to the transaction and the logic should be 

```
if cycle count TransactionAction.CYCLE_COUNT
def distinctProductCount = transactionEntries.collect { it.inventoryItem.product }.unique()  
if distinctProductCount == 1 return TransactionAction.RECORD_STOCK
else if distinctProductCount > 1 return TransactionAction.INVENTORY_IMPORT
return TransactionAction.INVENTORY_ADJUSTMENT
```

kchelstowski: We discussed in the tech huddle the reasoning behind this, so I'm marking this conversation as resolved for a better visibility.

grails-app/controllers/org/pih/warehouse/api/InventoryTransactionSummaryApiController.groovy
jmiranda: Or better yet create a separate command class (TransactionReportCommand). The dateRecorded property is not the same as the transactionDate so it's probably fine to have a separate command class for this one since they are not the same, even though they have similar properties.

One other small change I would make would be to keep the name singular InventoryTransactionSummary, inventory_transaction_summary, etc.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,111 @@
...
+
+    Transaction transaction
+
+    Product product
+
+    Location facility
+
+    Integer quantityBefore
+
+    Integer quantityAfter
```
jmiranda, line 20: A slightly more descriptive naming convention like quantityOnHandBefore, quantityOnHandAfter might be better here.

grails-app/domain/org/pih/warehouse/inventory/InventoryTransactionsSummary.groovy
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    VarianceTypeCode getVarianceTypeCode() {
+        if (quantityDifference > 0) {
+            return VarianceTypeCode.MORE
+        }
+        if (quantityDifference < 0) {
+            return VarianceTypeCode.LESS
+        }
+        return VarianceTypeCode.EQUAL
```
jmiranda, line 64: Not sure how people feel about the elvis operator, but I would love to see something more concise.
```
VarianceTypeCode getVarianceTypeCode() {
    quantityDifference > 0 ? VarianceTypeCode.MORE :
    quantityDifference < 0 ? VarianceTypeCode.LESS :
    VarianceTypeCode.EQUAL
}
```
Also, we're seeing this same logic in a bunch of places so it might make more sense to include it in a method on the enum itself.
```
enum VarianceTypeCode {
    MORE, LESS, EQUAL

    static VarianceTypeCode fromQuantityVariance(BigDecimal quantityVariance) {
        quantityVariance > 0 ? MORE : (quantityVariance < 0 ? LESS : EQUAL)
    }
}
```
And then when you want to convert the quantity difference to an enum, you use the method on the enum.
```
VarianceTypeCode varianceTypeCode = VarianceTypeCode.fromQuantityVariance(quantityVariance)
```

jmiranda (commented):
Nice work, @kchelstowski. I made some comments that could be addressed, but I'll leave the approval to @awalkowiak and @ewaterman.

PR #5380, opened 2025-07-15
OBPIH-7370 button for skipping count assignment

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7370

**Description:**
- Add 'skip' button when assigning a count after a count with discrepancies
- Doesn't show the recount module if 'not now' is clicked after a count with discrepancies

---

Screenshot from 2025-07-14 15-54-48

### DIFF
--- a/src/js/hooks/cycleCount/useCountStep.jsx
+++ b/src/js/hooks/cycleCount/useCountStep.jsx
@@ -43,8 +43,6 @@ import { checkBinLocationSupport } from 'utils/supportedActivitiesUtils';
 const useCountStep = () => {
   const [isAssignCountModalOpen, setIsAssignCountModalOpen] = useState(false);
   const assignCountModalData = useRef([]);
-  // Page on which user should land after closing assign cycle count modal
-  const redirectAfterClosingModal = useRef(null);
   // Table data is stored using useRef to avoid re-renders onBlur
   // (it removes focus while selecting new fields)
   const tableData = useRef([]);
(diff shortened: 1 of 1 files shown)


src/js/hooks/cycleCount/useCountStep.jsx
```diff
@@ -469,8 +469,7 @@ const useCountStep = () => {
         }
         hide();
         onClose()?.();
-        redirectAfterClosingModal.current = CYCLE_COUNT.list(TO_RESOLVE_TAB);
-        openAssignCountModal();
+        history.push(CYCLE_COUNT.list(TO_RESOLVE_TAB));
```
alannadolny, line 470: Is the `redirectAfterClosingModal` necessary in other places after that change? I think it can be removed.

ewaterman: it's still used to redirect to the resolve flow after you finish with the assignment when you click "resolve now". But you added this logic so maybe you're understanding something that I'm not

alannadolny: usages of `redirectAfterCloisingModal` in that file:
```
  // Page on which user should land after closing assign cycle count modal
  const redirectAfterClosingModal = useRef(null);
```
```
    history.push(redirectAfterClosingModal.current);
```
and
```
        redirectAfterClosingModal.current = CYCLE_COUNT.resolveStep();
```
because the `redirectAfterClosingModal` can have only one value at this moment (redirect to the resolve step), so I think you can directly use:
```
    history.push(CYCLE_COUNT.resolveStep());
```

alannadolny (approved):

PR #5381, opened 2025-07-15
OBPIH-7314 Include all inventory transaction in inventory audit report (also fixes OBPIH-7373, OBPIH-7325)

jmiranda:


grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1163,24 +1165,34 @@ class ReportService implements ApplicationContextAware {
...
 
             Location facility = (Location) it[0]
             Product product = (Product) it[1]
             String abcClass = it[2]
+            Integer quantityAdjusted = it[3]
+            Integer countAdjustments = it[4]
+
+            // Retrieve all product inventories completed during the given date range
+            List<TransactionType> inventoryTypes = TransactionType.findAllByTransactionCode(TransactionCode.PRODUCT_INVENTORY)
+            Integer countCycleCounts = TransactionEntry.countByTransactionTypes(facility, product, inventoryTypes, command.startDate, command.endDate).get()
```
jmiranda, line 1180: Reminder: As discussed on the tech huddle, this is not the most accurate way to calculate "number of counts", but it's the best we can do at the moment. Once we either (a) add a baseline for every count (record stock, inventory import, cycle count) OR (b) add a reference document for each adjustment transaction that has a reference document. Note: standalone adjustments (using Adjust Stock feature) probably won't have a reference document.

The former solution (a) still has the problem that we'd be counting baseline transactions that might not be associated with a count. So we still need (b) to make it work properly.

alannadolny (approved):

PR #5389, opened 2025-07-17
OBS-1876 Add HS Code Product Attribute to Commercial Invoice doc

awalkowiak:
https://pihemr.atlassian.net/browse/OBS-1876


grails-app/services/org/pih/warehouse/core/FileService.groovy
```diff
@@ -177,10 +178,10 @@ class FileService {
             def manufacturerName = manufacturerNames ? manufacturerNames.first() : ''
 
             return [
+                    hsCode           : shipmentItem.inventoryItem?.product?.attributes?.find { ProductAttribute pa -> pa.attribute.code == Constants.HS_CODE_PRODUCT_ATTRIBUTE_CODE }?.value ?: "",
```
ewaterman: You could make this a convenience method in Product if you wanted:
```
ProductAttribute getAttribute(String attributeCode) {
    return attributes?.find { ProductAttribute pa -> pa.attribute.code == attributeCode }
}
```
but the way you have it is fine.

ewaterman (approved):

grails-app/services/org/pih/warehouse/core/FileService.groovy
awalkowiak: Actually there is already `getProductAttribute` :see_no_evil: I am going to use it

alannadolny (approved):

PR #5393, opened 2025-07-18
OBPIH-7376 Errors encountered when using Record Stock to create or update inventory

alannadolny:


grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -584,21 +584,22 @@ class InventoryItemController {
         }
 
         log.info("User chose to validate or there are errors")
-        def warehouseInstance = Location.get(session?.warehouse?.id)
-
+        Location warehouseInstance = Location.get(session?.warehouse?.id)
+        commandInstance.product = Product.load(commandInstance.product.id)
```
ewaterman, line 589: why is this needed? Was the product in the command not getting initialized properly?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -2708,7 +2707,9 @@ class InventoryService implements ApplicationContextAware {
                 }
             }
             transaction {
-                lt("transactionDate", date)
+                // We are able to select dates with the precision set to minutes,
+                // so we should have at most one transaction within the same minute
+                lt("transactionDate",  new Date(date.time + 60_000))
```
ewaterman: I don't think this is what we want to do universally. In cycle count we use current timestamp for transaction date, which is precise to the second. So this logic is overly restrictive for that case.

But also, if a date picker only gives minute level precision, any transaction date that uses that picker within the same minute should get a date that is the start of the minute. I don't see why we need to consider any other times within that minute unless we're worried about other flows changing the same products at the same time that have precision to the second.

Is the issue not just that we're missing a baseline if there is no stock previously so this duplicate check doesn't fail? Can we not just change the logic to create the baseline for the default lot and bin with quantity 0?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -2733,11 +2734,19 @@ class InventoryService implements ApplicationContextAware {
...
-                eq("transactionDate", date)
+                // Catch transactions that are created within the same minute
+                // We are saving the adjustment transaction one second after
+                // inventory baseline, so to get that transaction, we should
+                // match transaction based on the time range.
+                // For a date: 18.07.2025 15:28:00 we are matching transactions
+                // that are created: 18.07.2025 15:28:00 <= date < 18.07.2025 15:28:59
+                // We are able to select dates with the precision set to minutes,
+                // so we should have at most one transaction within the same minute
+                ge("transactionDate", date)
```
ewaterman: You can use `between` I think to make this a single line.

https://docs.grails.org/latest/ref/Domain%20Classes/createCriteria.html

But I still don't think this is necessary. See my other comment

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
alannadolny: There was a lazy initialization exception error. It was not visible before, because some of the errors were not properly handled.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1484,12 +1486,17 @@ class InventoryService implements ApplicationContextAware {
...
                         null,
                         availableItems.values() as List<AvailableItem>,
-                        cmd.transactionDate
+                        cmd.transactionDate,
+                        null,
+                        [:],
+                        true,
+                        false,
+                        // Check if there are adjustments created for that specific time
+                        adjustmentTransactionDate
```
ewaterman: If we're at time 00:00:00, this will correctly fail if there is a transaction at 00:00:01 (ie an adjustment) but not fail if there's a transaction at 00:00:00 (ie there's already a baseline that had no adjustment, ex: a CC with no discrepancies).

So if you want to use your logic, you need to check both dates: the given date (ie bseline) and the given date + 1 (ie adjustment).

But I think simpler would be to add an extra check when creating the adjustments to make sure that there's no transaction at that time. That way baseline can check its dates and adjustment can check its dates. The logic doesn't get intermingled.

That's what I did for inventory import: https://github.com/openboxes/openboxes/blob/develop/grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy#L180

Essentially all you should need to do is add the following to `recordStockProductInventoryTransactionService.createAdjustmentTransaction`:
```
// We'd have weird behaviour if we allowed two transactions to exist at the same exact time (precision at the
// database level is to the second) so fail if there's already a transaction on the items for the given date.
if (inventoryService.hasTransactionEntriesOnDate(facility, transactionDate, inventoryImportData.inventoryItems)) {
    throw new IllegalArgumentException("A transaction already exists at time ${transactionDate}")
}
```
And let the baseline validation work like it was before.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1384,9 +1384,26 @@ class InventoryService implements ApplicationContextAware {
...
-                    cmd.transactionDate
+                    new Date(cmd.transactionDate.time + 2000)
             )
 
+            // 1. Create the baseline transaction
+            if (isInventoryBaselineEnabled) {
+                // We'd have weird behaviour if we allowed two transactions to exist at the same exact time (precision at the
+                // database level is to the second) so fail if there's already a transaction on the items for the given date.
+                List<InventoryItem> inventoryItems = availableItems.values().collect{ it.inventoryItem }
+                if (hasTransactionEntriesOnDate(currentLocation, adjustmentTransactionDate, inventoryItems)) {
```
ewaterman: This is looking closer to what I was expecting but still a question:

Why put this check for hasTransactionEntriesOnDate inside the isInventoryBaselineEnabled block? Shouldn't we do it even if baseline transactions are disabled?

That's why I was suggesting to put it in `recordStockProductInventoryTransactionService.createAdjustmentTransaction`. That way we only check for duplicates at adjustmentTransactionDate if we actually have an adjustment transaction to create.

alannadolny: I moved the validation outside the isInventoryBaselineEnabled block. I am not moving it to the `createAdjustmentTransaction` function because the code shouldn't be executed when there is a validation error. So, I am just validating it before proceeding deeper into that function.

ewaterman: > because the code shouldn't be executed when there is a validation error

This is all done in a transaction so if an exception is thrown won't it be rolled back?

If you're curious, you can look at how I did it in InventoryImportDataService.createAdjustmentTransaction. That seems to work fine, which is why I was suggesting you do it that way.

It's not a big deal though so if you prefer to do it here that's fine. However now you have the opposite problem: Even if you won't have any adjustment transaction, you still fail validation if a transaction exists at that time. This is almost never going to happen though so I'm not really worried about it.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1384,9 +1384,26 @@ class InventoryService implements ApplicationContextAware {
             Map<String, AvailableItem> availableItems = productAvailabilityService.getAvailableItemsAtDateAsMap(
                     currentLocation,
                     [cmd.product],
-                    cmd.transactionDate
+                    new Date(cmd.transactionDate.time + 2000)
```
ewaterman, line 1387: do you need to do this? See my comment here: https://github.com/openboxes/openboxes/blob/develop/grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy#L171-L177

alannadolny: Yeah, because while pulling data in the `getAvailableItemsAtDateAsMap`, we are looking for transaction entries that are newer than the passed date, so:
- transaction date: 17:14:00,
- adjustment date: 17:14:01,
- date in filtering: 17:14:02,
So having it like that, I can get the transaction entry, because adjustment date < date in filtering. Without that, it will be filtered out because adjustment date > transaction date.
The code that is responsible for that filtering:
```
    List<TransactionEntry> getTransactionEntriesBeforeDate(Location facility, Collection<Product> products, Date date) {
        if (!date) {
            return []
        }

        return criteria.list {
            if (products) {
                inventoryItem {
                    'in'("product", products)
                }
            }
            transaction {
                lt("transactionDate", date)
                eq("inventory", facility?.inventory)
                order("transactionDate", "asc")
                order("dateCreated", "asc")
            }
        } as List<TransactionEntry>
    }
```
The `lt` is responsible for that.

ewaterman: hmm okay that makes sense. I wonder if getTransactionEntriesBeforeDate should actually be using `le("transactionDate", date)` then? I know I was the one who wrote this method but I don't remember why I did it with `lt`.

I'm okay with you merging this as is just so that it can get to testing but I think it's worth discussing because we do it slightly differently in inventory import. Lets talk about it tomorrow during tech huddle.

jmiranda: I don't understand this well at the moment so I'd love to have a discussion on the tech huddle tomorrow. If possible, it would be great to turn all of the +1s +2s into constants that provide an explanation for why they're required.

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1384,9 +1384,26 @@ class InventoryService implements ApplicationContextAware {
...
+            // We'd have weird behaviour if we allowed two transactions to exist at the same exact time (precision at the
+            // database level is to the second) so fail if there's already a transaction on the items for the given date.
+            List<InventoryItem> inventoryItems = availableItems.values().collect{ it.inventoryItem }
+            if (hasTransactionEntriesOnDate(currentLocation, adjustmentTransactionDate, inventoryItems)) {
+                throw new IllegalArgumentException("A transaction already exists at time ${adjustmentTransactionDate}")
+            }
+
+            // 1. Create the baseline transaction
+            if (isInventoryBaselineEnabled) {
+                recordStockProductInventoryTransactionService.createInventoryBaselineTransactionForGivenStock(
```
jmiranda, line 1399: I'm not loving the name of this service. RecordStockService? BaselineTransactionService? TransactionService?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
awalkowiak: @jmiranda we already have a to do for this topic in other place, @alannadolny could you create a ticket for that?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
awalkowiak: @jmiranda this was named like that based on the `ProductInventoryTransactionService` used by Cycle Count when it was still creating "Product Inventory" transaction, after refactor from Product Inventory to Baseline + Adjustment these were not changed. @alannadolny Could you create a ticket to review all those names, and rename where needed?

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
awalkowiak: I don't like the +2, feels like this should be handled by a LE, instaed LT with increment by 1. @ewaterman is it possible to figure out what was a cause behind this? Why you used LT? Or it's hard to figure out now. Either way, i think we should create a ticket for this refactor, because this is used by few other places, so it's safer to tackle as a separate task after the release (@alannadolny could you create a ticket for this?)

alannadolny: @awalkowiak 
1. https://pihemr.atlassian.net/browse/OBPIH-7422?atlOrigin=eyJpIjoiZGZkMzI3ZDI2MTEwNDhkYTlkMWU3M2UzOTkzYzdlODAiLCJwIjoiaiJ9
2. https://pihemr.atlassian.net/browse/OBPIH-7423?atlOrigin=eyJpIjoiNDFjODM1ODU5OWU0NGU4NmFiOWNhZjRhMTc4ZmNmMGYiLCJwIjoiaiJ9

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
alannadolny: @awalkowiak 
https://pihemr.atlassian.net/browse/OBPIH-7424

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
jmiranda, line 1387:
```
new Date(cmd.transactionDate.time + Constants.AWESOME_NAME_THAT_EXPLAINS_EVERYTHING
```
Consttants.groovy
```
-- Big description of why we're doing this.
AWESOME_NAME_THAT_EXPLAINS_EVERYTHING = 2000
```

PR #5399, opened 2025-07-22
OBPIH-7322 change products inventoried to use items inventoried indic…

ewaterman:
…ator logic

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7322

**Description:** Change the logic for the "products inventoried" indicator to use the same logic as the "items inventoried" dashboard indicator (as was suggested by Manon). Kacper already refactored the dashboard indicator to account for the baseline + adjustment refactor so this seems safe to do. We want to display the same data in both indicators.

---

Screenshot from 2025-07-22 11-09-32

### DIFF
--- a/grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
@@ -4,6 +4,8 @@ import grails.core.GrailsApplication
 import grails.plugin.cache.Cacheable
 import org.grails.plugins.web.taglib.ApplicationTagLib
 import org.joda.time.LocalDate
+
+import org.pih.warehouse.DateUtil
 import org.pih.warehouse.api.StockMovementDirection
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.Location
@@ -103,6 +105,18 @@ class NumberDataService {
     @Cacheable(value = "dashboardCache", key = { "getItemsInventoried-${location?.id}" })
     NumberData getItemsInventoried(Location location) {
         Date firstOfMonth = LocalDate.now().withDayOfMonth(1).toDate()
+        return getItemsInventoriedInRange(location, firstOfMonth)
+    }
+
+    /**
+     * Fetch the count of distinct products that have been inventoried in the given time range.
+     * "Inventoried" means any operation that performs a full quantity count for the product.
+     *
+     * @param location The facility that we want the count at.
+     * @param startDate The datetime in the past to start the check at. If null, will use the start of time.
+     * @param endDate The datetime in the past to check up until. If null, will use the current time.
+     */
+    NumberData getItemsInventoriedInRange(Location location, Date startDate=null, Date endDate=null) {
         /**
         * After migrating from single product inventory transaction to baseline + adjustment transactions
         * there is a case when we might have only an adjustment if it's first inventory record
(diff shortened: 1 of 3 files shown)


alannadolny (approved):

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
```diff
@@ -121,14 +135,15 @@ class NumberDataService {
...
             WHERE t.inventory = :inventory
             AND t.transactionType.id IN :transactionTypeIds 
-            AND t.transactionDate >= :firstOfMonth""",
+            AND t.transactionDate BETWEEN :startDate AND :endDate""",
                 [
                         inventory          : location?.inventory,
                         transactionTypeIds : transactionTypeIds,
-                        firstOfMonth       : firstOfMonth,
+                        startDate          : startDate ?: DateUtil.EPOCH_DATE,
+                        endDate            : endDate ?: new Date(),
```
jmiranda, line 143: I like the idea here (re: defaults) but we should confirm that these are good "expected" values for start and end dates. Regardless, we either need default values OR logic to handle NULLs. 

I had been thinking about this when handling some of the reporting use cases. I don't love the excessive if/then/else conditions for when start date and/or end date are NULL, so I'd be very happy if we could always use BETWEEN for these queries as long as we can come up with sensible defaults.

grails-app/services/org/pih/warehouse/dashboard/NumberDataService.groovy
```diff
@@ -121,14 +135,15 @@ class NumberDataService {
...
                 ])
 
-        return new NumberData(itemsInventoried[0])
+        return new NumberData(itemsInventoried[0] as Double)
```
jmiranda, line 146: I've been mindlessly using BigDecimal for any floating-point numbers. Do you have a justification for using Double over BigDecimal? Or are we just picking our favorite floating-point class at this point?

grails-app/utils/org/pih/warehouse/DateUtil.groovy
```diff
@@ -54,6 +54,12 @@ class DateUtil {
     static final DateTimeFormatter DISPLAY_DATE_FORMATTER = DateTimeFormatter.ofPattern("dd/MMM/yyyy")
     static final DateTimeFormatter DISPLAY_DATE_TIME_OFFSET_FORMATTER = DateTimeFormatter.ofPattern("dd/MMM/yyyy HH:mm XXX")
     static final DateTimeFormatter DISPLAY_TIME_FORMATTER = DateTimeFormatter.ofPattern("HH:mm:ss")
+
+    /**
+     * A Date representing the epoch instant, which is January 1, 1970, 00:00:00 GMT.
+     */
+    static final Date EPOCH_DATE = new Date(0)
```
jmiranda, line 61: Wow, after all of that ZonedDateTime and LocalDateTime discovery work ... you're just going to `new Date(0)` us? 

I'm kidding. I have no opinion or even concept of what could/should be used here. And given the meaninglessness of timezones re: the epoch date, this seems like a fine approach.

Side note: Seeing `new Date(0)` reminded me of the line from the SNL Papyrus skit (https://www.youtube.com/watch?v=jVhlJNJopOQ)
"like a thoughtless child wandering by a garden, yanking leaves along the way"

jmiranda (approved):
Looks good.

PR #5401, opened 2025-07-23
OBPIH-7373 Fix calculation of number of counts and fix the last count…

kchelstowski:
…ed date issue on Products report

**Description:**

---


grails-app/migrations/views/counts.sql
```diff
@@ -0,0 +1,115 @@
...
+          )
+    )
+
+-- In the end make a union of all three subviews - since we separated all possible cases (baseline + adjustment, adjustment, baseline)
+-- we can be sure that no duplicates will appear
+SELECT * FROM baseline_with_adjustments
+UNION
+SELECT * FROM adjustment_without_baseline
+UNION
+SELECT * FROM baseline_without_adjustment
```
kchelstowski: The idea behind this view is:

- we first collect all the baselines + adjustments
- then we collect "alone" adjustments and "alone" baselines - we needed to do the first step before, so that we can easily filter out the baseline + adjustment that are in pair.
- in the end we make a union of three subviews

I added some additional, helper properties for a better visibility even though they are not used now.
Those are:

- `transaction_profile` - I couldn't figure out a better name for that, but basically it just indicates whether a count is: baseline + adjustment, alone adjustment or alone baseline
- `associated_transaction_id` - if we were to use this view for some other stuff, since the `baseline_with_adjustments`' base is the baseline transaction (the `transaction_id` would be the baseline one), I wanted a way to know which adjustment it is associated with, so in case baseline + adjustment, the `associatedTransaction` must not be `null`. For other two cases, it is supposed to be `null`.

grails-app/domain/org/pih/warehouse/inventory/TransactionEntry.groovy
```diff
@@ -58,7 +59,11 @@ class TransactionEntry implements Comparable, Serializable {
             }
             eq 'inventoryItem.product', product
             eq 'transaction.inventory', facility.inventory
-            eq "transactionType.transactionCode", TransactionCode.PRODUCT_INVENTORY
+            inList "transactionType.id", [
+                    Constants.ADJUSTMENT_CREDIT_TRANSACTION_TYPE_ID,
+                    Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID,
+                    Constants.INVENTORY_BASELINE_TRANSACTION_TYPE_ID
+            ]
```
ewaterman: I've seen this list in a few places now so I think it's worth putting in a constant. I'm not sure what to call it though.

InventoryTransactionTypes? BaseliningTransactionTypes? ProductCountTransactionTypes?

src/main/groovy/org/pih/warehouse/inventory/TransactionProfile.groovy
```diff
@@ -0,0 +1,12 @@
+package org.pih.warehouse.inventory
+
+enum TransactionProfile {
```
ewaterman: This name is fine, if a little vague.

Brainstorming some alternatives:
- Outcome
- Format
- Structure
- Composition

Personally I like "Composition". Maybe one of:
- BaseliningTransactionsComposition
- CountTransactionsComposition
- TransactionComposition

grails-app/domain/org/pih/warehouse/inventory/Count.groovy
```diff
@@ -0,0 +1,32 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class Count implements Serializable {
```
ewaterman: Maybe ProductCount since it's product based? I'm not sure what the best name is here, but Count feels a little too vague IMO

jmiranda: See comment below.

~~I think we can just call it InventoryCount or InventoryCountTransaction since it's not always going to be related to a product.~~

~~For example, I might~~
```
InventoryCount.findAllByProduct(product)
InventoryCount.findAllByFacility(facility)
InventoryCount.findAllByFacilityAndMonth(facility, startDate, endDate)
```

~~On the tech huddle, I brought up the fact that we might want to think through how we might use this class i.e. would it ever make sense to return records at the inventory item level. If so we could potentially let that become InventoryCountDetails (we don't need to implement it today) and this view could be InventoryCountRollup (transaction-level).~~

grails-app/domain/org/pih/warehouse/inventory/TransactionEntry.groovy
kchelstowski: I skipped this part for now since we don't have an exact idea how to name it + this would not be used at all for now - I just refactored it in case someone used that named query in the future.

grails-app/domain/org/pih/warehouse/inventory/Count.groovy
kchelstowski: I renamed it to ProductCount. I like it :+1:

grails-app/domain/org/pih/warehouse/inventory/ProductCount.groovy
```diff
@@ -0,0 +1,32 @@
+package org.pih.warehouse.inventory
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.product.Product
+
+class ProductCount implements Serializable {
```
jmiranda: Sorry, I was trying to get my comment in before you made any changes. 

I think we can just call it InventoryCount, InventoryCountRollup, or InventoryCountTransaction since the granularity is not always going to be at the product-level as far as I can tell. 

On the tech huddle, I mentioned that we might want to think through how we might use this class (and other similar views) more broadly.

For example, I might want to get counts based on a product at a facility over time or all counts within a facility between dates (using the findAll just to illustrate the different granularity).
```
InventoryCount.countByFacilityAndDateRecordedBetween(facility, startDate, endDate)
InventoryCount.countByFacilityAndProductAndDateRecordedBetween(facility, product, startDate, endDate)
```

i.e. would it ever make sense to return records at the inventory item level. If so we could potentially let that view become InventoryCountDetails (we don't need to implement it today) and this view could be InventoryCountRollup (facility + transaction-level + product).

jmiranda: One other caveat is that this view only deals with counts that have created transactions. It's possible that there are edge cases in which a cycle count has been completed, but doesn't create any transactions i.e. start cycle count when product is in stock, someone removes all inventory items, count does not add adjustments. This seems fairly benign (and unlikely), but just wanted to mention it.

grails-app/domain/org/pih/warehouse/inventory/TransactionEntry.groovy
jmiranda: InventoryCountTransactionTypes and I would like to make this a configuration. 

Also, now that I'm seeing them together

```
Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID,
Constants.INVENTORY_BASELINE_TRANSACTION_TYPE_ID
```
I think we should eventually change the INVENTORY_BASELINE_TRANSACTION_TYPE_ID to BASELINE_INVENTORY_TRANSACTION_TYPE_ID for consistency. Doesn't need to happen in this PR.

grails-app/migrations/views/changelog.xml
```diff
@@ -164,4 +164,7 @@
   <changeSet id="1748970410098-5" author="kchelstowski" runOnChange="true" runAlways="true" failOnError="true">
     <sqlFile path="views/inventory-transactions-summary.sql"/>
   </changeSet>
+  <changeSet id="1748970410098-6" author="kchelstowski" runOnChange="true" runAlways="true" failOnError="true">
+    <sqlFile path="views/product_counts.sql"/>
```
jmiranda: inventory-count.sql 

And use dashes for the filename, not underscores.

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+    -- First, just get all of the baseline transactions and adjustment transactions separately, so we can join them later
+    WITH inventory_baseline_candidates AS (
+        SELECT t.id AS transaction_id,
+               ii.id AS inventory_item_id,
+               t.transaction_date,
+               t.inventory_id
+        FROM transaction t
+        JOIN transaction_entry te ON te.transaction_id = t.id
+        JOIN inventory_item ii ON ii.id = te.inventory_item_id
+        WHERE t.transaction_type_id = '12'
```
jmiranda: Add comment re: tranasction type name

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+    ),
+    adjustments_candidates AS (
+        SELECT t.id AS transaction_id,
+               ii.id AS inventory_item_id,
+               t.transaction_date,
+               t.inventory_id
+        FROM transaction t
+        JOIN transaction_entry te ON te.transaction_id = t.id
+        JOIN inventory_item ii ON ii.id = te.inventory_item_id
+        WHERE t.transaction_type_id = '3'
```
jmiranda: Add comment re: transaction type name

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+    -- Adjustment transaction is hardcoded to be created 1 second after the baseline, so we can rely on TIMESTAMPDIFF
+    baseline_adjustment_matches AS (
+        SELECT
+            a.transaction_id AS adjustment_id,
+            b.transaction_id AS baseline_id
+        FROM adjustments_candidates a
+        JOIN inventory_baseline_candidates b
+          ON a.inventory_item_id = b.inventory_item_id
+          AND a.inventory_id = b.inventory_id
+          AND TIMESTAMPDIFF(SECOND, a.transaction_date, b.transaction_date) = -1
```
jmiranda: Add FIXME / TODO comment to remind us that we'd like to fix this once the reference document refactoring is done.

grails-app/domain/org/pih/warehouse/inventory/ProductCount.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+    String id
+
+    Transaction transaction
+
+    Product product
+
+    Location facility
+
+    Date dateRecorded
```
jmiranda: Can we call this dateCounted instead of dateRecorded?

grails-app/domain/org/pih/warehouse/inventory/ProductCount.groovy
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Counts can be created via baseline transaction + adjustment transaction together, alone adjustment or alone baseline.
+     * If it's baseline + adjustment pair, the associatedTransaction would be the adjustment transaction.
+     * For cases with alone adjustment or alone baseline, associatedTransaction is supposed to be null
+     * This means - if transactionProfile is BASELINE_WITH_ADJUSTMENT, we expect the associatedTransaction not to be null.
+     */
+    Transaction associatedTransaction
+
+    static constraints = {
+        table "product_counts"
```
jmiranda: inventory_count

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+            facility.id AS facility_id,
+            t.transaction_date AS date_recorded,
+            'BASELINE_WITH_ADJUSTMENT' as transaction_profile,
+            bam.adjustment_id as associated_transaction_id
+            FROM transaction t
+            JOIN baseline_adjustment_matches bam ON t.id = bam.baseline_id
+            JOIN transaction_entry te ON te.transaction_id = t.id
+            JOIN inventory_item ii ON ii.id = te.inventory_item_id
+            JOIN location facility ON facility.inventory_id = t.inventory_id
+            WHERE t.transaction_type_id = '12'
```
jmiranda: Same as above

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+            facility.id AS facility_id,
+            t.transaction_date AS date_recorded,
+            'BASELINE_WITHOUT_ADJUSTMENT' as transaction_profile,
+            -- Alone baseline doesn't have any associated transaction (like baseline + adjustment), so hardcode it to NULL
+            NULL as associated_transaction_id
+        FROM transaction t
+        JOIN transaction_entry te ON te.transaction_id = t.id
+        JOIN inventory_item ii ON ii.id = te.inventory_item_id
+        JOIN location facility ON facility.inventory_id = t.inventory_id
+        WHERE t.transaction_type_id = '12'
```
jmiranda: Same as above

grails-app/migrations/views/product_counts.sql
```diff
@@ -0,0 +1,115 @@
...
+            facility.id AS facility_id,
+            t.transaction_date AS date_recorded,
+            'ADJUSTMENT_WITHOUT_BASELINE' as transaction_profile,
+            -- Alone adjustment doesn't have any associated transaction (like baseline + adjustment), so hardcode it to NULL
+            NULL as associated_transaction_id
+        FROM transaction t
+        JOIN transaction_entry te ON te.transaction_id = t.id
+        JOIN inventory_item ii ON ii.id = te.inventory_item_id
+        JOIN location facility ON facility.inventory_id = t.inventory_id
+        WHERE t.transaction_type_id = '3'
```
jmiranda: Same as above

grails-app/domain/org/pih/warehouse/inventory/ProductCount.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+    Transaction transaction
+
+    Product product
+
+    Location facility
+
+    Date dateRecorded
+
+    TransactionProfile transactionProfile
```
jmiranda: Let's call this InventoryCountTypeCode with values

BASELINE
BASELINE_ADJUSTMENT
ADJUSTMENT

src/main/groovy/org/pih/warehouse/inventory/TransactionProfile.groovy
jmiranda: I like where you're going with the "composition" concept, but that's also somewhat vague. How about InventoryCountTypeCode or InventoryCountSourceTypeCode since it's less about type and more about the source of the information (baseline vs adjustment vs baseline + adjustment).

awalkowiak (approved):

PR #5404, opened 2025-07-23
OBPIH-7414 fix duplicate tags and categories being created during pro…

ewaterman:
…duct import

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7414

**Description:** Two issues:
- We were trying to fetch newly created tags/categories from the database before the cache was flushed, so even when one row created a category, when the next row fetches the same category, it isn't found, resulting in a duplicate being created
- We weren't stripping whitespace so an import value of `tags: "x, y"` would result in tags "x" and " y" (note the space). So even if we already have a tag "y", "y" != " y", so we'd still create a new one

I refactored the code to first gather a list of all tags and categories being imported, then get or create them all in a batch, then use that result (without fetching from the db again) to populate the tags and categories in the product.

---

Here's a video showcasing the bug: https://www.loom.com/share/1a600be8b48e423eba8ba871bdff2b44

### DIFF
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -13,6 +13,7 @@ import grails.core.GrailsApplication
 import grails.gorm.transactions.Transactional
 import grails.validation.ValidationException
 import groovy.xml.Namespace
+import org.apache.commons.lang.StringUtils
 import org.hibernate.criterion.CriteriaSpecification
 import org.hibernate.criterion.Restrictions
 import org.hibernate.sql.JoinType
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1025,67 +1073,46 @@ class ProductService {
...
-            log.info "Product ${product} already contains tag ${tag}"
+    private Tag findOrCreateTag(String tagName) {
+        Tag tag = Tag.findByTag(tagName)
+        if (tag) {
+            return tag
         }
-    }
 
+        tag = new Tag(tag: tagName)
+        if (!tag.save()) {
```
ewaterman, line 1111: Note that after this refactor, we don't need to flush the cache when saving tags anymore since we're directly returning and using them (whereas before we'd try to fetch from the db later and so we had to ensure the db session was flushed).

awalkowiak (approved):

PR #5408, opened 2025-07-24
OBPIH-7313 Fix Inventory Transaction Report qty before calculation wh…

kchelstowski:
…en proceeding inventory import with different product

**Description:**

---

### DIFF
--- a/grails-app/migrations/views/inventory-transactions-summary.sql
+++ b/grails-app/migrations/views/inventory-transactions-summary.sql
@@ -12,7 +12,7 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
             JOIN location facility ON transaction.inventory_id = facility.inventory_id
             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
             WHERE transaction_type.id = '12' -- baseline inventory transaction
-            GROUP BY transaction.id
+            GROUP BY transaction.id, inventory_item.product_id
     ),
     -- Helper view for adjustments to query against while searching for the latest baseline transaction
     -- without any adjustment transaction between queried adjustment and the latest baseline


grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -12,7 +12,7 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
             JOIN location facility ON transaction.inventory_id = facility.inventory_id
             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
             WHERE transaction_type.id = '12' -- baseline inventory transaction
-            GROUP BY transaction.id
+            GROUP BY transaction.id, inventory_item.product_id
```
kchelstowski, line 15: When importing inventory we can import multiple products - this means - one transaction id can be shared by many entries with different product id.
Having imported e.g. 10001, 10002 and 10003, we'd have transaction with id X, and entries A, B, C.
As we query `FROM` entries, join transaction and group by `transaction.id`, we'd have only one row representing this transaction, and it would just randomly (for the `product_id`) take the row with entry A, B or C (let's say it used A).
So we might miss two baseline rows for entries B and C that we were supposed to be joining later (`LEFT JOIN product_inventory_summary pis`). 
Having not joined the `product_inventory_summary`, it was assumed that the stock is recorded for the first time for B and C, which could result in qty before == 0 or more dangerously - since we do the `SUM(transaction_entry.quantity)`, for entry A, the qty before that was displayed, would not be the SUM of product 10001, but the SUM of 10001, 10002, 10003.

The fix is straight forward - since the inventory import allows to import different products, `transaction.id` does not define the uniqueness anymore, and we also have to group by the `product.id`

Issue #1351, opened 2019-12-05
System property user.home not being set when using Tomcat installed from download


jmiranda:
This causes the system to try to create the uploads directory under root (/). This fails because the tomcat user that executes the Tomcat server does not have permission to create a directory under root. When we install Tomcat from APT package repository, user.home seems to be set properly.

```
directoryPath = directoryPath.replaceFirst("^~", System.getProperty("user.home"))
```

Issue #47, opened 2015-03-28
New installation should create uploads directory by default


jmiranda:
```Unable to upload file due to exception: uploads/inventoryLevel.xls (No such file or directory)```

jmiranda:
Also, add to documentation to explain how to manually create the directory.

jmiranda:
Fixed as part of release 0.7.29.

PR #5415, opened 2025-07-28
Add docs describing how to override the default uploads folder

ewaterman:

**Link to GitHub issue or Jira ticket:** https://github.com/openboxes/openboxes/pull/3774

**Description:** I was looking through old PRs and I saw https://github.com/openboxes/openboxes/pull/3774 and realized it'd be valuable to include. The docs have changed a lot since then though so instead of trying to resolve conflicts in that PR, I just created a new one

### DIFF
--- a/docs/admin-guide/configuration/uploads.md
+++ b/docs/admin-guide/configuration/uploads.md
@@ -1,15 +1,7 @@
+Defines the directory to use for (temporarily) storing uploaded files such as product images, shipment documents, and data imports.
+If the uploads folder does not exist, tomcat will attempt to create it (which will only succeed if tomcat has write permissions in the parent location).
 
-## File Uploads Location
-Directory for storing temporary uploaded files, including but not limited to product images, 
-shipment documents, data imports.
-
-```shell
-openboxes:
-    uploads:
-        location: "uploads"
-```
-
-### Customization
+## Configuration
 
 ```shell
 openboxes:
@@ -17,11 +9,15 @@ openboxes:
         location: "uploads"
 ```
 
+## Customization
+The default configuration adds the uploads directory to the root of the server. This is fine for local testing, but in production you'll likely want to override this behaviour to provide a more specific path for the uploads directory.
+
+Something like `"/opt/tomcat/uploads"` would likely work, or `"/opt/tomcat/webapps/openboxes/uploads"` if you have multiple tomcat apps.
+
+!!! note "Reminder"
+    The uploads directory must be a location that tomcat has permission to write to. The default `/uploads` directory won't be created automatically since tomcat won't have permission to create a folder here. If you want to use the default configuration, you will need to create the folder yourself manually and grant the tomcat user permissions to write to the folder.
 
 ### Related
 * https://community.openboxes.com/t/unable-to-upload-file-due-to-exception-uploads-inventory-xls-no-such-file-or-directory/211
 * https://community.openboxes.com/t/how-do-i-create-a-new-product/135
 * https://community.openboxes.com/t/product-pictures/290
-
-
-## 


docs/admin-guide/configuration/uploads.md
```diff
@@ -1,27 +1,23 @@
...
 
 ```shell
 openboxes:
     uploads:
         location: "uploads"
 ```
 
-### Customization
+## Customization
+The default configuration adds the uploads directory to the root of the server. This is fine for local testing, but in production you'll likely want to override this behaviour to provide a more specific path for the uploads directory.
```
jmiranda: > adds the uploads directory to the root of the server

The configuration provides a relative path ("uploads") so that the directory should be created under the Tomcat directory (i.e. /opt/tomcat). However, we've always had a problem in which the directory is not created on startup (https://pihemr.atlassian.net/browse/OBGM-219)

```
May 04 20:35:27 OBNAV-test1 tomcat9[104502]: 2020-05-04 20:35:27.396  INFO --- [           main] org.pih.warehouse.core.FileService       : Attempting to create directory /var/lib/tomcat9/uploads
May 04 20:35:27 OBNAV-test1 tomcat9[104502]: 2020-05-04 20:35:27.397 ERROR --- [           main] org.pih.warehouse.core.FileService       : - Directory /var/lib/tomcat9/uploads cannot be created
```

This leads to errors like the following one when attempting to import data or upload files. I think there were other errors as well. 

```
Unable to upload file due to exception: uploads/inventory.xls (No such file or directory
```
Source: 
* https://community.openboxes.com/t/unable-to-upload-file-due-to-exception-uploads-inventory-xls-no-such-file-or-directory/211
* https://pihemr.atlassian.net/browse/OBGM-231

Note: At the end of the aforementioned ticket there was also a need to fix a permission issue within the Tomcat 9 configuration. **This might be important for the configuration OR troubleshooting docs.**

We originally fixed the issue here, but it never worked 100% of the time.
https://github.com/openboxes/openboxes/issues/47

As a workaround, I think we usually recommend that users create the directory themselves. Here's a community forum post about the issue (although there are likely others).

https://community.openboxes.com/t/unable-to-upload-file-due-to-exception-uploads-inventory-xls-no-such-file-or-directory/211/2

So the default of using the relative directory (which should create the uploads directory under /opt/tomcat) is still the proper approach, even for production. The problem is that the directory is never properly created on startup. I could have sworn we've created tickets and investigated this issue, but I think those efforts weren't as exhaustive as we'd like because we had a workaround. In fact, I think the solution to OBGM-219 was "apply workaround". 

Note: When we actually put some effort into containerizing OpenBoxes (read: ephemeral storage), we should eventually remove the need for temporary file storage. 

I believe the following ticket was created as a starting point for that, but I think the work should be to remove the need for temporary file storage (hint: we can store the input file in memory), not find an alternative. There's a case for storing documents on an alternative file storage medium (i.e. S3) but that's a separate problem with a seperate ticket.

https://pihemr.atlassian.net/browse/OBPIH-3963

jmiranda: By the way, here's another github issue that might explain the issue with creating the directory.

https://github.com/openboxes/openboxes/issues/1351

docs/admin-guide/configuration/uploads.md
```diff
@@ -1,27 +1,23 @@
...
 
-```shell
-openboxes:
-    uploads:
-        location: "uploads"
-```
+Something like `"/opt/tomcat/uploads"` would likely work, or `"/opt/tomcat/webapps/openboxes/uploads"` if you have multiple tomcat apps.
```
jmiranda: We also need to create the directory (which is the reason why "uploads" doesn't work). See some of the suggestions from the community forum posts in case there's something there that you can copy and paste. 

https://community.openboxes.com/t/unable-to-upload-file-due-to-exception-uploads-inventory-xls-no-such-file-or-directory/211
https://community.openboxes.com/t/filenotfoundexception-tp-link-eap245-ac1750-wireless-gigabit-ceil-eap245-jpg-permission-denied/281

docs/admin-guide/configuration/uploads.md
```diff
@@ -1,27 +1,23 @@
...
 
+!!! note "Reminder"
+    The uploads directory must be a location that tomcat has permission to write to. The default `/uploads` directory won't be created automatically since tomcat won't have permission to create a folder here. If you want to use the default configuration, you will need to create the folder yourself manually and grant the tomcat user permissions to write to the folder.
```
jmiranda: Nice, thanks for including that. 

We could/should include the solution alluded to in the  from the OBGM ticket (https://pihemr.atlassian.net/browse/OBGM-231?focusedCommentId=105011) in which this was resolved.  first link below about how to configure 

> The default `/uploads` directory won't be created automatically

This is a little misleading, so we should be more clear about what's happening here. The intent of the default config i.e. "uploads" was not to create a directory under root i.e. "/uploads". If Tomcat was doing what it was supposed to, the default config ("uploads") should actually be resolved to "/opt/tomcat/uploads". Due to a "bug?" within Tomcat that we've never fully investigated (https://github.com/openboxes/openboxes/issues/1351), the path is resolved to "/uploads". 

Because of that, Tomcat fails to create the uploads directory on startup, leaving us in the position where we need to manually create the directory. 

I don't know how to convey that in this "Reminder" (or if we even need to) but we should be clear that the intent was not to create "/uploads".

jmiranda: The permission issue is discussed in the following comment
https://pihemr.atlassian.net/browse/OBGM-231?focusedCommentId=105011

jmiranda (changes requested):

docs/admin-guide/configuration/uploads.md
ewaterman: thanks for this. It's really helpful context. I'll update the docs here to reflect all this. Probably via a small troubleshooting section with some links to the community post

docs/admin-guide/configuration/uploads.md
```diff
@@ -1,27 +1,82 @@
...
-* https://community.openboxes.com/t/unable-to-upload-file-due-to-exception-uploads-inventory-xls-no-such-file-or-directory/211
-* https://community.openboxes.com/t/how-do-i-create-a-new-product/135
-* https://community.openboxes.com/t/product-pictures/290
+For example, if `$CATALINA_HOME=/var/lib/tomcat9` and you're using the default configuration of "uploads":
 
+```
+# Security
+...
+ReadWritePaths=/var/lib/tomcat9/uploads/
+```
```
jmiranda, line 53: I would feel more comfortable using `/opt/tomcat` here since our installation instructions use this path. Or if there's some way to make it clear that this should be WHATEVER_CATALINA_HOME_IS_APPROPRIATE_FOR_YOUR_SYSTEM

docs/admin-guide/configuration/uploads.md
```diff
@@ -1,27 +1,82 @@
...
+
+```
+INFO  core.UploadService  - Find or create uploads directory uploads
+INFO  core.FileService  - Attempting to create directory /uploads
+ERROR core.FileService  - - Directory /uploads cannot be created
+```
+
+If you encounter the above error, change the upload location property to be an absolute path.
+
+For example, if `$CATALINA_HOME=/var/lib/tomcat9`:
```
jmiranda: Same here.

docs/admin-guide/configuration/uploads.md
ewaterman: yeah I'll switch it to `/opt/tomcat` as a default and be more clear that it'll be whatever your CATALINA_HOME is

PR #5416, opened 2025-07-28
OBPIH-7259 Fix calculation of number of counts for an adjustment with…

kchelstowski:
… different lot number than in the baseline transaction

**Description:**

---

### DIFF
--- a/grails-app/migrations/views/inventory-counts.sql
+++ b/grails-app/migrations/views/inventory-counts.sql
@@ -3,6 +3,7 @@ CREATE OR REPLACE VIEW inventory_counts AS
     WITH inventory_baseline_candidates AS (
         SELECT t.id AS transaction_id,
                ii.id AS inventory_item_id,
+               ii.product_id as product_id,
                t.transaction_date,
                t.inventory_id
         FROM transaction t
@@ -13,6 +14,7 @@ CREATE OR REPLACE VIEW inventory_counts AS
     adjustments_candidates AS (
         SELECT t.id AS transaction_id,
                ii.id AS inventory_item_id,
+               ii.product_id as product_id,
                t.transaction_date,
                t.inventory_id
         FROM transaction t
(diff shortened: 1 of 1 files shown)


grails-app/migrations/views/inventory-counts.sql
```diff
@@ -29,8 +29,7 @@ CREATE OR REPLACE VIEW inventory_counts AS
             b.transaction_id AS baseline_id
         FROM adjustments_candidates a
         JOIN inventory_baseline_candidates b
-          ON a.inventory_item_id = b.inventory_item_id
```
kchelstowski, line 32: the JOIN on inventory item is not necessary and caused the bug that for such record stock:
Screenshot from 2025-07-28 19-04-10

two counts would be counted (1 is the expected result).

The reason for that is that such adjustment was created for an inventory item (`TEEST` lot) that has not been present in the baseline that is associated with such adjustment.
Previously I thought that used inventory items would always be present in both adjustment and the baseline, but it turned out this might be possible to have different inventory items throughout the transaction process.
I believe this JOIN on inventory item is not needed at all, as JOIN by inventory and the TIMESTAMPDIFF still define the uniqueness of baseline + adjustment pair.

ewaterman: yeah I think this is safe to do since going forward we can only have a single baseline/adjustment transaction for a given datetime. It's not a perfect solution since it's not actually a strict requirement in the database (transactionDate is not unique) but since we're already relying on the +1 second assumption so heavily throughout the code, this is fine for now.

There's one scenario that I'm thinking of though:
- we do a record stock (or whatever) and there is no difference so only a baseline is created
- we create a separate adjustment transaction (for example using the create adjustment feature) that is one second after the baseline

Then this query will assume the baseline and adjustment are related, when they might not be. But that's such a rare edge case that I think it would have to be done intentionally to get it so I'm not worried.

ewaterman (approved):

grails-app/migrations/views/inventory-counts.sql
kchelstowski: The two points you mentioned are already handled in two other subviews.
Why I’m replying though is that after I turned off my PC I realized that this won’t work for the case when we do the import with different items. 
The transaction date would still be the same for those different products, so I will change the code tomorrow to still join the inventory item table, but via the product_id not the inventory item id.

alannadolny (approved):

awalkowiak (approved):

Issue #5417, opened 2025-07-28
Allow GitHub Actions to run on PRs in forks


ewaterman:
We have a number of actions that run when a PR is created (one for labelling the PR, one for writing tests, one for creating a new mkdocs build). Currently, when a new PR is created from a fork of the repository, these jobs fail with auth/permission errors. This is because jobs with the `pull_request` event run in the context of the origin/source repository (ie the fork), which does not have access to the workflow secrets in openboxes. This is intentionally done for security reasons so that a random user can't create a fork, write a malicious GitHub action, make a PR to the original repository, then have that action run immediately with access to internal secrets.

However we want to be able to run actions against forks since we want to see tests pass before merging a change into the main repository. So we need to find a balance between running actions freely on PRs while still protecting secrets.

There is the pull_request_target action that allows the action to run in the context of the the target/destination repository (ie openboxes). We can use this, but it has the security vulnerability stated above. https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

As such we need to do two things:
- First, change our github settings to require repo admin approval before running actions in a forked PR
- Then (and ONLY then), update our actions to use pull_request_target instead of pull_request

PR #5418, opened 2025-07-28
5417: use pull_request_target in github actions so they run on fork PRs

ewaterman:

**Link to GitHub issue or Jira ticket:** https://github.com/openboxes/openboxes/issues/5417

**Description:** Allow GitHub Actions to run on PRs from forks. 

For security purposes, I also set all PRs from forks to require approval (from a person with write access to the repo) before any actions are run:
image

This is so people can't maliciously change an action in their fork and have it immediately run in a PR against our repo.


.github/workflows/dbdocs.yml
```diff
@@ -1,29 +1,21 @@
-# This is a basic workflow to help you get started with Actions
-
 name: dbdocs
 
-# Controls when the action will run. 
 on:
-  # Triggers the workflow on push or pull request events but only for the develop branch
+  # Update the docs whenever a new change is pushed to the develop branch. As such, dbdocs will always
+  # contain the 'latest' database state (instead of being in sync with 'main' / the latest stable release).
```
ewaterman: @jmiranda we are currently updating the dbdocs whenever we create a new PR to develop. Is this intentional? I'd assume we only want to do it when a commit is actually merged to develop. As far as I can see, dbdocs only has the one instance, so pushing whenever a PR is created will stomp over the last docs.

ewaterman: I also made it so it only rebuilds if there was a change in `grails-app/migrations/**` since otherwise there will be no schema difference

ewaterman: I talked with Justin about this in Slack. I'm going to revert the change for now and simply disable this job since it doesn't actually do anything (the dbml file doesn't get regenerated)

ewaterman:
btw the reason why the jobs don't run is because changing from pull_request to pull_request_target makes it run using the workflow file on the target (ie develop branch). Since develop doesn't have pull_request_target yet, it does nothing. Subsequent PRs should pick it up (I hope)

Issue #4049, opened 2023-05-19
Automate dbdocs action to convert generated SQL to DBML


jmiranda:
After talking to the dbdocs team about the service I realized we're not automatically updating openboxes.dbml in the source code. In fact, we're not even periodically updating it manually. It would be great to use Hibernate export-schema or Liquibase to export SQL that can be converted to DBML and used to publish the documentation. 

We might need to manually update the openboxes.dbml file in order to add grouping, descriptions, etc. 

We might also want to restrict updates to PRs merged to develop rather than ALL commits/PRs.

jmiranda:
Thinking this might be low-hanging fruit, I took a crack at this tonight.
https://community.dbdiagram.io/t/anyone-have-success-setting-up-a-continuous-integration-using-github-actions/1465

Checkout code 
```
git checkout feature/upgrade-to-grails-3.3.10
git checkout -b 4094-automate-dbdocs
```
Install / configure dependencies
```
npm install dbdocs
sdk use grails 3.3.18
sdk use java 8.0.302-open
```
Export schema 
```
grails schema-export
```
Error
```
Caused by: org.gradle.execution.TaskSelectionException: Task 'schemaExport' not found in root project 'openboxes'.
	at org.gradle.execution.TaskSelector.getSelection(TaskSelector.java:116)
	at org.gradle.execution.TaskSelector.getSelection(TaskSelector.java:81)
	at org.gradle.execution.commandline.CommandLineTaskParser.parseTasks(CommandLineTaskParser.java:42)
	at org.gradle.execution.TaskNameResolvingBuildConfigurationAction.configure(TaskNameResolvingBuildConfigurationAction.java:44)
	at org.gradle.execution.DefaultBuildConfigurationActionExecuter.configure(DefaultBuildConfigurationActionExecuter.java:48)
	at org.gradle.execution.DefaultBuildConfigurationActionExecuter.access$000(DefaultBuildConfigurationActionExecuter.java:25)
	at org.gradle.execution.DefaultBuildConfigurationActionExecuter$1.proceed(DefaultBuildConfigurationActionExecuter.java:54)
	at org.gradle.execution.DefaultTasksBuildExecutionAction.configure(DefaultTasksBuildExecutionAction.java:44)
	at org.gradle.execution.DefaultBuildConfigurationActionExecuter.configure(DefaultBuildConfigurationActionExecuter.java:48)
[...]

PR #5420, opened 2025-07-28
autogenerate and publish dbdocs when a schema change is submitted

ewaterman:

**Link to GitHub issue or Jira ticket:** closes https://github.com/openboxes/openboxes/issues/4049

**Description:** GitHub Action for auto-generating our openboxes.dbml file and uploading it to the dbdocs site

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -7,9 +7,6 @@ on:
       CODECOV_TOKEN:
         required: true
 
-env:
-  JAVA_VERSION: 8
-
 jobs:
   run-backend-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/backend-tests.yml
```diff
@@ -26,7 +23,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          java-version: ${{env.JAVA_VERSION}}
+          java-version: ${{vars.JAVA_VERSION}}
```
ewaterman, line 26: unrelated to this change. I realized while I was working on this PR that we could set up these static variables directly in GitHub variables so no need to set a new env var in each file. Kinda neat!

PR #5423, opened 2025-07-29
OBPIH-7376 Errors encountered when using Record Stock to create or update inventory (Fix)

alannadolny:
The issue that caused the reopen is no longer visible. The fix here is for the error with an empty product ID after refreshing the page

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
@@ -583,27 +583,8 @@ class InventoryItemController {
             return
         }
 
-        log.info("User chose to validate or there are errors")
-        Location warehouseInstance = Location.get(session?.warehouse?.id)
-
-        commandInstance.product = Product.load(commandInstance.product.id)
-        commandInstance.inventory = warehouseInstance?.inventory
-        commandInstance.inventoryLevel = InventoryLevel.findByProductAndInventory(commandInstance?.product, commandInstance?.inventory)
-
-        Product productInstance = commandInstance.product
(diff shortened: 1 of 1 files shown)


grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -583,27 +583,8 @@ class InventoryItemController {
...
-        commandInstance.totalQuantity = inventoryService.getQuantityByProductMap(transactionEntryList)[productInstance] ?: 0
-
-        log.info "commandInstance.recordInventoryRows: "
-        commandInstance?.recordInventoryRows.each {
-            log.info "it ${it?.id}:${it?.lotNumber}:${it?.oldQuantity}:${it?.newQuantity}"
-        }
-
-        render(view: "showRecordInventory", model: [commandInstance: commandInstance])
+        flash.errors = commandInstance.errors
+        redirect(action: "showRecordInventory", params: ['product.id': commandInstance.product.id])
```
ewaterman, line 587: I'm confused as to why passing `model: [commandInstance: commandInstance]` to the view isn't needed anymore. It looks like it's always used in the GSP (even when there are errors) but I'm never totally sure how those files are working so maybe I'm wrong. Is the fact that we're flashing errors bypass something?

alannadolny: The `showRecordInventory` initializes the command object in the same way as it was done here before. So I removed the initialization, and I made a redirect to that action. The same view is displayed; I just had to pass the command errors that occurred during saving the inventory record. That change fixed removing the `id` from the URL (we were not able to refresh the page before -- that was also the reason for the `null` product that you showed on the tech huddle, on Monday). The render action that was used here doesn't allow us to pass params as it does with redirect, so that was the reason why the id from params was missing.

PR #5425, opened 2025-07-30
OBPIH-7329 Product Inventory migration improvements

awalkowiak:
- Display other transactions with dates overlapping with product inventory
- Properly zero out stock that is available, but is not in the transaction entries
- Save migration report CSV
- Do not remove multi-product transactions for the same transaction date

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
@@ -20,6 +20,8 @@ import org.pih.warehouse.DateUtil
 import org.pih.warehouse.api.AvailableItem
 import org.pih.warehouse.auth.AuthService
 import org.pih.warehouse.core.Constants
+import org.pih.warehouse.core.Document
+import org.pih.warehouse.core.DocumentType
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.core.LocationType
 import org.pih.warehouse.core.LocationTypeCode
@@ -462,7 +464,7 @@ class MigrationService {
                         log.debug "Migrating transactions for product: " +
                                 "${transactionList?.transactionEntries?.inventoryItem?.product?.productCode?.unique()} " +
                                 "at location ${location.name}"
-                        processProductInventoryTransactions(transactionList, location, migratedBy, results)
+                        processProductInventoryTransactions(transactionList, location, migratedBy, results, true)
                     }
                 }
             }
@@ -490,22 +492,42 @@ class MigrationService {
 
             // Refresh PA for all products that were migrated
             productAvailabilityService.triggerRefreshProductAvailability(location.id, products?.id, true)
+
+            createCSVMigrationReportForResults(location, results)
         }
 
         return [
                 "Location"         : location?.name,
-                "Migration Results": results
+                "Migration Results": results.collectEntries { key, values ->
+                    [(key): [] <<
+                        ["Status".padRight(10) +
+                        "Transaction name".padRight(25) +
+                        "Transaction Number".padRight(25) +
+                        "Transaction date".padRight(35) +
+                        "QTY".padRight(10) +
+                        "ITEMKEY".padRight(70)] +
+                        values.collect {
+                            "${it.status}".padRight(10) +
+                            "${it.transactionTypeName}".padRight(25) +
+                            "${it.transactionNumber}".padRight(25) +
+                            "${it.transactionDate}".padRight(35) +
+                            "${it.quantity}".padRight(10) +
+                            "${it.product}:${it.lotNumber}:${it.binLocation}".padRight(70)
+                        }
+                    ]
+                },
         ]
     }
 
-    void processProductInventoryTransactions(List<Transaction> transactions, Location location, User migratedBy, Map results) {
+    void processProductInventoryTransactions(List<Transaction> transactions, Location location, User migratedBy, Map results, boolean singleProduct = false) {
         if (!transactions) {
             return
         }
 
         Date previousTransactionDate = null
         transactions.each { Transaction it ->
-            if (previousTransactionDate && it.transactionDate == previousTransactionDate) {
+            // For single product transactions, we can keep only the newest transaction with the same transaction date
+            if (singleProduct && previousTransactionDate && it.transactionDate == previousTransactionDate) {
                 log.debug "Transaction ${it.transactionNumber} has a transaction date equal to the previously processed " +
                         "transaction. Skipping migrating this one as it won't have effect on the stock."
                 it.disableRefresh = true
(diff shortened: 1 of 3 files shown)


grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -863,23 +885,54 @@ class MigrationService {
...
 
-        transactionEntries?.each {  TransactionEntry entry ->
+        Map<String, Map> groupedEntries = transactionEntries?.groupBy { [it.inventoryItem, it.binLocation] }?.collectEntries { key, entries ->
+            String itemKey = productAvailabilityService.constructAvailableItemKey(key[1], key[0])
+            String comments = entries.comments.findAll { it }.join(', ')
+            [(itemKey): [
+                binLocation: key[1],
+                inventoryItem: key[0],
+                quantity: entries.sum { it.quantity },
+                comments: comments.length() > 255 ? comments.substring(0, 255) : comments
```
ewaterman, line 896: Good idea to truncate it if its too long. I didn't think to do that in InventoryImportDataService. Maybe we can think of making a util for this kind of stuff (not in this PR though).

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -863,23 +885,54 @@ class MigrationService {
         // Don't bother populating the transaction's fields until we know we'll need one.
         Transaction transaction = new Transaction()
 
-        transactionEntries?.each {  TransactionEntry entry ->
+        Map<String, Map> groupedEntries = transactionEntries?.groupBy { [it.inventoryItem, it.binLocation] }?.collectEntries { key, entries ->
```
ewaterman, line 889: in what scenarios do we create multiple transaction entries for the same product + bin + lot? Seems like weird behaviour.

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -993,4 +1048,86 @@ class MigrationService {
...
+            FROM transaction_entry te1
+                     JOIN inventory_item ii1 ON te1.inventory_item_id = ii1.id
+                     JOIN transaction t1 ON te1.transaction_id = t1.id
+                     JOIN transaction_entry te2 ON te1.inventory_item_id != te2.inventory_item_id
+                     JOIN inventory_item ii2 ON te2.inventory_item_id = ii2.id
+                     JOIN transaction t2 ON te2.transaction_id = t2.id
+            WHERE
+              ii1.product_id = ii2.product_id
+              AND t1.transaction_date = t2.transaction_date
+              AND t1.id < t2.id
```
ewaterman, line 1070: is this safe to do? Are new ids always > old ones?

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -993,4 +1048,86 @@ class MigrationService {
...
+              AND (
+                (t1.transaction_type_id = :transactionTypeId and t2.transaction_type_id != :transactionTypeId) 
+                OR 
+                (t1.transaction_type_id != :transactionTypeId and t2.transaction_type_id = :transactionTypeId)
+              )
+              AND t1.inventory_id = :inventoryId
+              AND t2.inventory_id = :inventoryId
```
ewaterman, line 1077: I'm not a DBA expert, but won't this cause a lot of comparisons?

If you have these transactions:
- id: 1, product_id: 1, transaction_date: 2025-01-01
- id: 2, product_id: 1, transaction_date: 2025-01-01
- id: 3, product_id: 1, transaction_date: 2025-01-01
- id: 4, product_id: 1, transaction_date: 2025-01-01
- ...

It'll match all of the following:
- 1 with 2, 3, 4...
- 2 with 3, 4...
- 3 with 4...

Why do we need to compare every transaction with every other transaction that comes after it?

I guess I'm not understanding why we need the transaction_type_id check. You're finding a pair of transactions where one is the type and one isn't (ex: one is a CREDIT adjustment). Why is this check necessary?

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman I'd say in a user error scenario, here is an example https://obdev3.pih-emr.org/openboxes/inventory/showTransaction/cab2b1078c7e02c4018d213e8593484d, there are more transactions in the system that have same product + lot + bin combination.

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman yeah, I had an issue with longer comments for transaction, so I applied it here too

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman Actually, this is probably some garbage leftover. Let me check if I can get rid of it

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman 
> I guess I'm not understanding why we need the transaction_type_id check. You're finding a pair of transactions where one is the type and one isn't (ex: one is a CREDIT adjustment). Why is this check necessary?

The issue is that I am trying to find all transactions at the same transaction date for a given product, where one **is** a product inventory transaction and the other **is not** product inventory transaction, because simultanous product inventroy transactions are handled, but these are for the user to inform him, that he should check these first before migration (or after). 

I am not super happy with the query, but it works (or at least I think so), hence I did not try to simplify or improve it

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman oh, I know why it was there. As you can see, entries are joining other entries with the same inventory item so essentially we end up with duplicates, and this is a simple way to filter them out. Here is an example **without** this filter:

product  | t1 number         | t2 number          |  transaction date
11874    | FGB-72F-7AE   | QDN-D6E-HU9 | 2021-10-29 18:20:00
11874    | FGB-72F-7AE   | QDN-D6E-HU9 | 2021-10-29 18:20:00
11874    | QDN-D6E-HU9 | FGB-72F-7AE   | 2021-10-29 18:20:00
11874    | QDN-D6E-HU9 | FGB-72F-7AE   | 2021-10-29 18:20:00

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: @ewaterman if you are asking specifically for the transaction_type_id checking, then I think it needs to stay, because of the duplicates filtered out (the `t1.id < t2.id` part). Unless you are asking about something else let me know.

ewaterman: > The issue is that I am trying to find all transactions at the same transaction date for a given product, where one is a product inventory transaction and the other is not product inventory transaction, because simultanous product inventroy transactions are handled, but these are for the user to inform him, that he should check these first before migration (or after).

hmm yeah this makes sense. I'm comfortable with how it is now if it's performant enough, especially if it's just for display to inform the user.

There's maybe a way to do it more efficient by doing something like:
- selecting product_id, transaction_date where type is product inventory
- selecting where type is not product inventory and product_id and transaction_date are in the above query

That way you don't need to check every transaction on a date to every other transaction, but I don't know if that's actually more efficient.

Anways, the way you have it is fine assuming you're comfortable with the current performance.

ewaterman (approved):

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: I did not spot any issues on that end, so I guess we're good

PR #5426, opened 2025-07-30
OBPIH-7341 Fix assigning users when product quantity is changed to 0 during counting, and disable the Assign button if nothing is selected

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/hooks/cycleCount/useAssignCycleCountModal.jsx
+++ b/src/js/hooks/cycleCount/useAssignCycleCountModal.jsx
@@ -127,10 +127,9 @@ const useAssignCycleCountModal = ({
         const countData = getCycleCountItemsWithAssignedCountData(data.data, commands);
 
         if (countData.length > 0) {
-          await cycleCountApi.updateCycleCountItems(
+          await cycleCountApi.updateCycleCountItemsBatch(
             { itemsToUpdate: countData },
             currentLocation?.id,
-            cycleCountIds,
           );
         }
       }
--- a/src/js/hooks/cycleCount/useAssignCycleCountModal.jsx
+++ b/src/js/hooks/cycleCount/useAssignCycleCountModal.jsx
@@ -76,7 +76,7 @@ const useAssignCycleCountModal = ({
         recount: true,
         assignee: dataToAssign?.[cycleCount.maxCountIndex]?.assignee,
       }))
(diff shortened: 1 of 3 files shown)


src/js/hooks/cycleCount/useAssignCycleCountModal.jsx
```diff
@@ -75,7 +75,9 @@ const useAssignCycleCountModal = ({
         id: item.id,
         recount: true,
         assignee: dataToAssign?.[cycleCount.maxCountIndex]?.assignee,
-      }));
+      }))
+        // we only want to update items that have an assignee
+        .filter(item => item.assignee);
```
SebastianLib, line 80: The issue here was that if we changed the product quantity to 0 before saving, the countIndex 1 would not be created. As a result, in this line:

```
return cycleCountItems.map((item) => ({
  id: item.id,
  recount: true,
  assignee: dataToAssign?.[cycleCount.maxCountIndex]?.assignee,
}))
```

for the cycle count with quantity product the maxCountIndex was 0, while dataToAssign had an entry only at index 1, which caused the assignee to be null and led to further issues. Therefore, if the assignee is null, there's no point in calling updateCycleCountItemsBatch for it. In that case, the behavior will be the same as when assigning the assignee via the "To Count" or "To Resolve" tab, and everything will work as expected.

kchelstowski: This feels weird to me. User might now know which items were and which were not sent to save. Why do we actually want to update items that have the assignee only? What's the context? Isn't assignee supposed to be filled automatically if you don't set it?
I just want to avoid another duct tape on frontend to make something work.

SebastianLib: So here's the thing: we had an issue where, after finishing a count with product quantity 0, setting an assignee for the resolve step caused an error. That's not true that we don't save the assignee, it's just saved in the verificationCount so everything is saved in this fuction:       
```
const response = await cycleCountApi.updateCycleCountRequests(
        currentLocation?.id,
        {
          commands,
        },
      ); 
```
image
 
and if the product quantity is later changed (e.g. to 2), the assignee will appear in the resolve workflow. I showed Manon how it works now, and she's okay with it

alannadolny (approved):

PR #5428, opened 2025-08-01
OBPIH-7313 Fix inventory adjustment recognition and qty before/after …

kchelstowski:
…calculation + improve performance of inventory transactions report

**Description:**

- fixed calculation of qty before/after for inventory adjustment based transactions
- improved performance of inventory transactions report with only date range filter from ~ 5,2s to 400ms

---

### DIFF
--- a/grails-app/migrations/views/inventory-transactions-summary.sql
+++ b/grails-app/migrations/views/inventory-transactions-summary.sql
@@ -14,27 +14,102 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
             WHERE transaction_type.id = '12' -- baseline inventory transaction
             GROUP BY transaction.id, inventory_item.product_id
     ),
-    -- Helper view for adjustments to query against while searching for the latest baseline transaction
-    -- without any adjustment transaction between queried adjustment and the latest baseline
-    -- Without the helper view, when making the EXISTS statement explicit, it caused performance issues
-    -- because the select was called for each row
-    adjustments_candidates AS (
-       SELECT ii.id as inventory_item_id,
+    -- Helper view to calculate quantity per adjustment transaction and product
+    quantity_sum_for_adjustment AS (
+       SELECT ii.product_id as product_id,
               t.transaction_date as transaction_date,
-              t.inventory_id as inventory_id
+              t.inventory_id as inventory_id,
+              t.id as transaction_id,
+              SUM(te.quantity) as quantity_sum
            FROM transaction t
                     JOIN transaction_entry te ON te.transaction_id = t.id
                     JOIN inventory_item ii ON ii.id = te.inventory_item_id
            WHERE t.transaction_type_id = '3'
-    )
+           GROUP BY t.id, ii.product_id
+       )
 
 SELECT
     CRC32(CONCAT(transaction.id, inventory_item.product_id, facility.id)) as id,
     transaction.id AS transaction_id,
     inventory_item.product_id as product_id,
     facility.id as facility_id,
-    COALESCE(pis.quantity_balance, 0) AS quantity_before,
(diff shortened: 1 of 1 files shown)


alannadolny (approved):

grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -14,27 +14,102 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
...
+       SELECT ii.product_id as product_id,
               t.transaction_date as transaction_date,
-              t.inventory_id as inventory_id
+              t.inventory_id as inventory_id,
+              t.id as transaction_id,
+              SUM(te.quantity) as quantity_sum
            FROM transaction t
                     JOIN transaction_entry te ON te.transaction_id = t.id
                     JOIN inventory_item ii ON ii.id = te.inventory_item_id
            WHERE t.transaction_type_id = '3'
```
ewaterman, line 27: while you're here, can you please add a comment about what '3' means (like the other lines have)

grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -14,27 +14,102 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
...
+    -- and check if there are more "alone" adjustment between such baseline and the current calculated adjustment
+    CASE
+        WHEN pis.transaction_id IS NULL THEN (
+              COALESCE((
+                  -- In 99% scenarios there is going to be only one baseline with a particular transaction_date for a product
+                  -- but due to some stale old data, where we could experience duplicate baselines.
+                  -- So we take MAX here for two reasons:
+                  -- 1) to avoid exception with "Subquery did not return unique result"
+                  -- 2) we don't take SUM, but MAX not to double the result
+                  SELECT MAX(quantity_balance)
```
ewaterman, line 48: this feels arbitrary. Why not take the one with MAX(date_created)? Or can we not guarantee that only returns a single entry either?

I'd take a look at what product availability does in that scenario (I think it looks at date_created) and mirror that behaviour. That way the reports are more likely to be in sync with QoH / stock history

grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -14,27 +14,102 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
...
+                      AND product_inventory_summary.facility_id = facility.id
+                      AND product_inventory_summary.baseline_transaction_date < transaction.transaction_date
+                 )
+                 AND quantity_sum_for_adjustment.transaction_date < transaction.transaction_date
+             ), 0)
+        )
+        -- If we have an associated baseline, we just take its quantity, we don't need to search for anything "between"
+        ELSE COALESCE(pis.quantity_balance, 0)
+    END AS quantity_before,
+     -- The condition for calculate quantity after is just quantity before + current sum of adjustment
```
ewaterman, line 82: I'm confused why we have both this logic and the above "quantity_before" logic, and why they seem to be different. You can't just take "quantity_before" and add it to the below?

grails-app/migrations/views/inventory-transactions-summary.sql
kchelstowski: Yes, there might be some old stale data where you have two baselines with exactly the same date. I wouldn’t have caught that if I hadn’t got an exception with „subquery did not return a single result”, and I started debugging why. 
I was also surprised that the MAX(date) did not cover the uniqueness.

grails-app/migrations/views/inventory-transactions-summary.sql
kchelstowski: You can’t use aliases at this point. When you declare e.g.
SELECT 
 SUM(foo) as some_sum,
 some_sum + 5 <—— you would fail at this, as some_sum would not yet be recognized

PR #5430, opened 2025-08-05
OBPIH-7313 Fix inventory transactions report crashing MariaDB

kchelstowski:

**Description:**

---

### DIFF
--- a/grails-app/migrations/views/inventory-transactions-summary.sql
+++ b/grails-app/migrations/views/inventory-transactions-summary.sql
@@ -2,6 +2,7 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
        WITH product_inventory_summary AS (
            SELECT
               inventory_item.product_id as product_id,
+              product.product_code as product_code,
               facility.id as facility_id,
               transaction.id as transaction_id,
               transaction.transaction_date as baseline_transaction_date,
@@ -11,6 +12,7 @@ CREATE OR REPLACE VIEW inventory_transactions_summary AS
             JOIN transaction_type ON transaction.transaction_type_id = transaction_type.id
             JOIN location facility ON transaction.inventory_id = facility.inventory_id
             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
+            JOIN product ON product.id = inventory_item.product_id
             WHERE transaction_type.id = '12' -- baseline inventory transaction
             GROUP BY transaction.id, inventory_item.product_id
     ),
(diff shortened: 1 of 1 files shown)


grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -47,13 +51,13 @@ SELECT
...
                   FROM product_inventory_summary pis
-                  WHERE pis.product_id = inventory_item.product_id
+                  WHERE pis.product_code = product.product_code
                     AND pis.facility_id = facility.id
                     -- Find the latest baseline before the calculated adjustment
                     AND pis.baseline_transaction_date = (
                       SELECT MAX(pis2.baseline_transaction_date)
                       FROM product_inventory_summary pis2
-                      WHERE pis2.product_id = inventory_item.product_id
+                      WHERE pis2.product_code = product.product_code -- comparing the product_code instead of inventory_item.product_id is intentional to avoid MariaDB bug when accessing the inventory_item.product_id with the table prefix (OBS-1901)
```
kchelstowski, line 60: It's a shame what is going on in the MariaDB for this clause. I don't even know how to explain it properly, but basically, this WHERE clause:
```sql
WHERE pis2.product_id = inventory_item.product_id
```
and the others where the same comparison is done, cause the MariaDB to crash. I still don't understand the underlying cause of that, because we do a similar thing with the facility:
```sql
AND pis2.facility_id = facility.id
```
where we add the table prefix like `facility.id`, and it works fine, whereas it crashes when we do a similar thing with the `inventory_item` table prefix.

What made the MariaDB works at first sight? When I refactored the WHERE to access the `product_id` directly (as we do the `inventory_item.product_id as product_id`):
```sql
 WHERE pis.product_id = product_id
```
The above worked fine for MariaDB, but did not for MySQL....... I mean, it did, but I was not even able to run the query, it ran for about 10 minutes and I surrendered. The reason for that was that without the table prefixes, the MySQL did not use any index for that field, it did not know what index key it should use, which caused the performances issues.

What else I tried?

- removing the `as product_id` and leaving the `inventory_item.product_id` alone in the SELECT, but it still did not work
- when I realized `facility` is the same case, and it works fine, I thought maybe it's due to the fact that we do the `JOIN location facility` whereas for inventory_item we did not add any shortcut, so I added `JOIN inventory_item ii` and then access the field via `ii.product_id` - it still didn't work
- since we JOIN the product table, I tried to look for `product.id` directly like: `WHERE pis.product_id = product.id` - still did not work


Randomly, I tried to look for any other unique field of product, so I tried to use `product_code` - surprisingly, it worked and it since both `product_id` and `product_code` are unique, we can use them interchangeably.  
Pros of that change is that it works ( :laughing: ) and it provides the same performance for MySQL and MariaDB.
Cons of that change is that it adds an additional JOIN to product table, but fortunately this doesn't impact the performance.

ewaterman (approved):
yeah this is strange. I don't understand the inner workings of mysql/mariadb to know why this is the case. Maybe something to do wth the field being a different data type or a primary key? Idk...

I would like to investigate it further to get a better understanding as to why this change is required but since we're needing to get release stable, lets just get this merged in to unblock testing

grails-app/migrations/views/inventory-transactions-summary.sql
ewaterman: thanks for the in depth explanation. I don't have much to add at this point but it is strange...

PR #5436, opened 2025-08-07
OBPIH-7443 fix import demo products error when tag is null and when category nam…

ewaterman:
…e doesn't match the case

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7443

**Description:** Two fixes:
- `tagNamesForAllProducts + products.tags.flatten() as List<String>` throws an error when tagNamesForAllProducts is null 
- Some product categories in the demo data aren't a case-sensitive match to a default category. In the base install changelog it was creating a category with name "ARVS" but in the `products.csv` demo data it was trying to assign "ARVs" to the products. This caused the import service to not be able to resolve the category and thus threw an error. I changed the csv file to use "ARVS" so that it matches the real category, but I also refactored the import logic to be case-insensitive so that even without the csv change it'll see them as the same category. The current behaviour on prod is that it's case-insensitive so the change in this PR preserves existing behaviour. (I refactored product import recently which changed the behaviour to be case-sensitive.)

The behaviour is slightly different depending on if the categories already exist:

Say we want to import product 1 with category "HI" and product 2 with category "hi"

Scenario 1) Categories "HI" and "hi" both already exist. When we import, p1 gets category "HI" and p2 gets category "hi"
Scenario 2) Category "HI" already exists, "hi" does not. When we import, both p1 and p2 get category "HI"
Scenario 3) Category "hi" already exists, "HI" does not. When we import, both p1 and p2 get category "hi"
Scenario 4) Neither category "hi" nor "HI" already exist. Whichever category appears first in the csv will get created and used for both products. If "hi" is first, both p1 and p2 get the categcategory "hi".

Kinda weird behaviour but it does make sense since if we already have two categories with the same name but a different case, we presumably want to match those directly, and if we only have one or neither, we assume user input error and do a case-insensitive match. And again, this is the existing behaviour.

---


grails-app/conf/templates/configuration/products.csv
```diff
@@ -1,10 +1,10 @@
...
 ,QX039,Default,Morphine 10mg immediate release tablet,,Pain,,,Each,Tertiary Care,0.25,TRUE,,TRUE,,,,,,,,,,,,,
 ,TS894,Default,"Morphine injection, 10mg/mL in 1mL ampoule",,Pain,,,Each,Tertiary Care,1.25,TRUE,,TRUE,,,,,,,,,,,,,
-,CR388,Default,Lamivudine 150mg + nevirapine 200mg + zidovudine 300mg tablet,,ARVs,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
-,TZ835,Default,Lamivudine 150mg + zidovudine 300mg tablet,,ARVs,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
-,CM871,Default,Lamivudine 150mg tablet,,ARVs,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
-,MC017,Default,Lamivudine 30mg + nevirapine 50mg + zidovudine 60mg tablet,,ARVs,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
+,CR388,Default,Lamivudine 150mg + nevirapine 200mg + zidovudine 300mg tablet,,ARVS,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
+,TZ835,Default,Lamivudine 150mg + zidovudine 300mg tablet,,ARVS,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
+,CM871,Default,Lamivudine 150mg tablet,,ARVS,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
+,MC017,Default,Lamivudine 30mg + nevirapine 50mg + zidovudine 60mg tablet,,ARVS,,,Each,Primary Care,0.29,TRUE,,,,,,,,,,,,,,,
```
ewaterman, line 7: Take a look at `default_OB_categories.csv` (and `changelog-insert-data-core.groovy` but I don't think that's used). The categories are created with ARVS but this file tried to set ARVs.

The code change in ProductService makes this csv change unnecessary (because it doesn't case-insensitive matching), but I figured it'd be best to keep the data in sync.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -904,11 +909,47 @@ class ProductService {
...
     /**
-     * Find or create a list of categories with the given names as a map keyed on category name. Nulls, duplicates,
-     * and whitespace will be ignored.
+     * A map of Category keyed on the "desensitized" (ie trimmed and lower-cased) category name.
+     *
+     * We desensitize the name because we want to treat names like "NAME" and "name  " as equivalent, which is useful
+     * when importing data since user input errors are common. In other scenarios this equivalency might not be useful
+     * (we might want to treat "name" and "NAME" as different), so make sure to be intentional when using this class.
+     */
+    private class CategoriesByDesensitizedName {
```
ewaterman, line 918: I opted to make this a class because it allowed for clearer usage of the map. Users of the class don't need to remember to desensitize the data, it happens automatically. All they need to do is provide the category for puts, and category name for gets

PR #5437, opened 2025-08-08
OBPIH-7440 Frontend pagination and virtualization improvements for count and recount workflows

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7440

**Description:**
In this ticket, the main changes include fixes to the "counted by / recounted by" and "date counted / date recounted" selects. Now, when selecting a new date, only one table re-renders instead of all of them as it did before. Additionally, I improved the arrow navigation, so we no longer need to call the resetFocus function in multiple places, which also results in fewer re-renders.

---

### DIFF
--- a/src/js/hooks/cycleCount/useCountStepTable.jsx
+++ b/src/js/hooks/cycleCount/useCountStepTable.jsx
@@ -30,16 +30,12 @@ const useCountStepTable = ({
   formatLocalizedDate,
   addEmptyRow,
   triggerValidation,
-  refreshFocusCounter,
   isFormDisabled,
+  forceRerender,
 }) => {
   const columnHelper = createColumnHelper();
   const [rowIndex, setRowIndex] = useState(null);
   const [columnId, setColumnId] = useState(null);
-  // If prevForceResetFocus is different from refreshFocusCounter,
-  // it triggers a reset of rowIndex and columnId.
-  const [prevForceResetFocus, setPrevForceResetFocus] = useState(0);
-
   const translate = useTranslate();
 
   const { users, currentLocation, binLocations } = useSelector((state) => ({
@@ -47,14 +43,6 @@ const useCountStepTable = ({
     binLocations: state.cycleCount.binLocations,
     currentLocation: state.session.currentLocation,
   }));
-
-  useEffect(() => {
-    if (refreshFocusCounter !== prevForceResetFocus) {
-      setRowIndex(null);
-      setColumnId(null);
-      setPrevForceResetFocus(refreshFocusCounter);
-    }
-  }, [refreshFocusCounter]);
 
   const showBinLocation = useMemo(() =>
     checkBinLocationSupport(currentLocation.supportedActivities), [currentLocation?.id]);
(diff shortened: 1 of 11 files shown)


src/js/hooks/cycleCount/useCountStepTable.jsx
```diff
@@ -195,6 +183,37 @@ const useCountStepTable = ({
...
         }
+        const handleClick = (event) => {
+          if (rowIndex === null && columnId === null) {
+            return;
+          }
+
+          const { target } = event;
+          const isInputElement = target.closest('input, select, textarea, .date-field-input, .react-datepicker, .react-select__control');
+          const isDatePickerDay = target.closest('.react-datepicker__day');
+          const isSelectOption = target.closest('.react-select__option');
```
kchelstowski: please add comments explaining what this code is responsible for - how does it fix performance issues etc.
Additionally, the naming doesn't feel quite right here:

- `isSelectOption` - is selection option what?
- `isDatePickerDay` - what exactly? is it present? is it picked? is it empty?

src/js/hooks/cycleCount/useResolveStep.js
```diff
@@ -533,6 +530,9 @@ const useResolveStep = () => {
   }) => {
     try {
       show();
+      // Before saving, we need to change the state updated to true for all items that were changed
+      itemsToUpdate.current.map((cycleCountId) => setAllItemsUpdatedState(cycleCountId, true));
+      itemsToUpdate.current = [];
```
kchelstowski: it feels to be done here. I mean - in one line we map the items, in another line we clear them, below we use completely different items to proceed the save with.
Why couldn't we keep the `markAllItemsAsUpdated` method?

src/js/hooks/cycleCount/useResolveStepTable.jsx
```diff
@@ -270,6 +259,37 @@ const useResolveStepTable = ({
...
+          if (isDatePickerDay || isSelectOption) {
+            setRowIndex(null);
+            setColumnId(null);
+          }
+        };
+        document.addEventListener('click', handleClick);
+
+        return () => {
+          document.removeEventListener('click', handleClick);
+        };
```
kchelstowski, line 301: the same as in the previous hook - add more comments explaining how it really solves the performance issue.

kchelstowski (changes requested):

src/js/hooks/cycleCount/useResolveStep.js
SebastianLib: We used the `markAllItemsAsUpdated` function when we changed something in counted by / recounted by, date counted / date recounted.

Now, instead of storing this information in useState, we keep it in a useRef, which means that when we change the date or the users, we don’t trigger a re-render caused by useState - and as a result, all tables are not unnecessarily re-rendered.

The problem was that during `markAllItemsAsUpdated` we set the updated property in tableData to true, but the components `useCountStepTable / useResolveStepTable` didn’t always receive the latest information, because there was no re-render (because tableData is also stored by useRef). This led to a situation where, when using arrow navigation and typing something, the entered value wasn’t saved until a re-render occurred. That was actually the hardest part to solve in this ticket, because we already had many issues with these forms before.

I came up with the idea that we don’t actually need to set updated to true in each item right away - instead, we can store this in an itemsToUpdate ref until we perform a save. Now, when we save and set all items to updated, we also want to clear the itemsToUpdate array.

kchelstowski: I understood the purpose of that, my point was that it should have been called in a separate method to understand the flow.
Moreover - since in 534th line you use `.map`, which is an immutable method and you don't use the returned value further. I'd suggest using `.forEach` in that case or `for of` loop.
`itemsToUpdate` also don't feel quite right, as some items might not be present in this array, but are still supposed to be saved (e.g. you change the quantity counted). I'd suggest maybe something like `cycleCountsMarkedToUpdate`?

PR #5439, opened 2025-08-08
OBPIH-7438 Move validation logic before transaction processing (improvement)

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7438

**Description:** 
The issue with disappearing lines after validation on missing lot displays had already been fixed earlier. The problem now was that if someone, for example, forgot to enter the lot number but did enter the expiration date, they would correctly get an error saying the lot number is missing, but we were still creating a baseline. So on the second attempt, if they tried to save within the same minute, they would receive an error message saying "A transaction already exists at that time..." I’ve now added validation before creating baselines, etc., because it’s unnecessary to proceed if there’s a validation error.
---

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
@@ -1373,6 +1373,29 @@ class InventoryService implements ApplicationContextAware {
     RecordInventoryCommand saveRecordInventoryCommand(RecordInventoryCommand cmd, Map params) {
         log.debug "Saving record inventory command params: " + params
 
+        // First, we need to check if there is no validation error (OBPIH-7438)
+        cmd.recordInventoryRows.each { RecordInventoryRowCommand row ->
+            if (!row) {
+                return
+            }
+
+            if (row.expirationDate && !row.lotNumber) {
+                cmd.errors.reject("inventoryItem.invalid", "Items with an expiry date must also have a lot number")
+                row.error = true
+                return cmd
+            }
+
+            if (cmd.product && cmd.product.lotAndExpiryControl && (!row.expirationDate || !row.lotNumber)) {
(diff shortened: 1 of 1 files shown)


kchelstowski (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1373,6 +1373,29 @@ class InventoryService implements ApplicationContextAware {
     RecordInventoryCommand saveRecordInventoryCommand(RecordInventoryCommand cmd, Map params) {
         log.debug "Saving record inventory command params: " + params
 
+        // First, we need to check if there is no validation error (OBPIH-7438)
```
ewaterman, line 1376: I suspect that this is due to database transactions. If there are errors, we set them on the command via `cmd.errors.reject` and return the command to the controller (and let the controller throw the actual exceptions). But since the service is annotated with `@Transactional`, because no actual exception is thrown from the service class, when we return to the controller it persists everything it has so far to the database.

Your change is fine for now, but I think we should eventually refactor this flow to properly throw exceptions directly within the service itself so that rollbacks happen.

Setting errors on the command via `cmd.errors.reject` then returning the command will always result in any data persistence up to that point to not be rolled back.

ewaterman (approved):

grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
kchelstowski: @ewaterman usually it's fine to set errors via command like this, but in the end we should always do something like:
```groovy
if (cmd.hasErrors()) {
  throw new ValidationException(...)
}
```
by doing so, we would actually throw 500 and make the rollback of the transaction.
I think this feature is a bit more complicated as we sometimes want to allow some actions even if some errors are visible to a user.

PR #5440, opened 2025-08-08
OBPIH-7444 fail creating adjustment when transaction already exists a…

ewaterman:
…t time

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7444

**Description:** If we do a cycle count or record stock or inventory import and set quantity to zero for a product, then next CC or RS or II on that product won't have a baseline (because we don't create baselines if quantity == 0, this is something we're going to change in 0.9.6). Because of that, we can successfully create multiple adjustment transactions at the exact same time. To fix this, we need to also check that there are no transactions at the time of the adjustments.

---

Video showing the bug: https://jam.dev/c/6005951b-9362-454e-8e7b-bc64e31a24a6

Video shows that after this fix, if we do a record stock to set quantity to zero, then an inventory import at the same time should fail, even if it's on different items of the product.

[Screencast from 2025-08-08 11:33:34 AM.webm](https://github.com/user-attachments/assets/e9c1a961-8820-46db-bcfd-43716b85c65c)

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
@@ -1389,8 +1389,7 @@ class InventoryService implements ApplicationContextAware {
 
             // We'd have weird behaviour if we allowed two transactions to exist at the same exact time (precision at the
             // database level is to the second) so fail if there's already a transaction on the items for the given date.
-            List<InventoryItem> inventoryItems = availableItems.values().collect{ it.inventoryItem }
-            if (hasTransactionEntriesOnDate(currentLocation, adjustmentTransactionDate, inventoryItems)) {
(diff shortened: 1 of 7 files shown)


grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -2726,25 +2731,31 @@ class InventoryService implements ApplicationContextAware {
...
         def criteria = TransactionEntry.createCriteria()
         return criteria.list {
-            'in'("inventoryItem", inventoryItems)
+            // TransactionEntry.product is optional, so we need to check the inventory item of the entry as well.
+            or {
+                inList("product", products)
+                inventoryItem {
+                    inList("product", products)
+                }
+            }
```
ewaterman, line 2758: I realized that we should have been checking products, not inventory items. If we only check inventory items, we don't capture the case where we have a transaction at the same time for the same product but with different items.

Seems obvious in hindsight...

PR #5441, opened 2025-08-11
OBPIH-7449 Correct Most Recent Stock Count date in Export Latest Inventory Date report

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7449

**Description:**

---

### DIFF
--- a/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
@@ -26,6 +26,7 @@ import java.text.SimpleDateFormat
 
 @Transactional(readOnly=true)
 class DashboardService {
+
     ConfigService configService
     def productAvailabilityService
 
@@ -337,8 +338,8 @@ class DashboardService {
         List<Object[]> latestInventoryDatesList = TransactionEntry.executeQuery("""
                 select ii.product.id, max(t.transactionDate)
                 from TransactionEntry as te
-                join te.inventoryItem as ii
-                join te.transaction as t
+                left join te.inventoryItem as ii
+                left join te.transaction as t
                 where t.inventory = :inventory
                 and t.transactionType.id in (:transactionTypeIds)
                 group by ii.product.id
--- a/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
@@ -335,19 +335,19 @@ class DashboardService {
 
         List<String> transactionTypes = configService.getProperty('openboxes.inventoryCount.transactionTypes', List) as List<String>
 
-        List<Object[]> latestInventoryDatesList = TransactionEntry.executeQuery("""
+        List<Object[]> latestInventoryDates = TransactionEntry.executeQuery("""
                 select ii.product.id, max(t.transactionDate)
                 from TransactionEntry as te
                 left join te.inventoryItem as ii
                 left join te.transaction as t
                 where t.inventory = :inventory
                 and t.transactionType.id in (:transactionTypeIds)
-                group by ii.product.id
+                group by ii.product
                 """,
                 [inventory: location.inventory, transactionTypeIds: transactionTypes])
 
         // Convert to map
-        Map<String, Timestamp> latestInventoryDates = latestInventoryDatesList.collectEntries { [it[0], it[1]] }
+        Map<String, Timestamp> latestInventoryDateMap = latestInventoryDates.collectEntries { [it[0], it[1]] }
 
         Map<Product, InventoryLevel> inventoryLevelMap = InventoryLevel
                 .findAllByInventory(location.inventory)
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -333,20 +336,30 @@ class DashboardService {
...
-                select ii.product.id, max(t.transactionDate)
-                from TransactionEntry as te
-                left join te.inventoryItem as ii
-                left join te.transaction as t
-                where t.inventory = :inventory
-                and t.transactionType.transactionCode in (:transactionCodes)
-                group by ii.product
-                """,
-                [inventory: location.inventory, transactionCodes: [TransactionCode.PRODUCT_INVENTORY, TransactionCode.INVENTORY]])
+        List transactionTypes = Holders.grailsApplication.config.openboxes.inventoryCount.transactionTypes
```
kchelstowski:
```suggestion
        List<String> transactionTypes = Holders.grailsApplication.config.openboxes.inventoryCount.transactionTypes
```

ewaterman: our preferred way of fetching properties is via the ConfigService:
```
configService.getProperty('openboxes.inventoryCount.transactionTypes', List<String>)
```
I don't actually know if it'll work with a typed list so you might need to do something like:
```
configService.getProperty('openboxes.inventoryCount.transactionTypes', List) as List<String>
```

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -333,20 +336,30 @@ class DashboardService {
...
+        Map<String, Timestamp> lastInventoryCountMap = TransactionEntry.createCriteria().list {
+            createAlias("inventoryItem", "ii")
+            createAlias("transaction", "t")
+            projections {
+                groupProperty("ii.product.id")
+                max("t.transactionDate")
+            }
+            eq("t.inventory", location.inventory)
+            inList("t.transactionType.id", transactionTypes)
+        }.collectEntries { [it[0], it[1]] }
```
ewaterman: Why change it to a createCriteria (vs the existing executeQuery)? Unless I'm not understanding something it looks like the behaviour is the same so can't you just change the executeQuery to use the transactionTypes variable?

I don't know if I have a preference between the two options but unless there's a good reason to change the behaviour I'd prefer to keep it however it was before because we're so close to release. Better to make as few changes as possible.

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -333,20 +336,30 @@ class DashboardService {
...
+        List transactionTypes = Holders.grailsApplication.config.openboxes.inventoryCount.transactionTypes
 
-
-        // Convert to map
-        Map<String, Timestamp> latestInventoryDateMap = latestInventoryDates.collectEntries { [it[0], it[1]] }
+        // Fetch latest inventory counts for all products in a single query to avoid N+1 query issue
+        // This improves performance by reducing the number of database calls (OBPIH-7449)
+        Map<String, Timestamp> lastInventoryCountMap = TransactionEntry.createCriteria().list {
+            createAlias("inventoryItem", "ii")
+            createAlias("transaction", "t")
```
ewaterman: do you even need these aliases? Can you just do something like: `eq("transaction.inventory", location.inventory)` and `groupProperty("inventoryItem.product.id")`?

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -333,20 +336,30 @@ class DashboardService {
...
+            projections {
+                groupProperty("ii.product.id")
+                max("t.transactionDate")
+            }
+            eq("t.inventory", location.inventory)
+            inList("t.transactionType.id", transactionTypes)
+        }.collectEntries { [it[0], it[1]] }
+
+        Map<String, Timestamp> lastCycleCountDateMap = CycleCountItem.createCriteria().list {
```
ewaterman: Why is this needed? If we do a cycle count, either we create a baseline or adjustment or both, so won't that be returned in the above?

kchelstowski: Sebastian was probably following how it was initially done in the Products report by Justin, where we compared last cycle count date to the last inventory date. I don't remember the context that was back then, but maybe we did not always create a transaction for the cycle count, but I agree with you this should be handled above, so @SebastianLib make sure that if you complete a cycle count, the date would be updated in the query above.

SebastianLib: Yes, what Kacper wrote above is correct. I also verified what you @ewaterman mentioned, and it does look like the date would indeed be updated in the query above

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
SebastianLib: I guess I might have unnecessarily written this using createAlias. I thought we usually prefer to use createAlias, but in this case it probably doesn’t make sense to rewrite it from scratch

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
SebastianLib: I guess I might have unnecessarily written this using createCriteria. I thought we usually prefer to use createCriteria, but in this case it probably doesn’t make sense to rewrite it from scratch

kchelstowski (approved):

PR #5442, opened 2025-08-13
OBPIH-7434 Fix performance issues in Products Changes report - materi…

kchelstowski:
…alize the helper views and optimize the main views

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
@@ -48,7 +48,7 @@ class Transaction implements Comparable, Serializable {
 
     def publishSaveEvent() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshProductAvailabilityEvent(this, forceRefresh))
-        Holders.grailsApplication.mainContext.publishEvent(new RefreshCycleCountReportViewsEvent(this))
+        Holders.grailsApplication.mainContext.publishEvent(new RefreshInventoryCountEvent(this))
     }
 
     def publishDeleteEvent() {
--- a/grails-app/migrations/views/inventory-counts-helper-views.sql
+++ b/grails-app/migrations/views/inventory-counts-helper-views.sql
@@ -1,4 +1,5 @@
-CREATE OR REPLACE TABLE adjustments_candidates AS
+DROP TABLE IF EXISTS adjustment_candidate;
+CREATE TABLE adjustment_candidate AS
     SELECT te.transaction_id AS transaction_id,
            ii.product_id as product_id,
            t.transaction_date,
@@ -9,9 +10,15 @@ CREATE OR REPLACE TABLE adjustments_candidates AS
              JOIN inventory_item ii ON ii.id = te.inventory_item_id
              JOIN location facility ON facility.inventory_id = t.inventory_id
     WHERE t.transaction_type_id = '3' -- adjustments
-    GROUP BY te.transaction_id, ii.product_id;
+    GROUP BY
+        te.transaction_id,
+        ii.product_id,
+        t.transaction_date,
+        t.inventory_id,
+        facility.id;
 
-CREATE OR REPLACE TABLE inventory_baseline_candidates AS
+DROP TABLE IF EXISTS inventory_baseline_candidate;
+CREATE TABLE inventory_baseline_candidate AS
     SELECT te.transaction_id AS transaction_id,
            ii.product_id as product_id,
            t.transaction_date,
@@ -22,18 +29,23 @@ CREATE OR REPLACE TABLE inventory_baseline_candidates AS
              JOIN inventory_item ii ON ii.id = te.inventory_item_id
              JOIN location facility ON facility.inventory_id = t.inventory_id
     WHERE t.transaction_type_id = '12' -- baseline inventory transaction
-    GROUP BY te.transaction_id, ii.product_id;
+    GROUP BY
+        te.transaction_id,
+        ii.product_id,
+        t.transaction_date,
+        t.inventory_id,
+        facility.id;
 
 -- Helps the TIMESTAMPDIFF match in baseline_adjustment_matches
 CREATE INDEX idx_inventory_product_date
-    ON inventory_baseline_candidates (inventory_id, product_id, transaction_date);
+    ON inventory_baseline_candidate (inventory_id, product_id, transaction_date);
 
 -- Helps matching by transaction_id in joins
 CREATE INDEX idx_transaction_id
-    ON inventory_baseline_candidates (transaction_id);
+    ON inventory_baseline_candidate (transaction_id);
 
 CREATE INDEX idx_product_inventory_date
-    ON adjustments_candidates (product_id, inventory_id, transaction_date);
+    ON adjustment_candidate (product_id, inventory_id, transaction_date);
 
 CREATE INDEX idx_product_inventory_base
-    ON inventory_baseline_candidates (product_id, inventory_id, transaction_date);
+    ON inventory_baseline_candidate (product_id, inventory_id, transaction_date);
--- a/grails-app/migrations/views/inventory-counts.sql
+++ b/grails-app/migrations/views/inventory-counts.sql
@@ -8,8 +8,8 @@ CREATE OR REPLACE VIEW inventory_counts AS
             b.transaction_id AS baseline_id,
             a.product_id AS product_id,
             a.facility_id AS facility_id
-        FROM adjustments_candidates a
-        JOIN inventory_baseline_candidates b
+        FROM adjustment_candidate a
+        JOIN inventory_baseline_candidate b
           ON a.product_id = b.product_id
           AND a.inventory_id = b.inventory_id
           AND TIMESTAMPDIFF(SECOND, a.transaction_date, b.transaction_date) = -1
--- a/grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
@@ -1,18 +1,29 @@
 package org.pih.warehouse.inventory
 
 import grails.gorm.transactions.Transactional
-import org.pih.warehouse.data.DataService
+import groovy.sql.Sql
+import org.pih.warehouse.core.Constants
+
+import javax.sql.DataSource
 
 @Transactional
 class InventoryCountService {
 
-    DataService dataService
+    DataSource dataSource
 
     void refreshAdjustmentCandidatesView(Inventory inventory, List<String> productIds, String transactionId, Date transactionDate) {
         productIds.each {
-            String dateString = transactionDate.format("yyyy-MM-dd HH:mm:ss")
+            String dateString = transactionDate.format(Constants.ISO_DATE_TIME_FORMAT)
+            Sql sql = new Sql(dataSource)
+            Map<String, Object> params = [
+                    transactionId: transactionId,
+                    productId: it,
+                    transactionDate: dateString,
+                    inventoryId: inventory.id,
+                    facilityId: inventory.warehouse.id
+            ]
             String query = """
-                INSERT INTO adjustments_candidates (
+                INSERT INTO adjustment_candidate (
                     transaction_id,
                     product_id,
                     transaction_date,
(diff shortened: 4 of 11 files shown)


grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -125,6 +125,12 @@ class UrlMappings {
             action = [GET: "getInventoryItem"]
         }
 
+        "/api/products/getLatestInventoryCountDate" {
+            controller = { "productApi" }
+            action = [GET: "getLatestInventoryCountDate"]
+        }
+
+
```
kchelstowski, line 133: an endpoint to get the last counted date. Sometimes it is expensive so to get additional few seconds I asked Manon if she feels OK with this column being fetched separately:
Peek 2025-08-13 14-16

grails-app/migrations/views/inventory-counts-helper-views.sql
```diff
@@ -0,0 +1,39 @@
...
+           ii.product_id as product_id,
+           t.transaction_date,
+           t.inventory_id,
+           facility.id as facility_id
+    FROM transaction_entry te
+             JOIN transaction t ON te.transaction_id = t.id
+             JOIN inventory_item ii ON ii.id = te.inventory_item_id
+             JOIN location facility ON facility.inventory_id = t.inventory_id
+    WHERE t.transaction_type_id = '12' -- baseline inventory transaction
+    GROUP BY te.transaction_id, ii.product_id;
```
kchelstowski: I materialized two subviews from the `inventory-counts` view. Sometimes the amount of data was that huge to be calculated in memory (via subquery) that it sometimes crashed the DB. The only way to make it work was to materialize some of them for now.

grails-app/migrations/views/inventory-counts.sql
```diff
@@ -88,23 +64,25 @@ CREATE OR REPLACE VIEW inventory_counts AS
...
             ii.product_id AS product_id,
             facility.id AS facility_id,
             t.transaction_date AS date_recorded,
             'BASELINE' as inventory_count_type_code,
             -- Alone baseline doesn't have any associated transaction (like baseline + adjustment), so hardcode it to NULL
             NULL as associated_transaction_id
-        FROM transaction t
-        JOIN transaction_entry te ON te.transaction_id = t.id
+        FROM transaction_entry te
+        JOIN transaction t ON t.id = te.transaction_id
```
kchelstowski, line 76: it was one of my first changes here - I don't know if it's needed anymore, but back then, when I started working on this ticket, changing the granularity here helped a bit. Since it is working fine now, I didn't want to try to make it work again via `FROM transaction`.

grails-app/migrations/views/inventory-counts.sql
```diff
@@ -88,23 +64,25 @@ CREATE OR REPLACE VIEW inventory_counts AS
...
         JOIN inventory_item ii ON ii.id = te.inventory_item_id
         JOIN location facility ON facility.inventory_id = t.inventory_id
+        LEFT JOIN transaction adjustment ON TIMESTAMPDIFF(SECOND, t.transaction_date, adjustment.transaction_date) = 1
         WHERE t.transaction_type_id = '12' -- baseline inventory transaction
-          AND NOT EXISTS (
+        AND NOT EXISTS (
               SELECT 1
               FROM baseline_adjustment_matches bam
               WHERE bam.baseline_id = t.id
+              AND bam.product_id = ii.product_id
```
kchelstowski, line 84: This is also a huge improvement and also a fix of a bug - when we do the inventory import, we actually create a baseline and adjustment for all items in the file, which means we can have multiple matches for one product (as X baselines contain this product) where X is amount of distinct items in the file, so by looking at the `baseline_id` only, we might have not matched the proper adjustment and vice versa.

src/js/hooks/cycleCount/useLastCountedDate.js
```diff
@@ -0,0 +1,36 @@
...
+
+import { PRODUCT_API } from 'api/urls';
+import apiClient from 'utils/apiClient';
+
+const useLastCountedDate = (tableData, loading) => {
+  const [lastCountedDateMap, setLastCountedDateMap] = useState({});
+
+  const fetchLatestInventoryCountDate = async () => {
+    if (tableData.data.length) {
+      const res = await apiClient.get(`${PRODUCT_API}/getLatestInventoryCountDate`, {
```
kchelstowski: todo: add a frontend api for that

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
jmiranda: Nice! This is the data I've wanted to materialize due to the performance issues on the All Products tab. I like the idea of making this a separate request for now. If we decide to materialize this data as well (which we probably should) we need to be thoughtful. There are a number of events that will cause the date last counted to require refreshing. We should create a separate ticket for this scope so it gets a proper investigation.

grails-app/migrations/views/inventory-counts-helper-views.sql
jmiranda: How long does it take to create and refresh these tables?

grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
```diff
@@ -0,0 +1,58 @@
...
+                    transaction_date,
+                    inventory_id,
+                    facility_id
+                )
+                VALUES (
+                    '${transactionId}',
+                    '${it}',
+                    '${dateString}',
+                    '${inventory.id}',
+                    '${inventory.warehouse.id}'
```
jmiranda: Vulnerable to SQL injection. Use named parameters here.

grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
```diff
@@ -0,0 +1,58 @@
...
+import org.pih.warehouse.data.DataService
+
+@Transactional
+class InventoryCountService {
+
+    DataService dataService
+
+    void refreshAdjustmentCandidatesView(Inventory inventory, List<String> productIds, String transactionId, Date transactionDate) {
+        productIds.each {
+            String dateString = transactionDate.format("yyyy-MM-dd HH:mm:ss")
```
jmiranda: Move the format to a constant and give this a name. You actually shouldn't need this if the 

Is there a reason we can't do a INSERT INTO SELECT here?

grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
```diff
@@ -0,0 +1,58 @@
...
+                    transaction_date,
+                    inventory_id,
+                    facility_id
+                )
+                VALUES (
+                    '${transactionId}',
+                    '${it}',
+                    '${dateString}',
+                    '${inventory.id}',
+                    '${inventory.warehouse.id}'
```
jmiranda: SQL injection

grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
```diff
@@ -48,6 +48,7 @@ class Transaction implements Comparable, Serializable {
 
     def publishSaveEvent() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshProductAvailabilityEvent(this, forceRefresh))
+        Holders.grailsApplication.mainContext.publishEvent(new RefreshCycleCountReportViewsEvent(this))
```
jmiranda: Shouldn't we be updating this when transaction deletes are happening? Or a transaction entry has been deleted for a particular product? Are there other events we need to watch for as well?

src/main/groovy/org/pih/warehouse/inventory/RefreshCycleCountReportViewsEvent.groovy
```diff
@@ -0,0 +1,26 @@
...
+
+class RefreshCycleCountReportViewsEvent extends ApplicationEvent {
+
+    Inventory inventory
+    List<String> productIds
+    Date transactionDate
+    String transactionId
+    String transactionTypeId
+
+    RefreshCycleCountReportViewsEvent(Transaction source) {
```
jmiranda: Let's call this RefreshCycleCountTransactionEvent.

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
ewaterman: I like this idea a lot in general. If we build our APIs to fix exactly what a page needs, it makes them inflexible. Breaking APIs down to serve individual "components" allows them to be re-used where needed and we can do exactly this kind of stuff (querying for different components independently in parallel then merging the result to build a page).

Obviously we don't want to overdo it and put too much responsibility on the frontend to build what it needs, but in this case I like it

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1479,4 +1484,39 @@ class ProductService {
...
+    Map<String, Timestamp> latestInventoryDateForProducts(List<String> productIds) {
+        return latestInventoryDateForProducts(AuthService.currentLocation, productIds)
+    }
+
+    Map<String, Timestamp> latestInventoryDateForProducts(Location location, List<String> productIds) {
+        List<String> transactionTypeIds = [
+                Constants.ADJUSTMENT_CREDIT_TRANSACTION_TYPE_ID,
+                Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID,
+                Constants.INVENTORY_BASELINE_TRANSACTION_TYPE_ID
+        ]
```
ewaterman: better to use `openboxes.inventoryCount.transactionTypes` instead

src/main/groovy/org/pih/warehouse/inventory/RefreshCycleCountReportViewsEvent.groovy
ewaterman: Why not RefreshInventoryCountEvent? This logic isn't specific to cycle counts since it triggers for all transactions.

Other options:
- RefreshInventoryReportsEvent
- InventoryTransactionEvent

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -1479,4 +1484,39 @@ class ProductService {
         }
         return synonym
     }
+
+    Map<String, Timestamp> latestInventoryDateForProducts(List<String> productIds) {
+        return latestInventoryDateForProducts(AuthService.currentLocation, productIds)
+    }
+
+    Map<String, Timestamp> latestInventoryDateForProducts(Location location, List<String> productIds) {
```
ewaterman, line 1494: Why Timestamp and not Date? The field in Transaction is Date

grails-app/services/org/pih/warehouse/inventory/InventoryCountService.groovy
```diff
@@ -0,0 +1,58 @@
...
+                )
+            """
+
+            dataService.executeStatement(query, true)
```
ewaterman: If you added the AdjustmentCandidates and InventoryBaselineCandidates domain classes, wouldn't you be able to do inserts like:
```
new AdjustmentCandidates(...).save()
```
Or does it not work that way for materialized views?

grails-app/migrations/views/inventory-counts-helper-views.sql
```diff
@@ -0,0 +1,39 @@
+CREATE OR REPLACE TABLE adjustments_candidates AS
```
ewaterman: Our tables are usually named in singular format. `adjustment_candidate` and `inventory_baseline_candidate`.

Not hugely important though

grails-app/services/org/pih/warehouse/report/ReportService.groovy
```diff
@@ -1170,6 +1173,17 @@ class ReportService implements ApplicationContextAware {
...
+        long inventoryCountTime = System.currentTimeMillis()
+        List<Object[]> inventoryCountList = InventoryCount.createCriteria().list {
+            projections {
+                rowCount()
+                groupProperty("product")
+            }
+            eq("facility", org.pih.warehouse.auth.AuthService.currentLocation)
+            inList("product", results.collect { it[1] })
+            between("dateRecorded", command.startDate, command.endDate)
+        }
```
ewaterman, line 1185: so you're just returning row count and product? Why not make it a map:
```
InventoryCount.createCriteria().list { ... }.collectEntries { [it[1], it.[0] }
```
Then later you can just key into the map instead of searching the list.

SebastianLib (approved):

PR #5449, opened 2025-08-18
OBPIH-7455 Fix before/after qty issues in the inventory transaction r…

kchelstowski:
…eport - include transfer in/out, materialize helper views

**Description:**

---

### DIFF
--- a/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
+++ b/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
@@ -12,7 +12,7 @@ SELECT
 FROM (
          SELECT id, transaction_date, inventory_id, transaction_type_id
          FROM transaction
-         WHERE transaction_type_id IN ('3','8','9')
+         WHERE transaction_type_id IN ('3','8','9') -- adjustments, transfer ins, transfer outs
            AND transaction.order_id IS NULL
      ) t
          JOIN transaction_entry te ON te.transaction_id = t.id
--- a/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
+++ b/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
@@ -18,9 +18,12 @@ FROM (
          JOIN transaction_entry te ON te.transaction_id = t.id
          JOIN inventory_item ii ON ii.id = te.inventory_item_id
          JOIN product p ON p.id = ii.product_id
-GROUP BY t.id, ii.product_id;
-
-CREATE INDEX IF NOT EXISTS idx_product_code_inventory on inventory_movement_summary(product_code, inventory_id);
+GROUP BY
+    t.id,
+    ii.product_id,
+    t.transaction_date,
+    p.product_code,
+    t.inventory_id;
 
 CREATE TABLE IF NOT EXISTS product_inventory_summary AS (
     SELECT
@@ -37,10 +40,10 @@ CREATE TABLE IF NOT EXISTS product_inventory_summary AS (
              JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
              JOIN product ON product.id = inventory_item.product_id
     WHERE transaction_type.id = '12' -- baseline inventory transaction
-    GROUP BY transaction.id, inventory_item.product_id
+    GROUP BY
+        transaction.id,
+        inventory_item.product_id,
+        product.product_code,
+        facility.id,
+        transaction.transaction_date
 );
-
-
-CREATE INDEX IF NOT EXISTS idx_product_code_facility on product_inventory_summary(product_code, facility_id, baseline_transaction_date);
-CREATE INDEX IF NOT EXISTS idx_product_id_facility on product_inventory_summary(product_id, facility_id, baseline_transaction_date);
-CREATE INDEX IF NOT EXISTS idx_product_inventory_summary_lookup ON product_inventory_summary (product_code, facility_id, baseline_transaction_date DESC);
--- a/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
+++ b/grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
@@ -1,0 +1,46 @@
+CREATE TABLE IF NOT EXISTS inventory_movement_summary AS
+SELECT
+    ii.product_id,
+    t.transaction_date,
+    p.product_code,
+    t.inventory_id,
+    t.id AS transaction_id,
+    SUM(
+        -- If it's outbound (transfer out), negate the quantity, as its entries have positive quantity
+        CASE WHEN t.transaction_type_id = '9' THEN -te.quantity ELSE te.quantity END
+    ) AS quantity_sum
+FROM (
+         SELECT id, transaction_date, inventory_id, transaction_type_id
+         FROM transaction
+         WHERE transaction_type_id IN ('3','8','9')
+           AND transaction.order_id IS NULL
+     ) t
+         JOIN transaction_entry te ON te.transaction_id = t.id
+         JOIN inventory_item ii ON ii.id = te.inventory_item_id
+         JOIN product p ON p.id = ii.product_id
+GROUP BY t.id, ii.product_id;
+
+CREATE INDEX IF NOT EXISTS idx_product_code_inventory on inventory_movement_summary(product_code, inventory_id);
+
+CREATE TABLE IF NOT EXISTS product_inventory_summary AS (
+    SELECT
+        inventory_item.product_id as product_id,
+        product.product_code as product_code,
+        facility.id as facility_id,
+        transaction.id as transaction_id,
+        transaction.transaction_date as baseline_transaction_date,
+        SUM(transaction_entry.quantity) as quantity_balance
+    FROM transaction_entry
+             JOIN transaction ON transaction.id = transaction_entry.transaction_id
+             JOIN transaction_type ON transaction.transaction_type_id = transaction_type.id
+             JOIN location facility ON transaction.inventory_id = facility.inventory_id
+             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
+             JOIN product ON product.id = inventory_item.product_id
+    WHERE transaction_type.id = '12' -- baseline inventory transaction
+    GROUP BY transaction.id, inventory_item.product_id
+);
+
+
+CREATE INDEX IF NOT EXISTS idx_product_code_facility on product_inventory_summary(product_code, facility_id, baseline_transaction_date);
+CREATE INDEX IF NOT EXISTS idx_product_id_facility on product_inventory_summary(product_id, facility_id, baseline_transaction_date);
+CREATE INDEX IF NOT EXISTS idx_product_inventory_summary_lookup ON product_inventory_summary (product_code, facility_id, baseline_transaction_date DESC);
--- a/grails-app/migrations/views/inventory-transactions-summary.sql
+++ b/grails-app/migrations/views/inventory-transactions-summary.sql
@@ -1,37 +1,4 @@
 CREATE OR REPLACE VIEW inventory_transactions_summary AS
-       WITH product_inventory_summary AS (
-           SELECT
-              inventory_item.product_id as product_id,
-              product.product_code as product_code,
-              facility.id as facility_id,
-              transaction.id as transaction_id,
-              transaction.transaction_date as baseline_transaction_date,
-              SUM(transaction_entry.quantity) as quantity_balance
-            FROM transaction_entry
-            JOIN transaction ON transaction.id = transaction_entry.transaction_id
-            JOIN transaction_type ON transaction.transaction_type_id = transaction_type.id
-            JOIN location facility ON transaction.inventory_id = facility.inventory_id
-            JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
-            JOIN product ON product.id = inventory_item.product_id
-            WHERE transaction_type.id = '12' -- baseline inventory transaction
-            GROUP BY transaction.id, inventory_item.product_id
-    ),
-    -- Helper view to calculate quantity per adjustment transaction and product
-    quantity_sum_for_adjustment AS (
-       SELECT ii.product_id as product_id,
-              t.transaction_date as transaction_date,
-              product.product_code as product_code,
-              t.inventory_id as inventory_id,
-              t.id as transaction_id,
-              SUM(te.quantity) as quantity_sum
-           FROM transaction t
-                    JOIN transaction_entry te ON te.transaction_id = t.id
-                    JOIN inventory_item ii ON ii.id = te.inventory_item_id
-                    JOIN product ON product.id = ii.product_id
-           WHERE t.transaction_type_id = '3'
-           GROUP BY t.id, ii.product_id
-       )
-
 SELECT
     CRC32(CONCAT(transaction.id, inventory_item.product_id, facility.id)) as id,
     transaction.id AS transaction_id,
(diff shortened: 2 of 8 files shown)


grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
```diff
@@ -0,0 +1,46 @@
...
+             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
+             JOIN product ON product.id = inventory_item.product_id
+    WHERE transaction_type.id = '12' -- baseline inventory transaction
+    GROUP BY transaction.id, inventory_item.product_id
+);
+
+
+CREATE INDEX IF NOT EXISTS idx_product_code_facility on product_inventory_summary(product_code, facility_id, baseline_transaction_date);
+CREATE INDEX IF NOT EXISTS idx_product_id_facility on product_inventory_summary(product_id, facility_id, baseline_transaction_date);
+CREATE INDEX IF NOT EXISTS idx_product_inventory_summary_lookup ON product_inventory_summary (product_code, facility_id, baseline_transaction_date DESC);
```
kchelstowski: This migration materializes subviews from `inventory-transaction-summary.sql` view.
One weakness of that is that an initial migration of those, take about 12 minutes to run, so I decided not to rebuild them everytime the app is run, but to rely only on the manual refreshes and monitor the eventual problems.

grails-app/migrations/views/inventory-transactions-summary.sql
```diff
@@ -66,18 +33,27 @@ SELECT
...
+                        AND product_inventory_summary.facility_id = facility.id
+                        AND product_inventory_summary.baseline_transaction_date < transaction.transaction_date
+                     ) IS NULL
+                         OR inventory_movement_summary.transaction_date > (
+                         SELECT MAX(baseline_transaction_date)
+                         FROM product_inventory_summary
+                         WHERE product_inventory_summary.product_code = product.product_code
+                           AND product_inventory_summary.facility_id = facility.id
+                           AND product_inventory_summary.baseline_transaction_date < transaction.transaction_date
+                         )
```
kchelstowski, line 53: There was a bug:

> Qty Before and After in the inventory transaction report display incorrectly on adjust inventory transactions that are performed after adding stock to out of stock item through rs or ii, example:

Having such scenario, we might not have ANY baseline for a product, so this statement:
```sql
AND quantity_sum_for_adjustment.transaction_date > (
                    SELECT MAX(baseline_transaction_date)
                    FROM product_inventory_summary
                    WHERE product_inventory_summary.product_code = product.product_code -- comparing the product_code instead of inventory_item.product_id is intentional to avoid MariaDB bug when accessing the inventory_item.product_id with the table prefix (OBS-1901)
                      AND product_inventory_summary.facility_id = facility.id
                      AND product_inventory_summary.baseline_transaction_date < transaction.transaction_date
```
could eventually return `null`, and the whole sum of `null`+`<some value>` would be evaluated to 0 in the end.
So what really changed is checking if the date is `null` (if it is, it means none baseline exists before), and only if it's not null, take this date in the AND statement.

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -21,4 +32,100 @@ class InventoryTransactionSummaryService {
...
+    }
+
+    void refreshProductInventorySummaryView(RefreshInventoryTransactionsSummaryEvent event) {
+        Sql sql = new Sql(dataSource)
+        if (event.isDelete) {
+            Map<String, Object> params = [
+                    transactionId: event.transactionId
+            ]
+            String query = """
+                DELETE FROM product_inventory_summary WHERE transaction_id = :transactionId
```
kchelstowski, line 89: not tested yet, but should be working, as I tested the delete from above for `inventory-movement-summary` that is working

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -21,4 +32,100 @@ class InventoryTransactionSummaryService {
...
+            Timestamp transactionDate = new Timestamp(event.transactionDate.time)
+
+            Map<String, Object> params = [
+                    transactionId: event.transactionId,
+                    productId: product.id,
+                    transactionDate: transactionDate,
+                    inventoryId: event.inventoryId,
+                    productCode: product.productCode,
+                    quantitySum: entries.sum { entry ->
+                        event.transactionTypeId == Constants.TRANSFER_OUT_TRANSACTION_TYPE_ID ? -entry.quantity : entry.quantity
```
kchelstowski, line 57: if it's transfer out, we have to negate the quantity as its entries are positive (not like in the record stock where they are negative themselves)

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -21,4 +32,100 @@ class InventoryTransactionSummaryService {
...
+            Map<String, Object> params = [
+                    transactionId: event.transactionId
+            ]
+            String query = """
+                DELETE FROM product_inventory_summary WHERE transaction_id = :transactionId
+            """
+            sql.executeUpdate(params, query)
+            return
+        }
+        Location facility = Location.findByInventory(Inventory.read(event.inventoryId))
```
kchelstowski, line 94: ugly, will clean that up tomorrow

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -21,4 +32,100 @@ class InventoryTransactionSummaryService {
...
+        event.products.each { key, entries ->
+            Product product = Product.read(key)
+            Timestamp transactionDate = new Timestamp(event.transactionDate.time)
```
kchelstowski, line 96: will add a comment about that - basically we need to save it as `Timestamp` to make sure it uses the server/db timezone instead of user's system timezone.

grails-app/services/org/pih/warehouse/inventory/RefreshInventoryTransactionsSummaryEventService.groovy
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.ApplicationListener
+
+@Transactional
+class RefreshInventoryTransactionsSummaryEventService implements ApplicationListener<RefreshInventoryTransactionsSummaryEvent> {
+
+    InventoryTransactionSummaryService inventoryTransactionSummaryService
+
+    @Override
+    void onApplicationEvent(RefreshInventoryTransactionsSummaryEvent event) {
+        if (event.transactionTypeId == Constants.INVENTORY_BASELINE_TRANSACTION_TYPE_ID) {
```
ewaterman, line 14: why only baseline? What if there is a record stock (or whatever) that only results in an adjustment?

src/main/groovy/org/pih/warehouse/inventory/RefreshInventoryTransactionsSummaryEvent.groovy
```diff
@@ -0,0 +1,35 @@
+package org.pih.warehouse.inventory
+
+import org.springframework.context.ApplicationEvent
+
+class RefreshInventoryTransactionsSummaryEvent extends ApplicationEvent {
+
+    String transactionId
+
+    Map<String, List<TransactionEntry>> products
```
ewaterman: just "products" is a little misleading to me since that would imply a list IMO. Could do `entriesByProduct` or  `transactionEntriesByProduct`

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -21,4 +32,100 @@ class InventoryTransactionSummaryService {
...
+                    transactionId: event.transactionId
+            ]
+            String query = """
+                DELETE FROM inventory_movement_summary WHERE transaction_id = :transactionId
+            """
+            sql.executeUpdate(params, query)
+            return
+        }
+        event.products.each { key, entries ->
+            Product product = Product.read(key)
```
ewaterman: you could change the map to actually key on product `Map<Product, List<TransactionEntry>>` to avoid needing to fetch the product each time.

I assume there won't be too many products in a single transaction

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
ewaterman: I think we should take the time to understand the actual implications of this (I don't remember the behavioural differences between Date and Timestamp besides the fact that Timestamp is a child class of Date meant to represent the TIMESTAMP type).
Maybe this works because Timestamp understands better than Date the timezone configured in the jdbc url (vs simply using the system time) but I don't know.

I'm uncertain if we actually want to use Timestamp. I suggest you check the behaviour by changing your system settings around. My rule of thumb is whenever I test timezone stuff, I always make sure to set my system time to be the same as the jdbc url time.

So either change your database url from:
```
dataSource:
  url: jdbc:mysql://localhost:3306/openboxes?serverTimezone=UTC&useSSL=false
```
to
```
dataSource:
  url: jdbc:mysql://localhost:3306/openboxes?useSSL=false
```
so that it uses your local timezone, but IMO much better would be to just set your local timezone to UTC.

Things get wonky when local testing when your timezone is different from the jdbc url, but it also is wonky if you copy stage data to your local without changing your local tz to UTC since that data was all created in UTC. If you're not configured for UTC, the server will convert the data to your TZ which will make it look like its off by x hours. Testing in UTC simplifies all that stuff.

See here for some more context as to why this breaks: https://pihemr.atlassian.net/wiki/spaces/OB/pages/3544154117/Configuring+App+Timezone#%E2%9D%8C-App-Server-and-DB-Server-are-different

grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
```diff
@@ -0,0 +1,46 @@
...
+    t.inventory_id,
+    t.id AS transaction_id,
+    SUM(
+        -- If it's outbound (transfer out), negate the quantity, as its entries have positive quantity
+        CASE WHEN t.transaction_type_id = '9' THEN -te.quantity ELSE te.quantity END
+    ) AS quantity_sum
+FROM (
+         SELECT id, transaction_date, inventory_id, transaction_type_id
+         FROM transaction
+         WHERE transaction_type_id IN ('3','8','9')
```
ewaterman: 3 is CREDIT Adjustment right? I thought we weren't using that for movements, only for inventory events (record stock, cycle count...).

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
ewaterman: why do you need to delete first? In what scenario are you refreshing a transaction that already exists in the summary?

kchelstowski: it's not about deleting first. If I delete, I return, so I don't proceed with anything that is further.
If `isDelete` flag is true, then it's a delete, if it's not, then it's insert. 
Since the methods are quite the same, I decided to just add one `if` in the method to handle both cases.

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: >  suggest you check the behaviour by changing your system settings around.

I don't think changing my system settings is an ideal solution - my point was to make that date aligned with the server/system settings no matter what settings are used by a particular server and client. If I were to make this aligned with my settings, then this would break if a user/server used different settings. 
Having used `Timestamp` here I'm making sure the date will be persisted in the DB like any other dates via GORM, that recognize system's (DB's) timezone.

grails-app/services/org/pih/warehouse/inventory/RefreshInventoryTransactionsSummaryEventService.groovy
kchelstowski: because `product-inventory-summary` mv stores only baselines:
```sql
CREATE TABLE IF NOT EXISTS product_inventory_summary AS (
    SELECT
        inventory_item.product_id as product_id,
        product.product_code as product_code,
        facility.id as facility_id,
        transaction.id as transaction_id,
        transaction.transaction_date as baseline_transaction_date,
        SUM(transaction_entry.quantity) as quantity_balance
    FROM transaction_entry
             JOIN transaction ON transaction.id = transaction_entry.transaction_id
             JOIN transaction_type ON transaction.transaction_type_id = transaction_type.id
             JOIN location facility ON transaction.inventory_id = facility.inventory_id
             JOIN inventory_item ON inventory_item.id = transaction_entry.inventory_item_id
             JOIN product ON product.id = inventory_item.product_id
    WHERE transaction_type.id = '12' -- baseline inventory transaction
    GROUP BY transaction.id, inventory_item.product_id
);
```
adjustments are handled by the `inventory-movement-summary`.

src/main/groovy/org/pih/warehouse/inventory/RefreshInventoryTransactionsSummaryEvent.groovy
kchelstowski: yeah. this was one of the parts I was about to clean up today. Thanks for the immediate suggestions though.

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: yeah, initially I needed only id, and didn't change it further. Anyway the products are already fetched while calling the `groupBy`:
```groovy
source.transactionEntries?.groupBy { it.inventoryItem?.product?.id }
```
so this `.read` in the end does not cause any n+1, but I will change it to get rid of that line.

grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
kchelstowski: I will add comments, `3` has been here forever, what came in are `8` and `9` so transfer in/out.

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
ewaterman: I'm on the bus so I'm replying to this one comment then going to bed :sweat_smile: 

It's not about server and client (I agree we shouldn't care about client tz) it's about server and db. The server tz will be your local tz by default, and so will the tz of the database connection when serverTimezone is not specified in the jdbc url. If your local tz is UTC+2 but your jdbc url specifies serverTimezone=UTC, then there's a tz desync between the server and the db because the server says "I'm in UTC+2 but treat the data in the database as if it's UTC so shift the dates by 2 hours whenever doing a read". Dates are stored without a tz in the database and its the tz that we specify in the jdbc url that dictates what we assume the tz of those dates is (and if we don't set a serverTimezone then it uses the system local tz).

I'd argue that setting the server tz to one value and the db tz to a different one is a bad practice and leads to wonky behaviour as I mention in my discover that I linked above. That's why I suggest to change your system tz when testing tz stuff specifically. Because that changes the tz of your server to make it sync up with the data that exists in the db (which on our envs is store in UTC).

We can discuss more on a tech huddle call if you want but I just want to make sure we're careful about what types we're choosing to use. I'm not saying that Timestamp is a wrong choice, it could be correct, I just want to make sure we take the time to consider the implications of that choice fully

PR #5451, opened 2025-08-20
OBPIH-7460 Remove empty field options and prevent adding empty Tags, Catalogs, Cateogries etc.

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7460

**Description:**

---

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
@@ -47,26 +47,31 @@ class SelectOptionsApiController {
     def catalogOptions() {
         boolean hideNumbers = params.boolean("hideNumbers", false)
 
-        def catalogs = genericApiService.getList(ProductCatalog.class.simpleName, [sort: "name"]).collect {
-            String label = hideNumbers ? it.name : "${it.name} (${it?.productCatalogItems?.size()})"
-            [id: it.id, label: label]
-        }
+        def catalogs = genericApiService.getList(ProductCatalog.class.simpleName, [sort: "name"])
+                .findAll { it?.name }
+                .collect {
+                    [id: it.id, label: hideNumbers ? it.name : "${it.name} (${it?.productCatalogItems?.size()})"]
+                }
         render([data: catalogs] as JSON)
     }
 
     def categoryOptions() {
-        List<Category> categories = genericApiService.getList(Category.class.simpleName, [:]).collect {
-            [id: it.id, label: it.getHierarchyAsString(" > ")]
-        }
+        List<Category> categories = genericApiService.getList(Category.class.simpleName, [:])
+                .findAll { it.getHierarchyAsString(" > ") }
+                .collect {
+                    [id: it.id, label: it.getHierarchyAsString(" > ")]
+                }
         render([data: categories] as JSON)
     }
 
     def tagOptions() {
         boolean hideNumbers = params.boolean("hideNumbers", false)
 
-        def tags = genericApiService.getList(Tag.class.simpleName, [sort: "tag"]).collect {
-            [id: it.id, label: hideNumbers ? it.tag : "${it.tag} (${it?.products?.size()})"]
-        }
+        def tags = genericApiService.getList(Tag.class.simpleName, [sort: "tag"])
+                .findAll { it?.tag }
+                .collect {
+                    [id: it.id, label: hideNumbers ? it.tag : "${it.tag} (${it?.products?.size()})"]
+                }
         render([data: tags] as JSON)
     }
 
--- a/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
@@ -32,7 +32,7 @@ class SelectOptionsApiController {
 
     def glAccountOptions() {
         List<GlAccount> glAccounts = glAccountService.getGlAccounts(params)
-                .findAll { it?.code && it?.name }
+                .findAll { it?.code }
                 .collect {
                     [id: it.id, label: "${it.code} - ${it.name}"]
                 }
@@ -61,7 +61,7 @@ class SelectOptionsApiController {
 
     def categoryOptions() {
         List<Category> categories = genericApiService.getList(Category.class.simpleName, [:])
-                .findAll { it.getHierarchyAsString(" > ") }
+                .findAll { it.name }
                 .collect {
                     [id: it.id, label: it.getHierarchyAsString(" > ")]
                 }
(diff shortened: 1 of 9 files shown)


grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -31,49 +31,60 @@ class SelectOptionsApiController {
...
+                }
         render([data: catalogs] as JSON)
     }
 
     def categoryOptions() {
-        List<Category> categories = genericApiService.getList(Category.class.simpleName, [:]).collect {
-            [id: it.id, label: it.getHierarchyAsString(" > ")]
-        }
+        List<Category> categories = genericApiService.getList(Category.class.simpleName, [:])
+                .findAll { it.getHierarchyAsString(" > ") }
```
ewaterman: is this supposed to be `.findAll { it.name }`?

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -31,49 +31,60 @@ class SelectOptionsApiController {
...
 
     def productGroupOptions() {
-        List<ProductGroup> productGroups = genericApiService.getList(ProductGroup.class.simpleName, [:]).collect {
-            [id: it.id, label: "${it.name}"]
-        }
+        List<ProductGroup> productGroups = genericApiService.getList(ProductGroup.class.simpleName, [:])
+                .findAll { it?.name }
+                .collect {
+                    [id: it.id, label: "${it.name}"]
+                }
```
ewaterman, line 47: What if we've assigned products to the blank group? We wouldn't be able to see them in the list of options and so wouldn't be able to filter for them.

I understand that seeing blank groups/tags/categories/... doesn't make much sense but to me this is a data issue, not a filtering issue. If we have blanks in our data we should be fixing the data.

Now that you added all those `blank: false` constraints which make it impossible to create groups/tags/categories/... with a blank name/code, once this is merged we should go through our envs and fix the old, bad data so that we don't have blanks.

If we simply filter blanks out, then we'll still have the blanks in the system, they'll just be invisible, which is more confusing IMO.

TLDR: I'd rather revert this filtering and fix the data manually.

grails-app/domain/org/pih/warehouse/core/GlAccount.groovy
```diff
@@ -43,7 +43,7 @@ class GlAccount implements Serializable {
 
     static constraints = {
         code(nullable: false, blank: false, unique: true)
-        name(nullable: true)
+        name(nullable: true, blank: false)
```
ewaterman: Is this safe to do? It prevents saving all of these entities if the field is blank. This is obviously good for future data but what about existing data that already has a blank name?

For PIH envs we have very few blanks. On prod we have just 1 tag and 1 payment term so we can simply fix the data like I mentioned in my other comment. But for non-PIH managed environments they could have a lot more blanks.

If there are any features that update these tables behind the scenes, they'll fail with constraint errors. This is fine for things like edit product because the error is clear and they can simply set the name field for Tag or Attribute or Category or Catalog. But maybe there's a feature that does XYZ and then also updates these tables.

Ex: Feature XYZ, lets say inbound, allows you to pick a bunch of products and add stock to them, but then also calls save on each of the selected products and their attributes in case you updated any fields. A user picked a pre-existing product that had a blank attribute and now the feature fails for that product and they don't know why. They weren't updating the product or its attributes but now they can't proceed with feature XYZ.

It's probably fine (I can't think of any features that work like that) but maybe there's a case I'm not thinking of.

Anyways, maybe the error message is clear enough and people can simply go fix the blanks in their tables, but I'm scared to put a change like this into the release after almost all testing is already complete.

CC @awalkowiak if you have any thoughts. Maybe I'm being too cautious.

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
SebastianLib: What’s written now also works, but it’s probably better to filter the way you suggested

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
SebastianLib: I also think that filtering out blank records may not be the best idea, but when I asked about it in yesterday’s tech huddle channel nobody raised any objections to doing this filtering. I think it would be good if others also shared their thoughts on this so that we might find a better solution

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
```diff
@@ -31,49 +31,60 @@ class SelectOptionsApiController {
     UserService userService
 
     def glAccountOptions() {
-        List<GlAccount> glAccounts = glAccountService.getGlAccounts(params).collect {
-            [id: it.id, label: "${it.code} - ${it.name}"]
-        }
+        List<GlAccount> glAccounts = glAccountService.getGlAccounts(params)
+                .findAll { it?.code }
```
SebastianLib, line 35: I’m not sure if I’m being too cautious here with this findAll, because the code for GlAccount has had **blank: false** set for a long time already. So it shouldn’t happen, unlike in other places, that the code could be empty just because **blank: false** was missing before

grails-app/domain/org/pih/warehouse/core/GlAccount.groovy
SebastianLib: I removed this change (blank: false), because name was previously nullable. Maybe that was a mistake, and it should have always been (nullable: false), but for now Manon wants this to stay as it is (nullable: true). 

Referring to what you wrote we already discussed on TH that it doesn’t make sense to create tags, categories etc. without a name. There’s no real point in polluting the system with such data because they were most likely created by mistake and aren’t actually used.

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
ewaterman: as long as you can still see them on some "list" page. For example, do blank catalogs still show up on the list product catalogs page?
 
image

We want to be able to see them somewhere so that users can edit them to give them a name or delete them from the system. Otherwise they'll be in the system but inaccessible.

I'm going to approve the PR but could you double check that each of the catalogs, tags, categories... list pages still show blanks?

ewaterman (approved):

grails-app/controllers/org/pih/warehouse/api/SelectOptionsApiController.groovy
SebastianLib: I double-checked, and all blank catalogs, tags etc. still show up on the list pages. The changes I made in `SelectOptionsApiController.groovy` only affect the filters – empty options are no longer displayed there, but they remain visible on the list pages.

PR #5452, opened 2025-08-20
OBPIH-7451 Add validation for orderAdjustment invoices

SebastianLib:

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
+++ b/grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
@@ -81,10 +81,15 @@ class InvoiceItem implements Serializable {
         glAccount(nullable: true)
         budgetCode(nullable: true)
         quantity(nullable: false, min: 0, validator: { Integer quantity, InvoiceItem obj ->
+            // Order adjustments are not quantity-based, they should always have quantity = 1
+            if(obj.orderAdjustment && quantity > 1) {
+                return ['invoiceItem.invalidQuantity.label']
+            }
+
             // If the invoice is a prepayment or the item is an order adjustment,
             // the validation below doesn't make sense, because
             // we do not have shipmentItem at this point.
-            if (obj.invoice?.isPrepaymentInvoice || obj.orderAdjustment || obj.inverse) {
+            if (obj.invoice?.isPrepaymentInvoice || obj.inverse) {
                 return true
             }
 


grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -81,10 +81,15 @@ class InvoiceItem implements Serializable {
...
+            // Order adjustments are not quantity-based, they should always have quantity = 1
+            if(obj.orderAdjustment && quantity > 1) {
+                return ['invoiceItem.invalidQuantity.label']
+            }
+
             // If the invoice is a prepayment or the item is an order adjustment,
             // the validation below doesn't make sense, because
             // we do not have shipmentItem at this point.
-            if (obj.invoice?.isPrepaymentInvoice || obj.orderAdjustment || obj.inverse) {
+            if (obj.invoice?.isPrepaymentInvoice || obj.inverse) {
```
ewaterman, line 91: The comment above makes it seem like you should still have the `|| obj.orderAdjustment` here so that you return early if there's an adjustment. Is there a reason to remove it?

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -81,10 +81,15 @@ class InvoiceItem implements Serializable {
         glAccount(nullable: true)
         budgetCode(nullable: true)
         quantity(nullable: false, min: 0, validator: { Integer quantity, InvoiceItem obj ->
+            // Order adjustments are not quantity-based, they should always have quantity = 1
+            if(obj.orderAdjustment && quantity > 1) {
```
ewaterman: Should it be `quantity != 1`?

Also, I feel like a more proper fix for this would be to grey out the quantity field for adjustments on the frontend and not let you edit it at all but this is fine for now.

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
SebastianLib: oh I forgot to remove it from the comment 😆

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
SebastianLib: In theory, we could do it that way, but is there really a point? If I implemented it like that, I think entering 0 would trigger an error, which I don’t think is intended. So I believe the current approach is fine: when we enter 0, there’s no error and the line just gets removed on save. And if the value is negative, validation still throws an error anyway

awalkowiak (approved):

grails-app/domain/org/pih/warehouse/invoice/InvoiceItem.groovy
```diff
@@ -81,10 +81,14 @@ class InvoiceItem implements Serializable {
         glAccount(nullable: true)
         budgetCode(nullable: true)
         quantity(nullable: false, min: 0, validator: { Integer quantity, InvoiceItem obj ->
-            // If the invoice is a prepayment or the item is an order adjustment,
-            // the validation below doesn't make sense, because
+            // Order adjustments are not quantity-based, they should always have quantity = 1
+            if(obj.orderAdjustment && quantity > 1) {
```
alannadolny:
```suggestion
            if (obj.orderAdjustment && quantity > 1) {
```

PR #5458, opened 2025-08-25
OBLS-126 Create initial version of putaway task API

jmiranda:
Video describing current state of the putaway task API
https://app.screencastify.com/watch/sGBRQdjjlNDbnTslhkru

### DIFF
--- a/grails-app/controllers/org/pih/warehouse/api/putaway/PutawayTaskApiController.groovy
+++ b/grails-app/controllers/org/pih/warehouse/api/putaway/PutawayTaskApiController.groovy
@@ -1,6 +1,7 @@
 package org.pih.warehouse.api.putaway
 
 import grails.converters.JSON
+import grails.rest.RestfulController
 import org.pih.warehouse.api.StatusCategory
 import org.pih.warehouse.core.Location
 import org.pih.warehouse.product.Product
@@ -8,7 +9,7 @@ import org.pih.warehouse.putaway.PutawayTask
 import org.pih.warehouse.putaway.PutawayTaskService
 import org.springframework.http.HttpStatus
 
-class PutawayTaskApiController {
+class PutawayTaskApiController extends RestfulController<PutawayTask> {
 
     static responseFormats = ['json']
     static allowedMethods = [
@@ -23,6 +24,10 @@ class PutawayTaskApiController {
 
     PutawayTaskService putawayTaskService
 
+    PutawayTaskApiController() {
+        super(PutawayTask)
+    }
+
     def search(SearchPutawayTaskCommand command) {
         log.info "search params " + params
         // Return a not found error if request specifies a product, but the product does not exist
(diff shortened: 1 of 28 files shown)


grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+            default:
+                throw new UnsupportedOperationException("Unsupported action: ${action}")
+        }
+
+        log.info "task.toPutaway " + task.toPutaway().toJson()
+
+        // FIXME The putaway task is a view-backed domain, but Grails doesn't know that so any changes to the
+        //  task trigger a persistence event. I need to figure out how to make it a read-only domain class.
+        task.discard()
```
awalkowiak, line 122: Is not 
``` groovy
static mapping = {
    cache usage: "read-only"
}
```
what you need?

grails-app/controllers/org/pih/warehouse/api/putaway/PutawayTaskApiController.groovy
```diff
@@ -0,0 +1,84 @@
...
+            delete: 'DELETE',
+            generateForOrder: 'POST'
+    ]
+
+    PutawayTaskService putawayTaskService
+
+    def search(SearchPutawayTaskCommand command) {
+        log.info "search params " + params
+        // Return a not found error if request specifies a product, but the product does not exist
+        if (command.product && !command.product) {
```
awalkowiak: It could be useful to allow sending a product code or upc, and do a product search here too (if no product in the payload). Right now we have to hit the barcode product search api, and then this one. But it's fine if you want to keep it separated.

jmiranda: Right but that would be more BFF, so I was trying to avoid it unless we really needed it. I think if the binding mechanism for Product was such that id could be the id or a searchable barcode, then that would be fine with me.

grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
jmiranda: that's a cache setting. does that actually make the domain read only?

jmiranda: i'm getting weird errors deep in other parts of the code (i.e. when i complete a putaway) where gorm seem to be trying to persist the putaway task and none of the configurations i tried helped the situation.

awalkowiak: > that's a cache setting. does that actually make the domain read only?

iirc it tells that the data is immutable and you cannot save it, but it might not be a good fit if the data in the view changes often, I think we'd need to manually evict the cache, so maybe it is not a good idea.

PR #5460, opened 2025-08-25
OBLS-14 Make auto-receive to Inbound Sortation location

druchniewicz:


grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -528,4 +528,14 @@ class ReceiptService {
 
         saveAndCompletePartialReceipt(partialReceipt)
     }
+
+    Location findReceiveLocation(Shipment shipment) {
```
awalkowiak: `findReceivingLocation`

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -528,4 +528,14 @@ class ReceiptService {
...
+        def inboundSortationLocation = Location.findByName(Constants.INBOUND_SORTATION_LOCATION_NAME)
```
awalkowiak: I think I'd rather rely on the activity codes instead location name (needs to have receive stock, not have pick stock, anything else?). On top of that, you have to look for internal locations from a destination, because when doing a simple `findBy`, you can get internal locations from different depots that you are receiving into.

awalkowiak (changes requested):

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
awalkowiak: @druchniewicz Maybe for simplicity, for now, let's add an `InboundSortation` activity code instead defining a set of codes that need to be present

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
druchniewicz: fixed

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
druchniewicz: fixed in a new commit

grails-app/services/org/pih/warehouse/receiving/ReceiptService.groovy
```diff
@@ -528,4 +528,18 @@ class ReceiptService {
 
         saveAndCompletePartialReceipt(partialReceipt)
     }
+
+    Location findReceivingLocation(Shipment shipment) {
+        // TODO: below code returns first matching internal location that supports inbound sortation
```
jmiranda, line 533: I created the following ticket to deal with this TODO. 

https://openboxes.atlassian.net/browse/OBLS-130

druchniewicz: thanks for it

PR #5462, opened 2025-08-26
OBPIH-7462 Limit number of old product inventory transactions migrated at once

awalkowiak:
https://pihemr.atlassian.net/browse/OBPIH-7462

This change is purely for the locations that have a lot of transactions to be migrated, like the Distribution Center.

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
@@ -411,9 +411,15 @@ class MigrationService {
 
         // Find all transactions with the old Product Inventory type (by PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
         TransactionType oldProductInventoryType = TransactionType.get(Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
+
+        Integer batchSize = grailsApplication.config.openboxes.transactions.inventoryBaseline.migration.batchSize ?: 5000
+        boolean requiresBatching = Transaction.countByInventoryAndTransactionType(location.inventory, oldProductInventoryType) > batchSize
+
+        log.info("Fetching ${requiresBatching ? batchSize : 'all'} transactions candidates for migration")
         List<Transaction> transactions = Transaction.findAllByInventoryAndTransactionType(
-                location.inventory, oldProductInventoryType
-        )?.sort { [it.transactionDate, it.dateCreated] }?.reverse()
+                location.inventory, oldProductInventoryType,
+                [max: requiresBatching ? batchSize : -1, sort: [transactionDate: "desc", dateCreated: "desc"]]
+        )
 
         if (!transactions) {
             log.info("No transactions found for location ${location.name} with transaction type ${oldProductInventoryType.name}")
@@ -427,6 +433,7 @@ class MigrationService {
             User migratedBy = AuthService.currentUser
             // Split transactions into those with no entries, single product (record stock and single product's inventory import)
             // and those with multiple products (only inventory import)
+            log.info("Grouping transactions into empty, single product and multi-product transactions")
             List<Transaction> transactionsWithNoEntries = transactions.findAll { it.transactionEntries?.isEmpty() }
             List<Transaction> singleProductTransactions = (transactions - transactionsWithNoEntries).findAll {
                 (it.transactionEntries?.inventoryItem?.product?.unique()?.size() ?: 0) == 1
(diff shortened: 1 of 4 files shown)


grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -411,9 +411,15 @@ class MigrationService {
 
         // Find all transactions with the old Product Inventory type (by PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
         TransactionType oldProductInventoryType = TransactionType.get(Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)
+
+        Integer batchSize = grailsApplication.config.openboxes.transactions.inventoryBaseline.migration.batchSize ?: 5000
```
alannadolny: use `configService`

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -440,65 +447,91 @@ class MigrationService {
...
-            List<Product> shouldBeZeroedOut = products - currentInventoryBaseline?.transactionEntries?.collect { it.inventoryItem.product }?.unique()
-            if (cleanupAdjustmentEnabled && shouldBeZeroedOut) {
-                log.debug("Check if there is need for 'zeroing out' adjustments for products that had no stock before migration")
-
-                Map<String, AvailableItem> availableItems = productAvailabilityService.getAvailableItemsAtDateAsMap(
+            // differences for edge cases that were not addressed (if a product already has baseline as a most recent
+            // transaction, in this location, we can skip it)
+            // mostRecentTransactionTypeByProduct contains a list of [Product, transaction date, transaction type id]
+            List mostRecentTransactionTypeByProduct = inventoryService.getMostRecentTransactionTypeForProductsInInventory(location.inventory, allProducts)
+            List<Product> productsToBaseline = mostRecentTransactionTypeByProduct?.findAll { it[2] != Constants.INVENTORY_BASELINE_TRANSACTION_TYPE_ID }?.collect { it[0] } ?: []
```
alannadolny: What's the "it" here? Getting the elements using indexes looks weird

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -440,65 +447,91 @@ class MigrationService {
...
+            log.info("Creating inventory baseline for current stock for ${productsToBaseline?.size()} products found in old transactions")
+            Transaction currentInventoryBaseline = null
+            if (productsToBaseline) {
+                currentInventoryBaseline = productInventoryTransactionMigrationService.createInventoryBaselineTransaction(
```
alannadolny: Is the `currentInventoryBaseline` used after that if block? I can't see it, so maybe it's worth using ternary

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -440,65 +447,91 @@ class MigrationService {
...
+                                    "at location ${location.name}"
+                            processProductInventoryTransactions(transactionList, location, migratedBy, results, true)
+                        }
+                    }
+                }
+                log.info("Processing ${multiProductTransactions.size()} multi-product transactions")
+                processProductInventoryTransactions(multiProductTransactions, location, migratedBy, results)
+
+                // Zero out stock for products that were migrated but had no stock initially (hence hand no baseline created)
+                def cleanupAdjustmentEnabled = grailsApplication.config.openboxes.transactions.inventoryBaseline.migration.cleanupAdjustment
```
alannadolny: use `configService`

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -440,65 +447,91 @@ class MigrationService {
...
+            productAvailabilityService.triggerRefreshProductAvailability(location.id, allProducts?.id, true)
 
             createCSVMigrationReportForResults(location, results)
         }
 
+        String batchMessage = "No batching required, all transactions ${performMigration ? 'were' : 'will be'} processed."
+        if (requiresBatching) {
+            batchMessage = "This location has more transactions than specified allowed batch size: ${batchSize}. "
+            batchMessage += performMigration ? "Migration was run only for ${batchSize} most recent transactions." :
+                    "Migration will be run in batches of ${batchSize} most recent transactions."
```
alannadolny: Groovy doesn't allow multiline strings? :thinking:

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
ewaterman: It's a list of [Product, transaction date, transaction type id] (see the new method in InventoryService) but I agree it's not ideal to be indexing like this. Two alternatives:

1) Make it a map instead so that it's more obvious what each field is: `[product: Product, transactionDate: Date, transactionTypeId: String]`. This still isn't great since it has to assume what the keys are but it saves you from having to make a whole new class just to wrap this data.
2) Make an actual class that wraps these fields

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
ewaterman: You can do:
```
String s = "blah" +
    "blah" +
    "blah"
```
I think the "proper" way to do it is via string builders but I'm not really concerned about it in this scenario since it's just a single message

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -411,9 +411,15 @@ class MigrationService {
...
+
+        Integer batchSize = grailsApplication.config.openboxes.transactions.inventoryBaseline.migration.batchSize ?: 5000
+        boolean requiresBatching = Transaction.countByInventoryAndTransactionType(location.inventory, oldProductInventoryType) > batchSize
+
+        log.info("Fetching ${requiresBatching ? batchSize : 'all'} transactions candidates for migration")
         List<Transaction> transactions = Transaction.findAllByInventoryAndTransactionType(
-                location.inventory, oldProductInventoryType
-        )?.sort { [it.transactionDate, it.dateCreated] }?.reverse()
+                location.inventory, oldProductInventoryType,
+                [max: requiresBatching ? batchSize : -1, sort: [transactionDate: "desc", dateCreated: "desc"]]
```
ewaterman, line 420: it's hard for me to see the actual changes here since there's a big diff. So the idea is that it'll only migrate 5000 rows max and if you want to migrate the rest you need to run it again?

I was looking to see if there was a loop somewhere where it'd loop until there was nothing left to migrate but I didn't see one

awalkowiak: >  if you want to migrate the rest you need to run it again?

Yes, otherwise it would hit timeouts too, and it would still be confusing for the user whether it finished or not.

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: For some reason, I thought it would be clearer to separate the ternary operator to that version, but yeah, it is not needed

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: Yeah, I think I had in mind doing something else here, but it ended like that, will replace with ternary

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: Yeah, you are right, it is clearer as a map instead of a list, changed it.

alannadolny (approved):

PR #5463, opened 2025-08-26
OBPIH-7467 Remove facility-specific forecasting configuration from Location

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7467

**Description:**
After the Tech Huddle discussion, we decided to remove the option to add forecasting for locations, so I removed all related places, including the Forecasting tab from Edit Location page and the location configuration.

---


grails-app/domain/org/pih/warehouse/inventory/InventoryLevel.groovy
```diff
@@ -92,7 +92,7 @@ class InventoryLevel {
     static constraints = {
         status(nullable: true)
         inventory(nullable: false)
-        product(nullable: false)
+        product(nullable: true)
```
SebastianLib: We need to set product as **nullable: true** because in the **updateForecastingConfiguration** function, there’s no product during the update, which causes an error. The line 
`inventoryLevelDataService.save(inventoryLevelInstance)` in updateForecastingConfiguration was causing the error. 

Previously, product was **nullable: true**, but it was changed to **false** in this ticket: https://github.com/openboxes/openboxes/commit/4daf3ed7c8057f4e9d36b54d99b5dde95840cbf0. I think changing it back to **nullable: true** should not cause any new issues.

alannadolny: Yeah, it was changed in this ticket, but do you know the exact cause of that change?

ewaterman: @jmiranda commented in that original PR with:

> product and inventory should never be null. Or that's at least true so far, since we haven't encountered a requirement for a default system-side inventory level yet or a default inventory level across organizations.

So I agree with Alan that we should figure out what exactly is going on here. Why does forecasting need a null product? Is it because that's how we're representing a global default across the facility? So was the above assumption just incorrect and actually the column should be nullable or do we need to rethink our approach?

On prod we do have one inventory level where product is null:
```
select * from inventory_level where product_id is null;
+----------------------------------+---------+----------------------------------+--------------+------------+------------------+---------------------+---------------------+-----------+--------------+--------------+-----------+-----------+-------------------------+----------------------+-------------------+---------------------------+---------------------------+----------+----------------------+---------------------------+-------------------------+
| id                               | version | inventory_id                     | min_quantity | product_id | reorder_quantity | date_created        | last_updated        | status    | max_quantity | bin_location | abc_class | preferred | expected_lead_time_days | forecast_period_days | forecast_quantity | preferred_bin_location_id | replenishment_location_id | comments | internal_location_id | replenishment_period_days | demand_time_period_days |
+----------------------------------+---------+----------------------------------+--------------+------------+------------------+---------------------+---------------------+-----------+--------------+--------------+-----------+-----------+-------------------------+----------------------+-------------------+---------------------------+---------------------------+----------+----------------------+---------------------------+-------------------------+
| cab2b10790efdf5801916ad14443474c |       0 | cab2b10790efdf5801916acffb6e472e |         NULL | NULL       |             NULL | 2024-08-19 13:26:25 | 2024-08-19 13:26:25 | SUPPORTED |         NULL | NULL         | NULL      |         0 |                    NULL |                30.00 |              NULL | NULL                      | NULL                      | NULL     | NULL                 |                      NULL |                    NULL |
+----------------------------------+---------+----------------------------------+--------------+------------+------------------+---------------------+---------------------+-----------+--------------+--------------+-----------+-----------+-------------------------+----------------------+-------------------+---------------------------+---------------------------+----------+----------------------+---------------------------+-------------------------+
1 row in set (0.000 sec)
```

SebastianLib: If we change values in the forecasting tab, we don’t pass the product, so I’m not entirely sure, as I wasn’t involved in creating this, but it seems to me that changes in the forecasting tab are global for a given location, which is why we don’t pass any product. That’s how it worked before Justin’s changes in this PR: https://github.com/openboxes/openboxes/pull/5121. I think Justin might have overlooked the case where we can change values in the forecasting tab, which is why he set the product nullable to false. @jmiranda, do you know anything about this?

jmiranda: Can we discuss on the tech huddle tomorrow? I think I had forgotten about the forecasting configuration for location (i.e. OBAM-140 Add forecasting configuration to location edit page #2687) which means it was probably a bad idea to go that route. I think we might need to plan on separating configuration related to location vs location + product vs organization. 

I think the change to make the product nullable again will need to be merged (to avoid a regression) but we need to make sure we understand the side effects and maybe add workarounds for cases where product can be null.

src/js/components/locations-configuration/ZoneAndBinLocations.jsx
```diff
@@ -360,8 +379,8 @@ class ZoneAndBinLocations extends Component {
             <button type="button" onClick={() => this.previousPage()} className="btn btn-outline-primary float-left btn-xs">
               <Translate id="react.default.button.previous.label" defaultMessage="Previous" />
             </button>
-            <button type="button" onClick={() => this.nextPage()} className="btn btn-outline-primary float-left btn-xs">
-              <Translate id="react.default.button.next.label" defaultMessage="Next" />
+            <button type="button" onClick={() => this.finish()} className="btn btn-outline-primary float-right btn-xs">
```
awalkowiak, line 382: @SebastianLib Is there a need to submit anything now on this step? Or everything is already saved, and we don't have anything to submit?

PR #5469, opened 2025-08-29
OBLS-129: Creation of automatic putaway after auto-receiving

druchniewicz:


grails-app/conf/application.yml
```diff
@@ -640,6 +640,9 @@ openboxes:
         sequenceNumber:
             format: '00000'
 
+    inboundSortation:
```
jmiranda: Move this to the "receiving" section above.

grails-app/services/org/pih/warehouse/inboundSortation/InboundSortationService.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+    private PutawayItem createPutawayItem(ReceiptItem receiptItem, Shipment shipment) {
+        new PutawayItem(
+                product: receiptItem.product,
+                inventoryItem: receiptItem.inventoryItem,
+                quantity: receiptItem.quantityReceived,
+                recipient: receiptItem.recipient,
+                currentFacility: shipment.destination,
+                currentLocation: receiptItem.binLocation,
+                putawayLocation: receiptItem.product.getInventoryLevel(shipment.destination.id)?.preferredBinLocation,
```
jmiranda: This is a good start, but this is going to be much more complicated and probably require its own service. 
For example, let's say we received 100 EA into the Inbound Sortation location and we have outbound demand of
* 10 EA bound for service staging
* 50 EA bound for local delivery staging

We want to have a mechanism that takes the received quantity, calculates outbound demand, determines the preferred putaway location and then generates the appropriate putaway orders. 

So using the strategy pattern, we'd build a putaway context with the relevant input (i.e. 100ea, product A, facility B), pass it through the strategy service and it would create the following putaway orders
* 10 ea to service 
* 50 ea to local delivery 
* 40 ea to preferred bin location

jmiranda: I'll start working on that next week.

jmiranda: Note that this line might set the putaway location to NULL if there's no preferred bin. I believe a putaway order can be saved without a putaway destination, so as long as we can set it from the putaway UI, then perhaps it's fine to set the putaway destination to NULL here. If not we should find a random bin location (potentially unused, but not necessary) to assign as the destination OR throw an error if one is not found.

jmiranda (changes requested):
Minor changes requested.

grails-app/services/org/pih/warehouse/inboundSortation/InboundSortationService.groovy
druchniewicz: I've started refactoring code to more generic, with SlottingService, PutawayContext, PutawayStrategy etc.

I will push first draft version today. Then we can extend it next steps.

Regarding putawawy location == null - yes, putaway order can be saved with empty putaway destination with no problems.

grails-app/services/org/pih/warehouse/inboundSortation/SlottingService.groovy
```diff
@@ -0,0 +1,27 @@
...
+            def tasks = strategy.execute(context, locations, quantityRemaining)
+
+            if (tasks) {
+                putawayTasks.addAll(tasks)
+                quantityRemaining -= tasks*.quantity.sum(0) as int
+            }
+        }
+
+        if (quantityRemaining > 0) {
+            throw RuntimeException("Quantity remaining should not be 0 at this point. Something went wrong.")
```
jmiranda: This error message contradicts the logic.

grails-app/services/org/pih/warehouse/inboundSortation/PutawayTask.groovy
```diff
@@ -0,0 +1,14 @@
+package org.pih.warehouse.inboundSortation
+
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.inventory.InventoryItem
+import org.pih.warehouse.product.Product
+
+class PutawayTask {
```
jmiranda: Let's call this PutawayResult for now so we don't conflict with the PutawayTask.

grails-app/services/org/pih/warehouse/inboundSortation/SlottingService.groovy
jmiranda: Could you also add the FIXME/TODO comments from the design document so I have something to reference if I decide to make any changes.

druchniewicz: right with this error message, nice catch, thanks

I added only one FIXME/TODO that was in the [confluence page](https://openboxes.atlassian.net/wiki/spaces/OL/pages/2303655937/Sortation+Service)

btw. confluence docs also says:
```
    // Get a list of all possible putaway locations
    // For now this could just be facility.locations or [] since the
    // initial strategy will be to putaway to the preferred bin for 
    // the product. But if the preferred bin is not specified then 
    // we'll need the list of bins to pick the next best one (i.e.
    // random)
    List<Location> locations = 
      locationQueryService.getLocationsByXyz(ctx.facility.id, [..])
```

but is it really necessary to create a new service and method for searching locations? I used def `locations = context.facility.internalLocations` so far - should not that be sufficient? you have all internal locations of given destination, and you can filter however you want in specific strategy if needed?

jmiranda: For now, getting all of the internal locations is fine. But eventually we may have criteria that we'll want to use to narrow down and sort the available locations. A lot of that should be taken care of within each strategy, but we might want to narrow the scope of locations that we consider to get rid of as many as we can before passing them off to the strategies.

Search by 
* availability (free / busy)
* capability (i.e. supports putaway)
* capacity (volume)
* equipment required for putaway 
* compatibility (i.e. temperature control)
* * sort by distance, ABC class, etc

jmiranda (approved):
This looks great. Does the cross-docking strategy work? If so, let's merge this as-is and consider merging the putaway task API ticket so we can start testing.

grails-app/conf/spring/resources.groovy
```diff
@@ -1,5 +1,9 @@
 package spring
 
+import org.pih.warehouse.inboundSortation.CrossDockingStrategy
```
jmiranda: Let's make this `inbound` for now. I'd like to come up with a better package structure, but I think inbound is good enough.

grails-app/domain/org/pih/warehouse/requisition/OutboundDemand.groovy
```diff
@@ -0,0 +1,9 @@
+package org.pih.warehouse.requisition
+
+import org.pih.warehouse.product.Product
+
+class OutboundDemand {
```
jmiranda: Move this into src/groovy since it's not a domain.

grails-app/services/org/pih/warehouse/inboundSortation/CrossDockingStrategy.groovy
```diff
@@ -0,0 +1,39 @@
...
+
+        List<PutawayResult> putawayTasks = []
+        for (entry in outboundDemandsMap.entrySet()) {
+            DemandTypeCode demandTypeCode = entry.key
+            List<OutboundDemand> demandsForType = entry.value
+
+            boolean isProductDemanded = demandsForType.any { it.product == context.product }
+
+            if (isProductDemanded) {
+                Location putawayLocation = locations.find { it.locationType?.name?.equalsIgnoreCase(demandTypeCode.code) }
```
jmiranda: So this will return the first one that matches the name, but we should probably be looking for all locations with the prefix i.e. WC-00, WC-01, WC-02, sorting them (alphabetically for now) and then returning the first one.

grails-app/services/org/pih/warehouse/inboundSortation/PutawayResult.groovy
```diff
@@ -0,0 +1,14 @@
...
+import org.pih.warehouse.core.Location
+import org.pih.warehouse.inventory.InventoryItem
+import org.pih.warehouse.product.Product
+
+class PutawayResult {
+    Location facility
+    Product product
+    InventoryItem inventoryItem
+    Location currentBinLocation
+    Location putawayLocation
```
jmiranda: I just noticed that the previous two fields use the old. Let's match the name used in the PutawayTask to make the mapping more straightforward. I might eventually rename these fields a bit. We need container, even though we don't use it now, because eventually we'll try to guess which container should be used.
```
Location location
Location container
Location destination 
```

jmiranda (approved):

PR #5470, opened 2025-09-01
OBPIH-7478 Validation issue on record stock when item with zone in stock 

alannadolny:


awalkowiak (approved):

grails-app/controllers/org/pih/warehouse/inventory/InventoryItemController.groovy
```diff
@@ -546,7 +546,21 @@ class InventoryItemController {
 
         // We need if/else statement to avoid duplication data (OBPIH-7438)
         if (flash.recordInventoryRows) {
-            commandInstance.recordInventoryRows = flash.recordInventoryRows
+            commandInstance.recordInventoryRows = flash.recordInventoryRows.collect { RecordInventoryRowCommand it ->
+                Location binLocation = it?.binLocation ? Location.get(it.binLocation.id) : null
```
kchelstowski, line 550: I don't feel 100% comfortable with approving this, as we should find the underlying issue of having to fetch the location explicitly, but given "the age" of the record stock code and the timeline, I'm approving it.

kchelstowski (approved):

PR #5471, opened 2025-09-01
OBPIH-7468 Display wards and depots on PO shipment (fix)

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7468

**Description:**

---

### DIFF
--- a/src/js/utils/option-utils.jsx
+++ b/src/js/utils/option-utils.jsx
@@ -75,7 +75,7 @@ export const debounceLocationsFetch = (
           direction: fetchAll ? undefined : (direction || queryParams?.direction),
           isReturnOrder: isReturnOrder || undefined,
           activityCodes,
-          withOrganization,
+          withOrganization: withOrganization || undefined,
         },
       }).then((result) => callback(_.map(result.data.data, (obj) => {
         const locationTypeData = withTypeDescription ? ` [${obj.locationType.description}]` : '';


grails-app/services/org/pih/warehouse/core/LocationService.groovy
```diff
@@ -134,6 +134,10 @@ class LocationService {
             locations = locations.findAll { location -> user.hasPrimaryRole(location) }
         }
 
+        if (params.withOrganization) {
```
SebastianLib, line 137: In create purchase order view we also filter by organization, so we need to do it here as well

src/js/components/stock-movement-wizard/combined-shipments/SendMovementPage.jsx
```diff
@@ -268,6 +269,13 @@ class SendMovementPage extends Component {
     this.debouncedLocationsFetch = debounceLocationsFetch(
       this.props.debounceTime,
       this.props.minSearchLength,
+      [activityCode.RECEIVE_STOCK], // activityCodes
+      true, // fetchAll
+      true, // withOrgCode
+      false, // withTypeDescription
+      false, // isReturnOrder
+      null, // direction
+      true, // withOrganization
```
SebastianLib, line 278: this looks ugly so I'm wondering if it would be better to make this an object, but then we would have to change it in many places

awalkowiak: @SebastianLib You can create a ticket to change this in the near future

PR #5472, opened 2025-09-02
OBPIH-7477 Fix performance issues of inventory transactions report wh…

kchelstowski:
…en reaching big amount of transactions in the system

**Description:**
Having reached a particular number of transactions in the system, the MariaDB query optimizer starts using different indexes and refs than it should be supposed to be using, which can cause in performance issues and scanning more rows than it should. 
When I compared the `EXPLAIN` locally and on obstg, I realized that obstg started using different transaction refs in the query than previously, and suddenly the performance dropped from ~15s on Rwinkwavu to ~4 minutes....

I decided not to fight with the automatic query optimizer anymore and force some of the indexes, and take a step back which results in:

- getting rid of the `inventory-transactions-summary` view that made the pre-filter for `facility` and `transaction_date` impossible
- instead of querying the view, we just query the same SQL statement, but with pre-filters of `facility` and `transaction_date` that cuts the rows significantly that the performance should be around ~100ms now (locally):
```sql
152 rows in set (0.077 sec)
```
vs
```sql
152 rows in set (2 min 3.716 sec)
```

If we ever need to query this logic for the whole system, we might think about bringing back the view (having assumed that we refactored the transactions not to have to have the complexed logic to handle all the edge cases), but for now, the view unfortunately was an overkill.

---


grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -14,23 +18,188 @@ import java.sql.Timestamp
...
-                "in"("product", command.products)
-            }
-            order("dateRecorded", "desc")
+    NativeQuery buildInventoryTransactionsQuery(String queryString, CycleCountReportCommand command, boolean paginate) {
+        NativeQuery query = sessionFactory
+            .currentSession
+            .createNativeQuery(queryString)
+            .setParameter("facility", command.facility.id)
+            .setParameter("startDate", command.startDate)
+            .setParameter("endDate", command.endDate)
```
kchelstowski, line 29: had to go the native query route and adding parameters like that, due to the problem with injecting the list of product ids other than injecting it directly via string (sql injection) that we wanted to avoid

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -14,23 +18,186 @@ import java.sql.Timestamp
...
+            JOIN location facility ON facility.inventory_id = transaction.inventory_id
+            LEFT JOIN product_inventory_summary pis
+               ON pis.product_id = inventory_item.product_id
+               AND pis.facility_id = facility.id
+               -- An adjustment is treated as associated with the baseline if the time diff between them is 1 second (baseline is created 1 second before the adjustment)
+               AND TIMESTAMPDIFF(SECOND, pis.baseline_transaction_date, transaction.transaction_date) = 1
+            WHERE transaction.transaction_type_id = '3' -- Adjustments
+            AND facility.id = :facility 
+            AND transaction.transaction_date BETWEEN :startDate AND :endDate 
+            ${command.products ? 'AND product.id IN (:products)' : ''}
```
ewaterman, line 192: injection?

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -14,23 +18,186 @@ import java.sql.Timestamp
...
+            JOIN product ON inventory_item.product_id = product.id
+            JOIN location facility ON facility.inventory_id = transaction.inventory_id
+            LEFT JOIN product_inventory_summary pis
+               ON pis.product_id = inventory_item.product_id
+               AND pis.facility_id = facility.id
+               -- An adjustment is treated as associated with the baseline if the time diff between them is 1 second (baseline is created 1 second before the adjustment)
+               AND TIMESTAMPDIFF(SECOND, pis.baseline_transaction_date, transaction.transaction_date) = 1
+            WHERE transaction.transaction_type_id = '3' -- Adjustments
+            AND facility.id = :facility 
+            AND transaction.transaction_date BETWEEN :startDate AND :endDate
```
ewaterman, line 155: so this is the only "new" line? Everything else is copied from the inventory-transactions-summary view?

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
ewaterman: Same question for the above query

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -14,23 +18,186 @@ import java.sql.Timestamp
...
+            WHERE transaction.transaction_type_id = '3' -- Adjustments
+            AND facility.id = :facility 
+            AND transaction.transaction_date BETWEEN :startDate AND :endDate 
+            ${command.products ? 'AND product.id IN (:products)' : ''}
+            GROUP BY transaction.id, inventory_item.product_id
+        """
+
+        int totalCount = buildInventoryTransactionsQuery(totalCountQuery, command, false)
+                .list()
+                .size()
```
ewaterman, line 198: So all we care about is the total number of product + transaction groups?

Would a count distinct be better?
```
SELECT COUNT(DISTINCT transaction.id, inventory_item.product_id)
FROM xyz;
```

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
```diff
@@ -14,23 +18,186 @@ import java.sql.Timestamp
...
+                        transaction: Transaction.read(it[0]),
+                        product: Product.read(it[1]),
+                        facility: Location.read(it[2]),
+                        quantityBefore: it[3],
+                        quantityAfter: it[4],
+                        quantityDifference: it[5],
+                        dateRecorded: it[6],
+                        recordedBy: User.read(it[7]),
+                        baselineTransaction: Transaction.read(it[8])
+                )}
```
ewaterman, line 174: is it more performant to bind the entities like this afterwards instead of in the query itself?

You could probably make this faster by doing it in batches:
```
List<String> transactionIds = []
List<String> productIds = []
...
List queryList = buildInventoryTransactionsQuery().list()
queryList.each {
    transactionIds.add(it[0])
    productIds.add(it[1])
    ...
}

Map transactionsMap = Transaction.getAll(transactionIds).collectEntries { entry -> return [entry.id, entry] }
Map productsMap = Product.getAll(productIds).collectEntries { entry -> return [entry.id, entry] }
...

List<InventoryTransactionsSummary> inventoryTransactions = queryList.collect {
    new InventoryTransactionsSummary(
        transaction: transactionsMap.get(it[0]),
        product: productsMap.get[1]),
        ...
}
```
Up to you though

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: it’s provided via param, why do you think it’s an injection?

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: yeah

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: will give it a shot, I didn’t know it was possible to group while counting in SQL 🤔

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
kchelstowski: Since this is a pure SQL, it doesn’t bind the domains, as we return plain values. 
I didn’t want to overcomplicate this since this is paginated, and very quick now (around 100ms), so we wouldn’t gain much from reducing n+1 just here. 
Anyway your suggestion is worth noting if we ever experience performance issues here (e.g. we won’t use pagination here in the future)

grails-app/services/org/pih/warehouse/inventory/InventoryTransactionSummaryService.groovy
ewaterman: you're right... I must have misread it

jmiranda: I don't think it's SQL injection since we're not using string interpolation to include the product IDs in the query string. The parameters are still bound at the end. 

This is what it would look like if we were doing SQL injection here
```
${command.products ? 'AND product.id IN (${command.products.join(",")})' : ''}
```

PR #5473, opened 2025-09-02
OBPIH-7465 Record stock not creating the correct adjustment when backdated

alannadolny:


grails-app/services/org/pih/warehouse/inventory/RecordStockProductInventoryTransactionService.groovy
```diff
@@ -28,22 +28,18 @@ class RecordStockProductInventoryTransactionService extends ProductInventoryTran
...
-            ? recordInventoryRowCommand.newQuantity
-            : recordInventoryRowCommand.newQuantity - quantityOnHand
+        Integer quantityOnHand = availableItems.get(key)?.quantityOnHand
+        // When the row doesn't already have an existing quantity on hand on the date of creation, it means that it is a new
+        // row. In that case, we should add the whole value from that row to the already existing corresponding line. If
+        //The quantity in the already existing row is 5, and the new row has a quantity of 2. We should get one line with qty 7,
+        // and the transaction entry should contain quantity 2. In case we are editing an already existing line (old quantity
+        // greater than 0), we should set the new value for that line. The adjustment entry should be the difference between
+        // the new value and the original value. When we had quantity 5, and then we changed it to 7, the row should have
+        // quantity 7, but the adjustment should have quantity 2.
```
ewaterman, line 38: I'm confused by this comment. It sounds like it's talking about the scenario where we do a record stock with two rows for the same product + bin + lot. Something like:

Image

So for example, given the following record stock:
```
p1, bin1, lot1, quantity=5
p1, bin1, lot1, quantity=10
```
If QoH at the time was 6, then we should end up with a single transaction entry row with quantity +9 for a total of 15
If QoH at the time was 0, then we should end up with a single transaction entry row with quantity +15 for a total of 15

But is that what's actually happening? In inventory import we merge the rows together if they have the same product + bin + lot (https://github.com/openboxes/openboxes/blob/release/0.9.5/grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy#L253-L276) but I don't see that being done for record stock.

All I see here is that we're looping the RecordInventoryRowCommand rows https://github.com/openboxes/openboxes/blob/release/0.9.5/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy#L1437. So if we have the above scenario then we'll end up with:
If QoH at the time was 6, then we get two rows, one with quantity -1 and one +4 for a total of 9 (wrong)
If QoH at the time was 0, then we get two rows, one with quantity +5 and one with +10 for a total of 15 (this scenario works fine)

alannadolny: This comment is not new; I added it when I was fixing the issue with merging two lines. Btw - it works as expected :smile:

awalkowiak: @alannadolny aaaand it turned out that @ewaterman was right here :cry:

alannadolny: But the issue is caused by the previous pr, not this one! :sadsmile:

jmiranda: Good catch, @ewaterman. Has this always been the case or this new with the refactoring? 

Edit: Nevermind, this must be new since we didn't create adjustments before this release. I guess I was mostly wondering how Record Stock worked when we had two rows with the same inventory item and whether our test cases cover a lot of these edge cases already. I assume so since we found this bug in the first place.

@alannadolny I wanted to mention that I see you re: sadsmile. Thanks for trying to bring that to a wider audience. We can try to lobby github and the rest of the world to adopt Kacper's brilliant emoji.

PR #5474, opened 2025-09-02
OBPIH-7475 date formatters for GSPs and data exporters

ewaterman:

- https://pihemr.atlassian.net/browse/OBPIH-7474
- https://pihemr.atlassian.net/browse/OBPIH-7475

**Description:** I decided to do both OBPIH-7474 and OBPIH-7475 together since there ended up being significant overlap.

This change adds the ability for us to format dates for display on GSPs and CSV exporters.

To format a date for GSPs:
```
<span class="something">
    ${g.formatDate(date: productInstance?.dateCreated)}
</span>
```

To format a date for exporters:
```
def csvRow = [
    ...
    Created:  dateFormatter.formatForExport(product.dateCreated),
    Updated: dateFormatter.formatForExport(product.lastUpdated),
]
```

---

Date in GSP in UTC+6 timezone:
image

Same date in UTC-7 timezone:
image

Same date in Spanish locale:
image

Same date in Chinese locale (it "works" but it's not grammatically correct for Chinese):
image

Date in CSV export in ISO format (my server is PST so daylight savings time weirdness is happening):
image

### DIFF
--- a/grails-app/domain/org/pih/warehouse/product/Product.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/Product.groovy
@@ -11,6 +11,7 @@ package org.pih.warehouse.product
 
 import grails.databinding.BindUsing
 import grails.util.Holders
+import java.time.Instant
 import org.apache.commons.collections.FactoryUtils
 import org.apache.commons.collections.list.LazyList
 import org.apache.commons.lang.NotImplementedException
@@ -229,8 +230,8 @@ class Product implements Comparable, Serializable {
     GlAccount glAccount
 
     // Auditing
-    Date dateCreated
-    Date lastUpdated
+    Instant dateCreated
+    Instant lastUpdated
     User createdBy
     User updatedBy
 
--- a/grails-app/services/org/pih/warehouse/product/ProductService.groovy
+++ b/grails-app/services/org/pih/warehouse/product/ProductService.groovy
@@ -17,6 +17,8 @@ import org.apache.commons.lang.StringUtils
 import org.hibernate.criterion.CriteriaSpecification
 import org.hibernate.criterion.Restrictions
 import org.hibernate.sql.JoinType
+import org.springframework.beans.factory.annotation.Autowired
+
 import org.pih.warehouse.core.ApiException
 import org.pih.warehouse.core.Constants
 import org.pih.warehouse.core.GlAccount
@@ -27,7 +29,9 @@ import org.pih.warehouse.core.Tag
 import org.pih.warehouse.core.UnitOfMeasure
 import org.pih.warehouse.LocalizationUtil
 import util.ReportUtil
-import java.text.SimpleDateFormat
+
+import org.pih.warehouse.core.date.DateFormatterManager
+
 /**
  * @author jmiranda*
  */
@@ -41,6 +45,7 @@ class ProductService {
     def userService
     def dataService
     ProductGroupService productGroupService
+    @Autowired DateFormatterManager dateFormatter
 
     def getNdcResults(operation, q) {
         def hipaaspaceApiKey = grailsApplication.config.hipaaspace.api.key
@@ -853,7 +858,6 @@ class ProductService {
      */
     String exportProducts(List<Product> products, boolean includeAttributes) {
 
-        def formatDate = new SimpleDateFormat("dd/MMM/yyyy hh:mm:ss")
         def attributes = Attribute.findAllByExportableAndActive(true, true)
         def formatTagLib = grailsApplication.mainContext.getBean('org.pih.warehouse.FormatTagLib')
         boolean hasRoleFinance = userService.hasRoleFinance()
@@ -887,8 +891,8 @@ class ProductService {
                 VendorName          : product.vendorName ?: '',
                 UPC                 : product.upc ?: '',
                 NDC                 : product.ndc ?: '',
-                Created             : product.dateCreated ? "${formatDate.format(product.dateCreated)}" : "",
-                Updated             : product.lastUpdated ? "${formatDate.format(product.lastUpdated)}" : "",
+                Created             : dateFormatter.formatForExport(product.dateCreated),
+                Updated             : dateFormatter.formatForExport(product.lastUpdated),
             ]
 
             if (includeAttributes) {
--- a/grails-app/taglib/org/pih/warehouse/DateTagLib.groovy
+++ b/grails-app/taglib/org/pih/warehouse/DateTagLib.groovy
@@ -9,9 +9,12 @@
  **/
 package org.pih.warehouse
 
+import grails.util.Holders
+import java.time.temporal.TemporalAccessor
 import org.ocpsoft.prettytime.PrettyTime
 import groovy.time.TimeDuration
 import org.pih.warehouse.core.Constants
+import org.pih.warehouse.core.date.DateFormatterManager
 
 class DateTagLib {
 
@@ -22,7 +25,33 @@ class DateTagLib {
         out << (new Date().format(org.pih.warehouse.core.Constants.DEFAULT_YEAR_FORMAT))
     }
 
+    /**
+     * Formats a date for display and localizes it to a given locale and timezone.
+     */
     Closure formatDate = { attrs, body ->
+        if (!attrs.containsKey('date')) {
+            throwTagError("Attribute [date] is required.")
+        }
+
+        switch (attrs.date) {
+            // All dates going forward should be java.time classes
+            case TemporalAccessor:
+                out << formatJavaTimeDate(attrs)
+                return
+            // Old dates are java.util.Date instances so we keep this for backwards compatability.
+            case Date:
+                out << formatJavaUtilDate(attrs)
+                return
+            default:
+                throwTagError("Attribute [date] does not support type [${attrs.date.class}].")
+        }
+    }
+
+    /**
+     * Overrides Grails' built-in FormatTagLib, wrapping it with custom functionality.
+     * See https://gsp.grails.org/latest/ref/Tags/formatDate.html for more information.
+     */
+    private String formatJavaUtilDate(Object attrs) {
         def formatTagLib = grailsApplication.mainContext.getBean('org.grails.plugins.web.taglib.FormatTagLib')
 
         if (!attrs.format) {
@@ -31,7 +60,14 @@ class DateTagLib {
         if (!attrs.timeZone && session.timezone) {
             attrs.timeZone = session.timezone
         }
-        out << formatTagLib.formatDate.call(attrs)
+        return formatTagLib.formatDate.call(attrs)
+    }
+
+    private String formatJavaTimeDate(Object attrs) {
+        DateFormatterManager dateFormatterManager =
+                Holders.grailsApplication.mainContext.getBean("dateFormatterManager") as DateFormatterManager
+
+        return dateFormatterManager.format(attrs.date)
     }
 
     def expirationDate = { attrs, body ->
--- a/grails-app/views/inventoryItem/_productDetails.gsp
+++ b/grails-app/views/inventoryItem/_productDetails.gsp
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.inventory.InventoryStatus" %>
+<%@ page import="org.pih.warehouse.DateUtil; org.pih.warehouse.inventory.InventoryStatus" %>
 <style>
 .nailthumb-container {
     width: 100%;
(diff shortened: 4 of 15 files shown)


grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -229,8 +230,8 @@ class Product implements Comparable, Serializable {
     GlAccount glAccount
 
     // Auditing
-    Date dateCreated
-    Date lastUpdated
+    Instant dateCreated
+    Instant lastUpdated
```
ewaterman, line 234: Our APIs can already handle the java.time classes like Instant. They'll automatically get stringified to the ISO format when returning them in JSON

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -887,8 +891,8 @@ class ProductService {
                 VendorName          : product.vendorName ?: '',
                 UPC                 : product.upc ?: '',
                 NDC                 : product.ndc ?: '',
-                Created             : product.dateCreated ? "${formatDate.format(product.dateCreated)}" : "",
-                Updated             : product.lastUpdated ? "${formatDate.format(product.lastUpdated)}" : "",
+                Created             : dateFormatter.formatForExport(product.dateCreated),
+                Updated             : dateFormatter.formatForExport(product.lastUpdated),
```
ewaterman, line 897: We'll eventually need to change all our CSV exporters to do this for all of our date fields

grails-app/taglib/org/pih/warehouse/DateTagLib.groovy
```diff
@@ -22,7 +25,33 @@ class DateTagLib {
         out << (new Date().format(org.pih.warehouse.core.Constants.DEFAULT_YEAR_FORMAT))
     }
 
+    /**
+     * Formats a date for display and localizes it to a given locale and timezone.
+     */
     Closure formatDate = { attrs, body ->
```
ewaterman, line 31: Making this change in the formatDate taglib allows us to keep using `g.formatDate` as we were before, except now it accepts java.time classes as well. Date classes will behave as they were before so we can migrate entities at our own pace and our GSPs will automatically be switched over.

grails-app/views/inventoryItem/_productDetails.gsp
```diff
@@ -350,8 +350,7 @@
                 </td>
                 <td class="value">
                     <span class="fade">
-                        ${g.formatDate(date: productInstance?.dateCreated, format: 'd-MMM-yyyy')}
-                        ${g.formatDate(date: productInstance?.dateCreated, format: 'hh:mma')}
+                        ${g.formatDate(date: productInstance?.dateCreated)}
```
ewaterman, line 354: in this case I needed to modify the GSP because it was using a custom format, which we're trying not to do anymore. We can add support for the "format" keyword back in if we want to, but ideally we use it in rare cases where we absolutely need to change the format for some reason.

src/main/groovy/org/pih/warehouse/core/session/SessionManager.groovy
```diff
@@ -0,0 +1,44 @@
...
+import org.pih.warehouse.DateUtil
+
+/**
+ * Handles operations on a user's HTTP session.
+ *
+ * Grails exposes the "session" field to controllers and tag libs, so those classes can safely use that field.
+ * This class is for accessing the session from anywhere else in the application.
+ */
+@Component
+class SessionManager {
```
ewaterman, line 18: I made this because I needed a way to get session attributes outside of a controller and I figured it'd be good to unify the behaviour under a class.

src/main/groovy/org/pih/warehouse/core/date/DateFormatterManager.groovy
```diff
@@ -0,0 +1,116 @@
...
+import org.springframework.stereotype.Component
+
+import org.pih.warehouse.core.session.SessionManager
+
+/**
+ * Formats date objects for display. For use only by data exporters and GSPs. APIs should always return dates in
+ * ISO format using UTC.
+ */
+@Component
+class DateFormatterManager {
```
ewaterman, line 18: This is the important class that glues it all together. The format(date, context) method will stringify any date, applying the locale and timezone of the user.

src/main/groovy/org/pih/warehouse/core/date/DateFormatterManager.groovy
```diff
@@ -0,0 +1,116 @@
...
+            case ZonedDateTime:
+                formatter = new TemporalAccessorDateTimeFormatter<ZonedDateTime>(
+                        locale, context?.patternOverride, displayFormat, timezone)
+                break
+            case LocalDate:
+                formatter = new TemporalAccessorDateFormatter<LocalDate>(
+                        locale, context?.patternOverride, displayFormat)
+                break
+            default:
+                throw new UnsupportedOperationException("Cannot format date of type [${date.class}]")
```
ewaterman, line 56: I intentionally didn't include Date as an option because if we're refactoring something to use this manager, we're also refactoring the domain to not use Date anymore.

If for whatever reason we need to continue using Date for something, we should be able to create a JavaUtilDateFormatter class relatively easily and just add a case for it here.

src/main/groovy/org/pih/warehouse/core/date/TemporalAccessorFormatter.groovy
```diff
@@ -0,0 +1,90 @@
...
+     * Returns the DateTimeFormatter that will be used to format the date to a String.
+     */
+    DateTimeFormatter getFormatter() {
+        // If we're overriding the pattern, construct a new formatter from that pattern and return it.
+        if (StringUtils.isNotBlank(pattern)) {
+            return addContextToFormatter(DateTimeFormatter.ofPattern(pattern))
+        }
+
+        DateTimeFormatter formatter
+        switch (displayFormat) {
```
ewaterman, line 60: We want different date patterns depending on what display format we're formatting to. For example, JSON should always be ISO strings, and GSPs should be a localized "dd/MMM/yyyy..." string.

And of course that can still be overwritten if the displayFormat param is set.

src/main/groovy/org/pih/warehouse/core/date/TemporalAccessorDateFormatter.groovy
```diff
@@ -0,0 +1,38 @@
...
+        //       constant value (though we can still use the constant as a fallback if the message isn't defined).
+        return DateUtil.DEFAULT_DISPLAY_DATE_FORMATTER
+    }
+
+    DateTimeFormatter getCsvFormatter() {
+        // We format our CSV exports using ISO pattern (regardless of locale) because we want to ensure that if
+        // the same CSV is later imported, it won't fail on date format errors. We could export to a locale-specific
+        // pattern but we'd need to verify that each of those patterns can be handled by our data binders. For now,
+        // we stick with ISO pattern because it's the only universal date format and is guaranteed to work.
+        return DateTimeFormatter.ISO_DATE
```
ewaterman, line 36: lets discuss this. This causes LocalDates in CSVs to always be formatted like yyyy-MM-dd and the equivalent line in `TemporalAccessorDateTimeFormatter` causes Instants/ZonedDateTimes to be formatted like yyyy-MM-ddThh:mm:ssXXX, regardless of locale.

So for example: 2021-02-16T12:11:09-03:00

The reason I chose to do this was to guarantee that if we then import that CSV file, the dates would ALWAYS be marshalled sucessfully, regardless of locale, because they're always in a valid format.

If we instead had dates like 15/Jan/2015 then 1) I don't know if we want to assume that english users will need to know they need to type Jan, Feb, Mar... when adding new dates and 2) I don't think it'd be able to marshall it properly and so the API call would fail. We eventually want the "display" format of dates in the UI to be configurable per locale (ex: chinese dates look like YYYY 年 MM 月 DD 日). We'll achieve this by using something like the `default.date.format` translation in message.properties. As such, we need to do one of two things:

1. For CSVs, always stick to a known accepted format (in this case I chose ISO). This simplifies the logic but might be slightly confusing to users if they don't see that format anywhere else in the UI.
2. Allow CSVs to export in the same locale-specific "display" format that we use in the UI and then loosen the requirements on our databinders to support first checking if the date string being input is formatted like the "display" format of the locale of the request.

For example:
If you're in english, it'll check if the date is formatted like 15/Jan/2015 (or whatever format we choose).
If you're in chinese, it'll check if the date is formatted like 2015 年 01 月 15 日.

And if it's not formatted like that, then continue checking our other supported formats (which are all just variations of the ISO format). This is probably more intuitive for users since we can choose a format that makes sense to their locale but it means different formats become valid/invalid depending on your locale, so if you're in english and your dates look like 15/Jan/2015, the dates will import successfully. But if you send a csv to someone who is in french (or you just change your locale), then the csv suddenly becomes invalid and wont import. That IMO is quite bad and was the big reason why I chose 1 for now.

I'm open to other ideas though!

grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -41,6 +45,7 @@ class ProductService {
     def userService
     def dataService
     ProductGroupService productGroupService
+    @Autowired DateFormatterManager dateFormatter
```
alannadolny: What's the formatting that we should use in this case? I mean:
```
@Autowired DateFormatterManager dateFormatter
```
vs
```
@Autowired
DateFormatterManager dateFormatter
```

grails-app/views/inventoryItem/_productDetails.gsp
```diff
@@ -1,4 +1,4 @@
-<%@ page import="org.pih.warehouse.inventory.InventoryStatus" %>
+<%@ page import="org.pih.warehouse.DateUtil; org.pih.warehouse.inventory.InventoryStatus" %>
```
alannadolny: We are adding the imports, line by line
```suggestion
<%@ page import="org.pih.warehouse.DateUtil" %>
<%@ page import="org.pih.warehouse.inventory.InventoryStatus" %>
```

grails-app/services/org/pih/warehouse/product/ProductService.groovy
ewaterman: I agree that the second one is generally clearer but if we have a lot of classes that we're autowiring then they can take up a lot of room at the top of the file.

```
@Autowired X x
@Autowired Y y
@Autowired Z z
```
vs
```
@Autowired
X x

@Autowired
Y y

@Autowired
Z z
```

You can look at ApiSpec.groovy for an example of what I mean. They really start to get annoying with how much space they take up.

But I suppose if we're importing a ton of components its an indicator that our class is doing too much work and needs to be broken down, so I don't mind changing it to be the "proper" format

grails-app/domain/org/pih/warehouse/product/Product.groovy
jmiranda: :+1:, but do me a favor and execute the following command when you get a chance.
```
grails dbm-generate-changelog changelog.xml
```
See what data type is being assigned for the product table's date_created and last_updated fields. I want to make sure that the data type that Grails/Hibernate expects is the same as what we're using. 

To be clear, I'm expecting the mapping to use TIMESTAMP instead of DATETIME and if that's the case, how does TIMESTAMP differ from DATETIME and what are some of the implications of leaving the column as a DATETIME when the property is defined as an Instant.

ewaterman: Looks like it is still seeing it as a datetime:
```
<changeSet author="ewaterman (generated)" id="1757091419092-74">
    <createTable tableName="product">
        <column name="date_created" type="datetime">
            <constraints nullable="false"/>
        </column>
        <column name="last_updated" type="datetime">
            <constraints nullable="false"/>
        </column>
        ...
    </createTable>
</changeSet>
```

ewaterman: as to the differences between TIMESTAMP and DATETIME, expand this section of my old design: https://pihemr.atlassian.net/wiki/spaces/OB/pages/3382542337/OBPIH-6916+Standardizing+backend+datetime+parsing#Mapping-java.time.*-classes-to-Domain-models-and-the-DATETIME-database-type

TLDR: We should probably be using TIMESTAMP WITH ZONE for all our date fields (MySQL 8 fixes the [year 2038 ](https://en.wikipedia.org/wiki/Year_2038_problem)problem for 64 bit machines) but since our servers are always in UTC, it doesn't really matter much for our use case.

PR #5476, opened 2025-09-04
OBPIH-7390  Add tests for utils and hooks in cycle count

SebastianLib:

**Description:**

---

### DIFF
--- a/src/js/tests/hooks/useAssignCycleCountModal.test.jsx
+++ b/src/js/tests/hooks/useAssignCycleCountModal.test.jsx
@@ -1,5 +1,6 @@
 import { act, renderHook } from '@testing-library/react-hooks';
 import moment from 'moment/moment';
+import { baseUser } from 'tests/__mocks__/users';
 
 import cycleCountApi from 'api/services/CycleCountApi';
 import notification from 'components/Layout/notifications/notification';
@@ -7,8 +8,6 @@ import countIndex from 'consts/countIndex';
 import cycleCountColumn from 'consts/cycleCountColumn';
 import NotificationType from 'consts/notificationTypes';
 import useAssignCycleCountModal from 'hooks/cycleCount/useAssignCycleCountModal';
-
-import { baseUser } from '../__mocks__/users';
 
 jest.mock('react-redux', () => ({
   useSelector: jest.fn().mockImplementation(() => ({
--- a/src/js/tests/hooks/useResolveStepHeader.test.jsx
+++ b/src/js/tests/hooks/useResolveStepHeader.test.jsx
@@ -1,9 +1,8 @@
 import { act, renderHook } from '@testing-library/react-hooks';
 import moment from 'moment';
+import { baseUser, newUser } from 'tests/__mocks__/users';
 
 import useResolveStepHeader from 'hooks/cycleCount/useResolveStepHeader';
-
-import { baseUser, newUser } from '../__mocks__/users';
 
 describe('useResolveStepHeader', () => {
   const props = {
--- a/src/js/tests/hooks/useCountStepHeader.test.jsx
+++ b/src/js/tests/hooks/useCountStepHeader.test.jsx
@@ -1,9 +1,8 @@
 import { act, renderHook } from '@testing-library/react-hooks';
 import moment from 'moment';
+import { baseUser, newUser } from 'tests/__mocks__/users';
 
 import useCountStepHeader from 'hooks/cycleCount/useCountStepHeader';
-
-import { baseUser, newUser } from '../__mocks__/users';
 
 describe('useCountStepHeader', () => {
   const props = {


src/js/tests/hooks/useCycleCountProductAvailablity.test.js
```diff
@@ -0,0 +1,139 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { ALL_PRODUCTS_TAB, TO_RESOLVE_TAB } from 'consts/cycleCount';
+import cycleCountCandidateStatus from 'consts/cycleCountCandidateStatus';
+import useCycleCountProductAvailability from 'hooks/cycleCount/useCycleCountProductAvailability';
+import useQueryParams from 'hooks/useQueryParams';
+import * as checkStatusUtils from 'utils/checkCycleCountStatus';
```
alannadolny, line 7: is it necessary to import everything from that util?

src/js/tests/hooks/useResolveStepHeader.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  const newUser = {
+    id: 'user-2',
+    firstName: 'Alice',
+    lastName: 'Wonder',
+    name: 'Alice Wonder',
+    label: 'Alice Wonder',
+    value: 'user-2',
+  };
```
alannadolny: What do you think about moving this to a test util file? It looks like it can be shared between some test files

src/js/tests/utils/cycleCountUtils.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+    },
+  };
+
+  it('trims leading and trailing spaces from lotNumber', () => {
+    const props = {
+      ...mockCycleCountItem,
+      inventoryItem: {
+        ...mockCycleCountItem.inventoryItem,
+        lotNumber: '   ABC123   ',
+      },
```
alannadolny, line 26: You can also add a scenario when the lot number is: '   A  B  C 1 2    3 ', because it shouldn't touch the spaces between letters + the case when the lot is 'ABC123''

src/js/tests/utils/groupBinLocationsByZone.test.jsx
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('handles empty binLocations array safely', () => {
+    const result = groupBinLocationsByZone([], mockTranslate);
+    expect(result).toEqual([]);
+  });
+
+  it('sorts zones alphabetically with No Zone at the start', () => {
+    const result = groupBinLocationsByZone(mockBinLocations, mockTranslate);
+    expect(result.map((zone) => zone.name)).toEqual(['No Zone', 'ZoneA', 'ZoneB']);
```
alannadolny, line 126: Are you sure that `toEqual` works properly on arrays?

src/js/tests/hooks/useCycleCountProductAvailablity.test.js
SebastianLib: yes, everything is needed to run the tests

src/js/tests/utils/groupBinLocationsByZone.test.jsx
SebastianLib: Yes, that works fine. It's supposed to check if the alphabetical order is correct. If I change the order in toEqual, then the test fails

src/js/utils/Badge.jsx
```diff
@@ -6,14 +6,19 @@ import CustomTooltip from 'wrappers/CustomTooltip';
...
+const Badge = ({ label, variant, tooltip }) => {
+  if (!label) {
+    return null;
+  }
+  return (
+    <CustomTooltip
+      content={label}
+      show={tooltip}
+    >
+      <span className={`badge ${variant}`} data-testid="badge">{label}</span>
```
ewaterman, line 18: what is `data-testid` used for? I'm not a fan of needing to put test fields in our real components but maybe I'm not understanding the need here

ewaterman:
nice job writing tests! :muscle:

src/js/utils/Badge.jsx
SebastianLib: We use data-testid to detect component elements, for example Badge.

example:
```
  it('applies the correct variant class', () => {
    render(<Badge label="Primary Badge" variant="primary" />);
    const badgeElement = screen.getByTestId('badge');
    expect(badgeElement).toHaveClass('badge', 'primary');
  });
```
We could avoid that and detect the component in other ways (for example by text), but I thought this approach is nice and consistent, since we already use data-testid in many places.

image

alannadolny: data-testid is also used as a selector for e2e tests

alannadolny (approved):

src/js/tests/hooks/useAssignCycleCountModal.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+import moment from 'moment/moment';
+
+import cycleCountApi from 'api/services/CycleCountApi';
+import notification from 'components/Layout/notifications/notification';
+import countIndex from 'consts/countIndex';
+import cycleCountColumn from 'consts/cycleCountColumn';
+import NotificationType from 'consts/notificationTypes';
+import useAssignCycleCountModal from 'hooks/cycleCount/useAssignCycleCountModal';
+
+import { baseUser } from '../__mocks__/users';
```
alannadolny: The path should be absolute, not relative

src/js/utils/Badge.jsx
ewaterman: Thanks for the explanation. Ultimately I'm okay with it. I did some reading about it and it seems like some people like it and other people hate it so 🤷.

I'm still not a fan of needing to add fields to our real components (even if they ultimately won't be rendered to the user) but I get the convenience aspect. These are unit tests, so I don't mind selecting by text or label or whatever since we can pretty much guarantee that it will be unique as we're controlling what gets rendered. But I get that's more complex to do for e2e tests where that text might not be unique. However I don't really see how our use of data-testid helps that scenario. IMO in that scenario we'd probably benefit from either finding a way to generate unique ids for components in a way that makes them queryable by those ids, but maybe that's more trouble than is worth.

Again, I'm fine with this but it doesn't feel like a perfect solution

alannadolny (approved):

PR #5477, opened 2025-09-04
OBPIH-7487 Fix wrong qty before in the inventory transactions report …

kchelstowski:
…after using adjust stock

**Description:**

---

### DIFF
--- a/grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
+++ b/grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
@@ -48,6 +48,9 @@ class Transaction implements Comparable, Serializable {
 
     def publishSaveEvent() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshProductAvailabilityEvent(this, forceRefresh))
+    }
+
+    void publishReportsEvents() {
         Holders.grailsApplication.mainContext.publishEvent(new RefreshInventoryCountEvent(this))
         Holders.grailsApplication.mainContext.publishEvent(new RefreshInventoryTransactionsSummaryEvent(this))
     }
@@ -60,6 +63,7 @@ class Transaction implements Comparable, Serializable {
     // ID won't be available until after the record is inserted
     def afterInsert() {
         publishSaveEvent()
+        publishReportsEvents()
     }
 
     def afterUpdate() {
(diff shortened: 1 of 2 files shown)


grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
```diff
@@ -60,6 +63,7 @@ class Transaction implements Comparable, Serializable {
     // ID won't be available until after the record is inserted
     def afterInsert() {
         publishSaveEvent()
+        publishReportsEvents()
```
kchelstowski, line 66: since the refresh of helper views was triggered on each save of transaction, so insert and update, and since we can update transactions (e.g. identifier number job, that updates the transaction), we could have created duplicate rows in helper views with each update of a transaction) - see the 14:36:00 transaction for example
Screenshot from 2025-09-03 16-46-27

First row shows only one record, the second row shows two rows (as the identifier job updated the transaction and triggered the event again), then I updated a transaction by changing entries, and again the event was triggered (3 rows visible).

Summary: for those helper views, we only want to add transactions once to the view

grails-app/migrations/views/inventory-transaction-summary-helper-views.sql
```diff
@@ -6,14 +6,16 @@ SELECT
...
+        CASE
+            WHEN t.transaction_type_id IN ('2', '4', '5', '9') THEN -te.quantity
+            ELSE te.quantity END
     ) AS quantity_sum
 FROM (
          SELECT id, transaction_date, inventory_id, transaction_type_id
          FROM transaction
-         WHERE transaction_type_id IN ('3','8','9') -- adjustments, transfer ins, transfer outs
-           AND transaction.order_id IS NULL
+         WHERE transaction_type_id IN ('2', '3', '4', '5', '8','9') -- adjustments, transfer ins, transfer outs, expired, damaged, consumption
```
kchelstowski, line 17: it'a an additional fix not related to this ticket - except the transfer in/outs, we would also like to count Expired, Damaged, Consumption transactions for the quantity before calculation. 
This would be a rare case to include them in the calculation, because they would usually exist with a baseline, but if we have e.g. 30 in stock, and we mark 30 as Expired, and then record stock, we would expect quantity before as 0, not 30 that would be visible if we didn't count Expired here.

kchelstowski:
Note: remove the `inventory_movement_summary` and `product_inventory_summary` tables from dbstg before the PR is merged, so that the fix could be applied for the existing data.

PR #5480, opened 2025-09-05
OBPIH-6969 Improve approach to product creation and stock data (endpoints)

alannadolny:


grails-app/services/org/pih/warehouse/product/ProductService.groovy
```diff
@@ -824,10 +825,14 @@ class ProductService {
                 }
             }
 
-            if (!product.save(flush: true)) {
+            Product savedProduct = product.save(flush: true)
```
ewaterman: nitpick: I think it'd still work the way it originally was and so this assignment to savedProduct might not be needed. You can still call `importedProducts.add(product)` later and I believe it'd contain the saved product. I'm not 100% sure though

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -796,6 +801,15 @@ class UrlMappings {
             action = [GET: "getItemsWithBackdatedShipments"]
         }
 
+        /**
+         * Inventory API endpoints
+         */
+
+        "/api/locations/$id/inventories/import"(parseRequest: true) {
```
ewaterman: why not `/api/facilities/$facilityId/...` (instead of /locations) like we've been doing for other APIs?

Also is the parseRequest required?

grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
```diff
@@ -179,7 +179,10 @@ class Transaction implements Comparable, Serializable {
         comment(nullable: true)
         // transaction date cannot be in the future
         transactionDate(nullable: false,
-                validator: { value -> value <= new Date() })
+            validator: {
+                value -> value <= new Date(System.currentTimeMillis() + 1000)
```
ewaterman: why is this needed? With this change you can now make transactions 1 second in the future. I don't think we want that

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -339,23 +339,26 @@ class InventoryImportDataService implements ImportDataService {
...
-            expirationDate = expirationDateRaw.toDate()
+            return inventoryService.findAndUpdateOrCreateInventoryItem(product, lotNumber, calendar.getTime())
+        }
+
+        if (expirationDateRaw instanceof LocalDate) {
+            return inventoryService.findAndUpdateOrCreateInventoryItem(product, lotNumber, expirationDateRaw.toDate())
         }
 
-        return inventoryService.findAndUpdateOrCreateInventoryItem(product, lotNumber, expirationDate)
+        return inventoryService.findAndUpdateOrCreateInventoryItem(product, lotNumber, expirationDateRaw)
```
ewaterman: suggestion: I personally don't see this refactoring as cleaner since it now needs to call inventoryService.findAndUpdateOrCreateInventoryItem all over the place. If you're really committed to not using else if conditions then I suggest doing this instead:

```
Date expirationDate = parseExpirationDate(expirationDateRaw)
return inventoryService.findAndUpdateOrCreateInventoryItem(product, lotNumber, expirationDate)


Date parseExpirationDate(Object expirationDateRaw) {
    if (expirationDateRaw instanceof String) {
        Date date = EXPIRATION_DATE_FORMAT.parse(expirationDateRaw)
        Calendar calendar = Calendar.getInstance()
        calendar.setTime(expirationDate)
        return calendar.getTime()
    }
    if (expirationDateRaw instanceof Date) {
        return expirationDateRaw
    }
    if (expirationDateRaw instanceof LocalDate) {
        return expirationDateRaw.toDate()
    }
    return null
}
```

src/main/groovy/org/pih/warehouse/importer/CSVUtils.groovy
```diff
@@ -200,4 +201,30 @@ class CSVUtils {
...
+
+    static List<Map<String, String>> csvToObjects(String csvString) {
+        CSVParser parser = CSVParser.parse(
+                csvString,
+                CSVFormat.DEFAULT.withFirstRecordAsHeader().withTrim()
+        )
+
+        return parser.collect { record ->
+            record.toMap().collectEntries { k, v ->
+                [(toCamelCase(k)): v]
```
ewaterman, line 213: Why do we want to modify the input data? What does it give us?

grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -339,23 +339,26 @@ class InventoryImportDataService implements ImportDataService {
...
     }
 
     private Location parseBinLocation(String binLocationName, Location parentLocation) {
-        if (binLocationName == null) {
+        if (!binLocationName) {
```
ewaterman, line 365: ~~What about if we have a bin location with a "" name? I know we've had issues with blanks vs nulls before so I'd be weary of treating them both as equivalent to null.~~

Nvm I checked and locations cannot be blank so this is fine.

grails-app/services/org/pih/warehouse/product/ProductService.groovy
alannadolny: You're right, started thinking in JavaScript way on the backend :smile:

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
alannadolny: I can see that we mixed /facilities vs /locations (16 usages vs 14 usages), but I am going to follow the facility pattern

grails-app/domain/org/pih/warehouse/inventory/Transaction.groovy
alannadolny: In the `importData` method, the baseline date is the current date, and the adjustment date is the baseline date + 1. When the import happens quickly, the adjustment date is in the future, so the validation fails. The logic in `importData` doesn't seem right to me in that context. I am going to subtract one second while creating `ImportDataCommand`, unless you have nothing against creating a baseline transaction with the date -1s, and the adjustment just on the transaction date.

src/main/groovy/org/pih/warehouse/importer/CSVUtils.groovy
alannadolny: The headers of CSV that we are using are like "Product Code", etc., so this is parsing it to "productCode" to work with the import method

awalkowiak: What if someone mistypes P**or**duct Code? Will it try to map to p**or**ductCode, which would later result in some potential issues? Or there will be an exception meaningful to the user, or will he be stuck trying to figure out why his file is not being imported?

awalkowiak (commented):
Overall lgtm, but I have a question in one of the threads

src/main/groovy/org/pih/warehouse/importer/CSVUtils.groovy
alannadolny: The `importData` method will throw an error in the same way as it is done on the import on Grails pages

awalkowiak: ok

PR #5482, opened 2025-09-08
OBPIH-6969 Fix parsing expiration date

alannadolny:


grails-app/services/org/pih/warehouse/importer/InventoryImportDataService.groovy
```diff
@@ -337,7 +337,7 @@ class InventoryImportDataService implements ImportDataService {
     }
 
     private InventoryItem parseInventoryItem(Product product, def lotNumberRaw, def expirationDateRaw) {
-        Date expirationDate = parseExpirationDate(expirationDateRaw)
+        Date expirationDate = expirationDateRaw ? parseExpirationDate(expirationDateRaw) : null
```
ewaterman: nitpick: I'm going to approve, but why not put this check inside of parseExpirationDate like how parseBinLocation has?

ewaterman (approved):

PR #5485, opened 2025-09-08
OBPIH-7488 support localized date patterns

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7488

**Description:** As a part of the timezone fix, we're trying to narrow down to only a specific set of patterns that we use for displaying dates. In addition, we want each of these date formats to be localizable since different languages might display dates differently.

For example, in English we may want to display LocalDate fields like dd/MMM/yyyy but in Chinese we may want to display them like yyyy 年 MM 月 dd 日.

So this ticket does the following:
- adds the new set of localizable date formats to message.properties
- fixes a malformatted spanish locale date
- modifies the existing g:formatDate taglib to support localizing dates
- modifies the DateFormatterManager.format method to support localizing dates

Ultimately, to localize our fields we do:
- Instant: DateFormatterManager.format(Instant) which calls TemporalAccessorDateTimeFormatter.format(Instant)
- LocalDate: DateFormatterManager.format(LocalDate) which calls TemporalAccessorDateTimeFormatter.format(LocalDate)
- Messages: MessageLocalizer.localize('x.y.z')

---

Here's a diagram of all the important classes involved:
image


grails-app/i18n/messages_es.properties
```diff
@@ -352,7 +352,7 @@ default.credits.label=Créditos
 default.balance.label=Saldo
 default.dashboard.label=Tablero
 default.data.label=datos
-default.date.format=dd / MMM / aaaa hh\: mm\: ss
+default.date.format=dd/MMM/yyyy hh:mm:ss a z
```
ewaterman, line 355: while I was testing I noticed that the spanish message for the default date was looking weird. It actually wouldn't parse properly for me. "a" is "AM"/"PM" so I think this format was just wrong...

grails-app/i18n/messages.properties
```diff
@@ -232,6 +232,12 @@ dashboard.binLocationSummary.label=Bin Location Summary <small>(grouped by Inven
 dashboard.fastMovers.label=Fast Movers <small>(Last 30 Days)</small>
 datepicker.hint.label=Use CTRL + arrow keys to navigate the calendar
 
+# Date formats that support localization.
+date.format.date.label=dd/MMM/yyyy
+date.format.dateTime.label=dd/MMM/yyyy HH:mm:ss
+date.format.dateTimeZone.label=dd/MMM/yyyy HH:mm:ss XXX
+date.format.time.label=MM/yyyy
```
ewaterman: These are the localizations for the new date formats. We should only need to override them for locales that want a different format.

But lets discuss what patterns we want to go with. These will display like:
```
01/Jan/2025
01/Jan/2025 00:00:00
01/Jan/2025 00:00:00 +01:00
00:00:00
```
Which I picked since that's what we already use throughout most of the app.

grails-app/taglib/org/pih/warehouse/DateTagLib.groovy
```diff
@@ -48,26 +61,44 @@ class DateTagLib {
...
+                .withPatternOverride(attrs.format as String)
+                .withTimezoneOverride(attrs.timeZone as ZoneId)
+                // Note "displayStyle" differs from "style" from Grails' formatDate, though they function similarly.
+                .withDisplayStyleOverride(attrs.displayStyle as DateDisplayStyle)
+                // If either displayStyle or format are set, don't also set the display format or else it would
+                // trigger errors in the DateFormatterContext.
+                .withDisplayFormat(attrs.displayStyle || attrs.format ? null : DateDisplayFormat.GSP)
+                .build()
+
+        return dateFormatterManager.format(attrs.date, context)
```
ewaterman, line 101: This allows us to specify additional attrs when using `<g:formatDate>`. We shouldn't need to use these extra attrs very often but it gives us the flexibility to override behaviours as needed for any edge cases that we have

grails-app/taglib/org/pih/warehouse/LocalizationTagLib.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+import org.grails.plugins.web.taglib.ValidationTagLib
+
+import org.pih.warehouse.core.localization.MessageLocalizer
+
+class LocalizationTagLib {
+
+    // Use Grails' namespace because we're overriding the default behaviour and so that the tag lib
+    // is accessible automatically throughout the app.
+    static namespace = 'g'
```
ewaterman: I might revert this whole class. I wanted to standardize how we localize messages in general but it wasn't working with special characters while I was testing locally. TBD after testing my branch on a deployed environment.

src/main/groovy/org/pih/warehouse/core/date/DateDisplayStyle.groovy
```diff
@@ -0,0 +1,39 @@
...
+enum DateDisplayStyle {
+
+    /*
+     * We should strive to only ever use a single format per date type. For example, we should only have one format
+     * for displaying day + month + year. This allows us to be consistent in how we display date fields in the app.
+     */
+    DATE('date.format.date.label', DateTimeFormatter.ofPattern('dd/MMM/yyyy')),
+    DATE_TIME('date.format.dateTime.label', DateTimeFormatter.ofPattern('dd/MMM/yyyy HH:mm:ss')),
+    DATE_TIME_ZONE('date.format.dateTimeZone.label', DateTimeFormatter.ofPattern('dd/MMM/yyyy HH:mm:ss XXX')),
+    TIME('date.format.time.label', DateTimeFormatter.ofPattern('HH:mm:ss'))
```
ewaterman, line 23: The enum version of the dates in message.properties. We don't strictly need this class but I find it much nicer to be able to work with enums instead of hard coded strings. That way if we need to change the underlying labels we can do so without refactoring all the usages

src/main/groovy/org/pih/warehouse/core/date/DateFormatterContext.groovy
```diff
@@ -0,0 +1,84 @@
+package org.pih.warehouse.core.date
+
+import java.time.ZoneId
+
+/**
+ * Context object containing the configuration fields for formatting dates.
+ * For a majority of cases the default settings can be used and so this context object will not be required.
+ */
+class DateFormatterContext {
```
ewaterman, line 9: I just moved this class to its own file. The only addition is displayStyleOverride

src/main/groovy/org/pih/warehouse/core/localization/LocalizableMessage.groovy
```diff
@@ -0,0 +1,32 @@
+package org.pih.warehouse.core.localization
+
+/**
+ * A simple wrapper class on a translatable message/label.
+ * Used to indicate that translations can be applied to the message.
+ */
+class LocalizableMessage {
```
ewaterman, line 7: I didn't end up using this class but I would like to keep it in since I eventually want to do some work towards standardizing our error messages and this would play a part in that. I'm imagining we have either a custom error class or a custom error message class that can hold a localized message and any other data we need.

src/main/groovy/org/pih/warehouse/core/localization/MessageLocalizer.groovy
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.stereotype.Component
+
+import org.pih.warehouse.core.Localization
+import org.pih.warehouse.core.session.SessionManager
+
+/**
+ * Translates messages into a specific locale.
+ */
+@Component
+class MessageLocalizer {
```
ewaterman, line 16: This is meant to standardize how we localize our messages in the app. We have a ton of different approaches and I didn't know which to choose when localizing dates for this PR so got fed up and decided to try to unify them all under one component. Eventually I'd like to unify our errors and GSPs to all localize via the same method so that we can remove the other approaches. That's a refactoring I'll save for a future date but this at least gives us a starting point

src/main/groovy/org/pih/warehouse/core/localization/LocaleDeterminer.groovy
```diff
@@ -0,0 +1,50 @@
...
+import org.pih.warehouse.LocalizationUtil
+import org.pih.warehouse.core.session.SessionManager
+
+/**
+ * Handles fetching locales.
+ *
+ * If you're trying to translate a message, use {@link org.pih.warehouse.core.localization.MessageLocalizer}.
+ */
+@Component
+class LocaleDeterminer {  // Note I wanted to call this LocaleResolver but that's already a Spring component
```
ewaterman, line 16: I noticed we had a ton of different ways that we resolved locales so I decided that I was going to bite the bullet and try to unify our behaviour.

I'll eventually add a flow diagram to this PR to demonstrate all the components that go into this change since I feel like it's hard to visualize them all. I figured a bunch of single purpose classes gives us better flexibility and testability than a single, massive class that does locale, localization, and messaging all in one place.

Eventually I want to refactor LocalizationUtil and all the other l10n related services as well but that'd be too much for one PR

grails-app/taglib/org/pih/warehouse/LocalizationTagLib.groovy
```diff
@@ -0,0 +1,43 @@
+package org.pih.warehouse
+
+import org.grails.plugins.web.taglib.ValidationTagLib
+
+import org.pih.warehouse.core.localization.MessageLocalizer
+
+class LocalizationTagLib {
```
ewaterman, line 7: This isn't currently used anywhere. I figured we'd want to experiment with the new messageLocalizer before applying it everywhere. Once we're comfortable that this flow works, we can switch the namespace of this class to 'g' and then everywhere that we have <g:message> will go through this taglib.

alannadolny (approved):

grails-app/i18n/messages.properties
```diff
@@ -232,6 +232,12 @@ dashboard.binLocationSummary.label=Bin Location Summary <small>(grouped by Inven
 dashboard.fastMovers.label=Fast Movers <small>(Last 30 Days)</small>
 datepicker.hint.label=Use CTRL + arrow keys to navigate the calendar
 
+# Date formats that support localization.
+date.format.date.label=dd/MMM/yyyy
+date.format.dateTime.label=dd/MMM/yyyy HH:mm:ss
+date.format.dateTimeZone.label=dd/MMM/yyyy HH:mm:ss XXX
+date.format.time.label=HH:mm:ss
```
jmiranda, line 239: As i mentioned on the call last week (?), i would really like to keep these all under the default namespace in messages.properties.
```
default.date.format=dd/MMM/yyyy
default.time.format=HH:mm:ss
default.date.time.format=dd/MMM/yyyy HH:mm:ss
default.date.time.timezone.format=dd/MMM/yyyy HH:mm:ss XXX
```
If that's going to cause too big of a change (since default.date.format has always been used to render the full datetime format) then we could preserve that by leaving it alone 
```
default.date.format=dd/MMM/yyyy hh:mm:ss a z
```
and adding more context-specific formats 
```
default.date.only.format=dd/MMM/yyy
default.datetime.format=dd/MMM/yyyy HH:mm:ss
default.datetime.timezone.format=dd/MMM/yyyy hh:mm:ss a z
default.time.format=HH:mm:ss
```
and then adding more date formats in the future, as needed if codifying these became necessary.
```
default.date.expiry.format=MMM yyyy
default.date.pretty.format=EEE, MMM dd, yyyy
default.date.weekday.format=EEE
default.date.mysql.format=yyyy-MM-dd
default.date.filename.format.=yyyyMMdd
default.datetime.filename.format=yyyyMMddHHmmss
```

PR #5491, opened 2025-09-10
OBPIH-7493 Last counted date incorrect when migration created baselin…

kchelstowski:
…e at the date of migration

**Description:**

---


grails-app/domain/org/pih/warehouse/product/Product.groovy
```diff
@@ -570,7 +570,7 @@ class Product implements Comparable, Serializable {
      * @param locationId
      * @return
      */
-    Date latestTransactionDate(String locationId, List<String> transactionTypeIds) {
+    Date latestTransactionDate(String locationId, List<String> transactionTypeIds, String commentToFilter) {
```
ewaterman: nitpick: seems like you could just use the constant in this method instead of making it a param since I don't see why we'd need to custom filter by comment for any other reason.

PR #5497, opened 2025-09-15
OBLS-178: Complete order only when all order items are completed (WIP)

druchniewicz:


src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
```diff
@@ -34,25 +35,14 @@ class PutawayTaskAdapter {
...
      */
     static OrderItem toOrderItem(PutawayTask task, OrderItem orderItem) {
-
-        Order order = orderItem.order
-
         // In most cases, we just want to make sure we're setting the most recent status
-        order.status = toOrderStatus(task.status)
-
-        // start - putaway task has been assigned and has been started
-        order.approvedBy = task.assignee
```
druchniewicz, line 44: I removed operations where me modify **order** and leave only these ones where we modify **orderItem** object. I think it suits more if this method is OrderItem toOrderItem(..)
Operations with order I extracted one level higher

PR #5498, opened 2025-09-15
OBPIH-7342 support java.time types in datepicker taglibs

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7342

**Description:** As a part of switching away from java.util.Date and towards java.time.LocalDate and java.time.Instant, we needed to modify the datepicker taglibs that we use on our GSPs to be able to accomodate the new java.time classes.

We have two date pickers in the old GSPs:

`jqueryDatePicker` which uses a date selector widget similar to what we do on the React side:

image

and `datePicker` which is just some simple dropdown selectors:

image

With this change, we can now call both of them without error:
- `<g:jqueryDatePicker value="${someLocalDateField}" />`
- `<g:datePicker value="${someJavaTimeField}" />`

### DIFF
--- a/grails-app/domain/org/pih/warehouse/product/ProductAssociation.groovy
+++ b/grails-app/domain/org/pih/warehouse/product/ProductAssociation.groovy
@@ -9,6 +9,8 @@
  **/
 package org.pih.warehouse.product
 
+import java.time.Instant
+
 /**
  * Represents an association between two products.
  */
@@ -31,8 +33,8 @@ class ProductAssociation {
 
     ProductAssociation mutualAssociation
 
-    Date dateCreated
-    Date lastUpdated
+    Instant dateCreated
+    Instant lastUpdated
 
     static belongsTo = [product: Product]
 
--- a/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
+++ b/grails-app/services/org/pih/warehouse/core/DashboardService.groovy
@@ -12,6 +12,8 @@ package org.pih.warehouse.core
 import grails.gorm.transactions.Transactional
 import grails.plugins.csv.CSVWriter
 import org.hibernate.sql.JoinType
+
+import org.pih.warehouse.DateUtil
 import org.pih.warehouse.inventory.InventoryItem
 import org.pih.warehouse.inventory.InventoryLevel
 import org.pih.warehouse.inventory.InventoryStatus
@@ -163,11 +165,11 @@ class DashboardService {
             }
 
             if (command.startDate) {
-                ge("expirationDate", command.startDate + 1)
+                ge("expirationDate", DateUtil.asDate(command.startDate.plusDays(1)))
             }
 
             if (command.endDate) {
-                le("expirationDate", command.endDate)
+                le("expirationDate", DateUtil.asDate(command.endDate))
             }
 
             order("expirationDate", "desc")
@@ -206,11 +208,11 @@ class DashboardService {
             }
 
             if (command.startDate) {
-                ge("expirationDate", command.startDate + 1)
+                ge("expirationDate", DateUtil.asDate(command.startDate.plusDays(1)))
             }
 
             if (command.endDate) {
-                le("expirationDate", command.endDate)
+                le("expirationDate", DateUtil.asDate(command.endDate))
             }
 
             order("expirationDate", "asc")
--- a/grails-app/taglib/org/pih/warehouse/DatePickerTagLib.groovy
+++ b/grails-app/taglib/org/pih/warehouse/DatePickerTagLib.groovy
@@ -9,16 +9,67 @@
  **/
 package org.pih.warehouse
 
-import java.text.SimpleDateFormat
+import java.time.Instant
+import java.time.LocalDate
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import org.apache.commons.lang.StringUtils
+import org.grails.plugins.web.taglib.FormTagLib
+
+import org.pih.warehouse.core.Constants
 
 class DatePickerTagLib {
 
+    /**
+     * Simple date + time form field selectors for dates.
+     * More info: https://docs.grails.org/latest/ref/Tags%20-%20GSP/datePicker.html
+     */
     Closure datePicker = { attrs, body ->
-        def datePickerTagLib = grailsApplication.mainContext.getBean('org.grails.plugins.web.taglib.FormTagLib')
+        FormTagLib datePickerTagLib = grailsApplication.mainContext.getBean(
+                'org.grails.plugins.web.taglib.FormTagLib') as FormTagLib
+
+        // Grails' date picker taglib doesn't support java.time classes and so the simplest thing we can
+        // do is to convert those types to Date before passing them to the taglib.
+        switch (attrs.value) {
+            case (Instant):
+                attrs.value = DateUtil.asDate(attrs.value as Instant)
+                attrs.precision = 'minute'
+                break
+            case (ZonedDateTime):
+                attrs.value = DateUtil.asDate(attrs.value as ZonedDateTime)
+                attrs.precision = 'minute'
+                break
+            case (LocalDate):
+                attrs.value = DateUtil.asDate(attrs.value as LocalDate)
+                attrs.precision = 'day'
+                break
+            default:
+                break
+        }
+
         out << datePickerTagLib.datePicker.call(attrs)
     }
 
-    def jqueryDatePicker = { attrs, body ->
+    /**
+     * A date picker widget for date-only dates.
+     * More info: https://jqueryui.com/datepicker/
+     *
+     * @attr id The id of the datepicker element. If not set, will use the value of name.
+     * @attr name Required. The name of the field to bind the date to.
+     * @attr value A Date or LocalDate instance to populate the field with initially.
+     * @attr autoSize If true, automatically resizes the input field to fit the date string.
+     *                Defaults to true if size is null, otherwise defaults to false.
+     * @attr size The number of characters wide the input field should be. Defaults to 10.
+     * @attr cssClass An optional CSS class to associate with the date picker element.
+     * @attr changeMonthAndYear If true, shows month and year as dropdown selectors. Defaults to true.
+     * @attr numberOfMonths The number of months to display at one time in the picker. Defaults to 1.
+     * @attr readOnly If true, the picker will be disabled. Defaults to false.
+     * @attr minDate Date string representing the earliest date that can be selected. Defaults to no limit.
+     * @attr maxDate Date string representing the latest date that can be selected. Defaults to no limit.
+     * @attr placeholder The text to display in the input field before any date has been selected. Defaults to empty.
+     * @attr autocomplete If true, enables autocomplete for the input field. Defaults to false.
+     */
+    Closure jqueryDatePicker = { attrs, body ->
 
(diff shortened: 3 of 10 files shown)


grails-app/services/org/pih/warehouse/core/DashboardService.groovy
```diff
@@ -163,11 +165,11 @@ class DashboardService {
             }
 
             if (command.startDate) {
-                ge("expirationDate", command.startDate + 1)
+                ge("expirationDate", DateUtil.asDate(command.startDate.plusDays(1)))
```
ewaterman, line 168: We need to convert back to java.util.Date because we haven't yet migrated InventoryItem.expirationDate to be a LocalDate field. Once we do that, we can remove these calls to asDate

grails-app/views/report/showTransactionReport.gsp
```diff
@@ -51,12 +51,10 @@
...
 
-								<g:jqueryDatePicker id="startDate"
-													name="startDate"
+								<g:jqueryDatePicker name="startDate"
 													cssClass="filter"
-													value="${command?.startDate }"
-													format="MM/dd/yyyy"
-													autocomplete="off"/>
+													value="${command?.startDate}"
+                                />
```
ewaterman, line 57: the changes to `g:jqueryDatePicker` that I made are backwards compatable and so the changes to these GSPs that use that taglib are actually not required. However I did it anyways to demonstrate the minimal set of fields that we'd need to provide. Everything else has defaults for when the field isn't provided.

Note that the format attr doesn't do anything anymore (in fact it never really did anything significant which is why I removed it).

grails-app/views/productAssociation/list.gsp
```diff
@@ -119,7 +119,7 @@
 
                                         <td>${fieldValue(bean: productAssociationInstance, field: "comments")}</td>
 
-                                        <td><format:date obj="${productAssociationInstance.dateCreated}" /></td>
+                                        <td><g:formatDate date="${productAssociationInstance.dateCreated}" /></td>
```
ewaterman, line 122: Instead of refactoring format:date to account for the new java.time classes I opted to simply replace it with calls to g:formatDate. format:date should be deprecated anyways IMO since it doesn't have all the features that g:formatDate has and we should be trying to unify our approach to l10n.

grails-app/views/productAssociation/create.gsp
```diff
@@ -95,7 +95,7 @@
 									<label for="dateCreated"><warehouse:message code="productAssociation.dateCreated.label" default="Date Created" /></label>
 								</td>
 								<td valign="top" class="value ${hasErrors(bean: productAssociationInstance, field: 'dateCreated', 'errors')}">
-									<g:datePicker name="dateCreated" precision="minute" value="${productAssociationInstance?.dateCreated}"  />
+									<g:datePicker name="dateCreated" value="${productAssociationInstance?.dateCreated}"  />
```
ewaterman: This change wasn't required, but now that we're using java.time classes it can figure out the precision on its own based on the type (LocalDate is day precision, everything else is minute) and so this field no longer needs to be provided.

grails-app/domain/org/pih/warehouse/product/ProductAssociation.groovy
```diff
@@ -31,8 +33,8 @@ class ProductAssociation {
 
     ProductAssociation mutualAssociation
 
-    Date dateCreated
-    Date lastUpdated
+    Instant dateCreated
+    Instant lastUpdated
```
ewaterman, line 37: This is for testing the g:datePicker

src/main/groovy/org/pih/warehouse/report/InventoryReportCommand.groovy
```diff
@@ -18,8 +20,8 @@ class InventoryReportCommand implements Validateable {
 
     Product product
     Location location
-    Date startDate
-    Date endDate
+    LocalDate startDate
+    LocalDate endDate
```
ewaterman, line 24: This is for testing the g:jqueryDatePicker

grails-app/views/productAssociation/list.gsp
ewaterman: update: I think I figured out how to refactor format:date to be smarter, but I'm going to do that change in a different PR

alannadolny (approved):

grails-app/views/productAssociation/create.gsp
jmiranda: I understand the thought here (sort of a "hey look, it works"), but precision is an implementation detail that is decided by the developer of the page/component. Having it be explicit feels like a good thing for various reasons so I would hate to move in the other direction.

* self-documenting
* unambiguous
* prevents errors if the underlying code changes
* no magic

Convention: "Explicit is better than implicit" 

Reference: https://peps.python.org/pep-0020/

I know we're not using Python, but I like most of the guiding principles from PEP20

grails-app/views/productAssociation/list.gsp
jmiranda: Yes, I would love to replace format:date with the default mechanism! I think a migration path like you're proposing (fix format:date soonish, replace format:date later) sounds great.

grails-app/views/report/showTransactionReport.gsp
jmiranda: Careful getting rid of the "id" attribute. I remember that being there for a reason. I don't understand the move ment of the value line. Was that just so the value is the last attribute specified for readability, everything else is configuration? If so, I think I like that (bookend with the two most important attributes like the author list of a peer-reviewed academic paper).

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
jmiranda: Yeah, I'm glad you brought this up. I know we need it here, but I really don't want to use DateUtil more than is required. 

Quick question ... what happens when you leave it as this?
```
ge("expirationDate", command.startDate.plusDays(1))
```

grails-app/views/report/showTransactionReport.gsp
jmiranda: Here's the line that I was thinking about. 
```
jQuery('#${id}-datepicker')
```
Some pages might have an input element (i.e. datepicker) with the same "name" attribute. The only way for jquery to distinguish between the two is if the "id" is unique. A good example might be a page that let's you create a record dynamically (adding a new row to Record Stock).

grails-app/taglib/org/pih/warehouse/DatePickerTagLib.groovy
```diff
@@ -38,10 +111,28 @@ class DatePickerTagLib {
...
+        // Otherwise we get weird behaviour (the selected dates shift by one day every time the page refreshes).
+        String value
+        switch (attrs.value) {
+            case Date:
+                value = Constants.MONTH_DAY_YEAR_DATE_FORMATTER.format(attrs.value)
+                break
+            case LocalDate:
+                value = DateTimeFormatter.ofPattern('MM/dd/yyyy').format(attrs.value as LocalDate)
+                break
+            case String:
```
jmiranda, line 155: The String case feels fine, but does the value need to be normalized somehow if there is a value provided? Can you make sure to add some test cases for various string values?

grails-app/views/report/showTransactionReport.gsp
ewaterman: > I don't understand the move ment of the value line.

I'm not sure I know what you mean. I just removed all other attrs so value happened to be at the end.  As an aside, if I was doing this from scratch I'd typically put the "important" attrs (in this case value) first and all the optional config attrs after. That way it matches what groovy does for method param optionals. But I wanted to refactor as little as possible here so I left it as is.

> Careful getting rid of the "id" attribute.

We still support providing an id, it just defaults to use the same value as the "name" attribute if id isn't explicitly provided (this was the existing behaviour, I didn't change it). In this case because name and id have the same value, it was redundant to define them both

grails-app/views/productAssociation/create.gsp
ewaterman: The reason I removed it is because we're switching to a model where the field type dictates everything. With java.util.Date it makes sense to need to be provided the precision because it's ambiguous, but with the java.time types, it's not only not ambiguous anymore but also typically incorrect to try to use something else.

LocalDate does not support a time or zone component, so setting a precision value of minute doesn't make sense. It would do nothing at best and break at worst.

Similarly, Instant/ZonedDateTime represent a full date + time, and so we should have date pickers on those types also be full date + time pickers. We can display an Instant however we want (maybe the time component isn't relevant in some screen) but if we find ourselves in a situation where we want the date picker to only be able to specify date and not time, then we should be asking ourselves if that field should be a LocalDate instead. If for some reason we decide that we want it to stay an Instant but only want to pick the date and not the time, then we can specify the precision explicitly, but that should be a rare exception IMO.

grails-app/taglib/org/pih/warehouse/DatePickerTagLib.groovy
ewaterman: I assume any normalization is handled in the jquery date picker itself. I only added support for Strings it because we have a small handful of cases where we pass a string directly, typically an empty one (`listTransactions.gsp` is an example where we pass a string). I don't actually remember why I added support for these extra null scenarios in. Maybe it's simplest to just let it do what it was doing before and pass the string through without any modifications at all

grails-app/services/org/pih/warehouse/core/DashboardService.groovy
ewaterman: Agreed. I'm hoping that once we get through this refactor we can pull out almost all usages of the date util.

Unfortunately it can't handle the conversion gracefull on its own 

Image

grails-app/views/inventory/listExpiredStock.gsp
```diff
@@ -101,11 +101,8 @@
 											   default="Expires after"
 									/>
 								</label>
-								<g:jqueryDatePicker id="startDate"
-													name="startDate"
+								<g:jqueryDatePicker name="startDate"
```
jmiranda, line 104: @ewaterman this is the change I was talking about (re: removing "id"). This change works in this case, but just be careful when the name property is duplicated i.e. dynamically added rows.

grails-app/views/productAssociation/create.gsp
```diff
@@ -90,24 +90,6 @@
...
-							</tr>
-						
-							<tr class="prop">
-								<td valign="top" class="name">
-									<label for="lastUpdated"><warehouse:message code="productAssociation.lastUpdated.label" default="Last Updated" /></label>
-								</td>
-								<td valign="top" class="value ${hasErrors(bean: productAssociationInstance, field: 'lastUpdated', 'errors')}">
-									<g:datePicker name="lastUpdated" precision="minute" value="${productAssociationInstance?.lastUpdated}"  />
-								</td>
-							</tr>
```
ewaterman, line 109: jokes on me, I was testing against a date picker whose value wasn't actually used. I removed it here so that we don't get confused about it in the future

grails-app/views/productAssociation/create.gsp
ewaterman: to update this thread, it turns out that we actually do need to always require precision because if the given value is null (which will happen if the associated field is null and so the date picker should display empty), then it can't be smart and figure out the precision based on type because all it has is null.

PR #5499, opened 2025-09-16
OBPIH-7160 Add filter for expired to stock transfer

alannadolny:


grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
```diff
@@ -1068,16 +1069,22 @@ class ProductAvailabilityService {
...
+    List<ProductAvailability> getStockTransferCandidates(Location location, Boolean showExpiredItemsOnly = false) {
         return ProductAvailability.createCriteria().list {
             eq("location", location)
             gt("quantityOnHand", 0)
+
+            if (showExpiredItemsOnly) {
+                inventoryItem(JoinType.LEFT_OUTER_JOIN.joinTypeValue) {
+                    lt('expirationDate', new Date())
+                }
+            }
```
ewaterman, line 1080: Do we not want an inner join? If there is no inventory item, then we should hide the row, right? We only want to see expired, and so if there's no item, it can't be expired.

jmiranda: There should always be an inventory item (only bin location can be null, I think) so I think this makes sense.

PR #5502, opened 2025-09-17
OBPIH-7504cfixes to product inventory migration warnings

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7504

**Description:** See comments in the ticket. Two issues:

1) There are issues with two product inventory transactions existing at the same time so I changed it to also warn in that case
2) We weren't seeing warnings for two transactions at the same time on the same lot. I think this was just a bug

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
@@ -1090,18 +1090,14 @@ class MigrationService {
             FROM transaction_entry te1
                      JOIN inventory_item ii1 ON te1.inventory_item_id = ii1.id
                      JOIN transaction t1 ON te1.transaction_id = t1.id
-                     JOIN transaction_entry te2 ON te1.inventory_item_id != te2.inventory_item_id
+                     JOIN transaction_entry te2 ON te1.transaction_id != te2.transaction_id
                      JOIN inventory_item ii2 ON te2.inventory_item_id = ii2.id
                      JOIN transaction t2 ON te2.transaction_id = t2.id
             WHERE
               ii1.product_id = ii2.product_id
               AND t1.transaction_date = t2.transaction_date
               AND t1.id < t2.id
-              AND (
-                (t1.transaction_type_id = :transactionTypeId and t2.transaction_type_id != :transactionTypeId) 
-                OR 
-                (t1.transaction_type_id != :transactionTypeId and t2.transaction_type_id = :transactionTypeId)
-              )
+              AND (t1.transaction_type_id = :transactionTypeId OR t2.transaction_type_id = :transactionTypeId)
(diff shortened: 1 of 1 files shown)


grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -1090,18 +1090,14 @@ class MigrationService {
             FROM transaction_entry te1
                      JOIN inventory_item ii1 ON te1.inventory_item_id = ii1.id
                      JOIN transaction t1 ON te1.transaction_id = t1.id
-                     JOIN transaction_entry te2 ON te1.inventory_item_id != te2.inventory_item_id
+                     JOIN transaction_entry te2 ON te1.transaction_id != te2.transaction_id
```
ewaterman, line 1097: I think this is just a bug? We want to warn when there are two different transactions at the same time, not two different items.

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -1090,18 +1090,14 @@ class MigrationService {
...
             WHERE
               ii1.product_id = ii2.product_id
               AND t1.transaction_date = t2.transaction_date
               AND t1.id < t2.id
-              AND (
-                (t1.transaction_type_id = :transactionTypeId and t2.transaction_type_id != :transactionTypeId) 
-                OR 
-                (t1.transaction_type_id != :transactionTypeId and t2.transaction_type_id = :transactionTypeId)
-              )
+              AND (t1.transaction_type_id = :transactionTypeId OR t2.transaction_type_id = :transactionTypeId)
```
ewaterman: See this comment for context: https://pihemr.atlassian.net/browse/OBPIH-7504?focusedCommentId=167388.

From our investigation, we get issues when there are two product inventories at the same time that both cause adjustments since BOTH adjustments will be applied after we migrate.

I discussed with Manon and we thought the simplest solution was to just also warn about two product inventory transactions existing at the same time so that they can be fixed manually but let me know your thoughts @awalkowiak.

awalkowiak: @ewaterman Overall sounds good, but after this change, I think, all "single product" product inventory transactions will be displayed there as well, which are handled automatically, and these will be a noise on the migration page. So we'd need to just display these that  are made from import (multiple products) and have common products

grails-app/services/org/pih/warehouse/data/MigrationService.groovy
awalkowiak: > I think this is just a bug?

Yeah, I think it makes sense, because he previous version would skip two transactions made for the same inventory items

awalkowiak (approved):

PR #5503, opened 2025-09-18
OBLS-157: Handled partial putaway and shortage APIs

druchniewicz:


grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
```diff
@@ -267,6 +285,81 @@ class PutawayTaskService {
...
+                    it.quantity == quantityRemaining &&
+                    it.orderItemStatusCode == OrderItemStatusCode.PENDING
+        } as OrderItem
+
+        return PutawayTask.createFromOrderItem(remainingOrderItem)
+    }
+
+    void shortage(PutawayTask task, ReasonCode reasonCode) {
+        def putawayDiscrepancyLocation = task.facility.internalLocations
+            .find { it.supports(ActivityCode.PUTAWAY_DISCREPANCY) }
```
jmiranda: We should throw an exception here (or implicitly create it) if the putaway discrepancy location doesn't exist.

grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
```diff
@@ -267,6 +285,81 @@ class PutawayTaskService {
...
+        currentItem.orderItemStatusCode = OrderItemStatusCode.CANCELED
+        currentItem.save(flush: true, failOnError: true)
+
+        OrderItem remainingOrderItem = order.orderItems.find {
+            it.parentOrderItem?.id == originalItem.id &&
+                    it.quantity == quantityRemaining &&
+                    it.orderItemStatusCode == OrderItemStatusCode.PENDING
+        } as OrderItem
+
+        return PutawayTask.createFromOrderItem(remainingOrderItem)
```
jmiranda: Unless I'm missing something, I think this should be moved to the PutawayTaskAdapter if possible. 
```
PutawayTask toPutawayTask(OrderItem orderItem)
```

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
```diff
@@ -129,4 +133,12 @@ class PutawayTaskAdapter {
             default: return OrderStatus.PENDING
         }
     }
+
+    static OrderItemStatusCode toOrderItemStatusCode(PutawayTaskStatus putawayTaskStatus) {
+        switch (putawayTaskStatus) {
+            case PutawayTaskStatus.COMPLETED: return OrderItemStatusCode.COMPLETED
+            case PutawayTaskStatus.CANCELED: return OrderItemStatusCode.CANCELED
+            default: return OrderItemStatusCode.PENDING
```
jmiranda, line 71: If I'm not mistaken, I think we need to add two more status codes (STARTED, IN_PROGRESS) to handle those two states, right? When we were using the OrderStatus, I co-opted existing status codes, but now we can just add new ones because there's not enough specificity in the current status set.
```
enum OrderItemStatusCode {
    PENDING(1),
    STARTED(2),
    IN_PROGRESS(3),
    COMPLETED(4),   // Meaning depends on context i.e. picked, putaway, delivered, received
    CANCELED(5),
    BACKORDER(6)
...
```

src/main/groovy/org/pih/warehouse/core/ActivityCode.groovy
```diff
@@ -111,6 +111,9 @@ enum ActivityCode {
     // Inbound sortation
     INBOUND_SORTATION('INBOUND_SORTATION'),
 
+    // Putaway discrepancy
+    PUTAWAY_DISCREPANCY('PUTAWAY_DISCREPANCY'),
```
jmiranda, line 115: I hadn't thought of using the activity code for this, but couldn't figure out the best way to model this. So I think it's a good idea for now. 

Eventually, we might want to make these locations a little more generic at some point i.e. LOST_AND_FOUND. So we can look for a LOST_AND_FOUND location within the zone where the discrepancy occurs Inbound Sortation Zone (putaway discrepancy) or Storage Zone 1 (picking discrepancy). 

Actually, let's do this. Call the activity code LOST_AND_FOUND. And we can have one Lost & Found location for the whole facility. The actual location will be called Lost & Found and it will be of type Discrepancy Location with activity code (HOLD_STOCK, LOST_AND_FOUND).

jmiranda (commented):
A few changes requested, but looks good.

grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
druchniewicz: I changed it to throw exception if location does not exist. We go with this approach in other cases as well. The entire configuration should be provided.

src/main/groovy/org/pih/warehouse/core/ActivityCode.groovy
druchniewicz: I created new location type with name 'Discrepancy' and new activity code LOST_AND_FOUND. This new location type supports HOLD_STOCK and LOST_AND_FOUND activity codes.
When I look for the discrepancy location I use following:
`def putawayDiscrepancyLocation = task.facility.internalLocations
            .find { it.locationType.name == Constants.DISCREPANCY_LOCATION_TYPE}`
            
 It finds first location with discrepancy location type. This location should already be created. Otherwise exception will be thrown

src/main/groovy/org/pih/warehouse/order/OrderItemStatusCode.groovy
```diff
@@ -3,10 +3,11 @@ package org.pih.warehouse.order
 enum OrderItemStatusCode {
 
     PENDING(1),
-    COMPLETED(2),   // Meaning depends on context i.e. picked, putaway, delivered, received
-    CANCELED(3),
-    BACKORDER(4)
-
+    IN_PROGRESS(2),
+    IN_TRANSIT(3),
```
jmiranda: Ah, I see you did add the codes. Since we're adding them from scratch (not mapping to existing ones) let's use the correct language i.e. STARTED instead of IN_PROGRESS and IN_PROGRESS instead of IN_TRANSIT. 

Note: I would recommend replacing IN_PROGRESS first otherwise you'll need to untangle whether each IN_PROGRESS means STARTED or IN_PROGRESS.

grails-app/migrations/0.9.x/changelog-2025-09-17-1435-add-column-order-item-reason-code.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    <changeSet author="druchniewicz" id="170920251435-0">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="order_item" columnName="reason_code"/>
+            </not>
+        </preConditions>
+
+        <addColumn tableName="order_item">
+            <column name="reason_code" type="VARCHAR(255)"/>
```
jmiranda: Let's call this discrepancy_reason_code. 

And I just realized that it would be a good idea to track this any time we split putaway tasks. For example, if there's not enough room in the primary location we can set the reason code = INSUFFICIENT_CAPACITY. If we return the split quantity into a hold location we can provide a reason code like DAMAGED or QUARANTINE.

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
```diff
@@ -34,80 +35,21 @@ class PutawayTaskAdapter {
...
 
         // assign - putaway task has been loaded into putaway container (change if user provided override)
         orderItem.containerLocation = task.container
 
         orderItem.destinationBinLocation = task.destination
 
-        // complete - task is completed
-        order.completedBy = task.completedBy
-        order.dateCompleted = task.dateCompleted
+        orderItem.reasonCode = task.reasonCode
```
jmiranda: Is this being set on the original (parent) order item?

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
```diff
@@ -1095,6 +1095,11 @@ class UrlMappings {
             action = [POST: "getCycleCountSummary", GET: "getCycleCountSummary"]
         }
 
+        "/api/reasonCodes"(parseRequest: true) {
```
jmiranda: @awalkowiak don't we already have a reason codes API? Or have we not merged that change from develop yet?

jmiranda: @druchniewicz change this to `/api/reason-codes`.

jmiranda (changes requested):
@druchniewicz On Monday, please make the three suggested changes and then merge this so we can start testing demo scenarios.
* URL mapping: reasonCodes => reason-codes
* Database migration: reason_code => discrepancy_reason_code
* OrderItemStatusCode: Replace the two statuses (IN_PROGRESS=>STARTED, IN_TRANSIT=>IN_PROGRESS)

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
druchniewicz: > @awalkowiak don't we already have a reason codes API? Or have we not merged that change from develop yet?

we already have and I used it. I just added a missing mapping here.
API is avaiable in ReasonCodeApiController and it takes set of reason codes from ReasonCode depending on activity code param

grails-app/migrations/0.9.x/changelog-2025-09-17-1435-add-column-order-item-reason-code.xml
druchniewicz: ok so I will change behavior in the mobile app that discrepancy code will be required each time when partial putaway is performed

grails-app/controllers/org/pih/warehouse/UrlMappings.groovy
druchniewicz: I think we can stick current approach and just add a new set of reason codes for discrepancy and wrap it with some activity code (see [ReasonCodeApiController](https://github.com/openboxes/openboxes/blob/feature/obaf-integration/grails-app/controllers/org/pih/warehouse/api/ReasonCodeApiController.groovy))

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
druchniewicz: No, it is set on the task (order item) that is being processed now

src/main/groovy/org/pih/warehouse/order/OrderItemStatusCode.groovy
druchniewicz: I just wanted to map putaway task status to order item status similarly to mapping between putaway task status and order status (in_progress -> approved, in_transit -> placed) etc. but there were no appropriate statuses in OrderItemStatusCode enum that's why I added in_progress and in_transit because previously we had only Pending and next status was Completed

grails-app/migrations/0.9.x/changelog-2025-09-17-1435-add-column-order-item-reason-code.xml
jmiranda: Make the change to the reason_code property name but don't change the required behavior in the mobile app unless it's easy to do. I created a comment in the ticket to track all of my feedback from the video and will create separate tickets for all of it.

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
jmiranda: Ok, we can live with that for now but it might make sense to have it set on the parent order item in the future, just to keep everything in one place i.e. status = CANCELED, discrepancy_reason_code = INSUFFICIENT_CAPACITY

jmiranda: Although i'm just realizing that we could split multiple times so maybe the discrepancy reason code is at the right level.

src/main/groovy/org/pih/warehouse/order/OrderItemStatusCode.groovy
jmiranda: Yeah that made total sense. The only reason I used IN_PROGRESS and IN_TRANSIT is because it meant I didn't have to add new OrderStatusCode values. Since we need to add them to OrderItemStatusCode, I figured we'd map them correctly this time.

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
druchniewicz: does it mean that only CANCELED tasks should have the discrepancy reason codes? (if yes then probably column with reason code that I've added in web application is not needed - because there are no canceled tasks displayed)

druchniewicz: and regarding parent items - in one of the ticket you said that all order items should be created under the original order item.
simple example below:
**TASK A with quantity = 100**
we split it to two tasks:
**TASK B with quantity = 30 -> this has parent TASK A**
**TASK C with quantity = 70 -> this has parent TASK A**
(TASK A is canceled now)

more complex example below:
**TASK A with quantity = 100**
we split it to two tasks:
**TASK B with quantity = 30 -> this has parent TASK A**
**TASK C with quantity = 70 -> this has parent TASK A**
(TASK A is canceled now)
we make partial putaway once again, we split TASK C to the new two tasks:
**TASK D with quantity 45 -> this has parent TASK A**
**TASK E With quantity 25 -> this has parent TASK A**
(TASK C is canceled now)

as you can see all items are created under original order item (task A) but I'm not sure if this is 100% correct - because e.g. Task D and Task E have parent Task A but in real they have been created from Task C.

grails-app/domain/org/pih/warehouse/putaway/PutawayTask.groovy
```diff
@@ -42,6 +43,7 @@ class PutawayTask {
     Location facility
     Location container                  // container scanned during putaway
     Location destination                // target storage or outbound staging
+    ReasonCode reasonCode
```
jmiranda: In case it wasn't clear, change this to discrepancyReasonCode as well.

src/main/groovy/org/pih/warehouse/api/PutawayTaskAdapter.groovy
jmiranda: > does it mean that only CANCELED tasks should have the discrepancy reason codes? (if yes then probably column with reason code that I've added in web application is not needed - because there are no canceled tasks displayed)

Sorry I missed this comment the other day. Make sure to @ me to make sure I see it. I get so many notifications from jira, github, et al.

image

that I really only look at my mentions only a semi-regular basis.

image

jmiranda: I think I was initially thinking that the original item would just keep splitting into more "active" children, not necessarily creating multiple canceled items under the original item. We just want to create splits. And then potentially have one canceled item to handle the "we can't find this item" use case.

I think I can live with either but the canceled items as siblings (without an association back to what replaced it) seems problematic. So is there a way where we can get rid of the canceled item? If that means we need the new split items to carry the discrepancy reason code to explain why they exist, that might be enough to explain the history. 

@awalkowiak thoughts?

grails-app/services/org/pih/warehouse/putaway/PutawayTaskService.groovy
```diff
@@ -200,6 +238,16 @@ class PutawayTaskService {
             throw new IllegalStateException("Container does not exist")
         }
 
+        def allowedActivityCodes = [
+                ActivityCode.PUTAWAY_CART,
+                ActivityCode.PICK_CART,
```
jmiranda: Let's jus tuse PUTAWAY_CART for this one.

druchniewicz: corrected

PR #5504, opened 2025-09-22
OBPIH-7512 replace string key with AvailableItemKey when working with product av…

ewaterman:
…ailability

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7512

**Description:**

When calculating QoH, we construct a map of available items that is keyed on a string value that is constructed from product code, bin location name, and lot number:
```
static String constructAvailableItemKey(Location binLocation, InventoryItem inventoryItem) {
    return "${inventoryItem.product.productCode}-${binLocation?.name}-${inventoryItem?.lotNumber}"
}
```
https://github.com/openboxes/openboxes/blob/release/0.9.5-hotfix2/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy

This is problematic however because it results in both the null lot and a lot with string "null" resolving to the same key:
```
String lotNumber = 'null'
println "${lotNumber}"          -> "null"
println "${lotNumber}" == null  -> false

lotNumber = null
println "${lotNumber}"          -> "null"
println "${lotNumber}" == null  -> false
```

I changed this logic to use an actual object as the key: `AvailableItemKey`. AvailableItemKey still uses bin location name and lot, but it keeps them as separate String fields, which should resolve the issue of null vs 'null'. Using a proper key class also means we don't need to pass around magic strings and unlabelled keys. I also introduced the `AvailableItemMap` as a convenience wrapper on the map so that it'd be easier to work with.

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/MigrationService.groovy
@@ -926,7 +926,7 @@ class MigrationService {
                     AvailableItemKey itemKey = new AvailableItemKey(binLocation, inventoryItem)
                     String comments = entries.comments.findAll { it }.join(', ')
                     [(itemKey): [
-                        binLocation: inventoryItem,
+                        binLocation: binLocation,
                         inventoryItem: inventoryItem,
                         quantity: entries.sum { it.quantity },
                         comments: comments.length() > 255 ? comments.substring(0, 255) : comments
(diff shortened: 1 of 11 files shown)


grails-app/services/org/pih/warehouse/data/MigrationService.groovy
```diff
@@ -555,7 +558,7 @@ class MigrationService {
         }
 
         Date previousTransactionDate = null
-        transactions.each { Transaction it ->
+        for (Transaction it in transactions) {
```
ewaterman, line 561: Debugging was really difficult when this was in an "each" closure because intellij debug fields get all messed up. More importantly, stack traces get really garbled inside closures so I generally prefer to use regular loops unless it's a really simple operation

awalkowiak (approved):

PR #5505, opened 2025-09-23
OBPIH-7508 Header and form implementation

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7508

**Description:**
In this ticket, I updated the header’s status indicator and added buttons (currently mocked). I also implemented a form with working validation, which you can test by clicking the Send Shipment button.

---
image

### DIFF
--- a/src/js/hooks/inboundV2/send/useInboundSendForm.js
+++ b/src/js/hooks/inboundV2/send/useInboundSendForm.js
@@ -16,6 +16,7 @@ import useQueryParams from 'hooks/useQueryParams';
 import useSpinner from 'hooks/useSpinner';
 import useTranslate from 'hooks/useTranslate';
 import apiClient from 'utils/apiClient';
+import createInboundWorkflowHeader from 'utils/createInboundWorkflowHeader';
 
 const useInboundSendForm = () => {
   const {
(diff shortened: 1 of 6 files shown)


src/css/main.scss
```diff
@@ -273,14 +273,30 @@ body {
...
+}
+.font-size-md {
+  font-size: 1.25rem;
+}
+.font-size-lg {
+  font-size: 1.5rem;
+}
+.font-size-xl {
+  font-size: 2rem;
+}
```
SebastianLib, line 290: We don’t currently have utility classes for font sizes in our Bootstrap because we are using version 4.6.2. Starting from Bootstrap 5, there are font size classes like fs-4, but these don’t exist in our current version. It might be worth adding these classes for consistency, since we don’t have anything similar at the moment. Maybe we could also consider bumping up the Bootstrap version in the future?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -0,0 +1,164 @@
...
+        expectedDeliveryDate: data.expectedDeliveryDate || null,
+      });
+
+      dispatch(
+        updateWorkflowHeader([
+          { text: data.identifier, color: '#000000', delimeter: ' - ' },
+          { text: data.origin.name, color: '#004d40', delimeter: ' to ' },
+          { text: data.destination.name, color: '#01579b', delimeter: ', ' },
+          { text: data.dateRequested, color: '#4a148c', delimeter: ', ' },
+          { text: data.description, color: '#770838', delimeter: '' },
```
alannadolny: Is it something that will be reused? Is it worth assigning it to a const in a separate file?

SebastianLib: Good observation. I think so. I’ll change it in a moment

PR #5507, opened 2025-09-23
OBPIH-7489 React date picker and utils using date fns

alannadolny:
date-fns uses a date object for storing dates, so it automatically transfers dates from the timezone obtained from the API response to the user's timezone. (Date object is always converted to the user's timezone). In the need for having a date without timezone information, there is a method for clearing timezone information. Date after clearing timezone information can be passed directly to the datepicker.


package.json
```diff
@@ -84,6 +85,7 @@
     "react-color": "2.19.3",
     "react-confirm-alert": "2.6.1",
     "react-datepicker": "1.8.0",
+    "react-datepicker-6": "npm:react-datepicker@6.9.0",
```
alannadolny, line 88: Alias for react-datepicker version 6, we have both libraries at the same time to avoid causing a lot of issues with date libraries under the hood (react-datepicker 1.8.0 uses moment.js, every newer version uses date-fns). We are not able to bump up that library to the newest version because of changes between React 16.8 and 17.0+. Version 6.9.0 is the newest version that actually works for us.

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+
+  const formatDate = (dateToFormat) => {
+    if (!dateToFormat) {
+      return null;
+    }
+
+    try {
+      return parse(
+        dateToFormat,
+        getDateFormat(),
```
alannadolny, line 65: Using the parse method, we should ensure that the format of the passed date matches the result of getDateFormat()

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+    onChange(formatted);
+  };
+
+  const dateFnsLocale = () => {
+    if (!currentLocale || ['en', 'ar'].includes(currentLocale)) {
+      return locales.enUS;
+    }
+
+    return locales[currentLocale];
+  };
```
alannadolny, line 100: date-fns uses its own local codes, so there is a mapping between our local codes and those used in date-fns

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+      required={required}
+      tooltip={tooltip}
+      errorMessage={errorMessage}
+      button={button}
+      hideErrorMessageWrapper={hideErrorMessageWrapper}
+      className={wrapperClassName}
+    >
+      <DatePicker
+        {...fieldProps}
+        ariaLiveMessages={{}}
```
alannadolny, line 122: ariaLiveMessages is a new prop for displaying a message above the date field. By passing an empty object, we are hiding that message.

src/js/components/form-elements/v2/DateFieldInput.jsx
```diff
@@ -24,7 +24,16 @@ const DateFieldInput = forwardRef(({
...
+      return (
+        <button
+          className="border-0 bg-inherit"
+          type="button"
+          aria-label="clear"
+          onClick={onClear}
+        >
+          <RiCloseLine />
+        </button>
+      );
```
alannadolny, line 36: Wrapping the icon allows us to test the close button

src/js/components/stock-movement-wizard/outboundImport/subsections/OutboundImportBasicDetails.jsx
alannadolny, line 1: Example usage of new datepicker (in outbound import)

src/js/components/stock-movement-wizard/outboundImport/subsections/OutboundImportSendingOptions.jsx
alannadolny, line 1: Example usage of new datepicker (in outbound import) + time picker

src/js/consts/timeFormat.js
```diff
@@ -4,9 +4,35 @@ export const TimeFormat = {
...
+  const map = {
+    [DateFormat.MM_DD_YYYY]: DateFormatDateFns.MM_DD_YYYY,
+    [DateFormat.DD_MMM_YYYY]: DateFormatDateFns.DD_MMM_YYYY,
+    [DateFormat.MMM_DD_YYYY]: DateFormatDateFns.MMM_DD_YYYY,
+    [DateFormat.MMM_DD_YYYY_HH_MM_SS]: DateFormatDateFns.MMM_DD_YYYY_HH_MM_SS,
+    [DateFormat.MM_DD_YYYY_HH_MM_Z]: DateFormatDateFns.MM_DD_YYYY_HH_MM_Z,
+  };
+
+  return map[format] ?? DateFormatDateFns[format];
 };
```
alannadolny, line 38: date-fns uses different date formats, so to allow us to simply convert the formats, there is a converter that can be deleted after getting rid of moment.js

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,229 @@
+import * as locales from 'date-fns/locale';
```
ewaterman, line 1: nice tests 💪

src/js/utils/dateUtils.js
```diff
@@ -17,4 +21,133 @@ const dateWithoutTimeZone = ({
...
+  const dateWithoutOffset = date.replace(/([+-]\d{2}:\d{2}|Z)$/, '');
+  const formatWithoutOffset = providedFormat.replace('XXX', '');
+  return parse(dateWithoutOffset, formatWithoutOffset, new Date());
+};
+
+/**
+ * Converts an ISO date string to a Date object or a "local date"
+ * @param {Object} params
+ * @param {string} params.date - ISO date string (e.g. '2025-09-18' or '2025-09-18T14:30:00')
+ * @param {Boolean} params.localDate - if true, returns the full date-time; if false, returns
```
ewaterman: is "local date" a date fns thing or a term you came up with?

On the backend "LocalDate" is a java.time class that means a date with no time or zone component. So something like "2025-09-18" would be represented by a LocalDate. It seems like that's the opposite of what you have here, which is confusing.

Why not inverse the condition and make the boolean "dateOnly"? Or am I misunderstanding?

src/js/consts/timeFormat.js
```diff
@@ -4,9 +4,35 @@ export const TimeFormat = {
...
 
 export const DateFormat = {
-  MM_DD_YYYY: 'MM/DD/YYYY',
-  DD_MMM_YYYY: 'DD/MMM/YYYY',
-  MMM_DD_YYYY: 'MMM DD, YYYY',
+  MM_DD_YYYY: 'MM/DD/yyyy',
+  DD_MMM_YYYY: 'DD/MMM/yyyy',
+  MMM_DD_YYYY: 'MMM DD, yyyy',
   MMM_DD_YYYY_HH_MM_SS: 'MMM DD, YYYY HH:mm:ss',
```
ewaterman: why change to lowercase 'y'? Doesn't moment js use uppercase? Also why is MMM_DD_YYYY_HH_MM_SS unchanged?

src/js/consts/timeFormat.js
```diff
@@ -4,9 +4,35 @@ export const TimeFormat = {
...
+// see: https://github.com/date-fns/date-fns/blob/main/docs/unicodeTokens.md
+export const DateFormatDateFns = {
+  MM_DD_YYYY: 'MM/dd/yyyy',
+  DD_MMM_YYYY: 'dd/MMM/yyyy',
+  MMM_DD_YYYY: 'MMM dd, yyyy',
+  MMM_DD_YYYY_HH_MM_SS: 'MMM dd, yyyy HH:mm:ss',
+  MM_DD_YYYY_HH_MM_Z: 'MM/dd/yyyy HH:mm XXX',
+  YYYY_MM_DD: 'yyyy-MM-dd',
+  YYYY_MM_DD_HH_MM_Z: "yyyy-MM-dd'T'HH:mmXXX",
+  YYYY_MM_DD_HH_MM_SS: "yyyy-MM-dd'T'HH:mm:ss",
```
ewaterman, line 24: it'd be awesome if we could do like we did in: https://github.com/openboxes/openboxes/pull/5485/ and localize these formats instead of hard coding them but we can save that for a future PR.

We'd want to change this to be something like: https://github.com/openboxes/openboxes/blob/develop/src/main/groovy/org/pih/warehouse/core/date/DateDisplayStyle.groovy

Also I see that we have a lot more formats here. Again, not for this PR, but it'd be great if we could use this as an opportunity to reduce down to a smaller set of formats that we use. Having all of `MM/dd/yyyy` and `yyyy-MM-dd` and `dd/MMM/yyyy` seems redundant

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,229 @@
...
+    });
+    const plDate = parseDateToString({
+      date: new Date(2025, 8, 19),
+      dateFormat: DateFormatDateFns.MMM_DD_YYYY,
+      options: { locale: locales.pl },
+    });
+    expect(esDate).toBe('sep 19, 2025');
+    expect(frDate).toBe('sept. 19, 2025');
+    expect(plDate).toBe('wrz 19, 2025');
+  });
```
ewaterman, line 154: could also add some tests on parseDateToString when we have a Date with a time component

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,229 @@
...
+
+describe('formatDateToZonedDateTimeString()', () => {
+  it('should return ISO-formatted date', () => {
+    const date = formatDateToZonedDateTimeString(
+      new Date(2025, 8, 19),
+    );
+    // examples that should match the following expression:
+    // 09/19/2025 00:00 +02:00
+    // 09/19/2025 00:00 -05:00
+    expect(date).toMatch(/^09\/19\/2025 00:00 [+-]\d{2}:\d{2}$/);
```
ewaterman: what timezone is it defaulting to? System local time?

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,229 @@
...
+});
+
+describe('formatDateToLocalDateString()', () => {
+  it('should return date string without timezone', () => {
+    const date = formatDateToLocalDateString(
+      new Date(2025, 8, 19),
+    );
+
+    expect(date).not.toMatch(/^19\/Sep\/2025 00:00 [+-]\d{2}:\d{2}$/);
+    expect(date).toMatch(/^19\/Sep\/2025/);
```
ewaterman: why toMatch and not toBe?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +21,133 @@ const dateWithoutTimeZone = ({
...
+
+/**
+ * Converts an ISO date string to a Date object or a "local date"
+ * @param {Object} params
+ * @param {Date} params.date - date object
+ * @param {String} params.dateFormat - desired date format
+ * @param {String} options.locale - desired locale
+ * @returns {Date | null}
+ */
+export const parseDateToString = ({
```
ewaterman: I'd name this formatDateToString if you want to be consistent with the below methods.

Also the docstring looks incorrect.

Also also why do we need to export this method? Can't it just be private and the below formatDateToXStrings be the only ones that use it? In what scenario do we need to use this method instead of the below ones? Is it when we need to customize the format?

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,229 @@
...
+    );
+
+    const frDate = formatDateToLocalDateString(
+      new Date(2025, 8, 19),
+      locales.fr,
+    );
+
+    expect(esDate).toBe('19/sep/2025');
+    expect(frDate).toBe('19/sept./2025');
+  });
```
ewaterman, line 210: test case: what happens if you give formatDateToLocalDateString a Date with a time component?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +21,133 @@ const dateWithoutTimeZone = ({
...
+  dateFormat: DateFormatDateFns.DD_MMM_YYYY,
+  options: {
+    locale,
+  },
+});
+
+/**
+ * A method for converting Date to a localized date string for display (skipping timezone)
+ */
+export const formatDateToDateOnlyString = (date, locale = locales.enUS) => {
```
ewaterman, line 132: so this does the following:
1) takes a Date and converts it to a datetimezone string
2) converts that datetimezone string to a new Date but zeroes out the time and zone
3) converts the Date to a date string

Can't you just do?
```
const year = getYear(date);
const month = getMonth(date);
const day = getDate(date);

const dateOnly = new Date(year, month, day);
```

src/js/utils/dateUtils.js
```diff
@@ -17,4 +21,133 @@ const dateWithoutTimeZone = ({
...
+export const formatDateToZonedDateTimeString = (date) => parseDateToString({
+  date,
+  dateFormat: DateFormatDateFns.MM_DD_YYYY_HH_MM_Z,
+});
+
+/**
+ * A method for converting Date to a localized date string for display (shifted time by timezone
+ * differences)
+ */
+export const formatDateToLocalDateString = (date, locale = locales.enUS) => parseDateToString({
```
ewaterman: okay so "local date" means a localized datetime? Given the confusion with java.time.LocalDate, I think something like `formatDateToDatetimeString` is clearer.

src/js/consts/timeFormat.js
alannadolny: moment.js works with both. The change is made because we are fetching the date format with "y", so to be able to convert it to date-fns using the converter that I created, I changed it to "y"

src/js/tests/utils/dateUtils.test.jsx
alannadolny: I forgot to change that :smile:

src/js/tests/utils/dateUtils.test.jsx
alannadolny: Yup, the date object converts it to the system local time

src/js/utils/dateUtils.js
alannadolny: When I did it like that, I got a year after converting to the local timezone :/

src/js/utils/dateUtils.js
alannadolny: > Is it when we need to customize the format

:point_up:

src/js/consts/timeFormat.js
ewaterman: from the docs it looks like 'y' is "era year" which differs slightly from 'Y'.https://momentjs.com/docs/#/displaying/format/

image

But I'm okay with it as long as you've double checked that the change has no actual impact

src/js/utils/dateUtils.js
```diff
@@ -17,4 +21,134 @@ const dateWithoutTimeZone = ({
...
+  const formatWithoutOffset = providedFormat.replace('XXX', '');
+  return parse(dateWithoutOffset, formatWithoutOffset, new Date());
+};
+
+/**
+ * Converts an ISO date string to a Date object or a date-only Date without timezone information
+ * @param {Object} params
+ * @param {Object} params.options
+ * @param {string} params.date - ISO date string (e.g. '2025-09-18' or '2025-09-18T14:30:00')
+ * @param {Boolean} params.dateOnly - if true, returns the full date-time; if false, returns
```
ewaterman, line 41: this should be the inverse. dateOnly == true means we return a date with time zeroed. All the logic in this method will need to be flipped too.

src/js/utils/dateUtils.js
ewaterman: Maybe related to this? https://github.com/date-fns/date-fns/issues/1539

Could also try date.getFullYear() or date.toLocaleDateString()

idk it just feels like there are extra steps here that aren't needed

src/js/tests/utils/dateUtils.test.jsx
ewaterman: hmm do we want that? Shouldn't we always try to default to the user's timezone?

also, does this test pass for UTC (which is what the github test runners use)? Won't it output 'Z' as the timezone?

btw if you want to be able to assert against just a string you could probably get the system timezone then do something like this:
```
const timezone = <compute the system timezone somehow>
expect(date).toBe('19/Sep/2025 00:00' + timezone);
```

ewaterman:
I don't know where to talk about this so I'll just put it in a comment here.

With the Date refactor on the backend, we're changing the format of accepted/returned values:
- LocalDate fields:
  - request: ISO format preferred or `MM/dd/yyyy` (for backwards compatability) (if you provide a time or tz it errors)
  - response: always as `yyyy-MM-dd` ie ISO format. Ex: "2000-01-01"
- Instant fields:
  - request: ISO format preferred or `MM/dd/yyyy` + at least hh:mm and a zone (if you leave hour hours, minutes or tz it errors)
  - response: always as `yyyy-MM-dd'T'hh:mm:ss.SSSXXX` ie ISO format. Ex: "2000-01-01T00:00:00.000Z"

So I just want to double check that we'd be able to properly handle each of the above using these util methods.

I don't think you're using these util methods on the old dates, but if you are then until we complete the Date refactor we still have all the old Date fields:
  - request: ISO format preferred or `MM/dd/yyyy` or `MM/dd/yyyy HH:mm XXX`
  - response: defaults to the full datetimezone ISO format like Instant but we also have a ton of different ways of formatting it that are configured per entity/API.

src/js/utils/dateUtils.js
```diff
@@ -17,4 +27,149 @@ const dateWithoutTimeZone = ({
...
+  dateOnly = false,
+  options = {
+    providedDateFormat: DateFormatDateFns.MMM_DD_YYYY,
+  },
+}) => {
+  if (!date) {
+    return null;
+  }
+
+  if (dateOnly && !options.providedDateFormat) {
```
ewaterman, line 57: does this need to be `!dateOnly`? That's what the error message indicates

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+        scrollableYearDropdown
+        disabledKeyboardNavigation
+        placeholderText={placeholderText}
+        popperContainer={RootPortalWrapper}
+        popperPlacement="bottom-start"
+        selected={selectedDate}
+        highlightDates={highlightedDates}
+        onChange={onChangeRaw || onChangeHandler}
+        onSelect={() => {
+          datePickerRef.current?.setOpen(false);
```
kchelstowski, line 143: out of curiosity - is it suggested by the library itself to rely on refs, or is it you who did it for the performance reasons?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+    } catch (e) {
+      return null;
+    }
+  };
+
+  const selectedDate = formatDate(value);
+  const highlightedDates = [selectedDate || new Date()];
+
+  const placeholderText = typeof placeholder === 'object'
+    ? translate(placeholder?.id, placeholder?.default)
```
kchelstowski: shouldn't this be `defaultMessage` instead of `default`?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +27,149 @@ const dateWithoutTimeZone = ({
   return parsedDate.format(outputDateFormat);
 };
 
+export const parseDateAndStripTimezone = (date, providedFormat) => {
+  const dateWithoutOffset = date.replace(/([+-]\d{2}:\d{2}|Z)$/, '');
```
kchelstowski, line 31: what do we mean by "offset" here?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +27,149 @@ const dateWithoutTimeZone = ({
...
+  const formatWithoutOffset = providedFormat.replace('XXX', '');
+  return parse(dateWithoutOffset, formatWithoutOffset, new Date());
```
kchelstowski, line 33: adding this `new Date()` in a few places feels weird - why is it needed to include the current time in some methods?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +27,149 @@ const dateWithoutTimeZone = ({
...
+  const absMinutes = Math.abs(offsetMinutes);
+  const hours = Math.floor(absMinutes / 60);
+  const minutes = absMinutes % 60;
+
+  // parsing hours / minutes to appropriate format: hours: 02, 01, 14, minutes: 02, 01, 59
+  const paddedHours = hours.toString().padStart(2, '0');
+  const paddedMinutes = minutes.toString().padStart(2, '0');
+
+  return `${sign}${paddedHours}:${paddedMinutes}`;
+};
```
kchelstowski, line 173: I kinda dont understand this method. What is it supposed to return and where it could be useful?

src/js/utils/dateUtils.js
```diff
@@ -17,4 +27,149 @@ const dateWithoutTimeZone = ({
...
+
+/**
+ * Get timezone offset, defaulting to the user's timezone offset
+ * @param {Number} timezoneOffset
+ * @returns {`${string}${string}:${string}`}
+ */
+export const displayTimezoneOffset = (timezoneOffset = new Date().getTimezoneOffset()) => {
+  // timezoneOffset = difference in minutes comparing to utc (timezoneOffset is an argument for
+  // testing purposes, because we can't force Date object to use different timezone that the user's)
+  const offsetMinutes = -timezoneOffset;
```
kchelstowski, line 161: why is there a minus?

src/js/utils/dateUtils.js
alannadolny: It's ok. 
dateOnly = true -> we need to know what user passed, so the format is needed
dateOnly = false -> we assumed the date is in ISO format, so we can just parse it

src/js/components/form-elements/v2/DateFieldDateFns.jsx
alannadolny: I think it's done like that by Sebastian, it's needed for arrow navigation or something like that

src/js/components/form-elements/v2/DateFieldDateFns.jsx
alannadolny: it's copied from our current datepicker, but I assume it should

src/js/utils/dateUtils.js
alannadolny: this part of dates:
2025-10-02T08:15:30 **+01:00**
2025-10-02T12:45:00 **+02:00**
2025-10-02T19:05:10 **-04:00**

src/js/components/form-elements/v2/DateFieldDateFns.jsx
SebastianLib: In this case, it wasn't implemented due to arrow navigation :smile: 
image

src/js/utils/dateUtils.js
alannadolny: It's needed because of the way the library works. It looks weird, but it's what the doc suggests

src/js/utils/dateUtils.js
alannadolny: The getTimezoneOffset() method of Date instances returns the difference, in minutes, between this date as evaluated in the UTC time zone

src/js/utils/dateUtils.js
alannadolny: testing purposes for now

src/js/tests/utils/dateUtils.test.jsx
```diff
@@ -0,0 +1,276 @@
...
+    const emptyStringDate = formatDateToString({
+      date: '',
+    });
+    expect(nullDate).toBe(null);
+    expect(emptyStringDate).toBe(null);
+  });
+
+  it('should format date properly', () => {
+    const date = formatDateToString({
+      date: new Date(2025, 8, 19),
```
SebastianLib: From what I can see, this same date is used multiple times, so maybe it might be worth creating a separate variable to store it?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -0,0 +1,210 @@
...
+      onChange(null);
+      return;
+    }
+
+    const formatted = format(date, getDateFormat());
+    onChange(formatted);
+  };
+
+  const dateFnsLocale = () => {
+    if (!currentLocale || ['en', 'ar'].includes(currentLocale)) {
```
ewaterman, line 95: I'm just noticing this now but why do we need this part: `['en', 'ar'].includes(currentLocale)`?

I'm assuming it's because you need to map 'en' to 'enUS' for english.

But what about ar? That's arabic right? Not english.

SebastianLib: This is a workaround that I implemented. Using the 'ar' locale causes the app to crash when selecting a date, so for now we fallback to 'en' to avoid the crash. We talked about this a few months ago in an internal (soldevelo) standup, and agreed that this solution is fine for the moment.

src/js/utils/dateUtils.js
ewaterman: can you update the error message below then to reflect that?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
ewaterman: Thanks for the explanation. @alannadolny can you add a comment in the code explaining this so that we remember to fix it properly in the future?

src/js/tests/utils/dateUtils.test.jsx
ewaterman: I don't actually mind having "magic numbers" in tests. In this case I think it's fine either way, but it's sometimes best to have tests be a little "dumber" for the sake of not relying on anything outside of the individual test itself. We don't want tests depending on shared data since ideally if you change something for a test, it should break only that test. But again, in this scenario I think it's fine to use a constant.

src/js/components/form-elements/v2/DateFieldDateFns.jsx
awalkowiak: Maybe it is worth adding as a comment in the code directly

awalkowiak (approved):

PR #5509, opened 2025-09-24
OBLS-180 Expand requisition data model to handle priority and delivery type

jmiranda:


grails-app/domain/org/pih/warehouse/requisition/Requisition.groovy
```diff
@@ -124,6 +126,13 @@ class Requisition implements Comparable<Requisition>, Serializable {
     Date dateRejected
     Boolean approvalRequired
 
+    // Temporary until we add the original order associated with this request
+    OrderTypeCode orderTypeCode
+
+    // Fulfillment properties
+    Integer priority
```
awalkowiak: You could add constraints to match the available DeliveryTypeCodes. In this case, it would be min 1 and max 5

jmiranda: Good idea, but this is eventually going to allow for any value up to and beyond 100. We'll ignore this priority until we have time to implement the active queue. 

Note that this priority is different from the Delivery Type priority. I had imagined we'd do some type of inheritance, but it's more like two levels of priority (Service is ALWAYS higher priority than Will Call and within Service a requisition.priority of 100 is higher priority than a 10).  

There are complicated cases where they need to be mixed together, but we'll start by treating them separate.

So for now, we're going to be sorting requisitions based on the delivery type priority in descending sort order i.e. 1 is higher priority. When the active queue is being sync'd as requisitions, we'll assign a priority for each order. That means the two-level implementation will look like this 
``` 
ORDER BY delivery_type_code.priority, requisition.priority
```
Note: We can't actually do ^^^ because the delivery_type_code.priority isn't a thing. So we'll have some type of CASE statement to convert from the enum to priority inside a view or something.

druchniewicz (approved):

awalkowiak (approved):

PR #5510, opened 2025-09-24
OBPIH-7509 Add Read-only summary table

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7509

**Description:**

---
image
image
image

### DIFF
--- a/src/js/hooks/inboundV2/send/useInboundSendTable.jsx
+++ b/src/js/hooks/inboundV2/send/useInboundSendTable.jsx
@@ -75,7 +75,7 @@ const useInboundSendTable = () => {
         <TableCell
           className="rt-td-send-step"
           customTooltip
-          tooltipLabel={getValue()?.toString()}
+          tooltipLabel={getValue()}
         >
           <div className="text-truncate">
             {getValue()}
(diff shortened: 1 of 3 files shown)


src/js/hooks/inboundV2/send/useInboundSendTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+      cell: ({ getValue }) => (
+        <TableCell
+          className="rt-td-send-step"
+          customTooltip
+          tooltipLabel={getValue()?.toString()}
+        >
+          <div className="text-truncate">
+            {getValue()}
+          </div>
+        </TableCell>
```
alannadolny, line 83: toString is needed in tooltipLabel, isn't it also needed in that div?

alannadolny (approved):

src/js/hooks/inboundV2/send/useInboundSendTable.jsx
SebastianLib: I don’t think so, because in other similar components we didn’t parse the values inside divs to toString, for example in `useAllProductsTab.jsx`. In tooltipLabel I did convert it to a string, since in the PropTypes we defined it as a string

SebastianLib: You’re right since I’m parsing in tooltipLabel, it makes sense to do the same in the div, even though it would work without it. I’ve updated it so that parsing only happens when the initial value is a number.

alannadolny (approved):

PR #5511, opened 2025-09-24
OBLS-202 Sending an email when putaway discrepancy occurs

druchniewicz:


grails-app/i18n/messages.properties
```diff
@@ -703,6 +703,7 @@ email.requestRedirect.message=Click <a href="{0}">here</a> to review your reques
 email.statusChange.message=Your request status has been updated to: {0}. Click <a href="{1}">here</a> to review your requests.
 email.withComment.message=With comment
 email.fulfillment.message=Your request status has been updated to: {0}. Click <a href="{1}">here</a> to review your requests.
+email.putawayDiscrepancy.message=Putaway discrepancy occurred. Reason: {0}.
```
jmiranda: Could you add the product to the message?

jmiranda:
1. You can either use mailtrap, mailhog, or a mail service of some kind (gmail, sendgrid, mailgun) with a free option. But also, don't worry about testing SMTP at this point. You can hand it over to me when you're ready.
2. Correct. It would be good to only send an email when we are sending stock to the discrepancy location. The language around "eventing of all splits" allows us to handle the event more generically. (e.g. publish webhook event to n8n and let the n8n job take care of figuring out when to send emails). We don't need to do webhooks right now.
3. Basically the user that gets these notifications is going to want to check the Dashboard indicator that @olewandowski1 just finished . They will likely want to know why this item was moved to Lost & Found.
4. I've 
5. Do you prefer any special .gsp file for email template, I've added crazy view so far:

grails-app/conf/application.yml
```diff
@@ -117,7 +117,7 @@ grails:
         copyright: ""
         footer: ""
     mail:
-        enabled: false
+        enabled: true
```
druchniewicz: accidentally pushed

grails-app/services/org/pih/warehouse/report/NotificationService.groovy
```diff
@@ -211,6 +212,23 @@ class NotificationService {
         }
     }
 
+    def sendPutawayDiscrepancyNotification(PutawayTask putawayTask) {
+        try {
+            def emailValidator = EmailValidator.getInstance()
+            def recipients = userService.findUsersByRoleType(RoleType.ROLE_ERROR_NOTIFICATION)
```
druchniewicz: role type to be changed here

grails-app/conf/application.yml
jmiranda: Oh yeah, let's remove that when you get a chance.

PR #5513, opened 2025-09-25
OBPIH-7506 allow blank lot numbers when updating product availability

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7506

OBS tickets that this resolves:
- https://pihemr.atlassian.net/browse/OBS-1879
- https://pihemr.atlassian.net/browse/OBS-1908

**Description:** There's a detailed explanation of the issue in the OBS tickets but in brief, when an inventory snapshot refresh is triggered for an inventory item (which was happening during putaway creation), it also triggers a product availability refresh on that inventory item. The product availability refresh logic wasn't differentiating between null and blank lot numbers and so it'd try to update both to the same default lot number, causing duplicate key errors. This change simply allows product availability refresh records to contain rows for blank lots as well.

Note that we rarely experience this because this particular refresh logic is only triggered by inventory snapshot refreshes. I assume the "regular" product availability refresh logic already handles this case properly (otherwise we'd see this error everywhere)

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/ProductAvailabilityService.groovy
@@ -917,6 +917,11 @@ class ProductAvailabilityService {
     }
 
     void updateProductAvailability(InventoryItem inventoryItem) {
+        // OBPIH-7506: Only use the default lot if the lot number is null because we have valid inventory items in our
+        //             system with blank string lot numbers (ie " "). Until we can fix our importers and migrate our
+        //             data to not allow blank lots, we have to continue to allow both null and " " lots to coexist.
+        String lotNumber = inventoryItem.lotNumber == null ? Constants.DEFAULT_LOT_NUMBER : inventoryItem.lotNumber
+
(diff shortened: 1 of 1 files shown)


ewaterman:
This PR triggered a discussion about how we handle lot numbers in general in the application. I won't handle the following in this PR since it'll require a proper design/discovery, but some ideas for future improvements include:
- bonus: consider renaming the current InventoryItem entity to ProductLot and then creating a separate InventoryItem entity that has a `Location facility` field so that it represents a product lot at some facility.
- add `getOrCreateProductLot(String lotNumber, LocalDate expirationDate)` and `getProductLot(String lotNumber)` methods somewhere that standardizes how we handle lots in our system
    - should blank string lots be allowed or should we trim input? (probably trim but would need a data migration...)
    - should we treat "" as the null lot? (probably but again we'd need a data migration, which might be more trouble than is worth)
- standardizing data-binding for lots in imports. Use the above getOrCreateProductLot method!
- create a value converter for ProductLot/InventoryItem that if given a String lotNumber, LocalDate expirationDate combo will resolve the class by calling getOrCreateProductLot

PR #5517, opened 2025-09-26
OBPIH-7230 still create baseline transaction if QoH is zero

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7230

**Description:** We rely on the existance of transactions for last counted date, so even when we do a count on a product with no quantity and we count no quantity again, we still want that count action to be tracked and update last counted date.

This change will insert a baseline with quantity zero in the default lot and bin in the above scenario.

---

First I set the stock to zero with a record stock, then I did a second record stock, also setting stock to zero. You can see that we still create the second baseline in that scenario, even though there's no stock at the time
Screenshot from 2025-09-26 14-49-42

### DIFF
--- a/grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
+++ b/grails-app/services/org/pih/warehouse/data/LoadDataService.groovy
@@ -262,6 +262,7 @@ class LoadDataService {
         Map<String, Product> productMap = Product.findAllByProductCodeInList(productCodes).collectEntries {
             [it.productCode, it]
         }
+        List<Product> products = productMap.values().toList()
 
         // Calculate dates for inventory baseline and adjustment transactions
         Date adjustmentTransactionDate = DateUtil.asDate(Instant.now())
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/pih/warehouse/inventory/product/availability/AvailableItemMap.groovy
```diff
@@ -55,4 +55,8 @@ class AvailableItemMap {
     Set<Map.Entry<AvailableItemKey, AvailableItem>> entrySet() {
         return map.entrySet()
     }
+
+    boolean isEmpty() {
```
awalkowiak, line 59: Is this used?

ewaterman: I don't think so. I must have removed where I was using this.

ewaterman:
Just to be verbose, I reverted the change in https://github.com/openboxes/openboxes/pull/5526 as a part of this PR. That PR was designed to be a temporary fix that was meant only for release/0.9.5-hotfix3. This PR is the "proper" solution.

PR #5520, opened 2025-09-29
OBPIH-7510 Implement Save, Save and Exit, Send, Rollback, and Navigation Logic for Inbound Shipment Step Form

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7510

**Description:**

---

### DIFF
--- a/src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
+++ b/src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
@@ -28,14 +28,14 @@ const InboundSendForm = ({ previous }) => {
     previousPage,
     saveAndExit,
     statusCode,
-    isUserAdmin,
+    hasRoleAdmin,
     shipped,
     hasErrors,
     matchesDestination,
   } = useInboundSendForm({ previous });
 
   // Rollback button is visible only for admins when shipment has been dispatched
-  const rollbackButtonVisible = isUserAdmin && shipped;
+  const rollbackButtonVisible = hasRoleAdmin && shipped;
 
   // button is disabled when there are form errors
   // or when the shipment status is not dispatched
--- a/src/js/hooks/inboundV2/send/useInboundSendForm.js
+++ b/src/js/hooks/inboundV2/send/useInboundSendForm.js
@@ -6,7 +6,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import {
   getCurrentLocale,
   getCurrentLocation,
-  getShipmentTypes, getUserRights,
+  getShipmentTypes,
 } from 'selectors';
 
 import { fetchShipmentTypes, updateWorkflowHeader } from 'actions';
@@ -16,12 +16,14 @@ import { STOCK_MOVEMENT_URL } from 'consts/applicationUrls';
 import locationType from 'consts/locationType';
 import NotificationType from 'consts/notificationTypes';
 import requisitionStatus from 'consts/requisitionStatus';
+import RoleType from 'consts/roleType';
 import { DateFormat } from 'consts/timeFormat';
 import { OutboundWorkflowState } from 'consts/WorkflowState';
 import useInboundSendValidation from 'hooks/inboundV2/send/useInboundSendValidation';
 import useQueryParams from 'hooks/useQueryParams';
 import useSpinner from 'hooks/useSpinner';
 import useTranslate from 'hooks/useTranslate';
+import useUserHasPermissions from 'hooks/useUserHasPermissions';
 import confirmationModal from 'utils/confirmationModalUtils';
 import createInboundWorkflowHeader from 'utils/createInboundWorkflowHeader';
 import dateWithoutTimeZone from 'utils/dateUtils';
@@ -31,14 +33,14 @@ const useInboundSendForm = ({ previous }) => {
     currentLocation,
     currentLocale,
     shipmentTypes,
-    userRights,
   } = useSelector((state) => ({
     currentLocation: getCurrentLocation(state),
     currentLocale: getCurrentLocale(state),
     shipmentTypes: getShipmentTypes(state),
-    userRights: getUserRights(state),
   }));
-  const { isUserAdmin } = userRights;
+  const hasRoleAdmin = useUserHasPermissions({
+    minRequiredRole: RoleType.ROLE_ADMIN,
+  });
   const translate = useTranslate();
   const { id: stockMovementId } = useQueryParams();
   const dispatch = useDispatch();
@@ -214,7 +216,7 @@ const useInboundSendForm = ({ previous }) => {
       }
 
       await stockMovementApi.updateStatus(stockMovementId, { rollback: true });
-      window.location.reload();
+      await fetchStockMovementData();
     } finally {
       spinner.hide();
     }
@@ -360,7 +362,7 @@ const useInboundSendForm = ({ previous }) => {
     previousPage,
     saveAndExit,
     statusCode,
-    isUserAdmin,
+    hasRoleAdmin,
     shipped,
     hasErrors,
     matchesDestination,
--- a/src/js/selectors/index.js
+++ b/src/js/selectors/index.js
@@ -30,11 +30,3 @@ export const getCurrentUser = (state) => state.session.user;
 export const getCycleCountTranslations = (state) => state.session.fetchedTranslations.cycleCount;
 
 export const getShipmentTypes = (state) => state.stockMovementCommon.shipmentTypes;
-
-export const getUserRights = (state) => ({
-  isUserAdmin: state.session.isUserAdmin,
-  isUserApprover: state.session.isUserApprover,
-  isUserManager: state.session.isUserManager,
-  isUserRequestApprover: state.session.isUserRequestApprover,
-  isSuperuser: state.session.isSuperuser,
-});


src/js/api/services/StockMovementApi.js
```diff
@@ -39,4 +40,7 @@ export default {
     apiClient.get(STOCK_MOVEMENT_BY_ID(id), { params }),
   getStockMovementItems: (id, params) =>
     apiClient.get(STOCK_MOVEMENT_ITEMS(id), { params }),
+  updateStockMovementShipment: (id, payload) => apiClient.post(
```
kchelstowski: `updateShipment` should be descriptive enough

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -185,14 +199,29 @@ const InboundSendForm = ({ previous }) => {
...
-          <Button
-            label="shipping.sendShipment.label"
-            defaultLabel="Send shipment"
-            variant="primary"
-            type="submit"
+            onClick={previousPage}
+            disabled={statusCode === requisitionStatus.DISPATCHED}
           />
+          <div className="buttons-container">
+            {(hasAdminRights && shipped) && (
```
kchelstowski: `hasAdminRights` -> `isUserAdmin`

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -185,14 +199,29 @@ const InboundSendForm = ({ previous }) => {
...
+              <Button
+                label="react.default.button.rollback.label"
+                defaultLabel="Rollback"
+                variant="primary-outline"
+                onClick={rollbackStockMovement}
```
kchelstowski: to be safer I will always suggest making it `onClick={() => rollbackStockMovement()}`

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -185,14 +199,29 @@ const InboundSendForm = ({ previous }) => {
...
+                StartIcon={<RiArrowGoBackFill className="icon" />}
+                disabled={Object.keys(errors).length > 0
+                  || statusCode !== requisitionStatus.DISPATCHED}
```
kchelstowski: make it a separate const with name e.g. `rollbackButtonVisible` with a short comment when it is disabled

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -185,14 +199,29 @@ const InboundSendForm = ({ previous }) => {
...
+              />
+            )}
+            <Button
+              label="shipping.sendShipment.label"
```
kchelstowski: `react` prefix probably missing

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
```diff
@@ -77,14 +84,14 @@ const InboundSendFormHeader = ({ errors }) => (
         </div>
       </div>
       <Button
-        onClick={() => console.log('click')}
+        onClick={saveAndExit}
```
kchelstowski: same comment as in rollback

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
```diff
@@ -77,14 +84,14 @@ const InboundSendFormHeader = ({ errors }) => (
...
         StartIcon={<RiPictureInPictureExitLine className="icon" />}
         defaultLabel="Save and exit"
         label="react.default.button.saveAndExit.label"
         variant="primary-outline"
       />
       <Button
-        onClick={() => console.log('click')}
+        onClick={onSave}
```
kchelstowski: same comment as in rollback

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -4,33 +4,44 @@ import { zodResolver } from '@hookform/resolvers/zod';
...
   const {
     currentLocation,
     currentLocale,
     shipmentTypes,
+    hasAdminRights,
   } = useSelector((state) => ({
     currentLocation: getCurrentLocation(state),
     currentLocale: getCurrentLocale(state),
     shipmentTypes: getShipmentTypes(state),
+    hasAdminRights: getAdminStatus(state),
```
kchelstowski: isUserAdmin

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -48,26 +59,55 @@ const useInboundSendForm = () => {
...
+      comments,
+      expectedDeliveryDate,
+    } = getValues();
+
+    return {
+      dateShipped: dateWithoutTimeZone({
+        date: shipDate,
+        outputDateFormat: DateFormat.MM_DD_YYYY,
+      }),
+      shipmentType: shipmentType.id || null,
```
kchelstowski: don't you need nullsafes here? If `shipmentType` is `null` you would get a nullpointer instead of falling back to `|| null` 
Moreover if a shipmentType can't be an empty string you should probably use `??` instead of `||`

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -48,26 +59,55 @@ const useInboundSendForm = () => {
...
+      trackingNumber: trackingNumber || '',
+      driverName: driverName || '',
+      comments: comments || '',
```
kchelstowski: same comment as above regarding `??` vs `||`

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -84,18 +124,19 @@ const useInboundSendForm = () => {
...
-            name: data.shipmentType.name,
-            label: data.shipmentType.displayName,
-            value: data.shipmentType.id,
-          }
-          : null,
+        shipmentType: {
+          id: data.shipmentType.id,
+          name: data.shipmentType.name,
+          label: data.shipmentType.displayName,
+          value: data.shipmentType.id,
```
kchelstowski: same question/comment about the nullsafes

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,12 +150,11 @@ const useInboundSendForm = () => {
     }
   };
 
-  const onSubmit = async (values) => {
-    console.log(values);
-
+  const validateAndBuildShipmentPayload = (values) => {
```
kchelstowski: I haven't checked the previous PRs, but one question that comes to my mind here is that why do we need a separate validation method and call it separately instead of handling validation via Zod?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -123,7 +163,7 @@ const useInboundSendForm = () => {
           'You are not able to send shipment from a location other than origin. Change your current location.',
         ),
       });
-      return;
+      return null;
```
kchelstowski: if it turns out that this method is necessary after answering to my question, I think validation methods should return true/false for consistency, so eventually split this method to validate + build payload methods

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -137,26 +177,196 @@ const useInboundSendForm = () => {
...
+        },
+      ],
+    });
+  };
+
+  // Saves changes made by user in this step and redirects to the shipment view page
+  const saveAndExit = async () => {
+    await trigger();
+
+    if (Object.keys(errors).length === 0) {
```
kchelstowski: I can see you repeating this many times. Maybe create a separate const `hasErrors`/`errorsEmpty` or something like this (check if we use it somewhere and eventually adjust the naming)

src/js/selectors/index.js
```diff
@@ -30,3 +30,5 @@ export const getCurrentUser = (state) => state.session.user;
 export const getCycleCountTranslations = (state) => state.session.fetchedTranslations.cycleCount;
 
 export const getShipmentTypes = (state) => state.stockMovementCommon.shipmentTypes;
+
+export const getAdminStatus = (state) => state.session.isUserAdmin;
```
kchelstowski: `isUserAdmin`

kchelstowski (changes requested):

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: We need a separate validation here because it’s only used when sending a shipment. For example, when just saving, we don’t use validateAndBuildShipmentPayload, so it seems better to keep it separate from Zod. But now I started wondering, if we only perform this validation on send shipment, maybe we should apply it everywhere. For now, I’ve mirrored the behavior from the old inbound, but I’ll talk with Manon if we also want this validation on save, previous buttons, etc. If yes, I will handle this via Zod

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
```diff
@@ -1,111 +1,139 @@
 import React from 'react';
 
 import PropTypes from 'prop-types';
+import { IoMdExit } from 'react-icons/io';
```
alannadolny, line 4: Shouldn't we use only ri icons?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -1,40 +1,56 @@
...
 
 import { zodResolver } from '@hookform/resolvers/zod';
-import _ from 'lodash';
 import { useForm } from 'react-hook-form';
 import { useDispatch, useSelector } from 'react-redux';
-import { getCurrentLocale, getCurrentLocation, getShipmentTypes } from 'selectors';
+import {
+  getCurrentLocale,
+  getCurrentLocation,
+  getShipmentTypes, getUserRights,
```
alannadolny:
```suggestion
  getShipmentTypes,
  getUserRights,
```

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -1,40 +1,56 @@
...
     shipmentTypes,
+    userRights,
   } = useSelector((state) => ({
     currentLocation: getCurrentLocation(state),
     currentLocale: getCurrentLocale(state),
     shipmentTypes: getShipmentTypes(state),
+    userRights: getUserRights(state),
   }));
-
+  const { isUserAdmin } = userRights;
```
alannadolny: Don't we have a hook for getting that information? I don't like the fact that you had to pull it from the state and destructurize the object directly in the component.

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,54 +169,201 @@ const useInboundSendForm = () => {
...
+          message: 'Error',
+          details: translate(
+            'react.stockMovement.alert.sendStockMovement.label',
+            'You are not able to send shipment from a location other than origin. Change your current location.',
+          ),
+        });
+        return;
+      }
+      await stockMovementApi.updateShipment(stockMovementId, getShipmentPayload());
+      await stockMovementApi.updateStatus(stockMovementId,
```
alannadolny, line 189: I don't remember - can't we update the status through updateShipment?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,54 +169,201 @@ const useInboundSendForm = () => {
...
+        { status: requisitionStatus.DISPATCHED });
+      window.location = STOCK_MOVEMENT_URL.show(stockMovementId);
```
alannadolny, line 191: Can't we do it in a more "react" way? We shouldn't override the location by hand

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,54 +169,201 @@ const useInboundSendForm = () => {
...
+          details: translate(
+            'react.stockMovement.alert.rollbackShipment.label',
+            'You are not able to rollback shipment from your location.',
+          ),
+        });
+        return;
+      }
+
+      await stockMovementApi.updateStatus(stockMovementId, { rollback: true });
+      window.location.reload();
```
alannadolny: don't use that

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,54 +169,201 @@ const useInboundSendForm = () => {
...
+    });
+  };
+
+  // Saves changes made by user in this step and redirects to the shipment view page
+  const saveAndExit = async () => {
+    await trigger();
+
+    if (!hasErrors) {
+      await stockMovementApi.updateShipment(stockMovementId, getShipmentPayload());
+      window.location = STOCK_MOVEMENT_URL.show(stockMovementId);
```
alannadolny, line 331: the same here

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -109,54 +169,201 @@ const useInboundSendForm = () => {
...
+      return;
+    }
+
+    confirmActionModal({
+      messageId: 'react.stockMovement.confirmExit.message',
+      messageDefault: 'Validation errors occurred. Are you sure you want to exit and lose unsaved data?',
+      onConfirm: () => {
+        window.location = STOCK_MOVEMENT_URL.show(stockMovementId);
```
alannadolny, line 339: the same here

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
SebastianLib: the problem is that the Ri exit icons are ugly, I talked to manon about it and it's ok to use other ones

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: The problem is that when I was using the function from React Router, it didn’t work correctly when we needed to navigate someone to a view, so this seems to be the only option

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: It's not possible at the moment, it could probably be done if we made changes to the backend, but it doesn't make sense since we already have a ready endpoint for this

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: Oh, I actually didn't even know we had something like that. Thanks for the information

src/js/hooks/inboundV2/send/useInboundSendForm.js
kchelstowski: It's a GSP page so it has to be done like that (unfortunately)

alannadolny: image

alannadolny: If it's needed for switching between GSP, it's ok. However, the other cases are still weird to me.

alannadolny (approved):

Issue #5392, opened 2025-07-18
Error : Inventory Snapshot


NatipongSupatsorn:
URI
/openboxes/product/createProductSnapshot/2c938084975398ce01975751713a00ee
Class
groovy.lang.MissingPropertyException
Message
No such property: comments for class: org.pih.warehouse.inventory.BinLocationItem

def recordStock(Location location, List lineItems) {

        if (!location || !location.inventory) {
            throw new IllegalArgumentException("Record stock transactions require a location")
        }

        Transaction transaction = new Transaction()
        transaction.transactionDate = new Date()
        transaction.inventory = location.inventory
        transaction.transactionNumber = generateTransactionNumber(transaction)
        transaction.transactionType = TransactionType.get(Constants.PRODUCT_INVENTORY_TRANSACTION_TYPE_ID)

        lineItems.each { lineItem ->
            // Add transaction entry to transaction
            TransactionEntry transactionEntry = new TransactionEntry()
            transactionEntry.binLocation = lineItem.binLocation
            transactionEntry.inventoryItem = lineItem.inventoryItem
            transactionEntry.quantity = lineItem.quantity
            transactionEntry.comments = lineItem.comments <----------------------------- Please consider this line. Thank you.
            transaction.addToTransactionEntries(transactionEntry)
        }

        if (!transaction.hasErrors() && transaction.save()) {
            log.info("Transaction saved: " + transaction)
        }

        return transaction
    }

PR #5269, opened 2025-05-21
OBPIH-7203 change adjustment icon in stock history and remove migation buttons o…

EWaterman:
…n product page

### :sparkles: Description of Change

[//]: <> (A concise summary of what is being changed. Please provide enough context for reviewers to be able to understand the change and why it is necessary.)

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7203

**Description:**

---
### :camera: Screenshots & Recordings (optional)

[//]: <> (If this PR contains a UI change, consider attaching one or more screenshots or recordings to help reviewers visualize the change. Otherwise, you can remove this section.)

Screenshot from 2025-05-21 13-34-50

PR #5527, opened 2025-10-01
5392 remove unused product snapshot controller

ewaterman:

**Link to GitHub issue or Jira ticket:** https://github.com/openboxes/openboxes/issues/5392

**Description:** We removed the "product snapshot" feature (not to be confused with the inventory snapshot feature) from the stock card in https://github.com/openboxes/openboxes/pull/5269 (behind the scenes this feature did a record stock containing the current stock for the product, so essentially it did what we have now for baselines but using the product inventory transaction type). This PR is just removing the extra code that I forgot to also remove as a part of that original PR.

Note that you can still achieve this "snapshot" functionality by simply doing a record stock and clicking save without making any changes.

### DIFF
--- a/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
+++ b/grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
@@ -1948,51 +1948,6 @@ class InventoryService implements ApplicationContextAware {
         return command
     }
 
-
-    def createStockSnapshot(Location location, Product product) {
-        List lineItems = getProductQuantityByBinLocation(location, product)
-
-        // If there's no stock we should record that as well
-        if (!lineItems || lineItems?.empty) {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/pih/warehouse/inventory/InventoryService.groovy
```diff
@@ -1948,51 +1948,6 @@ class InventoryService implements ApplicationContextAware {
         return command
     }
 
-
-    def createStockSnapshot(Location location, Product product) {
```
awalkowiak, line 1967: I think this is still used in the MigrationService, so you'd need to get rid of it there or replace with the inventory baseline

ewaterman: how did I miss this? Thanks for finding it

PR #5528, opened 2025-10-02
OBLS-181 Removing child order items during putaway rollback

druchniewicz:


jmiranda (approved):
Looks good to me. Just make sure we're resetting all of the right fields on the original item.

awalkowiak (approved):

PR #5529, opened 2025-10-03
OBPIH-7511 Implement document download/upload actions

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7511

**Description:**

---

### DIFF
--- a/src/js/hooks/inboundV2/send/useInboundSendForm.js
+++ b/src/js/hooks/inboundV2/send/useInboundSendForm.js
@@ -19,8 +19,8 @@ import requisitionStatus from 'consts/requisitionStatus';
 import RoleType from 'consts/roleType';
 import { DateFormat } from 'consts/timeFormat';
 import { OutboundWorkflowState } from 'consts/WorkflowState';
-import useInboundFileActions from 'hooks/inboundV2/send/useInboundFileActions';
 import useInboundSendValidation from 'hooks/inboundV2/send/useInboundSendValidation';
+import useFileActions from 'hooks/useFileActions';
 import useQueryParams from 'hooks/useQueryParams';
 import useSpinner from 'hooks/useSpinner';
 import useTranslate from 'hooks/useTranslate';
@@ -28,7 +28,7 @@ import useUserHasPermissions from 'hooks/useUserHasPermissions';
 import confirmationModal from 'utils/confirmationModalUtils';
 import createInboundWorkflowHeader from 'utils/createInboundWorkflowHeader';
 import dateWithoutTimeZone from 'utils/dateUtils';
-import filterDocumentsByStepNumber from 'utils/filterDocumentsByStepNumber';
+import filterDocumentsByStepNumber from 'utils/stockMovementUtils';
 
 const useInboundSendForm = ({ previous }) => {
   const {
@@ -215,11 +215,11 @@ const useInboundSendForm = ({ previous }) => {
     handleRemoveFile,
     sendFiles,
     handleExportFile,
-  } = useInboundFileActions({
-    stockMovementId,
+  } = useFileActions({
+    entityId: stockMovementId,
     onSave,
-    spinner,
     isValid,
+    uploadDocuments: stockMovementApi.uploadDocuments,
   });
 
   // sends the whole stock movement
--- a/src/js/utils/filterDocumentsByStepNumber.js
+++ b/src/js/utils/filterDocumentsByStepNumber.js
@@ -1,4 +1,0 @@
-const filterDocumentsByStepNumber = (documents, stepNumber) =>
-  documents.filter((document) => document.stepNumber === stepNumber);
-
-export default filterDocumentsByStepNumber;
--- a/grails-app/i18n/messages.properties
+++ b/grails-app/i18n/messages.properties
@@ -3362,6 +3362,8 @@ react.default.enum.TransactionAction.CYCLE_COUNT=CC
 react.default.enum.TransactionAction.RECORD_STOCK=RS
 react.default.enum.TransactionAction.INVENTORY_IMPORT=II
 react.default.enum.TransactionAction.INVENTORY_ADJUSTMENT=IA
+react.default.alert.fileSuccess.label=File uploaded successfully!
+react.default.alert.filesSuccess.label=Files uploaded successfully!
 
 # Date formats
 react.default.defaultDate.format=MM/DD/YYYY HH:mm:ss Z
--- a/src/js/components/stock-movement-wizard/inboundV2/inboundV2.scss
+++ b/src/js/components/stock-movement-wizard/inboundV2/inboundV2.scss
@@ -4,7 +4,7 @@
 }
 
 .inbound-wrapper {
-  min-height: 800px;
+  min-height: 900px;
 }
 
 .inbound-add-items {
@@ -108,7 +108,7 @@ input.hide-arrows[type=number] {
 }
 
 .rt-tbody-v2 {
-  max-height: 400px;
+  max-height: 500px;
   overflow: auto;
 
   .input-xs {
--- a/src/js/hooks/useFileActions.js
+++ b/src/js/hooks/useFileActions.js
@@ -2,21 +2,32 @@ import { useState } from 'react';
 
 import _ from 'lodash';
 
-import stockMovementApi from 'api/services/StockMovementApi';
 import notification from 'components/Layout/notifications/notification';
 import NotificationType from 'consts/notificationTypes';
+import useSpinner from 'hooks/useSpinner';
 import useTranslate from 'hooks/useTranslate';
 import useWindowOpen from 'hooks/useWindowOpen';
 
-const useInboundFileActions = ({
-  stockMovementId,
+/**
+ * Custom hook that manages file upload, download, removal and export actions.
+ *
+ * @param {string} entityId - ID of the entity the files belong to.
+ * @param {Function} onSave - Function called before exporting a document.
+ * @param {boolean} isValid - React Hook Form's formState.isValid flag which determines
+ * if the form is valid before export.
+ * @param {Function} uploadDocuments - Async function for uploading documents.
+ *
+ */
+const useFileActions = ({
+  entityId,
   onSave,
-  spinner,
   isValid,
+  uploadDocuments,
 }) => {
-  // List of user selected File objects (documents to attach to the current stock movement).
+  // List of user selected File objects.
   // These files are stored locally until uploaded via 'sendFiles()'.
   const [files, setFiles] = useState([]);
+  const spinner = useSpinner();
   const translate = useTranslate();
   const { openWindow } = useWindowOpen();
 
@@ -35,11 +46,11 @@ const useInboundFileActions = ({
     files.forEach((file, idx) => {
       data.append(`filesContents[${idx}]`, file);
     });
-    await stockMovementApi.uploadDocuments(stockMovementId, data);
+    await uploadDocuments(entityId, data);
     if (files.length > 1) {
       notification(NotificationType.SUCCESS)({
         message: translate(
-          'react.stockMovement.alert.filesSuccess.label',
+          'react.default.alert.filesSuccess.label',
           'Files uploaded successfully!',
         ),
       });
@@ -47,7 +58,7 @@ const useInboundFileActions = ({
     }
     notification(NotificationType.SUCCESS)({
       message: translate(
-        'react.stockMovement.alert.fileSuccess.label',
+        'react.default.alert.fileSuccess.label',
         'File uploaded successfully!',
       ),
     });
@@ -77,4 +88,4 @@ const useInboundFileActions = ({
   };
 };
 
-export default useInboundFileActions;
+export default useFileActions;
--- a/src/js/utils/stockMovementUtils.js
+++ b/src/js/utils/stockMovementUtils.js
@@ -1,0 +1,10 @@
+/**
+ * Filters documents by a specific step number
+ * @param {Array} documents - Array of document objects to filter
+ * @param {number} stepNumber - Step number to match against document.stepNumber
+ * @returns {Array} Filtered array of documents matching the given step number
+ */
+const filterDocumentsByStepNumber = (documents, stepNumber) =>
+  documents.filter((document) => document.stepNumber === stepNumber);
+
+export default filterDocumentsByStepNumber;


src/js/components/form-elements/v2/DropzoneFileSelect.jsx
```diff
@@ -78,6 +80,22 @@ const DropzoneFileSelect = ({
...
+          StartIcon={<RiUpload2Line className="icon" />}
+          onClick={open}
+          variant={buttonVariant}
+          defaultLabel={buttonLabel.defaultMessage}
+          label={buttonLabel.id}
+        />
+      </div>
+    );
+  }
+
```
alannadolny, line 98: Is it for displaying a button instead of a dropzone? I think we have something similar currently, and it's working without that change

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
```diff
@@ -38,60 +45,62 @@ const InboundSendFormHeader = ({
...
-                defaultMessage="Certificate of Donation"
-              />
-            </a>
+            {documents?.map((document) => (
+              <a
+                key={document?.uri}
+                href="#"
+                className="dropdown-item"
+                onClick={(e) => {
+                  e.preventDefault();
```
alannadolny, line 96: Why is it needed?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -158,6 +166,7 @@ const useInboundSendForm = ({ previous }) => {
         statusCode: data.statusCode ?? '',
         hasManageInventory: data.hasManageInventory ?? false,
         shipped: data.shipped ?? false,
+        documents: _.filter(data.associations.documents, (document) => document.stepNumber === 5),
```
alannadolny: Can we have documents with different step numbers? If yes, maybe it's worth creating a util

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -171,6 +180,31 @@ const useInboundSendForm = ({ previous }) => {
...
+      });
+      return;
+    }
+    notification(NotificationType.SUCCESS)({
+      message: translate(
+        'react.stockMovement.alert.fileSuccess.label',
+        'File uploaded successfully!',
+      ),
+    });
+  };
```
alannadolny: What if uploadDocuments returns a fake 200? Then you will show information that the files are uploaded successfully, even if they are not.

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -223,9 +262,9 @@ const useInboundSendForm = ({ previous }) => {
   };
 
   // save the whole stock movement
-  const onSave = async () => {
-    await trigger();
-    if (hasErrors) {
+  const onSave = async ({ showNotification = true }) => {
+    const isValid = await trigger();
```
alannadolny: Does the trigger function return all validation errors? Are you sure it's working correctly? Maybe add a comment - it's weird that the trigger function returns errors, so it will be good to add an explanation

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -341,6 +382,24 @@ const useInboundSendForm = ({ previous }) => {
     });
   };
 
+  const handleExportFile = async (doc) => {
+    const isValid = await trigger();
+    if (!isValid || !doc?.uri) {
+      return;
+    }
+    await onSave({ showNotification: false });
+    window.open(doc.uri, '_blank');
```
alannadolny: Isn't there any more "react" way of doing that? - just out of curiosity. If not, maybe pack the logic into a hook for handling the window object operations?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -341,6 +382,24 @@ const useInboundSendForm = ({ previous }) => {
...
+  };
+
+  const handleDownloadFiles = (newFiles) => {
+    setFiles((prevFiles) =>
+      _.unionBy([...prevFiles, ...newFiles], 'name'));
```
alannadolny: I don't understand that :smiling_face_with_tear:

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -341,6 +382,24 @@ const useInboundSendForm = ({ previous }) => {
...
+  };
+
+  const handleRemoveFile = (fileToRemove) => {
+    setFiles((prevFiles) => prevFiles.filter((f) => f.name !== fileToRemove.name));
```
alannadolny: f -> file

src/js/components/form-elements/v2/DropzoneFileSelect.jsx
SebastianLib: I think you mean `ButtonFileSelect.jsx` but there we can only select a file and there’s no drag & drop option. In `DropzoneFileSelect.jsx` we use library `react-dropzone`, which allows drag & drop, so even when we display just the button, drag & drop still works in the background. In the old inbound, drag & drop works as well, so here I wanted to do something similar.

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendFormHeader.jsx
SebastianLib: Thanks to` e.preventDefault()`, clicking the link doesn’t add "#" to the URL

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: yes we can have documents with different steps

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: After our private conversation on Slack, we decided that it can stay as it is.

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: trigger() returns true/false, not the full errors. formState.errors updates on the next render, so right after await trigger() it's safer to check the returned boolean instead of hasErrors.

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: It merges old and new files into one list, and makes sure there are no duplicates by name. If a file with the same name already exists, the old one is kept and the new one is ignored. I’ll probably change the logic soon so that the new file replaces the old one instead

src/js/hooks/inboundV2/send/useInboundSendForm.js
kchelstowski: out of curiosity - why lodash for a simple filter?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -29,6 +30,7 @@ import createInboundWorkflowHeader from 'utils/createInboundWorkflowHeader';
 import dateWithoutTimeZone from 'utils/dateUtils';
 
 const useInboundSendForm = ({ previous }) => {
+  const [files, setFiles] = useState([]);
```
kchelstowski: what do we mean by "files"? Is it possible to be more descriptive here?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -366,6 +425,11 @@ const useInboundSendForm = ({ previous }) => {
     shipped,
     hasErrors,
     matchesDestination,
+    documents,
+    handleExportFile,
+    handleDownloadFiles,
+    files,
+    handleRemoveFile,
```
kchelstowski, line 407: I would suggest creating a separate hook for export/import/download actions. This hooks is already heavy enough that it's gonna be difficult to read.

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -171,6 +180,31 @@ const useInboundSendForm = ({ previous }) => {
...
   };
 
+  const sendFiles = async () => {
+    const data = new FormData();
+
+    files.forEach((file, idx) => {
+      data.append(`filesContents[${idx}]`, file);
+    });
+
+    await stockMovementApi.uploadDocuments(stockMovementId, data);
```
kchelstowski: don't call async action without wrapping it why try/catch and adding a spinner

src/js/hooks/inboundV2/send/useInboundSendForm.js
kchelstowski: this should be wrapped in the try/finally also

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -171,6 +180,31 @@ const useInboundSendForm = ({ previous }) => {
...
+    if (files.length > 1) {
```
kchelstowski: what does the API return? isn't it possible to somehow check the success of the action based on the API response? Have you checked it?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -223,9 +262,9 @@ const useInboundSendForm = ({ previous }) => {
   };
 
   // save the whole stock movement
-  const onSave = async () => {
-    await trigger();
-    if (hasErrors) {
+  const onSave = async ({ showNotification = true }) => {
```
kchelstowski: what "notification" By reading just `showNotification` I can't figure out what notification we mean here

src/js/hooks/inboundV2/send/useInboundSendForm.js
kchelstowski: I don't understand why it is needed. Why would we need to trigger validation again if we should rely on Zod + React Hook Form to handle that.
I think the flow should be having the form wrapped with the methods from `useForm` hook like `control` etc, and when triggering the `onSave` method we should have up to date errors etc.

kchelstowski (changes requested):

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: This function is only used inside the `sendShipment` function, which is already wrapped in a try/finally, so it didn’t make sense to do it here as well

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: I noticed that you often like to use Lodash, so I thought it would be okay to use it in this function, but I’ve changed it now

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: This function is only used inside the sendShipment function, which is already wrapped in a try/finally, so it didn’t make sense to do it here as well

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: As I mentioned earlier, this function is only used inside the sendShipment function, which is already wrapped in a try/catch, so it didn’t make sense to do it here as well. This way, if an error occurs, it will still throw and the function execution will stop. Maybe it would actually be better to wrap it in a try/catch too, just to avoid any confusion? :thinking:

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: showNotification is a boolean that determines whether we want to display and trigger the notification function. For example:       
```
      if (showNotification) {
        notification(NotificationType.SUCCESS)({
          message: translate(
            'react.stockMovement.alert.saveSuccess.label',
            'Changes saved successfully',
          ),
        });
      }
```

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: The issue here is that validation only runs automatically on buttons with `type="submit"`. Our form already has `onSubmit={handleSubmit(sendShipment)}`, so if we made all buttons submit buttons, the same submit function would always run with validation, which we don’t want. Validation still runs normally on inputs etc. For any button that isn’t a submit, the best approach is probably to call `await trigger()` manually to ensure validation runs

SebastianLib: I think that trigger() is useful when a button is not of type submit, for example a "Previous" or "Save" button. If the user hasn’t interacted with the form yet, validation errors won’t appear automatically. Calling trigger() in this case forces the errors to show, allowing the user to see them

Validation still works normally and updates in real-time as the user interacts with the fields. trigger() is mainly helpful when the user hasn’t made any changes yet

src/js/hooks/inboundV2/send/useInboundFileActions.js
```diff
@@ -0,0 +1,84 @@
...
+  const sendFiles = async () => {
+    const data = new FormData();
+
+    files.forEach((file, idx) => {
+      data.append(`filesContents[${idx}]`, file);
+    });
+    try {
+      spinner.show();
+      await stockMovementApi.uploadDocuments(stockMovementId, data);
+      if (files.length > 1) {
```
SebastianLib: After your comments, I added a try/catch block here to improve readability, although I think it's not really necessary at the moment since this function is only used inside sendShipment, which is already wrapped in a try/catch. That should be sufficient for now, but in the future we might use this function elsewhere, and in that case having its own try/catch could prevent potential issues.

src/js/hooks/inboundV2/send/useInboundFileActions.js
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  const sendFiles = async () => {
+    const data = new FormData();
+
+    files.forEach((file, idx) => {
+      data.append(`filesContents[${idx}]`, file);
+    });
+    try {
+      spinner.show();
```
kchelstowski: if you said there was already a try/finally in a method "higher" I think there is no need to add an additional spinner, unless there is no spinner in the method "higher"?

src/js/hooks/inboundV2/send/useInboundFileActions.js
```diff
@@ -0,0 +1,84 @@
...
+        return;
+      }
+      notification(NotificationType.SUCCESS)({
+        message: translate(
+          'react.stockMovement.alert.fileSuccess.label',
+          'File uploaded successfully!',
+        ),
+      });
+    } finally {
+      spinner.hide();
```
kchelstowski: if the method "higher" calls any other methods, hiding the spinner at this step feels invalid, as you would hide the spinner when some other methods might be still called in the background

src/js/hooks/inboundV2/send/useInboundFileActions.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+  };
+
+  const handleExportFile = async (document) => {
+    if (!isValid || !document?.uri) {
+      return;
+    }
+    try {
+      spinner.show();
+      await onSave({ showNotification: false });
```
kchelstowski: I will repeat my question - what notification we mean here? Is it a success notification? If so, change the name to `showSuccessNotification`

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -171,6 +179,48 @@ const useInboundSendForm = ({ previous }) => {
     }
   };
 
+  /**
+   * Saves the whole stock movement
```
kchelstowski, line 183: I appreciate the comment here

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -171,6 +179,48 @@ const useInboundSendForm = ({ previous }) => {
...
+   */
+  const onSave = async ({ showNotification = true }) => {
+    if (!isValid) {
+      return;
+    }
+    try {
+      spinner.show();
+      await stockMovementApi.updateShipment(stockMovementId, getShipmentPayload());
+
+      if (statusCode === requisitionStatus.DISPATCHED) {
```
kchelstowski, line 195: what's the flow behind it? where is `statusCode` taken from? Are we sure it's up to date if you call update shipment above? Shouldn't we keep the response from 192th in a const and take the current values from it?

src/js/hooks/useWindowOpen.js
```diff
@@ -0,0 +1,10 @@
+const useWindowOpen = () => {
```
kchelstowski, line 5: what is this hook responsible for? Could you add a comment explaining the usage of it?

src/js/utils/filterDocumentsByStepNumber.js
```diff
@@ -0,0 +1,4 @@
+const filterDocumentsByStepNumber = (documents, stepNumber) =>
+  documents.filter((document) => document.stepNumber === stepNumber);
+
+export default filterDocumentsByStepNumber;
```
kchelstowski: Does this method need a separate file?

src/js/hooks/inboundV2/send/useInboundFileActions.js
SebastianLib: I think when you wrote that comment, you hadn’t seen it yet, but I added a note in the `onSave` function explaining this boolean:
`// @param {boolean} showNotification - Determines whether to show a success toast after saving.`

I’ll rename it to showSuccessNotification as you suggested, since it currently only shows a success notification.

src/js/hooks/inboundV2/send/useInboundSendForm.js
SebastianLib: We only change this status when sending a shipment in the **sendShipment** function. The status is updated here:
`await stockMovementApi.updateStatus(stockMovementId, { status: requisitionStatus.DISPATCHED });`
So during this save, the status doesn't change.

src/js/utils/filterDocumentsByStepNumber.js
SebastianLib: I don’t have a strong opinion here since it’s a fairly simple function, but @alannadolny suggested making it a separate util because we can have documents with different step numbers, so it might be useful to have a reusable function

kchelstowski: I'll let @alannadolny have the last comment here.
My point is that even if it's good to have it reusable, it should be placed maybe in an existing util file. It feels weird to me to have a util file with the same name like a method inside it. I would see it in the same way like we have `cycleCountUtils` so maybe to have `inboundFormUtils`? something like that, but don't make any changes until the :gorilla: that suggested that (@alannadolny :laughing: ) makes the call.

SebastianLib: I think stockMovementUtils would be a better name, since inbound and outbound share similar logic

PR #5533, opened 2025-10-06
OBLS-249 Added 'Rollback and Delete' buttons on putaway tasks and stock movement pages

druchniewicz:


grails-app/conf/application.yml
```diff
@@ -624,8 +624,12 @@ openboxes:
         receivingLocation:
             prefix: "R"
         showAutoReceiveButton: true
-        inboundSortation:
-            enabled: true
+
+    stockMovement:
```
awalkowiak, line 629: Change these `showRollbackAndDeleteButton` into:
```yml
    stockMovement:
        rollbackAndDelete:
            enabled: true
    putaway:
        rollbackAndDelete:
            enabled: true
```
and put a check for it inside the action in the controller.

Additionally, let's rename the `showAutoReceiveButton` into:
```yaml
        manualAutoReceiving:
            enabled: true
```

awalkowiak (changes requested):

grails-app/conf/application.yml
druchniewicz: I've pushed fix with it.

But now I noticed that Rollback and Delete sometimes removes whole shipment successfully and sometimes not. Lately I didn't see this issue, today I'm getting Operation failed: null 
```2025-10-22 17:01:11.416 [http-nio-8080-exec-10] ERROR o.p.w.i.StockMovementController - Failed to rollback and delete stock movement with ID ff8081819a0b8d2e019a0b9f02b80010: null
java.lang.NullPointerException: null
	at org.grails.orm.hibernate.support.ClosureEventTriggeringInterceptor.getModifiedPropertiesWithAutotimestamp(ClosureEventTriggeringInterceptor.java:206)
	at org.grails.orm.hibernate.support.ClosureEventTriggeringInterceptor.synchronizeHibernateState(ClosureEventTriggeringInterceptor.java:188)
	at org.grails.orm.hibernate.support.ClosureEventTriggeringInterceptor.onPreUpdate(ClosureEventTriggeringInterceptor.java:249)
	at org.hibernate.action.internal.EntityUpdateAction.preUpdate(EntityUpdateAction.java:244)
```

with no special reason.

druchniewicz: @awalkowiak when it comes to removing whole Shipment maybe will be better to create a new method that removes everything related to shipment one by one. So far I have been relying on existing methods that rollbacks one thing, rollbacks another, deletes one thing, deletes another - flow similar when you want to delete shipment via UI - you have to go through rollback processes and at the end you are able to remove shipment. Combining it together I see still some weird issues (like sometimes it works, sometimes not) and I'm thinking about something 'more readable' like remove all receipt and receipt items, remove all transactions, remove all events, remove requisition and requisition items, remove shipment items and remove shipment~
Another benefit I can see that I wouldn't touch any existing method and avoid the risk of unknowingly breaking something somewhere else. 
What do you think?

druchniewicz: Fixed it in [PR](https://github.com/openboxes/openboxes/pull/5569)

PR #5534, opened 2025-10-06
OBPIH-7505 refactor cycle count dates

ewaterman:

**Link to GitHub issue or Jira ticket:** https://pihemr.atlassian.net/browse/OBPIH-7505

**Description:** Refactors cycle count to use the java.time classes instead of java.util.Date.


grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
```diff
@@ -11,13 +13,13 @@ class CycleCount {
 
     Location facility
 
-    Date dateLastRefreshed
+    Instant dateLastRefreshed
```
kchelstowski, line 16: out of curiosity - doesn't it require any DB migration? Is it supposed to say `DATETIME` there? (I'm sorry if you've explained it somewhere)

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
```diff
@@ -30,15 +32,15 @@ class CycleCountCandidate {
 
     Integer negativeItemCount
 
-    Date dateLastCount
+    LocalDate dateLastCount
```
kchelstowski, line 35: why one is Instant and another is LocalDate?

grails-app/domain/org/pih/warehouse/inventory/CycleCountItem.groovy
```diff
@@ -37,11 +40,11 @@ class CycleCountItem implements Comparable {
 
     Boolean custom
 
-    Date dateCounted
+    LocalDate dateCounted
```
kchelstowski, line 43: same question - why is this one `LocalDate` and others below are `Instant`?

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
```diff
@@ -36,6 +35,9 @@ class CycleCountService {
     CycleCountTransactionService cycleCountTransactionService
     CycleCountProductAvailabilityService cycleCountProductAvailabilityService
 
+    @Autowired
```
kchelstowski, line 38: I'm wondering if you need the Autowired annotation, and why you had to add `@Component` annotation in the `DateFormatterManager` - the reason I ask is that I was also used to the Spring way of defining beans and injecting them and it turned out that Grails detected them automically - moreover - I remember getting exceptions when adding `@Autowired` explicitly, especially when trying to inject beans via constructor.

src/main/groovy/org/pih/warehouse/core/date/DateFormatterContext.groovy
```diff
@@ -80,5 +88,10 @@ class DateFormatterContext {
             context.displayStyleOverride = displayStyleOverride
             return this
         }
+
+        DateFormatterContextBuilder withDefaultValue(String defaultValue) {
+            context.defaultValue = defaultValue
+            return this
+        }
```
kchelstowski, line 95: I enjoy reading those builder classes, even though this could probably be just handled by creating an instance like: 
```
new DateFormatterContext(
  localeOverride: X,
  displayFormat: Y,
  ...)
```

grails-app/domain/org/pih/warehouse/inventory/CycleCount.groovy
ewaterman: Instant is automatically mapped to DATETIME by JDBC.

How it works is that the Instant (which is seconds since epoch) is converted to a date time in the timezone specified in the JDBC url before storing, and then converted back to an Instant upon retrieval using that same timezone. So even though DATETIME has no timezone information, the implicit timezone is the one we specified (and if serverTimezone isn't specified, it uses server time). For us, this is always UTC because we have: `jdbc:mysql://localhost:3306/openboxes?serverTimezone=UTC`

We should probably be using TIMESTAMP for all our Date/Instant fields since it's better suited for moment-in-time types. It's not strictly required to use TIMESTAMP though for Instant from what I've seen, it's just more performant since there's less conversion that needs to happen.

src/main/groovy/org/pih/warehouse/core/date/DateFormatterContext.groovy
ewaterman: it's especially nice for when we want the `withX(X x)` behaviour to have a little bit of logic behind it or when we want to require that a group of fields all be specified together. It makes the constructors cleaner too

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
ewaterman: It's because the bean is defined under `/src` and not under `/grails-app`. I've been trying to reserve `/grails-app` only for the grails-specific component types like service, controller, taglib... and trying to put everything else outside of the grails space (and doing it the spring way). It's a bit annoying to have to mix and match annotations, but I'd rather not stick a component into grails just for the sake of convenience. I fear we're eventually going to have to move away from grails since support is getting slower and slower so whenever I can do things the spring way, I opt to do so.

and yeah if I remember correctly if you're injecting them via constructor you don't need the annotation. Actually I prefer defining beans via constructors. It makes it easier to mock behaviour in tests in certain scenarios since we can just pass in a stub at the time of constructing the bean.

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
ewaterman: Instant is for fields like: `2025-01-01T00:00:00.000Z`
LocalDate is for fields like: `2025-01-01`

We want to distinguish between when a date should be a moment in time that changes depending on the timezone you're in (Instant) vs just a date that doesn't change no matter what timezone you're in (LocalDate)

In this case, it's a date-only field:

Image

But thinking about it, it is a little weird since the date that was counted is timezone dependent (if I counted at 11 PM UTC-2, that's the next day for UTC).

Maybe it makes sense for this to be an Instant and have the "last counted date" date picker pick midnight for whatever day was selected in the user's timezone.

kchelstowski: Yeah I had the same feeling that in someone's timezone it could be a day earlier, and you would filter by just an another day.
Given this example I'm starting to wonder where we could rely on the `LocalDate` since we can ask the same question for every scenario in the system? :thinking:

grails-app/services/org/pih/warehouse/inventory/CycleCountService.groovy
kchelstowski: > and yeah if I remember correctly if you're injecting them via constructor you don't need the annotation

Injection via constructor has never worked for me in Grails :man_shrugging:  even though this should be the preferred way of injection

grails-app/domain/org/pih/warehouse/inventory/CycleCountCandidate.groovy
ewaterman: expiry date is probably the best example. If you look at any product with an expiry date (including food) it just lists a date as the expiry. The product doesn't expire at a specific moment in time. It's the same with estimated ship date. There's no time or timezone component, just an approximate day. Expiration dates are the biggest source of issues in openboxes currently. Since we currently treat expiry as a date at midnight utc, if you're on a timezone that is UTC-1 then the expiry suddenly becomes the day before, which is incorrect.

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
```diff
@@ -49,7 +51,7 @@ class CycleCountItemCommand implements Validateable {
 
     Person assignee
 
-    Date dateCounted
+    LocalDate dateCounted
```
kchelstowski, line 54: why don't we care about the exact time it was counted?

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemDto.groovy
```diff
@@ -32,9 +35,9 @@ class CycleCountItemDto {
 
     Boolean custom
 
-    Date dateCounted
+    LocalDate dateCounted
```
kchelstowski, line 38: same as above

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemCommand.groovy
awalkowiak: cc @ewaterman

src/main/groovy/org/pih/warehouse/inventory/CycleCountItemDto.groovy
awalkowiak: cc @ewaterman

PR #5536, opened 2025-10-07
OBPIH-7489 React date picker and utils using date fns (fix failing tests)

alannadolny:


alannadolny:
@ewaterman, it looks like tests are running on develop, not on this branch

PR #5541, opened 2025-10-09
OBPIH-7513 Add tooltips and replace old DatePickers in Inbound

SebastianLib:

https://pihemr.atlassian.net/browse/OBPIH-7513

**Description:**

---

### DIFF
--- a/src/js/components/form-elements/v2/DateFieldDateFns.jsx
+++ b/src/js/components/form-elements/v2/DateFieldDateFns.jsx
@@ -12,6 +12,7 @@ import componentType from 'consts/componentType';
 import { DateFormatDateFns, TimeFormat } from 'consts/timeFormat';
 import useFocusOnMatch from 'hooks/useFocusOnMatch';
 import useTranslate from 'hooks/useTranslate';
+import { formatDateToString } from 'utils/dateUtils';
 import InputWrapper from 'wrappers/InputWrapper';
 import RootPortalWrapper from 'wrappers/RootPortalWrapper';
 
@@ -37,6 +38,7 @@ const DateFieldDateFns = ({
   clearable,
   wrapperClassName,
   focusProps = {},
+  customTooltip,
   ...fieldProps
 }) => {
   const translate = useTranslate();
@@ -115,6 +117,12 @@ const DateFieldDateFns = ({
       button={button}
       hideErrorMessageWrapper={hideErrorMessageWrapper}
       className={wrapperClassName}
+      customTooltip={customTooltip}
+      value={formatDateToString({
+        date: value,
+        dateFormat: DateFormatDateFns.DD_MMM_YYYY,
+        options: { locale: dateFnsLocale() },
+      })}
     >
       <DatePicker
         {...fieldProps}
--- a/src/js/hooks/inboundV2/addItems/useInboundAddItemsColumns.jsx
+++ b/src/js/hooks/inboundV2/addItems/useInboundAddItemsColumns.jsx
@@ -6,11 +6,18 @@ import React, {
 } from 'react';
 
 import { createColumnHelper } from '@tanstack/react-table';
+import * as locales from 'date-fns/locale';
 import _ from 'lodash';
 import PropTypes from 'prop-types';
 import { Controller, useController } from 'react-hook-form';
 import { RiDeleteBinLine } from 'react-icons/ri';
 import { useSelector } from 'react-redux';
+import {
+  getCurrentLocale,
+  getDebounceTime,
+  getMinSearchLength,
+  getUsers,
+} from 'selectors';
 
 import { TableCell } from 'components/DataTable';
 import TableHeaderCell from 'components/DataTable/TableHeaderCell';
@@ -22,6 +29,7 @@ import StockMovementDirection from 'consts/StockMovementDirection';
 import { DateFormatDateFns } from 'consts/timeFormat';
 import useArrowsNavigation from 'hooks/useArrowsNavigation';
 import useTranslate from 'hooks/useTranslate';
+import { formatDateToString } from 'utils/dateUtils';
 import { debouncePeopleFetch, debounceProductsFetch } from 'utils/option-utils';
 
 const useInboundAddItemsColumns = ({
@@ -58,12 +66,12 @@ const useInboundAddItemsColumns = ({
     debounceTime,
     minSearchLength,
     users,
-    locale,
+    currentLocale,
   } = useSelector((state) => ({
-    debounceTime: state.session.searchConfig.debounceTime,
-    minSearchLength: state.session.searchConfig.minSearchLength,
-    users: state.users.data,
-    locale: state.session.activeLanguage,
+    debounceTime: getDebounceTime(state),
+    minSearchLength: getMinSearchLength(state),
+    users: getUsers(state),
+    currentLocale: getCurrentLocale(state),
   }));
   const debouncedProductsFetch = useCallback(
     debounceProductsFetch(
@@ -377,7 +385,11 @@ const useInboundAddItemsColumns = ({
           <TableCell
             className="rt-td rt-td-xs rt-td-add-items"
             customTooltip
-            tooltipLabel={errors?.[row.index]?.expirationDate?.message ?? value}
+            tooltipLabel={errors?.[row.index]?.expirationDate?.message ?? formatDateToString({
+              date: value,
+              dateFormat: DateFormatDateFns.DD_MMM_YYYY,
+              options: { locale: locales[currentLocale] },
+            })}
           >
             <Controller
               name={`values.lineItems.${row.index}.expirationDate`}
@@ -401,7 +413,6 @@ const useInboundAddItemsColumns = ({
                     setValue(`values.lineItems.${row.index}.expirationDate`, newDate);
                     await trigger();
                   }}
-                  customTooltip
                 />
               )}
             />
@@ -553,7 +564,7 @@ const useInboundAddItemsColumns = ({
     debounceTime,
     minSearchLength,
     users,
-    locale,
+    currentLocale,
     headerRecipient,
     rowIndex,
     columnId,
--- a/src/js/hooks/inboundV2/create/useInboundCreateForm.js
+++ b/src/js/hooks/inboundV2/create/useInboundCreateForm.js
@@ -5,6 +5,7 @@ import queryString from 'query-string';
 import { useForm } from 'react-hook-form';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory, useLocation } from 'react-router-dom';
+import { getCurrentLocation } from 'selectors';
 
 import { fetchUsers, updateWorkflowHeader } from 'actions';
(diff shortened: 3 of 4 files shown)


src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -36,10 +37,17 @@ const DateFieldDateFns = ({
   clearable,
   wrapperClassName,
   focusProps = {},
+  triggerValidation,
```
kchelstowski: I don't like that. Why can't we follow the pattern from the old date picker to just pass `onChangeRaw` and then in the `DatePicker` here, just add `onChange` prop like we did in the old version?

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -45,13 +45,13 @@ const InboundSendForm = ({ previous }) => {
   // button is disabled when there are form errors
   // or when the shipment status is not dispatched
   // or selected destination doesn't match current location
-  const shouldDisableRollbackButton =
-    !isValid || statusCode !== requisitionStatus.DISPATCHED || !matchesDestination;
+  const shouldDisableRollbackButton = !isValid || statusCode !== requisitionStatus.DISPATCHED
```
kchelstowski: I must have missed it in some past code reviews.
Please change it to `isRollbackDisabled` or rather not to negate the const names: `isRollbackEnabled` - I assume if rollback is enabled, it means the button is enabled, so we can get rid of the "Button" word as well

src/js/components/form-elements/v2/TextInput.jsx
```diff
@@ -28,6 +28,8 @@ const TextInput = ({
   onKeyDown,
   focusProps = {},
   onWheel,
+  customTooltip,
```
kchelstowski, line 31: what was the purpose of the custom tooltip? How does it differ from the regular tooltip?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -161,7 +164,10 @@ const useInboundSendForm = ({ previous }) => {
         trackingNumber: data.trackingNumber ?? '',
         driverName: data.driverName ?? '',
         comments: data.comments ?? '',
-        expectedDeliveryDate: data.expectedDeliveryDate ?? null,
+        expectedDeliveryDate: formatDateToString({
+          date: data.expectedDeliveryDate,
+          dateFormat: DateFormatDateFns.DD_MMM_YYYY,
+        }) ?? null,
```
kchelstowski: is the `??` needed though? I can see the `formatDateToString` can only (already) return `null` the date is falsy, or `format` is supposed to return string.

src/js/components/form-elements/v2/TextInput.jsx
SebastianLib: A regular tooltip is just the 'i' icon where we can show some information, but we probably don't want that.

image


customTooltip displays the value that the user choose or typed:
image

src/js/components/form-elements/v2/DateFieldDateFns.jsx
SebastianLib: I did it this way because now we trigger validation in two functions: **onChange** and **onClear**. In the old date picker, function onChangeRaw didn’t fire when the date was cleared, so validation didn’t work correctly. Kuba noticed this during testing and asked for the change. I think this approach is cleaner and works better than the old DatePicker. @alannadolny  what do you think?

alannadolny: This prop shouldn't be passed here. It's the same datepicker that we had before; the library for handling date parsing is the only difference.

src/js/components/form-elements/v2/TextInput.jsx
alannadolny: Something feels wrong to me here. We have the same tooltip on cycle count. How is it working without that prop in cc?

src/js/hooks/inboundV2/send/useInboundSendForm.js
```diff
@@ -149,7 +149,10 @@ const useInboundSendForm = ({ previous }) => {
             label: `${data.destination.name} [${data.destination.locationType?.description ?? ''}]`,
           }
           : null,
-        shipDate: data.dateShipped ?? null,
+        shipDate: formatDateToString({
+          date: data.dateShipped,
+          dateFormat: DateFormatDateFns.DD_MMM_YYYY,
+        }),
```
alannadolny: If you are using some arguments in a function often, you can create curried functions:
```
const formatDate = (date) => (formatDateToString({
   date,
   dateFormat: DateFormatDateFns.DD_MMM_YYYY
}))
```

src/js/components/form-elements/v2/TextInput.jsx
SebastianLib: In the cycle count we already use the new tooltip, but that’s because it’s passed through components like TableHeaderCell, TableCell from the DataTable. There are cases where we don’t use DataTable to display fields, for example in steps 1 and 3 of inbound, and in those cases we need to display the tooltip in the InputWrapper itself, since it doesn’t have the new tooltip yet.

src/js/components/form-elements/v2/DateFieldDateFns.jsx
SebastianLib: The point here is that validation wasn’t triggered in the old DatePicker in the inbound flow when we cleared the date, because onChangeRaw didn’t detect it. That’s why I thought this approach would be the best one in the new DatePicker. I’ll try to find another way to handle it.

SebastianLib: I’ve found the root cause of the issue! I can handle everything within the onChange handler. Previously it wasn’t working, and I couldn’t figure out why, but it turned out that {...field} was overwriting my onChange :facepalm: :facepalm: :facepalm:  example:                  
```
                <DateFieldDateFns
                  title={{
                    id: 'react.stockMovement.dateRequested.label',
                    defaultMessage: 'Date Requested',
                  }}
                  errorMessage={errors.dateRequested?.message}
                  required
                  customDateFormat={DateFormatDateFns.DD_MMM_YYYY}
                  customTooltip
                  triggerValidation={trigger}
                  {...field}
                />
```
I came up with that solution, but I’ll clean it up now!

kchelstowski: @SebastianLib passing props via `{...props}` should be avoid. Even if you have to use it, never use it at the bottom as it might override props specified above it, but use it at the very top. If you had seen any places where we do the opposite, let us know.

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -36,10 +38,13 @@ const DateFieldDateFns = ({
   clearable,
   wrapperClassName,
   focusProps = {},
+  customTooltip,
   ...fieldProps
 }) => {
   const translate = useTranslate();
-  const onClear = () => onChange(null);
+  const onClear = async () => {
```
kchelstowski: why is async needed here? :open_mouth:

src/js/components/form-elements/v2/DateFieldDateFns.jsx
```diff
@@ -79,7 +84,7 @@ const DateFieldDateFns = ({
 
   const datePickerRef = useRef(null);
 
-  const onChangeHandler = (date) => {
+  const onChangeHandler = async (date) => {
```
kchelstowski: same as above - as @alannadolny said, it was an old datepicker just using a new library, why suddenly asyncs are needed here?

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
```diff
@@ -83,7 +85,7 @@ const InboundV2 = () => {
     isValid: addItemsIsValid,
     trigger: addItemsTrigger,
     getValues: addItemsGetValues,
-    setValue: addItemsValue,
+    setValue: addItemsSetValue,
```
kchelstowski, line 88: haven't caught it earlier - why do we actually call two hooks in the "root" component `InboundV2` instead of calling them in the actual components where they are needed?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
alannadolny: Why is this function async?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
alannadolny: And this one also?

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -110,19 +112,20 @@ const InboundSendForm = ({ previous }) => {
                 control={control}
                 disabled={!matchesDestination}
                 render={({ field }) => (
-                  <DateField
+                  <DateFieldDateFns
+                    {...field}
```
alannadolny, line 116: Are you sure it won't break anything?

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
```diff
@@ -208,19 +215,21 @@ const InboundSendForm = ({ previous }) => {
...
```
alannadolny, line 219: and here?

src/js/components/form-elements/v2/DateFieldDateFns.jsx
SebastianLib: I must have missed it and forgot to remove the async when I made the changes on friday :facepalm:

SebastianLib: I must have missed it and forgot to remove the async when I made the changes on friday :facepalm:

src/js/components/stock-movement-wizard/inboundV2/InboundV2.jsx
SebastianLib: I also noticed that and wrote about it in my document "Tech design: Planning delivery of inbound refactor". I will probably change it in the next ticket, where I will improve the performance of export items https://pihemr.atlassian.net/browse/OBPIH-7197

src/js/components/stock-movement-wizard/inboundV2/sections/InboundSendForm.jsx
SebastianLib: From what I tested, everything works fine

src/js/components/form-elements/v2/DateFieldDateFns.jsx
SebastianLib: I must have missed it and forgot to remove the async when I made the changes on friday 🤦

kchelstowski (approved):

